Issue #1669, opened 2020-02-25
Detect app type and mode and disable proper devtools pages


kenzieschmoll:
We do this in the dart:html version and need to port this functionality to the Flutter version of DevTools

Certain pages need to be disabled for profile builds as well as for different types of apps (e.g. vm apps)

PR #1757, opened 2020-03-20
Disable DevTools pages based on the connected app type.

kenzieschmoll:
When the page is not supported for the connected app, we show a full screen message explaining why the page is disabled.

Screen Shot 2020-03-20 at 10 40 00 AM

Fixes https://github.com/flutter/devtools/issues/1669

### DIFF
--- a/packages/devtools_app/lib/src/connected_app.dart
+++ b/packages/devtools_app/lib/src/connected_app.dart
@@ -61,7 +61,7 @@ class ConnectedApp {
 
     await serviceManager.serviceExtensionManager.extensionStatesUpdated.future;
 
-    // The debugAllowBanner extension is only available in debug builds.
+    // The debugAllowBanner extension is only available in debug builds
     final hasDebugExtension = serviceManager.serviceExtensionManager
         .isServiceExtensionAvailable(extensions.debugAllowBanner.extension);
     return !hasDebugExtension;
--- a/packages/devtools_app/lib/src/connected_app.dart
+++ b/packages/devtools_app/lib/src/connected_app.dart
@@ -16,7 +16,7 @@ class ConnectedApp {
   bool get appTypeKnown => _appTypeKnown;
   bool _appTypeKnown = false;
 
-  FutureOr<bool> get isFlutterApp async {
+  Future<bool> get isFlutterApp async {
     _isFlutterApp ??= await _libraryUriAvailable(flutterLibraryUri);
     _updateAppTypeKnown();
     return _isFlutterApp;
@@ -29,7 +29,7 @@ class ConnectedApp {
 
   bool _isFlutterApp;
 
-  FutureOr<bool> get isProfileBuild async {
+  Future<bool> get isProfileBuild async {
     _isProfileBuild ??= await _connectedToProfileBuild();
     _updateAppTypeKnown();
     return _isProfileBuild;
@@ -42,7 +42,7 @@ class ConnectedApp {
 
   bool _isProfileBuild;
 
-  FutureOr<bool> get isDartWebApp async {
+  Future<bool> get isDartWebApp async {
     _isDartWebApp ??= await _libraryUriAvailable(dartHtmlLibraryUri);
     _updateAppTypeKnown();
     return _isDartWebApp;
@@ -63,7 +63,7 @@ class ConnectedApp {
 
   bool get isRunningOnDartVM => serviceManager.vm.name != 'ChromeDebugProxy';
 
-  FutureOr<bool> get isDartCliApp async =>
+  Future<bool> get isDartCliApp async =>
       isRunningOnDartVM && !(await isFlutterApp);
 
   bool get isDartCliAppRaw => isRunningOnDartVM && !isFlutterAppRaw;
--- a/packages/devtools_app/lib/src/connected_app.dart
+++ b/packages/devtools_app/lib/src/connected_app.dart
@@ -13,60 +13,45 @@ const dartHtmlLibraryUri = 'dart:html';
 class ConnectedApp {
   ConnectedApp();
 
-  bool get appTypeKnown => _appTypeKnown;
-  bool _appTypeKnown = false;
+  bool get appTypeKnown =>
+      _isFlutterApp != null && _isProfileBuild != null && _isDartWebApp != null;
 
-  Future<bool> get isFlutterApp async {
-    _isFlutterApp ??= await _libraryUriAvailable(flutterLibraryUri);
-    _updateAppTypeKnown();
-    return _isFlutterApp;
-  }
+  Future<bool> get isFlutterApp async =>
+      _isFlutterApp ??= await _libraryUriAvailable(flutterLibraryUri);
 
-  bool get isFlutterAppRaw {
+  bool get isFlutterAppNow {
     assert(_isFlutterApp != null);
     return _isFlutterApp;
   }
 
   bool _isFlutterApp;
 
-  Future<bool> get isProfileBuild async {
-    _isProfileBuild ??= await _connectedToProfileBuild();
-    _updateAppTypeKnown();
-    return _isProfileBuild;
-  }
+  Future<bool> get isProfileBuild async =>
+      _isProfileBuild ??= await _connectedToProfileBuild();
 
-  bool get isProfileBuildRaw {
+  bool get isProfileBuildNow {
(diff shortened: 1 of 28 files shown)


packages/devtools_app/test/flutter/performance_screen_test.dart
```diff
@@ -20,9 +22,10 @@ void main() {
   PerformanceScreen screen;
   FakeServiceManager fakeServiceManager;
 
-  group('TimelineScreen', () {
+  group('PerformanceScreen', () {
```
devoncarew: 👍

packages/devtools_app/lib/src/flutter/screen.dart
```diff
@@ -74,6 +74,9 @@ abstract class Screen {
   Widget buildStatus(BuildContext context, TextTheme textTheme) {
     return null;
   }
+
+  /// Whether the screen is enabled.
```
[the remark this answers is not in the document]
devoncarew: Or, given that the logic for whether it's enabled (and why not) currently needs to be spread across the enabled() and build() methods, perhaps drop the enabled() method, and just rely on if statements in the build() method? I don't know where else in the app we'd call enabled() from.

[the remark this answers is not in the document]
kenzieschmoll: Moved `enabled` logic into build methods

packages/devtools_app/lib/src/info/flutter/info_screen.dart
```diff
@@ -25,7 +26,14 @@ class InfoScreen extends Screen {
   bool get showIsolateSelector => true;
 
   @override
-  Widget build(BuildContext context) => InfoScreenBody();
+  Widget build(BuildContext context) {
+    return enabled()
```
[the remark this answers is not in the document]
kenzieschmoll: moved enabled logic into build method. `serviceManager` is the only place from which `connectedApp` is accessible. `serviceManager` is a global so we can access it from anywhere without having to deal with inherited widgets.

packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
```diff
@@ -33,7 +34,22 @@ class InspectorScreen extends Screen {
...
   @override
-  Widget build(BuildContext context) => const InspectorScreenBody();
+  Widget build(BuildContext context) {
+    if (!enabled()) {
+      if (!serviceManager.connectedApp.isFlutterAppRaw) {
+        return const DisabledForNonFlutterAppMessage();
+      }
+      if (serviceManager.connectedApp.isProfileBuildRaw) {
+        return const DisabledForProfileModeMessage();
+      }
```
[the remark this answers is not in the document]
kenzieschmoll: There will not be a fall back. The only two cases that would make `enabled()` return false are covered by the if statements. This is more clear now that the enabled logic is directly in the build method.

packages/devtools_app/lib/src/info/flutter/info_screen.dart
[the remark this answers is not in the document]
devoncarew: (though likely not necessary as it looks like we keep app info off the global serviceManager)

packages/devtools_app/lib/src/connected_app.dart
```diff
@@ -8,36 +8,66 @@ import 'globals.dart';
...
+  void _updateAppTypeKnown() {
+    _appTypeKnown = _isFlutterApp != null &&
+        _isProfileBuild != null &&
+        _isDartWebApp != null;
+  }
+
   bool get isRunningOnDartVM => serviceManager.vm.name != 'ChromeDebugProxy';
 
-  Future<bool> get isDartCliApp async =>
+  FutureOr<bool> get isDartCliApp async =>
```
devoncarew: I think `Future<bool>` is correct here.

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -101,12 +101,7 @@ class DevToolsAppState extends State<DevToolsApp> {
       settings: settings,
       builder: (BuildContext context) {
         return DevToolsScaffold.withChild(
-          child: Center(
-            child: Text(
-              'Sorry, $uri was not found.',
-              style: Theme.of(context).textTheme.headline4,
-            ),
-          ),
+          child: CenteredMessage('Sorry, $uri was not found.'),
```
devoncarew: We should definitely include a version of the chrome 404 dinosaur game here: https://lmgtfy.com/?q=chrome+404+dinosaur+game.

packages/devtools_app/lib/src/connected_app.dart
kenzieschmoll: since isFlutterApp returns a FutureOr, wouldn't it be FutureOr?

devoncarew: I think they can both be changed to `Future`?

It's not a critical change.

packages/devtools_app/lib/src/connected_app.dart
```diff
@@ -8,36 +8,66 @@ import 'globals.dart';
...
-  Future<bool> get isPackageFlutterWeb async =>
-      _isPackageFlutterWeb ??= await _libraryUriAvailable(flutterWebLibraryUri);
-  bool _isPackageFlutterWeb;
+  Future<bool> get isProfileBuild async {
+    _isProfileBuild ??= await _connectedToProfileBuild();
+    _updateAppTypeKnown();
+    return _isProfileBuild;
+  }
+
+  bool get isProfileBuildRaw {
```
jacob314: `RemoteDiagnosticsNode` uses the convention of adding the suffix `Now` instead of  `Raw` to indicate getters that return a value immediately as opposed to matching getters that return a Future. I'm also open to using a different naming convention but if we do, we should be consistent across the application.

packages/devtools_app/lib/src/connected_app.dart
```diff
@@ -8,36 +8,66 @@ import 'globals.dart';
...
+  bool get isDartWebAppRaw {
+    assert(_isDartWebApp != null);
+    return _isDartWebApp;
+  }
 
-  Future<bool> get isDartWebApp async =>
-      _isDartWebApp ??= await _libraryUriAvailable(dartHtmlLibraryUri);
   bool _isDartWebApp;
 
+  void _updateAppTypeKnown() {
```
jacob314: this is cheap enough that 
_appTypeKnown could just be a getter with this expression rather than bothering with update logic.

packages/devtools_app/lib/src/connected_app.dart
```diff
@@ -55,10 +85,16 @@ class ConnectedApp {
...
         .map((ref) => ref.uri)
         .toList()
         .any((u) => u.startsWith(uri));
   }
+
+  Future<void> initializeValues() async {
+    await isFlutterApp;
+    await isProfileBuild;
+    await isDartWebApp;
+    _appTypeKnown = true;
```
jacob314: why is this set to true here as well as using the _updateAppTypeKnown getter in other cases?
Consider executing these three async requests in parallel for slightly faster startup.
E.g.
`await Futures.wait([isFlutterApp, isProfileBuild, isDartWebApp]);`
so that all vm requests are triggered simultaneously. would only really matter if remote debugging.

packages/devtools_app/test/flutter/network_screen_test.dart
```diff
@@ -0,0 +1,42 @@
...
+import 'package:mockito/mockito.dart';
+
+import '../support/mocks.dart';
+import 'wrappers.dart';
+
+void main() {
+  NetworkScreen screen;
+  FakeServiceManager fakeServiceManager;
+
+  group('NetworkScreen', () {
```
jacob314: nice tests!

packages/devtools_app/lib/src/connected_app.dart
kenzieschmoll: changed to now

packages/devtools_app/lib/src/connected_app.dart
kenzieschmoll: done

packages/devtools_app/lib/src/connected_app.dart
kenzieschmoll: done

PR #1758, opened 2020-03-20
Rename uses of 'gpu' to 'raster'.

kenzieschmoll:
Per, https://github.com/flutter/flutter/issues/29443, this CL updates the DevTools UI and code to use 'raster' instead of 'gpu'. This CL just includes renames, and leaves the html code alone where possible since we will be deleting it soon.

### DIFF
--- a/packages/devtools_app/lib/src/timeline/html_frames_bar_chart.dart
+++ b/packages/devtools_app/lib/src/timeline/html_frames_bar_chart.dart
@@ -207,7 +207,7 @@ class PlotlyDivGraph extends CoreElement {
       // TODO(terry): HACK - Ignore the event.
       log(
         'Ignored onFrameAdded - bad data.\n [uiDuration: '
-        '${frame.uiDuration}, gpuDuration: ${frame.gpuDuration}',
+        '${frame.uiDuration}, gpuDuration: ${frame.rasterDuration}',
         LogLevel.warning,
       );
     }
--- a/packages/devtools_app/lib/src/timeline/html_frames_bar_chart.dart
+++ b/packages/devtools_app/lib/src/timeline/html_frames_bar_chart.dart
@@ -191,7 +191,7 @@ class PlotlyDivGraph extends CoreElement {
 
   // Add current frame data to chunks of data for later plotting.
   void processNextFrame() async {
-    final frame =
+    final TimelineFrame frame =
         timelineController.frameBasedTimeline.frameAddedNotifier.value;
     if (frame == null) return;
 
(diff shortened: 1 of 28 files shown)


packages/devtools_app/lib/src/timeline/html_frames_bar_chart.dart
```diff
@@ -207,7 +207,7 @@ class PlotlyDivGraph extends CoreElement {
       // TODO(terry): HACK - Ignore the event.
       log(
         'Ignored onFrameAdded - bad data.\n [uiDuration: '
-        '${frame.uiDuration}, gpuDuration: ${frame.gpuDuration}',
+        '${frame.uiDuration}, gpuDuration: ${frame.rasterDuration}',
```
devoncarew: `gpuDuration` ==> `rasterDuration`?

packages/devtools_app/lib/src/timeline/html_frames_bar_plotly.dart
```diff
@@ -174,7 +174,7 @@ class FramesBarPlotly {
           ),
         ),
         marker: Marker(
-          color: colorToCss(mainGpuColor),
+          color: colorToCss(mainRasterColor),
```
devoncarew: Since this is in an html file, can you confirm that the timeline still works in the dart:html verison after this change?  That we're not doing something like renaming a string which we also reference (by name) from a css file or something?

packages/devtools_app/lib/src/trace_event.dart
```diff
@@ -104,8 +104,8 @@ class TraceEvent {
     if (_type == null) {
       if (args[typeKey] == 'ui') {
         _type = TimelineEventType.ui;
-      } else if (args[typeKey] == 'gpu') {
-        _type = TimelineEventType.gpu;
+      } else if (args[typeKey] == 'raster') {
```
devoncarew: Where are we getting `args[typeKey]` from? Is this data from the VM / engine, and will it have been updated to have the new name? If it is from the engine, perhaps check for both `gpu` and `raster`?

packages/devtools_app/lib/src/timeline/html_frames_bar_chart.dart
kenzieschmoll: whoops - this change wasn't supposed to be there. Removed. We didn't have a type on `frame` so the analyzer didn't check it.

packages/devtools_app/lib/src/trace_event.dart
kenzieschmoll: We asked the engine team a while back if they could add a UI/GPU distinction bit to the trace args, but they didn't have time to implement this.  So we actually never receive any events with a 'type' arg.  Probably best if we remove this all together.

packages/devtools_app/lib/src/timeline/html_frames_bar_plotly.dart
kenzieschmoll: Tested html app - all is WAI 👍  (after https://github.com/flutter/devtools/pull/1760 lands)

Issue #1711, opened 2020-03-10
Display warning messages when cpu and memory profilers are used in debug messages


jacob314:
Related issue to hide debug only pages in release mode.
https://github.com/flutter/devtools/issues/1669

PR #1764, opened 2020-03-25
Add banner message warnings and errors

kenzieschmoll:
Timeline and Performance pages:
Screen Shot 2020-03-25 at 4 21 58 PM
Screen Shot 2020-03-25 at 4 25 36 PM

Memory page:
Screen Shot 2020-03-25 at 4 22 10 PM
Screen Shot 2020-03-25 at 4 31 56 PM

Once a message is dismissed on a page, it will not be shown again on that page for a running DevTools instance. We should eventually add "Do not show this message again" functionality for warnings, and there is a TODO for that.

Fixes https://github.com/flutter/devtools/issues/1711


packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -0,0 +1,304 @@
...
+
+  Stream<Null> get onRefreshMessages => _refreshMessagesController.stream;
+
+  @override
+  void dispose() {
+    _refreshMessagesController.close();
+  }
+
+  void refreshMessages() {
+    _refreshMessagesController.add(null);
```
jacob314: the body of this method doesn't seem related to refreshing. can we make this a bit more idiomatic?

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -0,0 +1,304 @@
...
+
+    autoDispose(controller.onRefreshMessages.listen((_) => setState(() {})));
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final messagesForScreen = widget.screen.messages(context);
+    if (messagesForScreen.isNotEmpty) {
+      final messagesToShow = <Widget>[];
+      for (Widget message in messagesForScreen) {
```
jacob314: Widget --> var

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -0,0 +1,304 @@
...
+        assert(message is UniqueMessage);
```
jacob314: if we only handle UniqueMessage objects, change the type to
`List<UniqueMessage>`

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -0,0 +1,304 @@
...
+              child: Container(
+                child: RichText(
+                  text: TextSpan(
+                    children: textSpans,
+                  ),
+                ),
+              ),
+            ),
+            const SizedBox(width: denseSpacing),
+            CircularIconButton(
```
jacob314: this icon looks quite large. is that intentional?

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -0,0 +1,304 @@
...
+          foregroundColor: foreground,
+        );
+
+  static const foreground = Colors.white;
+  static const linkColor = Color(0xFF88D7FC);
+}
+
+// TODO(kenz): add "Do not show this again" option to warnings.
+class _BannerWarning extends BannerMessage {
+  const _BannerWarning({
```
jacob314: the warning banner looks good on a light theme but is too loud on a dark theme. Can we use the theme to get the foreground and background colors for this case or use a themed color with light and dark options.

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -0,0 +1,304 @@
...
+  String get id => 'DebugModePerformanceMessage - $screenType';
+
+  @override
+  Widget build(BuildContext context) {
+    return _BannerError(
+      id: id,
+      textSpans: [
+        const TextSpan(
+          text:
+              'You are running your app in debug mode. Debug mode performance '
```
jacob314: nit: this is probably cleaner as a ''' literal. You wouldn't need to escape anything.

packages/devtools_app/lib/src/flutter/banner_messages.dart
jacob314: you'll need to change UniqueMessage to a MessageWidget base class that has an id.
UniqueMessage is a bit of a confusing to me as it doesn't have any members related to being a message so that is probably a good thing.

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -0,0 +1,304 @@
...
+  Widget build(BuildContext context) {
+    final bannerMessagesController = Controllers.of(context).bannerMessages;
+    return Card(
+      color: backgroundColor,
+      child: Padding(
+        padding: const EdgeInsets.all(defaultSpacing),
+        child: Row(
+          children: <Widget>[
+            Expanded(
+              child: Container(
```
jacob314: @pq a lint should be able to catch that this Container widget can be removed.

packages/devtools_app/lib/src/flutter/common_widgets.dart
```diff
@@ -459,3 +459,43 @@ class CenteredMessage extends StatelessWidget {
     );
   }
 }
+
+class CircularIconButton extends StatelessWidget {
```
jacob314: isn't there an existing material design close button we can use instead? this button seems a bit large and not that consistent with material design.

packages/devtools_app/lib/src/flutter/scaffold.dart
```diff
@@ -191,11 +192,18 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
     // Build the screens for each tab and wrap them in the appropriate styling
     final tabBodies = [
       for (var screen in widget.tabs)
-        Align(
-          alignment: Alignment.topLeft,
-          child: Padding(
-            padding: DevToolsScaffold.appPadding,
-            child: screen.build(context),
+        Padding(
```
jacob314: @pq 
Padding + Align should trigger a lint to use Container.

packages/devtools_app/lib/src/flutter/scaffold.dart
```diff
@@ -191,11 +192,18 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
...
+          padding: DevToolsScaffold.appPadding,
+          child: Align(
+            alignment: Alignment.topLeft,
+            child: Column(
+              children: <Widget>[
```
jacob314: here and elsewhere omit <Widget> as it can be inferred.

packages/devtools_app/lib/src/flutter/screen.dart
```diff
@@ -68,6 +68,8 @@ abstract class Screen {
   /// Builds the body to display for this tab.
   Widget build(BuildContext context);
 
+  List<Widget> messages(BuildContext context) => [];
```
jacob314: change this type to indicate it is a list of MessageWidget objects with ids.

packages/devtools_app/lib/src/flutter/theme.dart
```diff
@@ -72,6 +72,13 @@ const devtoolsBlue = ColorSwatch<int>(600, {
   400: Color(0xFF13B9FD),
 });
 
+// Error color selected from
+// https://material.io/design/color/#color-theme-creation.
+const devtoolsError = Color(0xFFB00020);
+
+// Yellow 100 from https://material.io/design/color/#color-theme-creation.
+const devtoolsWarning = Color(0xFFFFF9C4);
```
jacob314: this color seems too bright for the dark theme.

packages/devtools_app/lib/src/performance/flutter/performance_screen.dart
```diff
@@ -39,6 +41,31 @@ class PerformanceScreen extends Screen {
...
+
+  @override
+  List<Widget> messages(BuildContext context) {
+    final showDebugModeWarning = serviceManager.connectedApp.isFlutterAppNow &&
+        !serviceManager.connectedApp.isProfileBuildNow;
+    final performanceController = Controllers.of(context).performance;
+    final showProfileGranularityWarning = performanceController
+            .cpuProfilerController
+            .service
+            .profileGranularityFlagNotifier
```
jacob314: it would be cleaner to listen for the value of profileGranularityFlagNotifier
instead of using the current value.

packages/devtools_app/lib/src/timeline/flutter/timeline_screen.dart
```diff
@@ -55,6 +57,31 @@ class TimelineScreen extends Screen {
...
+        !serviceManager.connectedApp.isProfileBuildNow;
+    final timelineController = Controllers.of(context).timeline;
+    final showProfileGranularityWarning = timelineController
+            .cpuProfilerController
+            .service
+            .profileGranularityFlagNotifier
+            .value
+            .valueAsString ==
+        ProfileGranularity.high.value;
+    return [
```
jacob314: this is duplicated code from the Performance  screen. Refactor to add a base class or a helper method.

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -0,0 +1,304 @@
...
+const _runInProfileModeDocsUrl =
+    'https://flutter.dev/docs/testing/ui-performance#run-in-profile-mode';
+
+const _profileGranularityDocsUrl =
+    'https://flutter.dev/docs/development/tools/devtools/performance#profile-granularity';
+
+class BannerMessagesController implements DisposableController {
+  final _dismissedMessageIds = <String>{};
+
+  final _refreshMessagesController = StreamController<Null>.broadcast();
```
jacob314: It would be nice to do something more idiomatic than a stream with only null elements. 
For example, a ValueListenable<List<IdMessage>> or a ChangeNotifier could be cleaner.
https://api.flutter.dev/flutter/foundation/ChangeNotifier-class.html

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -0,0 +1,304 @@
...
+    final messagesForScreen = widget.screen.messages(context);
+    if (messagesForScreen.isNotEmpty) {
+      final messagesToShow = <Widget>[];
+      for (Widget message in messagesForScreen) {
+        assert(message is UniqueMessage);
+        if (!controller.isMessageDismissed((message as UniqueMessage).id)) {
+          messagesToShow.add(message);
+        }
+      }
+      if (messagesToShow.isNotEmpty) {
```
jacob314: it would be slick to use AnimatedList so messages animate in rather than just appearing.
https://www.youtube.com/watch?v=ZtfItHwFlZ8

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -0,0 +1,304 @@
...
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final messagesForScreen = widget.screen.messages(context);
+    if (messagesForScreen.isNotEmpty) {
+      final messagesToShow = <Widget>[];
+      for (Widget message in messagesForScreen) {
+        assert(message is UniqueMessage);
+        if (!controller.isMessageDismissed((message as UniqueMessage).id)) {
```
jacob314: Unless you want an id for analytics in the future, you could just use widgets with Key objects.

jacob314:
High level I wonder if we can align the APIs for this with the Notifications.of and similar APIs.

packages/devtools_app/lib/src/flutter/theme.dart
kenzieschmoll: switched to colors from @raison00

packages/devtools_app/lib/src/performance/flutter/performance_screen.dart
kenzieschmoll: done

packages/devtools_app/lib/src/flutter/banner_messages.dart
kenzieschmoll: added todo

packages/devtools_app/lib/src/flutter/banner_messages.dart
kenzieschmoll: done - using inherited widget

packages/devtools_app/lib/src/flutter/banner_messages.dart
kenzieschmoll: done

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -68,62 +101,76 @@ class _BannerMessageContainerState extends State<BannerMessageContainer>
     final newController = Controllers.of(context)?.bannerMessages;
     if (newController == controller) return;
     controller = newController;
+  }
+
+  void push(BannerMessage message) {
+    if (controller.isMessageDismissed(message) ||
```
jacob314: this logic should be in the controller instead of here. nobody should ever accidentally add a dismissed message or an already visible message to the controller.

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -68,62 +101,76 @@ class _BannerMessageContainerState extends State<BannerMessageContainer>
...
+      if (!mounted) return;
+      setState(() {
+        controller.addMessage(message);
+      });
+    });
+  }
 
-    autoDispose(controller.onRefreshMessages.listen((_) => setState(() {})));
+  void remove(BannerMessage message, {bool dismiss = false}) {
+    // We push the banner message in a post frame callback because otherwise,we'd be
```
jacob314: missing space after comma

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -68,62 +101,76 @@ class _BannerMessageContainerState extends State<BannerMessageContainer>
...
+
+  void push(BannerMessage message) {
+    if (controller.isMessageDismissed(message) ||
+        controller.isMessageVisible(message)) return;
+    // We push the banner message in a post frame callback because otherwise,we'd be
+    // trying to call setState while the parent widget `BannerMessages` is in the middle
+    // of `build`.
+    WidgetsBinding.instance.addPostFrameCallback((_) {
+      if (!mounted) return;
+      setState(() {
```
jacob314: seems surprising to have a setState around a call to a controller. setState calls should be when the direct state of this widget changes. can you instead have this widget listen for when the messages associated with its screen id in the controller change?

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -68,62 +101,76 @@ class _BannerMessageContainerState extends State<BannerMessageContainer>
...
 
-    autoDispose(controller.onRefreshMessages.listen((_) => setState(() {})));
+  void remove(BannerMessage message, {bool dismiss = false}) {
+    // We push the banner message in a post frame callback because otherwise,we'd be
+    // trying to call setState while the parent widget `BannerMessages` is in the middle
+    // of `build`.
+    WidgetsBinding.instance.addPostFrameCallback((_) {
+      if (!mounted) return;
+      setState(() {
+        controller.removeMessage(message, dismiss: dismiss);
```
jacob314: as above. having a set state around a call to a controller is a sign this widget isn't listening for changes to the controller. 
For example, this is broken if you happen to have two banner message widgets listening on the same controller.

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -68,62 +101,76 @@ class _BannerMessageContainerState extends State<BannerMessageContainer>
...
-          children: messagesToShow,
-        );
-      }
-    }
-    return const SizedBox();
+    return _InheritedBannerMessages(
+      data: this,
+      child: Column(
+        children: [
+          ...controller?.messagesForScreen(widget.screen.type) ?? [],
```
jacob314: refactoring this to
```dart
Column(
  children: [
    Column(
      children: ValueListenableBuilder(controller?.messagesForScreen(widget.screen.type),
    ),
    Expanded(...),
)
```
will fix your issue with having to do a post frame callback and will make things work correctly if someone else uses the banner controller.

packages/devtools_app/lib/src/ui/flutter/vm_flag_widgets.dart
```diff
@@ -42,6 +47,23 @@ class ProfileGranularityDropdownState
         if (safeValue != flag.valueAsString) {
           _onProfileGranularityChanged(safeValue, context);
         }
+
+        final bannerMessages = BannerMessages.of(context);
+        if (safeValue == highProfilePeriod) {
+          bannerMessages.push(
+              HighProfileGranularityMessage(widget.screenType).build(context));
+        } else {
+          final currentMessages =
```
jacob314: this business logic should be in the BannerMessages controller and matching should not need to be done by calling toString on keys.

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -21,44 +19,79 @@ const _runInProfileModeDocsUrl =
 const _profileGranularityDocsUrl =
     'https://flutter.dev/docs/development/tools/devtools/performance#profile-granularity';
 
-class BannerMessagesController implements DisposableController {
-  final _dismissedMessageIds = <String>{};
+class BannerMessagesController {
+  final _messages = <DevToolsScreenType, List<BannerMessage>>{};
```
jacob314: make the values in this map
ValueNotifier<List<BannerMessage>>
and be sure to call 
notifyListeners on the ValueNotifier every time you change the content of the list.
Then users of this controller can be notified when the controller state changes instead of having to hope they call setState at the right time.

kenzieschmoll: the reason I didn't do this originally is because notifyListeners is protected
Screen Shot 2020-03-27 at 7 28 22 AM

packages/devtools_app/lib/src/flutter/banner_messages.dart
kenzieschmoll: doing this still results in the `setState() or markNeedsBuild() called during build.` error.

packages/devtools_app/lib/src/flutter/banner_messages.dart
kenzieschmoll: done

packages/devtools_app/lib/src/ui/flutter/vm_flag_widgets.dart
kenzieschmoll: done

packages/devtools_app/lib/src/flutter/banner_messages.dart
kenzieschmoll: switched to using value listenable

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -20,36 +21,61 @@ const _profileGranularityDocsUrl =
...
+    final messageWithKey = currentMessages.value.firstWhere(
+      (m) => m.key == key,
+      orElse: () => null,
+    );
+    if (messageWithKey != null) {
+      removeMessage(messageWithKey);
+    }
   }
 
   bool isMessageDismissed(BannerMessage message) {
```
jacob314: nit: you can probably make this message @protected or private now.

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -20,36 +21,61 @@ const _profileGranularityDocsUrl =
...
     return _dismissedMessageKeys.contains(message.key);
   }
 
   bool isMessageVisible(BannerMessage message) {
```
jacob314: nit: you can probably make this message @protected or private now.

packages/devtools_app/lib/src/flutter/banner_messages.dart
jacob314: sounds good. We could add a TODO to create our own sublcass that has a method to add an element to an existing List and call notifyListeners.

packages/devtools_app/lib/src/flutter/banner_messages.dart
kenzieschmoll: made visible for testing - here and below

packages/devtools_app/lib/src/flutter/banner_messages.dart
jacob314: darn. I guess the right mental model is you can't change any ValueListenable objects during build either.

packages/devtools_app/lib/src/flutter/banner_messages.dart
```diff
@@ -236,10 +273,10 @@ class DebugModePerformanceMessage {
       key: Key('DebugModePerformanceMessage - $screenType'),
       textSpans: [
         const TextSpan(
-          text:
-              'You are running your app in debug mode. Debug mode performance '
-              'is not indicative of release performance.\n\nRelaunch your '
-              'application with the \'--profile\' argument, or ',
+          text: '''
+You are running your app in debug mode. Debug mode performance is not indicative of release performance.
```
jacob314: does the extra '\n' before  "You are running" 
show up in how the text span is rendered?

packages/devtools_app/lib/src/ui/flutter/vm_flag_widgets.dart
```diff
@@ -53,16 +53,10 @@ class ProfileGranularityDropdownState
...
-              bannerMessages.controller.messagesForScreen(widget.screenType);
-          final highProfileGranularityMessage = currentMessages?.firstWhere(
-              (m) => m.key
-                  .toString()
-                  .contains(HighProfileGranularityMessage.keyPrefix),
-              orElse: () => null);
-          if (highProfileGranularityMessage != null) {
-            BannerMessages.of(context).remove(highProfileGranularityMessage);
-          }
+          BannerMessages.of(context).removeMessageByKey(
```
jacob314: nice!

packages/devtools_app/lib/src/flutter/banner_messages.dart
kenzieschmoll: no it does not

packages/devtools_app/lib/src/flutter/banner_messages.dart
pq: Indeed! [`avoid_unnecessary_containers`](https://dart-lang.github.io/linter/lints/avoid_unnecessary_containers.html) should do just the trick.

pq: (PR enabling it on the way...)

pq: https://github.com/flutter/devtools/pull/1802

packages/devtools_app/lib/src/flutter/scaffold.dart
pq: Cool!  I could use some context...  Issue opened: https://github.com/dart-lang/linter/issues/2059.  Maybe you could elaborate there when you have a moment?  Thanks!

kenzieschmoll: When an `Align` widget has a child of `Padding` or vice versa, we should trigger a lint that suggests instead using Container:
```
Container(
  alignment: ...,
  padding: ...,
);

pq: Thanks @kenzieschmoll!

packages/devtools_app/lib/src/flutter/banner_messages.dart
raison00: Google's minimum recommended touch target size is around 48px.  WCAG suggests 44px or larger.


If the dismiss button had width/height of 24px, a smaller image imprint, the dismiss button should have additional padding to ensure the target size is within the 48px area.   

Going below the 24px size is not recommended for the dismiss button.  

This is a 24px with a 48px touch target for reference:
24-close

PR #1765, opened 2020-03-25
lint for relative imports

devoncarew:
- lint for relative imports

Ensure we consistently use relative imports (before, we'd have the lints for both `avoid_relative_lib_imports` and `prefer_relative_imports` enabled, which is definitely a confusing situation).


jacob314:
Do you have an example of where avoid_relative_lib_imports was causing problems?
I wouldn't expect we want to have relative imports that include the lib directory even though we otherwise prefer relative imports.
https://dart-lang.github.io/linter/lints/avoid_relative_lib_imports.html

devoncarew:
Ah, I do think we want this lint enabled.

I'd thought this was about using package: self references vs. relative paths within the lib/ dir.

PR #1772, opened 2020-03-29
add an info / about dialog

devoncarew:
add an info / about dialog:
- add an 'info' dialog, and a toolbar item to open it; this dialog shows the app version, connected device info, and a feedback / file issues link
- remove the top-level feedback action from the toolbar
- remove the 'info' tab from the app
- add tests

This change means we have fewer top-level tabs - saving space for higher priority items. It also means that we're not showing VM flags to general users, who won't likely have much need for seeing the flags.

We can add the vm flags back in once we have a way to show more advanced, VM specific pages.

Screen Shot 2020-03-28 at 4 53 53 PM

Screen Shot 2020-03-28 at 4 50 21 PM


packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -239,6 +233,137 @@ class OpenSettingsAction extends StatelessWidget {
   }
 }
 
+class DevToolsAboutDialog extends StatefulWidget {
```
kenzieschmoll: should we name this `DevToolsInfoDialog` for consistency with `DevToolsInfoAction`?

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -239,6 +233,137 @@ class OpenSettingsAction extends StatelessWidget {
...
+    );
+  }
+
+  Widget _createDeviceInfo(
+    BuildContext context,
+    TextTheme textTheme,
+    FlutterVersion flutterVersion,
+  ) {
+    if (flutterVersion == null) {
+      return const Text('No Flutter device connected.');
```
kenzieschmoll: what do we currently show in the info page for Dart CLI apps? Do we still have a way of getting the Dart SDK version?

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -239,6 +233,137 @@ class OpenSettingsAction extends StatelessWidget {
...
+    final versions = {
+      'Flutter': flutterVersion.version,
+      'Framework': flutterVersion.frameworkRevision,
+      'Engine': flutterVersion.engineRevision,
+      'Dart': flutterVersion.dartSdkVersion,
+    };
+
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.start,
+      children: <Widget>[
```
kenzieschmoll: remove `<Widget>`, here and elsewhere

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -239,6 +233,137 @@ class OpenSettingsAction extends StatelessWidget {
...
+  }
+
+  Widget _createFeedbackLink(BuildContext context, TextTheme textTheme) {
+    return InkWell(
+      onTap: () async {
+        // TODO(devoncarew): Support analytics.
+        // ga.select(ga.devToolsMain, ga.feedback);
+
+        const reportIssuesUrl = 'https://github.com/flutter/devtools/issues';
+        if (await url_launcher.canLaunch(reportIssuesUrl)) {
```
kenzieschmoll: use helper method `launchUrl`  in flutter/utils.dart

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -239,6 +233,137 @@ class OpenSettingsAction extends StatelessWidget {
...
+          await url_launcher.launch(reportIssuesUrl);
+        } else {
+          Notifications.of(context).push('Unable to open url.');
+          Notifications.of(context).push('File issues at $reportIssuesUrl.');
+        }
+      },
+      child: Text(
+        'github.com/flutter/devtools/issues',
```
kenzieschmoll: maybe pull into a const `reportIssuesUrlShort` or similar. Then above `reportIssuesUrl` could be `'https://$reportIssuesUrlShort'`

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -239,6 +233,137 @@ class OpenSettingsAction extends StatelessWidget {
...
+    );
+  }
+
+  Widget _createFeedbackLink(BuildContext context, TextTheme textTheme) {
+    return InkWell(
+      onTap: () async {
+        // TODO(devoncarew): Support analytics.
+        // ga.select(ga.devToolsMain, ga.feedback);
+
+        const reportIssuesUrl = 'https://github.com/flutter/devtools/issues';
```
kenzieschmoll: add `/new` to end of url

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -239,6 +233,137 @@ class OpenSettingsAction extends StatelessWidget {
...
+      },
+    )..entering();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final textTheme = Theme.of(context).textTheme;
+
+    return AlertDialog(
+      actions: <Widget>[
```
kenzieschmoll: remove `<Widget>`

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -239,6 +233,137 @@ class OpenSettingsAction extends StatelessWidget {
...
+        FlatButton(
```
kenzieschmoll: create a widget in custom_widgets for this button. It is also used for the settings dialog.

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -239,6 +233,137 @@ class OpenSettingsAction extends StatelessWidget {
...
+      content: Column(
+        mainAxisSize: MainAxisSize.min,
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          Text('DevTools', style: textTheme.headline6),
+          const PaddedDivider(
+            padding: EdgeInsets.only(bottom: denseRowSpacing),
+          ),
+          const SelectableText('DevTools version ${devtools.version}'),
+          //
```
kenzieschmoll: remove empty comment here and below

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -239,6 +233,137 @@ class OpenSettingsAction extends StatelessWidget {
...
+            Navigator.of(context, rootNavigator: true).pop('dialog');
+          },
+          child: const Text('CLOSE'),
+        ),
+      ],
+      content: Column(
+        mainAxisSize: MainAxisSize.min,
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          Text('DevTools', style: textTheme.headline6),
```
kenzieschmoll: trailing comma after headline6, here and below

kenzieschmoll: add a helper `_header(String title)` or similar:
```
List<Widget> _header(String title) {
  return [
    Text(title, style: textTheme.headline6),
    const PaddedDivider(padding: EdgeInsets.only(bottom: denseRowSpacing)),
  ];
}
```

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -239,6 +233,137 @@ class OpenSettingsAction extends StatelessWidget {
...
+    return AlertDialog(
+      actions: <Widget>[
+        FlatButton(
+          onPressed: () {
+            Navigator.of(context, rootNavigator: true).pop('dialog');
+          },
+          child: const Text('CLOSE'),
+        ),
+      ],
+      content: Column(
```
kenzieschmoll: break this build method up into three helpers for _devtoolsInfo, _connectedDeviceInfo, and _feedback.

packages/devtools_app/lib/src/info/flutter/info_screen.dart
```diff
@@ -1,204 +0,0 @@
...
-                Text(versions[name]),
-              ],
-            ),
-          ),
-      ],
-    );
-  }
-}
-
-class _FlagList extends StatelessWidget {
```
jacob314: Showing the flag list has been useful for us a few times in helping users debug issues. 
My preference would be to not delete this UI. It is not nearly as advanced vm author specific as other functionality that would be exposed only in advanced mode.
For example, I would like all offline snapshots to start including a snapshot of the flag data so that when we debug a user issue we can see exactly which flags were set.
Example of a similar page in chrome:
chrome://flags/

packages/devtools_app/test/flutter/about_dialog_test.dart
```diff
@@ -0,0 +1,70 @@
...
+    testWidgets('Feedback section', (WidgetTester tester) async {
+      await tester.pumpWidget(wrap(aboutDialog));
+      expect(find.text('Feedback'), findsOneWidget);
+      expect(_findSubstring(aboutDialog, 'github.com/flutter/devtools'),
+          findsOneWidget);
+    });
+  });
+}
+
+Finder _findSubstring(Widget widget, String text) {
```
jacob314: should this helper be added to a utils.dart file under the test directory?

packages/devtools_app/test/flutter/about_dialog_test.dart
```diff
@@ -0,0 +1,70 @@
...
+import 'wrappers.dart';
+
+void main() {
+  DevToolsAboutDialog aboutDialog;
+  FakeServiceManager fakeServiceManager;
+
+  group('About Dialog', () {
+    setUp(() {
+      fakeServiceManager = FakeServiceManager(useFakeService: true);
+      when(fakeServiceManager.connectedApp.isDartWebAppNow).thenReturn(false);
```
kenzieschmoll: is this necessary anymore?

packages/devtools_app/test/flutter/about_dialog_test.dart
```diff
@@ -0,0 +1,70 @@
...
+      expect(_findSubstring(aboutDialog, 'github.com/flutter/devtools'),
+          findsOneWidget);
+    });
+  });
+}
+
+Finder _findSubstring(Widget widget, String text) {
+  return find.byWidgetPredicate((widget) {
+    if (widget is Text) {
+      final Text textWidget = widget;
```
kenzieschmoll: do we need this? shouldn't type promotion happen automatically?

packages/devtools_app/test/flutter/about_dialog_test.dart
```diff
@@ -0,0 +1,70 @@
...
+
+    testWidgets('DevTools section', (WidgetTester tester) async {
+      await tester.pumpWidget(wrap(aboutDialog));
+      expect(find.text('DevTools'), findsOneWidget);
+      expect(_findSubstring(aboutDialog, devtools.version), findsOneWidget);
+    });
+
+    testWidgets('Device Info section', (WidgetTester tester) async {
+      await tester.pumpWidget(wrap(aboutDialog));
+      expect(_findSubstring(aboutDialog, 'Device Info'), findsOneWidget);
```
kenzieschmoll: nit: why not 'Connected Device Info' since that is the full section title?

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -239,6 +233,137 @@ class OpenSettingsAction extends StatelessWidget {
...
+      'Engine': flutterVersion.engineRevision,
+      'Dart': flutterVersion.dartSdkVersion,
+    };
+
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.start,
+      children: <Widget>[
+        for (var name in versions.keys)
+          Padding(
+            padding: const EdgeInsets.only(bottom: 8.0),
```
jacob314: nit: the output looks great but having padding only at the bottom here is a bit surprising.
Can you instead add padding of 4.0 to both the top and bottom and tweak the padding elsewhere so the UI still looks the same as it does now?

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -239,6 +233,137 @@ class OpenSettingsAction extends StatelessWidget {
...
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          Text('DevTools', style: textTheme.headline6),
+          const PaddedDivider(
+            padding: EdgeInsets.only(bottom: denseRowSpacing),
+          ),
+          const SelectableText('DevTools version ${devtools.version}'),
+          //
+          const SizedBox(height: defaultSpacing),
+          Text('Connected Device Info', style: textTheme.headline6),
```
kenzieschmoll: nit: what about 'Connected App Info'? Connected device makes me think it would give information like Android/iOS/Chrome/MacOs etc.

packages/devtools_app/lib/src/info/flutter/info_screen.dart
kenzieschmoll: I agree. I'd be in favor of creating a "Flags" page with a flag icon to house this UI. This page will probably need to be disabled for dart web apps, as the info page was.

devoncarew: Ah, OK. I may restore the info page then, or a new similar Flags page, with just the DevTools version removed.

devoncarew:
Screen Shot 2020-03-30 at 10 24 10 PM

Screen Shot 2020-03-31 at 7 49 08 AM

packages/devtools_app/test/flutter/about_dialog_test.dart
devoncarew: (removed this from the dialog)

PR #1776, opened 2020-03-31
Refactor splitter to allow for N children and explicit minimum sizes.

kenzieschmoll:
N split:
n_splitter
N split with minimum sizes:
splitter_with_min_size

### DIFF
--- a/packages/devtools_app/lib/src/flutter/split.dart
+++ b/packages/devtools_app/lib/src/flutter/split.dart
@@ -4,6 +4,7 @@
 
 import 'dart:math';
 
+import 'package:devtools_app/src/utils.dart';
 import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:intl/intl.dart';
@@ -16,9 +17,6 @@ import 'package:intl/intl.dart';
 ///
 /// [initialFractions] defines how much space to give each child when building
 /// this widget.
-///
-/// The user can drag the widget with key [dividerKey[index]] to change the
-/// space allocated between children[index] and children[index + 1].
 class Split extends StatefulWidget {
   /// Builds a split oriented along [axis].
   Split({
@@ -36,7 +34,7 @@ class Split extends StatefulWidget {
     for (var fraction in initialFractions) {
       sumFractions += fraction;
     }
-    assert(sumFractions == 1.0);
+    assert((1.0 - sumFractions).abs() < defaultEpsilon);
 
     if (minSizes != null) {
       assert(minSizes.length == children.length);
@@ -66,7 +64,8 @@ class Split extends StatefulWidget {
   /// The minimum size each child is allowed to be.
   final List<double> minSizes;
 
-  /// The key passed to the divider(s) between each child in [children].
+  /// The key passed to the divider between children[index] and
+  /// children[index + 1].
   ///
   /// Visible to grab it in tests.
   @visibleForTesting
@@ -138,36 +137,46 @@ class _SplitState extends State<Split> {
           isHorizontal ? dragDetails.delta.dx : dragDetails.delta.dy;
       final fractionalDelta = dragDelta / axisSize;
 
+      // Returns the actual delta applied to elements before the splitter.
       double updateSpacingBeforeSplitterIndex(double delta) {
+        final startingDelta = delta;
         var index = splitterIndex;
         while (index >= 0) {
           fractions[index] += delta;
           final minFractionForIndex = _minFractionForIndex(index);
           if (fractions[index] >= minFractionForIndex) {
             _clampFraction(index);
-            return 0.0;
+            return delta;
           }
           delta = fractions[index] - minFractionForIndex;
           _clampFraction(index);
           index--;
         }
-        return delta;
+        // At this point, we know that both [startingDelta] and [delta] are
+        // negative, and that [delta] represents the overflow that did not get
+        // applied.
+        return startingDelta - delta;
       }
 
+      // Returns the actual delta applied to elements after the splitter.
       double updateSpacingAfterSplitterIndex(double delta) {
+        final startingDelta = delta;
         var index = splitterIndex + 1;
         while (index < fractions.length) {
           fractions[index] += delta;
           final minFractionForIndex = _minFractionForIndex(index);
           if (fractions[index] >= minFractionForIndex) {
             _clampFraction(index);
-            return 0.0;
+            return delta;
           }
           delta = fractions[index] - minFractionForIndex;
           _clampFraction(index);
           index++;
         }
-        return delta;
+        // At this point, we know that both [startingDelta] and [delta] are
+        // negative, and that [delta] represents the overflow that did not get
+        // applied.
+        return startingDelta - delta;
       }
 
       setState(() {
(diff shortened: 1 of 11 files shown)


packages/devtools_app/lib/src/flutter/split.dart
```diff
@@ -22,17 +22,22 @@ import 'package:flutter/material.dart';
...
+  })  : assert(axis != null),
+        assert(children != null && children.length >= 2),
+        assert(initialFractions != null && initialFractions.length >= 2),
+        assert(children.length == initialFractions.length),
+        super(key: key) {
+    var sumFractions = 0.0;
+    for (var fraction in initialFractions) {
+      sumFractions += fraction;
+    }
+    assert(sumFractions == 1.0);
```
jacob314: need to watch for rounding error in this assert. Use an epsilon.

packages/devtools_app/lib/src/flutter/split.dart
```diff
@@ -22,17 +22,22 @@ import 'package:flutter/material.dart';
...
-    @required this.firstChild,
-    @required this.secondChild,
-    double initialFirstFraction,
-  })  : initialFirstFraction = initialFirstFraction ?? 0.5,
-        assert(axis != null),
-        assert(firstChild != null),
-        assert(secondChild != null),
-        super(key: key);
+    @required this.children,
+    @required this.initialFractions,
```
jacob314: I like this new api.

packages/devtools_app/lib/src/flutter/split.dart
```diff
@@ -45,22 +50,20 @@ class Split extends StatefulWidget {
...
   ///
-  /// [secondChild] will receive a fraction of `1 - initialFirstFraction`.
-  final double initialFirstFraction;
+  /// The index of [initialFractions] corresponds to the child at index of
+  /// [children].
+  final List<double> initialFractions;
 
-  /// The key passed to the divider between [firstChild] and [secondChild].
+  /// The key passed to the divider(s) between each child in [children].
   ///
```
jacob314: indicate that this is the divider between index and index + 1 assuming that is what index represents.

packages/devtools_app/test/flutter/split_test.dart
```diff
@@ -13,243 +13,511 @@ void main() {
...
         ));
-        expect(tester.element(find.byKey(_w1)).size, const Size(0, 200));
-        expect(tester.element(find.byKey(_w2)).size, const Size(190, 200));
+        expectEqualSizes(
+            tester.element(find.byKey(_k1)).size, const Size(0, 200));
+        expectEqualSizes(
+            tester.element(find.byKey(_k2)).size, const Size(190, 200));
+      });
+
+      testWidgets('in a horizontal layout with n children',
```
jacob314: add a test with epsilon rounding errors so initialFractions add up to 0.99999999

packages/devtools_app/lib/src/flutter/split.dart
```diff
@@ -100,29 +101,70 @@ class _SplitState extends State<Split> {
...
+            return 0.0;
+          }
+          delta = fractions[index];
+          _clampFraction(index);
+          index--;
+        }
+        return delta;
+      }
+
+      double updateSpacingAfterSplitterIndex(double delta) {
```
jacob314: nit: the return value from these methods doesn't seem like it is used. Make it void if it isn't used.

packages/devtools_app/lib/src/flutter/split.dart
```diff
@@ -100,29 +101,70 @@ class _SplitState extends State<Split> {
...
 
-    // Remove space from each child to place the divider in the middle.
-    firstSize = firstSize - halfDivider;
-    secondSize = secondSize - halfDivider;
+    void updateSpacing(DragUpdateDetails dragDetails, int splitterIndex) {
+      final dragDelta =
+          isHorizontal ? dragDetails.delta.dx : dragDetails.delta.dy;
+      final fractionalDelta = dragDelta / axisSize;
+
+      double updateSpacingBeforeSplitterIndex(double delta) {
```
jacob314: comment that these two methods return the actual delta applied.

packages/devtools_app/lib/src/flutter/split.dart
```diff
@@ -100,29 +101,70 @@ class _SplitState extends State<Split> {
...
 
-    void updateSpacing(DragUpdateDetails dragDetails) {
-      final delta = isHorizontal ? dragDetails.delta.dx : dragDetails.delta.dy;
-      final fractionalDelta = delta / axisSize;
       setState(() {
-        // Update the fraction of space consumed by the children,
-        // being sure not to allocate any negative space.
-        firstFraction += fractionalDelta;
-        firstFraction = firstFraction.clamp(0.0, 1.0);
+        // Update the fraction of space consumed by the children. Always update
```
jacob314: nice comment!

packages/devtools_app/lib/src/flutter/split.dart
```diff
@@ -100,29 +101,70 @@ class _SplitState extends State<Split> {
...
+        // the shrinking children first so that we do not over-increase the size
+        // of the growing children and cause layout overflow errors.
+        if (fractionalDelta <= 0.0) {
+          final overflowDelta =
+              updateSpacingBeforeSplitterIndex(fractionalDelta);
+          final actualDelta = overflowDelta != 0.0
```
jacob314: this conditional isn't needed as both conditions are quivalent.

packages/devtools_app/lib/src/flutter/split.dart
```diff
@@ -100,29 +101,70 @@ class _SplitState extends State<Split> {
...
+          updateSpacingAfterSplitterIndex(-actualDelta);
+        } else {
+          final overflowDelta =
+              updateSpacingAfterSplitterIndex(-fractionalDelta);
+          final actualDelta = overflowDelta != 0.0
+              ? fractionalDelta + overflowDelta
+              : fractionalDelta;
+          updateSpacingBeforeSplitterIndex(actualDelta);
+        }
       });
```
jacob314: this conditional isn't needed as both conditions are equivalent.

packages/devtools_app/lib/src/flutter/split.dart
jacob314: > comment that these two methods return the actual delta applied.

I misread what the method did. I mean change the implementation to return that as that is simpler than returning the delta remaining which makes the logic a bit confusing.

packages/devtools_app/lib/src/flutter/split.dart
```diff
@@ -6,27 +6,27 @@ import 'dart:math';
...
-/// [initialFirstFraction] defines how much space to give the [firstChild]
-/// when first building this widget. [secondChild] will take the remaining
-/// space.
+/// [initialFractions] defines how much space to give each child when building
+/// this widget.
 ///
-/// The user can drag the widget with key [dividerKey] to change
-/// the space allocated between [firstChild] and [secondChild].
-// TODO(djshuckerow): introduce support for a minimum fraction a child is allowed.
+/// The user can drag the widget with key [dividerKey[index]] to change the
```
jacob314: nit: make this comment more generic. Just indicate the splitters can be dragged rather than explaining the details of the splitter in the class level comment.

kenzieschmoll: done

packages/devtools_app/lib/src/flutter/split.dart
kenzieschmoll: done

packages/devtools_app/lib/src/flutter/split.dart
kenzieschmoll: done

packages/devtools_app/lib/src/flutter/split.dart
kenzieschmoll: removed

packages/devtools_app/lib/src/flutter/split.dart
kenzieschmoll: good idea. done.

packages/devtools_app/lib/src/flutter/split.dart
kenzieschmoll: removed

packages/devtools_app/test/flutter/split_test.dart
kenzieschmoll: done

kenzieschmoll:
Landing this on red to avoid merge conflicts with Dave's CL. The failure is happening on master and is not related to this CL.

PR #1777, opened 2020-03-31
nits around the logging page

devoncarew:
nits around the logging page:
- only show one line in the summary portion of the UI for stdout/stderr text
- ensure that the logger line contents are centered in the line (previously, some was centered, and some was tall enough to look top-aligned)


packages/devtools_app/lib/src/logging/logging_controller.dart
```diff
@@ -665,9 +665,14 @@ class _StdoutEventHandler {
       buffer = null;
     }
 
+    const maxLength = 200;
     String summary = message;
-    if (message.length > 200) {
-      summary = message.substring(0, 200) + '…';
+    var index = summary.indexOf('\n');
```
jacob314: I believe Flutter supports its own more robust text rendering truncation with '...'
we should just use that rather than having our own implementation once we remove the dart:html version.

jacob314: maybe add a TODO for that?

devoncarew: What I want to do here is only display 1 line if there are multiple, and if the first line is long, to display `...` on overflow.

I'll play around with getting that working w/ Flutter mechanisms, and otherwise add a todo: and keep the string truncation stuff.

PR #1779, opened 2020-04-01
Update to use the latest Linux desktop runner.

terrylucas:
Flutter's Linux desktop may need a newer runner, a version mismatch error will occur.  This can happen when using a new master branch of Flutter (Linux desktop currently only works on the master branch). If this happens then after the steps:

- run `flutter update-packages` 
- run `pub get` on each `devtools/packages/` subdirectory (`devtools`, `devtools_server`, and `devtools_shared`)
- run `flutter pub get` for directory `devtools_app`

Re-generate files in the directories linux and macos go to the `devtools_app` directory and run `flutter create .`


terrylucas:
Jacob please TBR - any suggestions I'll make but I need this to continue working in Linux.

devoncarew:
@terrylucas - thanks for updating this!

After syncing to master and running `flutter create .` in flutter_app/, I see a bunch of changed file. Should some of these be committed to the repo, and for the others, should we update our .gitignore file?

PR #1780, opened 2020-04-01
update some of the generated files

devoncarew:
- update some of the generated files

This commits some of the generated web/ resources, and ignores the ios/ and android/ directories. We'll still need to rationalize what if anything we check in for the linux/ and macos/ directories.

@terrylucas


packages/devtools_app/web/manifest.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "devtools_app",
+    "short_name": "devtools_app",
+    "start_url": ".",
+    "display": "minimal-ui",
+    "background_color": "#0175C2",
+    "theme_color": "#0175C2",
+    "description": "A new Flutter project.",
```
jacob314, line 8: nit: put a real description

devoncarew: done

Issue #1568, opened 2020-01-23
Reduce calls to `FrameworkCore.init`


DaveShuckerow:

devoncarew:
@DaveShuckerow - what would be a good milestone for this issue?

DaveShuckerow:
Backlog.  This is not high-priority.

PR #1783, opened 2020-04-01
fix an issue where we tried to initiate a connection twice

devoncarew:
- fix an issue where we tried to initiate a connection twice

When using an `onSubmitted` on the `TextField`, both the callback for the onSubmitted and for the `RaisedButton`'s onPressed are called. This makes sure that we only try and initiate a connection once.

This would show up as an error after connection - displayed as a 'Completer already completed' error.


kenzieschmoll:
I think this PR fixes https://github.com/flutter/devtools/issues/1568

devoncarew:
> I think this PR fixes #1568

There may be additional calls (and some rationalization we can make to the init sequence). This at least does address part of #1568.

PR #1784, opened 2020-04-01
Use FlexSplitColumn for debugger panes.

kenzieschmoll:
This CL:
- Creates the `FlexSplitColumn` widget, which builds a `Split` widget with custom splitters and minor fraction and size transformations. Functionality explained in comments of respective class members.
- Adds functionality to specify custom splitter widgets in `Split`
- Uses `FlexSplitColumn` for the debugger panes so that the section headers are now the "splitter" widgets that respond to dragging and resizing.

debugger_panes

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
@@ -6,9 +6,11 @@ import 'package:flutter/material.dart';
 import 'package:vm_service/vm_service.dart';
 
 import '../../flutter/common_widgets.dart';
+import '../../flutter/flex_split_column.dart';
 import '../../flutter/octicons.dart';
 import '../../flutter/screen.dart';
 import '../../flutter/split.dart';
+import '../../flutter/theme.dart';
 import '../../globals.dart';
 
 class DebuggerScreen extends Screen {
@@ -41,6 +43,12 @@ class DebuggerScreenBody extends StatefulWidget {
 }
 
 class DebuggerScreenBodyState extends State<DebuggerScreenBody> {
+  static const callStackTitle = 'Call Stack';
+  static const variablesTitle = 'Variables';
+  static const breakpointsTitle = 'Breakpoints';
+  static const librariesTitle = 'Libraries';
+  static const debuggerPaneHeaderHeight = 60.0;
+
   ScriptRef loadingScript;
   Script script;
   ScriptList scriptList;
@@ -82,13 +90,7 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody> {
       initialFractions: const [0.25, 0.75],
       // TODO(https://github.com/flutter/devtools/issues/1648): Debug panes.
       children: [
-        OutlinedBorder(
-          child: ScriptPicker(
-            scripts: scriptList,
-            onSelected: onScriptSelected,
-            selected: loadingScript,
-          ),
-        ),
+        OutlinedBorder(child: debuggerPanes()),
         // TODO(https://github.com/flutter/devtools/issues/1648): Debug controls.
         OutlinedBorder(
           child: loadingScript != null && script == null
@@ -98,6 +100,60 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody> {
                 ),
         ),
       ],
+    );
+  }
+
+  Widget debuggerPanes() {
+    return LayoutBuilder(
+      builder: (context, constraints) {
+        final paneWidth = constraints.maxWidth;
+        return FlexSplitColumn(
+          totalHeight: constraints.maxHeight,
+          initialFractions: const [0.25, 0.25, 0.25, 0.25],
+          minSizes: const [0.0, 0.0, 0.0, 0.0],
+          headers: [
+            _debuggerPaneHeader(callStackTitle, paneWidth),
+            _debuggerPaneHeader(variablesTitle, paneWidth),
+            _debuggerPaneHeader(breakpointsTitle, paneWidth),
+            _debuggerPaneHeader(librariesTitle, paneWidth),
+          ],
+          children: [
+            const Center(child: Text('TODO: call stack')),
+            const Center(child: Text('TODO: variables')),
+            const Center(child: Text('TODO: breakpoints')),
+            ScriptPicker(
+              scripts: scriptList,
+              onSelected: onScriptSelected,
+              selected: loadingScript,
+            ),
+          ],
+        );
+      },
+    );
+  }
+
+  FlexSplitColumnHeader _debuggerPaneHeader(
+    String title,
+    double debuggerPaneWidth,
+  ) {
+    return FlexSplitColumnHeader(
+      height: debuggerPaneHeaderHeight,
+      width: debuggerPaneWidth,
+      child: Container(
+        decoration: BoxDecoration(
+          border: Border(
+            bottom: BorderSide(color: Theme.of(context).focusColor),
+          ),
+          color: Theme.of(context).primaryColor,
+        ),
+        padding: EdgeInsets.only(left: defaultSpacing),
+        alignment: Alignment.centerLeft,
+        height: debuggerPaneHeaderHeight,
+        child: Text(
+          title,
+          style: Theme.of(context).textTheme.headline6,
+        ),
+      ),
     );
   }
 }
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
```diff
@@ -100,6 +102,60 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody> {
...
+          headers: [
+            _debuggerPaneHeader(callStackTitle, paneWidth),
+            _debuggerPaneHeader(variablesTitle, paneWidth),
+            _debuggerPaneHeader(breakpointsTitle, paneWidth),
+            _debuggerPaneHeader(librariesTitle, paneWidth),
+          ],
+          children: [
+            const Center(child: Text('TODO: call stack')),
+            const Center(child: Text('TODO: variables')),
+            const Center(child: Text('TODO: breakpoints')),
```
kenzieschmoll, line 123: @DaveShuckerow this is where `BreakpointPicker` will go.

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
```diff
@@ -100,6 +102,60 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody> {
...
+  Widget debuggerPanes() {
+    return LayoutBuilder(
+      builder: (context, constraints) {
+        final paneWidth = constraints.maxWidth;
+        return FlexSplitColumn(
+          totalHeight: constraints.maxHeight,
+          initialFractions: const [0.25, 0.25, 0.25, 0.25],
+          minSizes: const [0.0, 0.0, 0.0, 0.0],
+          headers: [
+            _debuggerPaneHeader(callStackTitle, paneWidth),
```
jacob314, line 115: seems odd we have to pass in the pane width to the headers

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
```diff
@@ -100,6 +102,60 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody> {
       ],
     );
   }
+
+  Widget debuggerPanes() {
+    return LayoutBuilder(
+      builder: (context, constraints) {
+        final paneWidth = constraints.maxWidth;
+        return FlexSplitColumn(
+          totalHeight: constraints.maxHeight,
```
jacob314, line 111: seems a bit odd we have to pass in the totalHeight to the FlexSplitColumn. This seems like the natural default height for a FlexSpltColumn same as it would be for a regular column

packages/devtools_app/lib/src/flutter/flex_split_column.dart
```diff
@@ -0,0 +1,132 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+import 'package:flutter/material.dart';
+
+import 'split.dart';
+
+class FlexSplitColumn extends StatelessWidget {
+  FlexSplitColumn({
+    Key key,
+    @required this.totalHeight,
```
jacob314, line 11: instead of passing in the height let users wrap the FlexSplitColumn in the widget they want to give it the size they want.

packages/devtools_app/lib/src/flutter/flex_split_column.dart
```diff
@@ -0,0 +1,132 @@
...
+  final List<double> adjustedInitialFractions;
+
+  /// The minimum size each child is allowed to be.
+  final List<double> adjustedMinSizes;
+
+  /// The total height of the column, including all [headers] and [children].
+  final double totalHeight;
+
+  @visibleForTesting
+  static List<Widget> adjustChildren(
```
jacob314, line 68: the name adjust is confusing as this method doesn't have side effects. maybe a method name that indicates what the widget list it is building looks like. E.g. buildSomethingSomething

packages/devtools_app/lib/src/flutter/flex_split_column.dart
```diff
@@ -0,0 +1,132 @@
...
+    ];
+  }
+
+  @visibleForTesting
+  static List<double> adjustInitialFractions(
+    List<double> initialFractions,
+    List<FlexSplitColumnHeader> headers,
+    double totalHeight,
+  ) {
+    final intendedContentHeight = totalHeight -
```
jacob314, line 89: nit picking but I think the simple imperative code
```dart
var height = 0.0;
for (var header in header) {
  height += header.height;
}
```
is quicker to understand than the functional code using map and reduce.

packages/devtools_app/lib/src/flutter/split.dart
```diff
@@ -64,15 +63,21 @@ class Split extends StatefulWidget {
...
   /// The key passed to the divider between children[index] and
   /// children[index + 1].
   ///
   /// Visible to grab it in tests.
   @visibleForTesting
   Key dividerKey(int index) => Key('$this dividerKey $index');
 
-  /// The size of the divider between children in logical pixels (dp, not px).
-  static const double dividerMainAxisSize = 10.0;
+  /// The default size of the divider between children in logical pixels (dp,
```
jacob314, line 78: no need to indicate dp, not px as all calculations we do for flutter should be in logical pixels.

packages/devtools_app/lib/src/flutter/split.dart
```diff
@@ -214,47 +264,34 @@ class _SplitState extends State<Split> {
...
+      width: isHorizontal ? Split.defaultSplitterSize : layoutWidth,
+      height: isHorizontal ? layoutHeight : Split.defaultSplitterSize,
+      child: Center(
+        child: defaultDragIndicator,
+      ),
+    );
+  }
+}
+
+void _verifyFractionsSum(List<double> fractions) {
```
jacob314, line 288: nit: make the method name indicate what it is you are verifying the fractions sum to.

packages/devtools_app/lib/src/flutter/flex_split_column.dart
```diff
@@ -0,0 +1,132 @@
...
+          headers[0],
+          Expanded(child: children[0]),
+        ],
+      ),
+      ...children.sublist(1),
+    ];
+  }
+
+  @visibleForTesting
+  static List<double> adjustInitialFractions(
```
jacob314, line 84: make this method name indicate how they are adjusted.

packages/devtools_app/lib/src/flutter/flex_split_column.dart
```diff
@@ -0,0 +1,132 @@
...
+      children: adjustedChildren,
+      initialFractions: adjustedInitialFractions,
+      minSizes: adjustedMinSizes,
+      splitters: headers.sublist(1),
+    );
+  }
+}
+
+class FlexSplitColumnHeader extends SizedBox {
+  const FlexSplitColumnHeader({
```
jacob314, line 127: it is a little unfortunate we have to make the headers be fixed size widgets like this but I understand why that isn't easy.
Add a TODO to allow arbitrary widgets as column headers.  No need to do it now as we don't have enough use cases for this to justify the complexity.

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
kenzieschmoll: we need `totalHeight` to perform calculations in a static method `FlexSplitColumn.adjustInitialFractions`, not for building the widget.

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
kenzieschmoll: turns out we don't need it - removed.

packages/devtools_app/lib/src/flutter/flex_split_column.dart
kenzieschmoll: we need totalHeight for the calculation done in the static method `adjustInitialFractions`

packages/devtools_app/lib/src/flutter/flex_split_column.dart
kenzieschmoll: done

packages/devtools_app/lib/src/flutter/flex_split_column.dart
kenzieschmoll: changed to `adjustInitialFractionsToIncludeFirstHeader`

packages/devtools_app/lib/src/flutter/flex_split_column.dart
kenzieschmoll: added todo

packages/devtools_app/lib/src/flutter/split.dart
kenzieschmoll: done.

packages/devtools_app/lib/src/flutter/split.dart
kenzieschmoll: done

packages/devtools_app/lib/src/flutter/flex_split_column.dart
kenzieschmoll: changed to `buildChildrenWithFirstHeader`

PR #1787, opened 2020-04-02
[WIP] Connect the Flutter web app version to the devtools server

devoncarew:
- connect the Flutter web app version to the devtools server
- add a PreferencesController which uses the server connection to persist user preferences

Currently WIP (several todos to address)

- init the DevToolsServerConnection with a Flutter web connection controller
- make sure the Flutter web version sends the events: connected, currentPage, disconnected

PR #1790, opened 2020-04-03
add a SettingsController

devoncarew:
- add a SettingsController; use that in the app to control the current theme (and address a todo)

In a future refactor, this controller will be backed by storage on disk, and changes will propagate via the SSE connection to the devtools server. The theme variable in the url is now used for initial setup of the theme value, but the SettingsController's value is used after that. 

Pulled out of https://github.com/flutter/devtools/pull/1787 to make that PR smaller.

### DIFF
--- a/packages/devtools_app/lib/src/flutter/app.dart
+++ b/packages/devtools_app/lib/src/flutter/app.dart
@@ -277,10 +277,6 @@ class SettingsDialog extends StatelessWidget {
   Widget build(BuildContext context) {
     final preferences = DevToolsApp.of(context).preferences;
 
-    void _toggleTheme() {
-      preferences.darkModeTheme.value = !preferences.darkModeTheme.value;
-    }
-
     return AlertDialog(
       actions: [
         DialogCloseButton(),
@@ -291,7 +287,9 @@ class SettingsDialog extends StatelessWidget {
         children: [
           ..._header(Theme.of(context).textTheme, 'Settings'),
           InkWell(
-            onTap: _toggleTheme,
+            onTap: () {
+              preferences.toggleDarkModeTheme(!preferences.darkModeTheme.value);
+            },
             child: Row(
               children: [
                 ValueListenableBuilder(
@@ -300,7 +298,7 @@ class SettingsDialog extends StatelessWidget {
                     return Checkbox(
                       value: value,
                       onChanged: (bool value) {
-                        preferences.darkModeTheme.value = value;
+                        preferences.toggleDarkModeTheme(value);
                       },
                     );
                   },
--- a/packages/devtools_app/lib/src/flutter/app.dart
+++ b/packages/devtools_app/lib/src/flutter/app.dart
@@ -17,11 +17,11 @@ import '../network/flutter/network_screen.dart';
 import '../performance/flutter/performance_screen.dart';
 import '../timeline/flutter/timeline_screen.dart';
 import '../ui/flutter/service_extension_widgets.dart';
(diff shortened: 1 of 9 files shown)


packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -297,32 +272,13 @@ class DevToolsAboutDialog extends StatelessWidget {
...
-    void _toggleTheme([bool value]) {
-      value ??= !devtools_theme.isDarkTheme;
-      setState(() {
-        devtools_theme.useDarkTheme = value;
+    final preferences = DevToolsApp.of(context).preferences;
 
-        DevToolsApp.of(context).updateTheme();
-      });
+    void _toggleTheme() {
+      preferences.darkModeTheme.value = !preferences.darkModeTheme.value;
```
kenzieschmoll, line 281: instead of having the view change this, let's have the view call the controller, which changes the value of darkModeTheme.  We should also only expose the value listenable from the controller and have the notifier be private

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -338,9 +294,16 @@ class _SettingsDialogState extends State<SettingsDialog> {
...
-                Checkbox(
-                  value: devtools_theme.isDarkTheme,
-                  onChanged: (bool value) => _toggleTheme(value),
+                ValueListenableBuilder(
+                  valueListenable: preferences.darkModeTheme,
+                  builder: (context, value, _) {
+                    return Checkbox(
+                      value: value,
+                      onChanged: (bool value) {
+                        preferences.darkModeTheme.value = value;
```
kenzieschmoll, line 303: do we need to change the theme value here? It looks like we are already doing this in the Inkwell's onTap handler

packages/devtools_app/lib/src/flutter/preferences.dart
```diff
@@ -0,0 +1,56 @@
...
+    // TODO(devoncarew): Enable when we have storage backed settings.
+    //storage.getValue('analytics.enabled').then((String value) {
+    //  analyticsEnabled.value = value == 'true';
+    //  analyticsEnabled.addListener(() {
+    //    storage.setValue('analytics.enabled', '${analyticsEnabled.value}');
+    //  });
+    //});
+  }
+
+  final ValueNotifier<bool> darkModeTheme =
```
kenzieschmoll, line 47: make `darkModeTheme` a ValueListenable getter that returns a private notifier _darkModeTheme. That way only the controller can modify the value. You'll need to also add a method `toggleDarkModeTheme(bool isDarkMode)` or similar that changes the value of the private notifier

packages/devtools_app/lib/src/flutter/preferences.dart
```diff
@@ -0,0 +1,56 @@
...
+      ValueNotifier(devtools_theme.isDarkTheme);
+
+  // TODO(devoncarew): Enable when we have storage backed settings.
+  final ValueNotifier<bool> analyticsEnabled = ValueNotifier(false);
```
kenzieschmoll, line 51: same comment for this notifier

packages/devtools_app/lib/src/flutter/preferences.dart
```diff
@@ -0,0 +1,56 @@
...
+
+  void dispose() {
+    // Nothing to do here.
```
kenzieschmoll, line 54: remove this method for now?

devoncarew:
(landing TBR so I can continue work on the main #1787 PR; will address any other comments in follow-up)

PR #1796, opened 2020-04-07
Add ConditionalScreen functionality.

kenzieschmoll:
This CL adds functionality to conditionally add a screen in DevTools based on the presence of a library URI.

Here is an example of a ConditionalScreen added for the Flutter Gallery app based on the presence of "package:gallery/"
Screen Shot 2020-04-07 at 12 50 29 PM

### DIFF
--- a/packages/devtools_app/lib/main.dart
+++ b/packages/devtools_app/lib/main.dart
@@ -6,12 +6,15 @@ import 'package:flutter/material.dart';
 
 import 'src/config_specific/flutter/framework_initialize/framework_initialize.dart';
 import 'src/flutter/app.dart';
+import 'src/flutter/screen.dart';
 
 void main() {
+  const conditionalScreens = <ConditionalScreen>[];
+
   initializeFramework();
 
   // Now run the app.
   runApp(
-    DevToolsApp(),
+    const DevToolsApp(conditionalScreens),
   );
 }
--- a/packages/devtools_app/lib/main.dart
+++ b/packages/devtools_app/lib/main.dart
@@ -10,7 +10,7 @@ import 'src/flutter/screen.dart';
 
 void main() {
   // TODO(kenz): add some conditional screens.
-  const conditionalScreens = <ConditionalScreen>[PackageProviderScreen()];
+  const conditionalScreens = <ConditionalScreen>[];
 
   initializeFramework();
 
--- a/packages/devtools_app/lib/main.dart
+++ b/packages/devtools_app/lib/main.dart
@@ -9,7 +9,8 @@ import 'src/flutter/app.dart';
 import 'src/flutter/screen.dart';
 
 void main() {
-  const conditionalScreens = <ConditionalScreen>[];
+  // TODO(kenz): add some conditional screens.
+  const conditionalScreens = <ConditionalScreen>[PackageProviderScreen()];
 
   initializeFramework();
 
--- a/packages/devtools_app/lib/src/flutter/app.dart
+++ b/packages/devtools_app/lib/src/flutter/app.dart
@@ -65,6 +65,12 @@ class DevToolsAppState extends State<DevToolsApp> {
     });
   }
 
+  @override
+  void didUpdateWidget(DevToolsApp oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    _clearCachedRoutes();
+  }
+
   /// Generates routes, separating the path from URL query parameters.
   Route _generateRoute(RouteSettings settings) {
     final uri = Uri.parse(settings.name);
--- a/packages/devtools_app/lib/src/flutter/app.dart
+++ b/packages/devtools_app/lib/src/flutter/app.dart
@@ -39,7 +39,7 @@ class DevToolsApp extends StatefulWidget {
   final List<ConditionalScreen> conditionalScreens;
 
   @override
-  State<DevToolsApp> createState() => DevToolsAppState(conditionalScreens);
+  State<DevToolsApp> createState() => DevToolsAppState();
 
   static DevToolsAppState of(BuildContext context) {
     return context.findAncestorStateOfType<DevToolsAppState>();
@@ -53,10 +53,6 @@ class DevToolsApp extends StatefulWidget {
 // TODO(https://github.com/flutter/devtools/issues/1146): Introduce tests that
 // navigate the full app.
 class DevToolsAppState extends State<DevToolsApp> {
-  DevToolsAppState([this.conditionalScreens]);
-
-  final List<ConditionalScreen> conditionalScreens;
-
   final PreferencesController preferences = PreferencesController();
 
   @override
@@ -64,7 +60,7 @@ class DevToolsAppState extends State<DevToolsApp> {
     super.initState();
     serviceManager.isolateManager.onSelectedIsolateChanged.listen((_) {
       setState(() {
-        _regenerateRoutes();
+        _clearCachedRoutes();
       });
     });
   }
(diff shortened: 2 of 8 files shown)


packages/devtools_app/lib/main.dart
```diff
@@ -6,12 +6,15 @@ import 'package:flutter/material.dart';
 
 import 'src/config_specific/flutter/framework_initialize/framework_initialize.dart';
 import 'src/flutter/app.dart';
+import 'src/flutter/screen.dart';
 
 void main() {
+  const conditionalScreens = <ConditionalScreen>[];
```
jacob314, line 12: add a TODO to add some conditional screens so this code makes more sense.

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -32,8 +34,12 @@ const showNetworkPage = false;
 /// Top-level configuration for the app.
 @immutable
 class DevToolsApp extends StatefulWidget {
+  const DevToolsApp([this.conditionalScreens]);
+
+  final List<ConditionalScreen> conditionalScreens;
+
   @override
-  State<DevToolsApp> createState() => DevToolsAppState();
+  State<DevToolsApp> createState() => DevToolsAppState(conditionalScreens);
```
jacob314, line 42: To make this code consistent with Flutter, don't pass this into the state constructor. Instead grab it off of the widget within the DevToolsAppState object.

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -47,20 +53,34 @@ class DevToolsApp extends StatefulWidget {
 // TODO(https://github.com/flutter/devtools/issues/1146): Introduce tests that
 // navigate the full app.
 class DevToolsAppState extends State<DevToolsApp> {
+  DevToolsAppState([this.conditionalScreens]);
```
jacob314, line 56: state objects shouldn't have constructor parameters and should instead read this data off the associated widget.

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -82,31 +102,53 @@ class DevToolsAppState extends State<DevToolsApp> {
...
-              HotRestartButton(),
-              OpenSettingsAction(),
-              OpenAboutAction(),
-            ],
-          ),
-        ),
-    '/connect': (_, __) =>
-        DevToolsScaffold.withChild(child: ConnectScreenBody()),
-  };
+  Map<String, UrlParametersBuilder> get routes => _routes ??= {
```
jacob314, line 105: nit: consider using `{` instead of `=>` in this case to make the code easier to read.

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -82,31 +102,53 @@ class DevToolsAppState extends State<DevToolsApp> {
...
+            ),
+        '/connect': (_, __) =>
+            DevToolsScaffold.withChild(child: ConnectScreenBody()),
+      };
+
+  Map<String, UrlParametersBuilder> _routes;
+
+  void _regenerateRoutes() {
+    _routes = null;
+    routes;
```
jacob314, line 136: why is the routes getter called here? It doesn't seem needed.
Calling a getter just for its side effects is a code smell that should generally be avoided.
probably rename this method to _clearCachedRoutes or something.

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -82,31 +102,53 @@ class DevToolsAppState extends State<DevToolsApp> {
...
+  }
+
+  List<ConditionalScreen> _generateConditionalScreens() {
```
jacob314, line 139: nit: generate seems misleading. You are really computing the currentlyVisibleConditionalScreens
or something less wordy to that effect.

packages/devtools_app/lib/src/flutter/screen.dart
```diff
@@ -16,7 +16,7 @@ import 'connect_screen.dart';
 import 'scaffold.dart';
 import 'theme.dart';
 
-/// Defines pages shown in the tabbar of the app.
+/// Defines a page shown in the DevTools TabBar.
```
jacob314, line 19: nit: use `[` `]` in dartdoc comments for words that match class names.

packages/devtools_app/lib/src/flutter/screen.dart
```diff
@@ -76,6 +76,28 @@ abstract class Screen {
   }
 }
 
+/// Defines a page shown in the DevTools TabBar if [conditionalLibrary] is
+/// present in the connected application.
+abstract class ConditionalScreen extends Screen {
```
jacob314, line 81: It is worth adding a bit more of a comment for this class as it will have customers who aren't familiar with our code.
Maybe given an example showing how to use it for `package:provider`?

Also indicate limitations such as:
These screens will be completely hidden if the screen isn't available unlike pages such as the inspector that will intentionally show up as disabled if they are not available.

packages/devtools_app/lib/src/flutter/screen.dart
```diff
@@ -76,6 +76,28 @@ abstract class Screen {
...
+    IconData icon,
+    Key tabKey,
+  }) : super(
+          DevToolsScreenType.conditional,
+          title: title,
+          icon: icon,
+          tabKey: tabKey,
+        );
+
+  /// Library name on which we determine whether to include this screen in
```
jacob314, line 94: nit: on which we determine -->  that determines

packages/devtools_app/lib/src/flutter/screen.dart
```diff
@@ -76,6 +76,28 @@ abstract class Screen {
...
+  /// DevTools.
+  ///
+  /// This can either be a full library name or it can be a prefix.
```
jacob314, line 97: this is a library uri, not name, correct?
Give a coupe examples of valid inputs to use.

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -82,31 +102,53 @@ class DevToolsAppState extends State<DevToolsApp> {
...
+                tabs: [
+                  const InspectorScreen(),
+                  const TimelineScreen(),
+                  const MemoryScreen(),
+                  const PerformanceScreen(),
+                  const DebuggerScreen(),
+                  if (showNetworkPage) const NetworkScreen(),
+                  const LoggingScreen(),
+                  const InfoScreen(),
+                  ..._generateConditionalScreens(),
```
devoncarew, line 118: We may want to have some of the always included screens at the end of the list (logging? info?). Perhaps put the conditional screens ahead of the logging screen entry.

packages/devtools_app/lib/src/flutter/screen.dart
```diff
@@ -87,8 +109,11 @@ enum DevToolsScreenType {
   debugger,
   network,
   simple,
+  conditional,
```
jacob314, line 112: add a TODO to check if we are using the screen type for logging? we'll probably want to stop doing that and start just using string ids and all conditional screens would get grouped together which wouldn't be great.

devoncarew, line 112: I wonder if we should drop this enum and just load screens based on a unique string id?

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -288,6 +288,24 @@ class ServiceConnectionManager {
 
     return fps.roundToDouble();
   }
+
+  bool libraryUriAvailableNow(String uri) {
+    assert(serviceAvailable.isCompleted);
+    assert(isolateManager.selectedIsolateAvailable.isCompleted);
+    return isolateManager.selectedIsolateLibraries
+        .map((ref) => ref.uri)
```
jacob314, line 296: duplicated code between libraryUriAvailableNow and libraryUriAvailable.  
`libraryUriAvailableNow` can be called after awaiting the `selectedIsolateAvailable.future`
the code can also be simplified to 
isolateManager.selectedIsolateLibraries.any((u) => u.uri.startsWith(uri))

packages/devtools_app/lib/src/flutter/screen.dart
kenzieschmoll: added docs

packages/devtools_app/lib/src/flutter/screen.dart
kenzieschmoll: done

packages/devtools_app/lib/src/flutter/screen.dart
kenzieschmoll: done

packages/devtools_app/lib/src/flutter/screen.dart
kenzieschmoll: The original reason this enum was created was so that we could import snapshots, and then based on a string in the snapshot, we could create the correct `Screen` class. The challenge is to make this change while leaving the functionality in [import_export.dart](https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/config_specific/flutter/import_export/import_export.dart#L59) in tact, and also making conditional screens  compatible with import / export.

kenzieschmoll: Latest commit includes changes. PTAL

packages/devtools_app/lib/src/flutter/screen.dart
```diff
@@ -76,8 +76,25 @@ abstract class Screen {
...
 
-/// Defines a page shown in the DevTools TabBar if [conditionalLibrary] is
+/// Defines a page shown in the DevTools [TabBar] if [conditionalLibrary] is
 /// present in the connected application.
+///
+/// If [conditionalLibrary] is not present, the screen will be hidden from the
+/// [TabBar], not just disabled like other DevTools [Screen]s.
+/// Example:
+///
+/// class PackageProviderScreen extends ConditionalScreen {
```
jacob314, line 87: Nit: see how the Flutter code includes examples in dartdoc comments to find how to make examples show up better in DartDoc.

packages/devtools_app/lib/src/flutter/screen.dart
```diff
@@ -118,46 +127,61 @@ abstract class ConditionalScreen extends Screen {
...
+  static const debugger = DevToolsScreenType(debuggerId);
+  static const logging = DevToolsScreenType(loggingId);
+  static const info = DevToolsScreenType(infoId);
+  static const connect = DevToolsScreenType(connectId);
+  static const simple = DevToolsScreenType(simpleId);
 
-// TODO(kenz): we may need to refactor this code if conditional screens need
-// to be created from imports in the same way that non-conditional screens do.
-extension DevToolsScreenTypeExtension on DevToolsScreenType {
   Screen create() {
```
jacob314, line 159: lets eliminate the create() method completely. It would be simpler and equally effective to instead filter the list of screens to only include the screen matching the DevToolsScreenId rather than requiring screens to have factories like this.
I'd suggest we remove the DevToolsScreenType class and just use string Ids.

packages/devtools_app/lib/main.dart
```diff
@@ -9,7 +9,8 @@ import 'src/flutter/app.dart';
 import 'src/flutter/screen.dart';
 
 void main() {
-  const conditionalScreens = <ConditionalScreen>[];
+  // TODO(kenz): add some conditional screens.
+  const conditionalScreens = <ConditionalScreen>[PackageProviderScreen()];
```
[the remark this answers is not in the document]
kenzieschmoll: whoops no meant to remove

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -53,18 +53,14 @@ class DevToolsApp extends StatefulWidget {
...
 // navigate the full app.
 class DevToolsAppState extends State<DevToolsApp> {
-  DevToolsAppState([this.conditionalScreens]);
-
-  final List<ConditionalScreen> conditionalScreens;
-
   final PreferencesController preferences = PreferencesController();
 
   @override
   void initState() {
```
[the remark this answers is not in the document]
kenzieschmoll: done

packages/devtools_app/lib/src/flutter/screen.dart
kenzieschmoll: fixed

packages/devtools_app/lib/src/flutter/screen.dart
kenzieschmoll: Will do some refactoring for this in a follow up PR

kenzieschmoll:
Landing this and following up with a PR to do the mentioned refactor.

jacob314:
lgtm

PR #1801, opened 2020-04-08
Fix error in debugger page.

kenzieschmoll:
We need to be initializing the controller and listeners in didChangeDependencies so that we have access to `context`.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
@@ -63,14 +63,10 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
   ScriptList scriptList;
   Stack stack;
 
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
```diff
@@ -64,8 +64,8 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
   Stack stack;
 
   @override
-  void initState() {
-    super.initState();
+  void didChangeDependencies() {
```
jacob314, line 67: The serviceManager.service.getScripts code should move into initState or better yet that logic should move to the controller.
Right now there are a fair number of bugs in that logic related to lifecycles if new isolates load or dependencies change.

kenzieschmoll:
I started working on a CL earlier this week that starts to do some of the lifecycle cleanup. For now I moved get scripts back to initState.

PR #1803, opened 2020-04-08
Consolidate Screen and ConditionalScreen classes.

kenzieschmoll:
This CL also pulls screen definition out into `main.dart` so that conditional screens can be ordered in a more flexible manner.

### DIFF
--- a/packages/devtools_app/lib/src/flutter/app.dart
+++ b/packages/devtools_app/lib/src/flutter/app.dart
@@ -23,7 +23,7 @@ import 'utils.dart';
 /// Top-level configuration for the app.
 @immutable
 class DevToolsApp extends StatefulWidget {
-  const DevToolsApp([this.screens]);
+  const DevToolsApp(this.screens);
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -28,15 +20,12 @@ import 'screen.dart';
...
 import 'utils.dart';
 
-// TODO(bkonyi): remove this bool when page is ready.
-const showNetworkPage = false;
-
 /// Top-level configuration for the app.
 @immutable
 class DevToolsApp extends StatefulWidget {
-  const DevToolsApp([this.conditionalScreens]);
+  const DevToolsApp([this.screens]);
```
devoncarew, line 26: The list of screens is really optional?

kenzieschmoll: good catch! they should not be optional

PR #1805, opened 2020-04-08
connect the devtools server to the flutter web build of the app

devoncarew:
- connect the devtools server to the flutter web build of the app

Portions of this PR are from https://github.com/flutter/devtools/pull/1787; I have some continued iteration to do on this PR before it'll be ready for review.


devoncarew:
cc @DanTup for FYI

DanTup:
@devoncarew what's left to do on this? I did a quick test locally and it seemed to work (the app seemed to connect back to the client and report the events as I moved around).

I'm not sure why the tests failed on CI, the same ones passed for me locally. Are they supposed to be running with the html version or Flutter version?

(I did find one issue with the new version, and that's that it doesn't seem to handle automatically connecting when there's uri= in the URL.. there does seem to be a little code that seems like it's intended to support that, but I couldn't make it work)

devoncarew:
I see failures when run w/ the dart:html version compile to dart2js. I was going to investigate why (you can repro locally by hardcoding `testInReleaseMode` to true in server_test.dart).

If you have bandwidth to take this over that would be great; otherwise I'll investigate later this week.

DanTup:
> you can repro locally by hardcoding testInReleaseMode to true in server_test.dart

Ah, I suspected I was missing a flag for something. I'll take a look, thanks!

DanTup:
@devoncarew I think the issue is that `_setGlobals` was trying to use the connection (to initialise the framework) before the connection was made, so it was skipped.

Moving the initialisation to where the connection is made appears to solve the issue:

https://github.com/DanTup/devtools/commit/d223ec361cf0fb3befa9d9ec7d1eb84ac7d6ef6c

I'm not sure if this is a reasonable fix though, or whether there was a reason this was in `setGlobals`. If so, we might need to expose an event for when the connection is made that could be used to call this.

devoncarew:
Thanks for the investigation! I don't think it's important to init the connection in the setGlobals() call - that was likely an artifact of me iterating towards a general solution here.

devoncarew:
@DanTup - this is ready for review now

DanTup:
I found some issues while testing this (my "fix" fixed the HTML version, but broke the Flutter version, but also some of the API calls aren't working correctly in the Flutter version). Working through them now.

DanTup:
@devoncarew this LGTM with two issues:

1. My "fix" above was bad - it fixed the HTML version and broke the Flutter version. This is because the two run some setup code on opposite orders. A possible best fix is to have the connection initialise itself when it connects and not wait for it for the rest of setup (otherwise a timeout could delay initialisation):

https://github.com/DanTup/devtools/commit/ca8e3566de653d53e8d065f18b3c054e4d4450f5

2. The API changing page doesn't seem to work - there's code to push a history entry (`_pushScreenByPageId`), but it doesn't seem to actually change the current tab. I can take a look at this (as well as trying to get the server tests running against the Flutter version to test all this) in another PR if you want (it requires some other changes, as `?uri=` querystring doesn't seem to work yet either).

DanTup:
I opened https://github.com/flutter/devtools/pull/1886 which is a copy of this + the fix noted above.

Issue #1111, opened 2019-10-01
Add support for offline snapshots when DevTools is not connected to an app


DaveShuckerow:

kenzieschmoll:
As a note, a user should be able to drag and drop onto any page of the app to import snapshots.

jacob314:
@kenzieschmoll can this be closed?

kenzieschmoll:
modified the description to track what is still missing.

PR #1808, opened 2020-04-10
Refactor screen initialization, controller providers, and offline imports.

kenzieschmoll:
This PR
- Refactors the import code so that we are pushing a different `'/snapshot'` route when a file is imported into DevTools. This allows us to use a new set of controllers so that we do not disturb the state of DevTools when an import is loaded, and enables offline imports for conditional screens that implement the handling.
- Allow Screens to provide a ControllerProvider at initialization.
- Enables importing snapshots when DevTools is not connected to an app. Fixes https://github.com/flutter/devtools/issues/1111
- Removes `DevToolsScreenType` class.
- adds an examples directory to put example implementations of classes that are better described through code than comments. Now contains `conditional_screen.dart`


packages/devtools_app/lib/main.dart
```diff
@@ -40,3 +40,37 @@ void main() {
     const DevToolsApp(screens),
   );
 }
+
+// Example of conditional screen that could be added to [screens] above.
```
jacob314, line 44: Code examples should use the dartdoc style for including code snippets used by Flutter.

packages/devtools_app/lib/src/config_specific/flutter/import_export/import_export.dart
```diff
@@ -14,10 +13,8 @@ const nonDevToolsFileMessage = 'The imported file is not a Dart DevTools file.'
     ' At this time, DevTools only supports importing files that were originally'
     ' exported from DevTools.';
 
-String unsupportedDevToolsFileMessage(String devToolsScreen) {
-  return 'Could not import file. The imported file is from "$devToolsScreen", '
-      'which is not supported by this version of Dart DevTools. You may need to'
-      ' upgrade your version of Dart DevTools to view this file.';
+String attemptingToImportMessage(String devToolsScreen) {
+  return 'Attempting to import file for screen with id "$devToolsScreen".';
```
jacob314, line 17: add a TODO to specifically track the devtools versions with the snapshots. That way we can concretely say that "the screen was created with DevTools 0.3 but you opened with 0.2 which is likely the reason the screen is missing.

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
```diff
@@ -22,15 +22,12 @@ import '../../ui/theme.dart';
...
 class DebuggerScreen extends Screen {
-  const DebuggerScreen()
-      : super(
-          DevToolsScreenType.debugger,
-          title: 'Debugger',
-          icon: Octicons.bug,
-        );
+  const DebuggerScreen() : super(id, title: 'Debugger', icon: Octicons.bug);
+
+  static const id = 'debugger';
```
jacob314, line 27: nit: here and elsewhere it seems like we are defining the id variable more times than we need to.

If you change the constructor call to 
`const DebuggerScreen() : super(id: 'debugger', title: 'Debugger', icon: Octicons.bug);`
you can delete the const id and the override of the dogPageId member.

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -107,8 +118,18 @@ class DevToolsAppState extends State<DevToolsApp> {
               ],
             ),
           ),
-      '/connect': (_, __) =>
+      connectRoute: (_, __, ___) =>
           DevToolsScaffold.withChild(child: ConnectScreenBody()),
+      snapshotRoute: (_, __, args) {
+        return DevToolsScaffold.withChild(
+          child: Controllers.offline(
```
jacob314, line 125: nice!

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -151,10 +172,12 @@ class DevToolsAppState extends State<DevToolsApp> {
   }
 }
 
-/// A [WidgetBuilder] that takes an additional map of URL query parameters.
+/// A [WidgetBuilder] that takes an additional map of URL query parameters and
+/// args.
 typedef UrlParametersBuilder = Widget Function(
   BuildContext,
   Map<String, String>,
+  Object args,
```
jacob314, line 180: make args a named parameter and give it type SnapshotArguments if that is the only case we are using it?

packages/devtools_app/lib/src/flutter/controllers.dart
```diff
@@ -55,6 +53,25 @@ class ProvidedControllers implements DisposableController {
...
+      logging: null,
+      timeline: TimelineController(),
+      memory: null,
+      performance: null,
+      debugger: null,
+      bannerMessages: null,
+    );
+  }
+
+  Map<String, OfflineControllerMixin> get offlineLookup => {
```
jacob314, line 71: this seems redundant with the ProvidedControllers factory constructor for offline which provides the same data.

packages/devtools_app/lib/src/flutter/scaffold.dart
```diff
@@ -172,19 +171,25 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
...
-  /// Pushes a screen for an offline import.
-  void _pushScreenForImport(DevToolsScreenType screenType) {
+  /// Pushes the snapshot screen for an offline import.
+  void _pushSnapshotScreenForImport(
+    String screenId,
+    Object data,
+  ) {
+    final args = SnapshotArguments(screenId, data);
+    if (offlineMode) {
+      // If we are already in offline mode, only handle routing from existing
```
jacob314, line 181: what happens if we don't do this?

packages/devtools_app/lib/src/flutter/screen.dart
```diff
@@ -4,30 +4,39 @@
...
+  }) : this(
+          id,
+          conditionalLibrary: conditionalLibrary,
+          conditionalController: conditionalController,
+          title: title,
+          icon: icon,
+          tabKey: tabKey,
+        );
+
+  final String screenId;
```
jacob314, line 39: can we merge screenId and id?

packages/devtools_app/lib/src/flutter/screen.dart
```diff
@@ -48,6 +57,17 @@ abstract class Screen {
...
+  /// Screen controller for conditional screens.
+  ///
+  /// When non-null, this controller will be used for processing data from
+  /// offline imports. In order for conditional screens to have offline support,
+  /// the screen content returned by [build] needs to use this controller to
+  /// manage data.
+  ///
+  /// If null, we will use the corresponding controller from
+  /// [ProvidedControllers.offline] to process offline data for this screen.
+  final OfflineControllerMixin conditionalController;
```
jacob314, line 69: why do offline controllers for conditinoal imports need to be provided earlier than other controllers? It would be nice to work around that issue if practical instead of special casing how the offline controller needs to be provided for the conditional use case.

packages/devtools_app/lib/src/flutter/snapshot_screen.dart
```diff
@@ -0,0 +1,147 @@
...
+      (screen) => screen.screenId == widget.args?.screenId,
+      orElse: () => null,
+    );
+  }
+
+  // TODO(kenz): ensure conditional screens have a way to provide the controller
+  // they need since it will not be provided by `Controller.of(context)` as is.
+  Future<void> _buildScreenAndProcessData() async {
+    if (_screen != null && widget.args?.data != null) {
+      _screenController = _screen.conditionalController ??
```
jacob314, line 80: can we perform this task slightly later when Controllers.of(context) is always available so we don't have to flow the data in two different ways which makes the code harder to maintain?

packages/devtools_app/lib/src/flutter/snapshot_screen.dart
```diff
@@ -0,0 +1,147 @@
...
+      _screen = null;
+      _screenController = null;
+    });
+  }
+}
+
+class SnapshotArguments {
+  SnapshotArguments(this.screenId, this.data);
+  final String screenId;
+  final Object data;
```
jacob314, line 146: can data be given type Map<String, Object> if it is always really a JSON map?

packages/devtools_app/lib/src/info/flutter/info_screen.dart
```diff
@@ -16,10 +16,11 @@ import '../info_controller.dart';
 class InfoScreen extends Screen {
   const InfoScreen()
       : super(
-          DevToolsScreenType.info,
+          id,
           title: 'Info',
           icon: Octicons.info,
         );
+  static const id = 'info';
```
jacob314, line 23: as in the other cases, remove this const and write the value inline in the constructor.

packages/devtools_app/lib/src/flutter/app.dart
kenzieschmoll: done

packages/devtools_app/lib/src/config_specific/flutter/import_export/import_export.dart
kenzieschmoll: added a todo for better error handling and included the devtools version in the export code

packages/devtools_app/lib/src/flutter/scaffold.dart
kenzieschmoll: Added a comment to explain. copying and pasting here:

If we allow other routes that are not the '/snapshot' route to handle
       routing when we are already offline, the other routes will pop their
       existing screen ('/connect', or '/') and push '/snapshot' over the top.
       We want to avoid this because the routes underneath the existing
       '/snapshot' route should remain unchanged while '/snapshot' sits on
       top.

kenzieschmoll:
@jacob314 implemented changes we discussed offline to make each screen controller accessible via its own provider, which conditional screens can also provide. Conditional screens will now also be able to support offline imports if they implement the handling. PTAL.

packages/devtools_app/lib/src/logging/flutter/logging_controller.dart
```diff
@@ -0,0 +1,1000 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: copy of logging/logging_controller.dart to avoid breaking the html app. Changes outlined by comments.

packages/devtools_app/lib/src/logging/flutter/logging_controller.dart
```diff
@@ -0,0 +1,1000 @@
...
+  InstanceRef stringRef,
+) {
+  return service.retrieveFullStringValue(
+    isolateRef.id,
+    stringRef,
+    onUnavailable: (truncatedValue) => '${stringRef.valueAsString}...',
+  );
+}
+
+class LoggingControllerProvider extends ControllerProvider {
```
kenzieschmoll, line 57: start of new code

packages/devtools_app/lib/src/logging/flutter/logging_controller.dart
```diff
@@ -0,0 +1,1000 @@
...
+  const _InheritedLoggingController(
+      {@required this.data, @required Widget child})
+      : super(child: child);
+
+  final LoggingController data;
+
+  @override
+  bool updateShouldNotify(_InheritedLoggingController oldWidget) =>
+      oldWidget.data != data;
+}
```
kenzieschmoll, line 123: end of new code

packages/devtools_app/lib/src/performance/flutter/performance_controller.dart
```diff
@@ -0,0 +1,110 @@
...
+import 'package:meta/meta.dart';
+
+import '../../flutter/controllers.dart';
+import '../../globals.dart';
+import '../../profiler/cpu_profile_controller.dart';
+import '../../profiler/cpu_profile_model.dart';
+import '../../ui/fake_flutter/fake_flutter.dart';
+import '../../utils.dart';
+
+class PerformanceControllerProvider extends ControllerProvider {
```
kenzieschmoll, line 16: start of new code

packages/devtools_app/lib/src/performance/flutter/performance_controller.dart
```diff
@@ -0,0 +1,110 @@
...
+  const _InheritedPerformanceController(
+      {@required this.data, @required Widget child})
+      : super(child: child);
+
+  final PerformanceController data;
+
+  @override
+  bool updateShouldNotify(_InheritedPerformanceController oldWidget) =>
+      oldWidget.data != data;
+}
```
kenzieschmoll, line 76: end of new code

packages/devtools_app/lib/src/config_specific/flutter/import_export/_export_desktop.dart
```diff
@@ -15,7 +15,8 @@ class ExportControllerDesktop extends ExportController {
   static final _fs = FileIO();
 
   @override
-  void downloadFile(String filename, String contents) {
+  String downloadFile(String contents) {
```
jacob314, line 18: returning the filename from a method called downloadFile is surprising to me. Can you leave the method as void?

packages/devtools_app/lib/src/config_specific/flutter/import_export/import_export.dart
```diff
@@ -75,5 +63,29 @@ abstract class ExportController {
 
   const ExportController.impl();
 
-  void downloadFile(String filename, String contents);
+  String get filename {
```
jacob314, line 69: this shouldn't be a getter because it returns a new value every time it is called. Getters should whenever possible behave like fields which have the same value if you call them twice in a row.
something like
```dart
String generateFilename()
```
would make the intent more clear.

packages/devtools_app/lib/src/flutter/controllers.dart
```diff
@@ -207,53 +137,17 @@ class _InheritedProvider extends InheritedWidget {
...
-    SchedulerBinding.instance.addPostFrameCallback((_) {
-      _oldData?.dispose();
-    });
-  }
-}
-
-mixin OfflineControllerMixin<T> {
-  FutureOr<void> processOfflineData(T offlineData);
+/// For an example implementation, see `src/example/provided_controller.dart`.
+abstract class ControllerProvider extends StatefulWidget {
```
jacob314, line 150: Devtools is now depending on package:provider. I expect that will let you reduce the boilerplate defining all these providers manually.
https://pub.dev/packages/provider

packages/devtools_app/lib/src/flutter/screen.dart
```diff
@@ -111,3 +98,21 @@ abstract class Screen {
     return null;
   }
 }
+
+mixin OfflineScreenMixin<T extends StatefulWidget, U> on State<T> {
+  bool loadingOfflineData = false;
```
jacob314, line 103: make this a getter and private field. That way there is no way a caller can accidentally set the value.

packages/devtools_app/lib/src/config_specific/flutter/import_export/_export_desktop.dart
kenzieschmoll: we show a notification "Successfully exported $exportedFile to ~/Downloads directory", so we need the exported file name

packages/devtools_app/lib/src/config_specific/flutter/import_export/import_export.dart
kenzieschmoll: done

packages/devtools_app/lib/src/flutter/screen.dart
kenzieschmoll: done

packages/devtools_app/lib/src/flutter/controllers.dart
kenzieschmoll: good call! eliminated a lot of code. PTAL.

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
kenzieschmoll: removed id where we can. In some cases, we want to be able to access the static screen id for things like exporting / importing json and setting a screen id for banner messages

packages/devtools_app/lib/src/flutter/controllers.dart
```diff
@@ -144,10 +145,23 @@ class _InheritedProvider extends InheritedWidget {
...
 
-/// Superclass for all DevTools controller providers.
+/// Controller provider that [DevToolsScreen] instances can provide.
 ///
-/// For an example implementation, see `src/example/provided_controller.dart`.
-abstract class ControllerProvider extends StatefulWidget {
-  const ControllerProvider({Key key, this.child}) : super(key: key);
+/// Flutter UI components with this [ControllerProvider] as an ancestor can
+/// access the provided controller via `Provider.of<T>(context)`.
+class ControllerProvider<T> extends StatelessWidget {
```
jacob314: We can remove this ControllerProvider class as it isn't providing value beyond the base Provider class. Instead we should just pass in the `create` Functions which gives a cleaner API anyway.

packages/devtools_app/lib/main.dart
```diff
@@ -38,33 +38,57 @@ void main() {
     const DevToolsScreen(screen: InspectorScreen()),
     DevToolsScreen(
       screen: const TimelineScreen(),
-      controllerProvider: (child) => TimelineControllerProvider(child: child),
+      controllerProvider: (child) => ControllerProvider<TimelineController>(
```
jacob314: you can replace the controllerProvider line with 
createController: (_) =>TimelineController()
and then in the implementation of DevToolsScreen create a Provider<TController> that passes in the createController function as the `create` argument.

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -188,23 +187,33 @@ class DevToolsAppState extends State<DevToolsApp> {
...
 }
 
-class DevToolsScreen {
-  const DevToolsScreen({
-    @required this.screen,
-    this.controllerProvider,
+class DevToolsScreen<T> {
+  const DevToolsScreen(
+    this.screen, {
+    this.createController,
```
jacob314, line 193: nit: might be worthwhile to make this required and force users to pass in null if they don't want a controller just as a signal that this is something users should think about.

packages/devtools_app/lib/src/flutter/app.dart
```diff
@@ -188,23 +187,33 @@ class DevToolsAppState extends State<DevToolsApp> {
   }
 }
 
-class DevToolsScreen {
-  const DevToolsScreen({
-    @required this.screen,
-    this.controllerProvider,
+class DevToolsScreen<T> {
```
jacob314, line 190: Document what T is as part of the class level comment. Consider changing it to `C` as it is the Controller which might be a little unexpected.

PR #1812, opened 2020-04-12
update the .gitignore file to exclude macos/Runner/

devoncarew:
- update the .gitignore file to exclude macos/Runner/

cc @terrylucas


devoncarew:
(TBR)

PR #1817, opened 2020-04-14
Debugger state management cleanup

kenzieschmoll:
- Moves business logic into controller from screen
- Use ValueListenable pattern to update UI
- Add `debugger_service.dart` file to interact with vm_service (follows pattern used in the rest of DevTools)

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
@@ -9,25 +9,19 @@ import 'package:vm_service/vm_service.dart';
 
 import '../../auto_dispose.dart';
 import '../../globals.dart';
+import 'debugger_service.dart';
 
 /// Responsible for managing the debug state of the app.
 class DebuggerController extends DisposableController
     with AutoDisposeControllerMixin {
   DebuggerController() {
     switchToIsolate(serviceManager.isolateManager.selectedIsolate);
-    autoDispose(serviceManager.isolateManager.onSelectedIsolateChanged
-        .listen(switchToIsolate));
-
-    autoDispose(_service.onDebugEvent.listen(_handleIsolateEvent));
+    autoDispose(
+        debuggerService.onSelectedIsolateChanged.listen(switchToIsolate));
+    autoDispose(debuggerService.onDebugEvent.listen(_handleIsolateEvent));
   }
 
-  // TODO(djshuckerow): Separate the controller from a debugging service object
-  // such that the controller does not need to know about VmService.
-  VmService get _service => serviceManager.service;
-  InstanceRef get reportedException => _reportedException;
-
-  IsolateRef isolateRef;
-  List<ScriptRef> scripts;
+  final debuggerService = DebuggerService();
 
   final _scriptCache = <String, Script>{};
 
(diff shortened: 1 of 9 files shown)


packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
```diff
@@ -88,60 +74,52 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
...
-    } else {
+    });
+    addAutoDisposeListener(controller.scriptList, () {
       setState(() {
-        stack = null;
+        scriptList = controller.scriptList.value;
       });
-    }
+    });
+    breakpoints = controller.breakpoints.value;
```
jacob314, line 93: is there a reason we can't just use the breakpoints valuelistenable from the controller?
Same question for the scriptList.
It would be nice to not have two copies of the data.

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
```diff
@@ -88,60 +74,52 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
...
   }
 
-  Map<int, Breakpoint> getBreakpointsForLines() {
+  Map<int, Breakpoint> _breakpointsForLines() {
     if (script == null) return {};
-    String getScriptId(Breakpoint b) => b.location.script.id;
-
-    final bps = controller.breakpoints.value
-        .where((b) => b != null && getScriptId(b) == script.id);
+    final bps = breakpoints
```
jacob314, line 114: not caused by you but while you are here can you cleanup the `bps` variable name or eliminate it completely by rolling it right into the for loop.

packages/devtools_app/lib/src/debugger/flutter/debugger_service.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:vm_service/vm_service.dart';
+
+import '../../globals.dart';
+
+class DebuggerService {
```
jacob314, line 5: What purpose is the DebuggerService class serving? It seems to mainly just be forwarding messages from serviceManager.service.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -9,25 +9,19 @@ import 'package:vm_service/vm_service.dart';
...
   }
 
-  // TODO(djshuckerow): Separate the controller from a debugging service object
-  // such that the controller does not need to know about VmService.
-  VmService get _service => serviceManager.service;
-  InstanceRef get reportedException => _reportedException;
-
-  IsolateRef isolateRef;
-  List<ScriptRef> scripts;
+  final debuggerService = DebuggerService();
```
jacob314, line 24: All parts of this refactor are great except for adding the debuggerService class. Lets double down on everyone working on this code being familiar with the VmService API and avoid writing abstractions on top of it when they don't add clear value.  I would leave the`VmService get _service => ...` getter and delete the debuggerService field.
If you want the DebuggerService class for mocking and testing then I could be convinced but  it would be good to refactor the API a little so it is clear what the sependencies of the DebuggerService are. Currently it is getting all its data behind the scenes from the serviceManager singleton.

Issue #1648, opened 2020-02-20
Debugger page rewrite in Flutter


DaveShuckerow:
# Debugger page Flutter web rewrite

### General
- [x] when initing state, have the script for the rootLib selected
- [x] address some state management issues @kenzieschmoll 
- [x] the debugger should listen for reload events and see if any script source needs to be updated from the VM (`Isolate` / `IsolateReload` kind / `status: "success"`)
- [ ] debounce some of the debugger events (optionally) so things like stepping are smoother (https://github.com/flutter/devtools/issues/1924)

### Call stack 
- [x] implement @kenzieschmoll 

### Variables
- [x] implement @kenzieschmoll 

### breakpoints
- [x] improve display
- [x] update once the source line position has been resolved
- [x] show a count in the area title
- [x] click should jump to the source location
- [x] setting a breakpoint on an unsupported line should set the breakpoint anyway even if it is a noop (ed: we now silently ignore the breakpoint attempt; we could change to provide more feedback to the user)
- [x] have a way to clear all breakpoints
- [ ] persist breakpoints across sessions, per-app (https://github.com/flutter/devtools/issues/1925)

### search in the running app
- [x] improve scripts list display
- [x] move the libraries view to the right side of the screen, have it hidden by default, and have an action button to expand it out in order to search for libraries
- [x] allow searching for classes

### debugger buttons
- [x] Debugging controls for pause, resume, step in, step out, and step over
- [x] improve the appearance so they don't resize on mouse hover
- [x] show a control to configure break on exceptions

### code area
- [x] A view that can show code in Flutter (#1649)
- [x] Line numbering and inline breakpoint markers
- [x] show breakable lines in the gutter
- [x] display the char column of the execution point
- [ ] have forward / back nav. history buttons in the title area (https://github.com/flutter/devtools/issues/1929)
- [ ] show library imports, classes as hyperlinks in the view (based on VM info) (https://github.com/flutter/devtools/issues/1928)
- [ ] syntax highlighting (https://github.com/flutter/devtools/issues/1927)

### console area
- [x] display stdout and stderr text
- [ ] allow for expression evaluation @devoncarew (https://github.com/flutter/devtools/issues/1926)

---

Reference image for the old Flutter plugin view:

![image](https://user-images.githubusercontent.com/14947366/67529782-05788580-f672-11e9-923e-97ca677b8ed2.png)

jonahwilliams:
[...]

PR #1818, opened 2020-04-14
improve the look of the scripts view

devoncarew:
Improve the look of the scripts view:
- give slightly more room to the scripts view
- don't use horizontal scrolling, but allow horizontal overflow
- make the filter area slightly smaller
- add a badge for the script count in the script area header
- sort the scripts in the debugger controller
- make similar horizontal scrolling changes to the code area
- general work towards https://github.com/flutter/devtools/issues/1648

Screen Shot 2020-04-14 at 4 16 49 PM

Some future improvements are to make the badge slightly smaller (perhaps use https://pub.dev/packages/badges ?), and to remove the dedicated filter area and move it to the script area header.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/codeview.dart
@@ -245,6 +245,8 @@ class ScriptRow extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    // TODO(#1648): Implement syntax highlighting.
+
     return InkWell(
       onTap: onPressed,
       child: Container(
--- a/packages/devtools_app/lib/src/debugger/flutter/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/codeview.dart
@@ -163,6 +163,8 @@ class _CodeViewState extends State<CodeView> {
               child: SingleChildScrollView(
                 controller: _horizontalController,
                 scrollDirection: Axis.horizontal,
+                // TODO: Improve our sizing here - currently we do
+                // CodeView.assumedCharacterWidth * max line width
                 child: SizedBox(
                   width: CodeView.assumedCharacterWidth *
                       lines
--- a/packages/devtools_app/lib/src/debugger/flutter/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/codeview.dart
@@ -111,6 +111,8 @@ class _CodeViewState extends State<CodeView> {
 
   @override
   Widget build(BuildContext context) {
+    // TODO(#1648): Implement syntax highlighting.
+
     if (widget.script == null) {
       return Center(
         child: Text(
@@ -245,8 +247,6 @@ class ScriptRow extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    // TODO(#1648): Implement syntax highlighting.
-
     return InkWell(
       onTap: onPressed,
       child: Container(
--- a/packages/devtools_app/lib/src/debugger/flutter/common.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/common.dart
@@ -14,26 +14,6 @@ Widget densePadding(Widget child) {
   );
 }
 
-class Badge extends StatelessWidget {
-  const Badge({@required this.child});
-
-  final Widget child;
-
-  @override
-  Widget build(BuildContext context) {
-    // TODO(devoncarew): We'll likely want a badge implementation that's
-    // separate from [Chip].
-
-    return SizedBox(
-      child: Chip(
-        label: child,
-        visualDensity: VisualDensity.compact,
-        padding: const EdgeInsets.symmetric(vertical: -4.0),
-      ),
-    );
-  }
-}
-
 /// Create a header area for a debugger component.
 ///
 /// Either one of [text] or [child] must be supplied.
--- a/packages/devtools_app/lib/src/debugger/flutter/common.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/common.dart
@@ -14,6 +14,21 @@ Widget densePadding(Widget child) {
   );
 }
 
+class Badge extends StatelessWidget {
+  const Badge({@required this.child});
+
+  final Widget child;
+
+  @override
+  Widget build(BuildContext context) {
+    return Chip(
+      label: child,
+      visualDensity: VisualDensity.compact,
+      backgroundColor: Theme.of(context).accentColor,
+    );
+  }
+}
+
 /// Create a header area for a debugger component.
 ///
 /// Either one of [text] or [child] must be supplied.
@@ -31,11 +46,6 @@ Container debuggerSectionTitle(ThemeData theme, {String text, Widget child}) {
     padding: const EdgeInsets.only(left: defaultSpacing),
     alignment: Alignment.centerLeft,
     height: DebuggerScreen.debuggerPaneHeaderHeight,
-    child: child != null
-        ? child
-        : Text(
-            text,
-            style: theme.textTheme.subtitle2,
-          ),
+    child: child != null ? child : Text(text, style: theme.textTheme.subtitle2),
   );
 }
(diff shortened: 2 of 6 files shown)


packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
```diff
@@ -210,12 +216,19 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
...
         ),
-        padding: const EdgeInsets.only(left: defaultSpacing),
+        padding: const EdgeInsets.only(left: defaultSpacing, right: 4.0),
         alignment: Alignment.centerLeft,
         height: DebuggerScreen.debuggerPaneHeaderHeight,
-        child: Text(
-          title,
-          style: Theme.of(context).textTheme.subtitle2,
+        child: Row(
+          children: <Widget>[
```
kenzieschmoll, line 223: remove `<Widget>`

packages/devtools_app/lib/src/debugger/flutter/scripts.dart
```diff
@@ -6,85 +6,105 @@ import 'package:flutter/material.dart';
...
+  final void Function(ScriptRef scriptRef) onSelected;
 
   @override
   ScriptPickerState createState() => ScriptPickerState();
 }
 
 class ScriptPickerState extends State<ScriptPicker> {
-  List<ScriptRef> _filtered;
-  TextEditingController filterController = TextEditingController();
+  final TextEditingController _filterController = TextEditingController();
```
kenzieschmoll, line 33: remove type on LHS

packages/devtools_app/lib/src/debugger/flutter/codeview.dart
```diff
@@ -114,8 +111,6 @@ class _CodeViewState extends State<CodeView> {
 
   @override
   Widget build(BuildContext context) {
-    // TODO(https://github.com/flutter/devtools/issues/1648): Line numbers,
-    // syntax highlighting and breakpoint markers.
```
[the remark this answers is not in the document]
devoncarew: I can re-add this; I think most of the work is also tracked in the github issue.

packages/devtools_app/lib/src/debugger/flutter/codeview.dart
```diff
@@ -162,28 +157,18 @@ class _CodeViewState extends State<CodeView> {
             ),
             const SizedBox(width: denseSpacing),
             Expanded(
-              child: SingleChildScrollView(
```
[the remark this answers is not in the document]
devoncarew: I think this is an improvement from a presentation POV in the lists on the left. I'm less certain its a net win for the code area. The issue that I saw here was that we didn't know how wide the text display was, so we had a very large estimate (the longest line in the file * the est. max width of a char). This meant that we overscrolled 2-3x horizontally.

I think I'd prefer not having horizontal scrolling for now (and using ellipsis overflow, w/ the user having the ability to re-size their window and the code area) to the current est. horizontal size behavior.

In the future, we should likely move to a precise horizontal scroll - knowing how long the longest line is in pixels, and allowing users to scroll that far.

packages/devtools_app/lib/src/debugger/flutter/common.dart
```diff
@@ -14,6 +14,26 @@ Widget densePadding(Widget child) {
   );
 }
 
+class Badge extends StatelessWidget {
```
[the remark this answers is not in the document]
devoncarew: will do

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
```diff
@@ -78,18 +78,20 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
       });
     });
     addAutoDisposeListener(controller.currentStack);
-    addAutoDisposeListener(controller.scriptList);
+    addAutoDisposeListener(controller.sortedScripts);
```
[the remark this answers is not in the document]
devoncarew: No, just sortedScripts now.

In future work, we should be more precise about what we invalidate on changes. Currently rebuild more than we need to on stack and breakpoint changes.

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
devoncarew: My code completion is very insistent here for some reason.

devoncarew: Ah!

@pq, I still see the generic types inserted with the 'wrap with' refactorings (Row, Column, ...).

packages/devtools_app/lib/src/debugger/flutter/scripts.dart
devoncarew: done

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
pq: Ah!  Missed that.

Looking now (https://github.com/dart-lang/sdk/issues/41517).

packages/devtools_app/lib/src/debugger/flutter/codeview.dart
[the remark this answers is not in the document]
jacob314: Can you file a bug to re-enable horizontal scrolling? What you suggested sounds like exactly the right way to do this. If we assume a fixed with font, that should be easy to do efficiently.

packages/devtools_app/lib/src/debugger/flutter/scripts.dart
```diff
@@ -5,86 +5,106 @@
...
   @override
   void didUpdateWidget(ScriptPicker oldWidget) {
     super.didUpdateWidget(oldWidget);
 
-    if (_isNotLoaded) {
-      initFilter();
-    } else if (oldWidget.scripts != widget.scripts) {
-      updateFilter(filterController.text);
-    }
+    updateFilter();
```
jacob314, line 47: nit: it would be idiomatic to 
only call updateFilter if
`widget.scripts != oldWidget.scripts`
assuming that object equality is a safe way to check whether the list of scripts changed.

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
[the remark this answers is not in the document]
kenzieschmoll: Can you add a TODO for this? Also, can we remove the scriptList Notifier / listener on the controller then?

packages/devtools_app/test/flutter/debugger_screen_test.dart
```diff
@@ -46,6 +46,7 @@ void main() {
       when(debuggerController.breakpoints).thenReturn(ValueNotifier([]));
       when(debuggerController.scriptList)
           .thenReturn(ValueNotifier(ScriptList(scripts: [])));
+      when(debuggerController.sortedScripts).thenReturn(ValueNotifier([]));
```
jacob314, line 49: can we add a test where the sortedScripts list is not empty?

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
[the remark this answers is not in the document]
devoncarew: Yup, will do re: the todo.

Re: the `scriptList` Notifier - there aren't any current uses of it. However, it seems useful in terms of listening for basic VM event changes.

I think we'll also need more semantic changes - processed versions of the script lists, processed versions of the breakpoints list. Also things that are closer to UI state - the currently selected script +file location for instance. I'm not sure we want to track all those on the debugger controller, or have it provide more basic info and have another object which tracks post-processed data.

packages/devtools_app/lib/src/debugger/flutter/scripts.dart
devoncarew: OK, it looks like this method means that we're re-using the state object from an older statefulwidget.

I don't think object identify will work for us here (wrt the list of scripts). For the future though, we'd need to check both the list of scripts, and the filter controller have the same value? Or, the two fields on the `widget` - the scripts and the `selected` script?

packages/devtools_app/test/flutter/debugger_screen_test.dart
devoncarew: done!

PR #1828, opened 2020-04-16
Track PIDs of browsers spawned during tests and clean them up

DanTup:
I've noticed a few recent builds have been flaky (failing to connect to Chrome debug ports), especially since #1826 (which made more Chrome tests run). I think this may be related to the server spawning instances of Chrome, but not terminating them.

I'm not sure we want to automatically terminate them when the server quits (the user may have opened new tabs and be doing things in them), so I've added the `pid` to the return package so the client (or in this case, the tests) can decide whether to clean them up.

Running this locally prevents Chrome processes being left around after the tests.


DanTup:
Actually, I don't think these tests run on the bots I saw flake - they were test_dart2js and test_ddc, but these run in integration_ddc and integration_dart2js. So it probably doesn't explain the flakes, but it probably is worth having to avoid tests getting slower over time as we leave more Chrome instances around).

PR #1829, opened 2020-04-16
Add arrow left/right navigation to tree table

### LINKED ISSUE #1769: Support arrow key traversal in Tree Tables
This is implemented in the dart:html app. We likely need to add a "selected" state to the tree table items so that we have a modifiable starting point of traversal.

gmoothart:
Left key collapses the current node or selects the node's parent.
Right key expands the current node, or selects its first child.

Also fixed a math error in scrolling and added comments plus general
cleanup.

Fixes  #1769

### DIFF
--- a/packages/devtools_app/lib/src/flutter/table.dart
+++ b/packages/devtools_app/lib/src/flutter/table.dart
@@ -1,5 +1,4 @@
 import 'dart:math';
-import 'dart:ui';
 
 import 'package:flutter/material.dart' hide TableRow;
 import 'package:flutter/services.dart';
@@ -457,26 +456,28 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
       _moveSelection(ScrollKind.down, scrollController, constraints.maxHeight);
     } else if (event.logicalKey == LogicalKeyboardKey.arrowUp) {
       _moveSelection(ScrollKind.up, scrollController, constraints.maxHeight);
-    }
-
-    // On left arrow collapse the row if it is expanded. If it is not, move
-    // selection to its parent.
-    else if (event.logicalKey == LogicalKeyboardKey.arrowLeft) {
+    } else if (event.logicalKey == LogicalKeyboardKey.arrowLeft) {
+      // On left arrow collapse the row if it is expanded. If it is not, move
+      // selection to its parent.
       if (selectedNode.isExpanded) {
         _toggleNode(selectedNode);
       } else {
         _moveSelection(
-            ScrollKind.parent, scrollController, constraints.maxHeight);
+          ScrollKind.parent,
+          scrollController,
+          constraints.maxHeight,
+        );
       }
-    }
-
-    // On right arrow expand the row if possible, otherwise move selection down.
-    else if (event.logicalKey == LogicalKeyboardKey.arrowRight) {
+    } else if (event.logicalKey == LogicalKeyboardKey.arrowRight) {
+      // On right arrow expand the row if possible, otherwise move selection down.
       if (!selectedNode.isExpanded) {
         _toggleNode(selectedNode);
       } else {
         _moveSelection(
-            ScrollKind.down, scrollController, constraints.maxHeight);
+          ScrollKind.down,
+          scrollController,
+          constraints.maxHeight,
+        );
       }
     }
   }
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -1,4 +1,5 @@
 import 'dart:math';
+import 'dart:ui';
```
jacob314, line 2: nit: by convention typically `import 'dart:ui' as ui` as it has some names that can overlap with other libraries we use.

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -446,36 +453,59 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
...
+    else if (event.logicalKey == LogicalKeyboardKey.arrowLeft) {
+      if (selectedNode.isExpanded) {
+        _toggleNode(selectedNode);
+      } else {
+        _moveSelection(
+            ScrollKind.parent, scrollController, constraints.maxHeight);
+      }
+    }
+
+    // On right arrow expand the row if possible, otherwise move selection down.
```
jacob314, line 473: good comments!

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -446,36 +453,59 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     }
     assert(selectedNode == items[selectedNodeIndex]);
 
-    if (event.logicalKey == LogicalKeyboardKey.arrowDown ||
-        event.logicalKey == LogicalKeyboardKey.arrowUp) {
-      _moveSelection(event.logicalKey, scrollController, constraints.maxHeight);
+    if (event.logicalKey == LogicalKeyboardKey.arrowDown) {
```
kenzieschmoll, line 456: let's use a switch statement instead 
`switch (event.logicalKey) { ... }`

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -446,36 +453,59 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
...
+    // '-1' in the following calculations is needed because the header row
+    // occupies space in the viewport so we must subtract it out.
     final minCompleteItemsInView =
         (viewportHeight / _Table.defaultRowHeight).floor() - 1;
     final lastItemIndex = firstItemIndex + minCompleteItemsInView - 1;
     int newSelectedNodeIndex;
 
-    if (key == LogicalKeyboardKey.arrowDown) {
-      // move selection down if there is a node lower down.
+    if (scrollKind == ScrollKind.down) {
```
kenzieschmoll, line 500: could use a switch statement here as well to make it clear that there are three possible cases

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -446,36 +453,59 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
...
+    }
+
+    // On left arrow collapse the row if it is expanded. If it is not, move
+    // selection to its parent.
+    else if (event.logicalKey == LogicalKeyboardKey.arrowLeft) {
+      if (selectedNode.isExpanded) {
+        _toggleNode(selectedNode);
+      } else {
+        _moveSelection(
+            ScrollKind.parent, scrollController, constraints.maxHeight);
```
kenzieschmoll, line 469: trailing comma

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -446,36 +453,59 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
...
+      }
+    }
+
+    // On right arrow expand the row if possible, otherwise move selection down.
+    else if (event.logicalKey == LogicalKeyboardKey.arrowRight) {
+      if (!selectedNode.isExpanded) {
+        _toggleNode(selectedNode);
+      } else {
+        _moveSelection(
+            ScrollKind.down, scrollController, constraints.maxHeight);
```
kenzieschmoll, line 479: trailing comma

packages/devtools_app/lib/src/flutter/table.dart
gmoothart: I actually tried to do this, and got an error: `The switch case expression type 'LogicalKeyboardKey' can't override the == operator.`

LMK if there is a way to get around this.

kenzieschmoll: strange. not a blocker feel free to land without this change

packages/devtools_app/lib/src/flutter/table.dart
gmoothart: It turns out this line wasn't necessary.

packages/devtools_app/lib/src/flutter/table.dart
gmoothart: I moved things around to chain the if/else blocks more obviously.

packages/devtools_app/lib/src/flutter/table.dart
gmoothart: Done.

packages/devtools_app/lib/src/flutter/table.dart
gmoothart: Done

packages/devtools_app/lib/src/flutter/table.dart
gmoothart: Done

PR #1830, opened 2020-04-17
implement the debugger's console area

devoncarew:
Implement the debugger's console area:
- display stdout and stderr output from the process under debug in the 'Console' area
- auto-scroll content to the bottom when content is added (unless the user has manually scrolled back up a bit)
- have a short scroll animation when scrolling to show new content

I don't know if we have any guidance about our animations. I used a 100ms `Curves.linear` animation - short, and doesn't distract from the what the user's doing.

@kenzieschmoll @jacob314

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -3,24 +3,85 @@
 // found in the LICENSE file.
 
 import 'package:flutter/material.dart';
+import 'package:flutter/scheduler.dart';
 
 import '../../flutter/common_widgets.dart';
+import '../../flutter/theme.dart';
 import 'common.dart';
+import 'debugger_controller.dart';
 
-class Console extends StatelessWidget {
+// TODO(devoncarew): Allow scrolling horizontally as well.
+
+// TODO(devoncarew): Show some small UI indicator when we receive stdout/stderr.
+
+// TODO(devoncarew): Support hyperlinking to stack traces.
+
+/// Display the stdout and stderr output from the process under debug.
+class Console extends StatefulWidget {
+  const Console({
+    Key key,
+    this.controller,
+  }) : super(key: key);
+
+  final DebuggerController controller;
+
+  @override
+  _ConsoleState createState() => _ConsoleState();
+}
+
+class _ConsoleState extends State<Console> {
+  final scrollController = ScrollController();
+
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
+    final textStyle =
+        theme.textTheme.bodyText2.copyWith(fontFamily: 'RobotoMono');
 
     return OutlinedBorder(
       child: Column(
+        crossAxisAlignment: CrossAxisAlignment.stretch,
         children: [
           debuggerSectionTitle(theme, text: 'Console'),
-          const Expanded(
-            child: Center(child: Text('todo:')),
+          Expanded(
+            child: Padding(
+              padding: const EdgeInsets.all(denseSpacing),
+              child: SingleChildScrollView(
+                controller: scrollController,
+                child: ValueListenableBuilder(
+                  valueListenable: widget.controller.stdio,
+                  builder: (context, value, _) {
+                    // If we're at the end already, scroll to expose the new
+                    // content.
+                    final pos = scrollController.position;
+                    if (pos.pixels == pos.maxScrollExtent) {
+                      SchedulerBinding.instance.addPostFrameCallback((_) {
+                        _scrollToBottom();
+                      });
+                    }
+
+                    return Text(value, style: textStyle);
+                  },
+                ),
+              ),
+            ),
           ),
         ],
       ),
     );
+  }
+
+  void _scrollToBottom() async {
+    await scrollController.animateTo(
+      scrollController.position.maxScrollExtent,
+      duration: const Duration(milliseconds: 100),
+      curve: Curves.linear,
+    );
+
+    // Scroll again if we've received new content in the interim.
+    final pos = scrollController.position;
+    if (pos.pixels != pos.maxScrollExtent) {
+      scrollController.jumpTo(pos.maxScrollExtent);
+    }
   }
 }
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -75,7 +75,7 @@ class _ConsoleState extends State<Console> {
     await scrollController.animateTo(
       scrollController.position.maxScrollExtent,
       duration: const Duration(milliseconds: 100),
-      curve: Curves.linear,
+      curve: Curves.easeInOut,
     );
 
     // Scroll again if we've received new content in the interim.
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -3,24 +3,85 @@
...
         ],
       ),
     );
   }
+
+  void _scrollToBottom() async {
+    await scrollController.animateTo(
+      scrollController.position.maxScrollExtent,
+      duration: const Duration(milliseconds: 100),
+      curve: Curves.linear,
```
jacob314, line 78: linear isn't a typical choice for an animation curve. document why it is better hear than the more common easeInEaseOut.

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -3,24 +3,85 @@
...
+    );
+
+    // Scroll again if we've received new content in the interim.
+    final pos = scrollController.position;
```
jacob314, line 82: wouldn't this case would also trigger if the user interacted with the scrollable view in which case I don't think we would want to jump to the bottom?
It could be worth it to create a common widget that handles the auto-scroll-to-bottom case.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -120,6 +132,13 @@ class DebuggerController extends DisposableController
     commonScriptPrefix = scriptPrefix;
   }
 
+  /// Append to the stdout / stderr buffer.
+  void appendStdio(String text) {
+    // TODO(devoncarew): Check for greater than some amount of text and truncate
```
jacob314, line 137: why should we truncate?

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -74,6 +76,13 @@ class DebuggerController extends DisposableController
 
   ValueListenable<String> get exceptionPauseMode => _exceptionPauseMode;
 
+  final _stdio = ValueNotifier<String>('');
```
jacob314, line 79: making stdio and stdout be a List<String> would let views displaying this data in the case of a very large input be more efficient as the ListView would be able to display only the rows in view rather than having to first recompute line breaks.

jacob314:
Fyi @helin24 this CL will make it easier for you to test the ansi color coding logic in the Flutter version of the devtools app.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
devoncarew: Just in case the amount of text gets really large? 10k lines? 100k? I'm not sure how much we want to retain, and when we'd start having issues, either with the UI or the VM we're running on.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
devoncarew: Hmm, I'm guessing I could just use lines of `Text` in the `ListView`. This might also make it easier to start using RichTexts in order to differentiate between stdout and stderr.

packages/devtools_app/lib/src/debugger/flutter/console.dart
devoncarew: > wouldn't this case would also trigger if the user interacted with the scrollable view in which case I don't think we would want to jump to the bottom?

Yes, but it would be tough to distinguish between when we added more content and when the user manually scrolled.

What this is guarding against is us stopping auto-scrolling because we ourselves added more content before this animation completes (which I did see when testing).

I can add a todo: here to improve the behavior when the user manually scrolls, or think about ways to address the issue.

> It could be worth it to create a common widget that handles the auto-scroll-to-bottom case.

It would be nice to have for the logging view as well. Perhaps we can look at generalizing when we do look to implement it in one more place?

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
devoncarew: I'll implement this and follow up on the PR.

jacob314: The other thing to consider is whether you want to jump straight to using our TreeTable for the console output here. That would be nice for in the future displaying collapsible rows. The TreeTable also needs to support scroll to bottom functionality.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
jacob314: makes sense

packages/devtools_app/lib/src/debugger/flutter/console.dart
jacob314: Yes would be great to have a generalized robust scroll to bottom view.
Would also be nice to have a light weight abstraction on top of ScrollController that supports some of the cases we need here as well as in the inspector. The default implementation isn't good when you need to support scrolling to a new location when an animation is in progress and when you need to know the end destination of the in progress scroll.

packages/devtools_app/lib/src/debugger/flutter/console.dart
jacob314: see
theme.dart
which has our defaults for durations and curves.
`defaultCurve`
`defaultDuration`
`shortDuration`

jacob314:
lgtm_dog2

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
devoncarew: I'd like to just bite off the ListView work for this PR -

packages/devtools_app/lib/src/debugger/flutter/console.dart
devoncarew: I added a TODO: here to generalize the auto-scroll-to-bottom work.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -126,6 +135,34 @@ class DebuggerController extends DisposableController
...
+
+    // Parse out the new lines and append to the end of the existing lines.
+    var lines = _stdio.value.toList();
+    final newLines = text.split('\n');
+
+    if (lines.isNotEmpty && !lines.last.endsWith('\n')) {
+      lines[lines.length - 1] = '${lines[lines.length - 1]}${newLines.first}';
+      if (newLines.length > 1) {
+        lines.addAll(newLines.sublist(1));
+      }
```
ditman, line 151: Why is this needed? This is causing errors in my console to be fused to the last line of the log:

```
Error: Could not launch tel:
...
    at http://localhost:34077/dart_sdk.js:39165:9Error: Could not launch tel:
...                                             ^^ Here :(
    at http://localhost:34077/dart_sdk.js:39165:9

```

devoncarew: The stdio that the lines represent may not end in a newline. If not, we just append the new stdio chars to the end of the last line.

There could very well be issues w/ the logic here. Basically what we're trying to do is stream chars into this console, and only insert newlines if they're  part of the incoming stdio.

ditman: I couldn't find any cases where this was true, but I guess maybe with colored input it might change? (In my case this is making newer logs harder to find / I thought I had broken something :P)

Issue #1181, opened 2019-10-14
Support ansi color codes


DaveShuckerow:
We need to display ansi colored text in the logging and debug views.

DaveShuckerow:
We could do this by continuing to use ansi_up.js, but to get support on desktop too, I think we should just use a Dart implementation of the color coding.

DaveShuckerow:
For reference: https://github.com/google/ansicolor-dart

PR #1836, opened 2020-04-17
Convert ansi_up.js to dart (https://github.com/flutter/devtools/issues/1181)

helin24:
### Summary
This change converts most of the `ansi_up.js` file to dart so that we will be able to use it for the logging and debugging pages in flutter web as well.

Updates include:
- Moving JS functions into `ansi_up.dart`
- Adding unit test for main ansi function
- Removing references to JS file and removing the JS file

### Testing
To run the unit test:
```
cd packages/devtools_app
flutter test test/ansi_up_test.dart
```

To check that dart:html continues to work:
- Follow instructions in CONTRIBUTING.md to run dart:html web
- Run an app with logging messages including ansi codes, or run a modified version of the fixture file `color_console_output_app` which does this (as in https://github.com/helin24/devtools/commit/69dd41949be5665e33a78bd88442b16052a350d1)
- Verify that color codes appear in logging.


helin24:
I'm not sure how the packaging of `ansi_up` will work during release, since `pubspec.yaml` probably causes the library to be pulled from the public packages. While testing locally I've been specifying the path to the package, but how will the versioning work once this is deployed? Probably I will have to update the version in the `ansi_up/pubspec.yaml` file?

packages/devtools_app/test/ansi_up_test.dart
```diff
@@ -0,0 +1,93 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
```
jacob314, line 1: year is now 2020 :)

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
+  String version, _buffer;
+  bool bold;
+  List<List<AnsiUpColor>> ansiColors;
+  List<AnsiUpColor> palette256;
+  AnsiUpColor fg;
+  AnsiUpColor bg;
+  RegExp _csiRegex;
 
-  external TextPacket get_next_packet();
+  // ignore: sort_constructors_first
```
jacob314, line 17: sort the constructor first rather than suppressing this lint.

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
 /// codes.
-@JS()
-library ansi_up;
-
-import 'package:js/js.dart';
 
-@JS()
 class AnsiUp {
-  external AnsiUp();
+  String version, _buffer;
```
jacob314, line 9: Make _buffer a `final StringBuffer` and put it on its own line.

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
+  List<List<AnsiUpColor>> ansiColors;
+  List<AnsiUpColor> palette256;
+  AnsiUpColor fg;
+  AnsiUpColor bg;
+  RegExp _csiRegex;
 
-  external TextPacket get_next_packet();
+  // ignore: sort_constructors_first
+  AnsiUp() {
+    version = '4.0.3';
```
jacob314, line 19: nit: remove version. We don't care as the Dart `pub` package system lets us track versions.

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
+    _setupPalettes();
+    bold = false;
+    fg = bg = null;
```
jacob314, line 22: no need to initialize variables to null in Dart.

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
+        AnsiUpColor(rgb: [0, 255, 0], class_name: 'ansi-bright-green'),
+        AnsiUpColor(rgb: [255, 255, 85], class_name: 'ansi-bright-yellow'),
+        AnsiUpColor(rgb: [85, 85, 255], class_name: 'ansi-bright-blue'),
+        AnsiUpColor(rgb: [255, 85, 255], class_name: 'ansi-bright-magenta'),
+        AnsiUpColor(rgb: [85, 255, 255], class_name: 'ansi-bright-cyan'),
+        AnsiUpColor(rgb: [255, 255, 255], class_name: 'ansi-bright-white'),
+      ]
+    ];
+    palette256 = [];
+    for (var palette in ansiColors) {
```
jacob314, line 50: Will do the same thing as this for loop.
`palette256.addAll(palette);`

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
+    for (var i = 0; i < 24; ++i, greyLevel += 10) {
+      palette256.add(AnsiUpColor(
+          rgb: [greyLevel, greyLevel, greyLevel], class_name: 'truecolor'));
+    }
+  }
 
-  external void process_ansi(TextPacket textPacket);
-  external TextWithAttr with_state(TextPacket packet);
+  TextPacket _getNextPacket() {
+    final TextPacket pkt = TextPacket(kind: PacketKind.EOS, text: '', url: '');
```
jacob314, line 70: no need to specify the type on the lhs in these two cases as it is inferred.

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
+        return pkt;
+      }
+      if (nextChar == '[') {
+        _csiRegex ??= rgx(
+            '\n                        ^                           # beginning of line\n                                                    #\n                                                    # First attempt\n                        (?:                         # legal sequence\n                          \\x1b\\[                      # CSI\n                          ([\\x3c-\\x3f]?)              # private-mode char\n                          ([\\d;]*)                    # any digits or semicolons\n                          ([\\x20-\\x2f]?               # an intermediate modifier\n                          [\\x40-\\x7e])                # the command\n                        )\n                        |                           # alternate (second attempt)\n                        (?:                         # illegal sequence\n                          \\x1b\\[                      # CSI\n                          [\\x20-\\x7e]*                # anything legal\n                          ([\\x00-\\x1f:])              # anything illegal\n                        )\n                    ');
```
jacob314, line 100: you can take advantage of multiline dart strings to make this more readable.
```dart
'foo '
'bar '
'baz'
```
is equivalent to 
```dart
''foo bar baz'
```

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
+        final RegExpMatch match = _csiRegex.firstMatch(_buffer);
```
jacob314, line 101: as before omit type on LHS.

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
+        }
+        if (match.groupCount > 4) {
+          pkt.kind = PacketKind.ESC;
+          pkt.text = _buffer.substring(0, 1);
+          _buffer = _buffer.substring(1);
+          return pkt;
+        }
+        final String match1 = match.groupCount > 1 ? match.group(1) : null;
+        final String match3 = match.groupCount > 3 ? match.group(3) : null;
+        if ((match1 != '') || (match3 != 'm'))
```
jacob314, line 114: nit: nested parens are not needed.

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
+          pkt.kind = PacketKind.Unknown;
```
jacob314, line 115: nit: put `{` around if,else conditions

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
+      //  pkt.text = match[2];
+      //  var rpos = match[0].length;
+      //  this._buffer = this._buffer.slice(rpos);
+      //  return pkt;
+      //}
+    }
+    return pkt;
+  }
+
+  void _appendBuffer(String text) {
```
jacob314, line 187: remove this buffer and use
_buffer.write(text)
once you switch to a StringBuffer

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
+    _buffer = _buffer + text;
+  }
+
+  void _processAnsi(TextPacket textPacket) {
+    final List<String> sgrCmds = textPacket.text.split(';');
```
jacob314, line 192: as in other cases, remove type on LHS here.

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
+          }
+          if (modeCmd == '2' && sgrCmds.length > 2) {
+            final int r = int.tryParse(sgrCmds.removeAt(0), radix: 10);
+            final int g = int.tryParse(sgrCmds.removeAt(0), radix: 10);
+            final int b = int.tryParse(sgrCmds.removeAt(0), radix: 10);
+            if ((r >= 0 && r <= 255) &&
+                (g >= 0 && g <= 255) &&
+                (b >= 0 && b <= 255)) {
+              var c = AnsiUpColor(rgb: [r, g, b], class_name: 'truecolor');
+              if (isForeground)
```
jacob314, line 236: as elsewhere, make sure there are curly braces around if/else clauses.

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
 
-@JS()
 class TextWithAttr {
-  external AU_Color get fg;
-  external AU_Color get bg;
-  external bool get bold;
-  external String get text;
+  TextWithAttr({this.fg, this.bg, this.bold, this.text});
+
+  AnsiUpColor fg;
```
jacob314, line 255: make these fields all final.

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
-@JS()
-@anonymous
-class AU_Color {
-  external List<int> get rgb;
-  external String get class_name;
+class AnsiUpColor {
+  AnsiUpColor({this.rgb, this.class_name});
+
+  List<int> rgb;
+  String class_name;
```
jacob314, line 265: rename class_name to className.

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
 }
 
-class PacketKind {
-  static final int EOS = 0;
-  static final int Text = 1;
+mixin PacketKind {
```
jacob314, line 268: why is this a mixin?

jacob314: now that this code is all Dart you could as a quick followup turn this into an actual Dart enum
`enum PacketKind { EOS, Text, ...}`

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
-  external String get text;
-  external String get url;
+  int kind;
+  String text;
+  String url;
 }
 
 String _colorToCss(List/*<int>*/ rgb) => 'rgb(${rgb.join(',')})';
 
+RegExp rgx(String regexText) {
```
jacob314, line 299: pick a better name and document what this is doing.

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,306 @@
...
+  final RegExp wsrgx = RegExp(r"^\s+|\s+\n|\s*#[\s\S]*?\n|\n", multiLine: true);
+  final String txt2 = regexText.replaceAll(wsrgx, '');
```
jacob314, line 301: the variable names in this method don't mean anything. omit the variable names or pick more meaningful names. Dart variable names should not contain abbreviations or acronyms.

third_party/packages/ansi_up/lib/ansi_up.dart
helin24: Why `StringBuffer` instead of `String` here? It's a little strange that this is called a buffer in the first place because there's only one spot where a string is added.

third_party/packages/ansi_up/lib/ansi_up.dart
helin24: You mean for the `TextPacket` because it's being initialized on the right side, and `int` on the next line because `_buffer.length` is clearly an int? Does this apply to `_buffer.indexOf` below too?

third_party/packages/ansi_up/lib/ansi_up.dart
jacob314: sgtm. removing the append_buffer method completely sounds good.

jacob314:
> I'm not sure how the packaging of `ansi_up` will work during release, since `pubspec.yaml` probably causes the library to be pulled from the public packages. While testing locally I've been specifying the path to the package, but how will the versioning work once this is deployed? Probably I will have to update the version in the `ansi_up/pubspec.yaml` file?

You should update the version in the pubspec.yaml. fine.
the easiest way to do it would be to when it comes time to commit only commit the part of this CL that is under ansi_up. then publish the package `pub publish` then commit the other part of the CL as well as updating the version of ansi_up referenced in that cl.

That work flow is ok for this case because I expect after this CL we won't be tweaking ansi_up again aside for maybe bug fixes. For packages that we modify frequently, we have a separate workflow with dependency_overrides that makes iterating easier.

third_party/packages/ansi_up/lib/ansi_up.dart
helin24: It was originally a class so the automatic linting suggested a mixin instead 😅 Enum makes a lot more sense though!

third_party/packages/ansi_up/lib/ansi_up.dart
helin24: Did I do this part correctly? If I'm setting this `_text` field should it not be private with the underscore?

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,316 @@
...
-class TextWithAttr {
-  external AU_Color get fg;
-  external AU_Color get bg;
-  external bool get bold;
-  external String get text;
-}
+  void _processAnsi(TextPacket textPacket) {
+    final sgrCmds = textPacket.text.split(';');
+    while (sgrCmds.isNotEmpty) {
+      final sgrCmdStr = sgrCmds.removeAt(0);
```
jacob314, line 211: Add a TODO to change 
sgrCmds.removeAt(0) to tracking
an index in sgrCmds. That should be more efficient in Dart.

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,316 @@
...
-@JS()
-@anonymous
 class TextPacket {
-  /// enum like constant from PacketKind describing the packet.
-  external int get kind;
-  external String get text;
-  external String get url;
+  TextPacket({this.kind, this.text, this.url});
+
+  PacketKind kind;
```
jacob314, line 300: make all class fields that can be final be final for clarity.

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -2,76 +2,316 @@
...
+}
 
-  /// A valid CSI but not an SGR code.
-  static final int Unknown = 4;
+class AnsiUpColor {
+  AnsiUpColor({this.rgb, this.className});
 
-  /// Select Graphic Rendition.
-  static final int SGR = 5;
+  List<int> rgb;
```
jacob314, line 283: make all class fields that can be final be final for clarity.

PR #1838, opened 2020-04-19
improve the look of the badge control

devoncarew:
- improve the look of the badge control (it's slightly smaller and uses our theme color)

Screen Shot 2020-04-18 at 5 44 57 PM


packages/devtools_app/lib/src/flutter/common_widgets.dart
```diff
@@ -319,19 +319,27 @@ class BulletSpacer extends StatelessWidget {
 /// A small element containing some accessory information, often a numeric
 /// value.
 class Badge extends StatelessWidget {
-  const Badge({@required this.child});
+  const Badge({@required this.text});
```
jacob314, line 322: nit: make text be a required parameter instead of a required named parameter.
That would be consistent with widgets such as Text and save a bit of typing wherever this is used.

PR #1842, opened 2020-04-20
log all service protocol traffic

devoncarew:
- add a `debugLogServiceProtocolEvents` const to the debugger controller; when this is turned on, we log all service protocol traffic to the Logging page. This can be used to audit our service protocol traffic, and be better aware when we're making expensive calls

@jacob314, this shows up some issues wrt our controller lifetimes. It looks like we're creating controllers twice at app startup, and then once each time we hot reload, likely w/o disposing of the old controllers. I'm not sure why we re-create controllers on hot reload. I wasn't successful in untangling exactly what's going on. It has the effect of us sending a lot more traffic over the service protocol (requesting objects, the list of scripts, ...) than we need to.

@grouma - this should be useful in auditing our protocol traffic, reducing expensive calls, and delaying some calls until they're necessary.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
@@ -3,12 +3,20 @@
 // found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:convert';
 
 import 'package:flutter/foundation.dart';
 import 'package:vm_service/vm_service.dart';
 
 import '../../auto_dispose.dart';
+import '../../core/message_bus.dart';
 import '../../globals.dart';
+
+// Note: don't check this in enabled.
+/// Used to debug service protocol traffic. All requests to to the VM service
+/// connection are logged to the Logging page, as well as all responses and
+/// events from the service protocol device.
+const debugLogServiceProtocolEvents = false;
 
 /// Responsible for managing the debug state of the app.
 class DebuggerController extends DisposableController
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -3,13 +3,21 @@
...
+import 'dart:convert';
 
 import 'package:flutter/foundation.dart';
 import 'package:vm_service/vm_service.dart';
 
 import '../../auto_dispose.dart';
+import '../../core/message_bus.dart';
 import '../../globals.dart';
 
+// Note: don't check this in enabled.
```
jacob314, line 15: This doesn't seem debugger specific. Maybe we should add this to 
VmServiceWrapper?

packages/devtools_app/lib/src/flutter/controllers.dart
```diff
@@ -146,6 +146,13 @@ class _ControllersState extends State<Controllers> {
   @override
   void didUpdateWidget(Controllers oldWidget) {
     super.didUpdateWidget(oldWidget);
+
+    // TODO(#1732): We create controllers twice - here, and in initState().
```
jacob314, line 150: Can we short circuit the call to _initializeProviderData in this case when we notice that oldWidget and widget have the same overrideProviders? That along with calling dispose should fix all the issues.

devoncarew: I experimented with a few different things here - I'm a little less familiar with the state management going on here than I'd like to be. I'd like to leave this as a todo to revisit later, or have someone more familiar with this area take a look.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
devoncarew: Will do -

kenzieschmoll:
Let's hold off on landing this until https://github.com/flutter/devtools/pull/1808 lands. We are switching to using package:provider to provide our controllers, so the controller issue you're seeing here may be obsolete.

devoncarew:
Closing - will open an alternative PR.

PR #1843, opened 2020-04-20
logging view UI work and improved filtering

devoncarew:
- tweak the UI of the logging view to better match the look of other pages
- remove an opacity animation from the logging pages detail area - overall I don't think this animation helped the page
- add more sophisticated filtering to the page - support negative filter terms (`foo` matches all items w/ 'foo`; `-bar` matches all items that do not contain 'bar')

Screen Shot 2020-04-19 at 6 25 47 PM

### DIFF
--- a/packages/devtools_app/lib/src/core/filtering.dart
+++ b/packages/devtools_app/lib/src/core/filtering.dart
@@ -1,0 +1,60 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+typedef Matcher = bool Function(String text);
+
+/// Create a query a text based filter.
+/// 
+/// The search terms can be positive terms (the query items must contain the
+/// term) or negative terms (the query items must not contain the term).
+///
+/// The filter text `'foo'` would match all items that contain the word `foo`.
+/// `'-bar'` would exclude all items that contain `bar`.
+class Filter {
+  Filter._(this.items);
+
+  static Filter compile(String text) {
+    if (text == null) {
+      return Filter._(const []);
+    }
+
+    text = text.toLowerCase().trim().replaceAll(RegExp(r'\s+'), ' ');
+    final terms = text.split(' ');
+
+    return Filter._(terms.map((term) {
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/core/filtering.dart
```diff
@@ -0,0 +1,60 @@
...
+        return false;
+      }
+    }
+
+    return true;
+  }
+}
+
+class FilterItem {
+  FilterItem(this.text, [this.negative = false]);
```
jacob314, line 54: nit: would negative be clearer as a named parameter?

packages/devtools_app/test/core/filtering_test.dart
```diff
@@ -0,0 +1,67 @@
...
+      expect(applyFilter(filter, items), ['foo']);
+      expect(
+          applyFilter(filter, ['foo', 'foofoo', 'foobar']), ['foo', 'foofoo']);
+    });
+  });
+}
+
+List<String> applyFilter(Filter filter, List<String> items) {
+  return items.where((element) {
+    final m = filter.matches((String text) => element.contains(text));
```
jacob314, line 64: nit. remove local variable `m` that doesn't add value.

packages/devtools_app/lib/src/logging/flutter/logging_screen.dart
```diff
@@ -236,35 +223,16 @@ class _LogDetailsState extends State<LogDetails>
...
-              ),
-              Opacity(
-                opacity: 1 - crossFade.value,
-                child: _buildContent(context, _oldLog),
-              ),
-            ],
-          ),
-        );
-      },
+    return Container(
```
jacob314, line 233: The crossFade was intentionally removed, right?

packages/devtools_app/lib/src/core/filtering.dart
devoncarew: yup - changed

packages/devtools_app/test/core/filtering_test.dart
devoncarew: removed (I'd added this while debugging the tests)

PR #1844, opened 2020-04-20
Migrating to new Slider

JoseAlba:
Migrating Slider to latest specs from Material which includes improved accessibility.


devoncarew:
(re-starting the bot)

devoncarew:
The latest material spec: https://material.io/components/sliders

PR #1847, opened 2020-04-21
update the Badge widget to be slightly more round

devoncarew:
- update the Badge widget to be slightly more round


devoncarew:
> LGTM - unrelated to this PR, but have you checked the badge styles in both light and dark theme? I noticed in dark theme it looked like the background color was the same as the section header, but you may have fixed this since I last looked.

No, this still has that issue. It would be useful to have a separate background color for sure. The dark badge is still readable, and doesn't stand out as a UI glitch, even if it doesn't match the styling of the light theme.

Do we know the different theme colors to use it different places? I'm never quite sure where to use the accent vs other types of color (the app bar color, the background, normal text, selected text, selected table rows, ...). If we do have a system we should document and audit the app; if not, perhaps @raison00 could help here?

PR #1848, opened 2020-04-21
Use index instead of removeAt to process Ansi codes

helin24:
I figured it makes sense to make this small change before publishing this package to pub.dev. I'm not sure if this kind of change independently would warrant a minor version change.


packages/devtools_app/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -5,8 +5,10 @@
 import FlutterMacOS
 import Foundation
 
+import path_provider_macos
```
jacob314, line 8: revert this file as this seems unrelated. We should add GeneratedPluginRegistrant.swift to the `.gitignore` file and remove it from github.

PR #1852, opened 2020-04-21
improve breakpoint resolution code

devoncarew:
improve breakpoint resolution code:
- improve our wrapper around breakpoints that are resolved and ones that aren't (clients of the API can mostly use the typed wrappers now, instead of the `dynamic` location field)
- fix issues with jumping to locations of breakpoints for file that haven't yet been executed
- fix issues for setting breakpoints in scripts that are part files

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
@@ -46,6 +46,9 @@ class DebuggerController extends DisposableController
 
   ValueListenable<Script> get currentScript => _currentScript;
 
+  // A cached map of uris to ScriptRefs.
+  final Map<String, ScriptRef> _uriToScriptMap = {};
+
   final _currentStack = ValueNotifier<Stack>(null);
 
   ValueListenable<Stack> get currentStack => _currentStack;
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -77,6 +77,15 @@ class DebuggerController extends DisposableController
 
   ValueListenable<String> get exceptionPauseMode => _exceptionPauseMode;
 
+  final ValueNotifier<bool> _librariesVisible = ValueNotifier(false);
+
+  ValueNotifier<bool> get librariesVisible => _librariesVisible;
```
jacob314, line 82: Nit: make the LHS a ValueListenable to make writes more difficult.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -310,6 +319,11 @@ class DebuggerController extends DisposableController
     return _scriptCache[scriptRef.id];
   }
 
+  ScriptRef getScriptRefFromUri(String uri) {
```
jacob314, line 322: dart style: avoid prefixing method names with the word `get`.

Consider keeping a cached map from uri to ScriptRef to avoid the O(num scripts) search each time the scriptRefFromUri is needed. Probably not a needed optimization but scanning through the full list of 30,000 scripts might be noticeable for large applications.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
devoncarew: oops - removed; this was for a separate PR

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
devoncarew: Changed to `scriptRefForUri`.

I'm not sure it'll be a performance issue - this is currently only called (once) on an explicit user action. I'll cache the look up in map however - we might refactor the code in the future and end up calling this method more than expected.

PR #1856, opened 2020-04-21
Add arrow key navigation to the inspector tree.

### LINKED ISSUE #227: Inspector tree should be navigable with keyboard
Currently the cursor keys seem to scroll the inspector panes, but I think it'd be better if they behaved like other trees (up/down will go to the node rendered directly above/below, left/right should expand/collapse).

Also, I don't think you can currently tab to the trees.

gmoothart:
It seems that we only had to enable functionality already implemented in
InspectorTreeController.

Fixes #227

### DIFF
--- a/packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
+++ b/packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
@@ -104,14 +104,27 @@ class _InspectorScreenBodyState extends State<InspectorScreenBody>
       decoration: BoxDecoration(
         border: Border.all(color: Theme.of(context).focusColor),
       ),
-      child: InspectorTree(
-        controller: summaryTreeController,
-        isSummaryTree: true,
+      child: GestureDetector(
+        child: InspectorTree(
+          controller: summaryTreeController,
+          isSummaryTree: true,
+        ),
+        onTapUp: (TapUpDetails tapUpDetails) {
+          summaryTreeController.setFocus();
+        },
       ),
     );
-    final detailsTree = InspectorTree(
-      controller: detailsTreeController,
+    final detailsTree = GestureDetector(
+      child: InspectorTree(
+        controller: detailsTreeController,
+      ),
+      onTapUp: (TapUpDetails tapUpDetails) {
+        detailsTreeController.setFocus();
+      },
     );
+
+    // Set focus on the summary tree by default
+    summaryTreeController.setFocus();
 
     final splitAxis = Split.axisFor(context, 0.85);
     return Column(
--- a/packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
+++ b/packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
@@ -104,27 +104,14 @@ class _InspectorScreenBodyState extends State<InspectorScreenBody>
       decoration: BoxDecoration(
         border: Border.all(color: Theme.of(context).focusColor),
       ),
-      child: GestureDetector(
-        child: InspectorTree(
-          controller: summaryTreeController,
-          isSummaryTree: true,
-        ),
-        onTapUp: (TapUpDetails tapUpDetails) {
-          summaryTreeController.setFocus();
-        },
-      ),
-    );
-    final detailsTree = GestureDetector(
       child: InspectorTree(
-        controller: detailsTreeController,
+        controller: summaryTreeController,
+        isSummaryTree: true,
       ),
-      onTapUp: (TapUpDetails tapUpDetails) {
-        detailsTreeController.setFocus();
-      },
     );
-
-    // Set focus on the summary tree by default
-    summaryTreeController.setFocus();
+    final detailsTree = InspectorTree(
+      controller: detailsTreeController,
+    );
 
     final splitAxis = Split.axisFor(context, 0.85);
     return Column(
--- a/packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
+++ b/packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
@@ -139,12 +139,18 @@ class InspectorTreeControllerFlutter extends Object
     }
     return _maxIndent;
   }
+
+  void setFocus() {
+    client?.setFocus();
+  }
 }
 
 abstract class InspectorControllerClient {
   void onChanged();
 
   void scrollToRect(Rect rect);
+
+  void setFocus();
 }
 
 class InspectorTree extends StatefulWidget {
@@ -201,7 +207,6 @@ class _InspectorTreeState extends State<InspectorTree>
 
       _bindToController();
     }
-    FocusScope.of(context).requestFocus(focusNode);
     super.didUpdateWidget(oldWidget);
   }
 
@@ -212,6 +217,11 @@ class _InspectorTreeState extends State<InspectorTree>
     _scrollControllerX.dispose();
     _scrollControllerY.dispose();
     constraintDisplayController?.dispose();
+  }
+
+  @override
+  void setFocus() {
+    FocusScope.of(context).requestFocus(focusNode);
   }
 
   void _onScrollYChange() {
@@ -335,7 +345,6 @@ class _InspectorTreeState extends State<InspectorTree>
   void _handleKeyEvent(RawKeyEvent event) {
     if (event is! RawKeyDownEvent) return;
 
-    // Exit early if we aren't handling the key
     if (event.logicalKey == LogicalKeyboardKey.arrowDown) {
       controller.navigateDown();
     } else if (event.logicalKey == LogicalKeyboardKey.arrowUp) {
--- a/packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
+++ b/packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
@@ -7,6 +7,7 @@ import 'dart:math';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:pedantic/pedantic.dart';
+import 'package:flutter/services.dart';
 
 import '../../flutter/auto_dispose_mixin.dart';
 import '../../flutter/collapsible_mixin.dart';
@@ -177,6 +178,7 @@ class _InspectorTreeState extends State<InspectorTree>
   Rect currentAnimateTarget;
 
   AnimationController constraintDisplayController;
+  FocusNode focusNode = FocusNode();
 
   @override
   void initState() {
@@ -199,6 +201,7 @@ class _InspectorTreeState extends State<InspectorTree>
 
       _bindToController();
     }
+    FocusScope.of(context).requestFocus(focusNode);
     super.didUpdateWidget(oldWidget);
   }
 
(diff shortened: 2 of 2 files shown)


packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
```diff
@@ -329,6 +332,21 @@ class _InspectorTreeState extends State<InspectorTree>
     return maxOffset - viewportDimension;
   }
 
+  void _handleKeyEvent(RawKeyEvent event) {
+    if (event is! RawKeyDownEvent) return;
+
+    // Exit early if we aren't handling the key
```
kenzieschmoll, line 338: This comment doesn't seem to match the code below it. Can we just remove it?

gmoothart:
Note: arrow keys seem to only control the Details Tree, even when I click to place focus in the left tree. I'm not sure how to control that one.

inspector-keyboard-nav

kenzieschmoll:
What about requiring a click in the pane area to set focus instead of requesting focus in did update widget on both sections? An ancestor widget may need to maintain data on which tree has focus.

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
gmoothart: done.

packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
```diff
@@ -104,15 +104,28 @@ class _InspectorScreenBodyState extends State<InspectorScreenBody>
...
       ),
-      child: InspectorTree(
-        controller: summaryTreeController,
-        isSummaryTree: true,
+      child: GestureDetector(
+        child: InspectorTree(
+          controller: summaryTreeController,
+          isSummaryTree: true,
+        ),
+        onTapUp: (TapUpDetails tapUpDetails) {
```
gmoothart, line 112: This works as long as I click in the blank space of the tree. If I click on an actual node it doesn't fire.

packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
```diff
@@ -104,15 +104,28 @@ class _InspectorScreenBodyState extends State<InspectorScreenBody>
...
+      child: InspectorTree(
+        controller: detailsTreeController,
+      ),
+      onTapUp: (TapUpDetails tapUpDetails) {
+        detailsTreeController.setFocus();
+      },
     );
 
+    // Set focus on the summary tree by default
+    summaryTreeController.setFocus();
```
gmoothart, line 127: Is there a better place to set default focus than in the build method?

packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
kenzieschmoll: try modifying the `behavior` property of the GestureDetector. Maybe `HitTestBehavior.opaque`?

packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
kenzieschmoll: can we have the inspector screen track which tree was last selected? and then in didUpdateWidget, set the focus on the last selected tree?

kenzieschmoll: @jacob314 your thoughts?

packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
```diff
@@ -104,15 +104,28 @@ class _InspectorScreenBodyState extends State<InspectorScreenBody>
...
-      child: InspectorTree(
-        controller: summaryTreeController,
-        isSummaryTree: true,
+      child: GestureDetector(
+        child: InspectorTree(
+          controller: summaryTreeController,
+          isSummaryTree: true,
+        ),
+        onTapUp: (TapUpDetails tapUpDetails) {
+          summaryTreeController.setFocus();
```
jacob314, line 113: we should try hard to integrate with Flutter concepts of focus rather than adding our own concept of setting focus. Focus is something that is really hard to do robustly. Notice you will already get some flutter focus changes when you tab between controls in the flutter version of the inspector. the focus changes aren't good but that illustrates the baked in functionality.

packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
jacob314: we should try to integrate with the core flutter concepts of tracking the current focus. Having our controllers have a separate definition of focus will lead to confusing ui. You can add a hook in the inspector controller to query the associated view to determine it has focus if you need that info in the controller.

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
```diff
@@ -211,6 +219,11 @@ class _InspectorTreeState extends State<InspectorTree>
     constraintDisplayController?.dispose();
   }
 
+  @override
+  void setFocus() {
+    FocusScope.of(context).requestFocus(focusNode);
```
jacob314, line 224: rename this method to requestFocus to match the flutter naming conventions.

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
```diff
@@ -177,6 +184,7 @@ class _InspectorTreeState extends State<InspectorTree>
   Rect currentAnimateTarget;
 
   AnimationController constraintDisplayController;
+  FocusNode focusNode = FocusNode();
```
jacob314, line 187: make final.

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
```diff
@@ -211,6 +219,11 @@ class _InspectorTreeState extends State<InspectorTree>
     constraintDisplayController?.dispose();
```
jacob314, line 219: init the focusNode in initState and dispose in the dispose in the dispose method to handle the lifecycle better.

jacob314: you might also consider adding a https://api.flutter.dev/flutter/widgets/FocusScope-class.html widget to the tree but I'm minimally familiar with managing focus in Flutter.

packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
jacob314: I would reach out to the flutter users chat to get some feedback on this. I hope we can  just make the summary tree be the default focus target.

jacob314: The documentation for https://api.flutter.dev/flutter/widgets/FocusScope-class.html also has some good breadcrumbs describing how Flutter intends to support tracking focus.

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
```diff
@@ -138,12 +139,18 @@ class InspectorTreeControllerFlutter extends Object
     }
     return _maxIndent;
   }
+
+  void setFocus() {
```
jacob314, line 143: rename to requestFocus to match Flutter conventions and avoid methods that start with the word `set`

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
gmoothart: FocusScope sounds like what we want here.

packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
gmoothart: That didn't seem to have any effect. I added a line in the node's onTap() to explicitly focus the tree, along with a TODO to looks for a better way.

packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
gmoothart: Using a Focus node simplified the code a lot.

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
gmoothart: Done

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
gmoothart: It can't be final now because I'm initializing it in initState()

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
jacob314: We are only building with tip of trunk flutter so you can use FocusTraversalGroup now.

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
```diff
@@ -353,20 +380,24 @@ class _InspectorTreeState extends State<InspectorTree>
...
-                (context, index) {
-                  final InspectorTreeRow row = controller.root?.getRow(index);
-                  return _InspectorTreeRowWidget(
-                    key: PageStorageKey(row?.node),
-                    inspectorTreeState: this,
-                    row: row,
-                  );
-                },
-                childCount: controller.numRows,
+            child: RawKeyboardListener(
```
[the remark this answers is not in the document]
gmoothart: Done. This made the design a lot simpler, thanks!

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
```diff
@@ -353,20 +384,31 @@ class _InspectorTreeState extends State<InspectorTree>
...
-                  final InspectorTreeRow row = controller.root?.getRow(index);
-                  return _InspectorTreeRowWidget(
-                    key: PageStorageKey(row?.node),
-                    inspectorTreeState: this,
-                    row: row,
-                  );
-                },
-                childCount: controller.numRows,
+            child: GestureDetector(
+              onTap: () {
```
gspencergoog, line 388: This doesn't need the closure.

```suggestion
              onTap: _focusNode.requestFocus,
```

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
gmoothart: I tried enclosingScope and it didn't have the right behavior - I'm not sure what got scope, but it wasn't the tree. So I'm going to punt on this for now.

gspencergoog: Yeah, I think I was misreading your code.  What you want is `focusNode.requestFocus()`.

devoncarew:
@gmoothart - some of the issues w/ the bots should be addressed if you merge the latest from master into this PR

PR #1858, opened 2020-04-22
Manage selected breakpoint in debugger_controller.

kenzieschmoll:
- Use value listenables to manage UI state in breakpoints.dart
- get controller from Provider.of
- avoid rebuilding the entire Debugger screen when the selected breakpoint changes

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
@@ -75,6 +75,11 @@ class DebuggerController extends DisposableController
   ValueListenable<List<BreakpointAndSourcePosition>>
       get breakpointsWithLocation => _breakpointsWithLocation;
 
+  final _selectedBreakpoint = ValueNotifier<BreakpointAndSourcePosition>(null);
+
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -75,6 +75,11 @@ class DebuggerController extends DisposableController
   ValueListenable<List<BreakpointAndSourcePosition>>
       get breakpointsWithLocation => _breakpointsWithLocation;
 
+  final _selectedBreakpoint = ValueNotifier<BreakpointAndSourcePosition>(null);
```
devoncarew, line 78: Should we eventually set up a 2nd debugger controller, that's more focused on the UI state? There are a few items in this controller that are UI state related (what's focused, ...) - most of the other items track more the state of the VM.

kenzieschmoll: I think it is fine to have a single controller. This is what we do for other pages in DevTools as well. For example, the TimelineController maintains both the state of the data as well as the "selected" state of the UI (frame, timeline event, stack frame). In many cases, the selected UI dictate how the controller processes data, so I think it makes sense to have all "stateful" data maintained by the controller.

PR #1860, opened 2020-04-22
Display ansi colors in debugger and logging screens

helin24:
Debugger screen console:
Screen Shot 2020-04-22 at 2 13 49 PM

Logging screen:
Screen Shot 2020-04-22 at 2 02 37 PM

https://github.com/flutter/devtools/issues/1181

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -70,7 +70,9 @@ class _ConsoleState extends State<Console> {
                     controller: scrollController,
                     itemBuilder: (context, index) {
                       return RichText(
-                        text: TextSpan(children: maybeConvertToAnsiText(lines[index], textStyle)),
+                        text: TextSpan(
+                            children: maybeConvertToAnsiText(
+                                lines[index], textStyle)),
                         maxLines: 1,
                       );
                     },
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -71,7 +71,7 @@ class _ConsoleState extends State<Console> {
                     itemBuilder: (context, index) {
                       return RichText(
                         text: TextSpan(
-                          children: maybeConvertToAnsiText(
+                          children: processAnsiTerminalCodes(
                             lines[index],
                             textStyle,
                           ),
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -71,8 +71,11 @@ class _ConsoleState extends State<Console> {
                     itemBuilder: (context, index) {
                       return RichText(
                         text: TextSpan(
-                            children: maybeConvertToAnsiText(
-                                lines[index], textStyle)),
+                          children: maybeConvertToAnsiText(
+                            lines[index],
+                            textStyle,
+                          ),
+                        ),
                         maxLines: 1,
                       );
                     },
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -7,7 +7,7 @@ import 'package:flutter/scheduler.dart';
 
 import '../../flutter/common_widgets.dart';
 import '../../flutter/theme.dart';
-import '../../utils.dart';
+import '../../flutter/utils.dart';
 import 'codeview.dart';
 import 'common.dart';
 import 'debugger_controller.dart';
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -7,6 +7,7 @@ import 'package:flutter/scheduler.dart';
 
 import '../../flutter/common_widgets.dart';
 import '../../flutter/theme.dart';
+import '../../utils.dart';
 import 'codeview.dart';
 import 'common.dart';
 import 'debugger_controller.dart';
@@ -68,10 +69,9 @@ class _ConsoleState extends State<Console> {
                     itemExtent: CodeView.rowHeight,
                     controller: scrollController,
                     itemBuilder: (context, index) {
-                      return Text(
-                        lines[index],
+                      return RichText(
+                        text: TextSpan(children: maybeConvertToAnsiText(lines[index], textStyle)),
                         maxLines: 1,
-                        style: textStyle,
                       );
                     },
                   );
--- a/packages/devtools_app/lib/src/flutter/utils.dart
+++ b/packages/devtools_app/lib/src/flutter/utils.dart
@@ -16,33 +16,27 @@ Future<void> launchUrl(String url, BuildContext context) async {
   }
 }
 
-List<TextSpan> maybeConvertToAnsiText(String input, TextStyle defaultStyle) {
+List<TextSpan> processAnsiTerminalCodes(String input, TextStyle defaultStyle) {
   return decodeAnsiColorEscapeCodes(input, AnsiUp())
       .map((entry) => TextSpan(
             text: entry.text,
             style: entry.style.isEmpty
                 ? defaultStyle
                 : TextStyle(
-                    color: entry.fgColor != null && entry.fgColor.length > 2
-                        ? Color.fromRGBO(
-                            entry.fgColor[0],
-                            entry.fgColor[1],
-                            entry.fgColor[2],
-                            1,
-                          )
+                    color: entry.fgColor != null
+                        ? colorFromAnsi(entry.fgColor)
                         : null,
-                    backgroundColor:
-                        entry.bgColor != null && entry.bgColor.length > 2
-                            ? Color.fromRGBO(
-                                entry.bgColor[0],
-                                entry.bgColor[1],
-                                entry.bgColor[2],
-                                1,
-                              )
-                            : null,
+                    backgroundColor: entry.bgColor != null
+                        ? colorFromAnsi(entry.bgColor)
+                        : null,
                     fontWeight:
                         entry.bold ? FontWeight.bold : FontWeight.normal,
                   ),
           ))
       .toList();
+}
+
+Color colorFromAnsi(List<int> ansiInput) {
+  assert(ansiInput.length == 3, 'Ansi color list should contain 3 elements');
+  return Color.fromRGBO(ansiInput[0], ansiInput[1], ansiInput[2], 1);
 }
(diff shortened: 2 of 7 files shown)


packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -70,7 +70,9 @@ class _ConsoleState extends State<Console> {
                     controller: scrollController,
                     itemBuilder: (context, index) {
                       return RichText(
-                        text: TextSpan(children: maybeConvertToAnsiText(lines[index], textStyle)),
+                        text: TextSpan(
+                            children: maybeConvertToAnsiText(
+                                lines[index], textStyle)),
```
jacob314, line 75: general nit: add trailing commas to make the dartfmt use flutter style pretty much any time you get more than one comma on the same line.

helin24: In this case should it be a comma after `textStyle` and a comma after the first parens? And then flutter style puts every item ending in a comma in its own line?

packages/devtools_app/lib/src/flutter/utils.dart
```diff
@@ -14,3 +15,26 @@ Future<void> launchUrl(String url, BuildContext context) async {
     Notifications.of(context).push('Unable to open $url.');
   }
 }
+
+List<TextSpan> maybeConvertToAnsiText(String input, TextStyle defaultStyle) {
```
jacob314, line 19: nit: I wouldn't use the word maybe in this method name.
maybe name it
`processAnsiTerminalCodes` instead
The reason I recommend omitting the word maybe as a List<TextSpan> is generated regardless even if the text doesn't happen to have ansi terminal codes.

packages/devtools_app/lib/src/flutter/utils.dart
```diff
@@ -14,3 +15,26 @@ Future<void> launchUrl(String url, BuildContext context) async {
...
+  return decodeAnsiColorEscapeCodes(input, AnsiUp())
+      .map((entry) => TextSpan(
+            text: entry.text,
+            style: entry.style.isEmpty
+                ? defaultStyle
+                : TextStyle(
+                    color: entry.fgColor != null && entry.fgColor.length > 2
+                        ? Color.fromRGBO(entry.fgColor[0], entry.fgColor[1],
```
jacob314, line 27: add helper method to convert an array[int] to a Color object and use it for both cases.
Make it an assert the entry.fooColor.length == 3. It is a bug in AnsiUp if that isn't the case.

packages/devtools_app/lib/src/logging/flutter/logging_screen.dart
```diff
@@ -357,6 +362,33 @@ class _MessageColumn extends LogMessageColumn
...
         ],
       );
+    } else if (data.kind == 'stdout') {
+      // TODO(helin24): Move this to logging controller when dart:html is removed.
+      const Color color = Color.fromARGB(0xff, 0x00, 0x91, 0xea);
+
+      final RichText richText = RichText(
+        text: TextSpan(
+          children: maybeConvertToAnsiText(
+              // TODO(helin24): Recompute summary length considering ansi codes?
```
jacob314: Good todo. You can remove the question mark as this a valid case to optimize but not critical to optimize now.

packages/devtools_app/lib/src/logging/flutter/logging_screen.dart
```diff
@@ -357,6 +362,33 @@ class _MessageColumn extends LogMessageColumn
...
+        overflow: TextOverflow.ellipsis,
+      );
+
+      return Row(
+        children: <Widget>[
+          richText,
+          Flexible(
+            child: Container(
+              height: 12.0,
+              width: 0.0,
```
jacob314, line 389: is this width zero container needed?

devoncarew:
In the Logging page, it looks like the text in the Message column is sitting several pixels higher than the text baseline?

packages/devtools_app/lib/src/debugger/flutter/console.dart
jacob314: in this case I would add a comma before both parens. That way you get the flutter style with indents of only 2 spaces which makes deeply nested flutter build methods easier to read.

packages/devtools_app/lib/src/logging/flutter/logging_screen.dart
helin24: I don't really understand this param - if i remove `width` entirely, there are blue bars that show up to the right:
Screen Shot 2020-04-23 at 9 32 52 AM

jacob314: you don't want a blue bar. The blue bar is intended only for messages visualizing progress.

helin24: Oh I misunderstood! Yes, I can remove this container.

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -68,10 +69,11 @@ class _ConsoleState extends State<Console> {
                     itemExtent: CodeView.rowHeight,
                     controller: scrollController,
                     itemBuilder: (context, index) {
-                      return Text(
-                        lines[index],
+                      return RichText(
+                        text: TextSpan(
+                            children: maybeConvertToAnsiText(
+                                lines[index], textStyle)),
```
jacob314: there are still two parens next to each other here. In general, you have added enough commas when all indents are 2 spaces. If you see indents of 4 spaces it means you need to add more commas. This is an annoying edge case because the dartformatter doesn't know you are writing code within a widget build method where just defaulting to two spaces indent would be better.

packages/devtools_app/lib/src/flutter/utils.dart
```diff
@@ -14,3 +15,31 @@ Future<void> launchUrl(String url, BuildContext context) async {
...
+                    color: entry.fgColor != null
+                        ? colorFromAnsi(entry.fgColor)
+                        : null,
+                    backgroundColor: entry.bgColor != null
+                        ? colorFromAnsi(entry.bgColor)
+                        : null,
+                    fontWeight:
+                        entry.bold ? FontWeight.bold : FontWeight.normal,
+                  ),
+          ))
```
jacob314, line 38: nit: maybe add a trailing comma here to reduce the indent but this case is fine either way.

packages/devtools_app/test/flutter/debugger_screen_test.dart
```diff
@@ -75,7 +76,46 @@ void main() {
...
+      finder.evaluate().forEach((element) {
+        final richText = element.widget as RichText;
+        final textSpan = richText.text as TextSpan;
+        final secondSpan = textSpan.children[1] as TextSpan;
+        expect(
+          secondSpan.text,
+          'console',
+          reason: 'Text with ansi code should be in separate span',
+        );
+        expect(
```
jacob314, line 106: nit: you should be able to compare equality on the entire color rather than having to have a separate expect for red, green, and blue.

packages/devtools_app/test/flutter/logging_screen_test.dart
```diff
@@ -142,7 +144,9 @@ void main() {
         await tester.tap(find.byKey(ValueKey(fakeLogData[996])));
         await tester.pumpAndSettle();
         expect(
-          find.text('log event 996'),
+          find.byWidgetPredicate((widget) =>
```
jacob314, line 147: there is a test helper for finding a widget with matching text directly rather than having to match a predicate.

packages/devtools_app/test/flutter/logging_screen_test.dart
```diff
@@ -206,6 +216,50 @@ void main() {
...
+        );
+
+        finder.evaluate().forEach((element) {
+          final richText = element.widget as RichText;
+          final textSpan = richText.text as TextSpan;
+          final secondSpan = textSpan.children[1] as TextSpan;
+          expect(
+            secondSpan.text,
+            'log 995',
+            reason: 'Text with ansi code should be in separate span',
```
jacob314, line 247: these reason fields are helpful but don't feel like you always have to add them. in general, test code doesn't need to be quite the same quality as app code.
for example, the reason for the `findNWidgets(2)` case is great but you should feel free to omit this reason as this case as the reason is more just an implementation detail that is typical for any time flutter is rendering text.

packages/devtools_app/test/flutter/logging_screen_test.dart
helin24: It looks like `find.text` only works for `Text` and `EditableText` widgets. Ian suggested either adding keys to rich text or using this type of predicate (https://stackoverflow.com/questions/41152320/is-there-an-easy-way-to-find-particular-text-built-from-richtext-in-a-flutter-te). We could consider keys too, although maybe it's better to use a predicate to avoid adding code that's just for testing.

jacob314: That is too bad. We will be doing enough rich text display in DevTools that it might be worth it for us to add an extension method that makes this case easier.

helin24: Do you mean add a utility function in our devtools code or add something to the tester library?

helin24:
> In the Logging page, it looks like the text in the Message column is sitting several pixels higher than the text baseline?

This alignment issue seems complicated. When using `RichText`, the baseline displayed on inspector is higher than `Text` elements. I found [this suggestion](https://github.com/flutter/flutter/issues/30159#issuecomment-502875830) of using a `WidgetSpan` to specify a baseline, but then the `RichText` includes `WidgetSpan`s that include `Text` widget (instead of `RichText` containing `TextSpan`s), which seems to make testing harder since you can't call `toPlainText` for the full text inside. Maybe this is okay though.

Result:
Screen Shot 2020-04-24 at 2 26 58 PM

packages/devtools_app/test/flutter/logging_screen_test.dart
jacob314: you can add an extension method to our package and it will feel like it is directly part of the tester package in terms of how you call the api.

jacob314:
> > In the Logging page, it looks like the text in the Message column is sitting several pixels higher than the text baseline?
> 
> This alignment issue seems complicated. When using `RichText`, the baseline displayed on inspector is higher than `Text` elements. I found [this suggestion](https://github.com/flutter/flutter/issues/30159#issuecomment-502875830) of using a `WidgetSpan` to specify a baseline, but then the `RichText` includes `WidgetSpan`s that include `Text` widget (instead of `RichText` containing `TextSpan`s), which seems to make testing harder since you can't call `toPlainText` for the full text inside. Maybe this is okay though.
> 
> Result:
> Screen Shot 2020-04-24 at 2 26 58 PM

Try calling Text.rich instead of RichText. I expect that will give you more consistent display with the other text rendered.

helin24:
The alignment issue was due to not having `maxLines: 1` for the logging page, so `WidgetSpan` was not necessary!

PR #1861, opened 2020-04-22
minor UI tweaks to the network page

devoncarew:
Some minor UI tweaks to the network page for similarity with our other pages.

We may want to move away from using `PaginatedDataTable` to display the data - it's hard to get rid of the (small) padding around the Card that that widget uses. Also, I don't think we need the  pagination support that widget provides - we can just scroll for however long the list is.

cc @bkonyi

### DIFF
--- a/packages/devtools_app/lib/src/network/flutter/http_request_inspector.dart
+++ b/packages/devtools_app/lib/src/network/flutter/http_request_inspector.dart
@@ -71,6 +71,7 @@ class HttpRequestInspector extends StatelessWidget {
     );
 
     return Card(
+      margin: EdgeInsets.zero,
       color: Theme.of(context).canvasColor,
       child: Container(
         decoration: BoxDecoration(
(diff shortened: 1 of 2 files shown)


bkonyi:
> We may want to move away from using `PaginatedDataTable` to display the data - it's hard to get rid of the (small) padding around the Card that that widget uses. Also, I don't think we need the pagination support that widget provides - we can just scroll for however long the list is.
> 
> cc @bkonyi

If I recall correctly, there was some performance issues using a standard list view (mind you, I was testing with over 500+ events coming in over 20 seconds).

kenzieschmoll:
> > We may want to move away from using `PaginatedDataTable` to display the data - it's hard to get rid of the (small) padding around the Card that that widget uses. Also, I don't think we need the pagination support that widget provides - we can just scroll for however long the list is.
> > cc @bkonyi
> 
> If I recall correctly, there was some performance issues using a standard list view (mind you, I was testing with over 500+ events coming in over 20 seconds).

We can use a lazy list to solve this.

PR #1865, opened 2020-04-23
Network page cleanup.

jacob314:
Set min widths so ui is always readable.
Fix issue where the network page wouldn't work unless you had started
recording on the timeline first.
Fix crasher issue when part of a network request occurred before the
timeline samle.
Persist controller so network data doesn't disappear when you tab away.


packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -153,6 +168,8 @@ class NetworkController {
     // wall-time a request was made. This won't be 100% accurate, but it should
     // easily be within a second.
     _timelineMicrosOffset = DateTime.now().microsecondsSinceEpoch - timestamp;
+    await allowedError(serviceManager.service
+        .setVMTimelineFlags(<String>['GC', 'Dart', 'Embedder']));
```
kenzieschmoll, line 176: can we remove the type here?

packages/devtools_app/lib/src/flutter/scaffold.dart
```diff
@@ -43,8 +43,11 @@ class DevToolsScaffold extends StatefulWidget {
   /// A [Key] that indicates the scaffold is showing in full-width mode.
   static const Key fullWidthKey = Key('Full-width Scaffold');
 
+  // TODO(jacobr): compute this based on the width of the list of tabs rather
+  // than hardcoding. Computing this width dynamically is even more important
+  // in the presence of conditional screens.
   /// The width at or below which we treat the scaffold as narrow-width.
-  static const double narrowWidthThreshold = 1100.0;
+  static const double narrowWidthThreshold = 1300.0;
```
jacob314, line 50: fyi @kenzieschmoll, it would be nice to clean this up as part of the conditional import support.

packages/devtools_testing/lib/support/http_request_timeline_test_data.json
```diff
@@ -1 +1,5764 @@
...
+      "id": "46",
+      "args": {
+        "error": "SocketException: OS Error: Connection refused, errno = 111, address = 127.0.0.1, port = 35306",
+        "isolateId": "isolates/3907117677703047"
+      }
+    }
+  ],
+  "timeOriginMicros": 231932850307,
+  "timeExtentMicros": 4214988
+}
```
kenzieschmoll, line 5764: new line

packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -100,34 +103,46 @@ class NetworkController {
...
+    if (_httpRecordingNotifier.value == state) {
+      return;
+    }
+
     if (state) {
       // Start polling once we've enabled logging.
       assert(_pollingTimer == null);
       _pollingTimer = Timer.periodic(
-        const Duration(seconds: 1),
+        const Duration(milliseconds: 200),
```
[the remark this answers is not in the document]
jacob314: This timeout is highly dependent on how confident we are in the network page's performance so it shouldn't use a generic timeout just yet. Added a comment with a TODO.

packages/devtools_app/lib/src/network/network_controller.dart
jacob314: done. also removed the type from all other places in devtools we used this api

packages/devtools_testing/lib/support/http_request_timeline_test_data.json
jacob314: done.

packages/devtools_testing/lib/support/http_request_timeline_test_data.json
```diff
@@ -1 +1,5764 @@
...44","args":{"error":"SocketException: OS Error: Connection refused, errno = 111, address = 127.0.0.1, port = 35302","isolateId":"isolates/3907117677703047"}},{"name":"HTTP CLIENT POST","cat":"Dart","tid":52414,"pid":52406,"ts":231937016599,"ph":"b","id":"45","args":{"filterKey":"HTTP/client","method":"POST","uri":"http://127.0.0.1:8011?foo=bar&year=2019","isolateId":"isolates/3907117677703047"}},{"name":"HTTP CLIENT POST","cat":"Dart","tid":52414,"pid":52406,"ts":231937017768,"ph":"e","id":"45","args":{"error":"SocketException: OS Error: Connection refused, errno = 111, address = 127.0.0.1, port = 35304","isolateId":"isolates/3907117677703047"}},{"name":"HTTP CLIENT POST","cat":"Dart","tid":52414,"pid":52406,"ts":231937063645,"ph":"b","id":"46","args":{"filterKey":"HTTP/client","method":"POST","uri":"http://127.0.0.1:8011/foo/bar","isolateId":"isolates/3907117677703047"}},{"name":"HTTP CLIENT POST","cat":"Dart","tid":52414,"pid":52406,"ts":231937065295,"ph":"e","id":"46","args":{"error":"SocketException: OS Error: Connection refused, errno = 111, address = 127.0.0.1, port = 35306","isolateId":"isolates/3907117677703047"}}],"timeOriginMicros":231932850307,"timeExtentMicros":4214988}
```
jacob314, line 1: This file is identical to the previous file except that I intentionally omitted one start event to trigger a problem case I was seeing with a real world app where everything crashed if the first network event was incomplete due to the start of the network event being before the start of the snapshot.

jacob314:
Kenzie, please take another look. I had to add a bit more lifecycle cleanup than I wanted to this CL because making the tests pass without it was harder than fixing it.
TLDR: the controller now tracks the number of clients so we can stop polling when we have zero clients.

We still aren't handling app restart properly. I'll apply that with a followup cl.

packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -168,6 +171,7 @@ class NetworkController {
       // process of starting recording.
       return;
     }
+    _httpRecordingNotifier.value = true;
```
kenzieschmoll, line 174: why are we setting this here if we are calling _toggleHttpTimelineRecording (which also sets the value to true) below in the try statement?

packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -146,15 +166,32 @@ class NetworkController {
...
+
+      // Determine the offset that we'll use to calculate the approximate
+      // wall-time a request was made. This won't be 100% accurate, but it should
+      // easily be within a second.
+      _timelineMicrosOffset = DateTime.now().microsecondsSinceEpoch - timestamp;
+      // TODO(jacobr): add an intermediate manager class to track which flags are
+      // set. We are setting more flags than we probably need to here but setting
+      // fewer flags risks breaking functionality on the timeline view that
+      // assumes that all flags are set.
+      await allowedError(serviceManager.service
```
bkonyi, line 188: It might just be worth invoking `getVMTimelineFlags` first here to get the currently set flags. I doubt you'd see any performance hit from the extra RPC.

jacob314: yeah that would be fine from a performance perspective and is how I'd implement the client side flag manager in the TODO. We might want to add a bit of business logic to the VM in the future to simplify managing this further. For example, decouple it so setting timeline flags in one client doesn't impact other clients. Right now if you stopped recording the network events in one devtools instance you would break a second devtools instance that was also recording timeline events.

PR #1868, opened 2020-04-23
tweak the frames rendering slightly

devoncarew:
- tweak the frames rendering slightly (remove the parans)
- show async break frames in a different style
- address an NPE when rendering async frames (there is no associated source location)

Screen Shot 2020-04-23 at 1 54 57 PM

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/call_stack.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/call_stack.dart
@@ -61,6 +61,41 @@ class _CallStackState extends State<CallStack> {
     final subtleStyle = TextStyle(color: theme.unselectedWidgetColor);
     final selectedStyle = TextStyle(color: theme.textSelectionColor);
 
+    Widget child;
+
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/debugger/flutter/call_stack.dart
```diff
@@ -61,27 +61,49 @@ class _CallStackState extends State<CallStack> {
...
+      );
+    } else {
+      child = RichText(
+        maxLines: 1,
+        overflow: TextOverflow.ellipsis,
+        text: TextSpan(
+          text: _descriptionFor(frame),
+          style: selected ? selectedStyle : regularStyle,
+          children: [
+            if (!asyncMarker)
```
kenzieschmoll, line 89: this if check shouldn't be necessary since we are in the else statement of `if (asyncMarker)`

devoncarew: removed

PR #1869, opened 2020-04-23
address an exception when reparenting widgets

devoncarew:
- address an exception when reparenting widgets
- use a global key to have the code area remember its scroll position even when the parent widgets change (we add and remove a Split widget above it)

```
════════ Exception caught by widgets library ═══════════════════════════════════════════════════════
The following assertion was thrown building Expanded(flex: 1):
LinkedScrollControllerGroup does not have any scroll controllers attached.
'package:devtools_app/src/flutter/flutter_widgets/linked_scroll_controller.dart':
Failed assertion: line 45 pos 7: '_attachedControllers.isNotEmpty'

The relevant error-causing widget was: 
  Expanded devtools/packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart:133:11
#2      LinkedScrollControllerGroup.position (package:devtools_app/src/flutter/flutter_widgets/linked_scroll_controller.dart:45:7)
#3      _CodeViewState._updatePausedPositions (package:devtools_app/src/debugger/flutter/codeview.dart:115:43)
#4      _CodeViewState.initState (package:devtools_app/src/debugger/flutter/codeview.dart:63:5)
#5      StatefulElement._firstBuild (package:flutter/src/widgets/framework.dart:4644:58)
#6      ComponentElement.mount (package:flutter/src/widgets/framework.dart:4480:5)
...
════════════════════════════════════════════════════════════════════════════════════════════════════
```

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/codeview.dart
@@ -69,9 +69,6 @@ class _CodeViewState extends State<CodeView> {
 
     if (widget.script != oldWidget.script) {
       _updateLines();
-    }
-
-    if (widget.stack != oldWidget.stack) {
       _updatePausedPositions();
     }
   }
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/debugger/flutter/codeview.dart
```diff
@@ -69,9 +69,6 @@ class _CodeViewState extends State<CodeView> {
 
     if (widget.script != oldWidget.script) {
       _updateLines();
-    }
-
-    if (widget.stack != oldWidget.stack) {
```
jacob314, line 74: can you comment why we want to always update the paused positions even if the stack has not changed?

devoncarew:
I may drop this PR as I'm not 100% why it works, and I have some refactoring in this area for how we jump to source locations -

jacob314:
removing the setState calls from methods called only from initState and didUpdateDependencies is a good thing. I thought there was an exception thrown if you called setState from initState but I may be remembering that wrong as otherwise there should have been bigger problems. Dropping the equality check in the didUpdateWidget case is more worrying.

devoncarew:
closing in favor of an upcoming PR

PR #1873, opened 2020-04-24
Upgrade to vm_service 4.0.2 and use getClassList method in debugger.

kenzieschmoll:


packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -370,26 +369,19 @@ class DebuggerController extends DisposableController
...
-    (_service as VmServiceWrapper)
-        .getClassList(isolateRef.id)
-        .then((Response response) {
-      final List classes = response.json['classes'];
-      final List<ClassRef> classRefs = classes
-          .map((json) => ClassRef.parse(json))
-          .cast<ClassRef>()
+    try {
+      final classList = await _service.getClassList(isolateRef.id);
+      final classes = classList.classes
```
devoncarew, line 374: Shouldn't this end in a semi-colon? Perhaps I'm not reading it right.

You might want to make a copy of the list (`toList()`), to guard against that returned list not being mutable before you sort it.

bkonyi: Yes, it looks like it should.

bkonyi: No, it looks like there's more on the next line that wasn't changed.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -370,26 +369,19 @@ class DebuggerController extends DisposableController
...
+      classes.sort((ClassRef a, ClassRef b) {
         // We sort uppercase so that items like Foo sort before items like _Foo.
         return a.name.toUpperCase().compareTo(b.name.toUpperCase());
       });
-
-      _sortedClasses.value = classRefs;
-    }).catchError((e) {
+      _sortedClasses.value = classes;
+    } catch (e) {
       // Fail gracefully - not all clients support getClassList().
```
bkonyi, line 383: Is this actually the case? This RPC has been around for years but it just wasn't documented properly. Since it wasn't private, I doubt you'll run into a situation where this RPC isn't present.

devoncarew: Ah - the web backends (`package:dwds`) won't have this call yet :)

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
devoncarew: Ah, yeah. And it calls `toList()` at the end (after a `where`) so no need to make another copy.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
bkonyi: Of course... I always seem to forget that the VM service protocol isn't only implemented by the VM service :)

kenzieschmoll: we would run into the situation where the getClassList method isn't available from package:vm_service though (pre 4.0.2)

PR #1874, opened 2020-04-24
refactor how we navigate to scripts in the debugger

devoncarew:
Refactor how we navigate to scripts in the debugger:
- allow to jump to ScriptRefs (files) as well as associated SourceLocations (lines)
- handle cases where some Frames don't have associated lines
- be more targeted about what we listen to (that causes rebuilds) in DebuggerScreen
- have more granular listening for changes in CodeView (the entire widget only rebuilds if the ScriptRef its pointing to changes)
- update tests


devoncarew:
After this PR we can jump to things with positions associated with them (breakpoints, classes, frames), and have a notion of a single selected frame, which is where we draw the execution point.

packages/devtools_app/lib/src/debugger/flutter/breakpoints.dart
```diff
@@ -114,14 +114,6 @@ class _BreakpointPickerState extends State<BreakpointPicker> {
 
   Future<void> _onBreakpointSelected(BreakpointAndSourcePosition bp) async {
```
jacob314, line 115: this method no longer needs to be async.

jacob314: @pq do we have any lints that track when a method body is async but nothing async is actually happening?

packages/devtools_app/lib/src/debugger/flutter/codeview.dart
```diff
@@ -22,57 +22,92 @@ import 'debugger_model.dart';
...
+    widget.controller.scriptLocation.addListener(_handleScriptLocationChanged);
+  }
+
+  void _initScriptInfo() {
+    script = widget.controller.getScriptCached(scriptRef);
+
+    if (script == null) {
+      if (scriptRef != null) {
+        widget.controller.getScript(scriptRef).then((script) {
+          setState(() {
```
jacob314, line 72: make sure the scriptRef that was specified is still the relevant scriptRef for the codeview after the async operation completes and that the widget has not been disposed.

packages/devtools_app/lib/src/debugger/flutter/codeview.dart
```diff
@@ -22,57 +22,92 @@ import 'debugger_model.dart';
...
   }
 
   @override
   void didUpdateWidget(CodeView oldWidget) {
     super.didUpdateWidget(oldWidget);
 
-    if (widget.script != oldWidget.script) {
-      _updateLines();
+    if (widget.controller != oldWidget.controller) {
+      oldWidget.controller.scriptLocation
```
jacob314, line 103: use autoDispose here to avoid having to manually remove the old listener assuming that all old listeners can be disposed before adding new ones.

packages/devtools_app/lib/src/debugger/flutter/codeview.dart
```diff
@@ -82,140 +117,208 @@ class _CodeViewState extends State<CodeView> {
...
-      }
-    });
-  }
+    final location = widget.controller.scriptLocation.value?.location;
+    if (location?.line == null) {
+      return;
+    }
+
+    if (!verticalController.hasAttachedControllers) {
+      // TODO(devoncarew): I'm uncertain why this occurs.
```
jacob314, line 140: is this code path getting triggered after the stateful widget is disposed?

jacob314: change to log from print or remove.

packages/devtools_app/lib/src/debugger/flutter/debugger_model.dart
```diff
@@ -6,13 +6,48 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart' hide Stack;
 import 'package:vm_service/vm_service.dart';
 
+/// A tuple of a script and an optional location.
+class ScriptLocation {
+  ScriptLocation(this.scriptRef, {this.location});
+
+  final ScriptRef scriptRef;
+
+  /// This field can be nullable.
```
jacob314, line 15: nit: This field can be nullable --> This field is nullable.

packages/devtools_app/lib/src/debugger/flutter/debugger_model.dart
```diff
@@ -6,13 +6,48 @@ import 'package:flutter/foundation.dart';
...
+
+  @override
+  bool operator ==(other) {
+    return other is ScriptLocation &&
+        other.scriptRef == scriptRef &&
+        other.location == location;
+  }
+
+  @override
+  int get hashCode => (scriptRef.hashCode << 7) ^ (location?.hashCode ?? 0);
```
jacob314, line 26: this is a fine hashcode but might want to use the flutter helper method for generating a hashcode of a couple fields just to keep people on the right track if copying this code.

packages/devtools_app/lib/src/debugger/flutter/breakpoints.dart
devoncarew: removed async

packages/devtools_app/lib/src/debugger/flutter/debugger_model.dart
devoncarew: changed

packages/devtools_app/lib/src/debugger/flutter/debugger_model.dart
devoncarew: Ah, I didn't know about the `hashValues()` util - will change to use it.

packages/devtools_app/lib/src/debugger/flutter/codeview.dart
devoncarew: will do

packages/devtools_app/lib/src/debugger/flutter/codeview.dart
devoncarew: changed to `log()`

devoncarew: I added a guard on one of the code paths here (`if (mounted)`); I left the check log code to see if this still fires.

packages/devtools_app/lib/src/debugger/flutter/codeview.dart
devoncarew: Added a todo: to investigate this. There are some other things we're disposing, and I still need to get a better handle on the widget lifecycle and the AutoDispose mixin.

devoncarew:
PTAL - I believe I addressed all comments (modulo one about using AutoDispose, where I added a TODO:).

packages/devtools_app/lib/src/debugger/flutter/codeview.dart
jacob314: If use addAutoDisposeListener then the listener is automatically disposed when dispose on the widget is called. You can additionally dispose all existing listeners typically in `didChangeDependencies` by calling `cancel` manually.
There is a good example from the logging view.
https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/logging/flutter/logging_screen.dart#L88

https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/logging/flutter/logging_screen.dart#L88

devoncarew: OK, I re-read the code and that example, and updated this to use AutoDisposeMixin.

packages/devtools_app/lib/src/debugger/flutter/breakpoints.dart
pq: @jacob314: we don't but it's something to consider.

Tracking issue opened in https://github.com/dart-lang/linter/issues/2087.

PR #1878, opened 2020-04-27
show the column for the execution location

devoncarew:
- show the column for the execution location (this helps when we need to step several times in a single source line)
- use the getSourceReport() call from the VM to better know which lines can have breakpoints
- remove the 'Console' title area for the stdout area - people can figure out what it is on context, and it gives us more vertical space for debugger content

Screen Shot 2020-04-28 at 4 53 16 PM

Screen Shot 2020-04-28 at 4 54 13 PM


packages/devtools_app/lib/src/debugger/flutter/codeview.dart
```diff
@@ -422,29 +438,98 @@ class LineItem extends StatelessWidget {
...
+        children: [
+          Text(
+            lineContents,
+            maxLines: 1,
+            overflow: TextOverflow.ellipsis,
+          ),
+          Row(
+            children: [
+              Text(' ' * (column - 1)),
+              Transform.translate(
```
jacob314, line 475: consider tweaking the background color of all text after the column rather than drawing the arrow.

packages/devtools_app/lib/src/debugger/flutter/codeview.dart
```diff
@@ -422,29 +438,98 @@ class LineItem extends StatelessWidget {
...
     );
   }
 }
+
+Color _darken(Color c, [double percent = 0.05]) {
+  assert(0.0 <= percent && percent <= 1.0);
+
+  percent = 1.0 - percent;
+
+  return Color.fromARGB(
```
jacob314, line 518: these seem like reasonable extension methods on the `dart:ui` `Color` class.

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -40,44 +39,35 @@ class _ConsoleState extends State<Console> {
...
-                builder: (context, lines, _) {
-                  // If we're at the end already, scroll to expose the new
-                  // content.
-                  if (scrollController.hasClients &&
-                      scrollController.atScrollBottom) {
-                    scrollController.autoScrollToBottom();
-                  }
+      child: Padding(
+        padding: const EdgeInsets.all(denseSpacing),
+        child: ValueListenableBuilder<List<String>>(
```
jacob314, line 44: nit: consider removing the List<String> type as it can be inferred.

PR #1885, opened 2020-04-29
Memory Snapshot

terrylucas:
Enable collecting VM snapshot of of the heap.

image

Clicking on the "Snapshot" button will collect a snapshot of the objects in the Dart VM heap. There are 3 phases of building the snapshot:

1. collecting the raw data from the VM "Processing..."
2. building the graph "Graphing..."
3. organizing the data for display "Grouping..."

During these phases a spinning progress indicator appears displaying each phase e.g.,

image

Once the data has been processed, graphed and grouped then a visual display of the largest aggregate chunks of data is displayed in a heat map style presentation:

image

In this case, the **root** node (top-most parent displayed at the bottom of the window) is the total size for an application, all of its active memory, at a particular point in time (when the 'Snapshot' button was pressed). The size of the root should match the size of the graph, in this case ~156 MB

image

Back to the heap map, each rectangle stacked on top of the **root** rectangle (or node) is each major child's size by library name.  Some libraries are grouped to help aide in understand the larger memory picture e.g., _dart:*_ and _package:flutter/*_.  For example, clicking on package:flutter rectangle/node will display:

image

or typing package:fluter in the search area e.g.,

image

The search field will immediately select a rectangle once a name matches a library or class name.  The match must be exact, this will be relaxed in the future as well as an auto-complete list of matching nodes.  Deleting characters from the search area will immediately change the heat map displayed, if no matches the **root** based map is displayed (as if the **root** node was selected).  Pressing the `ESC key` clears the entire search area, as if the **root** node was selected.

The heat map is designed to be a 10,000 foot visual overview of memory usage grouped by library, such that a small memory footprint library may appear as a sliver while large memory usage libraries will appear as proportionally large rectangles in the heat map display.  Currently, the heat map is of all libraries filtering of libraries and classes is not applied to the heat map.

A second view, table view of the snapshot is available.  The table view is a detailed view of all memory and as such has filtering and inspecting which the heat map currently doesn't have.  In the future, the ability to move back and forth between heat map and table view should exist.  To switch to table view turn off the **Heat Map** switch

image

To display a table view:

image

An application may have hundreds of libraries, thousands of classes and millions of instances (objects) where the viewing and inspecting the tables, trees and lists of data may be daunting.  To help narrow down the sheer volume of data, filtering has been added (future search style access in table view will be added). 

Clicking on the **Filter** button will display a filter dialog:

image

image

Prior to a snapshot a few filters have been automatically turned on:
- Hide Private Classes
- Hide Classes with No Instances
- Hide Library with No Instances

Additionally, filtering is designed to eliminate extraneous libraries so framework libraries are default hidden:
- dart:*
- package:collection
- package:flutter*
- package:intl
- package:vector_math

Disabling any of these filters will display those libraries, classes and memory objects in the table view.

Once, a snapshot exist then the total known set of libraries will also appear under the 'Hide Libraries or Packages:' list e.g.,

image

In this example, 11 total libraries are known (given dart:* and package:flutter* are enabled).  If the wildcard filters are unchecked then 10 more libraries appear (dart:core, dart:_internal, dart:developer, dart:ui, etc.).

Table view has four columns (clicking on a column header sorts the rows by the column):
1. Library or Class
2. Count
3. Shallow
4. Retained

image

'Library or Class' column is a hierarchical column (tree) which is **Group by Library** is selected then the list is organized by library/package names.  There are two special entries, currently without children, the **External** is size of the external heap and **Filtered** is the size of all memory objects hidden.

Expanding a row e.g., **package:gallery** displays its children which are classes, clicking on a class e.g., **AppStateModel** will display the active instances AppStateModel.  When an instance is selected then an instance viewer table will appear to the right of the table e.g.,

image

The Instance View table has 3 columns, each sortable:
1. Type
2. Name
3. Value

Each row is a field of the class its type, name and value.  If a type is an object then it can be expanded as well. It's only a view of the values.


packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -48,6 +51,41 @@ class MemoryController extends DisposableController
 
   static const String logFilenamePrefix = 'memory_log_';
 
+  bool showHeatMap = true;
+
+  final List<Snapshot> snapshots = [];
+
+  Snapshot get lastSnapshot => snapshots.isNotEmpty ? snapshots.last : null;
```
jacob314, line 58: we have a `safeLast` extension method that you can call here to save a couple characters.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -48,6 +51,41 @@ class MemoryController extends DisposableController
...
+
+  /// Notifies that the source of the memory feed has changed.
+  ValueListenable get selectedSnapshotNotifier => _selectedSnapshotNotifier;
```
jacob314, line 61: specify the value type as part of this getter.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -48,6 +51,41 @@ class MemoryController extends DisposableController
...
+
+  final _selectedSnapshotNotifier = ValueNotifier<String>('');
+
+  set selectedSnapshotTimestamp(String snapshotTimestamp) {
+    _selectedSnapshotNotifier.value = snapshotTimestamp;
+  }
+
+  String get selectedSnapshotTimestamp => _selectedSnapshotNotifier.value;
+
+  HeapGraph theGraph;
```
jacob314, line 71: nit: rename from `theGraph` to `heapGraph`

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -48,6 +51,41 @@ class MemoryController extends DisposableController
...
+
+  /// Leaf node of tabletree snapshot selected?  If selected then the instance
+  /// view is displayed to view the fields of an instance.
+  /// Notifies that the timeline has been paused.
+  final _leafSelectedNotifier = ValueNotifier<HeapGraphElementActual>(null);
+
+  ValueListenable get leafSelectedNotifier => _leafSelectedNotifier;
```
jacob314, line 78: specify the type of the Value listenable here as well.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -48,6 +51,41 @@ class MemoryController extends DisposableController
...
+
+  ValueNotifier<HeapGraphElementActual> get selectedLeaf =>
+      _leafSelectedNotifier;
+
+  bool get isLeafSelected => selectedLeaf.value != null;
+
+  void setSelectedLeaf(HeapGraphElementActual selected) {
```
jacob314, line 85: make this a setter instead of `setSelectedLeaf`

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -223,9 +261,58 @@ class MemoryController extends DisposableController
 
   final SettingsModel settings = SettingsModel();
 
+  /// Tree to view Libary/Class/Instance
```
jacob314, line 264: avoid acronyms. is there a single word that you can use instead of `lci`?
`libraryClassInstanceTreeTable` would be accurate but a bit long :/

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -223,9 +261,58 @@ class MemoryController extends DisposableController
...
+  ///
+  /// The key is the group name (displayed in dialog), value is list of
+  /// libraries associated to the group. Structure is used to:
+  ///    - allow user to hide/show a library when filtering a snapshot.
+  ///    - create the hide list of libraries to drive the UX on when showing
+  ///      list of libraries/classes/objects in a snapshot table.
+  final filteredLibrariesByGroupName = <String, List<LibraryFilter>>{};
+
+  /// Notify that the filtering has changed.
+  ValueListenable get filterNotifier => _filterNotifier;
```
jacob314, line 291: add type to the lhs.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -223,9 +261,58 @@ class MemoryController extends DisposableController
...
+  bool filterLibraryNoInstances = true;
+
+  /// Table ordered by library, class or instance
+  static const String groupByLibrary = 'Library';
+  static const String groupByClass = 'Class';
+  static const String groupByInstance = 'Instance';
+
+  final groupingBy = ValueNotifier<String>(groupByLibrary);
+
+  ValueListenable get groupingByNotifier => groupingBy;
```
jacob314, line 315: add type to LHS

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
```diff
@@ -0,0 +1,391 @@
...
+    if (firstPart.startsWith(_dartLibraryUriPrefix)) {
+      return _dartLibraryUriPrefix;
+    } else if (firstPart.startsWith(_flutterLibraryUriPrefix)) {
+      return _flutterLibraryUriPrefix;
+    } else {
+      return firstPart;
+    }
+  }
+
+  List<String> get librariesFiltered =>
```
jacob314, line 48: nit: consider using Set<String> everywhere here for quicker membership tests.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
```diff
@@ -0,0 +1,391 @@
...
+}
+
+/// State of the libraries and packages (hidden or not) for the filter dialog.
+class LibraryFilter {
+  LibraryFilter(this.normalized, this.hide);
+
+  /// Displayed library name.
+  final String normalized;
+
+  /// Classes in this library hidden.
```
jacob314, line 102: `/// Classes in this library hidden.` --> `/// Whether classes in this library are hidden.`

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
```diff
@@ -0,0 +1,391 @@
...
+      // Is this library filtered?
+      _filteredLibraries.contains(libraryName);
+}
+
+/// State of the libraries and packages (hidden or not) for the filter dialog.
+class LibraryFilter {
+  LibraryFilter(this.normalized, this.hide);
+
+  /// Displayed library name.
+  final String normalized;
```
jacob314, line 100: nit: normalized --> displayName?

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
```diff
@@ -0,0 +1,391 @@
...
+
+    _controllers = LinkedScrollControllerGroup();
+    _letters = _controllers.addAndGet();
+  }
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+
+    controller = widget.controller;
```
jacob314, line 166: nit: short circuit if `controller == widget.controller` as can commonly happen after a hot reload.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
```diff
@@ -0,0 +1,391 @@
...
+
+    cancel();
+  }
+
+  @override
```
jacob314, line 171: remove no-op dispose method.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
```diff
@@ -0,0 +1,391 @@
...
+  void dispose() {
+    super.dispose();
+  }
+
+  void addLibrary(String libraryName, [bool hideState = false]) {
```
jacob314, line 176: nit: make hideState a named parameter.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
```diff
@@ -0,0 +1,391 @@
...
+
+    return Dialog(
+      shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(20.0)),
+      child: LayoutBuilder(
+        builder: (BuildContext context, BoxConstraints constraints) {
+          return Container(
+            width: MediaQuery.of(context).size.width / 3,
+            height: constraints.maxHeight < 500
+                ? constraints.maxHeight
+                : constraints.maxHeight / 3 + 300,
```
jacob314, line 261: document why `/3 + 300` is a good value here.
It probably makes sense to choose a value so that 500 (or the threshold) would have the same height regardless. 
So for example,
`constraints.maxHeight / 3 + (500*2/3)`

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
```diff
@@ -0,0 +1,391 @@
...
+                                  controller.filterZeroInstances = value;
+                                });
+                              }),
+                          const Text('Hide Classes with No Instances'),
+                        ],
+                      ),
+                      Row(
+                        mainAxisAlignment: MainAxisAlignment.start,
+                        children: [
+                          Checkbox(
```
jacob314, line 305: Can you add a ValueListenableCheckbox class to our shared UI library?
That would reduce the boilerplate for this code and would be handy elsewhere.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
```diff
@@ -0,0 +1,391 @@
...
+                      Row(
+                        mainAxisAlignment: MainAxisAlignment.start,
+                        children: [
+                          const Padding(padding: EdgeInsets.only(top: 30)),
+                          Text('Hide Libraries or Packages '
+                              '(${controller.filteredLibrariesByGroupName.length}):'),
+                        ],
+                      ),
+                      SizedBox(
+                        height: constraints.maxHeight / 4,
```
jacob314, line 324: Ah... one of the classic Flutter layout friction points. Fyi @csells, @InMatrix 
I'm guessing you had to add the SizedBox with a custom size to work around an exception placing a ListView within a flex layout.

Is making this listview scroll separately from the rest of the dialog what you want? If not,
`Column(
  children: <Widget>[
    ListView(
      shrinkWrap: true, // use it
    )
  ],
)`
and a SingleChildScrollView wrapping the column might be a better solution/.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
```diff
@@ -0,0 +1,391 @@
...
+                          }).toList(),
+                        ),
+                      ),
+                      const Padding(padding: EdgeInsets.only(top: 40)),
+                      Row(
+                        mainAxisAlignment: MainAxisAlignment.spaceEvenly,
+                        children: [
+                          DialogOkButton(
+                            () {
+                              // Re-generate librariesFiltered
```
jacob314, line 352: move this to a separate helper method as this is a lot of business logic to include inline in the build method.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
```diff
@@ -0,0 +1,391 @@
...
+  SnapshotFilterState createState() => SnapshotFilterState();
+}
+
+/// A FlatButton used to close a containing dialog - Cancel.
+class DialogCancelButton extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return FlatButton(
+      onPressed: () {
+        Navigator.of(context, rootNavigator: false).pop('dialog');
```
jacob314, line 121: can you make 'dialog' a constant?
Also promote these widgets to a shared ui component folder in devtools as they seem to apply to all dialogs we would want in the future as long as we use consistent conventions.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
```diff
@@ -0,0 +1,439 @@
...
+import 'package:vm_service/vm_service.dart';
+
+import '../flutter/memory_controller.dart';
+
+// TODO(terry): Ask Ben, what is a class name of ::?
+/// Internal class names :: automatically filter out.
+const String internalClassName = '::';
+
+// TODO(terry): Bake in types instead of comparing a fully qualified class name.
+const String predefinedNull = 'dart:core,Null';
```
jacob314, line 14: nit: make these be simple const objects with library and class fields  instead of comma separated strings

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
```diff
@@ -0,0 +1,439 @@
...
+
+/// List of classes to monitor, helps to debug particular class structure.
+final Map<int, String> _monitorClasses = {};
+
+/// Ensure the classId is zero based ()
+bool monitorClass({int classId, String className, String message = ''}) {
+  if (classId != null) {
+    if (_monitorClasses.containsKey(classId)) {
+      final className = _monitorClasses[classId];
+      print('STOP: class $className [classId=$classId]  $message');
```
jacob314, line 48: remove print statements here and elsewhere before committing

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
```diff
@@ -0,0 +1,439 @@
...
+    this.builtInClasses,
+    this.classSentinel,
+    this.classes,
+    this.elementSentinel,
+    this.elements,
+  );
+
+  final MemoryController controller;
+
+  bool instancesComputed = false;
```
jacob314, line 167: nit: make private perhaps with a getter.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
```diff
@@ -0,0 +1,439 @@
...
+  /// Group all classes by libraries - with applied filters.
+  final Map<String, List<HeapGraphClassActual>> groupByLibrary = {};
+
+  /// Group all instances by all classes.
+  final Map<String, List<HeapGraphElementActual>> rawGroupByClass = {};
+
+  /// Group all instances by all classes - with applied filters.
+  final Map<String, List<HeapGraphElementActual>> groupByClass = {};
+
+  String normalizeLibraryName(HeapSnapshotClass theClass) {
```
jacob314, line 196: document what the normalization is. Also move to a shared location. we'll want to normalize with other normalization schemes used in the debugger and in the future in the inspector.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
```diff
@@ -0,0 +1,439 @@
...
+
+    if (scheme == 'package' || scheme == 'dart') {
+      return '$scheme:${uri.pathSegments[0]}';
+    }
+
+    assert(theClass.libraryName.isEmpty);
+    return 'src';
+  }
+
+  bool computeRawGroups() {
```
jacob314, line 208: as it always returns true, make void.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
```diff
@@ -0,0 +1,439 @@
...
+    // Only compute once.
+    if (rawGroupByLibrary.isNotEmpty || rawGroupByClass.isNotEmpty) return true;
+
+    for (HeapGraphClassActual c in classes) {
+      final StringBuffer sb = StringBuffer();
```
jacob314, line 213: omit types on the lhs in this method that can be inferred.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
```diff
@@ -0,0 +1,439 @@
...
+    });
+
+    return true;
+  }
+
+  // TODO(terry): Need dominator graph for flow.
+  /// Compute all instances, needed for retained space.
+  void computeInstancesForClasses() {
+    if (!instancesComputed) {
+      for (var i = 0; i < elements.length; i++) {
```
jacob314, line 278: use `for (var instance in elements)` instead

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
```diff
@@ -0,0 +1,439 @@
...
+
+  List<HeapGraphElement> get references {
+    if (_references == null && referencesFiller != null) {
+      _references = [];
+      referencesFiller();
+    }
+    return _references;
+  }
+
+  String getPrettyPrint(Map<Uri, Map<String, List<String>>> prettyPrints) {
```
jacob314, line 302: avoid prefixing method names with the word `get`
this could also be a good method to document a little bit as it isn't immediately clear to me what it does.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
```diff
@@ -0,0 +1,439 @@
...
+}
+
+/// Object marked for removal on next GC.
+class HeapGraphElementSentinel extends HeapGraphElement {
+  @override
+  String toString() => 'HeapGraphElementSentinel';
+}
+
+/// Live element.
+class HeapGraphElementActual extends HeapGraphElement {
```
jacob314, line 344: Maybe `HeapGraphElementLive` or `LiveHeapGraphElement` instead of `HeapGraphElementActual`?

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -0,0 +1,848 @@
...
+  @visibleForTesting
+  static const searchButtonKey = Key('Snapshot Search');
+  @visibleForTesting
+  static const settingsButtonKey = Key('Snapshot Settings');
+
+  MemoryController controller;
+
+  Widget snapshotDisplay;
+
+  @override
```
jacob314, line 58: remove this method as it is a no-op

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -0,0 +1,848 @@
...
+    super.initState();
+  }
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+
+    controller = Provider.of<MemoryController>(context);
+
+    cancel();
```
jacob314, line 69: short circuit if the new controller is the same as the old controller as will usually be the case after a hot reload.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -0,0 +1,848 @@
...
+
+    addAutoDisposeListener(controller.selectedSnapshotNotifier, () {
```
jacob314, line 71: this autoDisposeListener calls are listening for some state on the controller to trigger some other state on the controller. 
Instead of doing that, add the listeners directly to the controller and instead have the view just listen for when state changes as a result of computeAllLibraries.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -0,0 +1,848 @@
...
+    });
+
+    addAutoDisposeListener(controller.filterNotifier, () {
+      setState(() {
+        controller.computeAllLibraries(true, true);
+      });
+    });
+
+    addAutoDisposeListener(controller.leafSelectedNotifier, () {
+      setState(() {});
```
jacob314, line 84: it is good to avoid empty setState calls whenever possible. can you make the ui that depends on the selected leaf just listen for this valuelistenable directly?

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -0,0 +1,848 @@
...
+    });
+  }
+
+  @override
```
jacob314, line 88: remove no-op dispose method

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -0,0 +1,848 @@
...
+  static const groupByMenuItem = Key('Filter Group By Menu Item');
+  @visibleForTesting
+  static const groupByKey = Key('Filter Group By');
+
+  Widget _groupByDropdown() {
+    final _groupByTypes = [
+      MemoryController.groupByLibrary,
+      MemoryController.groupByClass,
+      MemoryController.groupByInstance,
+    ].map<DropdownMenuItem<String>>(
```
jacob314, line 148: consider using list comprehension for loops instead. I think it makes code like this simpler.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -0,0 +1,848 @@
...
+      },
+    ).toList();
+
+    return DropdownButton<String>(
+      key: groupByMenuButtonKey,
+      value: controller.groupingBy.value,
+      iconSize: 20,
+      style: const TextStyle(fontWeight: FontWeight.w100),
+      onChanged: (String newValue) {
+        setState(
```
jacob314, line 169: in general, calls to a controller should be outside a setState block as the controller can be modified without calling setState on a particular view using it.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -0,0 +1,848 @@
...
+                          if (snapshotDisplay is MemorySnapshotTable) {
+                            controller.lciTreeTable.dataRoots.every((element) {
+                              element.collapseCascading();
+                              return true;
+                            });
+                            if (controller.instanceFieldsTreeTable != null) {
+                              // We're collapsing close the fields table.
+                              controller.setSelectedLeaf(null);
+                            }
+                            setState(() {});
```
jacob314, line 238: what would it take to eliminate this no-op setState?

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -0,0 +1,497 @@
...
+import 'package:flutter/widgets.dart' hide TextStyle;
+import 'package:intl/intl.dart';
+
+import 'memory_graph_model.dart';
+
+class FlameChart extends StatelessWidget {
+  const FlameChart(
+    this.sizes, {
+    // Flame chart has a blueish color.
+    this.lightColor = const Color(0xFFBBDEFB),
```
jacob314, line 20: Add these colors as constants to
https://github.com/flutter/devtools/blob/3ff20fc3d75fdba37f5340c0a3f96858f52e9ab2/packages/devtools_app/lib/src/ui/colors.dart
that way we can make sure we use identical colors for all cases we are doing the same thing. For example. can we reuse some of the existing flame chart colors used by the timeline.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -0,0 +1,497 @@
...
+
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart' hide TextStyle;
+import 'package:flutter/rendering.dart' hide TextStyle;
+import 'package:flutter/widgets.dart' hide TextStyle;
+import 'package:intl/intl.dart';
+
+import 'memory_graph_model.dart';
+
+class FlameChart extends StatelessWidget {
```
jacob314, line 16: Add a top level comment calling out how this flame chart differs from the other flame chart we support. The main difference I'm aware of is the difference in the navigation scheme. It could make sense to in the future have an option to toggle between the two navigation schemes.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -0,0 +1,497 @@
...
+    final Paragraph paragraph = builder.build()
+      ..layout(ParagraphConstraints(width: width - 5));
+    context.canvas.drawParagraph(
+      paragraph,
+      Offset(10 + left, size.height - top + 10),
+    );
+  }
+
+  double _paintAncestors(PaintingContext context, List<Node> nodes) {
+    double top = 50;
```
jacob314, line 217: make 50 a line height constant?

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -0,0 +1,497 @@
...
+            libraryName = libraryGroup;
+          }
+
+          // Map class names to familar user names.
+          final predefined = predefinedClasses['$libraryName,$className'];
+          if (predefined != null) {
+            className = predefined.prettyName;
+          }
+
+          // TODO(terry): Remove testing really big objects.
```
jacob314, line 302: remove this commented out code

packages/devtools_app/lib/src/memory/flutter/memory_screen.dart
```diff
@@ -85,6 +85,16 @@ class MemoryBodyState extends State<MemoryBody> {
     if (newController == controller) return;
     controller = newController;
 
+    // Update the chart when the memorySource changes.
+    addAutoDisposeListener(controller.selectedSnapshotNotifier, () {
```
jacob314, line 89: instead of doing this wire-up here, make sure the controller properly tracks flowing through changes to selectedSnapshotNotifier itself only making this widget update when the new snapshot gets created.

packages/devtools_app/lib/src/trees.dart
```diff
@@ -106,6 +106,9 @@ class TreeNode<T extends TreeNode<T>> {
     _isExpanded = false;
   }
 
+  /// Override to handle pressing on a Leaf node.
```
jacob314, line 109: is this `onSelectLeaf`, `selectLeaf`, or similar rather than `leaf`?

jacob314: Fyi @kenzieschmoll can you take a look at the tree changes to support leaf nodes as well?

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -0,0 +1,497 @@
...
+      if (className != null && className.isNotEmpty) className,
+      name,
+    ];
+  }
+}
+
+class HeatMapSizeAnalyzer extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return MaterialApp(
```
jacob314, line 450: why do we need an extra MaterialApp widget here?

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -0,0 +1,497 @@
...
+      color: Colors.blueGrey,
+      home: FlameChart(InstructionsSize.fromSnapshop(heapGraph)),
+      debugShowCheckedModeBanner: false,
+    );
+
+    // TODO(terry): Testing search needs to be stateful.
```
jacob314, line 456: remove this commented out code or implement

packages/devtools_app/lib/src/memory/flutter/memory_snapshot_models.dart
```diff
@@ -0,0 +1,657 @@
...
+      : controller = null,
+        name = sentinelName,
+        isLibrary = false,
+        isExternal = false,
+        isFiltered = false,
+        isClass = false,
+        isObject = false,
+        actualClass = null;
+
+  Reference.createLibrary(this.controller, this.name,
```
jacob314, line 42: nit. make this
`Reference.library` instead of `Reference.createLibrary` same for the others to follow dart conventions for constructor names.

packages/devtools_app/lib/src/memory/flutter/memory_snapshot_models.dart
```diff
@@ -0,0 +1,657 @@
...
+    controller.setSelectedLeaf(objectReference.instance);
+    if (onLeaf != null) onLeaf(this);
+
+    super.leaf();
+  }
+}
+
+class LibraryReference extends Reference {
+  LibraryReference(
+      MemoryController controller, String libraryName, this.actualClasses)
```
jacob314, line 149: nit: add trailing comma after `actualClasses`

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
```diff
@@ -0,0 +1,391 @@
...
+import 'package:vm_service/vm_service.dart';
+
+import '../../flutter/auto_dispose_mixin.dart';
+import '../../flutter/flutter_widgets/linked_scroll_controller.dart';
+
+import 'memory_controller.dart';
+import 'memory_snapshot_models.dart';
+
+/// First two libraries are special e.g., dart:* and package:flutter*
+const String _dartLibraryUriPrefix = 'dart:';
```
kenzieschmoll, line 15: these types can be removed

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -48,6 +51,41 @@ class MemoryController extends DisposableController
 
   static const String logFilenamePrefix = 'memory_log_';
 
+  bool showHeatMap = true;
+
+  final List<Snapshot> snapshots = [];
```
kenzieschmoll, line 56: nit: write as `final snapshots = <Snapshot>[];`

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
```diff
@@ -0,0 +1,391 @@
...
+          groupName.startsWith(_flutterLibraryUriPrefix)) {
+        return true;
+      }
+
+      return false;
+    });
+  }
+
+  @override
+  Widget build(BuildContext context) {
```
kenzieschmoll, line 250: can you break this build method up into helpers for readability?

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: Added:

/// Memory heat map blueish color from Blue 100 to Blue 900.
Color memoryHeatMapLightColor = const Color(0xFFBBDEFB);  // Material design BLUE 100
Color memoryHeatMapDarkColor = const Color(0xFF0D47A1);   // Material design BLUE 900

I agree should look at either mimicing or replacing to match flame chart colors and navigation.  For now I need a larger range of colors to see differences in the heat map.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -314,6 +314,17 @@ class MemoryController extends DisposableController
...
+  final _searchNotifier = ValueNotifier<String>('');
+
+  /// Notify that the search has changed.
+  ValueListenable get searchNotifier => _searchNotifier;
+
+  set search(String value) {
+    _searchNotifier.value = value;
+  }
+
+  String get search => _searchNotifier.value;
```
jacob314, line 326: nit: I think a good pattern is to only expose values that should be exposed through a ValueListenable through the value listenable. That way consumers know to listen for the value rather than relying on the current value. So leave the searchNotifier but remove this getter replacing its uses with searchNotifier.value

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
terrylucas: Done

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
terrylucas: Done

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
terrylucas: groupByTreeTable

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
terrylucas: // Dialog has three main vertical sections:
    //      - three checkboxes
    //      - one list of libraries with at least 5 entries
    //      - one row of buttons Ok/Cancel

I think your computation is better plus I adjusted for smaller ranges.

            width: MediaQuery.of(context).size.width / 3,
            height: constraints.maxHeight < 400
                ? constraints.maxHeight
                : constraints.maxHeight / 3 + (400 * .7),


Still think there needs to be UX cleanup but I'll await a UX review.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
terrylucas: It is because this list can be very large (list of known libraries 30, 50, etc.) and I don't want to lose the 3 top checkboxes either.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
terrylucas: This particular print I'd like to keep its for monitoring classes.  I'll add a test check for this output it means the user has forgot to turn off monitoring - a debug only thing.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
terrylucas: Yep, it was a bit tricky (since Checkbox is immutable) - but level of indirection solved the problem.

```
/// Checkbox Widget class using a ValueNotifier.
class NotifierCheckbox extends Checkbox {
  NotifierCheckbox({
    Key key,
    @required this.notifier,
  }) : super(
          key: key,
          value: notifier.value,
          onChanged: notifier.onChanged,
        );

  final CheckboxValueNotifier notifier;

  /// Notifies that the value of the checkbox has changed.
  ValueListenable<bool> get valueListenable => notifier;
}

/// ValueNotifier associated with a Checkbox widget.
class CheckboxValueNotifier extends ValueNotifier<bool> {
  CheckboxValueNotifier(bool value) : super(value);

  void onChanged(bool newValue) {
    value = newValue;
  }
}

```

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
terrylucas: Done

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
terrylucas: HeapGraphElementLive and HeapGraphClassLive

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
terrylucas: Removed as I was processing PR suggestions - gone now.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
terrylucas: This was a debugging function - removed.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -83,22 +84,39 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
...
+        snapshotDisplay = HeatMapSizeAnalyzer(
+          child: SizedBox.expand(
+            child: FlameChart(controller),
+          ),
+        );
+      } else {
+        snapshotDisplay = MemorySnapshotTable();
+      }
+    } else {
+      snapshotDisplay = const Text('');
```
jacob314, line 119: nit: use `const SizedBox()` instead of `const Text('')`

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -263,23 +281,59 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
...
+      ),
+    );
+  }
+
   Widget _buildSearchFilterControls() {
+    searchField = createSearchField();
+
+    final searchAndRawKeyboard = RawKeyboardListener(
+      child: searchField,
+      focusNode: FocusNode(),
```
jacob314: instead of creating a new focus node here, can you create it in a stateful widget so the lifecycle is a bit longer and you can call dispose on the focus node. 
that way selection won't be lost after for example doing a hot reload or otherwise rebuilding this portion of the widget tree.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -263,23 +281,59 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
...
-              key: searchButtonKey,
-              onPressed: _search,
-              child: Label(
-                search,
-                'Search',
-                minIncludeTextWidth: 200,
-              ),
-            ),
+          Container(
+            width: 300.0,
```
jacob314, line 361: Add a TODO to use a more adaptive layout than forcing this to 300.0

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -10,37 +10,160 @@ import 'package:flutter/material.dart' hide TextStyle;
...
+}
+
+class FlameChartState extends State<FlameChart> with AutoDisposeMixin {
+  FlameChartState(this.controller);
+
+  InstructionsSize sizes;
+
+  Map<String, Function> callbacks = {};
+
+  /// Flame chart has a blueish color.
```
jacob314: can we share some colors with our other flame chart?

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -10,37 +10,160 @@ import 'package:flutter/material.dart' hide TextStyle;
...
+  void initState() {
+    super.initState();
+  }
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+
+    controller = Provider.of<MemoryController>(context);
+
```
jacob314, line 76: short circuit return if `controller == Provider.of<MemoryController>(context)`

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -10,37 +10,160 @@ import 'package:flutter/material.dart' hide TextStyle;
...
   @override
   Widget build(BuildContext context) {
-    return _FlameChart(sizes, lightColor, darkColor);
+    sizes = InstructionsSize.fromSnapshop(controller);
+
+    if (sizes != null) {
+      _flameChart = _FlameChart(sizes, lightColor, darkColor, callbacks);
+      return _flameChart;
+    } else {
+      return const Text('');
```
jacob314, line 122: as elsewhere prefer SizedBox to a zero character string for all cases where we are showing empty content where the non-empty content is not a string.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -10,37 +10,160 @@ import 'package:flutter/material.dart' hide TextStyle;
...
+  const _FlameChart(
+      this.sizes, this.lightColor, this.darkColor, this.callbacks);
 
   final InstructionsSize sizes;
+
   final Color lightColor;
+
   final Color darkColor;
 
+  final Map<String, Function> callbacks;
```
jacob314, line 159: it would be more consistent with Flutter style to enumerate each callback with a typed Function rather than having a map of string names to callbacks.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -89,7 +222,10 @@ class FlameChartRenderObject extends RenderBox {
     markNeedsPaint();
   }
 
-  // TODO(terry): implemented not yet used to be used by search.
+  Node externalFindNode(String searchName) {
```
jacob314, line 224: what does the word "external" mean?
Perhaps remove the word external and make methods that should not be used outside this class private.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -106,12 +242,27 @@ class FlameChartRenderObject extends RenderBox {
...
+
+    // TODO(terry): Need to force a relayout to repaint (paintAncestors) with
+    //              new selection. Calling debugResetSize or size changing
+    //              works.  However, shouldn't hitTestSelf do the same e.g.,
+    //
+    //                  hitTestSelf(Offset(
+    //                    _selectedNode.rect.left + paintOffset.dx,
+    //                    _selectedNode.rect.top - paintOffset.dy,
+    //                  ));
+    size = size;
```
jacob314: does calling https://api.flutter.dev/flutter/rendering/RenderObject/markNeedsLayout.html
work? setting the size to itself is a little surprising and might stop having the desired effect if some optimization was added that detected that the size was unchanged.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -120,38 +271,33 @@ class FlameChartRenderObject extends RenderBox {
...
+        context, _selectedNode, 0 + paintOffset.dx, rootWidth, top - offset.dy);
 
     _paintChildren(
       context: context,
-      currentLeft: 1,
+      currentLeft: 1 + offset.dx,
       parentSize: _selectedNode.byteSize,
       children: _selectedNode.children.values,
-      topFactor: top + 51,
+      topFactor: top + 51 - offset.dy,
```
jacob314: there are a couple constants in this file that are about 50. Make a rowHeight or similar constant that is 50 and use it everywhere so that we can tweak the height without having to track down all occurrences of 50, 51, etc.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
```diff
@@ -250,19 +217,25 @@ class SnapshotFilterState extends State<SnapshotFilterDialog>
...
       shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(20.0)),
       child: LayoutBuilder(
         builder: (BuildContext context, BoxConstraints constraints) {
           return Container(
             width: MediaQuery.of(context).size.width / 3,
-            height: constraints.maxHeight < 500
+            height: constraints.maxHeight < 400
                 ? constraints.maxHeight
-                : constraints.maxHeight / 3 + 300,
+                : constraints.maxHeight / 3 + (400 * .7),
```
jacob314, line 264: this needs to be `constraints.maxHeight *.3 + (400 * .7)`, rather than
`constraints.maxHeight / 3 + (400 * .7)` to have a smooth transition.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -423,9 +425,9 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
   }
 
   Future<void> doGroupBy() async {
-    controller.theGraph.computeInstancesForClasses();
-    controller.theGraph.computeRawGroups();
-    controller.theGraph.computeFilteredGroups();
+    controller.heapGraph.computeInstancesForClasses();
```
jacob314, line 428: tiny nit: this would be a nice case for method cascades.

packages/devtools_app/lib/src/ui/flutter/utils.dart
```diff
@@ -0,0 +1,50 @@
...
+        );
+
+  final CheckboxValueNotifier notifier;
+
+  /// Notifies that the value of the checkbox has changed.
+  ValueListenable<bool> get valueListenable => notifier;
+}
+
+/// ValueNotifier associated with a Checkbox widget.
+class CheckboxValueNotifier extends ValueNotifier<bool> {
```
jacob314, line 44: this is interesting but would require that people use CheckboxValueNotifier instead of the standard ValueNotifier class to use NotiferCheckBox. I think it would be worth it to make NotifierCheckbox a little bit more complex and support any ValueNotifier rather than just a CheckboxValueNotifer.
Also, does the NotifierCheckbox currently update if the value of the notifier changes? I think it currently only updates in DevTools because setState happened to be called on a parent widget but I may not be understanding the Checkbox class.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
terrylucas: Done.  Bit of work needed operator == overloading but its much nicer.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: Gone. Create Stateful widget.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: Implemented.

packages/devtools_app/lib/src/memory/flutter/memory_snapshot_models.dart
terrylucas: Only issue was with createClass since class was a reserved name which is why I used createXXX.  But I changed everything and createClass is not aClass,

packages/devtools_app/lib/src/memory/flutter/memory_snapshot_models.dart
terrylucas: Done. Plus a few other places too.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
terrylucas: Why do you think these can be removed they are special for wildcard filtering.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -48,6 +51,41 @@ class MemoryController extends DisposableController
...
+    _selectedSnapshotNotifier.value = snapshotTimestamp;
+  }
+
+  String get selectedSnapshotTimestamp => _selectedSnapshotNotifier.value;
+
+  HeapGraph theGraph;
+
+  /// Leaf node of tabletree snapshot selected?  If selected then the instance
+  /// view is displayed to view the fields of an instance.
+  /// Notifies that the timeline has been paused.
```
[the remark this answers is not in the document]
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -354,33 +450,64 @@ class MemoryController extends DisposableController
...
-        })).then((isolates) {
-          libraryCollection = LibraryCollection(libraryFilters);
-          for (LibraryRef libraryRef in isolates.first.libraries) {
-            serviceManager.service
-                .getObject(_isolateId, libraryRef.id)
-                .then((theLibrary) {
-              libraryCollection.addLibrary(theLibrary);
-            });
-          }
+  static const String libraryRootNode = '___LIBRARY___';
```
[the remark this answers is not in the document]
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
terrylucas: Done - good catch similar to what I did for TextField too.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: Fixed it everywhere.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: Must have been old.  There is just one constant nodeHeight = 50 no 51, etc.

packages/devtools_app/lib/src/memory/flutter/memory_filter.dart
terrylucas: Ok, done.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: That was a bad word should have used exposeFindNode, exposeSelectNode would be better.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: Added a TODO this doesn't work only in the heatmap.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: I've tried all of these:

      markNeedsLayout();
      markParentNeedsLayout();
      markNeedsLayoutForSizedByParentChange();
      markNeedsSemanticsUpdate();
      markNeedsCompositingBitsUpdate();

Nothing but size=size works but not great.  There is a TODO around this need to ask the Flutter folks for suggestions.  Maybe custom painter is a solution?

jacob314: a TODO is fine. Maybe file a bug because this is quite odd. perhaps there is something strange about when we are calling externalSelectNode.

jacob314:
lgtm_dog

packages/devtools_app/test/flutter/memory_screen_test.dart
```diff
@@ -79,7 +79,7 @@ void main() {
 
       expect(controller.memorySource, MemoryController.liveFeed);
 
-      expect(find.byKey(MemoryScreen.snapshotButtonKey), findsOneWidget);
+//      expect(find.byKey(MemoryScreen.snapshotButtonKey), findsOneWidget);
```
jacob314, line 82: why is this commented out?

would be good to add some more tests as a followup. now that you are using a snapshot json file, writing unittests without requiring a vm service will be a lot easier.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: Done.

packages/devtools_app/lib/src/trees.dart
terrylucas: Mimicked to be like expand and collapse.  I would have preferred onSelectLeaf but thought it made more sense to be consistent with rest of tree.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: I don't think so the mainUiColorLight to mainUiColorDark isn't much color variation however, I'll add a TODO and check with the UX folks.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: Done.

PR #1887, opened 2020-04-29
Use a Focus widget to capture input to TreeTable

gmoothart:
Based on code review from #1856. Also improved test coverage and fixed a
few boundary condition bugs caught by better tests.

### DIFF
--- a/packages/devtools_app/lib/src/flutter/table.dart
+++ b/packages/devtools_app/lib/src/flutter/table.dart
@@ -27,7 +27,7 @@ typedef IndexedScrollableWidgetBuilder = Widget Function(
   int index,
 );
 
-typedef TableKeyEventHandler = void Function(RawKeyEvent event,
+typedef TableKeyEventHandler = bool Function(RawKeyEvent event,
     ScrollController scrollController, BoxConstraints constraints);
 
 enum ScrollKind { up, down, parent }
@@ -122,7 +122,6 @@ class FlatTableState<T> extends State<FlatTable<T>>
       sortColumn: widget.sortColumn,
       sortDirection: widget.sortDirection,
       onSortChanged: _sortDataAndUpdate,
-      focusNode: FocusNode(),
     );
   }
 
@@ -224,7 +223,7 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
   int selectedNodeIndex;
   List<double> columnWidths;
   List<bool> rootsExpanded;
-  FocusNode focusNode = FocusNode();
+  FocusNode focusNode;
 
   @override
   void initState() {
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -224,7 +223,7 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
   int selectedNodeIndex;
   List<double> columnWidths;
   List<bool> rootsExpanded;
-  FocusNode focusNode = FocusNode();
+  FocusNode focusNode;
```
jacob314, line 226: not due to this cl but...
nit: make this private with a getter if needed.

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -233,12 +232,13 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
...
         List.generate(dataRoots.length, (index) => dataRoots[index].isExpanded);
     _updateItems();
+    focusNode = FocusNode();
   }
 
   @override
   void didUpdateWidget(TreeTable oldWidget) {
     super.didUpdateWidget(oldWidget);
-    FocusScope.of(context).requestFocus(focusNode);
+    focusNode.requestFocus();
```
jacob314, line 241: sorry I missed this last round but calling focusNode from didUpdateWidget is terrifying. 
Why do we need to request focus from didUpdateWidget?
For example, didUpdateWidget will call with each hot reload and you really don't want to be grabbing focus on hot reload.
@gspencergoog what is the preferred way to indicate that a Widget should be the default focus target?

gspencergoog: Oh, yeah, that's is odd, now that I think about it.

The typical way is to use the `autofocus` attribute of the `Focus` widget.

Alternatively, you can call `autofocus` on the scope that the widget is in:

```dart
focusNode.enclosingScope.autofocus(focusNode);
```

This will give focus to the widget only if nothing else in that scope has focus.

The difference between the two methods is that the `Focus` widget will only call `FocusScope.autofocus()` once for the autofocused item, so that it only takes the focus the first time it is built, and not every time there isn't a focus in the scope.

gmoothart: Neither autofocus in the widget nor `focusNode.enclosingScope.autofocus(focusNode)` in didUpdateWidget worked; the TreeTable didn't get focus. And I don't know how to visualize scopes to understand why.

I could wrap the component in a GestureDetector and give focus on tap; I'm sure that would work.

gspencergoog: If something else in the scope is getting focus, then `autofocus` will be ignored. It could also be that more than one things in the scope has `autofocus` set.

If you want to know more about the focus state, you can turn on the focus system debugging.  Change line 20 in `focus_manager.dart` to be:

```dart
const bool _kDebugFocus = true;
```

and the focus system will output a lot of debug output about what it's doing, including a focus tree every time the focus changes.

gmoothart: from the debug information it looks like something in the child ListView is getting focus. GestureDetector is working reliably so I'm going to go with that for now.

PR #1888, opened 2020-04-29
add a key/value store to the devtools server

devoncarew:
- add a key/value store to the devtools server
- change the light/dark theme value to be persisted via the server connection

This adds the notion of a `Storage` class, and when we have a connection to the devtools server, backs that storage class w/ a key/value store implemented by the server.

### DIFF
--- a/packages/devtools_app/lib/src/config_specific/flutter/framework_initialize/_framework_initialize_desktop.dart
+++ b/packages/devtools_app/lib/src/config_specific/flutter/framework_initialize/_framework_initialize_desktop.dart
@@ -2,7 +2,14 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:convert';
+import 'dart:io';
+
 import 'package:flutter/foundation.dart';
+import 'package:path/path.dart' as path;
+
+import '../../../globals.dart';
+import '../../../storage.dart';
 
 /// Return the url the application is launched from.
 Future<String> initializePlatform() async {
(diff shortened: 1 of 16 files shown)


packages/devtools_app/lib/src/flutter/preferences.dart
```diff
@@ -22,26 +24,20 @@ class PreferencesController {
...
-    //    setTheme(darkTheme: darkModeTheme.value);
-    //    storage.setValue('ui.darkMode', '${darkModeTheme.value}');
-    //  });
-    //});
-
-    _darkModeTheme.addListener(() {
-      // ignore: deprecated_member_use_from_same_package
-      devtools_theme.setDarkTheme(_darkModeTheme.value);
+    storage.getValue('ui.darkMode').then((String value) {
+      _darkModeTheme.value = value == null || value == 'true';
```
DanTup, line 35: Is this intended to default to Dark mode if there's no stored value? (Other code, like in framework_core appears to default to light it `theme` isn't on the querystring).

packages/devtools_server/lib/src/client_manager.dart
```diff
@@ -90,6 +92,17 @@ class DevToolsClient {
...
           _respond(request);
           return;
+        case 'getPreferenceValue':
+          final String key = _currentPage = request['params']['key'];
+          final dynamic value = ServerApi.devToolsPreferences.properties[key];
+          _respondWithResult(request, value);
+          return;
+        case 'setPreferenceValue':
+          final String key = _currentPage = request['params']['key'];
+          final dynamic value = _currentPage = request['params']['value'];
```
DanTup, line 102: Are these `_currentPage =` supposed to be here? It looks like this will overwrite the `_currentPage` variable with the keys/values being stored?

packages/devtools_app/lib/src/config_specific/flutter/framework_initialize/_framework_initialize_desktop.dart
```diff
@@ -2,16 +2,71 @@
...
+  @override
+  Future setValue(String key, String value) async {
+    try {
+      _values = _readValues();
+      _values[key] = value;
+
+      const encoder = JsonEncoder.withIndent('  ');
+      _preferencesFile.writeAsStringSync('${encoder.convert(_values)}\n');
+    } catch (_) {
+      // ignore
```
DanTup, line 46: Is there anywhere we could usefully log a note if this fails, to help track it down if it doesn't seem to be working?

packages/devtools_app/lib/src/flutter/preferences.dart
jacob314: we might as well also change the default to the dark theme everywhere. Initially the default was the light theme in DevTools because that is what we worked on first and the light theme was a bit more polished in the dart:html version of the app. the dark theme in the flutter version of the app is about on par with the light theme.

packages/devtools_app/lib/src/config_specific/flutter/framework_initialize/_framework_initialize_desktop.dart
```diff
@@ -2,16 +2,71 @@
...
+
+  Map<String, dynamic> _readValues() {
+    try {
+      final File file = _preferencesFile;
+      if (file.existsSync()) {
+        return jsonDecode(file.readAsStringSync());
+      }
+      return {};
+    } catch (_) {
+      // ignore
```
jacob314, line 58: nit: if a user modifies the preference file manually they could get invalid json. Maybe report an error here with the exception thrown decoding the json?

packages/devtools_app/lib/src/config_specific/flutter/framework_initialize/_framework_initialize_desktop.dart
devoncarew: Yeah, I'll convert this to a `log()` call.

packages/devtools_app/lib/src/config_specific/flutter/framework_initialize/_framework_initialize_desktop.dart
devoncarew: sound good!

I changed this to pass the error through when both reading the values and writing the values

packages/devtools_app/lib/src/flutter/preferences.dart
devoncarew: yup, the code above will default to dark mode (sounds like this is reasonable at this point)

packages/devtools_server/lib/src/client_manager.dart
devoncarew: thanks for the eyes on this; this wasn't correct (and I'm not even sure how I wrote this; copy pasta? a merge issue?)

PR #1891, opened 2020-04-30
Support Uris on querystring

DanTup:
This adds support for ?uri= on the querystring to allow hard-refreshing and direct launching from VS Code.

The original issue is that code handling `/` would loading the `Initializer` which would push the `/connect` route if it was not able to connect. However, loading `/?uri=foo` would trigger two routes, `/` and `/?uri=foo` which resulted in us both pushing `/connect` and trying to handle the connection.

This removes the `/connect` route and instead conditional handles a single home route based on whether it has `?uri=`. The route `/` will just load the connection form, and then `/?uri=` will load the `Initializer`. If the `Initializer` decides to navigate back to the connection screen, it'll do that by popping itself off rather than pushing a new route.

This is my first foray into Routing, so don't assume I know what I'm doing - if anything looks weird or not best practice, do point it out 😄

### DIFF
--- a/packages/devtools_app/lib/src/flutter/initializer.dart
+++ b/packages/devtools_app/lib/src/flutter/initializer.dart
@@ -68,11 +68,6 @@ class _InitializerState extends State<Initializer>
     // If we become disconnected, attempt to reconnect.
     autoDispose(
       serviceManager.onStateChange.where((connected) => !connected).listen((_) {
-        // Generally, empty setState calls in Flutter should be avoided.
-        // However, serviceManager is an implicit part of this state.
-        // This setState call is alerting a change in the serviceManager's
-        // state.
-        setState(() {});
         // TODO(https://github.com/flutter/devtools/issues/1285): On losing
         // the connection, only provide an option to reconnect; don't
         // immediately go to the connection page.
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/flutter/initializer.dart
```diff
@@ -67,51 +65,51 @@ class _InitializerState extends State<Initializer>
...
+
     final uri = normalizeVmServiceUri(widget.url);
     final connected = await FrameworkCore.initVmService(
       '',
       explicitUri: uri,
       errorReporter: (message, error) =>
           Notifications.of(context).push('$message, $error'),
     );
+
     if (!connected) {
```
jacob314, line 101: not your bug but for some defensive programming, maybe wrape the initVmService call with a 
try {
} finally {
  if (!connected) ..
}

DanTup: There's already error handling inside `initVmService` which reports errors as notifications and then returns false. I can add it here, but I think it'd only be hit if the other error handling was removed (which would then effect the connect screen and others). Is it still worth adding here?

PR #1895, opened 2020-04-30
clean up around the use of the theme setting

devoncarew:
Clean up around the use of the theme setting:
- just use the theme setting to control the light/dark theme (no longer use the url)
- in some places, query the `theme.brightness` theme field to know whether we're in light or dark theme (instead of querying the global var)
- convert the `titleSolidBackgroundColor` global final field to a function; we'd had issues w/ some colors not updating on theme changes, I think possibly due to the framework making copies of colors? i.e., not expecting colors to be not be immutable (a `ThemedColor` can return different values when called two separate times)

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/codeview.dart
@@ -12,7 +12,6 @@ import '../../flutter/auto_dispose_mixin.dart';
 import '../../flutter/common_widgets.dart';
 import '../../flutter/flutter_widgets/linked_scroll_controller.dart';
 import '../../flutter/theme.dart';
-import '../../ui/theme.dart';
 import '../../utils.dart';
 import 'breakpoints.dart';
 import 'common.dart';
@@ -157,7 +156,6 @@ class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
 
     if (!verticalController.hasAttachedControllers) {
       // TODO(devoncarew): I'm uncertain why this occurs.
-      // todo: ???
       log('LinkedScrollControllerGroup has no attached controllers');
       return;
     }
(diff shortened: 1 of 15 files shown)


jacob314:
lgtm
After we remove the dart:html app we will want to get rid of ThemedColor due to the issue you mentioned.

packages/devtools_app/lib/src/debugger/flutter/codeview.dart
```diff
@@ -349,9 +347,10 @@ class GutterItem extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
+    final darkTheme = theme.brightness == Brightness.dark;
```
kenzieschmoll, line 350: should we make a utilityMethod `isDarkTheme(BuildContext context)` to avoid confusion about how to check for dark theme?

devoncarew: sounds good! I'll look at adding an extension method on `ThemeData`, so we can do something like `Theme.of(context).isDarkTheme`.

devoncarew: Added as `ThemeDataExtension`.

DanTup:
> just use the theme setting to control the light/dark theme (no longer use the url)

For the embedded view in VS Code, I was planning to pass `theme=dark` and `theme=light` to get the theme that matches the editor. Is there a way I can set the `brightness` that's used here, or would we need to add this back as a way to override it?

DanTup:
(I'm also passing through `&background=#f00f00` which would be good to use if it's close enough match for the theme because currently there's a slight mismatch - and if we hard-code it to VS Code's default dark background it probably won't match IntelliJ/Android Studio)

Issue #1889, opened 2020-04-30
Unable to connect to a second VM service after the first one disconnects


DanTup:
Using the Flutter version (though that might not be important), if I terminate the app I'm connected to, DevTools goes back to the Connect page. However, trying to use the Connect page to connect to another app fails with `Future already completed`. This looks like it comes from `serviceAvailable.complete();` in `service_manager`.

It looks like there's only ever a single instance of `ServiceConnectionManager` so probably upon disconnect there's some state that needs cleaning up (and the `Complete` reset to a new one).

Issue #1897, opened 2020-05-01
app lifecycle improvements


devoncarew:
A meta-issue for app lifecycle improvements:
- [x] we should be able to connect to an app, disconnect, and re-connect to a new app in the same devtools session https://github.com/flutter/devtools/issues/1889
- [x] after disconnecting from an app, it would be nice if the data from that session was still available to review; for this, each screen may need to be aware of and handle being in a disconnected state https://github.com/flutter/devtools/issues/1285
- [x] when not connected, it would be nice if some casual information about devtools (the devtools version and where to send feedback) was visible (i.e., make the settings and info buttons active even from the connect screen)
- [ ] diagnose why we see multiple requests to connect to the same connection https://github.com/flutter/devtools/issues/1568
- [x] have a way to disconnect to the existing connection https://github.com/flutter/devtools/issues/1900

elliette:
Closing this out since all tasks have been completed except for the one tracked by https://github.com/flutter/devtools/issues/1568

PR #1898, opened 2020-05-01
allow DevTools to re-connect to a new app after closing a previous app

devoncarew:
- allow DevTools to re-connect to a new app after closing a previous app
- fix https://github.com/flutter/devtools/issues/1889 (related to https://github.com/flutter/devtools/issues/1897)

This does some surgery on the `ServiceConnectionManager` class to make one Completer private, and ensure that we reset some minimal state between connections.

### DIFF
--- a/packages/devtools_app/lib/src/service_manager.dart
+++ b/packages/devtools_app/lib/src/service_manager.dart
@@ -47,7 +47,7 @@ class ServiceConnectionManager {
 
   Completer<VmService> _serviceAvailable = Completer();
 
-  Future get onServiceAvailable => _serviceAvailable.future;
+  Future<VmService> get onServiceAvailable => _serviceAvailable.future;
 
   bool get isServiceAvailable => _serviceAvailable.isCompleted;
 
--- a/packages/devtools_app/lib/src/service_manager.dart
+++ b/packages/devtools_app/lib/src/service_manager.dart
@@ -45,14 +45,18 @@ class ServiceConnectionManager {
   final StreamController<VmServiceWrapper> _connectionAvailableController =
       StreamController<VmServiceWrapper>.broadcast();
 
-  final serviceAvailable = Completer<void>();
+  Completer<VmService> _serviceAvailable = Completer();
+
+  Future get onServiceAvailable => _serviceAvailable.future;
+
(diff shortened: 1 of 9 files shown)


packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -45,14 +45,18 @@ class ServiceConnectionManager {
   final StreamController<VmServiceWrapper> _connectionAvailableController =
       StreamController<VmServiceWrapper>.broadcast();
 
-  final serviceAvailable = Completer<void>();
+  Completer<VmService> _serviceAvailable = Completer();
+
+  Future get onServiceAvailable => _serviceAvailable.future;
```
jacob314, line 50: missing type.
`Future` --> `Future<VmService>`

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -138,7 +142,7 @@ class ServiceConnectionManager {
       serviceTrafficLogger = VmServiceTrafficLogger(service);
     }
 
-    serviceAvailable.complete();
+    _serviceAvailable.complete(service);
```
jacob314, line 145: we might want to also add
```dart
if (_serviceAvaillabe.isCompleted) {_serviceAvailable = Completer();}
```
at the start of this method to further support opening a new service when there was an existing service. Maybe just add a TODO because we would really want to further guard things so only one request to switch to a new vm service could be triggered at once.

packages/devtools_app/lib/src/service_manager.dart
devoncarew: updated

packages/devtools_app/lib/src/service_manager.dart
devoncarew: If we're actively connected to a we likely want to guard against opening a new connection - I expect we'll first want to disconnect from the existing connection?

I'll add a TODO:, and open an issue about having a way to disconnect from the current connection in the UI.

devoncarew: Added the multiple connects issue to https://github.com/flutter/devtools/issues/1897, and filed https://github.com/flutter/devtools/issues/1900 about disconnecting from the current connection.

PR #1899, opened 2020-05-01
Add Ctrl+P shortcut to filter libraries

### LINKED ISSUE #1294: Debugger ctrl+p searching from the library view
We need to make sure the library panel is capturing the ctrl+f event for library search.

gmoothart:
In the spirit of the Chrome devtools "sources" panel, we'll use Ctrl+P to
search libraries and classes in the debugger view. Fixes #1294

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
@@ -250,23 +250,22 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
   }
 }
 
-class FilterLibraryIntent extends RequestFocusIntent {
+class FilterLibraryIntent extends Intent {
   const FilterLibraryIntent(
-    FocusNode focusNode,
+    this.focusNode,
     this.debuggerController,
   )   : assert(debuggerController != null),
-        super(focusNode);
+        assert(focusNode != null);
 
+  final FocusNode focusNode;
   final DebuggerController debuggerController;
 }
 
-class FilterLibraryAction extends RequestFocusAction {
+class FilterLibraryAction extends Action<FilterLibraryIntent> {
   @override
   void invoke(FilterLibraryIntent intent) {
-    debugPrint('invoking filter action');
-    super.invoke(intent);
-    // TODO: open, don't toggle
-    intent.debuggerController.toggleLibrariesVisible();
+    intent.debuggerController.openLibrariesView();
+    intent.focusNode.requestFocus();
   }
 }
 
--- a/packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
@@ -5,6 +5,7 @@
 import 'dart:async';
 
 import 'package:flutter/material.dart' hide Stack;
+import 'package:flutter/services.dart';
 import 'package:provider/provider.dart';
 import 'package:vm_service/vm_service.dart';
 
@@ -69,6 +70,13 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
   static const breakpointsTitle = 'Breakpoints';
 
   DebuggerController controller;
+  FocusNode _libraryFilterFocusNode;
+
+  @override
+  void initState() {
+    super.initState();
+    _libraryFilterFocusNode = FocusNode();
+  }
 
   @override
   void didChangeDependencies() {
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
```diff
@@ -147,28 +162,42 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
...
-        Column(
-          children: [
-            DebuggingControls(controller: controller),
-            const SizedBox(height: denseRowSpacing),
-            Expanded(
-              child: Split(
-                axis: Axis.vertical,
-                initialFractions: const [0.74, 0.26],
+    return Shortcuts(
+      shortcuts: <LogicalKeySet, Intent>{
```
gmoothart, line 166: These shortcuts seem to be ignored unless I manually focus something in the debugger pane; the only candidate I've found that works is the filter textbox itself which kind of defeats the purpose.

jacob314: Odd. If as a hack you place the shortcuts widget at the very root of the application do you get the expected behavior? The problem might be that the default focus is on the tab bar or some other portion of the UI outside the debugger screen.

gmoothart: Yes, it does work if I promote the shortcuts widget to the root of the application. I think that you are probably right about the focus being outside the debugger screen

jacob314: lets enable devtools screens to attach keyboard handlers placed at the root of the widget tree. seems reasonable for screens to always receive keyboard events when their screen is open.

gmoothart: I'll play around with that.

I currently have a theory that the way tab bodies are rendered by the devtools doesn't play nice with autofocus. It would explain a few things. I did verify that the primary focus continues to be the InspectorTree from the last tab when I want it to be in the debugger, and autofocus properties don't change that.

gmoothart: Got it! The secret is that every screen needs its own FocusScope, then things like autofocus work correctly. Thanks to @gspencergoog for helping me solve it.

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
```diff
@@ -147,28 +162,42 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
...
+        LogicalKeySet(LogicalKeyboardKey.control, LogicalKeyboardKey.keyP):
+            FilterLibraryIntent(_libraryFilterFocusNode, controller),
+      },
+      child: Actions(
+        actions: <Type, Action<Intent>>{
+          FilterLibraryIntent: FilterLibraryAction(),
+        },
+        child: Focus(
+          autofocus: true,
```
gmoothart, line 175: focusing the whole debugger page is a bit arbitrary. Alternatively I could autofocus, say, the "pause" button which is first on the screen? 
Screen Shot 2020-05-05 at 12 46 58 PM

devoncarew: That seems reasonable - nothing else on that screen jumps out as a good first focus candidate.

packages/devtools_app/lib/src/debugger/flutter/scripts.dart
```diff
@@ -24,12 +24,14 @@ class ScriptPicker extends StatefulWidget {
     @required this.scripts,
     @required this.classes,
     @required this.onSelected,
+    this.libraryFilterFocusNode,
```
kenzieschmoll, line 27: should this be required?

gmoothart: I don't think so, because if it isn't then the text field will just provide its own. The reason to provide it is if you want to control focus from the outside, as we do with the keyboard shortcut.

kenzieschmoll: right, but we aren't using the `ScriptPicker` widget anywhere but in the debugger, where we want to make sure we are always recognizing the ctrl + p shortcut. So making this parameter required would guarantee that we don't regress and stop including the focus node.

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
gmoothart: This seemed straightforward but I ran into more autofocus issues so I'm going to skip it for now.

packages/devtools_app/lib/src/debugger/flutter/scripts.dart
gmoothart: Added focus to the filter text field, even when you open it with the mouse.

PR #1902, opened 2020-05-04
Initial support for running integration tests using Flutter web version

DanTup:
This makes it possible to run integration tests using `flutter` instead of `pub build_runner` (spoiler: they don't pass) by setting an env variable `DEVTOOLS_FLUTTER_WEB`. This will cause the tests to build using Flutter, and the server to build the URI in a format for launching the Flutter version. Some of this conditional code can be dropped with the dart:html version.

Right now, most of these tests fail, because they rely on sending messages to the app using `HtmlApp` which is only set up in the HTML version. @devoncarew not sure what the plan is here - should that be fixed up to work with the Flutter version, or should the tests work differently?

The tests in `server_test.dart` don't need that model though, so almost pass (working on another PR for that).

It won't be possible to enable these tests on Travis until we fix them (most of the time out, so even having them as allowed failures may significantly increase the run duration), but this makes it easier to work on them locally:

```
WEBDEV_RELEASE=true TEST_FLUTTER_WEB=true flutter test -j1 test/integration_tests
```

### DIFF
--- a/packages/devtools_app/test/integration_tests/integration.dart
+++ b/packages/devtools_app/test/integration_tests/integration.dart
@@ -17,9 +17,6 @@ import '../support/cli_test_driver.dart';
 import '../support/utils.dart';
 
 const bool verboseTesting = false;
-final bool testInReleaseMode = Platform.environment['WEBDEV_RELEASE'] == 'true';
-final bool testFlutterWebVersion =
-    Platform.environment['DEVTOOLS_FLUTTER_WEB'] == 'true';
 
 WebdevFixture webdevFixture;
 BrowserManager browserManager;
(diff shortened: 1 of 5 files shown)


devoncarew:
> Right now, most of these tests fail, because they rely on sending messages to the app using HtmlApp which is only set up in the HTML version. @devoncarew not sure what the plan is here - should that be fixed up to work with the Flutter version, or should the tests work differently?

I suspect that we won't port that support over. We may want instead to use `flutter driver` tests to integration test the full app? cc @jacob314 for his thoughts.

jacob314:
We won't port these tests over and will instead use a combination of flutter widget tests and `flutter driver` tests for the cases the integration tests would cover.

DanTup:
Ok, I've reverted the changes related to the non-server integration tests. What remains allows running the server tests against the Flutter version (eg. ensuring the server can launch, launch on a specific page, change pages, etc.). I have those tests passing for an upcoming PR (which relies on a few of the open ones I have).

PR #1903, opened 2020-05-04
Add support for selecting an initial page using the querystring

DanTup:
This adds support for `page=foo` on the query string to load as the initial page (and broadcasts the initial page to the server).

However, it's slightly weird that it stays in the URL even when you move to other pages. I can fix that by changing the `addLocalHistoryEntry` code to instead push real routes (this would mean the `page=` in the querystring updates as you move around - and we could probably move that to be part of the route so it'll become `/inspector?uri=xxx` -> `/memory?uri=xxx`) - however, that seems to mess with the animations when you move between tabs because it's now loading whole new routes.

I can't find any way to update the browsers address bar (like pushing a named route) without us getting full page transitions. I think we need something like [`pushState`](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState) that lets us manipulate the URL without actually changing anything else. Named routes trigger Flutter to do work we don't want, and `addLocalHistoryEntry` doesn't allow us to update the URL (maybe it should, and that would be the nicest fix?).

Ideas welcome!

### DIFF
--- a/packages/devtools_app/lib/src/flutter/app.dart
+++ b/packages/devtools_app/lib/src/flutter/app.dart
@@ -127,6 +127,7 @@ class DevToolsAppState extends State<DevToolsApp> {
             url: params['uri'],
             builder: (_) => _providedControllers(
               child: DevToolsScaffold(
+                initialPage: params['page'],
                 tabs: _visibleScreens(),
                 actions: [
                   HotReloadButton(),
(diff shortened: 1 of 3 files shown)


DanTup:
If I understood correctly, done! There's also a server test that ensures the server launches on a specific page, though those tests aren't enabled yet (can be enabled after #1902).

PR #1906, opened 2020-05-04
show the device as busy during reload calls

devoncarew:
Show the device as busy during reload calls:
- add an `deviceBusy` ValueNotifier in the service connection manager
- when the device is busy, display a small indefinite progress widget near the device name
- have the reload and restart operations mark the device as busy

This gives the user more of an indication when longer term work is happening.

If we want to generalize this past this one use case, we could introduce the notion of a 'Jobs' abstraction. This would be longer lived tasks, with names and a manager class, that we could use to surface info about the jobs to the user.

- https://www.eclipse.org/articles/Article-Concurrency/jobs-api.html
- https://github.com/googlearchive/chromedeveditor/blob/master/ide/web/lib/jobs.dart

### DIFF
--- a/packages/devtools_app/lib/src/flutter/status_line.dart
+++ b/packages/devtools_app/lib/src/flutter/status_line.dart
@@ -183,9 +183,27 @@ class StatusLine extends StatelessWidget {
           // TODO(devoncarew): Add an interactive dialog to the device status
           // line.
 
+          final color = Theme.of(context).textTheme.bodyText2.color;
+
           return Row(
             mainAxisAlignment: MainAxisAlignment.end,
             children: [
+              ValueListenableBuilder<bool>(
+                valueListenable: serviceManager.deviceBusy,
+                builder: (context, isBusy, _) {
+                  return SizedBox(
+                    width: smallProgressSize,
+                    height: smallProgressSize,
+                    child: isBusy
+                        ? CircularProgressIndicator(
+                            strokeWidth: 2,
+                            valueColor: AlwaysStoppedAnimation<Color>(color),
+                          )
+                        : const SizedBox(),
+                  );
+                },
+              ),
+              const SizedBox(width: denseSpacing),
               Text(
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/flutter/status_line.dart
```diff
@@ -183,9 +183,27 @@ class StatusLine extends StatelessWidget {
           // TODO(devoncarew): Add an interactive dialog to the device status
           // line.
 
+          final color = Theme.of(context).textTheme.bodyText2.color;
+
           return Row(
             mainAxisAlignment: MainAxisAlignment.end,
             children: [
+              ValueListenableBuilder<bool>(
```
jacob314, line 191: nit: `<bool>` isn't required.

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -97,6 +97,18 @@ class ServiceConnectionManager {
...
 
+  final ValueNotifier<bool> _deviceBusy = ValueNotifier<bool>(false);
+
+  /// Whether the device is currently busy - performing a long-lived, blocking
+  /// operation.
+  ValueListenable<bool> get deviceBusy => _deviceBusy;
+
+  /// Set whether the device is currently busy - performing a long-lived,
+  /// blocking operation.
+  void setDeviceBusy(bool isBusy) {
```
jacob314, line 108: nit: avoid starting methods with the word `set`

jacob314:
lgtm

packages/devtools_app/lib/src/flutter/status_line.dart
devoncarew: removed

packages/devtools_app/lib/src/service_manager.dart
devoncarew: Sounds great! Will do.

PR #1913, opened 2020-05-05
Move flutter file under flutter/ directory

kenzieschmoll:
all flutter files need to be under a flutter/ dir


terrylucas:
Thanks Kenzie, this is a good catch. I committed the pub problem caused by this - unfortunately it was not.  Ultimately the pub with stable 2.7.2 worked but the latest pub 2.9.0-5.0.dev had the problem.

PR #1920, opened 2020-05-06
various tweaks to the memory page

devoncarew:
Various tweaks to the memory page:
- hook up the pause and resume buttons for the live memory feed (and, use a ValueNotifier in the memory controller)
- additional padding between some of the buttons
- reduce some vertical spacing in the memory screen to give more room to the charts
- have the button labels show up even for more narrow page layouts

@terrylucas 

Screen Shot 2020-05-06 at 8 02 31 AM

### DIFF
--- a/packages/devtools_app/lib/src/memory/flutter/memory_protocol.dart
+++ b/packages/devtools_app/lib/src/memory/flutter/memory_protocol.dart
@@ -17,7 +17,7 @@ import 'memory_controller.dart';
 class MemoryTracker {
   MemoryTracker(this.service, this.memoryController);
 
-  static const Duration _updateDelay = Duration(milliseconds: 1000);
+  static const Duration _updateDelay = Duration(milliseconds: 500);
 
   VmServiceWrapper service;
 
--- a/packages/devtools_app/lib/src/memory/flutter/memory_protocol.dart
+++ b/packages/devtools_app/lib/src/memory/flutter/memory_protocol.dart
@@ -8,6 +8,7 @@ import 'dart:math' as math;
 import 'package:devtools_shared/devtools_shared.dart';
 import 'package:vm_service/vm_service.dart';
 
+import '../../config_specific/logger/logger.dart';
 import '../../globals.dart';
 import '../../version.dart';
 import '../../vm_service_wrapper.dart';
@@ -16,7 +17,7 @@ import 'memory_controller.dart';
 class MemoryTracker {
   MemoryTracker(this.service, this.memoryController);
 
-  static const Duration updateDelay = Duration(milliseconds: 200);
+  static const Duration _updateDelay = Duration(milliseconds: 1000);
 
   VmServiceWrapper service;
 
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/memory/flutter/memory_screen.dart
```diff
@@ -255,23 +253,34 @@ class MemoryBodyState extends State<MemoryBody> with AutoDisposeMixin {
...
-            key: MemoryScreen.resumeButtonKey,
-            onPressed: controller.paused ? _resumeLiveTimeline : null,
-            child: Label(
-              playIcon,
-              'Resume',
-              minIncludeTextWidth: _primaryControlsMinVerboseWidth,
-            ),
+          const SizedBox(width: denseSpacing),
+          ValueListenableBuilder(
+            valueListenable: controller.paused,
```
kenzieschmoll, line 272: why not wrap the whole row in a the ValueListenableBuilder instead of wrapping both buttons in the same one?

devoncarew: done -

devoncarew:
@terrylucas - ptal, I did make some UI changes to this page; happy to discuss.

packages/devtools_app/lib/src/memory/flutter/memory_protocol.dart
```diff
@@ -16,7 +17,7 @@ import 'memory_controller.dart';
 class MemoryTracker {
   MemoryTracker(this.service, this.memoryController);
 
-  static const Duration updateDelay = Duration(milliseconds: 200);
+  static const Duration _updateDelay = Duration(milliseconds: 1000);
```
terrylucas, line 20: 200 ms is too often but I worry that 1 second is too little.  VM seems to report changes about every 400-500 ms would that be a better delay? ADB show changes every 1/2 a second or so.  With any polling we could lose the spike 1 second might be too much.

devoncarew: I'll update to 500ms.

devoncarew:
> 'Source: Live Feed' drop down looks much better.
> However, 'Display: 1 minute' is almost unreadable (font foreground color too close to background color).

Totally agree; I don't think that change was a result of this PR, but definitely something we should address.

> Also, when the dropdown is selected the popup menu of possible values are in a red foreground font color. Is this a bit jarring?

Ditto re: the UI and updating it.

> In general, these are all much better changes. Only one real worry on the delay interval see single line comment.

Yeah, not problem. I updated it to 500ms, and we can change to whatever we want.

When a newer API exists to query the VM embedder, we may have fewer concerns about frequent polling - I assume that API would be efficient to implement.

> One other thing, DevTools in Flutter Desktop always looks good (everything is visible and clear). Flutter Web is the problem. Is this a general thing we could fix so that both platforms match better? Or is it more endemic to theming right now in Flutter Web?

I don't know the reason for the difference; we should investigate (and definitely err towards favoring flutter web, as that's what our users see).

Issue #1910, opened 2020-05-05
The method '+' was called on null from _CpuProfileFlameChartState.leftForStackFrame 


devoncarew:
```
The following NoSuchMethodError was thrown building ValueListenableBuilder<CpuStackFrame>(state: _ValueListenableBuilderState<CpuStackFrame>#38a45):
The method '+' was called on null.
Receiver: null
Tried calling: +(0.9649776453055141)

The relevant error-causing widget was: 
  ValueListenableBuilder<CpuStackFrame> .../projects/flutter/devtools/packages/devtools_app/lib/src/profiler/flutter/cpu_profiler.dart:132:16
When the exception was thrown, this was the stack: 
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:53:5)
#1      _CpuProfileFlameChartState.leftForStackFrame (package:devtools_app/src/profiler/flutter/cpu_profile_flame_chart.dart:93:45)
#2      _CpuProfileFlameChartState.initFlameChartElements.createChartNodes (package:devtools_app/src/profiler/flutter/cpu_profile_flame_chart.dart:53:20)
#3      _CpuProfileFlameChartState.initFlameChartElements.createChartNodes (package:devtools_app/src/profiler/flutter/cpu_profile_flame_chart.dart:70:25)
#4      _CpuProfileFlameChartState.initFlameChartElements (package:devtools_app/src/profiler/flutter/cpu_profile_flame_chart.dart:74:21)
...
```

cc @kenzieschmoll

PR #1923, opened 2020-05-06
Fix CPU profiler bug where we were unintentionally mutating data.

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/1910.

We need to make a deep copy so that, when we change sort order in the Call Tree, we do not affect the data the flame chart is using to do layout.


kenzieschmoll:
Landing this on red - unrelated test failure

Issue #1756, opened 2020-03-20
App bar action button padding needs to be adjusted when hot restart / reload are not available.


kenzieschmoll:
In this screenshot, I am connected to a Dart VM app. The padding should be adjusted since hot restart and hot reload are unavailable

This applies for Profile builds of flutter apps also.

![Screen Shot 2020-03-20 at 10 27 28 AM](https://user-images.githubusercontent.com/43759233/77189967-a5a5cf80-6a95-11ea-9fb0-cb34b85acade.png)

haliliceylan:
Hi @kenzieschmoll  how can i reproduce this error ?

PR #1930, opened 2020-05-06
Only show hot restart and hot reload buttons for Flutter apps.

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/1756


kenzieschmoll:
Exactly

jacob314:
Instead of doing this we should continue to show and hide the buttons based on whether the app supports hot reload. For example, server side apps can also support hot reload.

PR #1933, opened 2020-05-07
Keep same page when switching to Fluter version

DanTup:
This tweaks the beaker icon to land you on the equivalent page on the new version.


DanTup:
Bot failures were unrelated to this change:

- https://github.com/dart-lang/package_config/issues/8
- Flake connecting to Chrome DevTools that I think we see periodically

PR #1939, opened 2020-05-07
Make TreeTable column headers not focusable

gmoothart:
This fixes a usability issue with treetable arrow-key navigation. We
don't want focus going to the column headers and scrolling vertically.

Before:
focus_before

After:
focus_after

### DIFF
--- a/packages/devtools_app/lib/src/flutter/table.dart
+++ b/packages/devtools_app/lib/src/flutter/table.dart
@@ -840,6 +840,7 @@ class _TableRowState<T> extends State<TableRow<T>>
         color: widget.backgroundColor ?? Theme.of(context).canvasColor,
         child: widget.onPressed != null
             ? InkWell(
+                canRequestFocus: false,
                 key: contentKey,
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -840,6 +840,7 @@ class _TableRowState<T> extends State<TableRow<T>>
         color: widget.backgroundColor ?? Theme.of(context).canvasColor,
         child: widget.onPressed != null
             ? InkWell(
+                canRequestFocus: false,
```
kenzieschmoll, line 843: does this prevent the row from being able to be selected?

gmoothart: Oh, good question. You can still click on the headers to sort, and they have a slight highlight on mouseover. From a mouse-interaction perspective the behavior seems identical. But with my change you can't use tab or arrow keys to move the focus to the individual headers. It goes straight from the buttons like "Collapse All" to the table body.

PR #1940, opened 2020-05-07
Protect against NPE caused by null `devtoolsServer`

kenzieschmoll:
@DanTup we were seeing exceptions in g3 because `devToolsServer` was null ("devtools server not available (204)" printed to console in this case).

This will fix the NPE but do we expect `devToolsServer to be null?


DanTup:
I think previously the `catch` block above would `return` and prevent all this code running. There are some changes in #1886 that allow `connect()` to log its own errors and `return null` though which break this. Change lgtm, though it might be cleaner to check for null further up and just `return` since it's likely none of this method is ever expected to run if there's no connection (this would prevent it coming back if there are further changes in this method that don't consider nulls).

PR #1943, opened 2020-05-08
Various UI tweaks based on UX review

devoncarew:
Various UI tweaks based on UX review:
- wrap `DropdownButton` widgets with `DropdownButtonHideUnderline`
- replace `Label` widgets with `MaterialIconLabel` widgets, which work better in both themes
- use the same live chart height and padding as the timeline page
- some padding and spacing tweaks
- don't use a Split widget to contain the live memory chart - just have a fixed size
- move the 'android memory' button to the main memory page toolbar
- remove the slider control from the memory page (I think having both this and the 1 min / 5 min / ... pulldown was confusing)
- when the android memory chart is visible, display it side-by-side with the dart heap chart (instead of below the dart chart)

@terrylucas - most of these changes are to the memory page; happy to discuss these on the PR or in person


devoncarew:
Screen Shot 2020-05-08 at 7 25 26 PM

Screen Shot 2020-05-08 at 7 25 38 PM

kenzieschmoll:
wdyt about making "Show Android Memory" a switch instead of a button? And can we hide the toggle if the connected app is not an android app?

devoncarew:
> wdyt about making "Show Android Memory" a switch instead of a button? And can we hide the toggle if the connected app is not an android app?

Hiding it if it's not connected to an android app sounds good. I don't know if I have an opinion about using a switch as opposed to a toggle button (though to be clear, the above button it's currently a toggle button).

I think longer term, we want the 'android memory' button and chart to go away, the chart to show all process memory used for any embedder (not just android), and to track coarse process usage (dart heap, skia memory, java/ObjC heap, ...).

I'll hide the button on non-android platforms for now.

terrylucas:
> wdyt about making "Show Android Memory" a switch instead of a button? And can we hide the toggle if the connected app is not an android app?

Good idea.  I suspect it will change from Android Memory to iOS Memory but might be for a much more sophisticated user not the first thing you see.

terrylucas:
> Various UI tweaks based on UX review:
> 
> * wrap `DropdownButton` widgets with `DropdownButtonHideUnderline`
> * replace `Label` widgets with `MaterialIconLabel` widgets, which work better in both themes
> * use the same live chart height and padding as the timeline page
> * some padding and spacing tweaks
> * don't use a Split widget to contain the live memory chart - just have a fixed size
> * move the 'android memory' button to the main memory page toolbar
> * remove the slider control from the memory page (I think having both this and the 1 min / 5 min / ... pulldown was confusing)
> * when the android memory chart is visible, display it side-by-side with the dart heap chart (instead of below the dart chart)
> 
> @terrylucas - most of these changes are to the memory page; happy to discuss these on the PR or in person

My only real concern is displaying horizontally the Dart VM heap and ADB memory instead of vertically -it loses it common X-axis timeline which I think is important.  I was looking at eliminating the 2 xaxis and combining to one.

raison00:
This mock concept shows padding changes and a legend for the Dart VM Heap and ADB memory.  It includes borders and small UI tweaks to unify DevTools pages.  

dart-android-memory-profiler

Changes include: Font color updates, padding, borders, layout, legend.

raison00:
This mock concept shows padding changes and a legend for the Dart VM Heap. It includes borders and small UI tweaks to unify DevTools pages.  

No Android device is detected, so a deactivated Android Memory button appears at the top, aligned with the other buttons.

-flutter--screen

devoncarew:
@terrylucas - I added back in the slider control (below the main memory graph). I left the Android memory graph to the right of the main memory graph when the android one is toggled on, and I restored the 'Android Memory' toggle button.

I think we can re-evaluate how best to scrub through the time samples in the graph once your fixes to the slider have landed. We may be fine with either the ability to change the visible range, or the ability to scroll through the x-values, but may not need both features.

@raison00 - thanks for the UI feedback! I think we should land this, then determine the deltas between the new UI and your proposal. We can perhaps track those as a punch-list issue for the memory page? That would also be a good place to discuss the changes in general.

Screen Shot 2020-05-12 at 2 50 46 PM

Screen Shot 2020-05-12 at 3 00 35 PM

PR #1946, opened 2020-05-08
Switch to a refresh-based timeline rather than record-stop.

kenzieschmoll:
Switching to a refresh-based timeline allows us to load any initial timeline data upon switching to the timeline page. This will also allow for tracing startup, which is a P1 use case for dream (g3).

Refreshing pulls all available timeline data. Clearing clears the timeline visually as well as the data on the VM. This is how observatory handles both these cases.

Screen Shot 2020-05-08 at 9 47 04 AM


packages/devtools_app/lib/src/timeline/flutter/timeline_controller.dart
```diff
@@ -330,9 +417,10 @@ class TimelineController
     _httpTimelineLoggingEnabledNotifier.value = state;
   }
 
-  Future<void> clearData() async {
-    if (serviceManager.hasConnection) {
+  Future<void> clearData({bool clearVmTimeline = true}) async {
```
jacob314, line 420: a quick comment could be helpful indicating when you should set clearVmTimeline to false.

packages/devtools_app/lib/src/timeline/flutter/timeline_processor.dart
```diff
@@ -139,9 +138,7 @@ class TimelineProcessor {
       // (e.g. thread_name events) as well as events from before we started
       // recording.
       final ts = event.event.timestampMicros;
-      final isTimeValid = ts != null &&
-          startRecordingMicros != null &&
-          ts >= startRecordingMicros;
+      final isTimeValid = ts != null;
```
jacob314, line 141: nit: this variable is now simple enough that you could inline it.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -130,7 +130,7 @@ void executeWithDelay(Duration delay, void callback(),
   }
 }
 
-Future<void> delayForBatchProcessing({int micros = 0}) async {
+Future<void> delayForBatchProcessing({int micros = 100}) async {
   // Even with a delay of 0 microseconds, awaiting this delay is enough to free
```
jacob314, line 134: this comment is now obsolete. did something change that made micros=0 insufficient? 100 microseconds is still tiny.

packages/devtools_testing/lib/support/flutter/timeline_test_data.dart
```diff
@@ -1001,3 +1001,22 @@ final durationEventsWithDuplicateTraces = [
...
+    },
+    {
+      'name': 'thread_name',
+      'tid': testRasterThreadId,
+      'args': {'name': '1.raster'},
+    },
+    ...goldenTraceEventsJson,
+  ],
+  'timeOriginMicros': 118039650802,
+  'timeExtentMicros': 118039679873 - 118039650802,
```
jacob314, line 1021: I like how you wrote timeExtentMicros as the delta between the two other hard coded times!

PR #1951, opened 2020-05-11
Make recorded Timeline streams configurable from DevTools dialog.

kenzieschmoll:
This PR also moves the Network Traffic toggle into the dialog and enables it on the Timeline by default.

Screen Shot 2020-05-11 at 2 55 45 PM
Screen Shot 2020-05-11 at 2 54 40 PM

### DIFF
--- a/packages/devtools_app/lib/src/timeline/flutter/timeline_screen.dart
+++ b/packages/devtools_app/lib/src/timeline/flutter/timeline_screen.dart
@@ -216,7 +216,6 @@ class TimelineScreenBodyState extends State<TimelineScreenBody>
           ),
         // TODO(kenz): hide or disable button if http timeline logging is not
         // available.
-        _logNetworkTrafficButton(),
         const SizedBox(width: defaultSpacing),
         Container(
           height: Theme.of(context).buttonTheme.height,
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/timeline/flutter/timeline_screen.dart
```diff
@@ -229,30 +228,25 @@ class TimelineScreenBodyState extends State<TimelineScreenBody>
...
+        const SizedBox(width: defaultSpacing),
+        ActionButton(
+          child: OutlineButton(
+            child: const Icon(
+              Icons.more_vert,
+              size: defaultIconSize,
+            ),
+            onPressed: _openSettingsDialog,
+          ),
+          tooltip: 'More Timeline Configurations',
```
devoncarew, line 240: `Timeline  Configuration`

packages/devtools_app/lib/src/timeline/flutter/timeline_screen.dart
```diff
@@ -333,3 +327,79 @@ class TimelineScreenBodyState extends State<TimelineScreenBody>
...
+          mainAxisAlignment: MainAxisAlignment.start,
+          children: [
+            checkbox,
+            title,
+          ],
+        ),
+      );
+    }
+
+    // Special case "Network Traffic" because it is not implemented as a
```
devoncarew, line 384: If this does have some differences from the POV of the user, we could use a `Divider` widget to visually separate it from the other items.

packages/devtools_app/lib/src/timeline/flutter/timeline_screen.dart
```diff
@@ -333,3 +327,79 @@ class TimelineScreenBodyState extends State<TimelineScreenBody>
...
+  }
+
+  List<Widget> _recordedStreams() {
+    final settings = <Widget>[];
+    for (final streamName in controller.recordedStreams.keys) {
+      final title = Text(streamName);
+      final checkbox = ValueListenableBuilder(
+        valueListenable: controller.recordedStreams[streamName],
+        builder: (context, enabled, _) {
+          return Checkbox(
```
devoncarew, line 366: For many of these items, the user may not have a sense for what toggling them does. For the ones where we can come up with good, brief descriptions, perhaps include the description in-line and to the right pg the checkbox, w/ a subtle text style?

packages/devtools_app/lib/src/timeline/flutter/timeline_screen.dart
kenzieschmoll: I thought about that. I don't think it does have a difference from the POV of the user. To the user, all the other toggles mean "do I want to see these events in the timeline or not", and this is what the result of the Network toggle will be for the user as well.

packages/devtools_app/lib/src/timeline/flutter/timeline_screen.dart
kenzieschmoll: That's a good idea. Added a TODO as I'm following up with @bkonyi to find accurate docs on these.

devoncarew: OK; we may just need a good enough description to let the user know why they want want to turn these on.

I assume for many of them they really won't want to touch them. Some they will know what they do and want to enable them, like the network toggle. Not sure if there's a way to call out the two types well in the UI - have a regular and an advanced section?

kenzieschmoll: The observatory categorizes different sets like so:
Screen Shot 2020-05-12 at 7 52 12 AM

Clicking one of these profiles enables a different set of timeline flags

PR #1954, opened 2020-05-13
[Debugger] Added "Copy to Clipboard" and "Clear" to stdio console.

ditman:
Added a couple of IconButtons to Copy to Clipboard and Clear the stdio output in the Debugger panel. (Those only show when there's *something* in `stdio`).

Screen Shot 2020-05-13 at 12 22 28 PM

Added a few unit tests for the logic.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -62,6 +62,7 @@ class _DebuggerConsoleState extends State<DebuggerConsole> {
       lines: _lines,
       controls: [
         if (_lines.isNotEmpty) ...[
+          // TODO(ditman) Extract these IconButtons to common_widgets.dart
           IconButton(
             icon: const Icon(Icons.content_copy),
             onPressed: () => copyToClipboard(_lines, context),
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -2,9 +2,12 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 
 import '../../flutter/common_widgets.dart';
+import '../../flutter/notifications.dart';
 import '../../flutter/theme.dart';
 import '../../flutter/utils.dart';
 import 'codeview.dart';
@@ -30,7 +33,84 @@ class Console extends StatefulWidget {
 }
 
 class _ConsoleState extends State<Console> {
-  final scrollController = ScrollController();
+  bool canDelete = false;
+
+  @override
+  Widget build(BuildContext context) {
+    return Stack(
+      children: [
+        ConsoleOutput(lines: widget.controller.stdio),
+        Container(
+          alignment: Alignment.bottomRight,
+          child: ConsoleControls(
+            lines: widget.controller.stdio,
+            onCopyPressed: () {
+              Clipboard.setData(ClipboardData(
+                text: widget.controller.stdio.value.join('\n'),
+              )).then((_) {
+                Notifications.of(context)?.push(
+                  'Copied!',
+                );
+              });
+            },
+            onDeletePressed: widget.controller.clearStdio,
+          ),
+        ),
+      ],
+    );
+  }
+}
+
+/// Optionally renders a ButtonBar with Console Controls,
+/// Copy and Clear.
+/// The callbacks for those buttons are passed from the outside.
+class ConsoleControls extends StatelessWidget {
+  const ConsoleControls({
+    @required ValueListenable<List<String>> lines,
+    this.onCopyPressed,
+    this.onDeletePressed,
+  })  : _lines = lines,
+        super();
+
+  final Function onCopyPressed;
+  final Function onDeletePressed;
+
+  final ValueListenable<List<String>> _lines;
+
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder<List<String>>(
+        valueListenable: _lines,
+        builder: (context, lines, _) {
+          return lines.isEmpty
+              ? Container()
+              : ButtonBar(
+                  buttonPadding: EdgeInsets.zero,
+                  alignment: MainAxisAlignment.end,
+                  children: [
+                    IconButton(
+                        icon: const Icon(Icons.content_copy),
+                        onPressed: onCopyPressed,
+                        tooltip: 'Copy to clipboard.'),
+                    IconButton(
+                      icon: const Icon(Icons.delete),
+                      onPressed: onDeletePressed,
+                      tooltip: 'Clear console output.',
+                    ),
+                  ],
+                );
+        });
+  }
+}
+
+/// Renders a widget with the output of the console.
+/// This is a ListView of text lines, with monospace font and a border.
+class ConsoleOutput extends StatelessWidget {
+  const ConsoleOutput({@required ValueListenable<List<String>> lines})
+      : _lines = lines,
+        super();
+
+  final ValueListenable<List<String>> _lines;
 
   @override
   Widget build(BuildContext context) {
@@ -38,11 +118,13 @@ class _ConsoleState extends State<Console> {
     final textStyle =
         theme.textTheme.bodyText2.copyWith(fontFamily: 'RobotoMono');
 
+    final scrollController = ScrollController();
+
     return OutlineDecoration(
       child: Padding(
         padding: const EdgeInsets.all(denseSpacing),
         child: ValueListenableBuilder<List<String>>(
-          valueListenable: widget.controller.stdio,
+          valueListenable: _lines,
           builder: (context, lines, _) {
             // If we're at the end already, scroll to expose the new content.
             if (scrollController.hasClients &&
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -29,7 +29,8 @@ class DebuggerConsole extends StatefulWidget {
   @override
   _DebuggerConsoleState createState() => _DebuggerConsoleState();
 
-  static const copyToClipboardButtonKey = Key('debugger_console_copy_to_clipboard_button');
+  static const copyToClipboardButtonKey =
+      Key('debugger_console_copy_to_clipboard_button');
   static const clearStdioButtonKey = Key('debugger_console_clear_stdio_button');
 }
 
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -4,10 +4,8 @@
 
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter/services.dart';
 
 import '../../flutter/common_widgets.dart';
-import '../../flutter/notifications.dart';
 import '../../flutter/theme.dart';
 import '../../flutter/utils.dart';
 import 'codeview.dart';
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -30,7 +33,102 @@ class Console extends StatefulWidget {
...
+      builder: (context, lines, _) {
+        // Extract this to a reusable ConsoleWithControls StatelessWidget?
+        return Stack(
+          children: [
+            ConsoleOutput(lines: lines),
+            Container(
+              alignment: Alignment.bottomRight,
+              child: ConsoleControls(
+                disabled: lines.isEmpty,
+                onCopyPressed: () {
```
devoncarew, line 49: Move the body of this callback out into an instance method?

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -30,7 +33,102 @@ class Console extends StatefulWidget {
...
+                  Clipboard.setData(ClipboardData(
+                    text: lines.join('\n'),
+                  )).then((_) {
+                    Notifications.of(context)?.push(
+                      'Copied!',
```
devoncarew, line 54: Perhaps, `Copied n lines.` to give the user a bit more info? `Copied nnn characters.`? I you change the text, we have some simple utility methods to help with pluralization.

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -30,7 +33,102 @@ class Console extends StatefulWidget {
 }
 
 class _ConsoleState extends State<Console> {
-  final scrollController = ScrollController();
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder<List<String>>(
+      valueListenable: widget.controller.stdio,
+      builder: (context, lines, _) {
+        // Extract this to a reusable ConsoleWithControls StatelessWidget?
```
devoncarew, line 41: I'd remove this comment, or convert it into a TODO(username): comment.

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -30,7 +33,102 @@ class Console extends StatefulWidget {
...
+/// Copy and Clear.
+/// The callbacks for those buttons are passed from the outside.
+class ConsoleControls extends StatelessWidget {
+  const ConsoleControls({
+    this.disabled,
+    this.onCopyPressed,
+    this.onDeletePressed,
+  }) : super();
+
+  final Function onCopyPressed;
```
devoncarew, line 78: Use `VoidFunction`?

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -30,7 +33,102 @@ class Console extends StatefulWidget {
...
+            buttonPadding: EdgeInsets.zero,
+            alignment: MainAxisAlignment.end,
+            children: [
+              IconButton(
+                icon: const Icon(Icons.content_copy),
+                onPressed: onCopyPressed,
+                tooltip: 'Copy to clipboard.',
+              ),
+              IconButton(
+                icon: const Icon(Icons.block),
```
devoncarew, line 96: Use `Icons.delete`? This is what we use to remove all breakpoints in the breakpoints view.

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -30,7 +33,102 @@ class Console extends StatefulWidget {
...
+  }) : super(key: key);
+
+  final List<String> lines;
+
+  @override
+  _ConsoleOutputState createState() => _ConsoleOutputState();
+}
+
+class _ConsoleOutputState extends State<ConsoleOutput> {
+  // The scroll controller must survive ConsoleOutput re-renders
```
devoncarew, line 120: Thanks! This is good cleanup.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -137,6 +137,11 @@ class DebuggerController extends DisposableController
   InstanceRef get reportedException => _reportedException;
   InstanceRef _reportedException;
 
+  /// Clears the contents of stdio
```
devoncarew, line 140: nit: end with a period.

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -30,7 +33,102 @@ class Console extends StatefulWidget {
...
+
+  final Function onCopyPressed;
+  final Function onDeletePressed;
+  final bool disabled;
+
+  @override
+  Widget build(BuildContext context) {
+    return disabled
+        ? Container()
+        : ButtonBar(
```
devoncarew, line 86: I wonder it we should be using `ButtonBar` in more places in devtools?

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -30,7 +33,102 @@ class Console extends StatefulWidget {
...
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder<List<String>>(
+      valueListenable: widget.controller.stdio,
+      builder: (context, lines, _) {
+        // Extract this to a reusable ConsoleWithControls StatelessWidget?
+        return Stack(
+          children: [
+            ConsoleOutput(lines: lines),
+            Container(
+              alignment: Alignment.bottomRight,
```
devoncarew, line 46: We may want an upper right alignment here, with a bit of padding from the top and right sides. We have some standard sizes in our theme.dart file - here we might use either `defaultSpacing` or `denseSpacing` depending on how they look.

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -30,7 +33,102 @@ class Console extends StatefulWidget {
...
+                onPressed: onDeletePressed,
+                tooltip: 'Clear console output.',
+              ),
+            ],
+          );
+  }
+}
+
+/// Renders a widget with the output of the console.
+/// This is a ListView of text lines, with a monospace font and a border.
```
kenzieschmoll, line 106: nit: dartdoc style - add a new line between the first sentence and all others.

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -30,7 +33,102 @@ class Console extends StatefulWidget {
...
+          ],
+        );
+      },
+    );
+  }
+}
+
+/// Optionally renders a ButtonBar with Console Controls,
+/// Copy and Clear.
+/// The callbacks for those buttons are passed from the outside.
```
kenzieschmoll, line 70: nit: dartdoc style - new line after first sentence

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -30,7 +33,102 @@ class Console extends StatefulWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder<List<String>>(
+      valueListenable: widget.controller.stdio,
+      builder: (context, lines, _) {
+        // Extract this to a reusable ConsoleWithControls StatelessWidget?
+        return Stack(
+          children: [
+            ConsoleOutput(lines: lines),
+            Container(
```
kenzieschmoll, line 45: instead of adding the disabled parameter to ConsoleControls, write `if (lines.isNotEmpty)` here, and then the container will only be built if that condition is met

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -30,7 +33,102 @@ class Console extends StatefulWidget {
...
+    return disabled
+        ? Container()
+        : ButtonBar(
+            buttonPadding: EdgeInsets.zero,
+            alignment: MainAxisAlignment.end,
+            children: [
+              IconButton(
+                icon: const Icon(Icons.content_copy),
+                onPressed: onCopyPressed,
+                tooltip: 'Copy to clipboard.',
```
kenzieschmoll, line 93: remove period at the end of tooltips

packages/devtools_app/lib/src/debugger/flutter/console.dart
ditman: Yep, will do.

packages/devtools_app/lib/src/debugger/flutter/console.dart
ditman: @devoncarew I did a quick search and I couldn't find much i18n, but I can change this to `Copied n lines.`. Where's the pluralization stuff?

packages/devtools_app/lib/src/debugger/flutter/console.dart
ditman: I used `delete`, but then I saw the "Clear" in the "Logging" panel using `block`, and changed it last minute (after I took the screenshot, btw! :P)

packages/devtools_app/lib/src/debugger/flutter/console.dart
devoncarew: If you search through the source for `pluralize` you'll see it - it'll look something like this: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart#L380

packages/devtools_app/lib/src/debugger/flutter/console.dart
ditman: Thanks! I want to make this console more reusable, once I get the TODO above done, I might be able to reuse this in the logging panel (so we get "Copy" there for free)

packages/devtools_app/lib/src/debugger/flutter/console.dart
devoncarew: Gotcha (this is closer to the logging console than the breakpoints view)

packages/devtools_app/lib/src/debugger/flutter/console.dart
ditman: Very good point, will do!

packages/devtools_app/lib/src/debugger/flutter/console.dart
ditman: I tried controlling that padding, but I think the ButtonBar is doing its own thing. I'll give this a shot, it's easy enough to test!

packages/devtools_app/lib/src/debugger/flutter/console.dart
ditman: In fact, I'm going to do what the comment suggests so I can get rid of it :P

packages/devtools_app/lib/src/debugger/flutter/console.dart
ditman: I think I'm going to refactor this so the "actions" can be passed from the outside as a `<Widget>[]`, instead of having to pass each callback on its own. That way the user of the Console widget can define what actions are available (for example, in Logging, only "Copy" would be available)

packages/devtools_app/lib/src/debugger/flutter/console.dart
ditman: (lib/src/utils.dart -> pluralize)

packages/devtools_app/lib/src/debugger/flutter/console.dart
ditman: Yep, done.

packages/devtools_app/lib/src/debugger/flutter/console.dart
ditman: Changed to `delete`

ditman:
This is an updated screenshot of how the widget currently looks:

Screen Shot 2020-05-13 at 12 22 28 PM

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -17,20 +18,144 @@ import 'debugger_controller.dart';
...
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Console(
+      lines: _lines,
+      controls: [
+        if (_lines.isNotEmpty) ...[
+          IconButton(
```
kenzieschmoll, line 65: add a TODO to pull these action widgets into common_widgets.dart if they will also be used in the logging screen.

ditman: Thanks for the super fast review!

I'll add a TODO there. I'll get back to the team to find the best place to put the extracted code next week!

PR #1956, opened 2020-05-13
Delete dart:html app.

kenzieschmoll:
Parts of the CL that need the most review:
- Changes to index.html and deleted entry point files (see commit "Delete html app entrypoint and update index.html")
- Icon cleanup code (see commit "Clean up icon rendering code.")

I kept the commit history as clean as I could so viewing specific commits may make it easier to review. @jacob314 @devoncarew @terrylucas @DanTup 

There likely is still more cleanup to do after this CL lands, but we can do that incrementally in our own areas of expertise within the app. Meta tracking issue for that here: https://github.com/flutter/devtools/issues/1965


DanTup:
The code changes look reasonable to me (though you might want more qualified opinions 😁).

However I did notice some strange graphical issues when running this version locally - I'm not sure if they're related to this change or just intermittent:

html

I've opened #1959 which should fix server integration tests for this.

kenzieschmoll:
The icons on the right of the text in the details view are correct. The rest is strange though, I am not seeing those issues. What platform are you running on?

packages/devtools_app/web/index.html
```diff
@@ -55,134 +26,20 @@
...
-    </footer>
-
-    <div id="toast-container" class="legacy-dart">
-    </div>
+<body>
+  <!-- This script installs service_worker.js to provide PWA functionality to
+       application. For more information, see:
+       https://developers.google.com/web/fundamentals/primers/service-workers -->
+  <script>
+    if ('serviceWorker' in navigator) {
```
devoncarew, line 37: I don't know that we need this, but we don't need to resolve for this PR. We should file an issue to investigate whether we want to use a service worker w/ DevTools.

DanTup:
> The rest is strange though, I am not seeing those issues. What platform are you running on?

macOS (Catalina). 

I repro'd without these changes, so seem unrelated. It happens if I just switch between a few tabs then go back to the inspector page. There are also some assertions failing which may be related - I'll file as a separate issue.

packages/devtools_app/web/index.html
kenzieschmoll: This was copied from the default configuration for a flutter web app. Filed issue to investigate further https://github.com/flutter/devtools/issues/1963

kenzieschmoll:
Disabled the integration tests to get the bots to go green. `flutter run -d macos` and `flutter run -d chrome` WAI to build the app, but the bundling for release needs to be fixed.

Re-enabling the tests and fixing the build from server are filed as P0s [here](https://github.com/flutter/devtools/issues/1965). Landing this so that we have a base to iterate off of.

PR #1958, opened 2020-05-13
Improve the debugger script cache

devoncarew:
- improve the debugger script cache; add entries to the cache for script requests that are in progress, as well as ones that we've already retreived
- move the cache into its own class (`ScriptCache`) instead of using a map in the DebuggerController class, now that the cache has gotten more complicated
- optimize frame display when we pause; we now use the `pauseEvent.topFrame` field, and display the pause immediately, while asynchronously populating the information about the other frames in the background. This will likely reduce the stepping latency a bit when debugging dart web apps (its currently expensive for package:dwds calculate the frame info)

Previously, when the script cache was empty, we were sending n requests for scripts when we paused, where n was the number of frames. I.e., the cache didn't handle multiple frames that referenced the same script, where that script was not yet populated in the cache. This issue was found using the new logging code in package:dwds - https://github.com/dart-lang/webdev/pull/1011.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart' hide Stack;
+import 'package:pedantic/pedantic.dart';
 import 'package:vm_service/vm_service.dart';
 
 import '../../auto_dispose.dart';
@@ -34,7 +35,7 @@ class DebuggerController extends DisposableController
 
   VmService get _service => serviceManager.service;
 
-  final _scriptCache = <String, Script>{};
+  final ScriptCache _scriptCache = ScriptCache();
 
   final _isPaused = ValueNotifier<bool>(false);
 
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -755,3 +763,59 @@ class DebuggerController extends DisposableController
...
+
+  /// Return a cached [Script] for the given [ScriptRef], returning null
+  /// if there is no cached [Script].
+  Script getScriptCached(ScriptRef scriptRef) {
+    // Check to see if this ScriptRef is really a Script.
+    if (scriptRef is Script) {
+      if (_scripts[scriptRef.id] == null) {
+        _scripts[scriptRef.id] = scriptRef;
+      }
+
```
kenzieschmoll, line 781: nit: not your fault, but I don't think these new lines add much value. Let's remove them to make this method more dense

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -446,25 +441,47 @@ class DebuggerController extends DisposableController
     }
   }
 
-  Future<void> _pause(bool pause) async {
-    _isPaused.value = pause;
```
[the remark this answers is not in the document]
devoncarew: No really; this made it slightly more clear what the two branches do, but I don't think it's critical to do it this way.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
devoncarew: Agreed, thinking about this, it's not likely that we'll have gotten a `Script` object without having already gone through this caching code; removed.

PR #1970, opened 2020-05-15
some tweaks to the debugger's console actions

devoncarew:
Some tweaks to the debugger's console actions:
- use a const for the icon size (from theme.dart, `actionsIconSize`)
- have the actions always visible - even if they don't apply at the time - but disabled when necessary
- a slight tweak to the 'copied n lines text'; it looked strange when there were 3 lines in the console (w/ a 4th empty newline), but the copied text said 'copied 4 lines'

cc @ditman

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -97,9 +97,10 @@ class Console extends StatelessWidget {
     return Stack(
       children: [
         _ConsoleOutput(lines: lines),
-        _ConsoleControls(
-          controls: controls,
-        ),
+        if (controls.isNotEmpty)
+          _ConsoleControls(
+            controls: controls,
+          ),
       ],
     );
   }
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -58,24 +58,26 @@ class _DebuggerConsoleState extends State<DebuggerConsole> {
 
   @override
   Widget build(BuildContext context) {
-    return Console(
-      lines: _lines,
-      controls: [
-        // TODO(ditman) Extract these IconButtons to common_widgets.dart
-        IconButton(
-          icon: const Icon(Icons.content_copy, size: actionsIconSize),
-          onPressed:
-              _lines.isEmpty ? null : () => copyToClipboard(_lines, context),
-          tooltip: 'Copy to clipboard',
-          key: DebuggerConsole.copyToClipboardButtonKey,
-        ),
-        IconButton(
-          icon: const Icon(Icons.delete, size: actionsIconSize),
-          onPressed: _lines.isEmpty ? null : widget.controller.clearStdio,
-          tooltip: 'Clear console output',
-          key: DebuggerConsole.clearStdioButtonKey,
-        ),
-      ],
+    return Material(
+      child: Console(
+        lines: _lines,
+        controls: [
+          // TODO(ditman) Extract these IconButtons to common_widgets.dart
+          IconButton(
+            icon: const Icon(Icons.content_copy, size: actionsIconSize),
+            onPressed:
+                _lines.isEmpty ? null : () => copyToClipboard(_lines, context),
+            tooltip: 'Copy to clipboard',
+            key: DebuggerConsole.copyToClipboardButtonKey,
+          ),
+          IconButton(
+            icon: const Icon(Icons.delete, size: actionsIconSize),
+            onPressed: _lines.isEmpty ? null : widget.controller.clearStdio,
+            tooltip: 'Clear console output',
+            key: DebuggerConsole.clearStdioButtonKey,
+          ),
+        ],
+      ),
     );
   }
 }
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -61,29 +61,28 @@ class _DebuggerConsoleState extends State<DebuggerConsole> {
     return Console(
       lines: _lines,
       controls: [
-        if (_lines.isNotEmpty) ...[
-          // TODO(ditman) Extract these IconButtons to common_widgets.dart
-          IconButton(
-            icon: const Icon(Icons.content_copy),
-            onPressed: () => copyToClipboard(_lines, context),
-            tooltip: 'Copy to clipboard',
-            key: DebuggerConsole.copyToClipboardButtonKey,
-          ),
-          IconButton(
-            icon: const Icon(Icons.delete),
-            onPressed: widget.controller.clearStdio,
-            tooltip: 'Clear console output',
-            key: DebuggerConsole.clearStdioButtonKey,
-          ),
-        ],
+        // TODO(ditman) Extract these IconButtons to common_widgets.dart
+        IconButton(
+          icon: const Icon(Icons.content_copy, size: actionsIconSize),
+          onPressed:
+              _lines.isEmpty ? null : () => copyToClipboard(_lines, context),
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
```diff
@@ -238,7 +238,7 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
           ActionButton(
             child: FlatButton(
               padding: EdgeInsets.zero,
-              child: const Icon(Icons.delete, size: 24.0),
+              child: const Icon(Icons.delete, size: actionsIconSize),
```
ditman, line 241: Good change, apologies for adding a magic number there!

packages/devtools_app/lib/src/flutter/utils.dart
```diff
@@ -24,9 +24,14 @@ Future<void> copyToClipboard(List<String> lines, BuildContext context) async {
     text: lines.join('\n'),
   ));
 
-  final numLines = lines.length;
+  var numLines = lines.length;
+  // Adjust line count if there's a trailing empty line.
+  if (numLines > 0 && lines.last.trim().isEmpty) {
+    numLines--;
+  }
```
ditman, line 31: IMO this should be fixed in the data source. Ensure that the data passed to the console never has weird empty lines (this would also allow us to remove the "render the data in the previous line" trick in the stdio stream controller)

packages/devtools_app/test/flutter/debugger_screen_test.dart
```diff
@@ -114,15 +114,20 @@ void main() {
     });
 
     group('ConsoleControls', () {
-      testWidgets('Console Controls are not rendered when stdio is empty',
+      testWidgets('Console Controls are rendered disables when stdio is empty',
```
ditman, line 117: Typo in "disables" should be: "disabled"

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -98,17 +97,16 @@ class Console extends StatelessWidget {
...
       children: [
         _ConsoleOutput(lines: lines),
-        if (controls.isNotEmpty) ...[
-          _ConsoleControls(
-            controls: controls,
-          ),
-        ],
+        _ConsoleControls(
+          controls: controls,
+        ),
```
ditman, line 102: I'd keep the `isNotEmpty` check here, so we don't end up rendering unneeded containers (I'm planning on reusing this console output in the "Logging" screen, and there are cases where we need to render a completely empty Console)

packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -61,29 +61,28 @@ class _DebuggerConsoleState extends State<DebuggerConsole> {
...
+        IconButton(
+          icon: const Icon(Icons.content_copy, size: actionsIconSize),
+          onPressed:
+              _lines.isEmpty ? null : () => copyToClipboard(_lines, context),
+          tooltip: 'Copy to clipboard',
+          key: DebuggerConsole.copyToClipboardButtonKey,
+        ),
+        IconButton(
+          icon: const Icon(Icons.delete, size: actionsIconSize),
+          onPressed: _lines.isEmpty ? null : widget.controller.clearStdio,
```
ditman, line 74: I did the "don't render disabled buttons" trick because the inkwell animation didn't finish when setting as disabled. I hope that works better now (didn't test it after migrating the whole page to a normal setState flow)

packages/devtools_app/lib/src/debugger/flutter/console.dart
devoncarew: Gotcha, I'll add this back.

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
devoncarew: No worries - we've been converting fixed constants to common, semantic ones in theme.dart over time.

packages/devtools_app/lib/src/flutter/utils.dart
devoncarew: I think the source data is correct here - we're streaming chars into this console from stdout and stderr. There are some fiddly bits around edge cases to handle the streaming and display (and, likely some better ways we could handle those cases).

In the case I saw this issue, there were 3 lines with content, each ending in a newline, and 4 lines displayed in the console, the last one empty (w/ the user seeing 3 lines of content).

I'll double check that we shouldn't fix this in the controller - that it should just pass along 3 lines in this case.

packages/devtools_app/test/flutter/debugger_screen_test.dart
devoncarew: will fix

packages/devtools_app/lib/src/debugger/flutter/console.dart
devoncarew: Hmm, yeah, that's annoying:

Screen Shot 2020-05-16 at 5 13 56 PM

Any idea how to fix the issue? (also, I would love if the inkwell didn't paint large enough to extend outside the console area...)

packages/devtools_app/lib/src/flutter/utils.dart
devoncarew: (updated to make changes to the controller)

packages/devtools_app/lib/src/debugger/flutter/console.dart
ditman: I didn't find a good solution, let me try and get this failing in dartpad.dev, and shop it around :)

ditman: All right, two things:

> I would love if the inkwell didn't paint large enough to extend outside the console area

We may be able to wrap the Console in a `Material` widget, that'll prevent the InkWell from overflowing the Console bounds!

Also, the InkWell issue:

https://dartpad.dev/3e21e4a1d81ec137beaeaae3865948d0

I think the problem is the "hover" state of the button is rendered even when disabled (and then it never changes because the mouse listeners are removed). This is probably a framework bug, that is showing in the web, but should show up as well in desktop (anywhere where "hover" makes sense).

I tried working around the issue with something like this:

```
hoverColor: _enabled ? null : Colors.white.withAlpha(0),
```

But when the button gets re-enabled, the "hover" state is shown (it shouldn't!).

I'm 99% sure this is an IconButton issue that is not handling correctly the disabled + hover events.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -149,7 +150,7 @@ class DebuggerController extends DisposableController
     var lines = _stdio.value.toList();
     final newLines = text.split('\n');
```
ditman, line 151: What if we always text.trim() before splitting? That way we can guarantee that there will never be any leading/trailing whitespace in the incoming text, and we don't need to keep track if we've seen a newline or not?

packages/devtools_app/lib/src/debugger/flutter/console.dart
devoncarew: I'll try the `Material` trick -

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
devoncarew: Hmm, I think we need to be ~faithful to the incoming chars. Esp. wrt triming leading whitespace - those may be indent that people would want to see in the output.

Similarly if we trim trailing new console output, that may not join well with the next set of chars we see.

I think what we want here logically is to stream chars into a buffer, and create new lines as we see line endings. That (with a little bit of bookkeeping) is what this code is doing.

PR #1971, opened 2020-05-15
use the vm service protocol's getMemoryUsage() API

devoncarew:
- switch to using the vm service protocol's `getMemoryUsage()` API (instead of polling isolates)
- address one related todo

cc @terrylucas

### DIFF
--- a/packages/devtools_shared/lib/src/memory/heap_sample.dart
+++ b/packages/devtools_shared/lib/src/memory/heap_sample.dart
@@ -6,17 +6,17 @@ import 'adb_memory_info.dart';
 
 /// DevTools Plotted and JSON persisted memory information.
 class HeapSample {
-  HeapSample(this.timestamp,
-      this.rss,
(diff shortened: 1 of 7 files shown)


packages/devtools_shared/lib/src/memory/heap_sample.dart
```diff
@@ -31,7 +29,6 @@ class HeapSample {
 
   Map<String, dynamic> toJson() => <String, dynamic>{
         'timestamp': timestamp,
-        'rss': rss,
```
terrylucas, line 34: rss should be stored to both the JSON files (that we export) and for reloading too.

devoncarew: restored

devoncarew:
@terrylucas, can you re-review?

devoncarew:
Landing, and can make additional revisions as necessary.

PR #1975, opened 2020-05-16
Remove dart:html from build_release.sh

DanTup:
I *think* this is all correct. We no longer need to build the dart:html version, nor do we need to do gymnastics to put the Flutter version in a folder and change its index to flutter.html.

With this change, running the "Run Server with Release Build" launch config from VS Code seems to work as expected.

### DIFF
--- a/packages/devtools_server/lib/src/external_handlers.dart
+++ b/packages/devtools_server/lib/src/external_handlers.dart
@@ -80,32 +80,11 @@ Future<shelf.Handler> defaultHandler(
     debugProxyHandler = proxyHandler(Uri.parse('http://localhost:$webPort/'));
   }
 
-  // The packages folder is renamed in the pub package so this handler serves
-  // out of the `pack` folder.
-  Handler packHandler;
-  if (!debugMode) {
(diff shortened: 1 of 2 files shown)


DanTup:
(also removed the old web handler for mapping /packages/ to /pack/ - a folder that no longer exists without the HTML version)

DanTup:
Yep, it launches fine for me. Previously we were still building the Flutter version at `/flutter.html` so it would serve up the (missing) `dart:html` version.

FWIW, if using VS Code there's a [launch config "Run Server with Release Build"](https://github.com/flutter/devtools/blob/201e68353105254bd307dfc60e46a1ad5d13d8c0/.vscode/launch.json#L28) that will do both steps (though just `build_release.sh` instead of `publish.sh`) plus automatically launch the browser and attach the debugger to the server which may be convenient for general testing of release builds (it doesn't modify the gitignore though, so when actually ready to publish you'd need to use the publish script).

kenzieschmoll:
Awesome. Thanks for the fix!

kenzieschmoll:
lgtm_army

PR #1978, opened 2020-05-18
improved feedback when stepping

devoncarew:
- when stepping, show the stepping controls as disabled until we get an indication that the app has resumed (this provides better feedback that hitting the step button did something)
- slow animations down for scrolling the editor when stepping
- add some dev time flags to help diagnose where we're spending time when stepping (to help diagnose slow package:dwds stepping)

I have a `ValueListenableBuilder` in a `ValueListenableBuilder` in this PR, which is correct I guess, but it would be nice if there was a technique that involved less widget nesting.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/controls.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/controls.dart
@@ -21,88 +21,97 @@ class DebuggingControls extends StatelessWidget {
   Widget build(BuildContext context) {
     return ValueListenableBuilder(
       valueListenable: controller.isPaused,
-      builder: (context, isPaused, child) {
-        final canStep = isPaused && controller.hasFrames.value;
+      builder: (context, isPaused, _) {
+        return ValueListenableBuilder(
+          valueListenable: controller.maybeResuming,
+          builder: (context, maybeResuming, Widget _) {
+            final canStep =
+                isPaused && !maybeResuming && controller.hasFrames.value;
 
-        return SizedBox(
-          height: Theme.of(context).buttonTheme.height,
-          child: Row(
-            children: [
-              RoundedOutlinedBorder(
-                child: Row(
-                  children: [
-                    DebuggerButton(
-                      title: 'Pause',
-                      icon: Icons.pause,
-                      autofocus: true,
-                      onPressed: isPaused ? null : controller.pause,
+            return SizedBox(
+              height: Theme.of(context).buttonTheme.height,
+              child: Row(
+                children: [
+                  RoundedOutlinedBorder(
+                    child: Row(
+                      children: [
+                        DebuggerButton(
+                          title: 'Pause',
+                          icon: Icons.pause,
+                          autofocus: true,
+                          onPressed: isPaused ? null : controller.pause,
+                        ),
+                        _LeftBorder(
+                          child: DebuggerButton(
+                            title: 'Resume',
+                            icon: Icons.play_arrow,
+                            onPressed: (isPaused && !maybeResuming)
+                                ? controller.resume
+                                : null,
+                          ),
+                        ),
+                      ],
                     ),
-                    _LeftBorder(
-                      child: DebuggerButton(
-                        title: 'Resume',
-                        icon: Icons.play_arrow,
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -41,6 +44,12 @@ class DebuggerController extends DisposableController
 
   ValueListenable<bool> get isPaused => _isPaused;
 
+  final _maybeResuming = ValueNotifier<bool>(false);
+
+  /// This indicates that we've requested a resume (or step) operation from the
+  /// VM, but haven't yet received the 'resumed' isolate event.
+  ValueListenable<bool> get maybeResuming => _maybeResuming;
```
kenzieschmoll, line 51: why is this maybeResuming instead of just resuming? "maybe" makes it seem like we can never really guarantee that this value is the true value of "resuming"

packages/devtools_app/lib/src/debugger/flutter/controls.dart
```diff
@@ -21,88 +21,97 @@ class DebuggingControls extends StatelessWidget {
...
-              RoundedOutlinedBorder(
-                child: Row(
-                  children: [
-                    DebuggerButton(
-                      title: 'Pause',
-                      icon: Icons.pause,
-                      autofocus: true,
-                      onPressed: isPaused ? null : controller.pause,
+      builder: (context, isPaused, _) {
+        return ValueListenableBuilder(
```
kenzieschmoll, line 25: another option would be to make DebuggingControls a stateful widget and then add listeners that call setState when either of these value listenables change, but I'm not sure that is better than what you already have.

You could break the content up into a helper method if you think that makes it more clear. e.g.

```
return ValueListenableBuilder(
  ...
    return ValueListenableBuilder(
    ...
      return _buildControls(paused: isPaused, maybeResuming: maybeResuming);
    ),
);
```

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
```diff
@@ -811,3 +846,28 @@ class ScriptCache {
     _inProgress.clear();
   }
 }
+
+/// A dev time class to help trace DevTools application events.
+class TimingLogger {
```
kenzieschmoll, line 851: let's move this to utils since it is not specific to the debugger. Also, maybe rename to DebugTimingLogger to make it clear that this a is a debugging class

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
```diff
@@ -27,6 +27,10 @@ import 'debugger_model.dart';
 import 'scripts.dart';
 import 'variables.dart';
 
+// This is currently a dev-time flag as the overall call depth may not be that
+// interesting to users.
+const bool showCallStackCount = false;
```
kenzieschmoll, line 32: nit: rename to `debugShowCallStackCount` to make it clear that this flag is for debugging purposes

packages/devtools_app/lib/src/debugger/flutter/controls.dart
devoncarew: Yeah, likely not something to optimize now, but perhaps if we really thought we could improve the code readability we could have a widget line `ValueListenableBuilder2<T, S>` that could handle two ValueListeners.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
devoncarew: I think I didn't know that this would be that precise when I wrote this; I'll change to something like `resuming` or `resumeInProgress`.

packages/devtools_app/lib/src/debugger/flutter/debugger_controller.dart
devoncarew: good idea, will do

packages/devtools_app/lib/src/debugger/flutter/debugger_screen.dart
devoncarew: changed

PR #1979, opened 2020-05-18
Auto-complete w/ overlay, object details in External heap and analysis of snapshot (early on images).

terrylucas:
- Support programmatic setting and detecting table selection.
- Added auto-complete drop-down overlay to TextField (search).
- Fixed heat map painting.
- Enable auto-complete search in both heat map and memory tree views.

Auto-complete overlay appears below the search TextField as characters are typed the matching classes, libraries, appear.  More UX work to do e.g., up/down arrow, bold matches in the popup list, etc. but helps to search, display and select items the heat map areas and expand and select an item in the library/classes tree e.g.,

image

- Added details about objects in the External heap.
- Added Analysis of the snapshot discoving large memory blocks consumed in either the ImageCache or raw data (_int32List).
- Fixed a few bugs when drilling in on a instance.

### DIFF
--- a/packages/devtools_app/lib/src/flutter/table.dart
+++ b/packages/devtools_app/lib/src/flutter/table.dart
@@ -7,6 +7,7 @@ import '../flutter/common_widgets.dart';
 import '../table_data.dart';
 import '../trees.dart';
 import '../utils.dart';
+import 'auto_dispose_mixin.dart';
 import 'collapsible_mixin.dart';
 import 'common_widgets.dart' show ColorExtension, ScrollControllerAutoScroll;
 import 'flutter_widgets/linked_scroll_controller.dart';
@@ -155,6 +156,29 @@ class FlatTableState<T> extends State<FlatTable<T>>
   }
 }
 
+class Selection<T> {
+  Selection(
+      {this.node,
+      this.nodeIndex,
+      this.scrollIntoView = false,
+      this.displayRow});
+
+  Selection.empty()
+      : node = null,
+        nodeIndex = null,
+        scrollIntoView = false,
+        displayRow = null;
+
+  final T node;
+  final int nodeIndex;
+  final bool scrollIntoView;
+  int displayRow;
+}
+
+class SelectionNotifier<T> extends ValueNotifier<T> {
+  SelectionNotifier(T) : super(T);
+}
+
 // TODO(https://github.com/flutter/devtools/issues/1657)
 
 /// A table that shows [TreeNode]s.
@@ -181,6 +205,7 @@ class TreeTable<T extends TreeNode<T>> extends StatefulWidget {
     @required this.keyFactory,
     @required this.sortColumn,
     @required this.sortDirection,
+    this.selectionNotifier,
   })  : assert(columns.contains(treeColumn)),
         assert(columns.contains(sortColumn)),
         assert(columns != null),
@@ -204,23 +229,25 @@ class TreeTable<T extends TreeNode<T>> extends StatefulWidget {
 
   final SortDirection sortDirection;
 
+  final SelectionNotifier<Selection<T>> selectionNotifier;
+
   @override
   TreeTableState<T> createState() => TreeTableState<T>();
 }
 
 class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
-    with TickerProviderStateMixin, TreeMixin<T>
+    with TickerProviderStateMixin, TreeMixin<T>, AutoDisposeMixin
     implements SortableTable<T> {
   /// The number of items to show when animating out the tree table.
   static const itemsToShowWhenAnimating = 50;
   List<T> animatingChildren = [];
   Set<T> animatingChildrenSet = {};
   T animatingNode;
-  T selectedNode;
-  int selectedNodeIndex;
   List<double> columnWidths;
   List<bool> rootsExpanded;
   FocusNode _focusNode;
+
+  SelectionNotifier<Selection<T>> selectionNotifier;
 
   FocusNode get focusNode => _focusNode;
 
@@ -238,6 +265,20 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
   void didUpdateWidget(TreeTable oldWidget) {
     super.didUpdateWidget(oldWidget);
 
+    cancel();
+
+    addAutoDisposeListener(selectionNotifier, () {
+      selectionNotifier.value.displayRow = _selectionRowNumber();
+
+      setState(() {
+        final node = selectionNotifier.value.node;
+        final parent = node.parent;
+        if (!parent.isExpanded) {
+          _toggleNode(parent);
+        }
+      });
+    });
+
     if (widget.sortColumn != oldWidget.sortColumn ||
         widget.sortDirection != oldWidget.sortDirection ||
         !collectionEquals(widget.dataRoots, oldWidget.dataRoots)) {
@@ -248,9 +289,61 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     _updateItems();
   }
 
+  /// Return the row number of the selected node.  This row number is the visual
+  /// row number including all rows above.
+  int _selectionRowNumber() {
+    final selectedNode = selectionNotifier.value.node;
+
+    final parents = List<T>(selectedNode.level);
+    T scanNode = selectedNode;
+    while (!scanNode.isRoot) {
+      parents[scanNode.parent.level] = scanNode.parent;
+      scanNode = scanNode.parent;
+    }
+
+    var displayRows = -1;
+    for (final T data in dataRoots) {
+      displayRows++;
+      final parentLevel = data.level;
+      if (parents[parentLevel] == data) {
+        // Found the parentage match of the selectedNode, drill down children
+        // of the selectedNode.
+        for (T child in parents[parentLevel].children) {
+          displayRows++;
+          if (child == selectedNode) {
+            return displayRows;
+          }
+        }
+      } else {
+        displayRows += _expandedRows(data);
+      }
+    }
+
+    return displayRows;
+  }
+
+  /// Return total rows displayed in node expanded children and
+  /// any of their expanded nested children.
+  int _expandedRows(T node) {
+    var displayRows = 0;
+
+    if (node.isExpanded) {
+      if (node.children.isNotEmpty) {
+        for (final child in node.children) {
+          displayRows += 1 + _expandedRows(child);
+        }
+      }
+    }
+
+    return displayRows;
+  }
+
   void _initData() {
     dataRoots = List.from(widget.dataRoots);
     sortData(widget.sortColumn, widget.sortDirection);
+    selectionNotifier = widget.selectionNotifier == null
+        ? SelectionNotifier<Selection<T>>(Selection<T>.empty())
+        : widget.selectionNotifier;
   }
 
   @override
@@ -278,8 +371,10 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
 
   void _onItemPressed(T node, int nodeIndex) {
     /// Rebuilds the table whenever the tree structure has been updated
-    selectedNode = node;
-    selectedNodeIndex = nodeIndex;
+    selectionNotifier.value = Selection(
+      node: node,
+      nodeIndex: nodeIndex,
+    );
 
     _toggleNode(node);
   }
@@ -363,6 +458,7 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
       onSortChanged: _sortDataAndUpdate,
       focusNode: _focusNode,
       handleKeyEvent: _handleKeyEvent,
+      selectionNotifier: selectionNotifier,
     );
   }
 
@@ -372,7 +468,7 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     int index,
   ) {
     Widget rowForNode(T node) {
-      final isNodeSelected = selectedNode == node;
+      final isNodeSelected = selectionNotifier.value.node == node;
       return TableRow<T>(
         key: widget.keyFactory(node),
         linkedScrollControllerGroup: linkedScrollControllerGroup,
@@ -432,11 +528,14 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     ].contains(event.logicalKey)) return false;
 
     // If there is no selected node, choose the first one.
-    if (selectedNode == null) {
-      selectedNode = items[0];
-      selectedNodeIndex = 0;
+    if (selectionNotifier.value.node == null) {
+      selectionNotifier.value = Selection(
+        node: items[0],
+        nodeIndex: 0,
+      );
     }
-    assert(selectedNode == items[selectedNodeIndex]);
+    assert(selectionNotifier.value.node ==
+        items[selectionNotifier.value.nodeIndex]);
 
     if (event.logicalKey == LogicalKeyboardKey.arrowDown) {
       _moveSelection(ScrollKind.down, scrollController, constraints.maxHeight);
@@ -445,8 +544,8 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     } else if (event.logicalKey == LogicalKeyboardKey.arrowLeft) {
       // On left arrow collapse the row if it is expanded. If it is not, move
       // selection to its parent.
-      if (selectedNode.isExpanded) {
-        _toggleNode(selectedNode);
+      if (selectionNotifier.value.node.isExpanded) {
+        _toggleNode(selectionNotifier.value.node);
       } else {
         _moveSelection(
           ScrollKind.parent,
@@ -456,8 +555,9 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
       }
     } else if (event.logicalKey == LogicalKeyboardKey.arrowRight) {
       // On right arrow expand the row if possible, otherwise move selection down.
-      if (selectedNode.isExpandable && !selectedNode.isExpanded) {
-        _toggleNode(selectedNode);
+      if (selectionNotifier.value.node.isExpandable &&
+          !selectionNotifier.value.node.isExpanded) {
+        _toggleNode(selectionNotifier.value.node);
       } else {
         _moveSelection(
           ScrollKind.down,
@@ -485,6 +585,8 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     final lastItemIndex = firstItemIndex + minCompleteItemsInView - 1;
     int newSelectedNodeIndex;
 
+    final selectionValue = selectionNotifier.value;
+    final selectedNodeIndex = selectionValue.nodeIndex;
     switch (scrollKind) {
       case ScrollKind.down:
         newSelectedNodeIndex = min(selectedNodeIndex + 1, items.length - 1);
@@ -493,7 +595,8 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
         newSelectedNodeIndex = max(selectedNodeIndex - 1, 0);
         break;
       case ScrollKind.parent:
-        newSelectedNodeIndex = max(items.indexOf(selectedNode.parent), 0);
+        newSelectedNodeIndex =
+            max(items.indexOf(selectionValue.node.parent), 0);
         break;
     }
 
@@ -519,8 +622,10 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     }
 
     setState(() {
-      selectedNode = newSelectedNode;
-      selectedNodeIndex = newSelectedNodeIndex;
+      selectionNotifier.value = Selection(
+        node: newSelectedNode,
+        nodeIndex: newSelectedNodeIndex,
+      );
     });
   }
 
@@ -543,6 +648,7 @@ class _Table<T> extends StatefulWidget {
     this.focusNode,
     this.handleKeyEvent,
     this.autoScrollContent = false,
+    this.selectionNotifier,
   }) : super(key: key);
 
   final int itemCount;
@@ -556,6 +662,7 @@ class _Table<T> extends StatefulWidget {
   final Function(ColumnData<T> column, SortDirection direction) onSortChanged;
   final FocusNode focusNode;
   final TableKeyEventHandler handleKeyEvent;
+  final SelectionNotifier<Selection<T>> selectionNotifier;
 
   /// The width to assume for columns that don't specify a width.
   static const defaultColumnWidth = 500.0;
@@ -564,7 +671,7 @@ class _Table<T> extends StatefulWidget {
   _TableState<T> createState() => _TableState<T>();
 }
 
-class _TableState<T> extends State<_Table<T>> {
+class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
   LinkedScrollControllerGroup _linkedHorizontalScrollControllerGroup;
   ColumnData<T> sortColumn;
   SortDirection sortDirection;
@@ -578,6 +685,26 @@ class _TableState<T> extends State<_Table<T>> {
     sortColumn = widget.sortColumn;
     sortDirection = widget.sortDirection;
     scrollController = ScrollController();
+  }
+
+  @override
+  void didUpdateWidget(_Table oldWidget) {
+    super.didUpdateWidget(oldWidget);
+
+    cancel();
+
+    // Detect selection changes but only care about scrollIntoView.
+    addAutoDisposeListener(oldWidget.selectionNotifier, () {
+      setState(() {
+        final selection = oldWidget.selectionNotifier.value;
+        if (selection.scrollIntoView) {
+          final newPosition = selection.displayRow * defaultRowHeight;
+          // TODO(terry): Optimize selecting row that's in the viewport
+          //              otherwise we jumpTo.
+          scrollController.jumpTo(newPosition);
+        }
+      });
+    });
   }
 
   @override
--- a/packages/devtools_app/lib/src/flutter/table.dart
+++ b/packages/devtools_app/lib/src/flutter/table.dart
@@ -157,10 +157,11 @@ class FlatTableState<T> extends State<FlatTable<T>>
 }
 
 class Selection<T> {
-  Selection(
-      {this.node,
-      this.nodeIndex,
-      this.scrollIntoView = false,});
+  Selection({
+    this.node,
+    this.nodeIndex,
+    this.scrollIntoView = false,
+  });
 
   Selection.empty()
       : node = null,
@@ -279,10 +280,7 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     addAutoDisposeListener(selectionNotifier, () {
       setState(() {
         final node = selectionNotifier.value.node;
-        final parent = node.parent;
-        if (parent != null) {
-          expandParents(parent);
-        }
+        expandParents(node?.parent);
       });
     });
 
@@ -299,9 +297,9 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
   void _initData() {
     dataRoots = List.from(widget.dataRoots);
     sortData(widget.sortColumn, widget.sortDirection);
-    selectionNotifier = widget.selectionNotifier == null
-        ? SelectionNotifier<Selection<T>>(Selection<T>.empty())
-        : widget.selectionNotifier;
+
+    selectionNotifier = widget.selectionNotifier ??
+        SelectionNotifier<Selection<T>>(Selection<T>.empty());
   }
 
   @override
@@ -492,6 +490,7 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
         nodeIndex: 0,
       );
     }
+
     assert(selectionNotifier.value.node ==
         items[selectionNotifier.value.nodeIndex]);
 
@@ -670,6 +669,7 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
           final newPos = selectedDisplayRow * defaultRowHeight;
           // TODO(terry): Optimize selecting row, if row's visible in
           //              the viewport just select otherwise jumpTo row.
+          // TODO(terry): Should animate too for a nicer interaction.
           scrollController.jumpTo(newPos);
         }
       });
@@ -677,9 +677,13 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
   }
 
   /// Return the number of visible rows above the selected node.
+  // TODO(terry): Must refactory should be T not dynamic but that requires hanlding
+  //              for both Table and TreeTable.
   int selectionRowNumber(dynamic selectedNode) {
     var scanNode = selectedNode;
-    var parent = scanNode.parent;
+    var parent = scanNode?.parent;
+
+    assert(parent != null);
 
     var totalVisibleRowsAboveNode = 0;
     while (!scanNode.isRoot) {
--- a/packages/devtools_app/lib/src/flutter/table.dart
+++ b/packages/devtools_app/lib/src/flutter/table.dart
@@ -338,6 +338,7 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
   void _toggleNode(T node) {
     if (!node.isExpandable) {
       node.leaf();
+      _updateItems();
       return;
     }
 
@@ -425,6 +426,7 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
   ) {
     Widget rowForNode(T node) {
       final isNodeSelected = selectionNotifier.value.node == node;
+      node.index = index;
       return TableRow<T>(
         key: widget.keyFactory(node),
         linkedScrollControllerGroup: linkedScrollControllerGroup,
@@ -582,6 +584,7 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
       selectionNotifier.value = Selection(
         node: newSelectedNode,
         nodeIndex: newSelectedNodeIndex,
+        scrollIntoView: true,
       );
     });
   }
@@ -655,21 +658,12 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
       setState(() {
         final selection = oldWidget.selectionNotifier.value;
         if (selection.scrollIntoView) {
-          final selectedDisplayRow = selectionRowNumber(selection.node);
-
-          // Compute total rows above immediate parent.
-          var parentNode = selection.node.parent;
-          if (parentNode.index != -1) {
-            parentNode = parentNode.parent;
-          }
-          while (parentNode.index != -1) {
-            parentNode = parentNode.parent;
-          }
-
-          final newPos = selectedDisplayRow * defaultRowHeight;
+          final int selectedDisplayRow = selection.node.index;
           // TODO(terry): Optimize selecting row, if row's visible in
           //              the viewport just select otherwise jumpTo row.
-          // TODO(terry): Should animate too for a nicer interaction.
+          final newPos = selectedDisplayRow * defaultRowHeight;
+
+          // TODO(terry): Should animate factor out _moveSelection to reuse here.
           scrollController.jumpTo(newPos);
         }
       });
--- a/packages/devtools_app/lib/src/flutter/table.dart
+++ b/packages/devtools_app/lib/src/flutter/table.dart
@@ -259,8 +259,10 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
   }
 
   void expandParents(T parent) {
-    if (parent?.parent?.index != -1) {
-      expandParents(parent?.parent);
+    if (parent == null) return;
+
+    if (parent.parent?.index != -1) {
+      expandParents(parent.parent);
     }
 
     if (parent != null && !parent.isExpanded) {
--- a/packages/devtools_app/lib/src/flutter/table.dart
+++ b/packages/devtools_app/lib/src/flutter/table.dart
@@ -160,19 +160,16 @@ class Selection<T> {
   Selection(
       {this.node,
       this.nodeIndex,
-      this.scrollIntoView = false,
-      this.displayRow});
+      this.scrollIntoView = false,});
 
   Selection.empty()
       : node = null,
         nodeIndex = null,
-        scrollIntoView = false,
-        displayRow = null;
+        scrollIntoView = false;
 
   final T node;
   final int nodeIndex;
   final bool scrollIntoView;
-  int displayRow;
 }
 
 class SelectionNotifier<T> extends ValueNotifier<T> {
@@ -261,6 +258,16 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     _focusNode = FocusNode();
   }
 
+  void expandParents(T parent) {
+    if (parent?.parent?.index != -1) {
+      expandParents(parent?.parent);
+    }
+
+    if (parent != null && !parent.isExpanded) {
+      _toggleNode(parent);
+    }
+  }
+
   @override
   void didUpdateWidget(TreeTable oldWidget) {
     super.didUpdateWidget(oldWidget);
@@ -268,13 +275,11 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     cancel();
 
     addAutoDisposeListener(selectionNotifier, () {
-      selectionNotifier.value.displayRow = _selectionRowNumber();
-
       setState(() {
         final node = selectionNotifier.value.node;
         final parent = node.parent;
-        if (!parent.isExpanded) {
-          _toggleNode(parent);
+        if (parent != null) {
+          expandParents(parent);
         }
       });
     });
@@ -287,55 +292,6 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     rootsExpanded =
         List.generate(dataRoots.length, (index) => dataRoots[index].isExpanded);
     _updateItems();
-  }
-
-  /// Return the row number of the selected node.  This row number is the visual
-  /// row number including all rows above.
-  int _selectionRowNumber() {
-    final selectedNode = selectionNotifier.value.node;
-
-    final parents = List<T>(selectedNode.level);
-    T scanNode = selectedNode;
-    while (!scanNode.isRoot) {
-      parents[scanNode.parent.level] = scanNode.parent;
-      scanNode = scanNode.parent;
-    }
-
-    var displayRows = -1;
-    for (final T data in dataRoots) {
-      displayRows++;
-      final parentLevel = data.level;
-      if (parents[parentLevel] == data) {
-        // Found the parentage match of the selectedNode, drill down children
-        // of the selectedNode.
-        for (T child in parents[parentLevel].children) {
-          displayRows++;
-          if (child == selectedNode) {
-            return displayRows;
-          }
-        }
-      } else {
-        displayRows += _expandedRows(data);
-      }
-    }
-
-    return displayRows;
-  }
-
-  /// Return total rows displayed in node expanded children and
-  /// any of their expanded nested children.
-  int _expandedRows(T node) {
-    var displayRows = 0;
-
-    if (node.isExpanded) {
-      if (node.children.isNotEmpty) {
-        for (final child in node.children) {
-          displayRows += 1 + _expandedRows(child);
-        }
-      }
-    }
-
-    return displayRows;
   }
 
   void _initData() {
@@ -698,13 +654,60 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
       setState(() {
         final selection = oldWidget.selectionNotifier.value;
         if (selection.scrollIntoView) {
-          final newPosition = selection.displayRow * defaultRowHeight;
-          // TODO(terry): Optimize selecting row that's in the viewport
-          //              otherwise we jumpTo.
-          scrollController.jumpTo(newPosition);
+          final selectedDisplayRow = selectionRowNumber(selection.node);
+
+          // Compute total rows above immediate parent.
+          var parentNode = selection.node.parent;
+          if (parentNode.index != -1) {
+            parentNode = parentNode.parent;
+          }
+          while (parentNode.index != -1) {
+            parentNode = parentNode.parent;
+          }
+
+          final newPos = selectedDisplayRow * defaultRowHeight;
+          // TODO(terry): Optimize selecting row, if row's visible in
+          //              the viewport just select otherwise jumpTo row.
+          scrollController.jumpTo(newPos);
         }
       });
     });
+  }
+
+  /// Return the number of visible rows above the selected node.
+  int selectionRowNumber(dynamic selectedNode) {
+    var scanNode = selectedNode;
+    var parent = scanNode.parent;
+
+    var totalVisibleRowsAboveNode = 0;
+    while (!scanNode.isRoot) {
+      final rowsAbove = parent.children.indexWhere((node) {
+        return scanNode == node;
+      });
+
+      if (rowsAbove > 0) {
+        // Add parent row to the count.
+        totalVisibleRowsAboveNode += rowsAbove + (parent.index >= 0 ? 1 : 0);
+      }
+
+      // Check all scanNode's parent siblings above current scanNode.
+      for (final sibling in parent.children) {
+        if (sibling == scanNode) break;
+
+        // Any parent siblings above expanded? Count those rows too.
+        if (sibling.isExpanded) {
+          // Count of a parent node's children and parent, if expanded,
+          // above selected node.
+          totalVisibleRowsAboveNode += sibling.children.length + 1;
+        }
+      }
+
+      scanNode = parent;
+      parent = scanNode.parent;
+    }
+
+    // Return zero based row.
+    return totalVisibleRowsAboveNode - 1;
   }
 
   @override
--- a/packages/devtools_app/lib/src/flutter/table.dart
+++ b/packages/devtools_app/lib/src/flutter/table.dart
@@ -163,11 +163,6 @@ class Selection<T> {
     this.scrollIntoView = false,
   });
 
-  Selection.empty()
-      : node = null,
-        nodeIndex = null,
-        scrollIntoView = false;
-
   final T node;
   final int nodeIndex;
   final bool scrollIntoView;
@@ -227,7 +222,7 @@ class TreeTable<T extends TreeNode<T>> extends StatefulWidget {
 
   final SortDirection sortDirection;
 
-  final SelectionNotifier<Selection<T>> selectionNotifier;
+  final ValueNotifier<Selection<T>> selectionNotifier;
 
   @override
   TreeTableState<T> createState() => TreeTableState<T>();
@@ -245,7 +240,7 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
   List<bool> rootsExpanded;
   FocusNode _focusNode;
 
-  SelectionNotifier<Selection<T>> selectionNotifier;
+  ValueNotifier<Selection<T>> selectionNotifier;
 
   FocusNode get focusNode => _focusNode;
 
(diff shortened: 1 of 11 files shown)


packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -494,7 +596,8 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
         newSelectedNodeIndex = max(selectedNodeIndex - 1, 0);
         break;
       case ScrollKind.parent:
-        newSelectedNodeIndex = max(items.indexOf(selectedNode.parent), 0);
+        newSelectedNodeIndex =
+            max(items.indexOf(selectionValue.node.parent), 0);
```
gmoothart, line 600: respect `scrollIntoView` in this method. I think you should either default it to true, or else you have to set it to true whenever the selection changes.

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -581,6 +688,26 @@ class _TableState<T> extends State<_Table<T>> {
...
+    super.didUpdateWidget(oldWidget);
+
+    cancel();
+
+    // Detect selection changes but only care about scrollIntoView.
+    addAutoDisposeListener(oldWidget.selectionNotifier, () {
+      setState(() {
+        final selection = oldWidget.selectionNotifier.value;
+        if (selection.scrollIntoView) {
+          final newPosition = selection.displayRow * defaultRowHeight;
```
gmoothart, line 702: can you re-use the code in _moveSelection?

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -155,6 +156,29 @@ class FlatTableState<T> extends State<FlatTable<T>>
...
+  Selection.empty()
+      : node = null,
+        nodeIndex = null,
+        scrollIntoView = false,
+        displayRow = null;
+
+  final T node;
+  final int nodeIndex;
+  final bool scrollIntoView;
+  int displayRow;
```
gmoothart, line 175: how is displayRow different from nodeIndex?

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -110,10 +112,41 @@ class FlameChartState extends State<FlameChart> with AutoDisposeMixin {
...
+
+            closeAutoCompleteOverlay();
+
+            controller.selectTheSearch = false;
+            controller.search = '';
+          } else {
+            // No exact match, return top matches.
+            var matches = matchNames(controller.search);
+            matches.sort();
+            matches = matches.sublist(0, min(topMatches, matches.length));
```
kenzieschmoll, line 129: you could combine these three calls:
```
final matches = matchNames(controller.search)
  ..sort()
  ..subList(0, min(topMatches, matches.length));
```

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -110,10 +112,41 @@ class FlameChartState extends State<FlameChart> with AutoDisposeMixin {
...
+
+    addAutoDisposeListener(controller.searchAutoCompleteNotifier, () {
+      setState(() {
+        if (autoCompleteOverlay == null) {
+          autoCompleteOverlay = createAutoCompleteOverlay(controller);
+          Overlay.of(context).insert(autoCompleteOverlay);
+        } else {
+          closeAutoCompleteOverlay();
+          autoCompleteOverlay = createAutoCompleteOverlay(controller);
+          Overlay.of(context).insert(autoCompleteOverlay);
```
kenzieschmoll, line 149: this code is duplicated in memory_heap_tree view. Consolidate somehow.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -153,6 +191,12 @@ class FlameChartState extends State<FlameChart> with AutoDisposeMixin {
   }
 }
 
+/// Key in callbacks map.
```
kenzieschmoll, line 194: Key in callbacks map for...? Add more descriptive comment. Same below for findNodeKey.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -240,19 +285,50 @@ class FlameChartRenderObject extends RenderBox {
     markNeedsPaint();
   }
 
+  List<String> _exposeMatchNames(String searchName) {
+    return matchNodeNames(_sizes.root.children, searchName);
+  }
+
+  /// Look for the node with a particular name (depth first traversal).
+  List<String> matchNodeNames(Map<String, Node> children, String searchName) {
```
kenzieschmoll, line 293: does Node extends TreeNode? If so, consider adding a `depthFirstTraversal` method to trees.dart (similar to `breadthFirstTraversal`), and then you could add a param `depthFirst` to `firstChildWithCondition` and search by the appropriate traversal type based on the value of `depthFirst`. If you need multiple matches, create a method `childrenWithCondition` in trees.dart. 

It would be nice to keep tree code in that file, as it is well tested and easy to add more tests for.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -240,19 +285,50 @@ class FlameChartRenderObject extends RenderBox {
...
+    final matches = <String>[];
+    final matchName = searchName.toLowerCase();
+    for (var child in children.entries) {
+      final node = child.value;
+      if (node.children.isEmpty) {
+        final lcNodeName = node.name.toLowerCase();
+        if (!lcNodeName.endsWith('.dart') && lcNodeName.startsWith(matchName)) {
```
kenzieschmoll, line 300: replace startsWith with contains? Here and below

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -240,19 +285,50 @@ class FlameChartRenderObject extends RenderBox {
...
+          matches.add(node.name);
+        }
+      } else {
+        final lcNodeName = node.name.toLowerCase();
+        if (!lcNodeName.endsWith('.dart') && lcNodeName.startsWith(matchName)) {
+          matches.add(node.name);
+        }
```
kenzieschmoll, line 307: this code is repeated. you could rewrite this as:
```
final node = child.value;
final lcNodeName = node.name.toLowerCase();
if (!lcNodeName.endsWith('.dart') && lcNodeName.startsWith(matchName)) {
  matches.add(node.name);
}
if (node.children.isNotEmpty) {
  // child matches code...
}
```

packages/devtools_app/lib/src/memory/flutter/memory_utils.dart
```diff
@@ -0,0 +1,60 @@
+//_memorySearchFieldKey
```
kenzieschmoll, line 1: remove comment

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -261,20 +258,28 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     _focusNode = FocusNode();
   }
 
+  void expandParents(T parent) {
+    if (parent?.parent?.index != -1) {
```
jacob314, line 262: nit: 
add
if (parent == null) return;
at the top of this method? This seems like it infinite loops currently unless I'm missing something which I probably am.

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -261,20 +258,28 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
...
 
     addAutoDisposeListener(selectionNotifier, () {
-      selectionNotifier.value.displayRow = _selectionRowNumber();
-
       setState(() {
         final node = selectionNotifier.value.node;
         final parent = node.parent;
-        if (!parent.isExpanded) {
-          _toggleNode(parent);
+        if (parent != null) {
```
jacob314, line 281: all this code could just be
expandParents(node.parent);
if the null check was added within the method.

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -699,15 +655,62 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
...
         final selection = oldWidget.selectionNotifier.value;
         if (selection.scrollIntoView) {
-          final newPosition = selection.displayRow * defaultRowHeight;
-          // TODO(terry): Optimize selecting row that's in the viewport
-          //              otherwise we jumpTo.
-          scrollController.jumpTo(newPosition);
+          final selectedDisplayRow = selectionRowNumber(selection.node);
+
+          // Compute total rows above immediate parent.
+          var parentNode = selection.node.parent;
```
jacob314, line 661: parentNode doesn't seem to be used? I'm missing where we are computing the total rows about the immediate parent.

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -699,15 +655,62 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
...
+          // TODO(terry): Optimize selecting row, if row's visible in
+          //              the viewport just select otherwise jumpTo row.
+          scrollController.jumpTo(newPos);
         }
       });
     });
   }
 
+  /// Return the number of visible rows above the selected node.
+  int selectionRowNumber(dynamic selectedNode) {
```
jacob314, line 679: why is selectedNode type `dynamic`?

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -536,17 +590,52 @@ class MemoryController extends DisposableController
...
+  void addAllToNode(
+      Reference root, Map<String, List<HeapGraphClassLive>> allItems) {
+    allItems.forEach((libraryName, classes) {
+      LibraryReference libReference = root.children.singleWhere((library) {
+        return libraryName == library.name;
+      }, orElse: () => null);
+
+      // Library not found add to list of children.
+      libReference ??= LibraryReference(this, libraryName, classes);
+      root.addChild(libReference);
```
jacob314, line 602: seems like we are calling addChild again even if the libReference already exists?

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -536,17 +590,52 @@ class MemoryController extends DisposableController
     return libraryRoot;
   }
 
+  void addAllToNode(
+      Reference root, Map<String, List<HeapGraphClassLive>> allItems) {
+    allItems.forEach((libraryName, classes) {
+      LibraryReference libReference = root.children.singleWhere((library) {
```
jacob314, line 596: this would be O(n^2) in terms of the #of libraries. We should probably add a set of all libraries so that this is O(n). That could matter for large projects with >4000 libraries.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -536,17 +590,52 @@ class MemoryController extends DisposableController
...
+  bool _gcing = false;
+  bool get isGcing => _gcing;
+
   Future<void> gc() async {
+    _gcing = true;
     await serviceManager.service.getAllocationProfile(
       _isolateId,
       gc: true,
     );
+    _gcing = false;
```
jacob314, line 638: put _gcing in a finally block so we never get stuck thinking we are gcing.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
```diff
@@ -62,8 +68,19 @@ Map<LibraryClass, Predefined> predefinedClasses = {
   predefinedString: const Predefined('String', true),
   predefinedList: const Predefined('List', false),
   predefinedMap: const Predefined('Map', false),
+  predefinedHashMap: const Predefined('HashMap', false),
 };
 
+bool isMap(HeapGraphClassLive live) =>
```
jacob314, line 74: can we check whether the class implements the Map interface or does that not make sense due to whether we have sufficient information about keys and values?

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
```diff
@@ -62,8 +68,19 @@ Map<LibraryClass, Predefined> predefinedClasses = {
...
+    live.fullQualifiedName == predefinedMap ||
+    live.fullQualifiedName == predefinedHashMap;
+
+bool isHashMap(HeapGraphClassLive live) =>
+    live.fullQualifiedName == predefinedHashMap;
+
+bool isList(HeapGraphClassLive live) =>
```
jacob314, line 81: maybe rename to make clear that this is the built in list class only?

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
```diff
@@ -169,24 +182,46 @@ HeapGraph convertHeapGraph(
...
+  final externals = List<HeapGraphExternalLive>(snapshotExternals.length);
+
+  // Construct all external objects and link to its live element.
+  for (int index = 0; index < snapshotExternals.length; index++) {
+    final snapshotObject = snapshotExternals[index];
+    final liveElement = elements[snapshotObject.object];
+    externals[index] = HeapGraphExternalLive(snapshotObject, liveElement);
+  }
+
+  // TODO(terry): Remove debug code.
```
jacob314, line 197: remove :)

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
```diff
@@ -222,6 +260,11 @@ class HeapGraph {
   /// Group all instances by all classes - with applied filters.
   final Map<String, List<HeapGraphElementLive>> groupByClass = {};
 
+  /// Group all filtered out instances by all classes (removed from groupByClass).
+  final Map<String, List<HeapGraphElementLive>> filteredElements = {};
```
jacob314, line 264: it could be handy to comment on what the String key for each of this maps is.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -42,6 +42,73 @@ enum SnapshotStatus {
...
+///     Extact Image:         ^Image$
+String buildRegExs(Map<WildcardMatch, List<String>> matchingCriteria) {
+  final resultRegEx = StringBuffer();
+  matchingCriteria.forEach((key, value) {
+    if (value.isNotEmpty) {
+      final name = value;
+      String regEx;
+      switch (key) {
+        case WildcardMatch.exact:
+          regEx = '^${name.join("\$|^")}\$';
```
jacob314, line 88: Add a TODO that this approach will blow up if the class names contain chars that impact regexes. E.g. searching for the class "$FOO" will cause problems.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -537,6 +620,97 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
     );
   }
 
+  bool _classMatcher(HeapGraphClassLive liveClass) {
+    final regExp = RegExp(knownClassesRegExs);
+    return regExp.allMatches(liveClass.name).isNotEmpty;
+  }
+
+  List<ExternalReference> _collect(List dupSearch) {
```
jacob314, line 628: specify the type of List that dupSearch is.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -537,6 +620,97 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
...
+
+          final liveExternal = external.liveExternal;
+          final size = liveExternal.externalProperty.externalSize;
+          final liveElement = liveExternal.live;
+          final HeapGraphClassLive liveClass = liveElement.theClass;
+          if (_classMatcher(liveClass)) {
+            externalsToAnalyze.add(external);
+          }
+
+          // TODO(terrry): Remove debug code.
```
jacob314, line 652: is this debug code or did you remove the debug code?

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -537,6 +620,97 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
...
+      final liveElement = externalRef.liveExternal.live;
+      final HeapGraphClassLive liveClass = liveElement.theClass;
+      print(liveClass.name);
+      if (liveClass.name == 'ImageCache') {
+        print("STOP");
+      }
+    }
+  }
+
+  // TODO(terry): Remove this debug code.
```
jacob314, line 687: fine to leave in this debug code as long as the method name is prefixed with the word debug and the method isn't called in production code.
that matches the sorts of patterns used by flutter reasonably well.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -749,67 +914,119 @@ class MemorySnapshotTableState extends State<MemorySnapshotTable>
...
     });
 
+    addAutoDisposeListener(controller.selectTheSearchNotifier, () {
+      setState(() {
+        _trySelectItem();
+      });
+    });
+
     addAutoDisposeListener(controller.searchNotifier, () {
       setState(() {
```
jacob314, line 924: nit: add the setState calls to _trySelectItem when the state changes rather than wrapping both calls to _trySelectItem with setState.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -710,6 +744,138 @@ class MemorySnapshotTableState extends State<MemorySnapshotTable>
...
+          // Use top 10 matches:
+          matches.sort();
+          controller.searchAutoComplete.value =
+              matches.sublist(0, min(topMatches, matches.length));
+        }
+      });
+    });
+  }
+
+  String matchSearch(Reference ref, String matchString) {
```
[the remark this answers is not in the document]
jacob314: I'd also be happy with a null string to indicate no match. With null safety, users would then have to check for null and it wouldn't be surprising that null indicates no match.

packages/devtools_app/lib/src/memory/flutter/memory_screen.dart
```diff
@@ -362,7 +362,16 @@ class MemoryBodyState extends State<MemoryBody> with AutoDisposeMixin {
     // TODO(terry): TBD real implementation needed.
   }
 
-  void _gc() {
-    // TODO(terry): Implementation needed.
+  Future<void> _gc() async {
+    // TODO(terry): Record GC in analytics.
+    try {
+      print("GC Starting");
```
jacob314, line 368: remove prints.

packages/devtools_app/lib/src/memory/flutter/memory_snapshot_models.dart
```diff
@@ -110,6 +132,8 @@ class Reference extends TreeNode<Reference> {
 
   final bool isLibrary;
 
+  final bool isExternals;
```
jacob314, line 135: maybe rename isExternals to something more verbose or document why isExternals and isExternal are different.

packages/devtools_app/lib/src/memory/flutter/memory_snapshot_models.dart
```diff
@@ -223,7 +275,21 @@ class ObjectReference extends Reference {
     MemoryController controller,
     int index,
     this.instance,
-  ) : super.object(controller, 'Instance $index');
+  ) : super.object(
+          controller,
+          'Instance $index',
+/*
```
jacob314, line 281: remove commented out code?

packages/devtools_app/lib/src/memory/flutter/memory_snapshot_models.dart
```diff
@@ -343,27 +409,73 @@ class ObjectFieldReference extends FieldReference {
...
-        );
+    var objFields = instanceToFieldNodes(
+      controller,
+      objectFieldReference.instance,
+    );
+
+    if (objFields.isEmpty) {
+      final className =
+          (objectFieldReference.instance.theClass as HeapGraphClassLive).name;
+      if (className.endsWith('Array') || className.endsWith('List')) {
```
jacob314, line 438: this seems really risky. Can we move this to a helper that ideally checks whether the class implements List?

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -155,6 +156,29 @@ class FlatTableState<T> extends State<FlatTable<T>>
...
+        scrollIntoView = false,
+        displayRow = null;
+
+  final T node;
+  final int nodeIndex;
+  final bool scrollIntoView;
+  int displayRow;
+}
+
+class SelectionNotifier<T> extends ValueNotifier<T> {
```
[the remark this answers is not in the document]
jacob314: typedefs don't support types yet but I agree that we should avoid creating ValueNotifier subclasses unless there is a good reason to (e.g. limitations tracking changes to a List.

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -699,15 +655,62 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
       setState(() {
         final selection = oldWidget.selectionNotifier.value;
         if (selection.scrollIntoView) {
-          final newPosition = selection.displayRow * defaultRowHeight;
-          // TODO(terry): Optimize selecting row that's in the viewport
-          //              otherwise we jumpTo.
-          scrollController.jumpTo(newPosition);
+          final selectedDisplayRow = selectionRowNumber(selection.node);
```
gmoothart, line 658: If you can't just do `final selectedDisplayRow = selection.nodeIndex;`, can you please add a comment explaining why? nodeIndex is supposed to be the true index of that row in the table.

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -238,6 +265,20 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
   void didUpdateWidget(TreeTable oldWidget) {
     super.didUpdateWidget(oldWidget);
 
+    cancel();
+
+    addAutoDisposeListener(selectionNotifier, () {
+      selectionNotifier.value.displayRow = _selectionRowNumber();
+
+      setState(() {
+        final node = selectionNotifier.value.node;
```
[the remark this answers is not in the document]
terrylucas: Only the node could be null not the value (see _initData).  But I did add node?.parent.

        final node = selectionNotifier.value.node;
        final parent = node?.parent;

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -248,9 +289,61 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     _updateItems();
   }
 
+  /// Return the row number of the selected node.  This row number is the visual
+  /// row number including all rows above.
+  int _selectionRowNumber() {
+    final selectedNode = selectionNotifier.value.node;
+
+    final parents = List<T>(selectedNode.level);
```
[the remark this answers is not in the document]
terrylucas: This has changed significantly but,

    var parent = scanNode?.parent;
    assert(parent != null);

is needed.

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -155,6 +156,29 @@ class FlatTableState<T> extends State<FlatTable<T>>
   }
 }
 
+class Selection<T> {
+  Selection(
+      {this.node,
+      this.nodeIndex,
+      this.scrollIntoView = false,
+      this.displayRow});
```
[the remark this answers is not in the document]
terrylucas: displayRow was eliminated.

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -238,6 +265,20 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
...
 
+    cancel();
+
+    addAutoDisposeListener(selectionNotifier, () {
+      selectionNotifier.value.displayRow = _selectionRowNumber();
+
+      setState(() {
+        final node = selectionNotifier.value.node;
+        final parent = node.parent;
+        if (!parent.isExpanded) {
```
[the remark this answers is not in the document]
terrylucas: Done.

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -248,9 +289,61 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
...
+      }
+    }
+
+    return displayRows;
+  }
+
   void _initData() {
     dataRoots = List.from(widget.dataRoots);
     sortData(widget.sortColumn, widget.sortDirection);
+    selectionNotifier = widget.selectionNotifier == null
```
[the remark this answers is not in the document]
terrylucas: Done.

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -581,6 +688,26 @@ class _TableState<T> extends State<_Table<T>> {
...
+
+    // Detect selection changes but only care about scrollIntoView.
+    addAutoDisposeListener(oldWidget.selectionNotifier, () {
+      setState(() {
+        final selection = oldWidget.selectionNotifier.value;
+        if (selection.scrollIntoView) {
+          final newPosition = selection.displayRow * defaultRowHeight;
+          // TODO(terry): Optimize selecting row that's in the viewport
+          //              otherwise we jumpTo.
+          scrollController.jumpTo(newPosition);
```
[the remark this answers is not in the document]
terrylucas: Add a todo but good idea.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -343,6 +346,11 @@ class MemoryController extends DisposableController
 
   String get search => _searchNotifier.value;
 
+  final searchAutoComplete = ValueNotifier<List<String>>([]);
```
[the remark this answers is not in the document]
terrylucas: I think the model is to not have automatic setters and getters on ValueNotifiers but to use

searchAutoComplete.value = 

so we now that we're changing a ValueNotifier. Plus its a final.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -89,11 +91,23 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
...
+    addAutoDisposeListener(controller.searchAutoCompleteNotifier, () {
+      setState(() {
+        if (autoCompleteOverlay == null) {
+          autoCompleteOverlay = createAutoCompleteOverlay(controller);
+          Overlay.of(context).insert(autoCompleteOverlay);
+        } else {
+          closeAutoCompleteOverlay();
+
+          autoCompleteOverlay = createAutoCompleteOverlay(controller);
+          Overlay.of(context).insert(autoCompleteOverlay);
```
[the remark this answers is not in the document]
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -364,20 +390,28 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
...
+      child: createSearchField(),
+      focusNode: rawKeyboardFocusNode,
+      onKey: (RawKeyEvent event) {
+        if (event is RawKeyDownEvent) {
+          if (event.logicalKey.keyId == LogicalKeyboardKey.escape.keyId) {
+            // ESCAPE key pressed clear search TextField.
+            clearSearchField();
+          } else if (event.logicalKey.keyId == LogicalKeyboardKey.enter.keyId) {
+            // Find exact match in autocomplete list - use that as our search value.
+            for (final autoEntry in controller.searchAutoComplete.value) {
```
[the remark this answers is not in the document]
terrylucas: We don't know if its an exact match could be prefix of 5 things if enter pressed then the first item in the list is used or the selected item in the list (when up/down arrow supported).

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -710,6 +744,138 @@ class MemorySnapshotTableState extends State<MemorySnapshotTable>
         controller.computeAllLibraries(true, true);
       });
     });
+
+    addAutoDisposeListener(controller.searchAutoCompleteNotifier, () {
```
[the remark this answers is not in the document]
terrylucas: That's nice, thanks for telling me.  I'll fix it.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -710,6 +744,138 @@ class MemorySnapshotTableState extends State<MemorySnapshotTable>
...
+          controller.searchAutoComplete.value =
+              matches.sublist(0, min(topMatches, matches.length));
+        }
+      });
+    });
+  }
+
+  String matchSearch(Reference ref, String matchString) {
+    final knownName = ref.name.toLowerCase();
+    if (knownName.startsWith(matchString.toLowerCase())) {
```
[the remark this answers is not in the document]
terrylucas: I like startsWith by I think contains is probably what most people expect. Done.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -364,20 +390,28 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
...
+              if (controller.search.toLowerCase() == autoEntry.toLowerCase()) {
+                searchTextFieldController.clear();
+                closeAutoCompleteOverlay();
+                controller.selectTheSearch = true;
+                controller.search = autoEntry;
               }
-            },
-          )
-        : const SizedBox();
+            }
```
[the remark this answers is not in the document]
terrylucas: I'll add the todo but there's a bigger problem two table with two selections.  How to tell which table selection has focus?  I think we'll need to have better focus handling, in general, in DevTools.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -710,6 +744,138 @@ class MemorySnapshotTableState extends State<MemorySnapshotTable>
...
+
+          switch (controller.groupingBy.value) {
+            case MemoryController.groupByLibrary:
+              for (final reference in controller.groupByTreeTable.dataRoots) {
+                if (reference.isLibrary) {
+                  final LibraryReference libraryReference = reference;
+                  final match = matchSearch(libraryReference, searchingValue);
+                  if (match.isNotEmpty) {
+                    matches.add(match);
+                  }
```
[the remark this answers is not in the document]
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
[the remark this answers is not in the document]
terrylucas: returns null if no match.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -710,6 +744,138 @@ class MemorySnapshotTableState extends State<MemorySnapshotTable>
...
+                }
+              }
+              break;
+            case MemoryController.groupByInstance:
+              // TODO(terry): TBD
+              break;
+          }
+
+          // Use top 10 matches:
```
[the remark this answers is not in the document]
terrylucas: Done.

```
      // Remove duplicates and sort the matches.
      final normalizedMatches = matches.toSet().toList()..sort();
      // Use the top 10 matches:
      controller.searchAutoComplete.value = normalizedMatches.sublist(

```

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -710,6 +744,138 @@ class MemorySnapshotTableState extends State<MemorySnapshotTable>
...
+        for (final reference in controller.groupByTreeTable.dataRoots) {
+          if (reference.isLibrary) {
+            final LibraryReference libraryReference = reference;
+            if (libraryReference.name == searchingValue) {
+              controller.selectionNotifier.value = Selection(
+                node: libraryReference,
+                nodeIndex: libraryReference.index,
+                scrollIntoView: true,
+              );
+              controller.searchAutoComplete.value = [];
```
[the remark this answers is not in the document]
terrylucas: I added the helper to the controller instead.

```
  void clearSearchAutoComplete() {
    searchAutoComplete.value = [];
  }
```

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -710,6 +744,138 @@ class MemorySnapshotTableState extends State<MemorySnapshotTable>
...
+          if (reference.isClass) {
+            final ClassReference classReference = reference;
+            if (classReference.name == searchingValue) {
+              controller.selectionNotifier.value = Selection(
+                node: classReference,
+                nodeIndex: classReference.index,
+                scrollIntoView: true,
+              );
+              controller.searchAutoComplete.value = [];
+              return true;
```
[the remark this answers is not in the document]
terrylucas: Done.

```
  bool _selectInTree(Reference reference, search) {
    if (reference.name == search) {
      controller.selectionNotifier.value = Selection(
        node: reference,
        nodeIndex: reference.index,
        scrollIntoView: true,
      );
      controller.clearSearchAutoComplete();
      return true;
    }
    return false;
  }
```

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
```diff
@@ -710,6 +744,138 @@ class MemorySnapshotTableState extends State<MemorySnapshotTable>
...
+
+  String matchSearch(Reference ref, String matchString) {
+    final knownName = ref.name.toLowerCase();
+    if (knownName.startsWith(matchString.toLowerCase())) {
+      return ref.name;
+    }
+    return '';
+  }
+
+  /// This finds and selects an exact match.
```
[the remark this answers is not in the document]
terrylucas: Done.
```
  /// This finds and selects an exact match in the tree.
  /// Returns `true` if [searchingValue] is found in the tree.
```

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
```diff
@@ -110,10 +112,41 @@ class FlameChartState extends State<FlameChart> with AutoDisposeMixin {
...
     addAutoDisposeListener(controller.searchNotifier, () {
       setState(() {
-        final searchingValue = controller.search;
-        if (searchingValue.isNotEmpty) {
-          final Node node = findNode(searchingValue);
-          selectNode(node);
+        if (controller.search.isNotEmpty) {
+          if (controller.selectTheSearch) {
+            // Select the node.
+            final Node node = findNode(controller.search);
```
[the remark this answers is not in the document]
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
```diff
@@ -330,7 +333,7 @@ class MemoryController extends DisposableController
 
   ValueListenable<String> get groupingByNotifier => groupingBy;
 
-  bool clearSearch = false;
+  bool selectTheSearch = false;
```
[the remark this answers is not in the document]
terrylucas: It was changed to a ValueNotifier but the name is still there added:

```
  /// Search is very dynamic, with auto-complete or programmatic searching,
  /// setting the value to true will fire off searching through a snapshot.
```

packages/devtools_app/lib/src/memory/flutter/memory_utils.dart
```diff
@@ -0,0 +1,60 @@
...
+
+import 'package:flutter/material.dart';
+
+import 'memory_controller.dart';
+
+/// Top 10 matches to display in auto-complete overlay.
+const topMatches = 10;
```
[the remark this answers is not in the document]
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: Not quite because matches is used for end index, but can use ..sort()

```
        var matches = matchNames(controller.search)..sort();
        // No exact match, return top matches.
        matches = matches.sublist(0, min(topMatchesLimit, matches.length));
```

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: Good catch.

```
addAutoDisposeListener(controller.searchAutoCompleteNotifier, () {
    setState(autoCompleteOverlaySetState(controller, context));
});


/// Helper setState callback when searchAutoCompleteNotifier changes, usage:
/// 
///     addAutoDisposeListener(controller.searchAutoCompleteNotifier, () {
///      setState(autoCompleteOverlaySetState(controller, context));
///     });
VoidCallback autoCompleteOverlaySetState(
  MemoryController controller,
  BuildContext context,
) {
  return () {
    if (autoCompleteOverlay != null) {
      closeAutoCompleteOverlay();
    }

    autoCompleteOverlay = createAutoCompleteOverlay(controller);
    Overlay.of(context).insert(autoCompleteOverlay);
  };
}
```

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: Done.
```
/// Definitions of exposed callback methods stored in callback Map the key
/// is the function name (String) and the value a callback function signature.

/// matchNames callback key.
const matchNamesKey = 'matchNames';

/// matchNames callback signature.
typedef MatchNamesFunction = List<String> Function(String);
```

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: No, these are nodes in the heat map.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: I've changed it to contains but I still prefer startswith especially since I know the classNames auto-complete shows more pertinent information earlier.  May want a setting for auto-complete that defaults to either start or contains.

packages/devtools_app/lib/src/memory/flutter/memory_heatmap.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_utils.dart
terrylucas: Done and added copyright.

packages/devtools_app/lib/src/flutter/table.dart
terrylucas: Re-wrote this some point after your review.  Should be fixed.

packages/devtools_app/lib/src/flutter/table.dart
terrylucas: Right, done.

packages/devtools_app/lib/src/flutter/table.dart
terrylucas: I'm adding a TODO that I need to fix immediately after committing this PR.  The selectionNotifier handling needs to be better refactored between _Table and TreeTable (TreeNode).  This works today because this code is only called when selection.scrollIntoView is true and that's only for TreeTable's.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
terrylucas: Yes, I fixed surprised I never saw the problem.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
terrylucas: Yes I'll add a TODO to eliminate the Map and have a Set of libraries.

packages/devtools_app/lib/src/memory/flutter/memory_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
terrylucas: Added TODO to investigate that to make sure we handle properly.  I'm not sure.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
terrylucas: done for all list, map, and hashmap.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
terrylucas: gone.

packages/devtools_app/lib/src/memory/flutter/memory_graph_model.dart
terrylucas: Good idea, done.

```
  /// Group all classes by libraries (key is library, value are classes).
  /// This is the entire set of objects (no filter applied).
  final Map<String, Set<HeapGraphClassLive>> rawGroupByLibrary = {};

  /// Group all classes by libraries (key is library, value is classes).
  /// Filtering out objects that match a given filter. This is always a
  /// subset of rawGroupByLibrary.
  final Map<String, Set<HeapGraphClassLive>> groupByLibrary = {};

  /// Group all instances by class (key is class name, value are class
  /// instances).  This is the entire set of objects (no filter applied).
  final Map<String, Set<HeapGraphElementLive>> rawGroupByClass = {};

  /// Group all instances by class (key is class name, value are class
  /// instances).  Filtering out objects that match a given filter. This
  /// is always a subset of rawGroupByClass.
  final Map<String, Set<HeapGraphElementLive>> groupByClass = {};

  /// Group of instances by filtered out classes (key is class name, value
  /// are instances). These are the instances not in groupByClass, together
  /// filteredElements and groupByClass are equivalent to rawGroupByClass.
  final Map<String, Set<HeapGraphElementLive>> filteredElements = {};

  /// Group of libraries by filtered out classes (key is library name, value
  /// are classes). These are the libraries not in groupByLibrary, together
  /// filteredLibraries and groupByLibrary are equivalent to rawGroupByLibrary.
  final Map<String, Set<HeapGraphClassLive>> filteredLibraries = {};
```

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
terrylucas: Done, good catch (I never use $ to start an identifier).

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
terrylucas: This code was eliminated.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
terrylucas: It was removed.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_screen.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_heap_tree_view.dart
terrylucas: I think this is what you want.  If not I think this is better.

```
      if (_trySelectItem()) {
        setState(() {
          closeAutoCompleteOverlay();
        });
      }
```

packages/devtools_app/lib/src/memory/flutter/memory_snapshot_models.dart
terrylucas:
```
  /// The External heap (contains all external heap items).
  final bool isExternals;

  /// An External heap item.
  final bool isExternal;
```

packages/devtools_app/lib/src/memory/flutter/memory_snapshot_models.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_snapshot_models.dart
terrylucas: This is old code I forgot to remove!  Now its gone.

terrylucas:
PTAL a few more things to clean up but close to committing.

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -248,9 +289,61 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     _updateItems();
   }
 
+  /// Return the row number of the selected node.  This row number is the visual
+  /// row number including all rows above.
+  int _selectionRowNumber() {
```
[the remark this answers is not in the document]
terrylucas: It is nodeIndex and no searching is needed just proper _updateItems

packages/devtools_app/lib/src/flutter/table.dart
terrylucas: there is no difference nodeIndex is used.

packages/devtools_app/lib/src/flutter/table.dart
terrylucas: done always true.

packages/devtools_app/lib/src/flutter/table.dart
terrylucas: I have a todo to support animation and I've added to factor that out of _moveSelection.

packages/devtools_app/lib/src/flutter/table.dart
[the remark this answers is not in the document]
terrylucas: Not sure what I'm to do here?

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -248,9 +289,61 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     _updateItems();
   }
 
+  /// Return the row number of the selected node.  This row number is the visual
```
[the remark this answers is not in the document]
terrylucas: I think this code was removed no longer exist.

packages/devtools_app/lib/src/flutter/table.dart
terrylucas: Code was cleaned up and removed.

packages/devtools_app/lib/src/flutter/table.dart
terrylucas: Its just a node index nothing more.

terrylucas:
PTAL I think I've fixed the issues.

packages/devtools_app/lib/src/flutter/table.dart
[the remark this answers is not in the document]
jacob314: I would remove this class and just use ValueNotifier. Typedefs won't work for this case and this class isn't providing value.,

packages/devtools_app/lib/src/flutter/table.dart
```diff
@@ -234,10 +258,34 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
...
+    if (parent != null && !parent.isExpanded) {
+      _toggleNode(parent);
+    }
+  }
+
   @override
   void didUpdateWidget(TreeTable oldWidget) {
     super.didUpdateWidget(oldWidget);
 
+    cancel();
```
jacob314, line 278: you should only update if the oldWidget is different from the new widget.

packages/devtools_app/lib/src/memory/flutter/memory_analyzer.dart
```diff
@@ -400,3 +587,84 @@ class _AnalysisFieldValueColumn extends ColumnData<AnalysisField> {
...
+/// Callback function when an ObjectReference's class name and fields all match.
+/// Parameters:
+///   className that matched
+///   fields that all matched
+///   value of the matched objectReference
+///
+typedef CompletedFunction = void Function(
+    String className, List<String> fields, dynamic value);
+
+class ObjectMatcher {
```
jacob314, line 653: nit: is this ImageObjectMatcher not ObjectMatcher?

packages/devtools_app/lib/src/memory/flutter/memory_analyzer.dart
```diff
@@ -400,3 +587,84 @@ class _AnalysisFieldValueColumn extends ColumnData<AnalysisField> {
...
+  bool findFieldMatch(ClassFields classFields, dynamic value) {
+    bool matched = false;
+
+    final className = classFields._fields.elementAt(0);
+    final listOfFieldsToMatch = _findClassMatch(className);
+
+    if (listOfFieldsToMatch != null) {
+      for (final fieldsToMatch in listOfFieldsToMatch) {
+        final fieldsSize = fieldsToMatch.length;
+        if (fieldsSize == classFields._fields.length - 1) {
```
jacob314, line 682: requiring all fields seems fragile. For example, we would miss a subclass that added a field and we would miss if an extra field was added to the class. Can we match if all fields are present or using a subclass relationship rather than requiring an exact match?

packages/devtools_app/lib/src/memory/flutter/memory_analyzer.dart
```diff
@@ -166,107 +165,295 @@ void imageAnalysis(
...
+
+  return null;
+}
+
+// Flag to monitor field/object drill in.
+String monitorDrillIn = '';
+
+void monitor(String msg) {
+  // TODO(terry): Debug code enable to view field drill in.
+  // if (monitorDrillIn.isNotEmpty) print("->$monitorDrillIn:$msg");
```
jacob314: maybe add a private top level variable to toggle on and off this debug logging if you find it is useful to debug this ui.

packages/devtools_app/lib/src/memory/flutter/memory_analyzer.dart
```diff
@@ -166,107 +165,295 @@ void imageAnalysis(
...
+    imageCacheNode.addChild(liveNode);
+
+    return imageCacheNode;
+  }
+
+  return null;
+}
+
+// Flag to monitor field/object drill in.
+String monitorDrillIn = '';
```
jacob314: to follow the flutter scheme consider also prefixing this field with the word _debug
to make it clear it is debugging only.

packages/devtools_app/lib/src/flutter/table.dart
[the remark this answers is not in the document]
terrylucas: Replace with just ValueNotifier.

packages/devtools_app/lib/src/flutter/table.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/flutter/memory_analyzer.dart
terrylucas: I've added a todo it should be loosey goosey so we don't have to rev every time the framework or user's code changes (eventually they can define analysis too).

```
  // TODO(terry): Change to be less strict.  Look for subclass or parentage
  //              relationships.  If a new field or subclass is added we can
  //              still find what we're looking for.  Maybe even consider the
  //              the type we're looking for - best to be loosey goosey so
  //              we're not brittle as the Framework or any code changes.

```

jacob314:
lgtm_dog2

packages/devtools_app/lib/src/memory/flutter/memory_analyzer.dart
terrylucas: Added.

packages/devtools_app/lib/src/memory/flutter/memory_analyzer.dart
terrylucas: Done.

Issue #1753, opened 2020-03-20
Devtools page in lib model write by flutter, can't select and copy text


yk3372:
For example：
lab:
![image](https://user-images.githubusercontent.com/1514899/77131006-fa951600-6a94-11ea-978a-21948f9dc959.png)
before:
![image](https://user-images.githubusercontent.com/1514899/77131030-0e407c80-6a95-11ea-90a9-7364cb75c3ec.png)

I'm often use this log details to get JSON strings for parsing classes. However, the left and right entries in lab cannot be selected and copy. 
Whether to consider using Selectable instead of Text, or other better methods, thanks

jacob314:
In addition to using Selectable we might want to add a "copy " link in the top right corner of the detail page to make copying all the json quicker and easier.

jacob314:
Fyi @ditman

ditman:
I think I need to enable the structured output, I was looking only at the console in the Debugger screen.

ditman:
I've just added the "Copy to clipboard" feature to the stdio console in the "Debugger" panel. I'll have a PR for the "Logging" panel next tuesday, most likely.

shashikantdurge:
Not able to select copy/paste anything.

kenzieschmoll:
This functionality won't be available until the next release

bryanhinkson:
Being able to select/copy text would also be really nice in the network tab.  There are times when hitting an API that I want to get the response and format it to make it easier to see if the data I am expecting is there.

ditman:
@bryanhinkson Please, can you create a new issue with the feature request? I think more people of the team would see it if you create a new issue, than if you update an issue that was closed almost a year ago. Thanks!

PR #1988, opened 2020-05-19
[Logging] Use reusable Console widget in Log Details

ditman:
Reuse the Console widget from the Debugger panel in the Logging panel.

* Extract the Console widget and the Control buttons to a separate, common, file.
* Modify the Debugger and Logging panels to use the same widget.
* Tests.

Fixes https://github.com/flutter/devtools/issues/1753 (or mitigates it, since this doesn't add selectability to the details, only the copy to clipboard functionality)

This is what it looks like:

console-in-logging-panel

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -56,7 +56,7 @@ class _DebuggerConsoleState extends State<DebuggerConsole> {
       lines: _lines,
       controls: [
         CopyToClipboardControl(
-          data: disabled ? null : _lines.join('\n'),
+          dataProvider: disabled ? null : () => _lines.join('\n'),
           successMessage: 'Copied $numLines ${pluralize('line', numLines)}.',
           buttonKey: DebuggerConsole.copyToClipboardButtonKey,
         ),
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -9,12 +9,6 @@ import '../../flutter/console.dart';
 import '../../utils.dart';
 import 'debugger_controller.dart';
 
-// TODO(devoncarew): Allow scrolling horizontally as well.
-
-// TODO(devoncarew): Show some small UI indicator when we receive stdout/stderr.
-
-// TODO(devoncarew): Support hyperlinking to stack traces.
-
 /// Display the stdout and stderr output from the process under debug.
 class DebuggerConsole extends StatefulWidget {
   const DebuggerConsole({
--- a/packages/devtools_app/lib/src/debugger/flutter/console.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/console.dart
@@ -5,10 +5,8 @@
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 
-import '../../flutter/common_widgets.dart';
-import '../../flutter/theme.dart';
-import '../../flutter/utils.dart';
-import 'codeview.dart';
+import '../../flutter/console.dart';
+import '../../utils.dart';
 import 'debugger_controller.dart';
 
 // TODO(devoncarew): Allow scrolling horizontally as well.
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -58,133 +50,22 @@ class _DebuggerConsoleState extends State<DebuggerConsole> {
...
-              maxLines: 1,
-            );
-          },
+    final numLines = _lines.length;
+    final disabled = numLines == 0;
+    return Console(
+      lines: _lines,
+      controls: [
+        CopyToClipboardControl(
+          data: disabled ? null : _lines.join('\n'),
```
devoncarew, line 59: This will join the lines each time this widget is rebuilt. I'm not sure if that's a performance consideration or not; the previous code wouldn't join the lines until the copy action was invoked.

packages/devtools_app/lib/src/flutter/console.dart
```diff
@@ -0,0 +1,219 @@
...
+/// If it succeeds, it displays a notification with `successMessage`.
+class CopyToClipboardControl extends StatelessWidget {
+  const CopyToClipboardControl({
+    this.data,
+    this.successMessage = 'Copied to clipboard.',
+    this.tooltip = 'Copy to clipboard',
+    this.buttonKey,
+  });
+
+  final String data;
```
devoncarew, line 202: This could take a data or a data provider - a callback that could calculate the data.

For an active console, we could be concatenating 5k lines (our current console limit) each time the console widget rebuilt.

packages/devtools_app/lib/src/logging/flutter/logging_screen.dart
```diff
@@ -230,38 +232,37 @@ class _LogDetailsState extends State<LogDetails>
...
-    if (log == null) return const SizedBox();
-    if (log.needsComputing) {
-      return const Center(child: CircularProgressIndicator());
+    if (showInspector(log)) {
+      return _buildInspector(context, log);
+    } else {
+      return Stack(
+        children: [
+          _buildSimpleLog(context, log),
+          if (log != null && log.needsComputing) ...[
```
devoncarew, line 246: or

```
if (log != null && log.needsComputing)
  const Center(child: CircularProgressIndicator()),
```

which uses slightly fewer bits of syntax.

packages/devtools_app/lib/src/debugger/flutter/console.dart
devoncarew: We could also have the `DebuggerController.stdio` API provide raw text (a String or StringBuilder) if we thought that would be easier for consumers.

packages/devtools_app/lib/src/logging/flutter/logging_screen.dart
ditman: You're right, I just don't like ifs without braces :P

packages/devtools_app/lib/src/flutter/console.dart
ditman: Good points, I'm going to try and make these computations lazy so they don't impact the main code path.

packages/devtools_app/lib/src/debugger/flutter/console.dart
ditman: Your suggestion of making `data` a `dataProvider` function is going to prevent this issue. About returning a raw text or not, I'm not sure, unless we need to do smarter things in the process of converting this to a string?

devoncarew: Making it lazy should be enough; in the future we may want to look at returning the raw char data to clients, as well as a pre-computed set of lines, so clients that want to operate on the unprocessed data can. Definitely not a today issue.

ditman: Done!

packages/devtools_app/lib/src/logging/flutter/logging_screen.dart
ditman: Done!

packages/devtools_app/lib/src/flutter/utils.dart
```diff
@@ -18,16 +17,18 @@ Future<void> launchUrl(String url, BuildContext context) async {
   }
 }
 
-/// Attempts to copy a bunch of `lines` to the clipboard.
-Future<void> copyToClipboard(List<String> lines, BuildContext context) async {
+/// Attempts to copy a String of `data` to the clipboard.
+///
+/// Shows a `successMessage` [Notification] on the passed in `context`.
+Future<void> copyToClipboard(
+    String data, String successMessage, BuildContext context) async {
```
kenzieschmoll, line 24: nit: trailing comma

packages/devtools_app/lib/src/logging/flutter/logging_screen.dart
```diff
@@ -230,38 +232,37 @@ class _LogDetailsState extends State<LogDetails>
...
-        child: richText,
-      ),
+    final disabled = log?.details == null || log.details.isEmpty;
+    return Console(
+      lines: log?.prettyPrinted?.split('\n'),
+      controls: [
+        CopyToClipboardControl(
+          dataProvider: disabled ? null : () => log?.prettyPrinted,
+          buttonKey: LogDetails.copyToClipboardButtonKey,
+        )
```
kenzieschmoll, line 264: trailing comma

packages/devtools_app/lib/src/flutter/utils.dart
ditman: Hahaha thanks for noticing. I tried to catch 'em all, but I'm not good at pokemon :P

packages/devtools_app/lib/src/logging/flutter/logging_screen.dart
ditman: Done!

packages/devtools_app/lib/src/flutter/utils.dart
ditman: Done!

ditman:
I'll squash+merge this once CI goes green.

PR #1992, opened 2020-05-20
Move querystring values over to URL fragment

### LINKED ISSUE #1990: Support URLs using querystrings without fragments
In the dart:html version of the app, launching DevTools from VS Code (when not using the service - which is the case for environments like Codespaces/GitPod/Containers) would use a URL like this:

http://localhost:9100/?uri=foo&hide=debugger&page=inspector

With the new Flutter version, the querystring goes into the fragment:

http://localhost:9100/#/?uri=foo&hide=debugger&page=inspector

Unfortunately there's a long-standing bug in VS Code with URI encoding (https://github.com/microsoft/vscode/issues/85930) which results in the fragment being badly encoded and the variables are not processed correctly. The issue is not currently assigned to a milestone so it seems likely it won't be fixed prior to the next DevTools release.

I think there are two possibilities:

1. Support the URL without the fragment
2. Support the mis-encoded fragment

An extra benefit of (1) (besides being less hacky) is that it doesn't require a change in Dart-Code (and thus, older versions of Dart-Code will also seamlessly work) so I think we should try to do that first (even if it's by just detecting the querystring and then redirecting to the fragment version).

DanTup:
This fixes #1990, though it doesn't feel particularly elegant. The details of why we need something like this are in #1990 (in short it's to work around a VS Code URL encoding bug that isn't likely to be fixed before we ship the Flutter version of DevTools ☹️).

Another option could be to have a separate HTML page with this JS in it, and launch that from VS Code to keep the index.html clean (though this won't maintain backwards compatibility with older versions of the extension).

### DIFF
--- a/packages/devtools_app/web/index.html
+++ b/packages/devtools_app/web/index.html
@@ -27,8 +27,10 @@
             window.location.href = '/unsupported-browser.html';
         }
 
-        // Handle legacy dart:html app URLs by moving query strings over to
-        // fragments.
+        // Handle URLs that pass all variables directly on a querystring without
+        // the fragment (for ex. VS Code while it has some encoding bugs preventing
+        // building the correct URLs using fragments
+        // https://github.com/microsoft/vscode/issues/85930).
         if (window.location.search && window.location.search.length > 1) {
           // Ensure each component is encoded, because if the URI contains / slashes
           // Flutter will split on them and try to push multiple routes.
--- a/packages/devtools_app/web/index.html
+++ b/packages/devtools_app/web/index.html
@@ -26,6 +26,16 @@
         if (!supportsES6Classes()) {
             window.location.href = '/unsupported-browser.html';
         }
+
+        // Handle legacy dart:html app URLs by moving query strings over to
+        // fragments.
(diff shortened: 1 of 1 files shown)


packages/devtools_app/web/index.html
```diff
@@ -26,6 +26,16 @@
         if (!supportsES6Classes()) {
             window.location.href = '/unsupported-browser.html';
         }
+
+        // Handle legacy dart:html app URLs by moving query strings over to
```
kenzieschmoll, line 30: are legacy dart:html app urls from launch attempts from older vs code plugin versions?

DanTup: Oops, I didn't keep the comment up-to-date. Right now, all VS Code extensions are going to launch with old URLs, as the VS Code bug mentioned above prevents us from building URLs using the fragment (VS Code encodes them badly - a long-standing bug that can't be fixed because it's a breaking change, so needs additional work to fix).

DanTup: Tweaked the comment and added a ref to the Code issue.

Issue #1944, opened 2020-05-08
ensure that the devtools' server use the correct cache header values


devoncarew:
We should ensure that the devtools' server use the correct cache header values - no more than caching for an hour or something. We're seeing some users that have older versions of DevTools in their browser cache even when they have the latest version on disk.

If the shelf server we're using is returning correct cache header values, then we may need to investigate other options (some cache busting? making sure we're not using service workers?).

PR #1039, opened 2019-09-11
Refactor handlers so that we can configure the way that we serve files

DaveShuckerow:

PR #1993, opened 2020-05-20
set a max limit for caching the compiled devtools resources

devoncarew:
- set a max limit for caching the compiled devtools resources served from the devtools server
- fix https://github.com/flutter/devtools/issues/1944

### DIFF
--- a/packages/devtools_server/lib/src/server.dart
+++ b/packages/devtools_server/lib/src/server.dart
@@ -130,10 +130,7 @@ final argParser = ArgParser()
 /// Wraps [serveDevTools] `arguments` parsed, as from the command line.
 ///
 /// For more information on `handler`, see [serveDevTools].
-Future<HttpServer> serveDevToolsWithArgs(
-  List<String> arguments, {
-  shelf.Handler handler,
-}) async {
+Future<HttpServer> serveDevToolsWithArgs(List<String> arguments) async {
   final args = argParser.parse(arguments);
 
   final help = args[argHelp];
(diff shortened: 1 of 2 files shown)


packages/devtools_server/lib/src/server.dart
```diff
@@ -130,10 +130,7 @@ final argParser = ArgParser()
 /// Wraps [serveDevTools] `arguments` parsed, as from the command line.
 ///
 /// For more information on `handler`, see [serveDevTools].
-Future<HttpServer> serveDevToolsWithArgs(
-  List<String> arguments, {
-  shelf.Handler handler,
```
DanTup, line 135: Is `handler` no longer required here? I thought it was added to support other tools (webdev or something interal?) that consume the server package. If so, this may be a breaking change to them.

devoncarew: Ah, I didn't see any uses in this repo, but didn't realize that webdev was relying on it. I'll revert this part of the change.

DanTup: I don't know for sure, it just seemed to ring a bell. Was added in https://github.com/flutter/devtools/pull/1039 but there's not much info so I'm not sure why I thought this (maybe from a discussion on a VC?).

If it is needed, it might be worth having a comment to avoid it being cleaned up.

devoncarew: (reverted this part of the change)

PR #1995, opened 2020-05-21
Add an &embed param to run without top/bottom app bars and change disconnect handling

DanTup:
This is still WIP, but looking for feedback on the approach. I have a generally-working VS Code PoC that embeds individual pages into tabs, though requires some changes to fit in properly.

This PR adds an `&embed=true` flag to the URL (though we could also change this to `/embed` or similar.. if we move to named routes for pages then we could append it, like `/inspector`, `/inspector/embed` or whatever) which will:

1. Hide the top/bottom nav bars
Originally I was going to do this by just skipping `DevToolsScaffold`, however there are other things in there like `Provider` that would break it)

2. Change from going back to the Connection screen to a "Disconnected" message
This looks better in the IDE than having several connection screens appear when you Stop a debug session. I've rigged it up so when a page is in this disconnected state, VS Code will automatically connect them to a new debug session (so if you stop debugging, then start again, all of the open DevTools tabs would show "Disconnected" but then automatically reconnect).

There are other things to think about (like whether we'd continue to spawn the server, and what the best API/method of communication between DevTools and the editor would be) but I think we can do that gradually - this change is relatively small and gives a much better experience in VS Code without significantly changing anything (it still spawns the server, just hosts the pages inside the editor instead of launching a browser).


DanTup:
@helin24 FYI - if you might be looking at embedding in IntelliJ, we should ensure any changes like this make sense for there too.

jacob314:
This approach looks good overall.
My one question is whether we should refine our flow on disconnect a bit both for the embed and regular case.
An alternate option to having a disconnected route or automatically pushing a route to the connect screen would be to just remain on the screen we were on when the app disconnects. The default screen class could show a greyed "disconnected" overlay on disconnect. If the app is not an embed, there would be a link to navigate to the connect route and connect to a different vm service. That approach would also be a little better for the non embed case as you could see the last state devtools was in when an unexpected disconnect event occurred.
I would expect IDEs would typically close embedded windows when apps disconnected for all cases but when the apps disconnected due to a crash in which case it could be helpful for users to not lose the last state they were in before the disconnect. E.g. what was the memory usage when my app crashed due to running out of memory, what was I doing with the inspector when I crashed my app?

DanTup:
> The default screen class could show a greyed "disconnected" overlay on disconnect. If the app is not an embed, there would be a link to navigate to the connect route and connect to a different vm service. That approach would also be a little better for the non embed case as you could see the last state devtools was in when an unexpected disconnect event occurred.

I like this idea, I'll have a go!

> I would expect IDEs would typically close embedded windows when apps disconnected

For how they work in VS Code, it feels much better keeping them open, for ex let's say you'd been debugging on one device and wanted to stop and switch to another:

- Launch a debug session
- Open Timeline
- Open Inspector
- Stop debug session
- Change selected device
- Start new debug session

By keeping the tabs open and visible, they will automatically connect to the new session when you've launched on the new device. If we closed the tabs when you stopped the debug session, you'd need to re-launch each DevTools page, and then arrange them back how you had them on the screen (in VS Code we con't dock these around the place, they behave exactly like normal editor tabs).

I'll try to get the changes above working first anyway, and then I can grab a vid or screenshare to show how it's working for feedback. Thanks!

DanTup:
I added an overlay instead of redirecting to another route, so it now looks like this:

In-browser:

Screenshot 2020-05-27 at 11 33 42

Embedded:

Screenshot 2020-05-27 at 11 38 33

In VS Code, we'll automatically reload the frame to connect to the new session you launch, so it says "Run a new debug session to connect". If the process may differ between editors, we might want to tweak that a bit.

packages/devtools_app/lib/src/flutter/initializer.dart
```diff
@@ -99,19 +107,54 @@ class _InitializerState extends State<Initializer>
...
+    final theme = Theme.of(context);
+    OverlayEntry overlay;
+    overlay = OverlayEntry(
+      builder: (context) => Container(
+        color: const Color.fromRGBO(128, 128, 128, 0.5),
+        child: Center(
+          child: Column(
+            children: [
+              const Spacer(),
+              Text(
```
DanTup, line 133: There's quite a lot of fluff here, but I don't know if there's a terser way to stack the text/button and center within the screen.

packages/devtools_app/lib/src/flutter/initializer.dart
```diff
@@ -99,19 +107,54 @@ class _InitializerState extends State<Initializer>
...
       }
     });
   }
 
+  OverlayEntry _createDisconnectedOverlay() {
+    final theme = Theme.of(context);
+    OverlayEntry overlay;
+    overlay = OverlayEntry(
+      builder: (context) => Container(
+        color: const Color.fromRGBO(128, 128, 128, 0.5),
```
DanTup, line 128: Not sure if hard-coding this colour is ok - it's a mid-grey so it seems to work ok for both themes.

jacob314: Add a TODO. This is fine for now but we should pick a color from the material theme if possible instead.

kenzieschmoll:
Are you close to landing this? If so, I will wait to land https://github.com/flutter/devtools/pull/2052/ and manually apply your changes.

DanTup:
I haven't had chance to look through at test failures yet. Won't be today so feel free to land yours. This change is trivial and I can easily reapply it when rebasing. Thanks for checking!

DanTup:
I've updated the tests slightly for the new functionality, and rebased on top of #2062 to fix the build.

Issue #1381, opened 2019-11-21
Timeline bar chart selection needs refactoring


kenzieschmoll:
Selecting a bar from the bar chart does not trigger any flutter lifecycle methods. The selection callback should not be triggered during paint, and currently the only way to get it to work is to call the callback in a timer.

jacob314:
The right refactoring is the selection should not be triggered from widget build, layout, or paint.
@kenzieschmoll can you note the exception that gets thrown when the callback is executed at the wrong time?

kenzieschmoll:
```

**Comment by kenzieschmoll** [comment]
Using `addPostFrameCallback` fixes the issue, but I'm still not convinced this is the best solution.
```
WidgetsBinding.instance
          .addPostFrameCallback((_) => onSelected(frameIndex));
```

jacob314:
We should still cleanup to detect outside of rendering the frame but with a simple post frame callback it is a little ugly rather than a hack.

jacob314:
![lgtm](https://i.imgur.com/JgNRalI.png)

Issue #1338, opened 2019-11-15
Investigate MPFlutterChart throws NOT_IMPLEMENTED 


terrylucas:
in flutter_flames_chart.dart a dashed line causes a NOT IMPLEMENTED exception for dash path.
This fails in the BOTS doesn't fail locally (could it be a headless Flutter issue?).  Enabling the below:

            // TODO(terry): Below crashed Flutter.
            // ..enableDashedLine(5, 5, 0)

Failure:

`
00:43 +119: /home/travis/build/flutter/devtools/packages/devtools_app/test/flutter/timeline_screen_test.dart: TimelineScreen builds proper content for timeline modes492══╡ EXCEPTION CAUGHT BY RENDERING LIBRARY 
╞═════════════════════════════════════════════════════════

The following UnimplementedError was thrown during paint():
UnimplementedError495496The relevant error-causing widget was:
CustomPaint org-dartlang-app:///packages/mp_chart/mp/chart/chart.dart:110:31498499When the exception was thrown, this was the stack:

package:dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 194:49 throw

package:build_web_compilers/lib/ui/src/ui/path_metrics.dart 72:24 get length
package:path_drawing/src/dash_path.dart 29:51 dashPath
package:mp_chart/mp/core/adapter_android_mp.dart 22:12 convert2DashPath
package:mp_chart/mp/core/render/y_axis_renderer.dart 333:26 renderLimitLines
package:mp_chart/mp/painter/bar_line_chart_painter.dart 296:49 onPaint
package:mp_chart/mp/painter/painter.dart 242:5 paint
package:flutter/src/rendering/custom_paint.dart 531:12 [_paintWithPainter]
package:flutter/src/rendering/custom_paint.dart 572:7 paint
package:flutter/src/rendering/object.dart 2219:7 [_paintWithContext]
package:flutter/src/rendering/object.dart 184:12 paintChild
package:flutter/src/rendering/proxy_box.dart 123:14 paint
package:flutter/src/rendering/object.dart 2219:7 [_paintWithContext]
package:flutter/src/rendering/object.dart 184:12 paintChild
package:flutter/src/rendering/proxy_box.dart 123:14 paint
package:flutter/src/rendering/object.dart 2219:7 

...

[...]

Issue #1030, opened 2019-09-10
Improve legend for the timeline graph


escamoteur:
Hi,

just playing with the devtools a bit. When on the timeline view I find the legend not expressive enough.

![image](https://user-images.githubusercontent.com/16635729/64608040-1757df00-d3ca-11e9-857d-a939c55b9fc5.png)

As you see not all colors are explained. It also would be helpfull to see the limits that change the colors in the top graph also displayed in the time chart of the single frame as vertical lines. Maybe also add the information from which time on the 60fps are no longer possible

Cheers and thanks for this great tools
Thomas

Issue #451, opened 2019-03-20
What makes a frame janky?


kenzieschmoll:
Starting an issue to track this discussion: **what makes a frame janky**?

Right now, we consider a frame to be janky when the sum of UI and GPU durations exceed the target time limit per frame (currently hard coded to ~16ms). We do not give each thread a 16ms threshold because we are also concerned with latency.

**Question**: To determine whether a frame is janky, should we A) _add the durations of UI work and GPU work_ (current implementation), or B) _use the duration of the frame_?

Case 1:
When both the sum of durations and frame duration meet the target, it does not matter which way we determine jank.
![Screen Shot 2019-03-20 at 2 44 37 PM](https://user-images.githubusercontent.com/43759233/54721279-c5bb4f80-4b1e-11e9-90c6-e50cacb87829.png)

Case 2 (problem case):
The sum of durations exceeds the target, but the frame duration does not.
![Screen Shot 2019-03-20 at 2 44 46 PM](https://user-images.githubusercontent.com/43759233/54721286-cd7af400-4b1e-11e9-9a9a-28659e49b0d8.png)

Case 3 (problem case):
The frame duration exceeds the target, but the sum of durations does not.
![Screen Shot 2019-03-20 at 2 44 53 PM](https://user-images.githubusercontent.com/43759233/54721302-db307980-4b1e-11e9-99b0-9a65c0c0072e.png)

kenzieschmoll:
@chinmaygarde @liyuqian @Hixie

liyuqian:
Good question. I think a clear definition would help us and our developers a lot.

My personal understanding is that "jank" is the opposite of "smooth". Latency is another issue. It's kind like when you buy a gaming monitor, you'll check 2 separate specs:

1. Refresh rate, which is about the smoothness. Often monitors have 60fps but 144hz is now quite popular for first-person shooting gamers.
2. Input latency. It could have a 240hz refresh rate but a high latency (say 10ms), or a 60hz refresh rate with low latency (say 1ms).

I think the separation of smoothness and latency also applies to Flutter. It's best to measure, monitor, and optimize them separately.

For the "jank" or smoothness specifically, it's just about how continuous your frames are. If you have two frames that differ from one another significantly in an unexpected way, users will detect such "jank" or non-smoothness.

In my experience, there are 2 ways to have such non-smoothness:
[...]

PR #1999, opened 2020-05-25
Rewrite flutter frames chart.

kenzieschmoll:
- Rewrite chart without the mp_chart dependency.
- Place UI / Raster times side-by-side instead of stacking them. We were previously giving noisy results for UI jank. Both the UI and Raster portion of the frame should be given 16.6 ms before we label the frame as "janky" - cc @liyuqian (related https://github.com/flutter/devtools/issues/451 https://github.com/flutter/devtools/issues/450)
- Color janky frames in a red overlay
- Color selected frame in Flutter yellow selection color (cc @raison00)
- Add chart legend on the right side

Screen Shot 2020-05-25 at 1 45 15 PM

**EDIT**
Screen Shot 2020-05-26 at 2 33 51 PM
Screen Shot 2020-05-26 at 1 12 54 PM

Fixes https://github.com/flutter/devtools/issues/1381
Fixes https://github.com/flutter/devtools/issues/1030
Fixes https://github.com/flutter/devtools/issues/1338

### DIFF
--- a/packages/devtools_app/lib/src/timeline/flutter/flutter_frames_chart.dart
+++ b/packages/devtools_app/lib/src/timeline/flutter/flutter_frames_chart.dart
@@ -20,6 +20,8 @@ class FlutterFramesChart extends StatefulWidget {
     this.displayRefreshRate,
   );
 
+  static const chartLegendKey = Key('Flutter frames chart legend');
+
   final List<TimelineFrame> frames;
 
   final int longestFrameDurationMs;
@@ -35,9 +37,8 @@ class _FlutterFramesChartState extends State<FlutterFramesChart>
   static const maxMsForDisplay = 48.0;
   static const minMsForDisplay = 18.0;
 
-  static const defaultFrameWidth = 32.0;
   static const defaultFrameWidthWithPadding =
-      defaultFrameWidth + densePadding * 2;
+      FlutterFramesChartItem.defaultFrameWidth + densePadding * 2;
 
   static const yAxisUnitsSpace = 48.0;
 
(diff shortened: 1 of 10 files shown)


devoncarew:
Some drive by thoughts:
- I'd make sure that the axis labels on the left align with our frame targets (so `16.7ms` or `17ms` for 60Hz devices)
- instead of changing the forground color for the selected frame, what if we change the background color? That would be for half the horiz. padding to either side, and the full height of the graph. That might be enough of an indication that the frame is selected, without adding another color that we need to then explain on the legend
- the the UI and Raster legend label, is there a terse way we can indicate that these are the two components of the frame time? `UI Frame Time` / `Raster Frame Time`? `UI Frame Portion` / `Raster Frame Portion`? `Frame Time (UI)` / `Frame Time (Raster)`?

kenzieschmoll:
Addressed DBCs. I think this looks better!
Screen Shot 2020-05-26 at 1 12 54 PM

packages/devtools_app/lib/src/timeline/flutter/flutter_frames_chart.dart
```diff
@@ -99,269 +78,402 @@ class _FlutterFramesChartState extends State<FlutterFramesChart>
...
+    @required this.selected,
+    @required this.onSelected,
+    @required this.msPerPx,
+    @required this.availableChartHeight,
+    @required this.displayRefreshRate,
+  });
 
-    // Create a data object with all the data sets - stacked bar.
-    _chartController.data = BarData([]..add(frameDurationsSet));
+  static const defaultFrameWidth = 32.0;
```
terrylucas, line 213: Should a TODO be added to compute this based on pixel density and device (32 might be too small for higher density screens).  I wouldn't fix just a note.

terrylucas: This is the legend size? not bar width just noticed that below.  I guess that should be computed we have this kind of thing everywhere in devtools.

packages/devtools_app/lib/src/timeline/flutter/flutter_frames_chart.dart
```diff
@@ -99,269 +78,402 @@ class _FlutterFramesChartState extends State<FlutterFramesChart>
...
+    return Container(
+      padding: const EdgeInsets.symmetric(horizontal: densePadding),
+      color: selected ? chartAccentColor : null,
+      child: Column(
+        children: [
+          // Dummy child so that the InkWell does not take up the entire column.
+          const Expanded(child: SizedBox()),
+          InkWell(
+            // TODO(kenz): make tooltip to persist if the frame is selected.
+            // TODO(kenz): change color on hover.
```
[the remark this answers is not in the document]
kenzieschmoll: The TODO is to change the color of the bars so that there is some indication of mouse hover.

Issue #1709, opened 2020-03-09
two finger scrolling in the inspector jumps around


devoncarew:
When doing two finger scrolling in the inspector (for the flutter web version of the app), vertical scrolling is very difficult and jumps around (horizontal scrolling seems fine though).

PR #2001, opened 2020-05-26
misc inspector work

devoncarew:
misc inspector work:
- some tweaks to the inspector screen to move a divider line from above the tab controls to just below the tabs
- fix an issue where the user couldn't scroll vertically for long distances - we'd interrupt the scroll with a horizontal scroll (and inadvertently cancel the user's vertical scroll); fix https://github.com/flutter/devtools/issues/1709; our animated scrolls to widgets still work
- remove the custom icons for widget types - just use a consistent 'round circle' icon for widgets

@jacob314

### DIFF
--- a/packages/devtools_app/lib/src/inspector/diagnostics_node.dart
+++ b/packages/devtools_app/lib/src/inspector/diagnostics_node.dart
@@ -598,11 +598,6 @@ class RemoteDiagnosticsNode extends DiagnosticableTree {
   Widget get icon {
     if (isProperty) return null;
 
-//    Widget icon = widget?.icon;
-//    if (icon == null && widgetRuntimeType != null) {
-//      icon ??= iconMaker.fromWidgetName(widgetRuntimeType);
-//    }
-
     return iconMaker.fromWidgetName(widgetRuntimeType);
   }
 
--- a/packages/devtools_app/lib/src/inspector/diagnostics_node.dart
+++ b/packages/devtools_app/lib/src/inspector/diagnostics_node.dart
@@ -597,11 +597,13 @@ class RemoteDiagnosticsNode extends DiagnosticableTree {
 
   Widget get icon {
     if (isProperty) return null;
-    Widget icon = widget?.icon;
-    if (icon == null && widgetRuntimeType != null) {
-      icon ??= iconMaker.fromWidgetName(widgetRuntimeType);
-    }
-    return icon;
+
+//    Widget icon = widget?.icon;
+//    if (icon == null && widgetRuntimeType != null) {
+//      icon ??= iconMaker.fromWidgetName(widgetRuntimeType);
+//    }
+
+    return iconMaker.fromWidgetName(widgetRuntimeType);
   }
 
   /// Returns true if two diagnostic nodes are indistinguishable from
--- a/packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
+++ b/packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
@@ -192,39 +192,32 @@ class _InspectorScreenBodyState extends State<InspectorScreenBody>
 
   Widget _expandCollapseButtons() {
     if (!_expandCollapseSupported) return null;
+
     return Align(
       alignment: Alignment.topRight,
-      child: Container(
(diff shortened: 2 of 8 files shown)


packages/devtools_app/lib/src/inspector/diagnostics_node.dart
```diff
@@ -597,11 +597,13 @@ class RemoteDiagnosticsNode extends DiagnosticableTree {
 
   Widget get icon {
     if (isProperty) return null;
-    Widget icon = widget?.icon;
-    if (icon == null && widgetRuntimeType != null) {
-      icon ??= iconMaker.fromWidgetName(widgetRuntimeType);
-    }
-    return icon;
+
+//    Widget icon = widget?.icon;
```
jacob314, line 601: remove instead of commenting out.

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
```diff
@@ -226,22 +225,6 @@ class _InspectorTreeState extends State<InspectorTree>
     _focusNode.requestFocus();
   }
 
-  void _onScrollYChange() {
```
jacob314, line 229: I agree the existing way this was done should be improved but doesn't removing this mean that we will end up scrolled to a X coordinate that doesn't show any useful content when we scroll to a new Y coordinate?

packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
```diff
@@ -192,39 +192,32 @@ class _InspectorScreenBodyState extends State<InspectorScreenBody>
 
   Widget _expandCollapseButtons() {
     if (!_expandCollapseSupported) return null;
+
     return Align(
       alignment: Alignment.topRight,
-      child: Container(
-        padding: const EdgeInsets.symmetric(horizontal: 6),
-        // Add a semi-transparent background to the
```
jacob314, line 199: does this comment about avoiding interfering with the expand collapse buttons no longer apply?

devoncarew: Yeah, I think that the buttons must have been overlaid over the content at one time? I don't think that's a bad idea, but currently they're above the content, on the same row as the tab buttons.

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
devoncarew: > we will end up scrolled to a X coordinate that doesn't show any useful content when we scroll to a new Y coordinate?

Yes - we will; the user will have to manually scroll around. We could improve things here, but the current behavior of very jerky vertical scrolling is IMO more of a friction for users than having automatic horizontal scrolling.

I could keep this commented out and have a todo? Add a todo, and reference this commit (to find the source later)?

devoncarew:
Updated based on review comments, and adding a screenshot of the tab area:

Screen Shot 2020-05-26 at 2 57 09 PM

packages/devtools_app/lib/src/inspector/diagnostics_node.dart
devoncarew: (done)

packages/devtools_app/lib/src/inspector/flutter/inspector_screen.dart
jacob314: oh now I remember! that was before there was the tab bar with the layout explorer.

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
jacob314: commented out with a todo sounds good. I don't think the right fix will be hard. we just need some logic layered on top of the scroll controller to smooth things out.

jacob314:
lgtm

packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart
devoncarew: Done! Added back in, commented out, with a comment referencing this PR.

Issue #1980, opened 2020-05-18
DevTools Network Page: UX Punch List 2020


raison00:
##
## Fix Known Issue(s):
- [ ] Fix Colors
- [ ] Fix Timing 
- [ ] Fix Scroll & Zoom
##
## Update Overall UX 
- [x] Update Table to match DevTools Standard tables
- [x] (P3) Display timing as mini-flamechart
- [x] (P3) Visualize timestamps:  currently Timing tab is forcing users to do math
- [x] Make table colors accessible and match DevTools pages
- [x] Ensure horizontal scrolling consistency
- [x] Update with unified light theme - many issues
- - [x] white text on white background for tab controller
- - [x] status colors lack contrast
- [x] Disable tab controller horizontal scrolling
- [x] Update Timestamps for relevancy
- - [x] Remove year
- - [x] remove any redundant info
- [x] Distinguish Network Tabs
- - [x] Network: Tracking Network Data
- - [x] Network: Log Network Traffic Update

#### Current ( DevTools 0.2.5 Network Page)
![image](https://user-images.githubusercontent.com/8105802/82237712-f2b5ee80-98ea-11ea-9f1e-e235024ce746.png)

![image](https://user-images.githubusercontent.com/8105802/82237727-fb0e2980-98ea-11ea-9635-9c02e1d4cdb2.png)

![image](https://user-images.githubusercontent.com/8105802/82237752-06615500-98eb-11ea-963d-6bdea703ab43.png)

devoncarew:
Here's the current UI - updated to use our standard devtools table:

<img width="1460" alt="Screen Shot 2020-05-24 at 11 43 46 AM" src="https://user-images.githubusercontent.com/1269969/83567225-204b8c00-a4d6-11ea-9d71-0d1b9ef461c1.png">

PR #2003, opened 2020-05-26
re-write the network page to use our standard table

devoncarew:
- re-write the network page to use our standard table
- work towards https://github.com/flutter/devtools/issues/1980


devoncarew:
Screen Shot 2020-05-24 at 11 43 46 AM

packages/devtools_app/lib/src/network/flutter/network_screen.dart
```diff
@@ -272,3 +193,128 @@ class NetworkScreenBodyState extends State<NetworkScreenBody> {
...
+          statusColumn,
+          durationColumn,
+          timestampColumn,
+        ],
+        data: requests,
+        keyFactory: (HttpRequestData data) => ValueKey<HttpRequestData>(data),
+        onItemSelected: (item) {
+          networkController.selectHttpRequest(item);
+        },
+        autoScrollContent: true,
```
kenzieschmoll, line 229: just want to verify that this is what we want here. We should be scrolling to the bottom?

packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -20,8 +20,17 @@ class NetworkController {
...
+  final _requests = ValueNotifier<HttpRequests>(HttpRequests());
+
+  void selectHttpRequest(HttpRequestData selection) {
+    _selectedHttpRequest.value = selection;
+  }
+
+  final _selectedHttpRequest = ValueNotifier<HttpRequestData>(null);
+
+  ValueListenable<HttpRequestData> get selectedHttpRequest =>
+      _selectedHttpRequest;
```
kenzieschmoll, line 33: switch order of getter and value

packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -20,8 +20,17 @@ class NetworkController {
...
 
   /// Notifies that new HTTP requests have been processed.
-  ValueListenable<HttpRequests> get requestsNotifier => _httpRequestsNotifier;
-  final _httpRequestsNotifier = ValueNotifier<HttpRequests>(HttpRequests());
+  ValueListenable<HttpRequests> get requests => _requests;
+  final _requests = ValueNotifier<HttpRequests>(HttpRequests());
+
+  void selectHttpRequest(HttpRequestData selection) {
+    _selectedHttpRequest.value = selection;
+  }
```
kenzieschmoll, line 28: move functions below variables and getters declared in this file

devoncarew:
I'm having persistent issues with the CI bots here - not sure it's related to these changes.

packages/devtools_app/lib/src/network/flutter/network_screen.dart
devoncarew: Yup! I think so - we want to make sure that the user is aware of new network events as they occur. If they do manually scroll the table away from the end, the table will stop auto-scrolling.

packages/devtools_app/lib/src/network/network_controller.dart
devoncarew: will do

packages/devtools_app/lib/src/network/network_controller.dart
devoncarew: hmm, getter's first I assume?

kenzieschmoll: yep I believe that is the recommended style

```
getter
value
setter
```

devoncarew:
The failing tests were the result of missing an `@teston('vm')` annotation.

PR #2006, opened 2020-05-27
remove the classes from the Libraries and Classes view

devoncarew:
- remove the classes from the Libraries and Classes view

This makes that view too busy. In the future, we should likely have a 'Search' field, with a drop down when you have content in the field, and code completion for the available items.


devoncarew:
(and, given how many libraries there are for large apps, we may want to change the Libraries view from a list to a tree of packages, directories, and files).

Issue #1972, opened 2020-05-15
DevTools Debugger Page: UX Punch List 2020


raison00:
## Update Overall UX 
- - [ ] Match Chrome DevTools key shortcuts where possible
- - [ ] Create consistent search / filter bindings
- - [ ] Audit `ctrl + p ` across linux / mac / windows
- - [ ] Update Filter shortcut with working command (replace `ctrl + p `)
- - [ ] Add key word accelerators
- - [ ] Add ability to set breakpoints quickly
- - [ ] Update breakpoint mapping and indications
- - [ ] Distinguish: Prominent notification of Debugger page State when on other DevTools pages
- - - [ ] Add notification when App is paused but we don't have frame data
- - - [ ] Add badge to Debugger Tab to indicate App is paused
- - - [ ] Add banner to App with a CTA **Continue Button** to resume
- - - [ ] Add animated notification when accessing during paused App state

- - [ ] Add label title to console area for consistency
- - [ ] Adjust pause/ play location UI & UX
- - [ ] Use DevTools standard table
- - [ ] Unify Search & Filter Expectations 
- - [ ] **Update/Add unified color themes**
- - - [ ] Rows need alternate colors 
- - - [ ] Add easy-to-see prominent notification of breakpoint/code content navigation
- - - - [ ] When changing breakpoints, add highlighted row in code area

#### Current ( DevTools 0.2.5 Debugger Page)

![image](https://user-images.githubusercontent.com/8105802/82094111-46d58e80-96b1-11ea-94e3-47ee635e28ce.png)
![image](https://user-images.githubusercontent.com/8105802/82094178-68367a80-96b1-11ea-8e07-ed5d7906eb84.png)

:sparkles:
####  UX Punch List Updates:
![image](https://user-images.githubusercontent.com/8105802/82093922-eba39c00-96b0-11ea-9b69-391ec5151a91.png)
![image](https://user-images.githubusercontent.com/8105802/82093947-f8c08b00-96b0-11ea-995f-d014ba8cefb4.png)

devoncarew:
@raison00, re: 'Adjust pause/ play location UI & UX' above - what did you have in mind?

devoncarew:
Also, generally, thank you for the detailed UX punch-lists! They're great, and will really help apply consistency and polish throughout the UI.

devoncarew:
For 'Add keyboard accelerators', I suggest we use:

- continue - F5
- pause - F6
- step over - F10
- step into - F11
[...]

PR #2008, opened 2020-05-28
add a title to the debugger console area

devoncarew:
- add a title to the debugger console area
- (general work towards https://github.com/flutter/devtools/issues/1972)

Screen Shot 2020-05-27 at 6 06 44 PM

### DIFF
--- a/packages/devtools_app/lib/src/debugger/flutter/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/flutter/codeview.dart
@@ -224,7 +224,35 @@ class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
     return OutlineDecoration(
       child: Column(
         children: [
-          debuggerSectionTitle(theme, text: scriptRef?.uri ?? ' '),
+          debuggerSectionTitle(
+            theme,
+            child: Row(
+              children: [
+                FlatButton(
+                  child: const Icon(
+                    Icons.chevron_left,
+                    size: actionsIconSize,
+                  ),
+                  padding: EdgeInsets.zero,
+                  materialTapTargetSize: MaterialTapTargetSize.shrinkWrap,
+                  onPressed: () => print('back'),
+                ),
+                FlatButton(
+                  child: const Icon(
+                    Icons.chevron_right,
+                    size: actionsIconSize,
+                  ),
+                  padding: EdgeInsets.zero,
+                  materialTapTargetSize: MaterialTapTargetSize.shrinkWrap,
+                  onPressed: () => print('forward'),
+                ),
+                const SizedBox(width: denseSpacing),
+                const BulletSpacer(),
+                const SizedBox(width: denseSpacing),
+                Text(scriptRef?.uri ?? ' ', style: theme.textTheme.subtitle2),
+              ],
+            ),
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/debugger/flutter/console.dart
```diff
@@ -52,20 +55,31 @@ class _DebuggerConsoleState extends State<DebuggerConsole> {
...
-          onPressed: disabled ? null : widget.controller.clearStdio,
-          tooltip: 'Clear console output',
-          buttonKey: DebuggerConsole.clearStdioButtonKey,
-        ),
-      ],
+
+    return OutlineDecoration(
+      child: Column(
+        children: [
+          debuggerSectionTitle(Theme.of(context), text: 'Console'),
```
ditman, line 62: Would it make sense that all Console widgets have a Title?

packages/devtools_app/lib/src/flutter/console.dart
```diff
@@ -104,25 +103,23 @@ class _ConsoleOutputState extends State<_ConsoleOutput> {
     final textStyle =
         theme.textTheme.bodyText2.copyWith(fontFamily: 'RobotoMono');
 
-    return OutlineDecoration(
-      child: Scrollbar(
```
ditman, line 108: I'm not a fan of removing the outline from the console, now everywhere that we use it we need to wrap it in the OutlineDecoration component, but I see how this'd be problematic when laying this out with an external title widget.

devoncarew: I think the presence of an outline is mostly dictated by the parent container and layout rather than something specific to the widget. For instance, for the debugger views on the left side of the page:

Screen Shot 2020-05-28 at 9 06 03 AM

if the area widgets had outlines around them as part of their definition, they'd have double outlines for the places where they abutted.

I guess I'm ok w/ a use of this widget knowing that it had to wrap it in an Outline for best results -

ditman: Haha OK, you convinced me. I wanted to have a completely self-contained widget, but that column looks like a good case for "the container controlling the outline".

packages/devtools_app/lib/src/debugger/flutter/console.dart
ditman: I think it'd be legit to have title in all Consoles. Also, you could make the "debuggerSectionTitle" a "SectionTitle" so you can reuse on top of other widgets elsewhere, so the whole UI gains consistency.

packages/devtools_app/lib/src/debugger/flutter/codeview.dart
```diff
@@ -224,7 +224,35 @@ class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
     return OutlineDecoration(
       child: Column(
         children: [
-          debuggerSectionTitle(theme, text: scriptRef?.uri ?? ' '),
+          debuggerSectionTitle(
+            theme,
+            child: Row(
+              children: [
+                FlatButton(
```
devoncarew, line 231: I'll revert this - this wasn't intended to be committed -

PR #2024, opened 2020-06-01
remove the animation for appbar layout changes

devoncarew:
- remove the animation for appbar layout changes

This fixes the issue where the memory page would deadlock when running on Flutter desktop. Because of where the animation was defined in the widget hierarchy, it would cause the app screens to rebuild often. There may still be a deadlock in the memory page, but if so we no longer seem to hit it.

Most pages only rebuild once now when switching screens. The memory page may rebuild twice (from brief observation). And - not topical for this PR - it seems to take about 300-400ms to build the frame when switching into the Inspector page; that all seems to be spent in layout.

### DIFF
--- a/packages/devtools_app/lib/src/flutter/scaffold.dart
+++ b/packages/devtools_app/lib/src/flutter/scaffold.dart
@@ -81,10 +81,7 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
     with TickerProviderStateMixin {
   /// A tag used for [Hero] widgets to keep the [AppBar] in the same place
   /// across route transitions.
-  static const String _appBarTag = 'DevTools AppBar';
-
-  AnimationController appBarAnimation;
-  CurvedAnimation appBarCurve;
+  static const Object _appBarTag = 'DevTools AppBar';
 
   /// The controller for animating between tabs.
   ///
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/flutter/scaffold.dart
```diff
@@ -81,10 +81,7 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
     with TickerProviderStateMixin {
   /// A tag used for [Hero] widgets to keep the [AppBar] in the same place
   /// across route transitions.
-  static const String _appBarTag = 'DevTools AppBar';
-
-  AnimationController appBarAnimation;
-  CurvedAnimation appBarCurve;
+  static const Object _appBarTag = 'DevTools AppBar';
```
kenzieschmoll, line 84: why did this change from String to Object

packages/devtools_app/lib/src/flutter/scaffold.dart
```diff
@@ -328,31 +306,27 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
...
-        end: const Size.fromHeight(kToolbarHeight + 40.0),
-      ).evaluate(appBarCurve);
-      final animatedAlignment = Tween<Alignment>(
-        begin: Alignment.centerRight,
-        end: Alignment.bottomLeft,
-      ).evaluate(appBarCurve);
+      preferredSize = isNarrow
+          ? const Size.fromHeight(kToolbarHeight + 40.0)
+          : const Size.fromHeight(kToolbarHeight);
+      final animatedAlignment =
```
kenzieschmoll, line 312: rename to `alignment` since it is no longer animated?

packages/devtools_app/lib/src/flutter/scaffold.dart
devoncarew: It's used as an `Object` later on, so I just updated the type to reflect the intend.

packages/devtools_app/lib/src/flutter/scaffold.dart
devoncarew: 👍

PR #2026, opened 2020-06-01
Fix bug in flutter frames chart axis.

kenzieschmoll:


kenzieschmoll:
I added a TODO to consider using more round values. I believe @devoncarew requested on the initial PR that we have the target ms always on the Y axis.

devoncarew:
> I believe @devoncarew requested on the initial PR that we have the target ms always on the Y axis.

From playing with the new chart, my feedback may be addressed just by having that 16.7 / 60 FPS horizontal line on the chart.

I think we'll want to go with either numbers that make sense in terms of the target frame times (i.e., just show 8ms, 16ms, 32ms, ...), or, show round numbers (10ms, 20ms, 30ms, ...).

We may also consider having a fixed y range for the chart - always display 50ms, or 66.4, or 100ms.

Issue #1957, opened 2020-05-13
DevTools Info Page: UX Punch List 2020


raison00:
## Update Overall UX 
- [x] Update borders
- [ ] Update UX for priority and emphasis
- [ ] Ensure unified layout with other pages
- [ ] Update/Add unified color themes
- - [x] Rows need alternate colors
- - [ ] Update selected color from gray  to `#36c6f4`

## VM Flag List
- [ ] Add two lines per entry
- [ ] Add description on separate line
- [x] Include column headers 
- [x] Add Label value to column
- [x] Add Search / Filter for flags
- [ ] Add Value / Percentages for flags

##  Open Questions
- [x] Move content/ consolidate to an observatory specific page?

#### Current ( DevTools 0.2.5 Info Page)
![image](https://user-images.githubusercontent.com/8105802/81869365-c84de500-9528-11ea-84e6-aee9997d3bc3.png)
![image](https://user-images.githubusercontent.com/8105802/81869409-db60b500-9528-11ea-82fb-830b91d7c837.png)

:sparkles:

devoncarew:
I believe that many of the items here will be addressed by https://github.com/flutter/devtools/pull/2027; that will:
- remove the info page
- add a 'device info' dialog, openable by clicking on the device name on the status line
- make the device name on the status line more clearly actionable
- have the vm flags info available from the device dialog, opening in a subsidiary dialog
- convert the flags presentation to use our std table, which should address several of the ux items raised here

devoncarew:
Here's the current state of affairs:

<img width="205" alt="Screen Shot 2020-06-02 at 1 33 06 PM" src="https://user-images.githubusercontent.com/1269969/83567029-cc40a780-a4d5-11ea-9a7a-e5733fbb5bc2.png">

<img width="421" alt="Screen Shot 2020-06-02 at 1 33 17 PM" src="https://user-images.githubusercontent.com/1269969/83567046-d2cf1f00-a4d5-11ea-8d71-7095cc40da2b.png">

<img width="948" alt="Screen Shot 2020-06-02 at 1 33 27 PM" src="https://user-images.githubusercontent.com/1269969/83567055-d82c6980-a4d5-11ea-8b49-ca56d3bf9cba.png">

<img width="943" alt="Screen Shot 2020-06-02 at 1 33 43 PM" src="https://user-images.githubusercontent.com/1269969/83567069-debae100-a4d5-11ea-8b9c-22464675c9c6.png">

[...]

PR #2027, opened 2020-06-01
move the contents of the info page into a device info dialog

devoncarew:
- move the contents of the info page into a device info dialog
- address much of the feedback in https://github.com/flutter/devtools/issues/1957, as well as some existing todos

Screen Shot 2020-05-30 at 4 55 56 PM

Screen Shot 2020-05-30 at 4 58 41 PM


packages/devtools_app/lib/src/flutter/device_dialog.dart
```diff
@@ -0,0 +1,265 @@
...
+
+  List<ColumnData<_DialogFlag>> get columns => [name, description, value];
+
+  @override
+  Widget build(BuildContext context) {
+    return OutlineDecoration(
+      child: FlatTable<_DialogFlag>(
+        columns: columns,
+        data: flags,
+        autoScrollContent: true,
```
kenzieschmoll, line 216: I'm surprised that we would be autoscrolling content to the end. Why do we need to do this?

packages/devtools_app/lib/src/flutter/device_dialog.dart
```diff
@@ -0,0 +1,265 @@
...
+        keyFactory: (_DialogFlag flag) => ValueKey<String>(flag.name),
+        sortColumn: name,
+        sortDirection: SortDirection.ascending,
+        onItemSelected: (_DialogFlag item) => null,
```
kenzieschmoll, line 220: What about just (_) {}

packages/devtools_app/lib/src/info/info_controller.dart
```diff
@@ -18,6 +18,19 @@ class InfoController extends DisposableController
     _listenForFlutterVersionChanges();
   }
 
+  /// Return the FlutterVersion for the current connected app, or `null`
+  /// if the `flutterVersion` service extension does not exist.
+  static Future<FlutterVersion> getFlutterVersion() async {
```
kenzieschmoll, line 23: Avoid get in method name

packages/devtools_app/lib/src/info/info_controller.dart
```diff
@@ -18,6 +18,19 @@ class InfoController extends DisposableController
...
+    final app = serviceManager.connectedApp;
+    if (!(await app.isFlutterApp)) {
```
kenzieschmoll, line 25: You may be able to use isFlutterAppNow depending on when this is called

packages/devtools_app/lib/src/flutter/device_dialog.dart
```diff
@@ -0,0 +1,265 @@
...
+  final ConnectedApp connectedApp;
+  final FlutterVersion flutterVersion;
+
+  @override
+  Widget build(BuildContext context) {
+    const boldText = TextStyle(fontWeight: FontWeight.bold);
+    final theme = Theme.of(context);
+
+    final vm = serviceManager.vm;
+
```
kenzieschmoll, line 36: Can you give an example what we expect the output to be to explain the substring code

packages/devtools_app/lib/src/flutter/device_dialog.dart
```diff
@@ -0,0 +1,265 @@
...
+    var version = vm.version;
+    if (version.contains(' ')) {
+      version = version.substring(0, version.indexOf(' '));
+    }
+
+    final versions = {
```
kenzieschmoll, line 42: Maybe rename since these aren't all versions? deviceInfo?

packages/devtools_app/lib/src/flutter/device_dialog.dart
```diff
@@ -0,0 +1,265 @@
...
+    return FlatButton(
+      onPressed: () {
+        Navigator.of(context, rootNavigator: true).pop('dialog');
+
+        unawaited(showDialog(
+          context: context,
+          builder: (context) => VMFlagsDialog(),
+        ));
+      },
+      child: Text('View VM Flags...'.toUpperCase()),
```
kenzieschmoll, line 93: Why the ellipsis?

packages/devtools_app/lib/src/flutter/device_dialog.dart
```diff
@@ -0,0 +1,265 @@
...
+
+    return AlertDialog(
+      actions: [
+        DialogCloseButton(),
+      ],
+      content: Column(
+        mainAxisSize: MainAxisSize.min,
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          Row(
```
kenzieschmoll, line 161: Since there are only two items you could do MainAxisAlignment.spaceBetween. or out the container In the expanded and align it to the right

packages/devtools_app/lib/src/flutter/device_dialog.dart
devoncarew: Sure (and I'll add as a comment as well):

Here's the full string: `Dart VM version: 2.9.0-13.0.dev (dev) (Fri May 29 15:59:05 2020 +0200) on "macos_x64"`; we just want the portion before the first space.

packages/devtools_app/lib/src/flutter/device_dialog.dart
devoncarew: sure, will rename

packages/devtools_app/lib/src/flutter/device_dialog.dart
devoncarew: To indicate that it opens a separate dialog (instead of just doing something like performing an action).

packages/devtools_app/lib/src/flutter/device_dialog.dart
devoncarew: This is a bit easier for be to read the code - to look at it and know how it'll lay out. Happy to have someone change it in the future however.

packages/devtools_app/lib/src/flutter/device_dialog.dart
devoncarew: We definitely don't good catch. This was a copy paste error.

packages/devtools_app/lib/src/flutter/device_dialog.dart
devoncarew: will shorten

packages/devtools_app/lib/src/info/info_controller.dart
devoncarew: This is delegating to the existing `serviceManager.getFlutterVersion()` call; I'd suggest if we choose to update we do that in a separate PR.

packages/devtools_app/lib/src/info/info_controller.dart
devoncarew: Gotcha; I'm not sure we know yet whether `isFlutterAppNow` has been populated or not at this location.

kenzieschmoll:
lgtm_army

PR #2033, opened 2020-06-03
Implement offline export / imports for Performance page

peterdjlee:
- Implement offline export and import for Performance page regardless of connection to a running app.
Screen Shot 2020-06-03 at 5 44 24 PM


packages/devtools_app/lib/src/performance/flutter/performance_screen.dart
```diff
@@ -131,37 +153,86 @@ class _PerformanceScreenBodyState extends State<PerformanceScreenBody>
...
+                    !controller.cpuProfileData.isEmpty
+                ? _exportPerformance
+                : null,
+            child: const MaterialIconLabel(
+              Icons.file_download,
+              'Export',
+              includeTextWidth: _secondaryControlsMinIncludeTextWidth,
+            ),
+          ),
+        ),
```
kenzieschmoll, line 231: pull this container out into a separate stateless widget class `ExportButton` and put it in common_widgets.dart. Then use an instance of ExportButton in both the performance and timeline screens

packages/devtools_app/lib/src/performance/flutter/performance_screen.dart
```diff
@@ -172,4 +243,25 @@ class _PerformanceScreenBodyState extends State<PerformanceScreenBody>
       recordedObject: 'CPU samples',
     );
   }
+
+  void _exportPerformance() {
+    final exportedFile = controller.exportData();
+
+    Notifications.of(context)
+        .push('Successfully exported $exportedFile to ~/Downloads directory');
```
kenzieschmoll, line 251: add a helper method `successfulExportMessage` to import_export.dart (similar to `attemptingToImportMessage` that takes a file name and returns this string.

use that helper here and in the timeline.

Also add a TODO to find a way to push the notification logic into the export controller.
we structure TODOs as "// TODO(username): todo message"

packages/devtools_app/lib/src/performance/flutter/performance_screen.dart
```diff
@@ -172,4 +243,25 @@ class _PerformanceScreenBodyState extends State<PerformanceScreenBody>
...
+  }
+
+  @override
+  FutureOr<void> processOfflineData(CpuProfileData offlineData) async {
+    if (!offlineMode) await controller.clear();
```
kenzieschmoll, line 256: is the call to clear necessary? I'm confused why we would ever be calling `processOfflineData` unless we are in `offlineMode`?

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -4,13 +4,17 @@
 
 import 'dart:async';
 
+import 'package:devtools_app/src/config_specific/flutter/import_export/import_export.dart';
+import 'package:devtools_app/src/performance/flutter/performance_screen.dart';
```
kenzieschmoll, line 8: relative and sorted imports

packages/devtools_app/lib/src/config_specific/flutter/import_export/import_export.dart
```diff
@@ -4,6 +4,9 @@
 
 import 'dart:convert';
 
+import 'package:devtools_app/src/performance/flutter/performance_screen.dart';
+import 'package:devtools_app/src/profiler/cpu_profile_model.dart';
```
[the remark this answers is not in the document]
peterdjlee: I realized those lines were actually unused imports so I'll take them out

packages/devtools_app/lib/src/performance/flutter/performance_screen.dart
```diff
@@ -55,44 +59,67 @@ class PerformanceScreenBody extends StatefulWidget {
...
   double processingProgress = 0.0;
 
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
     maybePushDebugModePerformanceMessage(context, PerformanceScreen.id);
 
-    final newController = Provider.of<PerformanceController>(context);
+    final newController =
+        Provider.of<PerformanceController>(context, listen: false);
```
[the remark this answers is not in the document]
peterdjlee: It was an attempted fix for a bug but I forgot to take it out after seeing that it doesn't actually fix the bug. Will remove it.

packages/devtools_app/lib/src/performance/flutter/performance_screen.dart
peterdjlee: Should it be named exportButton? Seems like the other stateless widgets are named as camelCase (e.g. stopButton, refreshButton).

jacob314: widget names should not start with a lower case letter. If we have widgets that start with a lower case letter those should be fixed. Likely you are seeing some helper methods that aren't actually widgets.

packages/devtools_app/lib/src/performance/flutter/performance_screen.dart
```diff
@@ -131,37 +153,86 @@ class _PerformanceScreenBodyState extends State<PerformanceScreenBody>
         ),
       ],
     );
+
+    // We put these two items in a stack because the screen's UI needs to be
```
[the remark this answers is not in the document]
kenzieschmoll: we need the screen to be built because there are listeners that are initialized when the screen widget is built. I ran into this issue in the Timeline screen and landed on the stack implementation because if the screen is not built ahead of time, the listeners do not respond to events that happen while the data is being processed.

packages/devtools_app/lib/src/performance/flutter/performance_screen.dart
peterdjlee: Yeah, I think the line is obsolete since this method is never called if it isn't offline mode. I should've realized to remove the lne instead of adding a boolean check.

packages/devtools_app/lib/src/flutter/common_widgets.dart
```diff
@@ -468,6 +467,38 @@ class ToggleButton extends StatelessWidget {
...
+/// Button to export data.
+///
+/// * `enabled`: Whether the button should be enabled.
+/// * `includeTextWidth`: The minimum width the button can be before the text is
+///    omitted.
+/// * `onPressed`: The callback to be called upon pressing the button.
+class ExportButton extends StatelessWidget {
+  const ExportButton({
+    Key key,
+    this.enabled = true,
```
kenzieschmoll, line 479: instead of passing enabled, just pass null to the onPressed parameter when the button should be disabled

PR #2034, opened 2020-06-04
Case Study Memory leak using large network images.

terrylucas:
Consider loading images in another isolate to fix UI stall e.g., downloading images on the UI thread.

However, this is a very simple app showing simple fixes using the imageCache to not explode memory with large images downloaded over the network.


case_study/memory_leaks/images_1/lib/main.dart
```diff
@@ -0,0 +1,268 @@
...
+            // action button
+            icon: const Icon(Icons.add),
+            onPressed: () {
+              _jumpNewPage();
+            },
+          )
+        ],
+        bottom: TabBar(
+            controller: _tabController,
+            tabs: tabs.map((e) => Tab(text: e)).toList()),
```
jacob314, line 102: nit: add trailing comma after toList()

case_study/memory_leaks/images_1/lib/main.dart
```diff
@@ -0,0 +1,268 @@
...
+      ),
+      body: listView(),
+    );
+
+    return scaffold;
+  }
+
+  /// *********************************************************************************
+  /// ***** ListView outside of TabBarView, ListView w/o width/height smaller.    *****
```
jacob314, line 111: Also consider removing the print statements as users won't have the benefit of them in typical apps they are debugging.

terrylucas: I meant to change these to developer.log.

case_study/memory_leaks/images_1/lib/main.dart
terrylucas: Done.

PR #2040, opened 2020-06-04
remove the date from network timestamps

devoncarew:
- remove the date from network timestamps
- work towards https://github.com/flutter/devtools/issues/1980

Screen Shot 2020-06-04 at 4 28 08 PM


kenzieschmoll:
landing this so that it is included in the g3 roll

PR #2041, opened 2020-06-04
misc changes

devoncarew:
misc local changes:
- remove a no-longer-used method
- add a new utility method to display size info in kb (this is used to some work that's not yet ready to land, but that I'd like to get off my disk for fewer merge changes)

### DIFF
--- a/packages/devtools_app/lib/src/utils.dart
+++ b/packages/devtools_app/lib/src/utils.dart
@@ -61,6 +61,8 @@ final NumberFormat _kbPattern = NumberFormat.decimalPattern()
   ..maximumFractionDigits = 0;
 
 String printKb(num bytes) {
+  // We add ((1024/2)-1) to the value before formatting so that a non-zero byte
+  // value doesn't round down to 0.
   return _kbPattern.format((bytes + 511) / 1024);
 }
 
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/utils.dart
```diff
@@ -57,8 +57,15 @@ final NumberFormat nf = NumberFormat.decimalPattern();
 
 String percent2(double d) => '${(d * 100).toStringAsFixed(2)}%';
 
+final NumberFormat _kbPattern = NumberFormat.decimalPattern()
+  ..maximumFractionDigits = 0;
+
+String printKb(num bytes) {
+  return _kbPattern.format((bytes + 511) / 1024);
```
kenzieschmoll, line 64: I understand the `/ 1024` part by why +511?

devoncarew: I think I was trying to adjust this for some rounding in the number formatter, but let me re-run this locally and see what was going on.

devoncarew: ok, the number formatter will round, so it's show 1 byte, as 0kb, 511 bytes as 0kb, and 512 bytes as 1kb. Adding 511 before the division will mean that we don't show a zero kb value when we do have some data.

This `printKb()` method is meant for things that are reasonably large (images on the order of 1-6MB), so we don't need the additional granularity of a decimal point (0.1kb), but I don't want to show 0 when there is in fact some data.

PR #2052, opened 2020-06-08
Delete flutter/ directories that no longer serve a purpose.

kenzieschmoll:
This CL moves all files inside of a `flutter/` dir up one level to the parent dir, and deletes all `flutter/` directories.

We originally placed all flutter files inside of a `flutter/` dir to split the flutter and html builds. Now that the html app is deleted, we no longer need these directories.


jacob314:
This will create a lot of pain for pending CLs. do we have any pending CLs we want to land before this goes in? I've got one but I can deal with the pain if it is the only major CL.

kenzieschmoll:
I didn't see any large CLs under review. Last time I mentioned this, we discussed waiting for Terry's memory CL to land and the table CLs Devon had in progress. The pending PRs under review seem relatively small.

kenzieschmoll:
@jacob314 I want to verify with you that the file move for `rendering_tester` is okay
`devtools_app/test/flutter/rendering_tester.dart → devtools_app/test/support/rendering_tester.dart`

When I moved the contents of `test/flutter` to `test/`, travis was trying to run rendering_tester.dart as a test, which it is not. Looks like you added this as part of the extent_delegate_list_view code.

tool/travis.sh
```diff
@@ -124,9 +124,9 @@ elif [ "$BOT" = "test_ddc" ]; then
     # The flutter tool doesn't support excluding a specific set of targets,
     # so we explicitly provide them.
     if [ "$PLATFORM" = "vm" ]; then
-        flutter test test/*.dart test/{core,fixtures,flutter,support,ui}/
+        flutter test test/*.dart test/{core,fixtures,support,ui}/
```
kenzieschmoll, line 127: @jacob314 Please review the changes in this file

PR #2058, opened 2020-06-09
make the heap data table the default view

devoncarew:
- make the heap data table the default view
- add an outline around the heap data table
- use common constants for the search box sizes

This PR has a few UI related changes, but the main change is that it changes the default view for the heap snapshot from the heat map to the heap data table.

cc @terrylucas

### DIFF
--- a/packages/devtools_app/lib/src/memory/memory_controller.dart
+++ b/packages/devtools_app/lib/src/memory/memory_controller.dart
@@ -52,7 +52,13 @@ class MemoryController extends DisposableController
 
   static const logFilenamePrefix = 'memory_log_';
 
-  bool showHeatMap = true;
+  final _showHeatMap = ValueNotifier<bool>(false);
+
+  ValueListenable<bool> get showHeatMap => _showHeatMap;
+
+  void toggleShowHeatMap(bool value) {
+    _showHeatMap.value = value;
+  }
 
   final snapshots = <Snapshot>[];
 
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -52,7 +52,13 @@ class MemoryController extends DisposableController
 
   static const logFilenamePrefix = 'memory_log_';
 
-  bool showHeatMap = true;
+  final _showHeatMap = ValueNotifier<bool>(false);
+
+  ValueListenable<bool> get showHeatMap => _showHeatMap;
+
+  void toggleShowHeatMap(bool value) {
```
jacob314, line 59: why not just expose showHeatMap as a ValueNotifier if we are fine with users reading and writing the value?

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -339,11 +344,11 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
           children: [
             const Text('Heat Map'),
             Switch(
-              value: controller.showHeatMap,
+              value: controller.showHeatMap.value,
```
jacob314, line 347: I'm not seeing where we are listening for changes to `controller.showHeatMap` to update this UI.

PR #2062, opened 2020-06-10
TBR: Add getSupportedProtocols to VM Service Wrapper + pin webdev

DanTup:
v4.1.0 of vm_service adds a new method (`getSupportedProtocols`). I don't think this is technically a breaking change, however both DevTools and webdev `implement` the `VmServiceInterface` which means they don't compile with that version.

This change:

1. Pins the webdev version activated on Travis back to version 2.5.6 which compiles (it gets vm_service v3)
2. Requires v4.1 of vm_service for DevTools so we can access the new types, and provides an implementation for `getSupportedProtocols` to address the same issue for DevTools

If this goes green, I may land TBR so I can rebase my other PR and ensure it's passing, but we'll want to revert the hard-coded webdev version here as soon as webdev publishes a fix.

Related:

- https://github.com/dart-lang/webdev/issues/1036
- https://github.com/dart-lang/webdev/issues/1037
- https://github.com/dart-lang/sdk/commit/54fb6ce3c07719d72eec0a14c00f905b724873c2

### DIFF
--- a/packages/devtools/CHANGELOG.md
+++ b/packages/devtools/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 0.2.6-dev.2
 * Ship Flutter Web version of DevTools by default
+* Update package:vm_service dependency to ^4.1.0
 
 ## 0.2.5 2020-05-07
 * Persist connected app URI when switching to Flutter web version of DevTools [#1933](https://github.com/flutter/devtools/pull/1933)
--- a/packages/devtools_app/lib/src/vm_service_wrapper.dart
+++ b/packages/devtools_app/lib/src/vm_service_wrapper.dart
@@ -729,6 +729,19 @@ class VmServiceWrapper implements VmService {
   }
 
   @override
+  Future<ProtocolList> getSupportedProtocols() async {
+    if (await isProtocolVersionSupported(
+        supportedVersion: SemanticVersion(major: 3, minor: 35))) {
+      return _trackFuture(
+        'getSupportedProtocols',
+        _vmService.getSupportedProtocols(),
+      );
+    } else {
+      return null;
+    }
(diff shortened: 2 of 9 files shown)


gmpassos:
You need to fix this too:

https://github.com/dart-lang/webdev/blob/master/dwds/pubspec.yaml
https://github.com/dart-lang/webdev/blob/master/dwds/lib/src/services/chrome_proxy_service.dart#L41

DanTup:
Pushed an additional change to add `ID`s to several constructors that now require them.

@gmpassos 

> Add `pub upgrade` to ensure that caches are not hiding any issue with new dependencies

The `pubspec.lock` file isn't committed here, so I believe the `pub get` call will already fetch the latest (allowed) versions from `pub`. I think `pub upgrade` is only required to check for newer versions than are in the lock file.

> You need to fix this too:

Yeah, those are in other repositories and will need fixing there. I'm not entirely sure what the implementation of `getSupportedProtocols` will be for `webdev` so that's probably best done by the webdev developers. Here I'm just trying to get this build back green (the changes here roll back to an older version of webdev (temporarily) to avoid the issue until it's fixed in webdev).

DanTup:
@jacob314 landing this TBR to get the build green - I'll address any feedback in a follow-up PR. The webdev version in travis.sh can be reverted once webdev is fixed, though all the other changes are required (to implement the updated VM Service interface, and provide IDs to all constructors that now mark them as required).

DanTup:
@jacob314 @devoncarew I landed this TBR. The hard-coded webdev version has since been reverted, but the other changes remain (required to implement the new vm_service interface) so still need to be reviewed. Thanks!

DanTup:
Thanks!

PR #2069, opened 2020-06-11
normalize icon usage across pages

devoncarew:
- use the `tune` material icon for settings in-line in a page (as a distinct icon from the gear icon at the application level)
- use the higher def resolution icons for the service extension actions
- make the reload and restart icons larger (to match the size of the other toolbar icons)
- use a material icon for restart (it looks better than the png we currently have for restart)

Screen Shot 2020-06-10 at 9 03 56 PM

Screen Shot 2020-06-10 at 9 03 11 PM

Screen Shot 2020-06-10 at 9 03 16 PM

Screen Shot 2020-06-10 at 9 03 21 PM

Screen Shot 2020-06-10 at 9 03 36 PM

### DIFF
--- a/packages/devtools_app/lib/src/service_extensions.dart
+++ b/packages/devtools_app/lib/src/service_extensions.dart
@@ -79,7 +79,7 @@ class ServiceExtensionDescription<T> {
 final debugAllowBanner = ToggleableServiceExtensionDescription<bool>._(
   extension: 'ext.flutter.debugAllowBanner',
   description: 'Debug Banner',
-  icon: createImageIcon('icons/debug_banner.png'),
+  icon: createImageIcon('icons/debug_banner@2x.png'),
   enabledValue: true,
   disabledValue: false,
   enabledTooltip: 'Hide Debug Banner',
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/service_extensions.dart
```diff
@@ -139,7 +139,7 @@ final profileWidgetBuilds = ToggleableServiceExtensionDescription<bool>._(
 final repaintRainbow = ToggleableServiceExtensionDescription<bool>._(
   extension: 'ext.flutter.repaintRainbow',
   description: 'Repaint Rainbow',
-  icon: createImageIcon('icons/repaint_rainbow.png'),
+  icon: createImageIcon('icons/repaint_rainbow@2x.png'),
```
jacob314, line 142: you don't need to add the 2x. Unless there is a bug in the flutter version, this icon code is designed to automatically use the 2x versions on a high dpi devices like the matching intellij code does.

devoncarew: I don't believe it's working - from local testing, it does look like there's less pixelation when choosing the 2x paths.

I believe that the `foo@wx.png` convention is what IntelliJ uses, and flutter uses something like `.../2.0x/my_icon.png` (i.e., based on directory name).

devoncarew: I opened https://github.com/flutter/devtools/issues/2072 to track addressing this.

PR #2077, opened 2020-06-11
Use default drag_handle icon for splitters.

kenzieschmoll:
Screen Shot 2020-06-11 at 2 54 40 PM
Screen Shot 2020-06-11 at 2 54 55 PM

Fixes TODO.


kenzieschmoll:
Because it is an `Icon` widget, the size is forced to be square, so if we shrink the size, the splitter gets smaller, even though we just want to decrease the space around the icon. We could build the icon manually (essentially breaking down what the Icon widget does for us) and allow width and height to be different.

devoncarew:
> Because it is an `Icon` widget, the size is forced to be square, so if we shrink the size, the splitter gets smaller, even though we just want to decrease the space around the icon. We could build the icon manually (essentially breaking down what the Icon widget does for us) and allow width and height to be different.

I think it's an improvement as is; if we want a more narrow splitter, but to keep the icon image the same size, we could crop the icon a bit.

Issue #1964, opened 2020-05-14
DevTools Timeline Page: UX Punch List 2020


raison00:
## Update Overall UX 
- [x] Enlarge Warning Panel:  Current size lends itself to be unengaged
- [x] ~Create Icon for Navigation~
- [ ] Update borders
- [x] ~UX Distinction: Performance Overlay, Track Widget Builds, Network~
- [x] Scrolling Tabs: use a consistent scroll physics for the TabViewBar widget

- [x] **Update/Add unified color themes**
- - [x] ~Multiple issues with light mode.~
- -  [x] Update page's default font color to default Light mode font color. 
- - [x] Rows need alternate colors: color serves as distinction between threads. (Currently: Light Blue = UI, Dark Blue = Raster)
- - -[x] UI Blues: `#88b1de`
- - -[x] Raster Blues: `#2c5daa`
- -  [x] Jank: `#c3595a`.  `#f5846b`

- - [x] Update selected color from gray to its original color. 
- - [x]  Bracket color and select indicator:`#36c6f4`

## Timeline and Flame Chart
- [x] Add sticky legend
- [ ] Add Mini Map (tracked separately here: https://github.com/flutter/devtools/issues/2094)
- [x] Use DevTools standard table
- [x] ~Make Overflow Menu enabled by default~

#### Current ( DevTools 0.2.5 Timeline Page)

![image](https://user-images.githubusercontent.com/8105802/81957819-f92e2880-95c1-11ea-8313-8fd5432c0c42.png)
![image](https://user-images.githubusercontent.com/8105802/81957936-1c58d800-95c2-11ea-9155-0156b6a03062.png)
![image](https://user-images.githubusercontent.com/8105802/81957979-2bd82100-95c2-11ea-9b79-c19d003ae2e6.png)

:sparkles:
####  UX Punch List Updates:

kenzieschmoll:
Note: some of these issues look like they belong to the Network page instead of the timeline page

raison00:
I've removed a few network items.  thanks for the catch!

kenzieschmoll:
what do you mean by "Create Icon for Navigation"?

raison00:
@kenzieschmoll  I will create some designs to express the navigation and post them here.  I have gone through my notes a few times today to see why this was added to the punch list. It's not really a line item anymore, so I will remove it - but the concept for navigation clarity remains the core of the issue and is based on the UX Distinction: Performance Overlay, Track Widget Builds, Network 
[...]

PR #2079, opened 2020-06-12
Make timeline flame chart section labels sticky

kenzieschmoll:
transparent_stickly_labels
transparent_sticky_labels_dark

- Make labels sticky to the right
- paint alternating colors for section backgrounds (same alt colors we use for tables)
- addresses part of https://github.com/flutter/devtools/issues/1964


jacob314:
Having the labels sticky is great. Discussed offline tweaking them to use horizontal real-estate a more efficiently.

kenzieschmoll:
Updated and ready for review. PTAL @jacob314

jacob314:
High level thoughts:
1. Maybe increase the alpha slightly so that the labels are a little more readable.
2. For blocks with more than one line consider placing the label completely on its own line. That matches what chrome devtools does.
3. maybe add padding so that time zero doesn't overlap the labels?

packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -767,3 +767,27 @@ extension ThemeDataExtension on ThemeData {
...
+  final color = theme.canvasColor;
+  return _modifyColorForIndex(color, index, isDarkTheme: theme.isDarkTheme);
+}
+
+Color alternatingColorForIndex(int index, {@required bool isDarkTheme}) {
+  final color = defaultBackgroundColor;
+  return _modifyColorForIndex(color, index, isDarkTheme: isDarkTheme);
+}
+
+Color _modifyColorForIndex(
```
jacob314, line 783: nit: remove word `modify`?

packages/devtools_app/lib/src/timeline/timeline_flame_chart.dart
```diff
@@ -34,15 +36,17 @@ class TimelineFlameChart extends FlameChart<TimelineData, TimelineEvent> {
...
-    // TODO(kenz): we need to calculate start inset based on the width of the
-    // section labels. We should also set a max, ellipsize, and rely on tooltip
-    // to give the full name in the event that the section name exceeds max.
-    //
-    // Alternatively, we could make the label section a column of it's own that
-    // is resizeable. It would need to link scroll controllers with the list
-    // view holding the flame chart nodes. This would make section labels sticky
-    // to the left as an inherent bonus.
-    return 140.0;
+    const maxStartInset = 300.0;
```
jacob314, line 39: nice. I was suggesting this but looks like you already did it. :)

packages/devtools_app/lib/src/common_widgets.dart
kenzieschmoll: done

kenzieschmoll:
added TODOs for some of your suggestions as some are more involved changes that should be done in a separate PR

PR #2080, opened 2020-06-14
Support --start-paused and retain profile history after disconnect/crash.

terrylucas:
- Flutter application started with --start-paused causes the main isolate to pause.  The debugger detects the pause states and correctly enables the 'Resume' button.  If Resume is pressed that isolate will resume.  Unfortunately, other isolates created (after the main isolate is resumed) will subsequently be created in a pause start state (requiring subsequent isolates to be resumed).  This PR will correctly resume any future created isolates in a pause state.
- Fixed inspector to not throw exceptions when launched (default DevTools) and Flutter was started with --start-paused.  Open [issue](https://github.com/flutter/devtools/issues/2081) to fix the inspector to properly reconnect when Flutter is running again (Resume from debugger).
- Added ability to review collected memory statistics if application is disconnected or crash (e.g., review graph, snapshots, and any analyzes).

### DIFF
--- a/packages/devtools_app/lib/src/connected_app.dart
+++ b/packages/devtools_app/lib/src/connected_app.dart
@@ -4,6 +4,8 @@
 
 import 'dart:async';
 
+import 'package:vm_service/vm_service.dart';
+
 import 'globals.dart';
 import 'service_extensions.dart' as extensions;
 
@@ -27,8 +29,24 @@ class ConnectedApp {
 
   bool _isFlutterApp;
 
-  Future<bool> get isProfileBuild async =>
-      _isProfileBuild ??= await _connectedToProfileBuild();
+  Future<bool> get isProfileBuild async {
+    // If we have one isolate check if paused on start?
+    final isolates = serviceManager.isolateManager.isolates;
+    if (isolates.length == 1) {
+      final isolate = await serviceManager.service.getIsolate(isolates[0].id);
+      if (isolate.pauseEvent.kind == EventKind.kPauseStart) {
+        // Application started with --start-paused, assume profile build is
+        // false - debugging memory. Otherwise, _connectedToProfileBuild
+        // waits forever when paused start.
+        // TODO(terry): Revisit this assumption.
+        _isProfileBuild = false;
+        return _isProfileBuild;
+      }
+    }
+
+    _isProfileBuild ??= await _connectedToProfileBuild();
+    return _isProfileBuild;
+  }
 
   bool get isProfileBuildNow {
     assert(_isProfileBuild != null);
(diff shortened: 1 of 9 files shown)


packages/devtools_app/lib/src/connected_app.dart
```diff
@@ -27,8 +29,24 @@ class ConnectedApp {
...
 
-  Future<bool> get isProfileBuild async =>
-      _isProfileBuild ??= await _connectedToProfileBuild();
+  Future<bool> get isProfileBuild async {
+    // If we have one isolate check if paused on start?
+    final isolates = serviceManager.isolateManager.isolates;
+    if (isolates.length == 1) {
+      final isolate = await serviceManager.service.getIsolate(isolates[0].id);
+      if (isolate.pauseEvent.kind == EventKind.kPauseStart) {
+        // Application started with --start-paused, assume profile build is
```
jacob314, line 38: can you file a bug? This seems like a strange assumption. I would suggest we instead fix the _connectedToProfileBuild method. The current implementation is more fragile than it needs to be. Profile builds are missing support like `eval` so we should use that instead of checking for a specific service extension.

packages/devtools_app/lib/src/initializer.dart
```diff
@@ -144,6 +144,11 @@ class _InitializerState extends State<Initializer>
                   style: theme.textTheme.bodyText2,
                 ),
               const Spacer(),
+              RaisedButton(
+                  onPressed: () {
+                    overlay.remove();
+                  },
+                  child: const Text('Review History')),
```
jacob314, line 151: nit" add trailing comma between `))` for better formatting.

packages/devtools_app/lib/src/connected_app.dart
terrylucas: Filed bug https://github.com/flutter/devtools/issues/2082 and changed to attempt eval if fails assume profile build.

packages/devtools_app/lib/src/initializer.dart
terrylucas: done.

packages/devtools_app/lib/src/inspector/inspector_service.dart
```diff
@@ -555,7 +561,13 @@ class ObjectGroup {
     String methodName,
     Map<String, Object> params,
   ) {
-    return _callServiceExtension('ext.flutter.inspector.$methodName', params);
+    final callMethodName = 'ext.flutter.inspector.$methodName';
+    if (serviceManager.serviceExtensionManager
```
kenzieschmoll, line 565: should this be if _not_ available? like above in `invokeServiceMethodDaemonNoGroup`?

terrylucas: Yes it should.  I caught it by debugging the problem - good eyes.

PR #2084, opened 2020-06-15
Replace heatmap with treemap in memory page

peterdjlee:
- Replace heatmap with treemap in memory page
memory_treemap


packages/devtools_app/lib/src/app.dart
```diff
@@ -459,10 +460,10 @@ List<DevToolsScreen> get defaultScreens => <DevToolsScreen>[
...
-//        createController: () => ExampleController(),
-//        supportsOffline: true,
-//      ),
+      // Uncomment to see a sample implementation of a conditional screen.
+      // DevToolsScreen<ExampleController>(
+      //   const ExampleConditionalScreen(),
+      //   createController: () => ExampleController(),
+      //   supportsOffline: true,
+      // ),
     ];
```
kenzieschmoll, line 468: revert all changes to this file

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -227,9 +227,9 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
       ]);
     } else if (controller.snapshotByLibraryData != null) {
       if (controller.showHeatMap.value) {
-        snapshotDisplay = HeatMapSizeAnalyzer(
+        snapshotDisplay = TreemapSizeAnalyzer(
           child: SizedBox.expand(
```
kenzieschmoll, line 231: is this SizedBox.expand still necessary? Can we just do `snapshotDisplay = MemoryTreemap(controller);`?

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -267,7 +267,8 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
       mainAxisAlignment: MainAxisAlignment.end,
       children: [
         Expanded(child: snapshotDisplay),
-        if (hasDetails) const SizedBox(width: defaultSpacing),
+        if (hasDetails)
+          const SizedBox(width: defaultSpacing),
```
kenzieschmoll, line 271: can this stay on the same line? if not, let's add brackets { }

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -342,7 +343,7 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
         const SizedBox(width: defaultSpacing),
         Row(
           children: [
-            const Text('Heat Map'),
+            const Text('Treemap'),
             Switch(
               value: controller.showHeatMap.value,
```
kenzieschmoll, line 348: rename notifier on controller from `showHeatMap` to `showTreeMap`

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -342,7 +343,7 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
         const SizedBox(width: defaultSpacing),
         Row(
           children: [
-            const Text('Heat Map'),
+            const Text('Treemap'),
```
kenzieschmoll, line 346: use camel case: Treemap --> TreeMap. Here and elsewhere

packages/devtools_app/lib/src/memory/memory_heatmap.dart
```diff
@@ -36,7 +36,6 @@ import 'memory_utils.dart';
 /// Mouse action over, upward, children will further expand its children in
```
kenzieschmoll, line 36: you should be able to delete this file completely

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+
+import '../auto_dispose_mixin.dart';
+import '../treemap/treemap.dart';
+import '../trees.dart';
+
+import 'memory_controller.dart';
+import 'memory_graph_model.dart';
+import 'memory_utils.dart';
+
+class TreemapSizeAnalyzer extends SingleChildRenderObjectWidget {
```
kenzieschmoll, line 21: I'm not sure that you need this widget. Can we just use MemoryTreeMap in place of this widget?

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+  void paint(PaintingContext context, Offset offset) {
+    if (child != null) {
+      context.paintChild(child, offset);
+    }
+  }
+}
+
+class MemoryTreemap extends StatefulWidget {
+  const MemoryTreemap(
+    this.controller,
```
kenzieschmoll, line 44: no need for trailing comma here, as the constructor would fit on one line

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+class RenderFlameChart extends RenderProxyBox {
+  @override
+  void paint(PaintingContext context, Offset offset) {
+    if (child != null) {
+      context.paintChild(child, offset);
+    }
+  }
+}
+
+class MemoryTreemap extends StatefulWidget {
```
kenzieschmoll, line 42: MemoryTreemap --> MemoryTreeMap

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+  Map<String, Function> callbacks = {};
+
+  MemoryController controller;
+
+  Widget snapshotDisplay;
+
+  @override
+  void initState() {
+    super.initState();
+  }
```
kenzieschmoll, line 67: no need to add the overridden method here as it is just calling super

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+
+    addAutoDisposeListener(controller.searchAutoCompleteNotifier, () {
+      setState(autoCompleteOverlaySetState(controller, context));
+    });
+  }
+
+  @override
+  void dispose() {
+    super.dispose();
+  }
```
kenzieschmoll, line 116: no need to override

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+/// selectNode callback name.
+const selectNodeKey = 'selectNode';
+
+/// selectNode callback signature.
+typedef SelectNodeFunction = void Function(TreemapNode);
+
+class InstructionsSize {
+  const InstructionsSize(this.root);
+
+  factory InstructionsSize.fromSnapshop(MemoryController controller) {
```
kenzieschmoll, line 163: typo: .fromSnapshop -->  .fromSnapshot

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+
+    root.byteSize = snapshotGraph.shallowSize + snapshotGraph.externalSize;
+
+    return InstructionsSize(root);
+  }
+
+  final TreemapNode root;
+}
+
+class TreemapNode extends TreeNode<TreemapNode> {
```
kenzieschmoll, line 262: TreemapNode --> TreeMapNode

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+  void addSize(int byteSize) {
+    this.byteSize += byteSize;
+  }
+
+  TreemapNode getChildWithName(String name) {
+    if (childrenMap.containsKey(name)) {
+      return childrenMap[name];
+    } else {
+      return null;
+    }
```
kenzieschmoll, line 284: this is equivalent to `return childrenMap[name]`. If `childrenMap` does not contain a key `name` null will be returned anyway

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+
+  final String name;
+  int byteSize;
+  final Map<String, TreemapNode> childrenMap;
+
+  void addSize(int byteSize) {
+    this.byteSize += byteSize;
+  }
+
+  TreemapNode getChildWithName(String name) {
```
kenzieschmoll, line 279: avoid `get` in method names. Rename to `childWithName`, or better yet, just access the map directly where this method is used (see next comment)

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+const selectNodeKey = 'selectNode';
+
+/// selectNode callback signature.
+typedef SelectNodeFunction = void Function(TreemapNode);
+
+class InstructionsSize {
+  const InstructionsSize(this.root);
+
+  factory InstructionsSize.fromSnapshop(MemoryController controller) {
+    final Map<String, TreemapNode> rootChildren = <String, TreemapNode>{};
```
kenzieschmoll, line 164: remove type from LHS. Type is inferred from declaration on RHS

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+    final TreemapNode root = TreemapNode(
```
kenzieschmoll, line 165: remove type from LHS

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+
+          final symbol = Symbol(
+            name: 'new $className',
+            size: shallowSize,
+            libraryUri: libraryName,
+            className: className,
+          );
+
+          Map<String, TreemapNode> currentChildren = rootChildren;
+          final TreemapNode parentReset = currentParent;
```
kenzieschmoll, line 208: remove type

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+          for (String pathPart in symbol.parts) {
+            currentChildren.putIfAbsent(
+              pathPart,
+              () {
+                final TreemapNode node = TreemapNode(
```
kenzieschmoll, line 213: remove type

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+                    name: pathPart, childrenMap: <String, TreemapNode>{});
```
kenzieschmoll, line 214: trailing comma after last param if the declaration breaks to multiple lines

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+      },
+    );
+
+    root.byteSize = root.childrenMap.values
+        .fold(0, (int current, TreemapNode node) => current + node.byteSize);
+
+    final snapshotGraph = controller.snapshots.last.snapshotGraph;
+    // Add the external heap to the heat map.
+    root.childrenMap.putIfAbsent('External Heap', () {
+      final TreemapNode node = TreemapNode(
```
kenzieschmoll, line 234: remove type

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+          currentParent = parentReset;
+        }
+      },
+    );
+
+    root.byteSize = root.childrenMap.values
+        .fold(0, (int current, TreemapNode node) => current + node.byteSize);
+
+    final snapshotGraph = controller.snapshots.last.snapshotGraph;
+    // Add the external heap to the heat map.
```
kenzieschmoll, line 232: comment: heat --> tree

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+      final TreemapNode node = TreemapNode(
+        name: 'External Heap',
+        childrenMap: <String, TreemapNode>{},
+      );
+      root.addChild(node);
+      node.byteSize = snapshotGraph.externalSize;
+      return node;
+    });
+
+    // Add the filtered libraries/classes to the heat map.
```
kenzieschmoll, line 243: heat --> tree

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+
+    final snapshotGraph = controller.snapshots.last.snapshotGraph;
+    // Add the external heap to the heat map.
+    root.childrenMap.putIfAbsent('External Heap', () {
+      final TreemapNode node = TreemapNode(
+        name: 'External Heap',
+        childrenMap: <String, TreemapNode>{},
+      );
+      root.addChild(node);
+      node.byteSize = snapshotGraph.externalSize;
```
kenzieschmoll, line 239: instead of this add `..byteSize = snapshotGraph.externalSize;` at the end of the TreeMapNode constructor call above

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+    });
+
+    // Add the filtered libraries/classes to the heat map.
+    root.childrenMap.putIfAbsent('All Filtered Libraries', () {
+      final node = TreemapNode(
+        name: 'All Filtered Libraries',
+        childrenMap: <String, TreemapNode>{},
+      );
+      root.addChild(node);
+      node.byteSize = snapshotGraph.shallowSize - root.byteSize;
```
kenzieschmoll, line 250: same comment here about using `..byteSize = ...`

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+  }
+
+  @override
+  void dispose() {
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    sizes = InstructionsSize.fromSnapshop(controller);
```
kenzieschmoll, line 120: this feels expensive to be doing in the build method. Can you put this in didChangeDependencies after we get the controller?

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+    }
+  }
+
+  @override
+  String toString() {
+    return '{name: $name, size: $byteSize}\n';
+  }
+}
+
+class Symbol {
```
kenzieschmoll, line 304: move this class to memory_graph_model.dart

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+/// findNode callback signature.
+typedef FindNodeFunction = TreemapNode Function(String);
+
+/// selectNode callback name.
+const selectNodeKey = 'selectNode';
+
+/// selectNode callback signature.
+typedef SelectNodeFunction = void Function(TreemapNode);
+
+class InstructionsSize {
```
kenzieschmoll, line 160: move this class to memory_graph_model.dart

packages/devtools_app/lib/src/treemap/treemap.dart
```diff
@@ -0,0 +1,459 @@
+import 'package:flutter/material.dart';
+
```
kenzieschmoll, line 2: instead of putting this inside its own treemap/ directory, let's put it inside src/charts/

packages/devtools_app/lib/src/treemap/treemap.dart
```diff
@@ -0,0 +1,459 @@
...
+import '../memory/memory_treemap.dart';
+
+class Treemap extends StatefulWidget {
+  const Treemap({
+    this.rootNode,
+    this.levelsVisible,
+    this.width,
+    this.height,
+    this.onTap,
```
kenzieschmoll, line 11: add @required to parameters that are required

packages/devtools_app/lib/src/memory/memory_treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
kenzieschmoll: Ah I didn't notice this was inside a lift. Leaving as is is fine then

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
kenzieschmoll: Oh okay. I think either is fine as long as we are consistent. Leaving as is is good.

packages/devtools_app/lib/src/app.dart
```diff
@@ -465,4 +465,4 @@ List<DevToolsScreen> get defaultScreens => <DevToolsScreen>[
 //        createController: () => ExampleController(),
 //        supportsOffline: true,
 //      ),
-    ];
+    ];
```
kenzieschmoll, line 468: add new line back at the end of this file

packages/devtools_app/lib/src/memory/memory_treemap.dart
kenzieschmoll: can you verify that the treemap rebuilds when taking a new snapshot?

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+    for (int i = start; i <= end; i++) {
+      sum += children[i].byteSize;
+    }
+    return sum;
+  }
+
+  int computePivot2(List<TreemapNode> children) {
+    switch (pivotType) {
+      case PivotType.pivotByMiddle:
+        {
```
kenzieschmoll, line 74: you can remove the brackets for each case

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+          for (int i = 0; i < children.length; i++) {
+            if (children[i].byteSize > maxSize) {
+              maxSize = children[i].byteSize.toDouble();
+              pivotIndex = i;
+            }
+          }
+          return pivotIndex;
+        }
+    }
+    return -1;
```
kenzieschmoll, line 90: do you need this return statement? all cases of the switch statement are met, so I would think you would not need it. If you get an error, you can also try adding this as a `default` case in the switch statement

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+  // the aspect ratio of PD is as low as it can be.
+
+  // Example layout:
+  // ----------------------------
+  // |      |  PD  |            |
+  // |      |      |            |
+  // |  L1  |------|    L3      |
+  // |      |  L2  |            |
+  // |      |      |            |
+  // ----------------------------
```
kenzieschmoll, line 111: use dart doc style ///

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+    }
+    return -1;
+  }
+
+  // Divides a given list of treemap nodes into four parts:
+  // L1, P, L2, L3
+
+  // P (pivot) is the treemap node chosen to be the pivot based on the pivot type.
+
+  // L1 inclues all treemap nodes before the pivot treemap nodes.
```
kenzieschmoll, line 98: typo: includes
typo: nodes --> node at the end of the sentence

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+
+  // L2 and L3 combined include all treemap nodes after the pivot treemap node.
+  // A combination of elements are put into L2 and L3 so that
+  // the aspect ratio of PD is as low as it can be.
```
kenzieschmoll, line 102: replace PD here and in the ascii art with P to make it consistent with above docs

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+  // |  L1  |------|    L3      |
+  // |      |  L2  |            |
+  // |      |      |            |
+  // ----------------------------
+  List<Cell> buildTreemap(
+    List<TreemapNode> children,
+    double width,
+    double height,
+    double x,
+    double y,
```
kenzieschmoll, line 117: make these required named parameters

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+  ) {
+    final bool isHorizontalRectangle = width > height;
```
kenzieschmoll, line 119: remove type from LHS

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+
+    final int totalSize = computeSize(children, 0, children.length - 1);
```
kenzieschmoll, line 121: remove type

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+
+    if (children.isEmpty) {
+      return [];
+    }
+
+    // Sort list of children, descending in size.
+    children.sort((a, b) => b.byteSize.compareTo(a.byteSize));
+
+    if (children.length == 1 || children.length == 2) {
```
kenzieschmoll, line 130: if (children <= 2)

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+      final List<Cell> positionedChildren = [];
```
kenzieschmoll, line 131: instead push the type to the RHS and write `final positionedChildren = <Cell>[];`

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+      double offset = isHorizontalRectangle ? x : y;
+
+      for (final child in children) {
+        final double ratio = child.byteSize / totalSize;
```
kenzieschmoll, line 135: remove type here and elsewhere in this file where it is not needed

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+  // The depth of children visible from this treemap widget.
+  // For example, levelsVisible = 2
+  // ---------------
+  // |    l = 1    |
+  // |  ---------  |
+  // |  | l = 2 |  |
+  // |  |       |  |
+  // |  |       |  |
+  // |  ---------  |
+  // ---------------
```
kenzieschmoll, line 25: from this I am unclear whether `l = 1` are root's children or root. can you clarify?

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+
+    if (children.length == 1 || children.length == 2) {
+      final List<Cell> positionedChildren = [];
+      double offset = isHorizontalRectangle ? x : y;
+
+      for (final child in children) {
+        final double ratio = child.byteSize / totalSize;
+        positionedChildren.add(
+          Cell(
+            key: UniqueKey(),
```
kenzieschmoll, line 138: do you need this key?

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+              message: rootNode.name,
+              child: buildNestedTreemap(),
+            ),
+          ),
+        );
+      } else {
+        return Container(
+          color: Colors.white38,
+          child: Center(
+            child: shouldDisplayText ? nameAndSizeText() : Container(),
```
kenzieschmoll, line 368: instead write 
```
Container(
          color: Colors.white38,
          child: shouldDisplayText ? Center(
            child: nameAndSizeText(),
          ) : SizedBox(),
)
```

(with prettier formatting of course :) )

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+      list3Height,
+      list3XCoord,
+      list3YCoord,
+    );
+
+    return list1Cells + [pivotCell] + list2Cells + list3Cells;
+  }
+
+  String nodeSizeText() {
+    final int size = rootNode.byteSize;
```
kenzieschmoll, line 295: remove type here and below

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+    final double sizeInKB = size / 1024;
```
kenzieschmoll, line 296: look at using printKb and printMb methods from utils.dart instead of doing all this manually

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+
+  // Treemap layout:
+  // ----------------------------
+  // |        Title Text        |
+  // |--------------------------|
+  // |                          |
+  // |         Content          |
+  // |                          |
+  // |                          |
+  // ----------------------------
```
kenzieschmoll, line 314: use dart doc ///

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+      child: Container(
+        decoration: BoxDecoration(border: Border.all(width: 0.5)),
+        child: Column(
+          children: [
+            if (widget.levelsVisible > 0 &&
+                rootNode.children.isNotEmpty &&
+                shouldDisplayText)
+              Center(
+                child: Text(
+                  rootNode.name + ' ' + nodeSizeText(),
```
kenzieschmoll, line 328: use dart string interpolation here: `'${rootNode.name} ${nodeSizeText()}'`

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+      }
+    } else if (widget.levelsVisible == 1) {
+      if (rootNode.children.isNotEmpty) {
+        return GestureDetector(
+          onTap: widget.onTap,
+          child: Container(
+            color: Colors.white38,
+            child: Tooltip(
+              message: rootNode.name,
+              child: buildNestedTreemap(),
```
kenzieschmoll, line 360: This should create a TreeMap widget instead of using the methods `buildTreemap` and `buildNestedTreemap`

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+    }
+
+    // Contruct list 1 sub-treemap.
+    final double list1SizeRatio = list1Size / totalSize;
+    final double list1Width =
+        isHorizontalRectangle ? width * list1SizeRatio : width;
+    final double list1Height =
+        isHorizontalRectangle ? height : height * list1SizeRatio;
+    List<Cell> list1Cells;
+    list1Cells = buildTreemap(
```
kenzieschmoll, line 232: create a new TreeMapWidget instead of using the method `buildTreemap`. Here and for the other cases in this method

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,493 @@
...
+    if (widget.levelsVisible == 0) {
+      if (shouldDisplayText) {
+        return Container(
+          width: widget.width,
+          child: Center(
+            child: nameAndSizeText(),
+          ),
+        );
+      } else {
+        return Container();
```
kenzieschmoll, line 350: return SizedBox() instead

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Updated comment to reflect that `l = 1` are root's children.

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Changed to default case: -1

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Fixed

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Fixed

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Without it, the treemap's names were not updating as it zoomed in on a cell although the cells were rearranging.

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/memory/memory_treemap.dart
peterdjlee: Yup, the treemap rebuilds with the new snapshot

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,517 @@
...
+
+  final double height;
+
+  final VoidCallback onTap;
+
+  @override
+  _TreemapState createState() => _TreemapState();
+}
+
+enum PivotType { pivotByMiddle, pivotBySize }
```
kenzieschmoll, line 44: stylistically we should keep a StatefulWidget class directly above the corresponding State class. Move enum above Treemap or below the state class

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,517 @@
...
+    super.initState();
+    rootNode = widget.rootNode;
+  }
+
+  bool get shouldDisplayText => widget.height > minHeightToDisplayText;
+
+  void cellOnTap(TreemapNode child) {
+    if (child != null) {
+      setState(() {
+        rootNode = child;
```
kenzieschmoll, line 63: can you add a todo to push rootNode state management into the controller responsible for managing the data? Ideally we'd like the memory controller to maintain the root node so that we can update the root node for the heap table accordingly. We'll want to do something like this for the code size page as well.

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,517 @@
...
+        }
+        return pivotIndex;
+      default:
+        return -1;
+    }
+  }
+
+  /// Divides a given list of treemap nodes into four parts:
+  /// L1, P, L2, L3.
+
```
kenzieschmoll, line 102: we want to keep the whole comment connected for dart doc, so instead of a blank line, write ///. here and elsewhere

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,517 @@
...
+        rootNode = child;
+      });
+    }
+  }
+
+  /// Computes the total size of a given list of treemap nodes.
+  int computeByteSizeForNodes({
+    @required List<TreemapNode> nodes,
+    @required int startIndex,
+    @required int endIndex,
```
kenzieschmoll, line 72: on second thought, let's make these optional named parameters that default to 0 and nodes.length - 1 respectively.

You won't be able to set `endIndex` to a default of `node.length - 1` in the params, but you can write `endIndex ??= nodes.length - 1;` at the beginning of the method.

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,517 @@
...
+    final pSize = pivotDataReference.byteSize;
+
+    final list1 = children.sublist(0, pivotIndex);
+    final list1Size = computeByteSizeForNodes(
+      nodes: list1,
+      startIndex: 0,
+      endIndex: list1.length - 1,
+    );
+
+    List<TreemapNode> list2 = [];
```
kenzieschmoll, line 178: replace with `var list2 = <TreemapNode>[];` here and for list3

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,517 @@
...
+    int list2Size = 0;
+    List<TreemapNode> list3 = [];
+    int list3Size = 0;
+
+    // The amount of data we have from pivot + 1 (exclusive)
+    // In another words, if we only put one data in l2, how many are left for l3?
```
kenzieschmoll, line 184: nit: typo "in other words"

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,517 @@
...
+    // [L1, pivotIndex, data, |d|] d = 2
+    final l3MaxLength = children.length - pivotIndex - 1;
+    int bestIndex = 0;
+    double pivotBestWidth = 0;
+    double pivotBestHeight = 0;
+
+    // We need to be able to put at least 3 elements in l3 for this algorithm.
+    if (l3MaxLength >= 3) {
+      double pBestAspectRatio = double.infinity;
+      // Iterate through different combinations of list2 and list3 to find
```
kenzieschmoll, line 194: in comments wrap variables and members with [ ] (e.g. [list2])

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,517 @@
...
+    final list1Size = computeByteSizeForNodes(
+      nodes: list1,
+      startIndex: 0,
+      endIndex: list1.length - 1,
+    );
+
+    List<TreemapNode> list2 = [];
+    int list2Size = 0;
+    List<TreemapNode> list3 = [];
+    int list3Size = 0;
```
kenzieschmoll, line 181: can we rename these "size" objects to "byteSize" (e.g. list3ByteSize) so it is clear we are talking about bytes and not list length?

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,517 @@
...
+        // Best aspect ratio that is the closest to 1.
+        if ((1 - pAspectRatio).abs() < (1 - pBestAspectRatio).abs()) {
+          pBestAspectRatio = pAspectRatio;
+          bestIndex = i;
+          // Kept track of width and height to construct the pivot cell.
+          pivotBestWidth = pWidth;
+          pivotBestHeight = pHeight;
+        }
+      }
+      // Split the rest of the data into list2 and list3
```
kenzieschmoll, line 224: wrap list2 and list3 with [ ]

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,517 @@
...
+      height: list2Height,
+      x: list2XCoord,
+      y: list2YCoord,
+    );
+
+    // Construct pivot cell.
+    final pivotXCoord = isHorizontalRectangle ? x + list1Width : x + list2Width;
+    final pivotYCorrd = isHorizontalRectangle ? y : y + list1Height;
+    final pivotCell = TreemapCell(
+      key: UniqueKey(),
```
kenzieschmoll, line 289: is this necessary?

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,517 @@
...
+  /// Example layout:
+  /// ```
+  /// ----------------------
+  /// |      |  P   |      |
+  /// |      |      |      |
+  /// |  L1  |------|  L3  |
+  /// |      |  L2  |      |
+  /// |      |      |      |
+  /// ----------------------
+  /// ```
```
kenzieschmoll, line 118: can you add a link here to the algorithm source this code is based on?

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Without the key, flutter tries to reuse the widgets and the names of each cell does not update accordingly when zooming in

packages/devtools_app/lib/src/charts/treemap.dart
kenzieschmoll: perhaps make the key the cells name and byte size so that we reuse the widgets where possible?

packages/devtools_app/lib/src/memory/memory_treemap.dart
jacob314: also update file names to tree_map if we are treating it as two words.

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,334 @@
...
+        sizes = InstructionsSize.fromSnapshop(controller);
+      });
+    });
+
+    addAutoDisposeListener(controller.filterNotifier, () {
+      setState(() {
+        controller.computeAllLibraries(true, true);
+      });
+    });
+    // TODO(peterdjlee): Need to check if applicable to treemap.
```
jacob314, line 60: @terrylucas any thoughts on this?

packages/devtools_app/lib/src/memory/memory_graph_model.dart
```diff
@@ -535,3 +538,138 @@ class HeapGraphClassLive extends HeapGraphClass {
   @override
   String toString() => name;
 }
+
+class InstructionsSize {
```
terrylucas, line 542: nit: I'm not sure these classes should be in memory_graph_model (it was a big file).  memory_graph_model is used by table view and analysis and treemap.  These classes are only used by treemap?  

Maybe a treemap_model.dart to hold all of these classes.  These are very specific to treemap - right?

packages/devtools_app/lib/src/memory/memory_treemap.dart
terrylucas: Search should be available from everywhere even treemap. Especially if the name can't be displayed at a zoomed out level e.g., navigation zero.

Search when treemap is visible should at least select the class name or library found?  But, I suspect it would highlight, select (click) and show current state in navigation bar.

Does Treemap expose a matchNames, findNode and selectNode?  If so, then I can add back this functionality post your commit.

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,460 @@
...
+
+    final pivotIndex = computePivot(children);
+
+    final pivotDataReference = children[pivotIndex];
+    final pByteSize = pivotDataReference.byteSize;
+
+    final list1 = children.sublist(0, pivotIndex);
+    final list1ByteSize = computeByteSizeForNodes(nodes: list1);
+
+    List list2 = <TreemapNode>[];
```
kenzieschmoll, line 161: use var instead of List. The type will be inferred by the RHS. here and below for list3

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,460 @@
...
+
+      final pdAndList2Ratio = (pByteSize + list2ByteSize) / totalByteSize;
+      final pdRatio = pByteSize / (pByteSize + list2ByteSize);
+      pivotBestWidth =
+          isHorizontalRectangle ? pdAndList2Ratio * width : pdRatio * width;
+      pivotBestHeight =
+          isHorizontalRectangle ? pdRatio * height : pdAndList2Ratio * height;
+    }
+
+    final List positionedTreemaps = <Positioned>[];
```
kenzieschmoll, line 224: remove List type

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    return positionedTreemaps;
+  }
+
+  Text buildNameAndSizeText({
+    @required Color fontColor,
+    @required bool onTwoLines,
+  }) {
+    final newline = onTwoLines ? '\n' : ' ';
```
kenzieschmoll, line 320: rename to separator since it is not always a new line

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,460 @@
...
+  TreemapNode({
+    @required this.name,
+    this.byteSize = 0,
+    this.childrenMap = const <String, TreemapNode>{},
+  })  : assert(name != null),
+        assert(byteSize != null),
+        assert(childrenMap != null);
+
+  final String name;
+  final Map<String, TreemapNode> childrenMap;
```
kenzieschmoll, line 442: rename to children

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,128 @@
...
+  const MemoryTreemap(this.controller);
+
+  final MemoryController controller;
+
+  @override
+  MemoryTreemapState createState() => MemoryTreemapState(controller);
+}
+
+class MemoryTreemapState extends State<MemoryTreemap> with AutoDisposeMixin {
+  MemoryTreemapState(this.controller);
```
kenzieschmoll, line 26: have you verified that taking a new snapshot builds a treemap for the new data?

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,460 @@
+import 'package:flutter/material.dart';
+
+import '../trees.dart';
+import '../ui/colors.dart';
+import '../utils.dart';
+
+enum PivotType { pivotByMiddle, pivotBySize }
+
+class NewTreemap extends StatefulWidget {
```
kenzieschmoll, line 9: rename to Treemap

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -1,70 +1,56 @@
...
-    @required this.height,
-    @required this.onTap,
-  });
+class NewTreemap extends StatefulWidget {
+  NewTreemap._({
+    this.rootNode,
+    this.nodes = const [],
+    this.levelsVisible,
+    this.onRootChangedCallback,
+  }) : assert(rootNode == null && nodes.isNotEmpty ||
```
jacob314, line 15: nit: prefer having two named constructors rather than one constructor where only one of two arguments can be set.

kenzieschmoll: I suggested that Peter do it this way. Since it is a widget, all fields have to be final, so the constructor that sets defaults for `rootNode` and `nodes` is private, and the only way to construct a Treemap widget is from `Treemap.fromRoot` or `Treemap.fromNodes`

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -1,70 +1,56 @@
...
-  final int levelsVisible;
-
-  final double width;
+  final List<TreemapNode> nodes;
 
-  final double height;
+  final int levelsVisible;
 
-  final VoidCallback onTap;
+  final void Function(TreemapNode) onRootChangedCallback;
```
jacob314, line 42: This should be 
`Function(TreemapNode node)`
Unless it got cleaned up when the `Function` keyword was added,
`Function(TreemapNode)`
actually means
`Function(dynamic TreemapNode)`

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -1,70 +1,56 @@
...
 
   @override
-  _TreemapState createState() => _TreemapState();
+  _NewTreemapState createState() => _NewTreemapState();
 }
 
-class _TreemapState extends State<Treemap> {
-  static const double minHeightToDisplayText = 20.0;
+class _NewTreemapState extends State<NewTreemap> {
   PivotType pivotType = PivotType.pivotBySize;
```
jacob314, line 49: this should be a stateless widget now that it doesn't have state.

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,460 @@
...
+    final newline = onTwoLines ? '\n' : ' ';
+    return Text(
+      '${widget.rootNode.name}$newline${nodeSizeText()}',
+      style: TextStyle(color: fontColor),
+      textAlign: TextAlign.center,
+      overflow: TextOverflow.ellipsis,
+    );
+  }
+
+  String nodeSizeText() {
```
kenzieschmoll, line 329: make this a private method on TreeNode class

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,460 @@
...
+    return positionedTreemaps;
+  }
+
+  Text buildNameAndSizeText({
+    @required Color fontColor,
+    @required bool onTwoLines,
+  }) {
+    final newline = onTwoLines ? '\n' : ' ';
+    return Text(
+      '${widget.rootNode.name}$newline${nodeSizeText()}',
```
kenzieschmoll, line 322: make `displayText({bool oneLine = true})` a method on `TreemapNode`

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,460 @@
...
+  }
+
+  Widget buildTreemapCell() {
+    if (widget.levelsVisible > 0 && widget.rootNode.children.isNotEmpty) {
+      return Padding(
+        padding: const EdgeInsets.all(1.0),
+        child: Column(
+          children: [
+            // TODO(peterdjlee): Abstract out to a widget or a method.
+            Tooltip(
```
kenzieschmoll, line 366: add waitDuration

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -0,0 +1,460 @@
...
+              message: widget.rootNode.name,
```
kenzieschmoll, line 367: let's put the byte size in this too. widget.rootNode.display

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,128 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    if (sizes != null) {
+      return LayoutBuilder(
+        builder: (context, constraints) {
+          return NewTreemap.fromRoot(
+              rootNode: root,
+              levelsVisible: 3,
+              onRootChangedCallback: _onRootChanged);
```
kenzieschmoll, line 100: add trailing comma

packages/devtools_app/lib/src/memory/memory_treemap.dart
kenzieschmoll: Let's hook the callbacks up as a follow up to this CL (which we will land after releasing next week so this won't break any users)

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -199,8 +182,8 @@ class _TreemapState extends State<Treemap> {
         );
 
         // Calculate the aspect ratio for the pivot treemap node.
-        final pAndList2Ratio = (pSize + list2Size) / totalSize;
-        final pRatio = pSize / (pSize + list2Size);
+        final pAndList2Ratio = (pByteSize + list2Size) / totalByteSize;
```
jacob314, line 185: avoid abbreviations in names. For example: use 
`pivot` instead of `p` in variable names. That makes it easier for someone new to the code to tell what the variables mean. Here I had to look up a line to browse up to the previous lines to figure out what `p` was.

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -170,10 +153,10 @@ class _TreemapState extends State<Treemap> {
     final pivotIndex = computePivot(children);
 
     final pivotDataReference = children[pivotIndex];
-    final pSize = pivotDataReference.byteSize;
+    final pByteSize = pivotDataReference.byteSize;
 
     final list1 = children.sublist(0, pivotIndex);
```
jacob314, line 158: you might want to add a helper datastructure with all the variables that currently end up with 1, 2, and 3 in them.
There is similar logic for each case where you get a list of nodes and then perform some computations on how to place it into a treemap. Here is an example with one of the metrics you compute. It looks like there are a number more metrics you could add to the little helper class.
E.g. 
```
class TreeMapMetrics {
  final List<TreeMapNode> nodes;
  final double byteSize;
  final double totalByteSize;
  final double pivotAndListRatio;
  TreeMapMetrics(this.nodes, this.totalBytesSize) : 
    byteSize = computeByteSizeForNodes(nodes);
   ...
}
```

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -230,74 +213,127 @@ class _TreemapState extends State<Treemap> {
...
+    if (list3.isNotEmpty) {
+      positionedTreemaps.add(
+        Positioned(
+          left: list3XCoord,
+          top: list3YCoord,
+          width: list3Width,
+          height: list3Height,
+          child: NewTreemap.fromNodes(
+            nodes: list3,
+            levelsVisible: widget.levelsVisible,
```
jacob314, line 306: comment why this isn't levelsVisible - 1 as that is slightly surprising unless a user understands the algorithm.

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -230,74 +213,127 @@ class _TreemapState extends State<Treemap> {
...
+      textAlign: TextAlign.center,
+      overflow: TextOverflow.ellipsis,
     );
+  }
 
-    return list1Cells + [pivotCell] + list2Cells + list3Cells;
+  String nodeSizeText() {
+    final size = widget.rootNode.byteSize;
+    final sizeInKB = size / 1024;
+    if (sizeInKB < 1024.0) {
```
jacob314, line 332: make this a utility method to pretty display memory sizes. There are printMb and printKb methods already in src/utils.dart that have some overlap. I agree that for this case you do want to sometimes display KB and sometimes display MB. The existing helpers in src/utils/dart handle a couple more edge cases to make the output easier to read so you should use that logic when it comes to displaying values in KB and MB.

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -306,174 +342,93 @@ class _TreemapState extends State<Treemap> {
...
+            widget.onRootChangedCallback(widget.rootNode);
+          },
           child: Container(
-            color: mainUiColor,
-            child: Tooltip(
-              message: rootNode.name,
-              child: buildNestedTreemap(),
+            decoration: BoxDecoration(
+              color: mainUiColor,
+              border: Border.all(width: 0.5),
```
jacob314, line 403: a border width 0.5 will look different on high DPI and regular DPI devices. Is there a reason we can't use border width 1.0 and a color that stands out less?
Not all high dpi devices will have a pixel ratio of 2.0 meaning you will have some subtle differences on high dpi devices as well.

kenzieschmoll:
> I know we talked about a navigation bar above the treemap is that post this PR?

Terry - the breadcrumbs will be included in this PR, but we will likely wait to land this until after we branch the release next week. The search / filter / selection will all need to be hooked up, which I'm thinking we can do as a follow up to this PR.

packages/devtools_app/lib/src/memory/memory_treemap.dart
peterdjlee: Yup, new data loads in and a different treemap is built when taking a new snapshot.

packages/devtools_app/lib/src/memory/memory_treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: It's named childrenMap because children is already defined in the TreeNode class, and we needed to add a separate map field for searching functionalities

packages/devtools_app/lib/src/memory/memory_graph_model.dart
peterdjlee: For now, I've moved it to memory_treemap.dart, but I think the file structure can be evaluated again after hooking up search/selection/filter after this CL

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: That's a good consideration, something I didn't think about. Fixed

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: To clarify, should I be adding this function to utils.dart? I do use the printKb and printMb functions to actually print the values, but I manually compute the size in KB first for the purpose of checking the size to decide whether to display the values in KB or MB.

kenzieschmoll: perhaps create a utility method `prettyPrintBytes(int size)` that leverages the existing `printMb` and `printKb` methods

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: I added some documentation for levelsVisible instead, since there are multiple instances where `levelsVisible - 1` is not passed in. Could you take a look and see if that's helpful?

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -315,27 +347,16 @@ class _NewTreemapState extends State<NewTreemap> {
 
   Text buildNameAndSizeText({
     @required Color fontColor,
-    @required bool onTwoLines,
+    @required bool oneline,
   }) {
-    final newline = onTwoLines ? '\n' : ' ';
     return Text(
-      '${widget.rootNode.name}$newline${nodeSizeText()}',
+      rootNode.displayText(oneline: oneline),
```
kenzieschmoll, line 353: use camel case for oneline --> oneLine

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -349,77 +370,78 @@ class _NewTreemapState extends State<NewTreemap> {
...
+              fontColor: Colors.black,
+              oneline: false,
             ),
           ),
         ),
       );
     }
   }
 
+  Tooltip buildSelectable({@required Widget child}) {
```
kenzieschmoll, line 426: add dart doc above this method that explains that selecting this widget will trigger a re-root of the tree to the associated rootNode

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -349,77 +370,78 @@ class _NewTreemapState extends State<NewTreemap> {
...
+    return Tooltip(
+      message: rootNode.displayText(),
+      waitDuration: tooltipWait,
```
kenzieschmoll, line 429: let's add preferBelow: false to match other tooltips in devtools

peterdjlee:
- Implemented breadcrumb navigation
memory_treemap3

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -405,31 +402,89 @@ class Treemap extends StatelessWidget {
           ],
         ),
       );
+    } else {
+      return Column(
+        children: [
+          if (levelsVisible == 2 && height > minHeightToDisplayTitleText)
```
kenzieschmoll, line 408: Why are we special casing 2 here? see other comment about adding an outerLevel param. Don't we want a title even for level 1 nodes?

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -405,31 +402,89 @@ class Treemap extends StatelessWidget {
...
+              ),
+              newRoot: pathFromRoot[index],
+            );
+          },
+        ),
+      );
     } else {
       return buildSelectable(
         child: Container(
+          width: double.infinity,
```
kenzieschmoll, line 460: is this width necessary?

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -405,31 +402,89 @@ class Treemap extends StatelessWidget {
...
+            border: Border.all(color: Colors.black54),
           ),
-          child: Center(
-            child: buildNameAndSizeText(
-              fontColor: Colors.black,
-              oneline: false,
-            ),
+          child: buildNameAndSizeText(
+            fontColor: Colors.white,
+            oneline: true,
```
kenzieschmoll, line 466: camel case for oneline --> oneLine

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -405,31 +402,89 @@ class Treemap extends StatelessWidget {
...
-            border: Border.all(),
+            border: Border.all(color: Colors.black54),
           ),
-          child: Center(
-            child: buildNameAndSizeText(
-              fontColor: Colors.black,
-              oneline: false,
-            ),
+          child: buildNameAndSizeText(
+            fontColor: Colors.white,
```
kenzieschmoll, line 465: this won't work for light theme. Use Theme.of(context).textTheme to grab a color

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Restructured code so this line is deleted.

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -405,31 +402,89 @@ class Treemap extends StatelessWidget {
...
+              ),
+            ),
+          ),
+        ],
+      );
+    }
+  }
+
+  Widget buildTitleText() {
+    if (levelsVisible == 2) {
```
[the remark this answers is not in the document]
peterdjlee: I added in a `outermostLevel` parameter for Treemap to make it more robust.

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Without the width, the title text boundaries don't fill up beyond the text width.

packages/devtools_app/lib/src/charts/treemap.dart

packages/devtools_app/lib/src/charts/treemap.dart
jacob314: ah I missed that this constructor is private. Actually I don't think this constructor is adding any value as you can just inline it into the other two constructors and then you don't need the assert as the assert is always true and using `this.` the code will be about ~15 lines shorter.

peterdjlee: Does that mean we should have something like this instead?
```
class Treemap extends StatelessWidget {

  const Treemap.fromRoot({
    @required this.rootNode,
    this.nodes,
    @required this.levelsVisible,
    @required this.outermostLevel,
    @required this.height,
    @required this.onRootChangedCallback,
  });

  const Treemap.fromNodes({
    this.rootNode,
    @required this.nodes,
    @required this.levelsVisible,
    @required this.outermostLevel,
    @required this.height,
    @required this.onRootChangedCallback,
  });

  final TreemapNode rootNode;

  final List<TreemapNode> nodes;
```

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,266 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    if (sizes != null) {
+      return LayoutBuilder(
+        builder: (context, constraints) {
+          return Treemap.fromRoot(
+            rootNode: root,
+            levelsVisible: 2,
+            outermostLevel: 2,
```
peterdjlee, line 100: @jacob314 I feel like it might not be the best design to make a developer change both `levelsVisible` and `outermostLevel` since both values need to be the same for the code to work as expected. What would be a better solution?

packages/devtools_app/lib/src/charts/treemap.dart
jacob314: correct.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -67,6 +67,15 @@ String percent2(double d) => '${(d * 100).toStringAsFixed(2)}%';
 final NumberFormat _kbPattern = NumberFormat.decimalPattern()
   ..maximumFractionDigits = 0;
 
+String prettyPrintBytes(int size) {
+  final sizeInKB = size / 1024.0;
+  if (sizeInKB < 1024.0) {
+    return '[${printKb(size)} KB]';
```
jacob314, line 73: the `[` and `]` shouldn't be in this method as that is a detail of how you want to display the # of byes in a larger string not part of how you want the bye string formatted.
Also, showing the KB and MB units is something that should be an option for the printKb and printMb helpers. In fact is is a bit odd we are displaying some sizes without a unit. For example, add a named parameter `includeUnit` to those methods..

peterdjlee: Done. I noticed that there are unit tests in place for `printMb` and `printKb` methods. Should there be some for `prettyPrintBytes` as well?

packages/devtools_app/lib/src/memory/memory_treemap.dart
jacob314: here's a cleaner design:
specify `levelsVisible` and `isOutermostLevel` as it looks like you are only using outermostLevel to check whether you are currently at the `outermostLevel`.

packages/devtools_app/lib/src/utils.dart
jacob314: yes it would be a good idea to write a quick unittest for this. you should also write a golden image test for the treemap code verifying that it renders as expected given some canned data.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -67,23 +67,25 @@ String percent2(double d) => '${(d * 100).toStringAsFixed(2)}%';
...
   }
 }
 
-String printKb(num bytes) {
+String printKb(num bytes, {bool includeUnit = false}) {
   // We add ((1024/2)-1) to the value before formatting so that a non-zero byte
   // value doesn't round down to 0.
-  return _kbPattern.format((bytes + 511) / 1024);
+  final unit = includeUnit ? ' KB' : '';
+  return _kbPattern.format((bytes + 511) / 1024) + unit;
```
jacob314, line 83: nit: unit is a misleading parameter name here because sometimes it is not a unit.
I would instead write either
```dart
return "_kbPattern.format((bytes + 511) / 1024)$}${includeUnit ? ' KB' : ''}";
```

or 
```dart
  var output =  _kbPattern.format((bytes + 511) / 1024);
  if (includeUnit) {
    output += ' KB';
  }
  return output;
```

peterdjlee: Done

packages/devtools_app/lib/src/utils.dart
peterdjlee: Unit test done. Writing golden test now

peterdjlee: I've been digging around the tests folder to learn more about the structure that's already in place. It seems like there aren't any tests written for the `HeapTree` widget which includes the bottom half of the memory screen. I was thinking it might best to add more regular tests in addition to a golden image test, but wasn't sure if all of the testing should be included in this PR or a separate PR.

kenzieschmoll:
closed in favor of https://github.com/flutter/devtools/pull/2131

PR #2085, opened 2020-06-16
Fix hang running as a macos desktop app.

jacob314:


packages/devtools_app/lib/src/config_specific/file/_file_desktop.dart
```diff
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:devtools_app/src/config_specific/logger/logger.dart';
```
devoncarew, line 5: use a relative import?

jacob314: done.

PR #2087, opened 2020-06-16
Fix case where table columns were missing adequate padding making the…

jacob314:
… UI hard to read.becausespacesaregood. This impacts allthetables.

Before:
Screen Shot 2020-06-15 at 9 11 12 PM

After:
Screen Shot 2020-06-15 at 9 48 22 PM


jacob314:
> Do we need to adjust the logic that calculates the width of all the columns? We have code that will add up all the fixed size columns, the left and right table padding, subtract from the overall table's width, and allocate the remaining space to the non-fixed size columns.

Good point. The treatment of fixed sized columns is a bit odd as their size is now the size with the spacing which is inconsistent for the first and last row. Cleaning this up and switching to spacing.

PR #2097, opened 2020-06-18
ui tweaks to the splitter icon

devoncarew:
ui tweaks to the splitter icon:
- make the color slightly gray
- change the splitter width to 12 px (from 24)

Screen Shot 2020-06-17 at 4 41 02 PM

Screen Shot 2020-06-17 at 4 41 06 PM

cc @kenzieschmoll

### DIFF
--- a/packages/devtools_app/lib/src/split.dart
+++ b/packages/devtools_app/lib/src/split.dart
@@ -278,7 +278,7 @@ class _SplitState extends State<Split> {
   double _totalSplitterSize() {
     final numSplitters = widget.children.length - 1;
     if (widget.splitters == null) {
-      return numSplitters * DefaultSplitter.size;
+      return numSplitters * DefaultSplitter.splitterWidth;
     } else {
       var totalSize = 0.0;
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/split.dart
```diff
@@ -292,17 +292,23 @@ class _SplitState extends State<Split> {
...
   final bool isHorizontal;
 
   @override
   Widget build(BuildContext context) {
     return Transform.rotate(
       angle: isHorizontal ? degToRad(90.0) : degToRad(0.0),
-      child: const Icon(
-        Icons.drag_handle,
-        size: size,
+      child: Align(
```
jacob314, line 304: @pq here is a simple flutter layout lint. Lint if Align is used in a case where `Center` would have the same effect and quick fix to using Center.

pq: https://github.com/dart-lang/linter/issues/2157

pq: Thanks!

PR #2099, opened 2020-06-18
Hide "Disconnected" overlay if DevTools is reconnected to a VM

DanTup:
When running in the existing non-embedded mode from VS Code, you can reconnect DevTools to  a new VM (using the server API). This didn't cause the disconnected overlay to be hidden so you'd end up with the overlay still visible but could see the tools in the background reconnected.

### DIFF
--- a/packages/devtools_app/lib/src/initializer.dart
+++ b/packages/devtools_app/lib/src/initializer.dart
@@ -125,6 +125,7 @@ class _InitializerState extends State<Initializer>
         disconnectedOverlayReconnectSubscription = serviceManager.onStateChange
             .where((connected) => connected)
             .listen((_) => hideDisconnectedOverlay());
+        autoDispose(disconnectedOverlayReconnectSubscription);
       }
     });
   }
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/initializer.dart
```diff
@@ -113,16 +116,29 @@ class _InitializerState extends State<Initializer>
...
   void _handleNoConnection() {
     WidgetsBinding.instance.addPostFrameCallback((_) {
-      if (!_checkLoaded() && ModalRoute.of(context).isCurrent) {
+      if (!_checkLoaded() &&
+          ModalRoute.of(context).isCurrent &&
+          currentDisconnectedOverlay == null) {
         Overlay.of(context).insert(_createDisconnectedOverlay());
+
+        // Set up a subscription to hide the overlay if we become reconnected.
+        disconnectedOverlayReconnectSubscription = serviceManager.onStateChange
```
jacob314, line 125: make sure to cancel the subscription in the dispose method or use the auto dispose mixin.

PR #2111, opened 2020-06-23
Improve flame chart performance by limiting row rebuilds on event selection

kenzieschmoll:
We were previously rebuilding the entire chart on selection. Now we only rebuild the rows with the previous and next selection.


jacob314:
lgtm

Issue #2101, opened 2020-06-18
DevTools Inspector Page: Layout Explorer UX Updates


raison00:
**Update/Add unified color themes**

- [x] Main Axis: `#1375bc`
 - [x] Row: `#2c5daa`
- [x] Cross Axis: `#8ac652`
 - [x] Column: `#77974d`
 - [x] Selected Widget: `#36c6f4`
- [x] Unselected: `#88b1de`
- [x] Emphasized Text, Arrows: `#009aca`
 - [x] Text, Arrows:  `#55767f`
- [x] Unconstrained: `#f5846b`
- [x] De-emphasized Unconstrained: `#dea089`

##
 **Empty Space, aka Negative Space**
 - [x] Negative Space:  `#51bade` at 30% opacity
 - [x] Negative Space:  `#386c7d` at 30% opacity
 - [x] Background Color Dark:  `#303030`
 - [x] Background Color Light:  `#ffffff`

##
 **Tiled Swatches for Negative Space:**
 **Dark Mode Specific Update: Tiled Swatch for Negative Space: 20% opacity to recede into background**
![image](https://user-images.githubusercontent.com/8105802/85625747-c3e51380-b620-11ea-90c6-7866f38d05e2.png) | ![swatch1-light-mode](https://user-images.githubusercontent.com/8105802/85625585-8bddd080-b620-11ea-94ee-d20b320ccc45.png)
 **Light Mode Specific Update: Tiled Swatch for Negative Space: 30% opacity to recede into background**
##
 **Dark Mode Specific Updates**
- [x] Selected Background Area: `#edeeef` at 30% opacity
- [x] Default Font Color: `#ffffff`

##
 **Light Mode Specific Updates**
- [x] Selected Background Area: `#edeeef`
- [x] Default Font Color: `#000000`
- [x] De-emphasized Font Color: `#505050` (if needed)

Dark Mode Updates:
![layout-explorer-dark](https://user-images.githubusercontent.com/8105802/85072259-a0781f80-b16d-11ea-961c-f3ce36883be2.png)

Light Mode Updates:

![layout-explorer-light](https://user-images.githubusercontent.com/8105802/85072283-b259c280-b16d-11ea-883e-ee2426c780ad.png)

PR #2112, opened 2020-06-24
Color cleanup

jacob314:
Finish color cleanup of the layout explorer page.

Eliminate all uses of ThemedColor as recent DevTools changes to support changing the color theme on the fly and loading it from local storage were messing up some color rendering in the inspector and timeline. The problem was ThemedColor was breaking Flutter internal caching and we were rendering an initial page with the wrong theme.
Now that we have removed the dart:html app, removing themed color was pretty easy.

All I did to do this refactor was make ThemedColor stop implementing Color, fix all type errors (by adding toColor calls), replaced all themed colors with extension methods on ColorScheme (recommended place for it not Theme), then regexp replace all toColor calls with colorCheme.colorName calls.

I had hoped to be able to stop at the toColor stage but that didn't shake out all the logic errors because we were using colors in slightly wrong places. It is key to stop passing colors into controllers and make sure you are flowing the colorScheme to places you need it.

The ugliest change is to the memory page where the chart controller needs colors and I really didn't want to change the third party package.
I've verified all pages work and all color regressions from ThemedColor are now fixed.

This matches the new spec for the layout explorer. There are some light layout explorer refactors as part of this to reduce duplicated logic for similar cases to make it easier to apply the ui refactors requested.
https://github.com/flutter/devtools/issues/2101
Screen Shot 2020-06-23 at 8 36 24 PM
Screen Shot 2020-06-23 at 8 36 14 PM
Screen Shot 2020-06-23 at 8 35 03 PM
Screen Shot 2020-06-23 at 8 34 47 PM


packages/devtools_app/lib/main.dart
```diff
@@ -6,12 +6,18 @@ import 'package:flutter/material.dart';
 
 import 'src/app.dart';
 import 'src/config_specific/framework_initialize/framework_initialize.dart';
+import 'src/preferences.dart';
 
 void main() async {
+  final preferences = PreferencesController();
+  // Wait for preferences to load before rendering the app to avoid a flash of
```
jacob314, line 13: I was hoping this was the main reason the inspector light mode theme was broken but that was too optimistic.

packages/devtools_app/lib/src/inspector/inspector_tree.dart
```diff
@@ -30,14 +30,13 @@ final RegExp assertionThrownBuildingError = RegExp(
...
 
-const Color selectedRowBackgroundColor = ThemedColor(
-  Color.fromARGB(255, 202, 191, 69),
-  Color.fromARGB(255, 99, 101, 103),
-);
-const Color hoverColor = ThemedColor(
-  Colors.yellowAccent,
-  Color.fromARGB(255, 70, 73, 76),
-);
+extension InspectorColorScheme on ColorScheme {
```
jacob314, line 33: as a future CL we should merge some of the various color scheme extensions that have grown organically.

packages/devtools_app/lib/src/inspector/layout_explorer/flex/flex.dart
```diff
@@ -73,64 +70,112 @@ const minHeightToAllowTruncating = 375.0;
 const minWidthToAllowTruncating = 375.0;
 
 // Story of Layout colors
```
jacob314, line 72: these colors were actually changed as part of the layout bug. This file is the only file with actual code changes. Trying to keep them together to avoid merge conflicts. The layout explorer was previously broken by ThemedColor combining with the color scheme not being fixed for the duration of running the app so it was hard to test this without fixing the theme issue.

packages/devtools_app/lib/src/inspector/layout_explorer/flex/flex.dart
```diff
@@ -922,17 +996,16 @@ class FlexChildVisualizer extends StatelessWidget {
   const FlexChildVisualizer({
```
jacob314, line 996: The flex child visualizer passed too many colors in when what we really wanted was to render based on the state. This resulted in color logic duplicated 3X

packages/devtools_app/lib/src/memory/memory_chart.dart
```diff
@@ -51,11 +51,6 @@ class MemoryChartState extends State<MemoryChart> with AutoDisposeMixin {
 
   @override
   void initState() {
-    _setupDartChartController();
```
jacob314, line 54: you cannot load the theme in initState() so this setup has to wait for when dependencies change.
The charts look fine so I think everything is alright. It would be much better if the controllers didn't define the colors when we shift this code over to use our own custom charting solution.

packages/devtools_app/lib/src/preferences.dart
```diff
@@ -6,35 +6,25 @@ import 'package:flutter/foundation.dart';
...
 /// A controller for global application preferences.
 class PreferencesController {
-  PreferencesController() {
-    _init();
-  }
-
-  final ValueNotifier<bool> _darkModeTheme =
-      // ignore: deprecated_member_use_from_same_package
-      ValueNotifier(devtools_theme.isDarkTheme);
+  final ValueNotifier<bool> _darkModeTheme = ValueNotifier(true);
```
jacob314, line 12: this is now the single source of truth for the color scheme instead of referencing the deprecated source of truth.

packages/devtools_app/lib/src/preferences.dart
```diff
@@ -6,35 +6,25 @@ import 'package:flutter/foundation.dart';
...
 
   ValueListenable get darkModeTheme => _darkModeTheme;
 
-  void _init() {
+  Future<void> init() async {
```
jacob314, line 16: previously the app would start running before the color scheme was returned from local storage which meant the scheme rendering one way and then changing.

kenzieschmoll:
have you tried switching the theme from dark to light via the settings dialog? I checked out your branch, and when I switch to light mode the flame chart doesn't update:

Screen Shot 2020-06-24 at 8 18 05 AM

packages/devtools_app/lib/src/connect_screen.dart
```diff
@@ -31,6 +31,7 @@ class _ConnectScreenBodyState extends State<ConnectScreenBody> {
   void initState() {
     super.initState();
     controller = TextEditingController();
+    controller.text = 'http://127.0.0.1:60775/i0PcNhWhGA0=/';
```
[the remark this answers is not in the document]
jacob314: you passed the test that you actually read the CL :)

packages/devtools_app/lib/src/inspector/diagnostics.dart
```diff
@@ -122,8 +125,8 @@ class DiagnosticsNodeDescription extends StatelessWidget {
               final int codePoint =
                   JsonUtils.getIntMember(properties, 'codePoint');
               if (codePoint > 0) {
-                final Icon icon =
-                    FlutterMaterialIcons.getIconForCodePoint(codePoint);
+                final Icon icon = FlutterMaterialIcons.getIconForCodePoint(
+                    codePoint, colorScheme);
```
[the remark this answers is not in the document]
jacob314: done

packages/devtools_app/lib/src/inspector/inspector_tree_flutter.dart
```diff
@@ -426,9 +427,9 @@ class _InspectorTreeState extends State<InspectorTree>
   bool get wantKeepAlive => true;
 }
 
-final _defaultPaint = Paint()
+Paint _defaultPaint(ColorScheme colorScheme) => Paint()
 // TODO(kenz): try to use color from Theme.of(context) for treeGuidelineColor
```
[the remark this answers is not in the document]
jacob314: indeed. done.

jacob314:
> have you tried switching the theme from dark to light via the settings dialog? I checked out your branch, and when I switch to light mode the flame chart doesn't update:
> 
> Screen Shot 2020-06-24 at 8 18 05 AM

I've tried switching using the dialog a lot. Let me double check the viewing the timeline page.

packages/devtools_app/lib/src/inspector/layout_explorer/flex/flex.dart
```diff
@@ -73,64 +70,112 @@ const minHeightToAllowTruncating = 375.0;
...
+    ),
+  );
+}
 
 const maxRequestsPerSecond = 3.0;
 
 /// Text widget for displaying width / height.
-Widget dimensionDescription(TextSpan description, bool overflow) {
+Widget dimensionDescription(
+    TextSpan description, bool overflow, ColorScheme colorScheme) {
```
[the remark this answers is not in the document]
jacob314: done

jacob314:
> > have you tried switching the theme from dark to light via the settings dialog? I checked out your branch, and when I switch to light mode the flame chart doesn't update:
> > Screen Shot 2020-06-24 at 8 18 05 AM
> 
> I've tried switching using the dialog a lot. Let me double check the viewing the timeline page.

Problem was aggressive equality operators for the custom painters causing old custom painters to be used after the theme changed.

PR #2113, opened 2020-06-24
Add padding to top of flame chart so selection brackets aren't clipped

kenzieschmoll:


packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -25,6 +25,7 @@ const double rowHeightWithPadding = rowHeight + rowPadding;
 const double sectionSpacing = 15.0;
 const double sideInset = 70.0;
 const double sideInsetSmall = 40.0;
+const int defaultRowOffsetForTopPadding = 3;
```
jacob314, line 28: comment why this is 3 as the hack is flutter profiling specific.

PR #2114, opened 2020-06-24
Fix exception in CPU profile transformer

kenzieschmoll:
An exception would be thrown when we select a UI timeline event and attempt to pull a CPU profile, and then select another  UI timeline event before the first profile was done being processed. Now we bail out early if the processId's do not match.

### DIFF
--- a/packages/devtools_app/lib/src/profiler/cpu_profile_transformer.dart
+++ b/packages/devtools_app/lib/src/profiler/cpu_profile_transformer.dart
@@ -29,13 +29,19 @@ class CpuProfileTransformer {
 
   int _stackFramesProcessed = 0;
 
-  Future<void> processData(CpuProfileData cpuProfileData) async {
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/profiler/cpu_profile_transformer.dart
```diff
@@ -50,6 +56,10 @@ class CpuProfileTransformer {
 
     // Use batch processing to maintain a responsive UI.
     while (_stackFramesProcessed < _stackFramesCount) {
+      if (processId != _activeProcessId) {
```
jacob314, line 59: nit: I think the best place for this throw is right after each `await` call as you will want to add more checks if you add more awaits.

kenzieschmoll: done

PR #2107, opened 2020-06-22
Build treemap from instruction size input data for code size page

peterdjlee:
* Render treemap after processing instruction sizes data with `vm_snapshot_analysis` package.
Screen Shot 2020-06-24 at 2 50 20 PM

PR #2116, opened 2020-06-24
Add treemap code and pretty-print for bytes

peterdjlee:
Originally reviewed and approved in #2084 - breaking it into a separate PR to land only the Treemap code and code related to pretty-printing bytes


packages/devtools_app/test/support/utils.dart
```diff
@@ -42,6 +44,46 @@ Future<Timeline> loadNetworkProfileTimeline() async {
...
+  final processedTestData = treemapFromJson(treemapTestData);
+  processedTestData['n'] = 'Root';
+
+  return generateTree(processedTestData);
+}
+
+// TODO(peterdjlee): Remove this method once code size PR (#2107) lands.
+/// Builds a tree with [TreemapNode] from [treeJson] which represents
+/// the hierarchical structure of the tree.
+TreemapNode generateTree(Map<String, dynamic> treeJson) {
```
peterdjlee, line 61: This code also lives in `CodeSizeController` which I was planning on using after deleting this method once the code size PR (#2107) lands. Instead, should I be taking this method out to something like `treemap_utils.dart`?

packages/devtools_app/test/treemap_test.dart
```diff
@@ -0,0 +1,83 @@
...
+
+void main() {
+  group('Treemap', () {
+    TreemapNode root;
+
+    setUp(() async {
+      root = await loadInstructionSizesJson();
+    });
+
+    void changeRoot(TreemapNode newRoot) {
```
peterdjlee, line 20: @kenzieschmoll This is also a duplicate that exists in `CodeSizeController`. Should I be taking it out to a different file or is it okay if I keep both?

packages/devtools_app/test/support/utils.dart
kenzieschmoll: yeah. let's add here and figure out how to clean up in your code size cl

packages/devtools_app/test/treemap_test.dart
kenzieschmoll: i don't think you actually need this method. just pass a null or noop callback ( `() {}` ) for onRootChangedCallback

peterdjlee: I think if we want to test what happens after zooming in we need the function

PR #2117, opened 2020-06-24
Replace heatmap with treemap in memory page

peterdjlee:
Originally reviewed and approved in #2084  - breaking it into a separate PR to land only the code that replaces the heatmap with treemap in memory page.


peterdjlee:
Closed to land existing PR (#2084) instead.

PR #2121, opened 2020-06-24
Update negative space images to use new assets that can be tiled.

jacob314:


jacob314:
Screen Shot 2020-06-24 at 3 46 46 PM
Screen Shot 2020-06-24 at 3 46 36 PM

PR #2123, opened 2020-06-25
Explain canvaskit crash and restart the tab before CanvasKit destroys the browser.

jacob314:
Please suggest a better warning message. I'm a bit conservative about warnings so we only blow up on canvaskit font warnings.

Screen Shot 2020-06-24 at 5 35 41 PM


packages/devtools_app/web/index.html
```diff
@@ -52,6 +52,22 @@
...
     }
   </script>
+
+  <!-- Script to fail gracefully when CanvasKit crashes. -->
+  <script>
+    var realWarn = window.console.warn.bind(console);
+    window.console.warn = function(obj) {
+      if( typeof(obj) == "string" && obj.indexOf("CanvasKit") != -1 && obj.indexOf("font") != -1) { 
+        window.alert(
+          "The Flutter CanvasKit Skia Backened has encountered a fatal error and needs to restart. " +
```
kenzieschmoll, line 62: typo: backend

devoncarew, line 62: Either:

```
¯\_(ツ)_/¯
```

or:

```
The Flutter web backend ('CanvasKit') has encountered a fatal error and needs to restart. ...
```

kenzieschmoll: maybe: 

"Oh no! Fatal error in the Flutter CanvasKit backend. We are aware of this issue and working hard on a fix. Press OK to restart DevTools."

jacob314: done. used the second. Also fixed an issue with multiple alerts because of cascading errors from canvaskit before the reload can happen.

jacob314: That sounds good as well. I pasted in Devon's suggestion first. Feel free to switch to this suggestion before committing.

PR #2131, opened 2020-06-26
Replace heatmap with treemap on the memory page

peterdjlee:
Originally reviewed and approved in #2084 -- isolating code that replaced the heatmap with treemap on the memory page.
Screen Shot 2020-06-26 at 11 27 27 AM


packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -369,7 +366,7 @@ class HeapTreeViewState extends State<HeapTree> with AutoDisposeMixin {
       mainAxisAlignment: MainAxisAlignment.end,
       children: [
         Expanded(child: snapshotDisplay),
-        if (hasDetails) const SizedBox(width: defaultSpacing),
+        hasDetails ? const SizedBox(width: defaultSpacing) : const SizedBox(),
```
kenzieschmoll, line 369: why is this change necessary?

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,258 @@
...
+
+        sizes = InstructionsSize.fromSnapshot(controller);
+      });
+    });
+
+    addAutoDisposeListener(controller.filterNotifier, () {
+      setState(() {
+        controller.computeAllLibraries(rebuild: true);
+      });
+    });
```
kenzieschmoll, line 63: so does filter work in this CL or not? If not, I don't think we should be listening to filterNotifier

packages/devtools_app/lib/src/memory/memory_treemap.dart
```diff
@@ -0,0 +1,258 @@
...
+    //   setState(() {
+    //     if (_trySelectItem()) {
+    //       closeAutoCompleteOverlay();
+    //     }
+    //   });
+    // });
+
+    addAutoDisposeListener(controller.searchAutoCompleteNotifier, () {
+      setState(autoCompleteOverlaySetState(controller, context));
+    });
```
kenzieschmoll, line 83: similar comment here. If search is not hooked up, I don't think we should be listening for search autocomplete

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
peterdjlee: I think initially it was because the line didn't fit in one line but it should fit in one line. Reverted.

packages/devtools_app/lib/src/memory/memory_treemap.dart
peterdjlee: Commented out

packages/devtools_app/lib/src/memory/memory_treemap.dart
peterdjlee: Commented out

PR #2132, opened 2020-06-26
Render treemap for v8 snapshot input on the code size page

peterdjlee:
* Ensure the treemap on code size page shows when given a v8 snapshot input
Screen Shot 2020-06-26 at 1 06 10 PM


packages/devtools_app/test/treemap_test.dart
```diff
@@ -44,43 +44,128 @@ void main() {
...
+        expect(find.text(text), findsNothing);
+
+        await expectLater(
+          find.byKey(treemapKey),
+          matchesGoldenFile('goldens/treemap_sizes_zoom.png'),
+        );
+        // Await delay for golden comparison.
+        await tester.pumpAndSettle(const Duration(seconds: 2));
+      },
+      skip: kIsWeb || !Platform.isMacOS,
```
kenzieschmoll, line 90: you can add this skip to the whole group instead of individually skipping each test. Same for below.

peterdjlee: It seems like the tests aren't actually skipped even when I set the group skip parameters to a hard-coded `true`. I'm not sure if the result is intended; should I keep the skip parameters for each test?

kenzieschmoll: Hmm that is surprising. Go ahead and move the skips back to the tests.

peterdjlee: Done

kenzieschmoll:
If you merge master, the bots should go green

PR #2154, opened 2020-07-04
[debugger] Add support for non-primitive map keys

jonahwilliams:
Currently a map with keys that do not have a meaningful `valueAsString` will just display `[null]:` in the dropdown. To support arbitrarily complex keys/maps, instead create an intermediate "Entry X" object, with a child bound variable for the key and value instanceRefs. This allows full inspection of key values.

EDIT: Updated image with grouped entries

localhost_63300_#__uri=ws%3A%2F%2F127 0 0 1%3A63309%2FH2X81qU1ZCI%3D%2Fws - Google Chrome 7_7_2020 1_07_02 PM


kenzieschmoll:
Can you add a screenshot of the change to the PR description? thanks

packages/devtools_app/test/association_variable_test.dart
```diff
@@ -0,0 +1,184 @@
...
+      kind: InstanceKind.kMap,
+      id: '123',
+      classRef: null,
+      associations: [
+        MapAssociation(
+          key: InstanceRef(
+              classRef: null,
+              id: '4',
+              kind: InstanceKind.kString,
+              valueAsString: 'Hey'),
```
kenzieschmoll, line 57: trailing comma

packages/devtools_app/test/association_variable_test.dart
```diff
@@ -0,0 +1,184 @@
...
+import 'package:vm_service/vm_service.dart';
+
+import 'package:devtools_app/src/debugger/debugger_controller.dart';
+import 'package:devtools_app/src/debugger/debugger_model.dart';
+import 'package:devtools_app/src/globals.dart';
+import 'package:devtools_app/src/service_manager.dart';
+import 'package:devtools_app/src/vm_service_wrapper.dart';
+
+void main() {
+  ServiceConnectionManager manager;
```
kenzieschmoll, line 17: can you use the FakeServiceManager class we use elsewhere in DevTools tests?

packages/devtools_app/test/association_variable_test.dart
```diff
@@ -0,0 +1,184 @@
...
+    when(manager.service.onIsolateEvent).thenAnswer((_) {
+      return const Stream.empty();
+    });
+    when(manager.service.onStdoutEvent).thenAnswer((_) {
+      return const Stream.empty();
+    });
+    when(manager.service.onStderrEvent).thenAnswer((_) {
+      return const Stream.empty();
+    });
+    globals[ServiceConnectionManager] = manager;
```
kenzieschmoll, line 35: use `setGlobal` method from globals.dart

packages/devtools_app/test/association_variable_test.dart
```diff
@@ -0,0 +1,184 @@
...
+    debuggerController = DebuggerController();
+    debuggerController.isolateRef =
```
kenzieschmoll, line 37: use .. notation (`debuggerController = DebuggerController()..isolateRef = ...;`)

packages/devtools_app/test/association_variable_test.dart
```diff
@@ -0,0 +1,184 @@
...
+        IsolateRef(id: '1', number: '2', name: 'main');
+  });
+
+  tearDown(() {
+    globals[ServiceConnectionManager] = null;
+  });
```
kenzieschmoll, line 43: is this necessary if the global  var is set in each setUp call?

jonahwilliams:
Screenshots attached to both PRs

packages/devtools_app/test/association_variable_test.dart
```diff
@@ -0,0 +1,184 @@
...
+        equals(value),
+      )
+      .having(
+          (v) => v.boundVar,
+          'boundVar',
+          const TypeMatcher<BoundVariable>()
+              .having((bv) => bv.name, 'name', equals(name)));
+}
+
+class MockVmServiceWrapper extends Mock implements VmServiceWrapper {}
```
kenzieschmoll, line 184: we already have a mock for VmServiceWrapper. See `MockVmService` in support/mocks.dart. If we can use FakeServiceManager, the mock vm service is included by default.

kenzieschmoll:
Top level comment - I wonder if it would be better to display `[value]` as a child of `[key]`? Or at least to indent it as if it were? What do other debuggers (like Chrome DevTools) do to display key-value pairs with non-primitive keys?

jonahwilliams:
> Top level comment - I wonder if it would be better to display [value] as a child of [key]?

I'd want to avoid confusing the value with a property of the key itself. Indenting slightly might work better, or drawing some sort of connector to show the are associated. Alternatively, I could create a synthetic object just for grouping the two. Though in all of these cases I would be worried about running out of horizontal real estate.

> What do other debuggers (like Chrome DevTools) do to display key-value pairs with non-primitive keys?

for `Map` objects, it looks like chrome groups each key/value together into a numbered group and then displays them side by side.

debugger  Add support for non-primitive map keys by jonahwilliams · Pull Request #2154 · flutter_devtools - Google Chrome 7_6_2020 10_07_13 AM (2)

kenzieschmoll:
I think grouping the key value pair together in an entry object makes it more clear that the key and value are associated. We could get rid of the `Entries` field and jump down one level further to preserve horizontal space. Perhaps `Entry 0`, `Entry 1`, etc.

jonahwilliams:
Sounds good, I'll take another crack at this tonight and post an update when it is ready to be reviewed. Thanks @kenzieschmoll !

packages/devtools_app/test/association_variable_test.dart
jonahwilliams: Fixed

packages/devtools_app/test/association_variable_test.dart
jonahwilliams: Done, though I hit some issues with the isolate manager which I worked around by updating DebuggerController to avoid switchToIsolate logic for testing.

packages/devtools_app/test/association_variable_test.dart
jonahwilliams: Done

packages/devtools_app/test/association_variable_test.dart
jonahwilliams: Done

packages/devtools_app/test/association_variable_test.dart
jonahwilliams: No, good point - removed.

packages/devtools_app/test/association_variable_test.dart
jonahwilliams: Done

jonahwilliams:
Updated with grouped entries.

jonahwilliams:
I also removed the special casing for primitives that used valueAsString

packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -24,8 +24,10 @@ final _log = DebugTimingLogger('debugger', mute: true);
 /// Responsible for managing the debug state of the app.
 class DebuggerController extends DisposableController
     with AutoDisposeControllerMixin {
-  DebuggerController() {
-    switchToIsolate(serviceManager.isolateManager.selectedIsolate);
+  DebuggerController({bool initialSwitchToIsolate = true}) {
+    if (initialSwitchToIsolate) {
```
jonahwilliams, line 28: I added this to prevent the fake from forcing the DebuggerController through isolate subscription. Otherwise, thats a lot of mocking each test needs

packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -688,7 +690,9 @@ class DebuggerController extends DisposableController
   /// We call this method as we expand variables in the variable tree, because
   /// building the tree for all variable data at once is very expensive.
   Future<void> buildVariablesTree(Variable variable) async {
-    if (!variable.isExpandable || variable.treeInitialized) return;
+    if (!variable.isExpandable ||
```
jonahwilliams, line 693: The addition of the entires clause means that there can be an expandable Variable without an instance ref. In this case, child variables are already added so this needs to short-circuit.

PR #2156, opened 2020-07-06
Add fraction digits parameter to all print byte methods

peterdjlee:
* Add fraction digits parameter to `prettyPrintBytes` and `printKB` to make byte display more flexible

### DIFF
--- a/packages/devtools_app/lib/src/charts/treemap.dart
+++ b/packages/devtools_app/lib/src/charts/treemap.dart
@@ -552,7 +552,7 @@ class TreemapNode extends TreeNode<TreemapNode> {
   String prettyByteSize() {
     // Negative sign isn't explicitly added since a regular print of a negative number includes it.
     final plusSign = showDiff && byteSize > 0 ? '+' : '';
-    return '$plusSign${prettyPrintBytes(byteSize, includeUnit: true)}';
+    return '$plusSign${prettyPrintBytes(byteSize, fractionDigits: 2, includeUnit: true)}';
   }
 
   /// Returns a list of [TreemapNode] in the path from root node to [this].
--- a/packages/devtools_app/lib/src/charts/treemap.dart
+++ b/packages/devtools_app/lib/src/charts/treemap.dart
@@ -552,7 +552,7 @@ class TreemapNode extends TreeNode<TreemapNode> {
   String prettyByteSize() {
     // Negative sign isn't explicitly added since a regular print of a negative number includes it.
     final plusSign = showDiff && byteSize > 0 ? '+' : '';
-    return '$plusSign${prettyPrintBytes(byteSize, fractionDigits: 2, includeUnit: true)}';
+    return '$plusSign${prettyPrintBytes(byteSize, kbFractionDigits: 1, includeUnit: true)}';
   }
 
   /// Returns a list of [TreemapNode] in the path from root node to [this].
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -552,7 +552,7 @@ class TreemapNode extends TreeNode<TreemapNode> {
   String prettyByteSize() {
     // Negative sign isn't explicitly added since a regular print of a negative number includes it.
     final plusSign = showDiff && byteSize > 0 ? '+' : '';
-    return '$plusSign${prettyPrintBytes(byteSize, includeUnit: true)}';
+    return '$plusSign${prettyPrintBytes(byteSize, fractionDigits: 2, includeUnit: true)}';
```
kenzieschmoll, line 555: what about doing 1 here? 2 seems like it may be a little too granular.

packages/devtools_app/lib/src/code_size/code_size_table.dart
```diff
@@ -82,8 +82,13 @@ class _SizeColumn extends ColumnData<TreemapNode> {
   dynamic getValue(TreemapNode dataObject) => dataObject.byteSize;
 
   @override
-  String getDisplayValue(TreemapNode dataObject) =>
-      prettyPrintBytes(dataObject.byteSize, includeUnit: true);
+  String getDisplayValue(TreemapNode dataObject) {
+    return prettyPrintBytes(
+      dataObject.byteSize,
+      fractionDigits: 2,
```
kenzieschmoll, line 88: same comment here. Maybe do 1?

packages/devtools_app/lib/src/utils.dart
```diff
@@ -65,22 +65,27 @@ final NumberFormat nf = NumberFormat.decimalPattern();
...
-  ..maximumFractionDigits = 0;
+String printKB(num bytes, {int fractionDigits = 0, bool includeUnit = false}) {
+  final NumberFormat _kbPattern = NumberFormat.decimalPattern()
+    ..maximumFractionDigits = fractionDigits;
 
-String printKB(num bytes, {bool includeUnit = false}) {
   // We add ((1024/2)-1) to the value before formatting so that a non-zero byte
-  // value doesn't round down to 0.
-  var output = _kbPattern.format((bytes + 511) / 1024);
+  // value doesn't round down to 0. If showing decimal points, let it round normally.
```
kenzieschmoll, line 86: add a todo here to add similar rounding logic to ensure non-zero values when fractionDigits > 0.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -65,22 +65,27 @@ final NumberFormat nf = NumberFormat.decimalPattern();
 
 String percent2(double d) => '${(d * 100).toStringAsFixed(2)}%';
 
-String prettyPrintBytes(num bytes, {bool includeUnit = false}) {
+String prettyPrintBytes(
+  num bytes, {
+  int fractionDigits = 0,
```
kenzieschmoll, line 70: we should probably have two parameters `kbFractionDigits` (defaults to 0), and `mbFractionDigits` (defaults to 1 to match `printMb`).

It feels natural to need different granularity defaults at different orders of magnitude.

packages/devtools_app/lib/src/charts/treemap.dart

packages/devtools_app/lib/src/code_size/code_size_table.dart
peterdjlee: Done

packages/devtools_app/lib/src/utils.dart
peterdjlee: Makes sense! Done

packages/devtools_app/lib/src/utils.dart
peterdjlee: Done

PR #2166, opened 2020-07-08
Added Event Pane.

terrylucas:
Added event pane displays user and VM events e.g., user initiated GC, Snapshots, Auto-Snapshots, Monitoring Allocations and GCs occurring in the VM.

image

Possible events are:

image

Various commands glyphs have changed snapshot, collapse/expand:

image

Tooltips exist e.g.,

image

Monitoring of memory allocations have been added with accumulators for instances and bytes (size of memory objects):

image

Clicking on 'Monitor Allocations':

image

creates a monitor allocation event in the event page and an entry in the table view

image

Clicking on the 'Monitor ...' will display the allocation/accumulators for all memory: 

image

Pressing 'Reset Accumulators' button:

image

Resets both Accum columns to zero and adds a 'Reset Accumulator' event to the event pane.  Monitoring the accumulators can help in detecting memory leaks.

image


packages/devtools_shared/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: Package of shared structures between devtools_app and devtools_serv
 
 # Note: this version should only be updated by running tools/update_version.sh
 # that updates all versions of packages from packages/devtools.
-version: 0.8.1-dev.1
```
jacob314, line 6: you shouldn't have to update this version as part of the CL. The publish script will now update all versions in lock step.

cobblest:
I have a few ideas: 
1) the Monitoring Allocation and Reset Accumulator icons are a bit confusing. It's hard to quickly realize what it means by looking at the icons. Shall we use text instead? 
2) What's the purpose of having monitor allocation - Reset? Is it similar function with Snapshots (monitoring changes between snapshots)? Right now there are so many icons in the first screenshot and I am concerned that users would be confused about different types of events. It could be confusing especially because we use different color+shape combinations. From human attentions's perspective, it's will cost a lot of cognitive resources to process multiple features https://en.wikipedia.org/wiki/Feature_integration_theory

packages/devtools_testing/lib/support/memory_test_data.dart
```diff
@@ -0,0 +1,8763 @@
...
+///     - allocation profile followed by an allocation profile reset
+///     - allocation profile only
+/// 70 VM GC events
+/// Dart Heap capacity ~684 MB, Used ~42 MB, External 634 MB, and RSS 534 MB
+/// 292 samples
+/// 
+/// How to create a new test data set. Run DevTools collecting stats then press
+/// the Export button. Then copy the contents of the exported file and assign to
+/// testHeapSampleData.
+final testHeapSampleData = {
```
jacob314, line 24: you can use a `r'''` string to make this easier for the analysis server to handle and make the test data match the actual parsed json a little more.

packages/devtools/pubspec.yaml
```diff
@@ -16,8 +16,8 @@ environment:
   sdk: '>=2.3.0 <3.0.0'
 
 dependencies:
```
jacob314, line 18: you can revert these pubspec changes and just rely on pinned dependencies until we publish a new version.

packages/devtools_app/lib/src/connect_screen.dart
```diff
@@ -115,7 +115,6 @@ class _ConnectScreenBodyState extends State<ConnectScreenBody> {
       );
       return;
     }
-
```
jacob314, line 118: nit: revert this probably accidental change.

packages/devtools_app/lib/src/memory/memory_allocation_table_data.dart
```diff
@@ -0,0 +1,159 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import '../table_data.dart';
+import 'memory_protocol.dart';
+
+const fieldWidthNumbers = 100.0;
```
jacob314, line 8: nit: `fieldWidthNumbers` --> `defaultNumberFieldWidth` or `numberFieldWidth`

packages/devtools_app/lib/src/memory/memory_allocation_table_data.dart
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  @override
+  double get fixedWidthPx => fieldWidthNumbers;
+}
+
+class FieldInstanceAccumulatorColumn extends ColumnData<ClassHeapDetailStats> {
+  FieldInstanceAccumulatorColumn()
+      : super(
+          'Accum',
```
jacob314, line 69: Is there another name we can show that is clearer than 'Accum'? Maybe we can use the same terminology as when diffing two snapshots so
Delta or Difference?

packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
```diff
@@ -0,0 +1,89 @@
...
+    with AutoDisposeMixin {
+  MemoryController controller;
+
+  final List<ColumnData<ClassHeapDetailStats>> columns = [];
+
+  @override
+  void initState() {
+    setupColumns();
+
+    super.initState();
```
jacob314, line 32: @pq here is an example of a lint that could help avoid a case that trips me up all the time.
https://api.flutter.dev/flutter/widgets/State/initState.html
calls to `super.initState()` should occur at the top op of the initState method by convention. Similarly, calls to super.dispose() should occur at the bottom.

cobblest:
We could also consider simplifying the design for events icons by using less color and shape combinations. I created some examples below: 
DevTools Snapshot Icons

packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
```diff
@@ -0,0 +1,89 @@
...
+      columns: columns,
+      data: controller.monitorAllocations,
+      keyFactory: (d) => Key(d.classRef.name),
+      onItemSelected: (ref) {},
+      sortColumn: controller.sortedMonitorColumn,
+      sortDirection: controller.sortedMonitorDirection,
+      onSortChanged: (column, direction) {
+        controller.sortedMonitorColumn = column;
+        controller.sortedMonitorDirection = direction;
+
```
jacob314, line 83: nit: remove blank line

packages/devtools_app/lib/src/memory/memory_chart.dart
```diff
@@ -204,31 +206,21 @@ class MemoryChartState extends State<MemoryChart> with AutoDisposeMixin {
...
-          const SizedBox(width: denseSpacing),
-        if (controller.isAndroidChartVisible)
-          Expanded(
-            child: SizedBox(
-              height: defaultChartHeight,
-              child: LineChart(androidChartController),
-            ),
-          ),
+        memoryTimeline.liveData.isEmpty
+            ? const SizedBox(height: 0)
```
jacob314, line 212: nit: is the height: 0 really needed?

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -583,27 +626,76 @@ class MemoryController extends DisposableController
...
+
+  var _monitorAllocations = <ClassHeapDetailStats>[];
+
+  List<ClassHeapDetailStats> get monitorAllocations => _monitorAllocations;
+
+  set monitorAllocations(List<ClassHeapDetailStats> allocations) {
+    _monitorAllocations = allocations;
+    // Clearing allocations reset ValueNotifier to zero.
+    allocations.isNotEmpty
+        ? _monitorAllocationsUpdated.value++
```
jacob314, line 649: nit: I would write
```dart
if (allocations.isEmpty) {
  _monitorAllocationsUpdated = 0;
} else {
  _monitorAllocationsUpdated.value++;
}
```

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -583,27 +626,76 @@ class MemoryController extends DisposableController
...
+  }
+
   Future<List<ClassHeapDetailStats>> resetAllocationProfile() =>
       getAllocationProfile(reset: true);
 
-  // 'reset': true to reset the object allocation accumulators
+  /// 'reset': true to reset the object allocation accumulators
   Future<List<ClassHeapDetailStats>> getAllocationProfile({
-    bool reset = false,
+    reset = false,
```
jacob314, line 658: add back `bool`. I don't think Dart infers that reset has to have type bool in this case but I could be wrong.

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -583,27 +626,76 @@ class MemoryController extends DisposableController
...
         .where((ClassHeapDetailStats stats) {
       return stats.instancesCurrent > 0 || stats.instancesAccumulated > 0;
     }).toList();
+
+    return allocations;
+  }
+
+  Future<void> computeLibraries() async {
+    // TODO(terry): Review why unawaited is necessary.
+    unawaited(serviceManager.service.getVM().then((vm) {
```
jacob314, line 686: this unawaited seems wrong. The Future returned by the method doesn't match when the libraries are actually computed. On the surface it seems like you could write
```dart
final vm = await serviceManager.service.getVM();
final isolates = await Future.wait(vm.isolates.map((IsolateRef ref) => serviceManager.service.getIsolate(ref.id))
...

```

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -811,18 +903,53 @@ class MemoryController extends DisposableController
...
 
-    if (topNode == null) {
-      topNode = LibraryReference(this, libraryRootNode, null);
+    AllocationsMonitorReference monitorRoot;
+    var anyAnalyses = false;
+    for (final reference in topNode.children) {
+      if (reference is AllocationsMonitorReference) {
+        monitorRoot = reference;
+      }
+      anyAnalyses |= reference is AnalysesReference;
```
jacob314, line 914: `nit: ||= instead of |=`

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -811,18 +903,53 @@ class MemoryController extends DisposableController
...
     }
 
+    if (monitorAllocations.isNotEmpty) {
+      var createRoot = false;
+      var createChild = false;
+
+      if (monitorRoot != null) {
+        final AllocationMonitorReference monitor = monitorRoot.children.first;
+        if (monitor.dateTime != monitorTimestamp) {
+          monitorRoot.removeLastChild();
```
jacob314, line 931: document why you remove the last child even though you compared against the first child.

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -992,33 +1140,118 @@ class MPChartData {
...
 
+/// Prepare data to plot in Event Chart.
+class MPEventsChartData {
+  bool _pruning = false;
+
+  /// Signal that every addTrace will cause a prune.
+  bool get pruned => _pruning;
+
+  /// Datapoint entries for ghost trace to stop auto-scaling of Y-axis.
+  final ghosts = <Entry>[];
```
jacob314, line 1151: it seems these lists all have the same # of entries which is a little bit surprising. Maybe document that and consider having 
List<List<Entry>> with enum ids for the various data point types.
As is, it looks like you would need to update a number of places in the code to add a new entry type.

packages/devtools_app/lib/src/memory/memory_events_pane.dart
```diff
@@ -0,0 +1,255 @@
...
+  void initState() {
+    super.initState();
+  }
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+
+    _memoryController = Provider.of<MemoryController>(context);
+    // TODO(jacobr): this is an ugly way to be using the theme. It would be
```
jacob314, line 51: TODO(terry) :)

packages/devtools_app/lib/src/memory/memory_events_pane.dart
```diff
@@ -0,0 +1,255 @@
...
+
+    _controller.setViewPortOffsets(
+        defaultSpacing * 3, denseSpacing, defaultSpacing, 0);
+  }
+
+  /// Override onNothingSelected and onValueSelected for selectionListener.
+  @override
+  void onNothingSelected() {
+    // TODO: implement onNothingSelected
+    print('>>>> onNothingSelected');
```
jacob314, line 121: resolve these TODOs and remove the prints.

packages/devtools_app/lib/src/memory/memory_events_pane.dart
```diff
@@ -0,0 +1,255 @@
...
+    setState(() {
+      // Display new events in the pane.
+      _updateEventPane();
+    });
+  }
+
+  /// Display any newly received events in the chart.
+  void _updateEventPane() {
+    setState(() {
+      _controller.data = ScatterData.fromList([
```
jacob314, line 235: Nit:
`final dataSets = [
 _gcVmDataSet,
        _allocationStartSet,
        _allocationContinueSet,
        _allocationResetSet,
        _gcUserDataSet,
        _snapshotDataSet,
        _snapshotAutoDataSet,
];
_controller.data = ScatterData.fromList(dataSets);
for (var dataSet in dataSets) {
  dataSet.notifyDataSetChanged();
}
```

File: packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart [hunk 19]
```diff
@@ -69,7 +73,8 @@ const knowClassesToAnalyzeForImages = <WildcardMatch, List<String>>{
 
   // Anything that exactly matches:
   WildcardMatch.exact: [
-    '_Int32List',
+    '_Int32List', // Emulator stores raw images as Int32List.
+    '_Int64List', // New devices e.g., Pixel 3 XL stores images as Int64List.
```
jacob314, line 77: wild guess: it isn't emulators vs new devices, it is 32 bit vs 64bit devices.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -211,6 +218,12 @@ class HeapTreeViewState extends State<HeapTree>
       ));
     });
 
+    addAutoDisposeListener(controller.monitorAllocationsChanged, () {
```
jacob314, line 221: calls to the controller shouldn't be wrapped in a call to setState. You can cleanup by listening to monitorAllocationsChanged directly from the controller
and listening for the libraries changing from directly in the Widget.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -469,48 +489,170 @@ class HeapTreeViewState extends State<HeapTree>
...
+                            controller.groupByTreeTable.dataRoots
+                                .every((element) {
+                              element.collapseCascading();
+                              return true;
+                            });
+                            if (controller.instanceFieldsTreeTable != null) {
+                              // We're collapsing close the fields table.
+                              controller.selectedLeaf = null;
+                            }
+                            setState(() {});
```
jacob314, line 508: not new but while you are tweaking this code, can we remove the empty setState? or wrap the setState call around the state that is modified?

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -469,48 +489,170 @@ class HeapTreeViewState extends State<HeapTree>
...
-                          setState(() {});
                         }
-                      }
-                    : null,
-                child: const Text('Collapse All'),
-              ),
+                      : null,
+                  child: const MaterialIconLabel(
+                    Icons.vertical_align_top,
+                    '',
```
jacob314, line 514: a label without text is a bit odd. You can look at the implementation of MaterialIconLabel to see how to just render the icon. Same for the ExpandAll case.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -469,48 +489,170 @@ class HeapTreeViewState extends State<HeapTree>
...
+
+        Tooltip(
+          message: 'Monitor Allocations',
+          child: OutlineButton(
+            key: allocationMonitorKey,
+            onPressed: () async {
+              await _allocationStart();
+            },
+            child: createImageIcon(
+              'icons/memory/communities_white@2x.png',
```
jacob314, line 551: is there no material icon we can use for this? It would be nice to not add more png icons. @cobblest I think you might have some suggestions.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -469,48 +489,170 @@ class HeapTreeViewState extends State<HeapTree>
...
+        ),
+        Tooltip(
+          message: 'Reset Accumulators',
+          child: OutlineButton(
+            key: allocationMonitorResetKey,
+            onPressed: () async {
+              await _allocationReset();
+            },
+            child: createImageIcon(
+              'icons/memory/reset_icon_white@2x.png',
```
jacob314, line 564: similarly, it would be nice to use a material icon for this as well. We should also show the label rather than a tooltip for these buttons as it isn't obvious enough what they are from the icons.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -469,48 +489,170 @@ class HeapTreeViewState extends State<HeapTree>
...
+                  (instancesCurrent - previousAllocation.instancesCurrent);
+          currentAllocation.bytesAccumulated =
+              previousAllocation.bytesAccumulated +
+                  (bytesCurrent - previousAllocation.bytesCurrent);
+
+          final instancesAccumulated = currentAllocation.instancesAccumulated;
+          final bytesAccumulated = currentAllocation.bytesAccumulated;
+
+          if (instancesAccumulated != 0 || bytesAccumulated != 0) {
+            debugLogger('previous,index=[$previousIndex][$currentIndex] '
```
jacob314, line 604: should this logging be removed? this seems like it is logging a ton of data given the if condition is usually true. if this is an exceptional case, explain what worrying condition occurred.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -469,48 +489,170 @@ class HeapTreeViewState extends State<HeapTree>
...
+          // Either a new class in currentAllocations or old that's no longer
+          // active in previousAllocations.
+          final currentClassId = currentAllocation.classRef.id;
+          final ClassHeapDetailStats first =
+              controller.monitorAllocations.firstWhere(
+            (element) => element.classRef.id == currentClassId,
+            orElse: () => null,
+          );
+          if (first != null) {
+            //It's a class that's gone?
```
jacob314, line 624: missing space and replace `?` with `.` as hopefully that is a statement about what is happening not a question.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -469,48 +489,170 @@ class HeapTreeViewState extends State<HeapTree>
...
+    controller.monitorTimestamp = allocationtimestamp;
+    controller.monitorAllocations = currentAllocations;
+  }
+
+  Future<void> _allocationReset() async {
+    controller.memoryTimeline.addMonitorResetEvent();
+    final currentAllocations = await controller.resetAllocationProfile();
+
+    // Reset all accumulators to zero.
+    currentAllocations.every((element) {
```
jacob314, line 648: nit:
```dart
for (var element in currentAllocations) {
    element.bytesAccumulated = 0;
    element.instancesAccumulated = 0;
}
```
is cleaner.

packages/devtools_app/lib/src/memory/memory_instance_tree_view.dart
```diff
@@ -0,0 +1,166 @@
...
+  MemoryController controller;
+
+  final TreeColumnData<FieldReference> treeColumn = _FieldTypeColumn();
+  final List<ColumnData<FieldReference>> columns = [];
+
+  @override
+  void initState() {
+    setupColumns();
+
+    super.initState();
```
jacob314, line 32: same as elsewhere, super should be called first for initState.

packages/devtools_app/lib/src/memory/memory_instance_tree_view.dart
```diff
@@ -0,0 +1,166 @@
...
+    //              listen to, so we don't need the setState calls.
+    // Update the chart when the memorySource changes.
+    addAutoDisposeListener(controller.selectedSnapshotNotifier, () {
+      setState(() {
+        controller.computeAllLibraries(rebuild: true);
+      });
+    });
+  }
+
+  void setupColumns() {
```
jacob314, line 56: if setupColumns is only called in initState, consider inlining it in initState.

packages/devtools_app/lib/src/memory/memory_instance_tree_view.dart
```diff
@@ -0,0 +1,166 @@
...
+  @override
+  String getDisplayValue(FieldReference dataObject) {
+    if (dataObject is ObjectFieldReference && !dataObject.isNull) {
+      // Real object that isn't Null value is empty string.
+      return '';
+    }
+
+    var value = getValue(dataObject);
+    if (value is String && value.length > 30) {
+      value = '${value.substring(0, 13)}…${value.substring(value.length - 17)}';
```
jacob314, line 149: why this is needed?
It would be nice to use standard Flutter string truncation with `...`
If truncating in the middle gives a much better result, I would prefer for us to at some point implement it robustly with text measurement so we actually truncate the correct # of characters rather than guessing that we can only show 30 characters.

packages/devtools_app/lib/src/memory/memory_protocol.dart
```diff
@@ -144,26 +140,74 @@ class MemoryTracker {
...
+    final keysToRemove = <String>[];
+
+    final isolateCount = isolateHeaps.length;
+    final keys = isolateHeaps.keys.toList();
+    for (var index = 0; index < isolateCount; index++) {
+      final isolateId = keys[index];
+      var usage = isolateHeaps[isolateId];
+      if (isolateCount > 1) {
+        // Is the isolate live (not sentinaled).
+        usage = await _isolateMemoryUsage(isolateId, usage);
```
jacob314, line 168: make this a different variable rather than reusing the usage object. 
this current logic seems a bit suspect or I am getting confused. it looks like we are assuming that if there is one isolate it can't be a sentinel which seems strange.

packages/devtools_app/lib/src/memory/memory_snapshot_models.dart
```diff
@@ -169,6 +208,10 @@ class Reference extends TreeNode<Reference> {
 
   final String name;
 
+  final bool isAllocation;
```
jacob314, line 211: document the difference between isAllocation and isAllocations.

packages/devtools_app/lib/src/memory/memory_timeline.dart
```diff
@@ -0,0 +1,483 @@
...
+      final javaHeap = sample.adbMemoryInfo.javaHeap.toDouble();
+      final nativeHeap = sample.adbMemoryInfo.nativeHeap.toDouble();
+      final code = sample.adbMemoryInfo.code.toDouble();
+      final stack = sample.adbMemoryInfo.stack.toDouble();
+      final graphics = sample.adbMemoryInfo.graphics.toDouble();
+      final other = sample.adbMemoryInfo.other.toDouble();
+      final system = sample.adbMemoryInfo.system.toDouble();
+      final total = sample.adbMemoryInfo.total.toDouble();
+
+      final graphicsEntry = Entry(
```
jacob314, line 288: track a y and use `+=` to simplify the code and make it easier to re-arange the order in the future.

packages/devtools_app/linux/.gitignore
```diff
@@ -1,7 +1 @@
-app_configuration.mk
-Makefile
-flutter
 flutter/ephemeral
```
jacob314, line 1: were these changes intentional?

packages/devtools_app/test/memory_screen_test.dart
```diff
@@ -159,4 +183,86 @@ void main() {
...
+
+    expect(snapshotButton.enabled, isFalse);
+    await tester.pumpAndSettle(const Duration(seconds: 3));
+
+    expect(
+      controller.selectedSnapshotTimestamp.millisecondsSinceEpoch,
+      lessThan(DateTime.now().millisecondsSinceEpoch),
+    );
+  });
 }
```
jacob314, line 268: a golden image test of this would be easy to add and would help check more of the state.

packages/devtools_shared/lib/src/memory/event_sample.dart
```diff
@@ -0,0 +1,166 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Displays heap object allocation monitoring (in the VM).  The mllocation
```
jacob314, line 5: mllocation --> allocation.
Add a blank line after the first line to make a proper class level comment.
The word "displays" confuses me a bit as this is more model logic than view logic. Maybe, "Controls heap object allocation monitoring..."?

packages/devtools_shared/lib/src/memory/event_sample.dart
```diff
@@ -0,0 +1,166 @@
...
+            ? AllocationAccumulator.fromJson(json['allocationAccumulatorEvent'])
+            : null,
+      );
+
+  Map<String, dynamic> toJson() => <String, dynamic>{
+        'timestamp': timestamp,
+        'gcEvent': isEventGC,
+        'snapshotEvent': isEventSnapshot,
+        'snapshotAutoEvent': isEventSnapshotAuto,
+        'allocationAccumulatorEvent': allocationAccumulator != null
```
jacob314, line 121: --> allocationAcumulator?.toJson()

packages/devtools_testing/lib/support/memory_test_data.dart
terrylucas: Good idea done.

packages/devtools_app/lib/src/memory/memory_allocation_table_data.dart
terrylucas: Changed to defaultNumberFieldWidth.

packages/devtools_app/lib/src/memory/memory_allocation_table_data.dart
terrylucas: I like Delta (it's short) I'll use that.  I will only show positive I think that more important, its showing leaks.  Maybe we could have a setting default that only shows growth unless turned off then displays both positive and negative growth?

terrylucas:
> We could also consider simplifying the design for events icons by using less color and shape combinations. I created some examples below:
> DevTools Snapshot Icons

I've changed the scatter plot shapes and colors to:

image

terrylucas:
> I have a few ideas:
> 
> 1. the Monitoring Allocation and Reset Accumulator icons are a bit confusing. It's hard to quickly realize what it means by looking at the icons. Shall we use text instead?
> 
> Youyang, the icons do have tooltips e.g., "Monitor Allocation" and "Reset Accumulator". What text would you suggest? One concern is the amount of horizontal space if the tooltip text was used. Would "Monitor" and "Reset" work?  Also, would you suggest I use the scatter plot shapes e.g., green hollow circle for Snapshot, red hollow circle for Auto-snapshot, yellow hollow circle for Monitor Allocations, glowing yellow filled circle for reset, blue hollow circle for GC?
> 
> 1. What's the purpose of having monitor allocation - Reset? Is it similar function with Snapshots (monitoring changes between snapshots)?
> 
> Reset is a technique that the VM/Observatory used, which our user study folks liked. Basically, clicking 'Reset' will reset accumulators to zero the next time a "Monitor Allocation" is clicked the Accumulators will show the number of objects allocated between last "Reset" and "Monitor Allocation". We could also display "Accumulators" as "Diff". I currently abbreviate Accumulators as "Accum" could also use "ACC". But I don't have a lot of horizontal space to have a column heading of "Accumulators". Any suggestions on this problem? Jacob suggestion "Delta" or "Difference". I like Delta (it's short) I'll use that. It only show growth (positive values negatives values displays as zero )I think that more important, its showing leaks. Maybe we could have a setting that default to only show growth unless turned off then displays both positive and negative growth?
> 
> Right now there are so many icons in the first screenshot and I am concerned that users would be confused about different types of events. It could be confusing especially because we use different color+shape combinations. From human attentions's perspective, it's will cost a lot of cognitive resources to process multiple features https://en.wikipedia.org/wiki/Feature_integration_theory
> 
> Your right, probably information overload! Saw your suggestion on glyphs in the event pane I'll respond to that suggestion.

packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_chart.dart
terrylucas: null implies the height of the box shouldn't be constrained.

packages/devtools_app/lib/src/memory/memory_controller.dart
terrylucas: Positive test is best. Done.

packages/devtools_app/lib/src/memory/memory_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_controller.dart
terrylucas: ||= isn't valid right?

packages/devtools_app/lib/src/memory/memory_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_controller.dart
terrylucas: They all have the same number of entries to keep the timeline (ticks) in proper sync with the main chart as we scroll around, zoom in/out, etc.  I think a better solution, if memory becomes a problem, would be to only expand the entries for the visible ticks.

However, List<List> with enum ids for each dataset is a good idea.

terrylucas: Fixed for all three charts (event, adb, and event pane).

packages/devtools_app/lib/src/memory/memory_events_pane.dart
terrylucas: onNothingSelected, etc. are derived from an abstract class will implement soon when next PR with legend is ready.

packages/devtools_app/lib/src/memory/memory_events_pane.dart
terrylucas: Rewrote all charting to handle list of datasets with enums code is simplified and more robust too.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: It's the device good guess bad comment.  Fixed.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: I'll need to change MaterialIconLabel to take a null or optional text since null doesn't work.

terrylucas: Added a constructor
MaterialIconLabel.noText(icon);

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: Added TODO will fix later.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: Again added TODO will check with Youyang.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: I've added this under a _debugAllocationMonitoring flag.  May remove after I'm happy that the monitoring is working well.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: // A class that no longer exist (live or sentinel).

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_instance_tree_view.dart
terrylucas: Fixed in all memory classes.

packages/devtools_app/lib/src/memory/memory_instance_tree_view.dart
terrylucas: Moved all setupColumns to initState.

packages/devtools_app/lib/src/memory/memory_instance_tree_view.dart
terrylucas: // TODO(terry): Can we use standard Flutter string truncation?
    //              Long variable names the beginning and end of the name is
    //              most significant the middle part of name less so.  However,
    //              we need a nice hover/tooltip to show the entire name.
    //              If middle, '...' then we need a more robust computation w/
    //              text measurement so we accurately truncate the correct # of
    //              characters instead guessing that we only show 30 chars.

packages/devtools_app/lib/src/memory/memory_timeline.dart
terrylucas: Done.

packages/devtools_app/linux/.gitignore
terrylucas: Yes is necessary for latest Linux desktop.

packages/devtools_app/test/memory_screen_test.dart
terrylucas: I'll add a golden image test post this PR that's a good idea.

packages/devtools_shared/lib/src/memory/event_sample.dart
terrylucas: Yeah display is the the end result the data is all stored with a HeapSample.

/// Monitor heap object allocations (in the VM).  The allocation monitor will
/// cause 'start' event exist in the HeapSample. Immediately afterwards a
/// 'continues' event is added on each subsequent timestamp tick (HeapSample)
/// until another monitor start event.  A 'reset' event stops the 'continues'
/// event for one timestamp tick with a 'reset' event. Immediately after the
/// reset event a 'continues' event will again appear in the HeapSample's
/// MemoryEventInfo - until a new monitor is started. One monitor exists per
/// VM connection.

packages/devtools_shared/lib/src/memory/event_sample.dart
terrylucas: Fixed.

terrylucas:
Jacob still have a few more suggestions to complete - PTAL most everything is done.

terrylucas:
> done with a first round of comments.

PTAL most of the suggestions are complete few little ones to fix.

packages/devtools/pubspec.yaml
terrylucas: done.

packages/devtools_shared/pubspec.yaml
terrylucas: reverted.

packages/devtools_app/lib/src/memory/memory_controller.dart
terrylucas: Removed function no longer used.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: The name was poorly choosen should be monitorAllocationsNotifier (its a ValueNotifier).

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: Fixed the empty setState.

packages/devtools_app/lib/src/memory/memory_protocol.dart
terrylucas: Thought I was being cute only checking if isolates > 1 however, should always do then in case something strange from the normal mechanism happens we'll still work.

packages/devtools_app/lib/src/memory/memory_snapshot_models.dart
terrylucas: /// Allocations are being monitored, user requested monitoring of any
  /// allocated memory.  Parent node containing all monitored classes.
  final bool isAllocations;

  /// A monitored class.
  final bool isAllocation;

packages/devtools_app/lib/src/connect_screen.dart
terrylucas: reverted.

packages/devtools_app/lib/src/memory/memory_allocation_table_data.dart
```diff
@@ -123,23 +127,25 @@ class FieldSizeColumn extends ColumnData<ClassHeapDetailStats> {
...
-      dataObject.bytesAccumulated;
+  int getValue(ClassHeapDetailStats dataObject) =>
+      dataObject.bytesDelta;
 
+  // TODO(terry): Only show grows (negative value returns 0). Consider setting
+  //              to display growth and decline.
   @override
   String getDisplayValue(ClassHeapDetailStats dataObject) =>
-      '${getValue(dataObject)}';
+      '${max(0, getValue(dataObject))}';
```
jacob314, line 148: why is there a max? I would have thought a negative delta is fine?

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -505,13 +502,14 @@ class HeapTreeViewState extends State<HeapTree>
                               // We're collapsing close the fields table.
                               controller.selectedLeaf = null;
                             }
-                            setState(() {});
+                            setState(() {
+                              controller.buildTreeFromAllData();
```
jacob314, line 506: is this setState call needed? we shouldn't have to wrap calls to the controller with a setState. Is there another piece of data from the controller we need the widget to listen on?

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -528,13 +526,14 @@ class HeapTreeViewState extends State<HeapTree>
                               element.expandCascading();
                               return true;
                             });
-                            setState(() {});
                           }
+                          setState(() {
```
jacob314, line 530: same question as above about setState call.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -528,13 +526,14 @@ class HeapTreeViewState extends State<HeapTree>
...
+                            controller.buildTreeFromAllData();
+                          });
                         }
                       : null,
-                  child: const MaterialIconLabel(
+                  child: const MaterialIconLabel.noText(
```
jacob314, line 535: Instead of adding MaterialIconLabel.noText you can just write
`Icon( Icons.vertical_align_bottom, size: 18.0)` or make a helper method with that default. We should probably set 18.0 as our default icon size somewhere.

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -1131,41 +1132,56 @@ class MPChartData {
...
-  final monitorStart = <Entry>[];
-  final monitorContinues = <Entry>[];
-  final monitorReset = <Entry>[];
+  final datasets = List<List<Entry>>.generate(
+    EventDataSets.values.length,
+    (int) => <Entry>[],
+  );
+
+  /// Compute each dataset Entry's.
+  List<Entry> get ghosts => datasets[EventDataSets.ghostsSet.index];
```
jacob314, line 1176: nice use of dart enums!

packages/devtools_app/lib/src/memory/memory_allocation_table_data.dart
terrylucas: Here's the comment I left to Youyang about Monitor/Reset.

Reset is a technique that the VM/Observatory used, which our user study folks liked. Basically, clicking 'Reset' will reset accumulators to zero the next time a "Monitor Allocation" is clicked the Accumulators will show the number of objects allocated between last "Reset" and "Monitor Allocation". We could also display "Accumulators" as "Diff". I currently abbreviate Accumulators as "Accum" could also use "ACC". But I don't have a lot of horizontal space to have a column heading of "Accumulators". Any suggestions on this problem? Jacob suggestion "Delta" or "Difference". I like Delta (it's short) I'll use that.

It only show growth (positive values negatives values displays as zero this is what observatory did) I think that's more important, to show leaks. Maybe we could have a setting that defaults to only show growth unless turned off then displays both positive and negative growth?

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: Good idea will create an createIcon helper.

packages/devtools_app/lib/src/memory/memory_allocation_table_data.dart
jacob314: If we show negative values they should definitely be sorted at the end.
I think there is some prior art for showing negative deltas as well. For example I believe this tool supports it.
https://www.yourkit.com/docs/java/help/compare_snapshots.jsp
Negative deltas aren't typically as actionable but could be useful if you want to sanity check that a bunch of memory was freed like you expected when for example navigating to a different route in your flutter app.

terrylucas: I agree I'll add a TODO.  What do we think about adding this is a setting?

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: I've fixed all the setState calls.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: Fixed.

Issue #2178, opened 2020-07-13
Can't zoom into diff tree map cells


kenzieschmoll:
When the code size diff tree map is loaded, I am unable to drill into a cell. I can verify in the tree table that this cell has children, so clicking the cell should re-root the tree.

![cant_drill_into_diff](https://user-images.githubusercontent.com/43759233/87351587-774d7380-c50e-11ea-97f4-aaf26b7f15b4.gif)

peterdjlee:
Fixed in #2181. Closing issue.

PR #2181, opened 2020-07-14
Fix zoom for diff page and add tests

peterdjlee:
* Fix zoom for diff page and add tests to address #2178


kenzieschmoll:
Is there a file missing from this PR? I'm only seeing the test change.

peterdjlee:
> Is there a file missing from this PR? I'm only seeing the test change.

My mistake! Didn't git add the file. Should be good now.

Issue #2044, opened 2020-06-05
Monitor grpc traffic via devtools


mehmetf:
Currently devtools can monitor HTTP traffic. Some transport implementations (in particular package:grpc) operate directly on dart:io Socket. The ideal goal of this would be:

- Monitor socket traffic.
- Be able to interpret and decode grpc traffic (protos) and show details about the payload in the tooling

@yohanln has mentioned this would be a good intern project.

mehmetf:
Also @jacob314 and @bkonyi as FYI

bkonyi:
@zichangg has actually already implemented socket profiling in dart:io, I just haven't gotten around to surfacing it in DevTools yet (I'll hopefully get around to it in Q3).

zichangg:
Right! But my implementation is only monitoring statistics of socket(without looking at the content). It probably needs more work for grpc.

thedejifab:
Is anyone working on this at the moment? Currently building stuff using GRPC and I need to monitor my traffic as I do with REST API calls but this is a blocker.

zichangg:
We will start the work very soon!

thedejifab:
Cool @zichangg. I think I'm actually interested in working on this

kenzieschmoll:
There is an in-progress CL for this. @bkonyi is going to pick up on the existing work when he has some bandwidth.

kenzieschmoll:
Some work for this effort was landed here: https://github.com/grpc/grpc-dart/pull/392, but we need to audit what work is left to be done to support debugging gRPC requests with DevTools

EnesKaraosman:
Any update here?

CoderDake:
Currently at the back of a long backlog.

loeffel-io:
+1

srawlins:
@kenzieschmoll: @bkonyi and I were thinking this might be better suited to be a DevTools extension. WDYT?

PR #2191, opened 2020-07-17
Add socket profiling to Network page

kenzieschmoll:
This CL
- adds socket information to the Network profiler
- adds an Overview UI that displays general info for both HTTP and WS requests
- removes the "Timing" tab in the network request inspector, as that information is now in the "Overview" tab

Table with both HTTP and WS requests and overview for selected HTTP request:
Screen Shot 2020-07-17 at 3 19 06 PM
Overview for socket:
Screen Shot 2020-07-17 at 1 52 26 PM

TODO before submitting: add tests and general cleanup


packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -99,17 +69,29 @@ class HttpRequestData {
     return data;
   }
 
-  final int _timelineMicrosBase;
+  static const _connectionInfoKey = 'connectionInfo';
+  static const _contentTypeKey = 'content-type';
+  static const _cookieKey = 'cookie';
+  static const _errorKey = 'error';
+  static const _filterKey = 'filterKey';
```
jacob314, line 76: nit: normalize the key names. This one has the word `Key` in the value but others don't.

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -99,17 +69,29 @@ class HttpRequestData {
...
+  static const _localPortKey = 'localPort';
+  static const _methodKey = 'method';
+  static const _requestHeadersKey = 'requestHeaders';
+  static const _responseHeadersKey = 'responseHeaders';
+  static const _statusCodeKey = 'statusCode';
+  static const _setCookieKey = 'set-cookie';
```
jacob314, line 82: make this key camelCase?

jacob314: perhaps add a TODO and file a bug upstream on the VM to cleanup the names.

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -135,32 +132,38 @@ class HttpRequestData {
 
   /// A map of general information associated with an HTTP request.
   Map<String, dynamic> get general {
```
jacob314, line 134: cache the value here so we stop creating a ton of copies.  Otherwise the code in getters like `port` is a bit wasteful as we generate the map twice and the only use a bool from it.

packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -126,46 +131,68 @@ class NetworkController {
         invalidRequests.add(requestData);
       }
     }
-    return HttpRequests(
+
+    currentValues.removeWhere((value) => value is WebSocket);
```
jacob314, line 135: nit: rather than removing all WebSocket, maybe just don't add them in the first place. it looks like we are adding them a couple lines earlier.

packages/devtools_app/lib/src/http/http_request_data.dart
kenzieschmoll: filterKey is actually the name of the key in the arguments map. This would need to be changed on the VM but I think it is actually an accurate description as is because it is describing a key to group events on.

packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -203,19 +234,21 @@ class NetworkController {
...
-  /// Clears the previously collected HTTP timeline events and resets the last
-  /// refresh timestamp to the current time.
+  /// Clears the previously collected HTTP timeline events, clears the socket
+  /// profile from the vm, and resets the last refresh timestamp to the current
+  /// time.
   Future<void> clear() async {
     await _networkService.updateLastRefreshTime();
-    _requests.value = HttpRequests();
-    _selectedHttpRequest.value = null;
+    await networkService.clearSocketProfile();
```
jacob314, line 250: can we have one networkService method that does all the clearing rather than having updateLastRefreshTime and clearSockerProfile?

packages/devtools_app/lib/src/http/http_request_data.dart
kenzieschmoll: added a TODO and filed a bug. Names need to be changed upstream

packages/devtools_app/lib/src/http/http_request_data.dart
kenzieschmoll: done

packages/devtools_app/lib/src/network/network_controller.dart
kenzieschmoll: At this point, `currentValues` contains all the current requests we have in the profiler, which will contain ws requests if they exist. When we refresh the network data after polling, the new data may contain web sockets with the same ids as ones we already have. For this reason, we need to remove the current web sockets and replace with the updated data in the profile. We could be smarter about this and keep unchanged sockets in the list if we need to.

packages/devtools_app/lib/src/network/network_controller.dart
kenzieschmoll: done

packages/devtools_app/lib/src/network/network_controller.dart
jacob314: sg. can you add a comment to that effect?

kenzieschmoll: done

kenzieschmoll:
Addresses part of https://github.com/flutter/devtools/issues/2044

PR #2193, opened 2020-07-19
Fix RTL alignment / VerticalDirection bug on Layout Explorer

### LINKED ISSUE #2138: RTL alignment in layout explorer
Dart Devtool layout explorer is miss alignment with locale and RTL option 
locale option in the code is :

`locale: Locale('ar', 'AE'),`

main direction for all of the app is RTL

DevTool:
image

UI:
image

so the main problem is layout explorer should be mirrored so result would be true.

Of course thanks for the great new DevTool.

adalberht:
This CL should fixes #2138 

@jacob314


packages/devtools_app/lib/src/inspector/inspector_data_models.dart
```diff
@@ -370,6 +370,9 @@ class FlexLayoutProperties extends LayoutProperties {
   final VerticalDirection verticalDirection;
   final TextBaseline textBaseline;
 
+  @override
+  List<LayoutProperties> get children => flipMainAxis ? super.children.reversed.toList() : super.children;
```
jacob314, line 374: maybe leave children in the real order and add a second getter called 
displayChildren that are in the display order? You should be able to rename children n the parent class to displayChildren and then add a new children getter to the parent class to make this change without having to do much manual work.

packages/devtools_app/lib/src/inspector/inspector_data_models.dart
```diff
@@ -417,6 +420,36 @@ class FlexLayoutProperties extends LayoutProperties {
     return height + overflowEpsilon < max(childrenHeights);
   }
 
+  /// Method and logic is taken from [RenderFlex] on `rendering/flex.dart`
+  bool _flipMainAxis;
+  bool get flipMainAxis {
+    if (_flipMainAxis != null) return _flipMainAxis;
+    assert(direction != null);
+    final startIsTopLeft = () {
```
jacob314, line 428: make this a method instead of a closure.
e.g
`bool startIsTopLeft() {`

adalberht:
Hey Jacob, sorry for my inactivity on this PR. I was busy with getting involved on hosting the Flutter Indonesia hackathon for the past month. I just got the time to work on this PR again but I need more time on fixing and writing the proper tests.
I'll let you know once this PR is ready for review again!

packages/devtools_app/test/layout_explorer/layout_explorer_serialization_delegate.dart
```diff
@@ -0,0 +1,69 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/rendering.dart';
+import 'package:flutter/widgets.dart';
+
+/// This class is merely a duplicate of the InspectorSerializationDelegate inside the `assets/scripts/inspector_polyfill_script.dart`.
+/// This delegate is used for testing Widget Diagnostics Node so that we don't have to create manual JSON each time we want to create new test cases.
+/// TODO(adalberht): Ask Jacob on what's the better solution on code reuse between the Layout Explorer polyfillscripts and the code inside test package?
```
adalberht, line 11: I wanted to reuse some code on the polyfill scripts but I'm not sure on how to do that properly without cloning/copy-pasting them.
This is useful so that our unit tests can just generate the Layout Explorer Remote Diagnostics Node based on actual widgets instead of us creating JSON data for each tests cases that we want to create.
But let me know if that's not preferrable! To be honest writing it this way is way makes writing the tests more concise and pleasant :D

packages/devtools_app/test/layout_explorer/layout_explorer_test_utils.dart
```diff
@@ -0,0 +1,30 @@
...
+  final element = find.byWidget(widget).evaluate().first;
+  final nodeJson =
+  element.toDiagnosticsNode(style: DiagnosticsTreeStyle.dense).toJsonMap(
+    LayoutExplorerSerializationDelegate(
+      subtreeDepth: subtreeDepth,
+      service: WidgetInspectorService.instance,
+    ),
+  );
+  return RemoteDiagnosticsNode(nodeJson, null, null, null);
+}
```
adalberht, line 30: This is what I meant by my previous comment

packages/devtools_app/test/layout_explorer/layout_explorer_serialization_delegate.dart
jacob314: Clever idea! Lets land like this and we can add a build script to copy the code so it can be kept completely in sync. The other way we could solve this is by tests communicating with a remote Dart vm like in the inspector_service tests. That has the downside that the tests are slower to run.

packages/devtools_app/test/layout_explorer/layout_explorer_test_utils.dart
jacob314: I like it!

jacob314:
You will want to remove the draft tag and fix the couple of lint errors causing the bots to fail. 
For the use of a protected member error, you can use the quick fix to allow that lint error for that line.

packages/devtools_app/test/layout_explorer/layout_explorer_serialization_delegate.dart
adalberht: Awesome, thanks, Jacob! Alright then, I agree maybe let's land like this and l will try to look on the alternative testing methods with remote Dart VM that you mentioned

adalberht:
Hey Jacob, sorry that I missed out on some test cases on my previous version of the PR, apparently we also need to flip the main axis alignment because that affects the calculation on the free spaces.
Can you please kindly take a look at it again? I added some lines on the main axis and free spaces calculation and also added some more unit tests for it.

adalberht:
Awesome, thanks @jacob314 !
I didn't have write access to the repository, so you can merge it whenever you want! :D

PR #888, opened 2019-08-07
Memory profiler fixes.

terrylucas:
Few fixes for re-entrancy, faster instance search, selection and indicators of work processing.

PR #2195, opened 2020-07-20
Allow theme to be overridden by background color in querystring

DanTup:
This adds a `ThemeOverrides` class that is created at the very top level and passed into `themeFor()` to allow the theme to be influenced by these values. Currently this is only populated in the web implementation.

Rather than passing the theme, the theme is inferred from the background color  - if all RGB values are < 50 (decimal) or > 205 it will switch to Dark/Light mode. Otherwise, the custom background color is ignored entirely (to avoid accidentally using colours that may not have enough contrast). I'm not sure if these specific values are reasonable though.

This also sets the body background immediately, so if there is a delay in the Flutter app rendering, we don't see a quite page in the meantime (this is different to the existing code that appears to do something similar for the theme preference, which is for avoiding using the wrong theme initially if the preference is not the same as the default).

### DIFF
--- a/packages/devtools_app/lib/main.dart
+++ b/packages/devtools_app/lib/main.dart
@@ -7,6 +7,7 @@ import 'package:flutter/material.dart';
 import 'src/app.dart';
 import 'src/config_specific/framework_initialize/framework_initialize.dart';
 import 'src/config_specific/load_fallback_app/load_fallback_app.dart';
+import 'src/config_specific/theme_overrides/theme_overrides.dart';
 import 'src/preferences.dart';
 
 void main() async {
@@ -49,6 +50,8 @@ void main() async {
     };
   }
 
+  final themeOverrides = ThemeOverrides.load();
+
   final preferences = PreferencesController();
   // Wait for preferences to load before rendering the app to avoid a flash of
   // content with the incorrect theme.
@@ -58,6 +61,6 @@ void main() async {
 
   // Now run the app.
   runApp(
-    DevToolsApp(defaultScreens, preferences),
+    DevToolsApp(defaultScreens, preferences, themeOverrides),
   );
 }
--- a/packages/devtools_app/lib/main.dart
+++ b/packages/devtools_app/lib/main.dart
@@ -8,7 +8,6 @@ import 'src/app.dart';
 import 'src/config_specific/framework_initialize/framework_initialize.dart';
 import 'src/config_specific/ide_theme/ide_theme.dart';
 import 'src/config_specific/load_fallback_app/load_fallback_app.dart';
-import 'src/config_specific/theme_overrides/theme_overrides.dart';
 import 'src/preferences.dart';
 
 void main() async {
--- a/packages/devtools_app/lib/main.dart
+++ b/packages/devtools_app/lib/main.dart
@@ -51,7 +51,7 @@ void main() async {
     };
   }
 
-  final ideTheme = IdeTheme.load();
+  final ideTheme = getIdeTheme();
 
   final preferences = PreferencesController();
   // Wait for preferences to load before rendering the app to avoid a flash of
--- a/packages/devtools_app/lib/main.dart
+++ b/packages/devtools_app/lib/main.dart
@@ -6,6 +6,7 @@ import 'package:flutter/material.dart';
 
 import 'src/app.dart';
 import 'src/config_specific/framework_initialize/framework_initialize.dart';
+import 'src/config_specific/ide_theme/ide_theme.dart';
 import 'src/config_specific/load_fallback_app/load_fallback_app.dart';
 import 'src/config_specific/theme_overrides/theme_overrides.dart';
 import 'src/preferences.dart';
@@ -50,7 +51,7 @@ void main() async {
     };
   }
 
-  final themeOverrides = ThemeOverrides.load();
+  final ideTheme = IdeTheme.load();
 
   final preferences = PreferencesController();
   // Wait for preferences to load before rendering the app to avoid a flash of
@@ -61,6 +62,6 @@ void main() async {
 
   // Now run the app.
   runApp(
-    DevToolsApp(defaultScreens, preferences, themeOverrides),
+    DevToolsApp(defaultScreens, preferences, ideTheme),
   );
 }
--- a/packages/devtools_app/lib/src/config_specific/ide_theme/ide_theme_desktop.dart
+++ b/packages/devtools_app/lib/src/config_specific/ide_theme/ide_theme_desktop.dart
@@ -2,15 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be found
 // in the LICENSE file.
 
-import 'package:flutter/widgets.dart';
+import 'ide_theme.dart';
 
-/// IDE-supplied theming.
-class IdeTheme {
-  IdeTheme({this.backgroundColor, this.foregroundColor, this.fontSize});
-
-  factory IdeTheme.load() => IdeTheme();
-
-  Color backgroundColor;
-  Color foregroundColor;
-  double fontSize;
-}
+/// Load any IDE-supplied theming.
+IdeTheme getIdeTheme() => null;
--- a/packages/devtools_app/lib/src/config_specific/ide_theme/ide_theme_desktop.dart
+++ b/packages/devtools_app/lib/src/config_specific/ide_theme/ide_theme_desktop.dart
@@ -4,12 +4,11 @@
 
 import 'package:flutter/widgets.dart';
 
-/// Environment-specific theme overrides, for example IDE-provided
-/// theming for embedded mode.
-class ThemeOverrides {
-  ThemeOverrides({this.backgroundColor, this.foregroundColor, this.fontSize});
+/// IDE-supplied theming.
+class IdeTheme {
+  IdeTheme({this.backgroundColor, this.foregroundColor, this.fontSize});
 
-  factory ThemeOverrides.load() => ThemeOverrides();
+  factory IdeTheme.load() => IdeTheme();
 
   Color backgroundColor;
   Color foregroundColor;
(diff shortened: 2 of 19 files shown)


packages/devtools_app/lib/main.dart
```diff
@@ -6,9 +6,12 @@ import 'package:flutter/material.dart';
 
 import 'src/app.dart';
 import 'src/config_specific/framework_initialize/framework_initialize.dart';
+import 'src/config_specific/theme_overrides/theme_overrides.dart';
```
jacob314, line 9: maybe rename theme_overrides to ide_theme as that is exactly what it is.
It is the theme of the IDE that we are embedding devtools in / using devtools with?

packages/devtools_app/lib/src/config_specific/theme_overrides/theme_overrides_desktop.dart
```diff
@@ -0,0 +1,17 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be found
+// in the LICENSE file.
+
+import 'package:flutter/widgets.dart';
+
+/// Environment-specific theme overrides, for example IDE-provided
+/// theming for embedded mode.
+class ThemeOverrides {
```
jacob314, line 9: move this class over to a shared location rather than duplicating it each conditional import.
Dart is really bad at type checking conditional imports so keeping the API imported conditionally small is important.

packages/devtools_app/lib/src/config_specific/theme_overrides/theme_overrides_web.dart
```diff
@@ -0,0 +1,81 @@
...
+
+    final overrides = ThemeOverrides(
+      backgroundColor: _tryParseColor(queryParams['backgroundColor']),
+      foregroundColor: _tryParseColor(queryParams['foregroundColor']),
+      fontSize: _tryParseDouble(queryParams['fontSize']),
+    );
+
+    // If the environment has provided a background color, set it immediately
+    // to avoid a white page until the first Flutter frame is rendered.
+    if (overrides.backgroundColor != null) {
```
jacob314, line 25: for this to work we need to also make sure that the main dart.js script is loaded synchronously. That should be fine to enforce as DevTools isn't at all useful until the script is loaded as we aren't a PWA.

packages/devtools_app/lib/src/config_specific/theme_overrides/theme_overrides_web.dart
```diff
@@ -0,0 +1,81 @@
...
+
+  static double _tryParseDouble(String input) {
+    try {
+      if (input != null) {
+        return double.parse(input);
+      }
+    } catch (e) {
+      // The user can manipulate the query string so if the value is invalid
+      // print the value but otherwise continue.
+      print(e);
```
jacob314, line 77: use logging and print a more descriptive message than e.

packages/devtools_app/lib/src/config_specific/theme_overrides/theme_overrides_web.dart
```diff
@@ -0,0 +1,81 @@
...
+        final fullHex = input
+            .replaceAll('#', '') // Remove any leading #
+            .replaceAll('%23', '') // Ignore over-escaped # too
+            .padLeft(8, 'f');
+        return Color(int.parse(fullHex, radix: 16));
+      }
+    } catch (e) {
+      // The user can manipulate the query string so if the value is invalid
+      // print the value but otherwise continue.
+      print(e);
```
jacob314, line 64: use logging and print a more descriptive message than e.

packages/devtools_app/lib/src/config_specific/theme_overrides/theme_overrides_web.dart
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  Color backgroundColor;
+  Color foregroundColor;
+  double fontSize;
+
+  static Map<String, String> _loadQueryParams() {
+    // The querystring could be in one of two places:
+    // http://localhost:123/?backgroundColor=00000
+    // http://localhost:123/#/?backgroundColor=00000
```
jacob314, line 40: should it be considered a bug that we support both?  it would be nice to just support standard url query parameters.

packages/devtools_app/lib/src/config_specific/theme_overrides/theme_overrides_web.dart
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  static Color _tryParseColor(String input) {
+    try {
+      if (input != null) {
+        final fullHex = input
+            .replaceAll('#', '') // Remove any leading #
+            .replaceAll('%23', '') // Ignore over-escaped # too
+            .padLeft(8, 'f');
+        return Color(int.parse(fullHex, radix: 16));
```
jacob314, line 59: doesn't really matter but technically if you get:
#888
that should be converted to
#ff888888
not 
#fffff888

packages/devtools_app/lib/src/config_specific/theme_overrides/theme_overrides_web.dart
DanTup: It's currently imported in `packages/devtools_app/web/index.html` just before the closing `</body>` without any async/defer tag:

```
<script src="main.dart.js" type="application/javascript"></script>
```

Is that correct?

packages/devtools_app/lib/src/config_specific/theme_overrides/theme_overrides_web.dart
DanTup: That's how I originally did it, but it seems that when the main method ran Flutter had already redirected us to the second. I thought it would be best to support both in case anything changed in Flutter that might affect this (it doesn't seem like parsing the raw querystring like this is a particularly Flutter-like thing to do, so I wasn't sure if we should rely on exactly how it behaved).

If we just want one, the second one is probably the right one. WDYT?

packages/devtools_app/lib/main.dart
DanTup: Yeah, makes sense - I was trying to make it a bit more generic, but I'm not sure it really was. Done!

packages/devtools_app/lib/src/config_specific/theme_overrides/theme_overrides_desktop.dart
DanTup: I did this by reusing the class definition in `_stub.dart` and then just making a function that loads the IDE theme. If this isn't what you had in mind or I can do it better, let me know!

packages/devtools_app/lib/src/config_specific/theme_overrides/theme_overrides_web.dart
DanTup: Done!

packages/devtools_app/lib/src/config_specific/theme_overrides/theme_overrides_web.dart
DanTup: Done!

packages/devtools_app/lib/src/config_specific/theme_overrides/theme_overrides_web.dart
DanTup: Oops, good point! Since these can come from user-defined themes, I don't know if VS Code would clean them up, so it's probably important to support. I've moved this function to `utils.dart` and fixed to support 3-digit colours w/ tests. Thanks!

packages/devtools_app/lib/src/config_specific/ide_theme/ide_theme_desktop.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be found
+// in the LICENSE file.
+
+import 'ide_theme.dart';
+
+/// Load any IDE-supplied theming.
+IdeTheme getIdeTheme() => null;
```
jacob314, line 8: dart style: methods should not start with the word get.

packages/devtools_app/lib/src/config_specific/ide_theme/ide_theme_web.dart
```diff
@@ -0,0 +1,78 @@
...
+    backgroundColor: _tryParseColor(queryParams['backgroundColor']),
+    foregroundColor: _tryParseColor(queryParams['foregroundColor']),
+    fontSize: _tryParseDouble(queryParams['fontSize']),
+  );
+
+  // If the environment has provided a background color, set it immediately
+  // to avoid a white page until the first Flutter frame is rendered.
+  if (overrides.backgroundColor != null) {
+    document.body.style.backgroundColor =
+        overrides.backgroundColor.value.toRadixString(16).padLeft(8, '0');
```
jacob314, line 27: don't you need a '#' at the start of the color string?

jacob314:
lgtm pending fixing the missing `#`

packages/devtools_app/lib/src/config_specific/ide_theme/ide_theme_desktop.dart
DanTup: Any preference to renaming this (`parseIdeTheme`? `readIdeTheme`?) versus making it a getter?

My (minor) concern with making it a getter, is that it would be easy to accidentally use it globally from any code (rather than the reference passed down the tree from `main.dart`). With a method, that might be a little more obvious.

Right now it doesn't matter much (as long as the getter caches the result so it's not being re-read), but if we wanted to trigger rebuilds if it changed in future, having code reaching out to the global getter seems a bit weird.

packages/devtools_app/lib/src/config_specific/ide_theme/ide_theme_web.dart
DanTup: Yep - apparently my testing of this was bad because the scaffold color was being set and I didn't notice the real bg was not.

While fixing this, I noticed there's a different in the position of alpha in CSS colors (#RRGGBBAA) and the order of Flutter's bytes (alpha is high bits) so I also handled this in case any editors pass through a 4/8 digit hex code (we would've previously interpreted this incorrectly).

PR #2196, opened 2020-07-21
Move clear button and diff tree type dropdown up to the tab bar level

peterdjlee:
* Move clear button and diff tree type dropdown up to the tab bar level.
Screen Shot 2020-07-20 at 10 48 09 PM


packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -82,10 +83,25 @@ class CodeSizeBodyState extends State<CodeSizeBody>
...
+        snapshotRoot = controller.snapshotRoot.value;
+      });
+    });
+
+    diffRoot = controller.diffRoot.value;
+    addAutoDisposeListener(controller.diffRoot, () {
+      setState(() {
+        diffRoot = controller.diffRoot.value;
+      });
+    });
```
kenzieschmoll, line 99: why did we move this state up the tree? Now if either root changes, both the snapshot and diff view will rebuild (even if it is not the active view)

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -98,24 +114,70 @@ class CodeSizeBodyState extends State<CodeSizeBody>
               controller: _tabController,
               tabs: tabs,
             ),
+            Row(
+              children: [
+                _buildDiffTreeTypeDropdown(currentTab),
```
kenzieschmoll, line 119: shouldn't we only show this if the current tab is the diff tab?

peterdjlee: Done

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -102,6 +102,7 @@ class CodeSizeBodyState extends State<CodeSizeBody>
   @override
   Widget build(BuildContext context) {
     final currentTab = tabs[_tabController.index];
+    final isSnapView = currentTab.key == snapshotTabKey;
```
kenzieschmoll, line 105: rename to `isSnapshotView`

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -98,24 +115,67 @@ class CodeSizeBodyState extends State<CodeSizeBody>
               controller: _tabController,
               tabs: tabs,
             ),
+            Row(
+              children: [
+                if (!isSnapView) _buildDiffTreeTypeDropdown(),
```
kenzieschmoll, line 120: probably better to check that it is the diff view here in case we end up with other views that also are "not the snap view". We can rename to `isDiffView` everywhere instead

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -98,24 +115,67 @@ class CodeSizeBodyState extends State<CodeSizeBody>
...
+    return DropdownMenuItem<DiffTreeType>(
+      value: diffTreeType,
+      child: Text(diffTreeType.display),
+    );
+  }
+
+  Widget buildClearButton(bool isSnapView) {
+    return clearButton(
+      busy: isSnapView ? snapshotRoot == null : diffRoot == null,
+      onPressed: isSnapView ? controller.clearSnapshot : controller.clearDiff,
```
kenzieschmoll, line 169: can we track which view is active in the controller, and then if you call `controller.clear()`, the controller will clear which ever view needs to be cleared. This will allow us to push the state back down the tree to where it was, and reduce the work of the UI code to manage state.

peterdjlee: Done

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -84,25 +77,14 @@ class CodeSizeBodyState extends State<CodeSizeBody>
...
-    });
+    addAutoDisposeListener(controller.activeDiffTreeType);
 
-    diffRoot = controller.diffRoot.value;
-    addAutoDisposeListener(controller.diffRoot, () {
-      setState(() {
-        diffRoot = controller.diffRoot.value;
-      });
-    });
+    addAutoDisposeListener(controller.activeViewKey);
```
kenzieschmoll, line 82: We don't need to listen to this. The controller just needs to be updated on tapping to switch tabs.

packages/devtools_app/lib/src/code_size/code_size_controller.dart
```diff
@@ -87,10 +88,28 @@ class CodeSizeController {
...
-      diffTreeType: newDiffTreeType,
     );
   }
 
+  /// The active tree type used to build the diff treemap.
+  ValueListenable<Key> get activeViewKey {
+    return _activeViewKey;
+  }
+
+  final _activeViewKey = ValueNotifier<Key>(CodeSizeBodyState.snapshotTabKey);
```
kenzieschmoll, line 99: we shouldn't need this to be a value notifier. Just a boolean that is tracked on the controller. UI code shouldn't need to listen to this

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -84,25 +77,14 @@ class CodeSizeBodyState extends State<CodeSizeBody>
...
   }
 
   @override
   Widget build(BuildContext context) {
-    final currentTab = tabs[_tabController.index];
-    final isSnapView = currentTab.key == snapshotTabKey;
+    final isDiffView = controller.activeViewKey.value == diffTabKey;
```
kenzieschmoll, line 87: revert to checking the tab controller's index

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -163,36 +146,39 @@ class CodeSizeBodyState extends State<CodeSizeBody>
     );
   }
 
-  Widget buildClearButton(bool isSnapView) {
+  Widget buildClearButton(bool isDiffView) {
```
kenzieschmoll, line 149: we don't need the isDiffView param

packages/devtools_app/lib/src/code_size/code_size_controller.dart
```diff
@@ -87,10 +88,28 @@ class CodeSizeController {
...
+
+  void changeActiveView(Key activeViewKey) {
+    _activeViewKey.value = activeViewKey;
+  }
+
+  void clear() {
```
kenzieschmoll, line 105: can we pass in the active view key as a parameter to clear()?

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -64,9 +60,7 @@ class CodeSizeBodyState extends State<CodeSizeBody>
   void initState() {
     super.initState();
     _tabController = TabController(length: tabs.length, vsync: this)
```
kenzieschmoll, line 62: wrap this in `addAutoDisposeListener`
Example: https://github.com/flutter/devtools/pull/2197

peterdjlee: Done

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -144,18 +150,22 @@ class CodeSizeBodyState extends State<CodeSizeBody>
...
+  const SnapshotView();
+
   @override
-  _SnapshotViewState createState() => _SnapshotViewState();
+  SnapshotViewState createState() => SnapshotViewState();
 }
 
-class _SnapshotViewState extends State<SnapshotView> with AutoDisposeMixin {
-  static const treemapKey = Key('Code Size Treemap');
+class SnapshotViewState extends State<SnapshotView> with AutoDisposeMixin {
```
kenzieschmoll, line 166: leave this class private and move const key into StatefulWidget `SnapshotView`

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -45,8 +45,13 @@ class CodeSizeBody extends StatefulWidget {
...
     with AutoDisposeMixin, SingleTickerProviderStateMixin {
-  static const snapshotTabKey = Key('Code Size Snapshot Tab');
-  static const diffTabKey = Key('Code Size Diff Tab');
+  static const snapshotTabKey = Key('Snapshot Tab');
+  static const diffTabKey = Key('Diff Tab');
+  
+  @visibleForTesting
+  static const clearButtonKey = Key('Clear Button');
+  @visibleForTesting
+  static const dropdownKey = Key('Diff Tree Type Dropdown');
```
kenzieschmoll: move these const keys into the widget class `CodeSizeBody`. The CodeSizeBodyState can stay private

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -257,12 +267,15 @@ class _SnapshotViewState extends State<SnapshotView> with AutoDisposeMixin {
...
+  const DiffView();
+
   @override
-  _DiffViewState createState() => _DiffViewState();
+  DiffViewState createState() => DiffViewState();
 }
 
-class _DiffViewState extends State<DiffView> with AutoDisposeMixin {
-  static const treemapKey = Key('Code Size Treemap');
+class DiffViewState extends State<DiffView> with AutoDisposeMixin {
```
kenzieschmoll, line 276: same comment here

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -27,6 +27,18 @@ class CodeSizeScreen extends Screen {
 
   static const id = 'codeSize';
 
+  static const snapshotTabKey = Key('Snapshot Tab');
+  static const diffTabKey = Key('Diff Tab');
+
+  @visibleForTesting
```
kenzieschmoll, line 33: I think we can remove the visibleForTesting annotation since the keys are now used outside of this class, so they'd need to be public anyway

peterdjlee: I think for now the only keys that we need outside are snapshotTabKey and diffTabKey. Should I still remove the annotation for the rest of the keys?

kenzieschmoll: Fine to leave as is

PR #2201, opened 2020-07-23
Hide unavailable screens from the navigation

DanTup:
Turns out the race here was introduced by [my embed changes](https://github.com/flutter/devtools/pull/1995/files#diff-9b4fa0fbc9e2d26e0f66b7d4b1b59eccR128-L131).. I'd moved the screen filtering out of a function that was being called after the connection to before.

This seems to work correctly now - however there was only one screen where we showed a message saying it wasn't available for Flutter (the Inspector screen). All of the others seem to "work" for Dart apps (as in, they render - I don't know whether it's possible to get data to the - for example the Timeline in a non-Flutter app?).

Are there any other screens that should be hidden? (Currently a Dart app will start off on the Timeline screen)

@jacob314 @kenzieschmoll @grouma

### DIFF
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -137,6 +137,7 @@ class DevToolsAppState extends State<DevToolsApp> {
             url: params['uri'],
             allowConnectionScreenOnDisconnect: !embed,
             builder: (_) {
+              // TODO(dantup): Handle when `page` is not in the list of visible screens.
               final tabs = embed && page != null
                   ? _visibleScreens().where((p) => p.screenId == page).toList()
                   : _visibleScreens();
@@ -183,22 +184,7 @@ class DevToolsAppState extends State<DevToolsApp> {
     _routes = null;
   }
 
-  List<Screen> _visibleScreens() {
-    final visibleScreens = <Screen>[];
-    for (var screen in _screens) {
-      if (screen.conditionalLibrary != null) {
-        if (serviceManager.isServiceAvailable &&
-            serviceManager
-                .isolateManager.selectedIsolateAvailable.isCompleted &&
-            serviceManager.libraryUriAvailableNow(screen.conditionalLibrary)) {
-          visibleScreens.add(screen);
-        }
-      } else {
-        visibleScreens.add(screen);
-      }
-    }
-    return visibleScreens;
-  }
+  List<Screen> _visibleScreens() => _screens.where(shouldShowScreen).toList();
 
   Widget _providedControllers({@required Widget child, bool offline = false}) {
     final _providers = widget.screens
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -133,28 +133,30 @@ class DevToolsAppState extends State<DevToolsApp> {
         if (params['uri']?.isNotEmpty ?? false) {
           final embed = params['embed'] == 'true';
           final page = params['page'];
-          final tabs = embed && page != null
-              ? _visibleScreens().where((p) => p.screenId == page).toList()
-              : _visibleScreens();
           return Initializer(
             url: params['uri'],
             allowConnectionScreenOnDisconnect: !embed,
-            builder: (_) => _providedControllers(
-              child: DevToolsScaffold(
-                embed: embed,
-                ideTheme: ideTheme,
-                initialPage: page,
-                tabs: tabs,
-                actions: [
-                  if (serviceManager.connectedApp.isFlutterAppNow) ...[
-                    HotReloadButton(),
-                    HotRestartButton(),
+            builder: (_) {
+              final tabs = embed && page != null
+                  ? _visibleScreens().where((p) => p.screenId == page).toList()
+                  : _visibleScreens();
+              return _providedControllers(
+                child: DevToolsScaffold(
+                  embed: embed,
+                  ideTheme: ideTheme,
+                  initialPage: page,
+                  tabs: tabs,
+                  actions: [
+                    if (serviceManager.connectedApp.isFlutterAppNow) ...[
+                      HotReloadButton(),
+                      HotRestartButton(),
+                    ],
+                    OpenSettingsAction(),
+                    OpenAboutAction(),
                   ],
-                  OpenSettingsAction(),
-                  OpenAboutAction(),
-                ],
-              ),
-            ),
+                ),
+              );
+            },
           );
         } else {
           return DevToolsScaffold.withChild(
--- a/packages/devtools_app/lib/src/inspector/inspector_screen.dart
+++ b/packages/devtools_app/lib/src/inspector/inspector_screen.dart
@@ -7,7 +7,6 @@ import 'package:vm_service/vm_service.dart' hide Stack;
 
 import '../auto_dispose_mixin.dart';
 import '../blocking_action_mixin.dart';
-import '../common_widgets.dart';
 import '../connected_app.dart';
 import '../globals.dart';
 import '../initializer.dart';
@@ -27,7 +26,8 @@ class InspectorScreen extends Screen {
   const InspectorScreen()
       : super.conditional(
           id: 'inspector',
-          conditionalLibrary: flutterLibraryUri,
+          requiresLibrary: flutterLibraryUri,
+          requiresDebugBuild: true,
           title: 'Flutter Inspector',
           icon: Octicons.deviceMobile,
         );
@@ -36,16 +36,7 @@ class InspectorScreen extends Screen {
   String get docPageId => screenId;
 
   @override
-  Widget build(BuildContext context) {
-    final isFlutterApp = serviceManager.connectedApp.isFlutterAppNow;
-    final isProfileBuild = serviceManager.connectedApp.isProfileBuildNow;
-    if (!isFlutterApp || isProfileBuild) {
-      return !isFlutterApp
-          ? const DisabledForNonFlutterAppMessage()
-          : const DisabledForProfileBuildMessage();
-    }
-    return const InspectorScreenBody();
-  }
+  Widget build(BuildContext context) => const InspectorScreenBody();
 }
 
 class InspectorScreenBody extends StatefulWidget {
--- a/packages/devtools_app/lib/src/inspector/inspector_screen.dart
+++ b/packages/devtools_app/lib/src/inspector/inspector_screen.dart
@@ -8,6 +8,7 @@ import 'package:vm_service/vm_service.dart' hide Stack;
 import '../auto_dispose_mixin.dart';
 import '../blocking_action_mixin.dart';
 import '../common_widgets.dart';
+import '../connected_app.dart';
 import '../globals.dart';
 import '../initializer.dart';
 import '../octicons.dart';
@@ -24,8 +25,9 @@ import 'inspector_tree_flutter.dart';
 
 class InspectorScreen extends Screen {
   const InspectorScreen()
-      : super(
-          'inspector',
+      : super.conditional(
+          id: 'inspector',
+          conditionalLibrary: flutterLibraryUri,
           title: 'Flutter Inspector',
           icon: Octicons.deviceMobile,
         );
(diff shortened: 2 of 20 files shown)


jacob314:
Either the timeline or the debugger screen would be a reasonable default for a regular Dart app. 
If you run a Flutter Web app you should see all the performance screens (timeline, memory, performance) hidden.

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -24,8 +25,9 @@ import 'inspector_tree_flutter.dart';
 
 class InspectorScreen extends Screen {
   const InspectorScreen()
-      : super(
-          'inspector',
+      : super.conditional(
+          id: 'inspector',
+          conditionalLibrary: flutterLibraryUri,
```
jacob314, line 30: funny that we missed that the conditionalLibrary property would work great for the inspector!

packages/devtools_app/lib/src/app.dart
```diff
@@ -129,27 +129,29 @@ class DevToolsAppState extends State<DevToolsApp> {
...
-                embed: embed,
-                initialPage: page,
-                tabs: tabs,
-                actions: [
-                  if (serviceManager.connectedApp.isFlutterAppNow) ...[
-                    HotReloadButton(),
-                    HotRestartButton(),
+            builder: (_) {
+              final tabs = embed && page != null
+                  ? _visibleScreens().where((p) => p.screenId == page).toList()
```
jacob314, line 137: if you try to open an embedded page on an app that doesn't support it we should display an error message explaining what went wrong.

DanTup:
Does it seem reasonable for me to change `conditionalLibrary` to instead just be a getter that a screen can override?

```
@override
bool get isAvailable => serviceManager.connectedApp.isFlutterWebAppNow
```

I think that'll simplify things as the Flutter Web ones need to check two libraries (flutter + dart:html). They're both libraries though, so I don't think it'll introduce any extra delays or races.

kenzieschmoll:
My main concern is that this will put more pressure on the conditional screen (which would be outside of our package internally) to call into DevTools service code. Right now, the conditional screen just needs to specify a String (`conditionalLibrary`) and code inside of DevTools takes care of the rest. Transitioning to a getter would mean that the conditional screen would have to have this code, and therefore have more responsibility to understand DevTools internals:

```
bool get isAvailable => serviceManager.isServiceAvailable &&
            serviceManager
                .isolateManager.selectedIsolateAvailable.isCompleted &&
            serviceManager.libraryUriAvailableNow(screen.conditionalLibrary)
```

We may be able to put this into a helper that would make it simpler for the conditional library case, and also allow flexibility for other conditions that are not just the presence of a library. We just need to be cautious here with how this will build inside of g3.

DanTup:
@kenzieschmoll makes sense. If it might complicate things internally, it's probably better for me to stick with something simple (and if it needs extending in future, it may be better done by someone with visibility of the internal use).

I could just swap `String conditionalLibrary` to `List<String> conditionalLibraries` (`requiredLibraries`?) for now? I think the only current requirement is to check for both flutter+dart:html in order to hide those we don't want for Flutter web.

kenzieschmoll:
That sounds like a good short term solution that we can tweak in the future if we need.

jacob314:
> @kenzieschmoll makes sense. If it might complicate things internally, it's probably better for me to stick with something simple (and if it needs extending in future, it may be better done by someone with visibility of the internal use).
> 
> I could just swap `String conditionalLibrary` to `List<String> conditionalLibraries` (`requiredLibraries`?) for now? I think the only current requirement is to check for both flutter+dart:html in order to hide those we don't want for Flutter web.

That won't work because you need the inverse of `conditionalLibraries`. You need to say the performance screen should only show up if the dart:html library is not present.
I'm fine with having multiple parameters that filter whether a library can be concluded. For example requiresDartVm would be reasonable for the performance pages and requiresDebug is needed for the inspector. As of now we would show the inspector in profile mode which isn't right.

packages/devtools_app/lib/src/app.dart
DanTup: I've added a TODO for this and will do it it a separate PR to avoid adding more to this one (it's already grown a fair bit!).

DanTup:
I've pushed changes that move all the previous in-screen conditions (and showing messages) to use the conditional flags on the `Screen` class (and added new flags as suggested above). The new tests check exactly what screens appear for each combination of Dart/Flutter web/not-web app. I've also manually checked what I see when running the app matches what the test sees.

A few notes:

- I extracted `shouldShowScreen` from the app class so it could be tested without needing to set up the whole app (we don't have any tests that do that except for an integration test, and it got complicated). I'm not sure if there's a better place than a top-level function in `screen.dart` though.
- I don't know if I got all the conditions correct for each type of app, so please check the specific screens in the tests (I've left the hidden screens in the list but commented out, so it's easier to see what's there/missing).
- I had to make the library field on the `Screen.conditional` constructor not-required to support some screens that only cared about the other flags.. this means the named constructor looks a bit pointless now but I wasn't sure about removing it if it's used internally (though I may have already broken any existing use by renaming the field for consistency with the others 🤔)

packages/devtools_app/lib/src/logging/logging_screen.dart
```diff
@@ -30,12 +29,7 @@ class LoggingScreen extends Screen {
   String get docPageId => screenId;
 
   @override
-  Widget build(BuildContext context) {
-    return !(serviceManager.connectedApp.isFlutterWebAppNow &&
-            serviceManager.connectedApp.isProfileBuildNow)
```
kenzieschmoll, line 35: I'm struggling to remember why this was disabled for flutter web profile builds to begin with. @jacob314 @devoncarew do either of you recall?

packages/devtools_app/lib/src/network/network_screen.dart
```diff
@@ -33,11 +32,7 @@ class NetworkScreen extends Screen {
   static const recordingInstructionsKey = Key('Recording Instructions');
 
   @override
-  Widget build(BuildContext context) {
-    return !serviceManager.connectedApp.isDartWebAppNow
-        ? const NetworkScreenBody()
-        : const DisabledForWebAppMessage();
-  }
+  Widget build(BuildContext context) => const NetworkScreenBody();
```
kenzieschmoll, line 35: shouldn't we still be disabling this for dart web apps? is `requiresDartVm: true` now the equivalent of the `!serviceManager.connectedApp.isDartWebAppNow` condition?

packages/devtools_app/test/visible_screens_test.dart
```diff
@@ -0,0 +1,183 @@
...
+      setGlobal(ServiceConnectionManager, fakeServiceManager);
+      setGlobal(FrameworkController, FrameworkController());
+
+      await serviceManager.isolateManager.selectedIsolateAvailable.future;
+    });
+
+    void setupMockValues({
+      bool web = false,
+      bool flutter = false,
+      bool profileMode = false,
```
kenzieschmoll, line 38: nit: change to debug mode to match `mockIsDebugFlutterApp`

packages/devtools_app/test/visible_screens_test.dart
```diff
@@ -0,0 +1,183 @@
...
+    });
+
+    testWidgets('are correct for Flutter (non-web) profile app',
+        (WidgetTester tester) async {
+      setupMockValues(flutter: true, profileMode: true);
+
+      expect(
+          visibleScreenTypes,
+          equals([
+            InspectorScreen,
```
kenzieschmoll, line 112: Inspector and debugger should be disabled for profile builds

packages/devtools_app/test/visible_screens_test.dart
```diff
@@ -0,0 +1,183 @@
...
+    });
+
+    testWidgets('are correct for Flutter web profile app',
+        (WidgetTester tester) async {
+      setupMockValues(flutter: true, web: true, profileMode: true);
+
+      expect(
+          visibleScreenTypes,
+          equals([
+            InspectorScreen,
```
kenzieschmoll, line 148: here too - inspector and debugger should be disabled for profile builds

packages/devtools_app/test/visible_screens_test.dart
```diff
@@ -0,0 +1,183 @@
...
+          visibleScreenTypes,
+          equals([
+            // InspectorScreen,
+            TimelineScreen, // Works offline, so appears regardless of web flag
+            // MemoryScreen,
+            PerformanceScreen, // Works offline, so appears regardless of web flag
+            DebuggerScreen,
+            NetworkScreen,
+            LoggingScreen,
+            if (codeSizeScreenEnabled) CodeSizeScreen,
```
kenzieschmoll, line 173: I believe this list should only contain TimelineScreen and PerformanceScreen, as those are the only two screens that have implemented offline support

packages/devtools_app/lib/src/logging/logging_screen.dart
DanTup: Looks like this is something our current flags don't allow us to represent. For now I've committed the test with it enabled for Profile Flutter Web apps and a TODO.

packages/devtools_app/lib/src/network/network_screen.dart
DanTup: Yep, I somehow missed the condition on this. Fixed!

> is `requiresDartVm: true` now the equivalent of the `!serviceManager.connectedApp.isDartWebAppNow` condition?

Yep, they're the same. We could invert it, but the naming might be worse (`requiresNonWebApp`, `disabledForWebApp`?). I can update if there are any preferences!

packages/devtools_app/test/visible_screens_test.dart
DanTup: Done!

packages/devtools_app/test/visible_screens_test.dart
DanTup: Done!

packages/devtools_app/test/visible_screens_test.dart
DanTup: Done!

packages/devtools_app/test/visible_screens_test.dart
DanTup: How do I access the offline mode? I've fixed the code/tests, but unsure how to check the app still works. Going to `/snapshot` gives a message about not being able to load a screen with ID `null`, but I can't figure out how to provide it an ID.

kenzieschmoll: the easiest way would be to record a CPU profile in the Performance page and click the "Export" button. then drag and drop that file into DevTools (this only works for the web version)

DanTup: Ah, got it! Though it seems like when doing this, the navigation/tabs aren't shown (no `tabs` are passed to `DevToolsScaffold`, so it doesn't look like this change affects anything there right now.

packages/devtools_app/lib/src/logging/logging_screen.dart
jacob314: it is reasonable to expose in profile builds. Just check it doesn't throw exceptions. The structured error handling won't work in profile builds.

DanTup: How can I get the VM service address for a web profile build to test? It doesn't seem to be provided from the terminal (using `flutter run -d chrome --profile`) nor in the JSON in `--machine` mode as far as I can see.

DanTup: Turns out you can't connect DevTools to web profile mode as there's no VM Service. So this existing check didn't do anything. I've removed the corresponding test.

DanTup:
I've removed the web+profile test and rebased on latest code.

@jacob314 it may be good for someone to check this doesn't break (/have races) for any internal app types I haven't tested with (I did test Dart web, Flutter web, Flutter mobile), either before or after landing.

packages/devtools_app/lib/src/app.dart
DanTup: Opened https://github.com/flutter/devtools/pull/2207

PR #2204, opened 2020-07-27
Add DragAndDropManager and hook up drag and drop in the code size page

kenzieschmoll:
Until this CL, the only consumer in DevTools that used the `DragAndDrop` widget was `DevToolsScaffold`. It is used to listen for offline snapshot imports via drag and drop. We now have a need for more `DragAndDrop` consumers on the Code Size page, so this CL enables that by 

- adding a `DragAndDropManager` that fires listeners on the "top-most" `DragAndDrop` widget in the tree. 
- stopping drag and drop event propagation when `DragAndDrop` with a null drop handler is present
- adding `DragAndDrop` widgets to the Code size page

We use a hitTest to find the "top-most" `DragAndDrop` widget on a page, so this inherently allows for multiple "drop zones" on a single screen.


kenzieschmoll:
FYI @peterdjlee

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
```diff
@@ -66,4 +109,13 @@ abstract class DragAndDropState extends State<DragAndDrop> {
   void drop() {
     _dragging.value = false;
   }
+
+  bool coordinatesInBound(double x, double y) {
+    final RenderBox renderBox = context.findRenderObject();
+    final globalToLocalOffset = renderBox.globalToLocal(Offset(x, y));
+    return globalToLocalOffset.dx >= 0.0 &&
```
jacob314, line 116: using the constraints maxWidth doesn't seem right. 
https://api.flutter.dev/flutter/rendering/RenderBox/paintBounds.html
or semantic bounds would probably be better. The rectangle class should also have a helper to check intersection.

packages/devtools_app/.vscode/launch.json
```diff
@@ -0,0 +1,13 @@
+{
```
jacob314, line 1: Revert this change

packages/devtools_app/lib/src/app.dart
```diff
@@ -160,11 +163,13 @@ class DevToolsAppState extends State<DevToolsApp> {
           return DevToolsScaffold.withChild(
             child: ConnectScreenBody(),
             ideTheme: ideTheme,
+            dragAndDropId: 'Scaffold - connect screen - dragAndDropKey',
           );
         }
       },
       snapshotRoute: (_, __, args) {
         return DevToolsScaffold.withChild(
+          dragAndDropId: 'Scaffold - snapshot screen - dragAndDropKey',
```
jacob314, line 172: I don't think these ids are needed.

packages/devtools_app/lib/src/config_specific/drag_and_drop/_drag_and_drop_web.dart
```diff
@@ -49,7 +45,8 @@ class _DragAndDropWebState extends DragAndDropState {
   }
 
   void _onDragOver(MouseEvent event) {
-    super.dragOver();
+    super.dragOver(event.offset.x, event.offset.y);
```
jacob314, line 48: not a new issue but it is surprising to call super with a method other than the name if the current method. Can that be avoided?

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
```diff
@@ -2,61 +2,169 @@
...
+    final manager = createDragAndDropManager(notifications: notifications);
+    manager.init();
+    return manager;
+  }
+
+  DragAndDropManager.impl({@required this.notifications});
+
+  final NotificationsState notifications;
+
+  final _statesById = <String, DragAndDropState>{};
```
jacob314, line 27: I don't think we need this map. Just walking the Element tree and using DragAndDropState is enough. Having the map just leaves rooms for bugs like two widgets with the same id causing a collision in the map and one object to get lost without unregister being called.

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
```diff
@@ -2,61 +2,169 @@
...
+  void hitTestAndUpdateActiveId(double x, double y) {
+    final hitTestResult = HitTestResult();
+    RendererBinding.instance.hitTest(hitTestResult, Offset(x, y));
+
+    // Starting at bottom of [hitTestResult.path], look for [DragAndDrop] or
+    // [DragAndDropEventAbsorber] widgets. These widgets will be marked by a
+    // [RenderMetaData] target with [DragAndDropMetaData] metaData.
+    for (final result in hitTestResult.path) {
+      if (result.target is RenderMetaData) {
+        final target = result.target as RenderMetaData;
```
jacob314, line 77: you can get the DragAndDropState directly from the target.metaData instead of needing the map.

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
```diff
@@ -2,61 +2,169 @@
...
+  /// widgets accordingly.
+  void hitTestAndUpdateActiveId(double x, double y) {
+    final hitTestResult = HitTestResult();
+    RendererBinding.instance.hitTest(hitTestResult, Offset(x, y));
+
+    // Starting at bottom of [hitTestResult.path], look for [DragAndDrop] or
+    // [DragAndDropEventAbsorber] widgets. These widgets will be marked by a
+    // [RenderMetaData] target with [DragAndDropMetaData] metaData.
+    for (final result in hitTestResult.path) {
+      if (result.target is RenderMetaData) {
```
jacob314, line 76: you can avoid this as check in dart by writing
```dart
final target = result.target;
if (target is RenderMetaData) {
  // now target is known to be a RenderMetaData so not as check required.
  // Dart is picky about this because hitTestResult.target could possibly return a different value the second time it is called.
```
```

File: packages/devtools_app/.vscode/launch.json (hunk 2 above)
Comment (kenzieschmoll) [reply packages/devtools_app/.vscode/launch.json:1]: done

File: packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart [hunk 8]
```diff
@@ -2,61 +2,169 @@
...
+        // The first [DragAndDropMetaData] we find will either be for the active
+        // [DragAndDrop] widget or for a [DragAndDropEventAbsorber] widget. We
+        // should return if we find either.
+        if (target.metaData is DragAndDropMetaData) {
+          final metaData = target.metaData as DragAndDropMetaData;
+
+          // Deactivate all drag and drop states if we find
+          // [_dragAndDropEventAbsorberId].
+          if (metaData.value == _dragAndDropEventAbsorberId) {
+            for (final state in _statesById.values) {
```
jacob314, line 87: you can track the last active state instead of looping through all of them.

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
```diff
@@ -2,61 +2,169 @@
...
+
+class DragAndDrop extends StatefulWidget {
+  const DragAndDrop({
+    @required this.id,
+    @required this.manager,
+    @required this.handleDrop,
+    @required this.child,
+  });
+
+  final String id;
```
jacob314, line 117: can we remove the id?

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
```diff
@@ -2,61 +2,169 @@
...
 
   /// Callback to handle parsed data from drag and drop.
   ///
   /// The current implementation expects data in json format.
-  final void Function(Map<String, dynamic> data) handleDrop;
+  final DevToolsJsonFileHandler handleDrop;
+
+  final DragAndDropManager manager;
```
jacob314, line 124: is there a reason we don't just assume this is a singleton?
E.g. use DragAndDropManager.instance everywhere it is needed.

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
```diff
@@ -2,61 +2,169 @@
...
+    widget.manager.registerDragAndDrop(widget.id, this);
+  }
 
   @override
-  void didChangeDependencies() {
-    notifications = Notifications.of(context);
-    super.didChangeDependencies();
+  void dispose() {
+    super.dispose();
+    widget.manager.unregisterDragAndDrop(widget.id);
```
jacob314, line 146: by convention, call super.disppose() last.
@pq this would be a nice flutter specific lint.

jacob314, line 146: use `this` instead of `widget.id`

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
```diff
@@ -66,4 +174,64 @@ abstract class DragAndDropState extends State<DragAndDrop> {
   void drop() {
     _dragging.value = false;
   }
+
+  void activate(bool active) {
+    _isActive = active;
+  }
+}
+
+const _dragAndDropEventAbsorberId = 'DragAndDropEventAbsorber';
```
jacob314, line 183: Can't this just be the DragAndDrop widget with no handler on drag. Behavior could be that if you have no handler on drag that means nothing should be visually rendered on drag.

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
kenzieschmoll: removed id

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
kenzieschmoll: done

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
kenzieschmoll: done

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
kenzieschmoll: done

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
kenzieschmoll: done

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
kenzieschmoll: done

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
kenzieschmoll: done

kenzieschmoll: done

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
kenzieschmoll: done

packages/devtools_app/lib/src/config_specific/drag_and_drop/_drag_and_drop_web.dart
kenzieschmoll: `super.dragOver()` doesn't take a MouseEvent because that is an html specific import. I removed `super` to make it more clear that this is not an override method

PR #2208, opened 2020-07-30
Use Object instead of Map<String, dynamic> for DevToolsJsonFile

kenzieschmoll:
We need to use Object instead of Map<String, dynamic> because instruction_sizes input is actually a List of Maps, and v8 input is a map. package:vm_snapshot_analysis uses Object as an expected type in utility methods for this reason.

### DIFF
--- a/packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
+++ b/packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
@@ -59,15 +59,17 @@ class ImportController {
     }
     previousImportTime = now;
 
-    final isDevToolsSnapshot = json[devToolsSnapshotKey];
+    final isDevToolsSnapshot =
+        json is Map<String, dynamic> && json[devToolsSnapshotKey];
     if (isDevToolsSnapshot == null || !isDevToolsSnapshot) {
       _notifications.push(nonDevToolsFileMessage);
       return;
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
```diff
@@ -59,15 +59,17 @@ class ImportController {
...
       _notifications.push(nonDevToolsFileMessage);
       return;
     }
 
+    final devToolsSnapshot = json as Map<String, dynamic>;
     // TODO(kenz): support imports for more than one screen at a time.
-    final activeScreenId = json[activeScreenIdKey];
-    offlineDataJson = json;
+    final activeScreenId = devToolsSnapshot[activeScreenIdKey];
+    offlineDataJson = devToolsSnapshot;
```
peterdjlee, line 72: Are the additions to this file just to make sure we don't accept instruction sizes/v8 snapshot for offline mode?

kenzieschmoll: This file uses the `DevToolsJsonFile` object as well. So we need to cast to a map so that we can check for the active screen id. If it is not a map, we know it's not an offline import and we will show the `nonDevToolsFileMessage` message.

PR #2214, opened 2020-07-31
Add code_size_package example to case_study/code_size

peterdjlee:
case_study/code_size/unoptimized/code_size_package is a simple application that uses package:encrypt which imports around 500KB data.

case_study/code_size/optimized/code_size_package is the same application that uses a smaller, specific package:steel_crypt to do the exact same task for around 500KB reduced code size.

We will use this application to see how the effective the DevTools code size tool is at identifying large packages in an APK analysis.


case_study/code_size/optimized/code_size_package/lib/main.dart
```diff
@@ -0,0 +1,104 @@
+import 'dart:convert';
+import 'dart:typed_data';
+
+import 'package:flutter/material.dart';
+import 'package:steel_crypt/steel_crypt.dart';
```
kenzieschmoll, line 5: can we import this `as encrypt` so it is clear where we are using the package?

case_study/code_size/optimized/code_size_package/lib/main.dart
```diff
@@ -0,0 +1,104 @@
...
+
+  @override
+  _MyHomePageState createState() => _MyHomePageState();
+}
+
+class _MyHomePageState extends State<MyHomePage> {
+  final textEdittingController = TextEditingController(
+    text: 'Lorem ipsum dolor sit amet',
+  );
+  String encryptedText = '';
```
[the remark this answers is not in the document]
peterdjlee: Done

case_study/code_size/optimized/code_size_package/lib/main.dart
peterdjlee: Done

PR #2217, opened 2020-07-31
Add instructions to file import containers in code size page

kenzieschmoll:
Please feel free to suggest better wording. Would be nice to link out to instructions on how to generate these files when/if they exist.

Snapshot / APK view
Screen Shot 2020-07-31 at 3 28 47 PM
Snapshot / APK view - file selected 
Screen Shot 2020-07-31 at 3 29 08 PM
Diff view
Screen Shot 2020-07-31 at 3 27 14 PM
Diff view - files selected
Screen Shot 2020-07-31 at 3 27 38 PM

TODO before landing - polish and adding tests

### DIFF
--- a/packages/devtools_app/lib/src/code_size/file_import_container.dart
+++ b/packages/devtools_app/lib/src/code_size/file_import_container.dart
@@ -13,6 +13,7 @@ import '../utils.dart';
 class FileImportContainer extends StatefulWidget {
   const FileImportContainer({
     @required this.title,
+    @required this.instructions,
     this.actionText,
     this.onAction,
     this.onFileSelected,
@@ -20,6 +21,8 @@ class FileImportContainer extends StatefulWidget {
   }) : super(key: key);
 
   final String title;
+
+  final String instructions;
 
   /// The title of the action button.
   final String actionText;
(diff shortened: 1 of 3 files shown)


peterdjlee:
nit: Is there a way to distinguish/separate the instruction line and "drag and drop or click import file" line? I feel like those two are different types of instructions. Maybe something like putting the drag and drop line under the import bar and making the color more like gray? Not really sure if that'd work. Would be great to get some UXR input for the import pages in general.

kenzieschmoll:
I'm not sure we should break it up visually - I think it makes more sense to have the visual break down of `text -> input action -> analyze action`, vs `text -> input action -> analyze action -> more text`. Instructions on how to input the file are logically paired which which file we should input, IMO. We could try to make the instructions more concise and merge into a single sentence, but then again, the instructions may become more verbose anyway once we have some documentation to link to.

packages/devtools_app/lib/src/code_size/file_import_container.dart
```diff
@@ -72,23 +71,65 @@ class _FileImportContainerState extends State<FileImportContainer> {
...
-            color: Theme.of(context).textTheme.headline1.color,
+        // Horizontal spacer with flex value of 1.
+        const Flexible(
+          child: SizedBox(height: rowHeight),
+        ),
+        Flexible(
+          flex: 4,
+          fit: FlexFit.tight,
+          child: Container(
+            height: 37.0,
```
peterdjlee, line 101: replace 37.0 with `rowHeight`

kenzieschmoll: good catch. done

PR #2218, opened 2020-08-03
Theming tweaks

DanTup:
This tweaks the threshold used for deciding whether the querystring-supplied background/foreground colors are dark/light enough to use Flutter's `calculateLuminance` (and currently has a threshold of 0.1 out of 1 at both ends).

It also tweaks a few places that were using a default grey colour to instead use the background lightened/darkened. There are still a few places that grey shows up, but some may be more difficult to fix (they come from our custom `DevToolsColorScheme` overrides that might not have access to the real theme/context).

Screenshot 2020-08-03 at 18 05 25

Screenshot 2020-08-03 at 18 06 00

Screenshot 2020-08-03 at 18 06 17

@jacob314

### DIFF
--- a/packages/devtools_app/lib/src/inspector/layout_explorer/flex/flex.dart
+++ b/packages/devtools_app/lib/src/inspector/layout_explorer/flex/flex.dart
@@ -9,6 +9,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter/rendering.dart';
 import 'package:flutter/widgets.dart';
 
+import '../../../common_widgets.dart';
 import '../../../theme.dart';
 import '../../../ui/theme.dart';
 import '../../../utils.dart';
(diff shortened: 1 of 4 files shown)


DanTup:
Thoughts on this? I'm not too sure about how the light theme looks 🤔

Screenshot 2020-08-03 at 18 33 49

Screenshot 2020-08-03 at 18 34 20

Screenshot 2020-08-03 at 18 34 32

jacob314:
The color change is a bit too extreme on the light theme.  I would suggest customizing the darken % in the light and dark theme to match the existing colors as closely as practical.

DanTup:
@jacob314 how about these? Left is new, right is old. The previous light theme got lighter, but here we always darken so it looks different, but is more subtle now. Alternatively, we could lighten when the theme is light instead?

Screenshot 2020-08-04 at 10 30 45

Screenshot 2020-08-04 at 10 31 05

Screenshot 2020-08-04 at 10 32 10

jacob314:
lgtm

DanTup:
I've rebased since it's been a while, will land as long as nothing breaks. Thanks!

DanTup:
(This took quite a few runs to go green, seems like some of the tests are currently quite flaky - unrelated to this change though!)

PR #2221, opened 2020-08-04
Add support for DDS connections when using Devtools on web

bkonyi:
As of 3.37 the VM service will attempt to redirect web socket connections to DDS if DDS is connected. This works fine with some web socket implementations (e.g., dart:io's WebSocket) but is unsupported in others (e.g., dart:html's WebSocket). On the web we need to query the VM service via HTTP first to determine which web socket URI we should use (see [getWebSocketTarget](https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#getwebsockettarget) RPC for details).

### DIFF
--- a/packages/devtools_app/lib/src/service.dart
+++ b/packages/devtools_app/lib/src/service.dart
@@ -3,9 +3,13 @@
 // found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:convert';
 
+import 'package:flutter/foundation.dart';
+import 'package:http/http.dart' as http;
 import 'package:pedantic/pedantic.dart';
 import 'package:vm_service/utils.dart';
+import 'package:vm_service/vm_service.dart';
 import 'package:web_socket_channel/web_socket_channel.dart';
 
 import 'config_specific/sse/sse_shim.dart';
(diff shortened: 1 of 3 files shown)


devoncarew:
I may try and set up a VC - I think it would be useful to see if there's a way we can do this without a redirect (either an http one or one based on client probing).

jacob314:
Maybe I'm misunderstanding this.
Would it be possible for us to only surface the DDS URI to debugging tools instead of the raw vm service one? That way they could just operate like they do today and flutter_tools and `dart` could be responsible for making sure we surface the `dds` uri instead of the raw one from the vm service?

bkonyi:
> Maybe I'm misunderstanding this.
> Would it be possible for us to only surface the DDS URI to debugging tools instead of the raw vm service one? That way they could just operate like they do today and flutter_tools and `dart` could be responsible for making sure we surface the `dds` uri instead of the raw one from the vm service?

For various reasons this isn't ideal, but the biggest issue is the `flutter attach` case. Flutter always has to output the VM service URI as DDS is hosted within `flutter_tools`. If the original `flutter_tools` instance shuts down for whatever reason, users won't be able to reattach as they won't know the actual VM service URI. Also, we still report the VM service connection information via `dart:developer` which would be another headache we'd have to deal with.

The original approach I took was to replace the VM service URI with the DDS URI and it ended up being a nightmare, particularly for `flutter_tools` which required some pretty invasive changes.

bkonyi:
Abandoning this since we're no longer using redirects for clients to connect to DDS.

PR #2229, opened 2020-08-05
Change gitignore so that we check it pubspec.loc files.

jacob314:
This fixes cases where local runs were inconsistent from runs on the build bots unless you ran `flutter pub upgrade` locally.


pubspec.lock
```diff
@@ -0,0 +1,551 @@
+# Generated by pub
```
devoncarew, line 1: I don't think we should have a pubspec file at the top level of the repo.

devoncarew:
For the travis failures, I think we'll need to change `flutter format --set-exit-if-changed lib/ test/ web/` to `dart format --output=none --set-exit-if-changed lib/ test/ web/`.

pubspec.lock
jacob314: that shows how long I've had the same local client.

jacob314:
> dart format --output=none --set-exit-if-changed lib/ test/ web/

trying that.

jacob314:
> One comment about a file that should probably be deleted.
> 
> Also, do we care about having a consistent solve between devtools_app / devtools_server / devtools_shared? If so, we might want to do something like what flutter/flutter does - solve for related packages at once, and them move those versions into the related pubspec / lock files. It's worth filing an issue for it if we think this is something we should do.

filed https://github.com/flutter/devtools/issues/2240 to track.

devoncarew:
> > dart format --output=none --set-exit-if-changed lib/ test/ web/
> 
> trying that.

I have https://github.com/flutter/devtools/pull/2231, but it's not happy for reasons I can't figure out.

Issue #2173, opened 2020-07-10
Improve performance for code size screen


peterdjlee:
* Improve performance when switching between tabs.
  - Keep each tab view persistent between tab switches.
* Improve performance when switching between different diff types on the diff tab.
  - Ensure loading for each type only happens once by saving each tree in a node.

peterdjlee:
> Is there a way to keep tabs truly persistent?
I've tried looking at [this article](https://medium.com/@diegoveloper/flutter-persistent-tab-bars-a26220d322bc) which states that using `AutomaticKeepAliveClientMixin` should help. I was able to stop the tabs from rebuilding, but that didn't seem to improve the performance.

peterdjlee:
> I've tried looking at this article which states that using AutomaticKeepAliveClientMixin should help. I was able to stop the tabs from rebuilding, but that didn't seem to improve the performance. Either rebuilding isn't the problem or my method of checking if the widget is rebuilding is wrong.

Rebuilding widgets:
![ezgif com-video-to-gif (3)](https://user-images.githubusercontent.com/22206026/87204495-fc4c4900-c2d2-11ea-8898-8769bab92385.gif)

Not rebuilding widgets:
![ezgif com-video-to-gif (4)](https://user-images.githubusercontent.com/22206026/87204547-1dad3500-c2d3-11ea-9019-d07f0c323f94.gif)

kenzieschmoll:
Is each tree map cell wrapped in a Tooltip?

peterdjlee:
> Is each tree map cell wrapped in a Tooltip?

Yup. All cells and also the title bars are wrapped in a Tooltip.

peterdjlee:
> Yup. All cells and also the title bars are wrapped in a Tooltip.

Can confirm that removing all Tooltip widgets and Inkwell widget shaves off about a second each, reducing around 66% of the loading time. So one optimization we can do would be to limit the number of widgets we wrap with Tooltip and Inkwell.

kenzieschmoll:
[...]

PR #2233, opened 2020-08-05
Build diff trees on analyze and cache the roots

peterdjlee:
Build all three (combined, increase only, decrease only) diff tree types when user presses the analyze button and cache the roots to prevent any re-processing of the data.

### DIFF
--- a/packages/devtools_app/lib/src/code_size/code_size_controller.dart
+++ b/packages/devtools_app/lib/src/code_size/code_size_controller.dart
@@ -53,11 +53,14 @@ class CodeSizeController {
   TreemapNode get _activeDiffRoot {
     switch (_activeDiffTreeType.value) {
       case DiffTreeType.increaseOnly:
+        assert(_increasedDiffTreeRoot != null);
         return _increasedDiffTreeRoot;
       case DiffTreeType.decreaseOnly:
+        assert(_decreasedDiffTreeRoot != null);
         return _decreasedDiffTreeRoot;
       case DiffTreeType.combined:
       default:
+        assert(_combinedDiffTreeRoot != null);
         return _combinedDiffTreeRoot;
     }
   }
@@ -145,7 +148,7 @@ class CodeSizeController {
 
     final diffMap = buildComparisonTreemap(oldFile.data, newFile.data);
     diffMap['n'] = 'Root';
-    
+
     // TODO(peterdjlee): Try to move the non-active tree generation to separate isolates.
     _combinedDiffTreeRoot = generateDiffTree(
       diffMap,
--- a/packages/devtools_app/lib/src/code_size/code_size_controller.dart
+++ b/packages/devtools_app/lib/src/code_size/code_size_controller.dart
@@ -82,6 +82,7 @@ class CodeSizeController {
     }
   }
 
+
   /// The active diff tree type used to build the diff treemap.
   ValueListenable<DiffTreeType> get activeDiffTreeType {
     return _activeDiffTreeType;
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/code_size/code_size_controller.dart
```diff
@@ -90,14 +90,23 @@ class CodeSizeController {
   final _activeDiffTreeType =
       ValueNotifier<DiffTreeType>(DiffTreeType.combined);
 
-  // TODO(peterdjlee): Cache each diff tree so that it's less expensive
-  //                   to change bettween diff tree types.
+  TreemapNode _increasedDiffTreeRoot;
+  TreemapNode _decreasedDiffTreeRoot;
+  TreemapNode _combinedDiffTreeRoot;
```
kenzieschmoll, line 95: make these getters that either return the cached value or generate the tree if the cached value is null

kenzieschmoll: all the cached values also need to be reset to null on clear

packages/devtools_app/lib/src/code_size/code_size_controller.dart
```diff
@@ -90,14 +90,23 @@ class CodeSizeController {
...
+
   void changeActiveDiffTreeType(DiffTreeType newDiffTreeType) {
     _activeDiffTreeType.value = newDiffTreeType;
-    loadDiffTreeFromJsonFiles(
-      _oldDiffSnapshotJsonFile.value,
-      _newDiffSnapshotJsonFile.value,
-    );
+    switch (_activeDiffTreeType.value) {
+      case DiffTreeType.increaseOnly:
+        changeDiffRoot(_increasedDiffTreeRoot);
```
kenzieschmoll, line 101: these should reference the respective getters with the null checks

packages/devtools_app/lib/src/code_size/code_size_controller.dart
```diff
@@ -131,9 +140,14 @@ class CodeSizeController {
...
-    final newRoot = generateDiffTree(diffMap);
 
-    changeDiffRoot(newRoot);
+    _combinedDiffTreeRoot = generateDiffTree(diffMap, DiffTreeType.combined);
+    _increasedDiffTreeRoot =
+        generateDiffTree(diffMap, DiffTreeType.increaseOnly);
+    _decreasedDiffTreeRoot =
+        generateDiffTree(diffMap, DiffTreeType.decreaseOnly);
+
+    changeDiffRoot(_combinedDiffTreeRoot);
```
kenzieschmoll, line 150: change to the active diff tree root. You may need to make `activeDiffTreeRoot` a getter that switches on the active tree type. As a user if I want to look at increase only for one diff, and then I clear and load a second diff, I don't want the default so switch back to combined, I was to see the diff type that I left off at.

packages/devtools_app/lib/src/code_size/code_size_controller.dart
```diff
@@ -131,9 +140,14 @@ class CodeSizeController {
 
     final diffMap = buildComparisonTreemap(oldFile.data, newFile.data);
     diffMap['n'] = 'Root';
-    final newRoot = generateDiffTree(diffMap);
```
kenzieschmoll, line 143: add a todo to try to move the non-active root generation to a separate isolate

packages/devtools_app/lib/src/code_size/code_size_controller.dart
peterdjlee: Done

packages/devtools_app/lib/src/code_size/code_size_controller.dart
```diff
@@ -50,6 +50,22 @@ class CodeSizeController {
     _diffRoot.value = newRoot;
   }
 
+  TreemapNode get _activeDiffRoot {
+    switch (_activeDiffTreeType.value) {
+      case DiffTreeType.increaseOnly:
```
kenzieschmoll, line 55: add asserts in each case that the root is not null

kenzieschmoll:
part of https://github.com/flutter/devtools/issues/2173

PR #2241, opened 2020-08-06
Upgrade script

jacob314:
Quick and dirty pub upgrade script so we always run upgrade on all the packages. Also runs it as part of the script to update the version across all packages.


jacob314:
@googlebot

tool/pub_upgrade.sh
```diff
@@ -0,0 +1,29 @@
...
+pushd devtools_server
+flutter pub upgrade
+popd
+
+pushd devtools_shared
+flutter pub upgrade
+popd
+
+
+popd
```
kenzieschmoll, line 29: new line at EOF?

jacob314: done

PR #2247, opened 2020-08-06
convert the scripts view to a tree

devoncarew:
- convert the scripts view from a flat list to a tree view

Screen Shot 2020-08-06 at 3 49 21 PM

### DIFF
--- a/packages/devtools_app/lib/src/debugger/debugger_model.dart
+++ b/packages/devtools_app/lib/src/debugger/debugger_model.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:flutter/material.dart';
 import 'package:meta/meta.dart';
 import 'package:vm_service/vm_service.dart';
 
@@ -25,12 +26,7 @@ class ScriptLocation {
   }
 
   @override
-  int get hashCode {
-    int hash = 17;
-    hash = 0x1fffffff & (hash * 31 + scriptRef.hashCode);
-    hash = 0x1fffffff & (hash * 31 + location.hashCode);
-    return hash;
-  }
+  int get hashCode => hashValues(scriptRef, location);
 
   @override
   String toString() => '${scriptRef.uri} $location';
@@ -271,7 +267,7 @@ class FileNode extends TreeNode<FileNode> {
   ScriptRef scriptRef;
 
   /// This exists to allow for O(1) lookup of children when building the tree.
-  Map<String, FileNode> _childrenAsMap = {};
+  final Map<String, FileNode> _childrenAsMap = {};
 
   bool get hasScript => scriptRef != null;
 
@@ -282,7 +278,15 @@ class FileNode extends TreeNode<FileNode> {
     final root = FileNode('<root>');
 
     for (var script in scripts) {
-      final parts = script.uri.split('/');
+      var parts = script.uri.split('/');
+
+      // Look for and handle dotted package names.
+      if (parts.isNotEmpty && parts.first.contains('.')) {
+        parts = [
+          ...parts.first.split('.'),
+          ...parts.sublist(1),
+        ];
+      }
 
       FileNode node = root;
       for (var name in parts) {
@@ -291,7 +295,7 @@ class FileNode extends TreeNode<FileNode> {
       node.scriptRef = script;
     }
 
-    // Null out the _childrenAsSet field.
+    // Clear out the _childrenAsMap map.
     root._trimChildrenAsMapEntries();
 
     return root.children;
@@ -306,9 +310,9 @@ class FileNode extends TreeNode<FileNode> {
     });
   }
 
-  /// Null out the _childrenAsMap field recursively to save memory.
+  /// Clear the _childrenAsMap map recursively to save memory.
   void _trimChildrenAsMapEntries() {
-    _childrenAsMap = null;
+    _childrenAsMap.clear();
 
     for (var child in children) {
       child._trimChildrenAsMapEntries();
--- a/packages/devtools_app/lib/src/debugger/debugger_model.dart
+++ b/packages/devtools_app/lib/src/debugger/debugger_model.dart
@@ -2,8 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'package:flutter/foundation.dart';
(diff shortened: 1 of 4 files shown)


devoncarew:
```
devtools_app/test/debugger_screen_test.dart: DebuggerScreen Libraries visible
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following TestFailure object was thrown running a test:
  Expected: exactly one matching node in the widget tree
  Actual: _TextFinder:<zero widgets with text "package:/test/script.dart" (ignoring offstage
widgets)>
   Which: means none were found but one was expected
```

devoncarew:
@googlebot ...

packages/devtools_app/lib/src/debugger/debugger_model.dart
```diff
@@ -26,7 +25,12 @@ class ScriptLocation {
   }
 
   @override
-  int get hashCode => hashValues(scriptRef, location);
+  int get hashCode {
+    int hash = 17;
+    hash = 0x1fffffff & (hash * 31 + scriptRef.hashCode);
```
jacob314, line 30: You might as well use the flutter `hashValues` helper method for consistency.

packages/devtools_app/lib/src/debugger/debugger_model.dart
```diff
@@ -252,3 +256,62 @@ class Variable extends TreeNode<Variable> {
...
+class FileNode extends TreeNode<FileNode> {
+  FileNode(this.name);
+
+  final String name;
+
+  // This can be null.
+  ScriptRef scriptRef;
+
+  /// This exists to allow for O(1) lookup of children when building the tree.
+  Map<String, FileNode> _childrenAsMap = {};
```
jacob314, line 274: in the spirit of NNBD can you make this final and clear the entries rather than setting it to null?

packages/devtools_app/lib/src/debugger/debugger_model.dart
devoncarew: I was trying to avoid a package:flutter import so that the tests for this could run on the dart cli. Perhaps I should add the import and assume that the test will run as a `flutter test`?

packages/devtools_app/lib/src/debugger/debugger_model.dart
devoncarew: Hmm, we'd save some of the memory then but not all (the n instances of empty maps). It could very well be that this is an unnecessary optimization (the `_trimChildrenAsMapEntries()` method generally).

I'll switch to clearing the map -

packages/devtools_app/lib/src/debugger/debugger_model.dart
```diff
@@ -252,3 +256,62 @@ class Variable extends TreeNode<Variable> {
...
+  bool get hasScript => scriptRef != null;
+
+  /// Given a flat list of service protocol scripts, return a tree of scripts
+  /// representing the best hierarchical grouping.
+  static List<FileNode> createRootsFrom(List<ScriptRef> scripts) {
+    // The name of this node is not exposed to users.
+    final root = FileNode('<root>');
+
+    for (var script in scripts) {
+      final parts = script.uri.split('/');
```
kenzieschmoll, line 285: what about google3 uris? e.g. `package:foo.bar.baz/src/foobar.dart`

Here's a method in vm_snapshot_analysis that handles this, though it doesn't seem like depending on vm_snapshot_analysis from the debugger makes much sense, so we could just copy it over to devtools:

```
List<String> partsForPath(String path) {
  final parts = path.split('/');
  if (parts.first.startsWith('package:')) {
    // Convert dot separated package name into a path from which this package originated.
    parts.replaceRange(0, 1, parts.first.split('.'));
  }
  return parts;
}
```

vm_snapshot_analysis commit where I added this: https://github.com/dart-lang/sdk/commit/99a233505f5ae4102f6537425e625fe8960c5cef

packages/devtools_app/test/debugger_model_test.dart
```diff
@@ -0,0 +1,57 @@
...
+import 'package:test/test.dart';
+import 'package:vm_service/vm_service.dart';
+
+void main() {
+  group('debugger_model', () {
+    group('FileNode', () {
+      final ScriptRef ref1 = ScriptRef(uri: 'package:foo', id: 'id-1');
+      final ScriptRef ref2 = ScriptRef(uri: 'package:foo/foo.dart', id: 'id-2');
+      final ScriptRef ref3 = ScriptRef(uri: 'package:bar/bar.dart', id: 'id-3');
+      final ScriptRef ref4 = ScriptRef(uri: 'package:bar/baz.dart', id: 'id-4');
```
kenzieschmoll, line 15: we should add a test for google3 style uris

packages/devtools_app/lib/src/debugger/debugger_model.dart
devoncarew: Good point - this should handle google3 urls as well.

Currently it would display them in a flat list at the top level:

```
package:foo.bar.baz/
package:foo.bar.qux/
...
```

I assume it would be better to display them heirarichally, but might need to actually play with it to be sure:

```
package:foo/
  bar/
    baz/
    qux/
...
```

devoncarew: cc @grouma who also brought this up

grouma: +1 for hierarchal.

packages/devtools_app/lib/src/debugger/debugger_model.dart
devoncarew: converted back to using `hashValues()`.

packages/devtools_app/lib/src/debugger/debugger_model.dart
devoncarew: Converted to `_childrenAsMap.clear()`.

packages/devtools_app/lib/src/debugger/debugger_model.dart
devoncarew: done - we handle dotted paths now

packages/devtools_app/test/debugger_model_test.dart
devoncarew: 👍 - added

PR #2251, opened 2020-08-07
Update scaffold's narrowWidthThreshhold now that the code size tab is visible

kenzieschmoll:


kenzieschmoll:
no, it's just a value that looks fine manually testing. We have a todo to dynamically calculate the value of `narrowWidthThreshold`, but that is likely a much more involved fix. This is just a band aid to stop the tabs from overlapping with "Dart DevTools".

PR #2252, opened 2020-08-07
Use CustomPainter to render Treemap cells instead of building widgets

peterdjlee:
Use CustomPainter to render Treemap cells instead of building widgets. This makes navigating between levels much faster than before. On a medium-sized app, it reduces the on-click loading time from 10 seconds to almost instant (profile mode).

### DIFF
--- a/packages/devtools_app/lib/src/charts/treemap.dart
+++ b/packages/devtools_app/lib/src/charts/treemap.dart
@@ -18,6 +18,7 @@ class Treemap extends StatelessWidget {
     this.nodes,
     @required this.levelsVisible,
     @required this.isOutermostLevel,
+    @required this.width,
     @required this.height,
     @required this.onRootChangedCallback,
   });
@@ -27,6 +28,7 @@ class Treemap extends StatelessWidget {
     @required this.nodes,
     @required this.levelsVisible,
     @required this.isOutermostLevel,
+    @required this.width,
     @required this.height,
     @required this.onRootChangedCallback,
   });
@@ -62,6 +64,8 @@ class Treemap extends StatelessWidget {
   /// Whether current levelsVisible matches the outermost level.
   final bool isOutermostLevel;
 
+  final double width;
+
   final double height;
 
   final void Function(TreemapNode node) onRootChangedCallback;
@@ -72,7 +76,8 @@ class Treemap extends StatelessWidget {
 
   static const minHeightToDisplayTitleText = 100.0;
 
-  static const minHeightToDisplayCellText = 40.0;
+  static const minWidthToDisplayCellText = 40.0;
+  static const minHeightToDisplayCellText = 50.0;
 
   /// Computes the total size of a given list of treemap nodes.
   /// [endIndex] defaults to nodes.length - 1.
@@ -131,8 +136,10 @@ class Treemap extends StatelessWidget {
   /// |      |      |      |
   /// ----------------------
   /// ```
-  List<Positioned> buildTreemaps({
+  List<PositionedCell> buildTreemaps({
     @required List<TreemapNode> children,
+    @required double x,
+    @required double y,
     @required double width,
     @required double height,
   }) {
@@ -146,23 +153,25 @@ class Treemap extends StatelessWidget {
     // Sort the list of treemap nodes, descending in size.
     children.sort((a, b) => b.unsignedByteSize.compareTo(a.unsignedByteSize));
     if (children.length <= 2) {
-      final positionedChildren = <Positioned>[];
+      final positionedChildren = <PositionedCell>[];
       double offset = 0;
 
       for (final child in children) {
         final ratio = child.unsignedByteSize / totalByteSize;
 
         positionedChildren.add(
-          Positioned(
-            left: isHorizontalRectangle ? offset : 0.0,
-            top: isHorizontalRectangle ? 0.0 : offset,
+          PositionedCell(
+            left: isHorizontalRectangle ? x + offset : x,
+            top: isHorizontalRectangle ? y : y + offset,
             width: isHorizontalRectangle ? ratio * width : width,
             height: isHorizontalRectangle ? height : ratio * height,
+            rootNode: child,
             child: Treemap.fromRoot(
               rootNode: child,
               levelsVisible: levelsVisible - 1,
               isOutermostLevel: false,
               onRootChangedCallback: onRootChangedCallback,
+              width: width,
               height: height,
             ),
           ),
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -382,44 +357,39 @@ class Treemap extends StatelessWidget {
...
+        child: buildSelectable(
+          selectable: !isOutermostLevel,
+          child: Column(
+            children: [
+              if (height > minHeightToDisplayTitleText) buildTitleText(context),
+              Expanded(
+                child: Treemap.fromNodes(
+                  nodes: rootNode.children,
+                  levelsVisible: levelsVisible,
+                  isOutermostLevel: isOutermostLevel,
```
kenzieschmoll, line 372: will isOutermostLevel ever be true if we are buildling a treemap `fromNodes`? if not, can we remove this parameter or make it default to false

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -429,53 +399,63 @@ class Treemap extends StatelessWidget {
...
-          decoration: BoxDecoration(
-            border: Border.all(color: Colors.black87),
-          ),
-          child: buildNameAndSizeText(
-            textColor: Theme.of(context).textTheme.bodyText2.color,
-            oneLine: true,
-          ),
+        child: Text(
+          rootNode.displayText(),
+          style: TextStyle(color: Theme.of(context).textTheme.bodyText2.color),
```
kenzieschmoll, line 412: this should be the same color you are using in the cell painters (Colors.black87). Either put in a const or use the color from theme in the painter

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -382,44 +357,39 @@ class Treemap extends StatelessWidget {
...
-                child: Center(
-                  child: height > minHeightToDisplayCellText
-                      ? buildNameAndSizeText(
-                          textColor:
-                              rootNode.showDiff ? Colors.white : Colors.black,
-                          oneLine: false,
-                        )
-                      : const SizedBox(),
-                ),
+            child: CustomPaint(
```
kenzieschmoll, line 386: Can't this stay the same and just return a widget via `buildSelectable`? this isn't the level we are painting for other nodes. Other nodes at this level are widgets whose children are painted.

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: So the `Colors.black87` is for drawing the cell border, and the current font color for cell painter is `showDiff ? Colors.white : Colors.black` to ensure the text is visible depending on the diff color. For this specific line, we're building the text for the title bar, which doesn't depend on the diff color but the theme.

packages/devtools_app/lib/src/charts/treemap.dart

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -590,3 +575,132 @@ class TreemapNode extends TreeNode<TreemapNode> {
...
+    final rootNode = positionedCell.rootNode;
+
+    final bounds = Rect.fromLTWH(
+      positionedCell.left,
+      positionedCell.top,
+      size.width,
+      size.height,
+    );
+
+    final rectPaint = Paint()..color = rootNode.displayColor;
```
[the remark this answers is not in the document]
kenzieschmoll: have you tried `flutter format`?

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:devtools_app/src/inspector/layout_explorer/flex/flex.dart';
```
kenzieschmoll, line 5: why are we importing code from the layout_explorer?

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -429,53 +399,63 @@ class Treemap extends StatelessWidget {
...
   ///
   /// Selecting this widget will trigger a re-root of the tree
   /// to the associated [TreemapNode].
   ///
   /// The default value for newRoot is [rootNode].
-  Tooltip buildSelectable({@required Widget child, TreemapNode newRoot}) {
+  Widget buildSelectable({
+    @required Widget child,
+    TreemapNode newRoot,
+    bool selectable = true,
```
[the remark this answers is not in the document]
kenzieschmoll: if it is not selectable, you are just returning child. So at the call site where you are passing selectable: false, don't call `buildSelectable` and just use the child directly when that condition is met. then you can remove the selectable param from this method, which is surprising since a "selectable" should always be selectable, intuitively.

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -498,16 +496,21 @@ class Treemap extends StatelessWidget {
...
+          height: constraints.maxHeight,
         );
+        if (levelsVisible <= 1) {
+          return CustomPaint(
+            painter: MultiCellPainter(nodes: positionedChildren),
+            size: Size(constraints.maxWidth, constraints.maxHeight),
+          );
+        } else {
+          return Stack(children: positionedChildren);
+        }
```
kenzieschmoll, line 513: can you add a comment to each case explaining when we hit either.

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -360,23 +359,9 @@ class Treemap extends StatelessWidget {
...
-                  onRootChangedCallback: onRootChangedCallback,
-                  width: width,
-                  height: height,
-                ),
-              ),
-            ],
-          ),
-        ),
+        child: isOutermostLevel
+            ? buildTreemapFromNodes(context)
```
kenzieschmoll, line 363: store the output of buildTreemapFromNodes in a var instead of calling it twice

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -504,11 +499,13 @@ class Treemap extends StatelessWidget {
           height: constraints.maxHeight,
         );
         if (levelsVisible <= 1) {
+          // If this is the second to the last level, paint all widgets in the last level.
```
kenzieschmoll, line 502: nit: add reasoning why (e.g. to improve performance)

kenzieschmoll: we want someone in the future who is looking at this code to understand why we are painting some cells and using widgets for others

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Sounds good. done

PR #2253, opened 2020-08-07
Add code size entry point to landing page

kenzieschmoll:
You can navigate back to the landing page (connect screen) by pressing back in the web brower.

Screen Shot 2020-08-07 at 3 17 55 PM
Screen Shot 2020-08-07 at 3 38 24 PM


packages/devtools_app/lib/src/landing_screen.dart
```diff
@@ -15,66 +16,111 @@ import 'theme.dart';
...
         ),
+        const PaddedDivider(),
+        child,
+        const PaddedDivider(padding: EdgeInsets.symmetric(vertical: 10.0)),
       ],
     );
   }
 
-  Widget _buildTextInput() {
+  Widget _headerText(String text) {
```
jacob314, line 109: nit: seems a little surprising that header is bodyText1 and subHeaderText is caption as the two don't seem that semantically related. Is there anything semantically similar in the material theme? It would be nice if these names generally matched names from the material theme.

kenzieschmoll: good point. fixed.

PR #2256, opened 2020-08-10
Analytics Notification

grouma:
- Update entry point HTML to load `devtools_analytics.js`
- Update `analtyics.js` to look for search query parameters across the entire URL
  - Note this was required since Flutter Web routing embeds a `#`
- Move all analytics related logic under folder called `analytics`
- Create `AnalyticsPrompt` which prompts a user to provide consent for usage analytics collection
- Add corresponding tests

Note this PR simply adds support for analytics and issues an initial call. I'll follow up with a PR that calls the analytics for each screen.

Screenshot:
Screen Shot 2020-08-10 at 9 52 05 AM


kenzieschmoll:
Can you add a screenshot of AnalyticsNotificaiton?

packages/devtools_app/lib/src/analytics_notification.dart
```diff
@@ -0,0 +1,196 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
+
+import 'theme.dart';
+import 'ui/analytics.dart' as analytics;
+import 'ui/analytics_platform.dart' as platform;
```
kenzieschmoll, line 10: analytics_platform has a direct dependency on dart:html, which will be problematic for the desktop version of DevTools. Have you tested the desktop version? I would expect something would be blowing up

packages/devtools_app/lib/src/analytics_notification.dart
```diff
@@ -0,0 +1,196 @@
...
+          onPressed: () {
+            _provider.setDontAllowAnalytics();
+            setState(() {
+              _isVisible = false;
+            });
+          },
+          color: Colors.grey,
+        ),
+        const Padding(
+          padding: EdgeInsets.only(left: 20.0),
```
kenzieschmoll: use `defaultSpacing` const

packages/devtools_app/lib/src/analytics_notification.dart
```diff
@@ -0,0 +1,196 @@
...
+                    padding: const EdgeInsets.all(defaultSpacing),
+                    child: Column(
+                      mainAxisSize: MainAxisSize.min,
+                      crossAxisAlignment: CrossAxisAlignment.start,
+                      children: [
+                        Text(
+                          'Send usage statistics for DevTools?',
+                          style: textTheme.headline5,
+                        ),
+                        const Padding(padding: EdgeInsets.only(top: 20.0)),
```
kenzieschmoll: use defauiltSpacing const or add another const to theme.dart if we need 20.0 (e.g. wideSpacing or similar)

packages/devtools_app/lib/src/analytics_notification.dart
grouma: Moved things around. It's now using a conditional import and uses a stub on the desktop version.

packages/devtools_app/lib/src/analytics_notification.dart
```diff
@@ -0,0 +1,196 @@
...
+  Future<void> initialize() async {
+    analytics.exposeGaDevToolsEnabledToJs();
+    if (analytics.isGtagsReset()) {
+      await analytics.resetDevToolsFile();
+    }
+  }
+
+  bool _isEnabled;
+  @override
+  Future<bool> get isEnabled async => _isEnabled ??= await analytics.isEnabled;
```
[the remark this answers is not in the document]
grouma: Will do. We should have a lint for this if we want to enforce it.

packages/devtools_app/lib/src/analytics_notification.dart
```diff
@@ -0,0 +1,196 @@
...
+                        Text(
+                          'Send usage statistics for DevTools?',
+                          style: textTheme.headline5,
+                        ),
+                        const Padding(padding: EdgeInsets.only(top: 20.0)),
+                        _analyticsDescription(textTheme),
+                        const SizedBox(height: denseRowSpacing),
+                        _actionButtons(),
+                      ],
+                    )),
```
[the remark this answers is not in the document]
grouma: Will do. Same as above though. We should have lints.

packages/devtools_app/lib/src/analytics/remote_provider.dart
```diff
@@ -19,10 +19,10 @@ class _RemoteAnalyticsProvider implements AnalyticsProvider {
   @override
   Future<bool> get isEnabled async => _isEnabled ??= await analytics.isEnabled;
```
kenzieschmoll, line 20: put getter before _isEnabled

packages/devtools_app/lib/src/analytics/prompt.dart
```diff
@@ -0,0 +1,157 @@
...
+    }();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final textTheme = Theme.of(context).textTheme;
+    return Column(
+      mainAxisAlignment: MainAxisAlignment.spaceBetween,
+      children: [
+        Expanded(
```
kenzieschmoll, line 64: can these two widgets be removed? I don't think we need the extra column because it just has a single child, (Expanded) which also has a single child

packages/devtools_app/lib/src/analytics/prompt.dart
```diff
@@ -0,0 +1,151 @@
...
+
+  final Widget _child;
+  final AnalyticsProvider _provider;
+
+  bool _isVisible = false;
+
+  @override
+  void initState() {
+    super.initState();
+    () async {
```
jacob314, line 39: having an unhandled await in initState is always a bit suspect.
I would suggest awaiting for _provider.initialize() outside of a widget. Ideally all this before any UI is rendered perhaps from main.dart
that way you won't get a flash of 1 frame of content without a prompt. Delaying the first frame rendered for the analytics status wouldn't be a big deal. Another reasonable option would be to move this to a controller.

grouma: I'll follow up with a PR that fixes this issue. Want to get analytics turned on asap.

PR #2259, opened 2020-08-10
Display loading message when processing files for code size page

peterdjlee:
* Display loading message when processing files for code size page
ezgif com-video-to-gif

### DIFF
--- a/packages/devtools_app/lib/src/code_size/code_size_controller.dart
+++ b/packages/devtools_app/lib/src/code_size/code_size_controller.dart
@@ -125,7 +125,7 @@ class CodeSizeController {
   // TODO(peterdjlee): Spawn an isolate to run parts of this function to
   //                   prevent the UI from freezing and display a circular
   //                   progress indicator on code size screen. Needs flutter
-  //                   web to support working with isolates. (See #33577)
+  //                   web to support working with isolates. See #33577.
   void loadTreeFromJsonFile(DevToolsJsonFile jsonFile) async {
     changeSnapshotJsonFile(jsonFile);
 
@@ -158,7 +158,7 @@ class CodeSizeController {
   // TODO(peterdjlee): Spawn an isolate to run parts of this function to
   //                   prevent the UI from freezing and display a circular
   //                   progress indicator on code size screen. Needs flutter
-  //                   web to support working with isolates. (See #33577)
+  //                   web to support working with isolates. See #33577.
   void loadDiffTreeFromJsonFiles(
     DevToolsJsonFile oldFile,
     DevToolsJsonFile newFile,
--- a/packages/devtools_app/lib/src/code_size/code_size_controller.dart
+++ b/packages/devtools_app/lib/src/code_size/code_size_controller.dart
@@ -118,8 +118,23 @@ class CodeSizeController {
     changeDiffRoot(_activeDiffRoot);
   }
 
-  void loadTreeFromJsonFile(DevToolsJsonFile jsonFile) {
+  /// Notifies that the json files are currently being processed.
+  ValueListenable get processingNotifier => _processingNotifier;
+  final _processingNotifier = ValueNotifier<bool>(false);
+
+  // TODO(peterdjlee): Spawn an isolate to run parts of this function to
+  //                   prevent the UI from freezing and display a circular
+  //                   progress indicator on code size screen. Needs flutter
+  //                   web to support working with isolates. (See #33577)
+  void loadTreeFromJsonFile(DevToolsJsonFile jsonFile) async {
     changeSnapshotJsonFile(jsonFile);
+
+    _processingNotifier.value = true;
+
+    // Free up the thread for the code size page to display the loading message.
+    // Without passing in a high value, the value listenable builder in the code
+    // size screen does not get updated.
+    await delayForBatchProcessing(micros: 10000);
 
     Map<String, dynamic> processedJson;
     if (jsonFile.isApkFile) {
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -45,6 +45,8 @@ class CodeSizeScreen extends Screen {
   @visibleForTesting
   static const diffViewTreemapKey = Key('Diff View Treemap');
 
+  static const loadingMessage = 'Loading data...\nUI may be unresponsive.';
```
kenzieschmoll, line 48: nit: how about "Loading data...\nPlease do not refresh or leave this page."

packages/devtools_app/lib/src/code_size/code_size_controller.dart
```diff
@@ -118,9 +118,18 @@ class CodeSizeController {
...
+  ValueListenable get processingNotifier => _processingNotifier;
+  final _processingNotifier = ValueNotifier<bool>(false);
+
+  Future<void> loadTreeFromJsonFile(DevToolsJsonFile jsonFile) async {
     changeSnapshotJsonFile(jsonFile);
 
+    _processingNotifier.value = true;
+
+    // Free up the thread for the code size page to display the loading message.
+    await delayForBatchProcessing(micros: 10);
```
kenzieschmoll, line 131: is this necessary? here and for the diff case below

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -204,6 +208,12 @@ class _SnapshotViewState extends State<SnapshotView> with AutoDisposeMixin {
     });
 
     addAutoDisposeListener(controller.snapshotJsonFile);
+
+    addAutoDisposeListener(controller.processingNotifier, () {
+      setState(() {
+        processing = controller.processingNotifier.value;
+      });
+    });
```
kenzieschmoll, line 216: instead of setting state for the whole widget, why not just use a ValueListenableBuilder widget that handles the state management for processing?

same comment below

packages/devtools_app/lib/src/code_size/code_size_screen.dart
peterdjlee: sg, done

packages/devtools_app/lib/src/code_size/code_size_controller.dart
peterdjlee: having micros as 0 didn't seem to be enough for the screen to display the loading text. Was surprised to see this given that the documentation states 0 micros should be enough.

packages/devtools_app/lib/src/code_size/code_size_screen.dart
peterdjlee: Makes sense! done

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -209,11 +208,7 @@ class _SnapshotViewState extends State<SnapshotView> with AutoDisposeMixin {
 
     addAutoDisposeListener(controller.snapshotJsonFile);
 
-    addAutoDisposeListener(controller.processingNotifier, () {
-      setState(() {
-        processing = controller.processingNotifier.value;
-      });
-    });
+    addAutoDisposeListener(controller.processingNotifier);
```
kenzieschmoll, line 211: you don't need to listen to the notifier if you use a value listenable builder. If you don't specify a listener in `addAutoDisposeListener`, `setState` is still called on the whole widget. So you can remove this line

packages/devtools_app/lib/src/code_size/code_size_controller.dart
```diff
@@ -123,7 +123,7 @@ class CodeSizeController {
   final _processingNotifier = ValueNotifier<bool>(false);
 
   // TODO(peterdjlee): Spawn an isolate to run parts of this function to
-  //                   prevent the UI from freezing and display a circular 
+  //                   prevent the UI from freezing and display a circular
```
kenzieschmoll, line 126: OH are you awaiting loadTreeFromJson file where it is called in the UI? this could be why the state was not being updated

peterdjlee: I tried running it after removing all the asyncs and awaits, but seems like that doesn't fix it. In `loadTreeFromJsonFile`, commenting out everything after `processingNotifier.value = true` makes the loading message pop up, so it seems like the thread is frozen as soon as we enter the function and stays that way until we get out. I'm not sure if this is expected behavior?

packages/devtools_app/lib/src/code_size/code_size_controller.dart
```diff
@@ -118,9 +118,24 @@ class CodeSizeController {
...
-  void loadTreeFromJsonFile(DevToolsJsonFile jsonFile) {
+  /// Notifies that the json files are currently being processed.
+  ValueListenable get processingNotifier => _processingNotifier;
+  final _processingNotifier = ValueNotifier<bool>(false);
+
+  // TODO(peterdjlee): Spawn an isolate to run parts of this function to
+  //                   prevent the UI from freezing and display a circular
+  //                   progress indicator on code size screen. Needs flutter
+  //                   web to support working with isolates. (See #33577)
+  void loadTreeFromJsonFile(DevToolsJsonFile jsonFile) async {
```
kenzieschmoll, line 129: I'm suggesting that this should return a Future, and then where you call `loadTreeFromJson` (in code_size_screen.dart) instead call `await loadTreeFromJson`

peterdjlee: Ah yeah, before the last commit I was calling `await loadTreeFromJson`

PR #2266, opened 2020-08-11
check in untracked git files

devoncarew:
- check in two untracked git files

One of them is the `tool/` dir's lock file (`tool/pubspec.lock`). I don't think it's as important to use the exact same package versions here as it is for the code we ship to users (devtools_app, ...). We can either check this lock file in, or add it to a gitignore specific to the tool/ directory - happy to do either there.


devoncarew:
note, I switched this to adding a .gitignore for the lock file in `tool/`.

PR #2269, opened 2020-08-11
Enable zoom by click-on-cell by coordinate matching for Treemap

peterdjlee:
Enable zoom by click-on-cell by coordinate matching for Treemap
  * Iterate through the list of painted cells to find overlapping cell with clicked point.


packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -358,31 +359,26 @@ class Treemap extends StatelessWidget {
...
     if (rootNode.children.isNotEmpty) {
       final treemapFromNodes = buildTreemapFromNodes(context);
-      return Padding(
-        padding: const EdgeInsets.all(1.0),
-        child: isOutermostLevel
-            ? treemapFromNodes
-            : buildSelectable(child: treemapFromNodes),
-      );
+      return Padding(padding: const EdgeInsets.all(1.0), child: treemapFromNodes
+          // : buildSelectable(child: treemapFromNodes),
```
jacob314, line 363: remove commented out code.

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -513,6 +517,32 @@ class Treemap extends StatelessWidget {
...
     );
   }
+
+  void _onTapDown(
+    TapDownDetails details,
+    List<PositionedCell> positionedCells,
+  ) {
+    final x = details.localPosition.dx;
+    final y = details.localPosition.dy;
+    // Define target point as rectangle with width and height as 0.01 to
```
jacob314, line 527: it is more standard to just intersect a point with the rectangle rather than giving the mouse a "width"

packages/devtools_app/lib/src/charts/treemap.dart

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: Done

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -513,6 +517,32 @@ class Treemap extends StatelessWidget {
...
+      final cellRect = Rect.fromLTWH(
+        cell.left,
+        cell.top,
+        cell.width,
+        cell.height,
+      );
+      // TODO(peterdjlee): Optimize with better algorithm to find the overlapping cell.
+      if (touchPoint.overlaps(cellRect)) {
+        // Ensure the new root has children to show when zoomed into.
+        if (cell.node.children.isNotEmpty) {
```
jacob314, line 547: This special case seems a little surprising. Why not let users zoom all the way in?
If we don't want to we should select the parent of the selected node instead.

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -513,6 +520,36 @@ class Treemap extends StatelessWidget {
...
+  ) {
+    final x = details.localPosition.dx;
+    final y = details.localPosition.dy;
+    // Define target point as rectangle with width and height as 0.01 to
+    // represent the touch point.
+    final touchPoint = Offset(x, y);
+    // TODO(peterdjlee): Optimize with better algorithm to find the overlapping cell.
+    for (final cell in positionedCells) {
+      final cellRect = Rect.fromLTWH(
+        cell.left,
```
kenzieschmoll, line 540: we should just store these values in a rect on the cell instead of having individual fields for left, top, width, height on the cell

packages/devtools_app/lib/src/charts/treemap.dart
kenzieschmoll: delete the comment now that it is obsolete

packages/devtools_app/lib/src/charts/treemap.dart

packages/devtools_app/lib/src/charts/treemap.dart
peterdjlee: For now I made it select the parent of the node, but will take a look at just zooming down to a single cell. Might need to shift around the code a bit.

kenzieschmoll:
Can you add a test?

peterdjlee:
Now users zoom all the way down to a single cell.
Screen Shot 2020-08-13 at 11 11 04 AM

kenzieschmoll:
Why does the table say "no children to show"? The table should match the tree and have a single node (e.g. libflutter.so in this example)

packages/devtools_app/lib/src/code_size/code_size_table.dart
```diff
@@ -43,14 +43,26 @@ class CodeSizeSnapshotTable extends StatelessWidget {
...
-      dataRoots: rootNode.children,
-      columns: columns,
-      treeColumn: treeColumn,
-      keyFactory: (node) => PageStorageKey<String>(node.name),
-      sortColumn: sortColumn,
-      sortDirection: SortDirection.descending,
-    );
+    if (rootNode.children.isNotEmpty) {
+      return TreeTable<TreemapNode>(
+        dataRoots: rootNode.children,
```
kenzieschmoll, line 48: This should probably just be `[rootNode]`

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -477,41 +526,30 @@ class Treemap extends StatelessWidget {
...
+  ///
+  /// If so, reroot to the matching cell if the cell has children to show.
+  void _onTapDown(
+    TapDownDetails details,
+    List<PositionedCell> positionedCells,
+  ) {
+    final x = details.localPosition.dx;
+    final y = details.localPosition.dy;
+    final touchPoint = Offset(x, y);
+    // TODO(peterdjlee): Optimize with better algorithm to find the overlapping cell.
```
jacob314, line 547: nit: better -> more efficient
just to be clear that this algorithm is correct and you are concerned about the case where positionedCells is large. You could say this is O(positionedCells.length) while an optimized algorithm could be O(log(positionCells.length) if you want to be a little clearer.

PR #2270, opened 2020-08-12
Hide the timeline slider and expose Legend.

terrylucas:
Clicking on the Legend button will display a semi-opaque the legend overlay:

image

If Android Memory isn't visible then the legend only shows the legends of the charts visible e.g.,

image


packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -345,10 +367,188 @@ class MemoryBodyState extends State<MemoryBody> with AutoDisposeMixin {
...
 
+  final legendKey = GlobalKey(debugLabel: MemoryScreen.legendKeyName);
+  static const legendXOffset = 20;
+  static const legendYOffset = 7.0;
+  static const legendWidth = 200.0;
+  static const legendTextWidth = 55.0;
+  static const legendHeight1Chart = 185.0;
+  static const legendHeight2Charts = 340.0;
+
+  // TODO(terry): Consider custom painter?
```
jacob314, line 392: A custom painter would be nice so that these stay in sync with the images used in the chart itself.

PR #2273, opened 2020-08-12
Make hint text for entering a vm service url less confusing.

jacob314:
@csells noted that the hint text for the vm service url makes it look like one has already been typed.
Before:
Screen Shot 2020-08-12 at 11 00 30 AM
After:
Screen Shot 2020-08-12 at 10 38 26 AM

Tip: when reviewing CLs like this be sure to turn off white space only diffs. Wish we could set that as the default.

### DIFF
--- a/packages/devtools_app/lib/src/landing_screen.dart
+++ b/packages/devtools_app/lib/src/landing_screen.dart
@@ -147,8 +147,8 @@ class _LandingScreenBodyState extends State<LandingScreenBody> {
               Padding(
                 padding: const EdgeInsets.all(8.0),
                 child: Text(
-                  '(http://127.0.0.1:12345/auth_code=)',
-                  textAlign: TextAlign.end,
+                  '(e.g., http://127.0.0.1:12345/auth_code=...)',
+                  textAlign: TextAlign.start,
                   style: Theme.of(context).textTheme.caption,
                 ),
               ),
--- a/packages/devtools_app/lib/src/landing_screen.dart
+++ b/packages/devtools_app/lib/src/landing_screen.dart
@@ -145,7 +145,7 @@ class _LandingScreenBodyState extends State<LandingScreenBody> {
                 controller: connectDialogController,
               ),
               Padding(
-                padding: const EdgeInsets.all(8.0),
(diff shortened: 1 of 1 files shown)


jacob314:
Here is how to turn off whitespace diffs:
Screen Shot 2020-08-12 at 11 03 45 AM

kenzieschmoll:
DBC - can we left align this and add "e.g. " inside the parenthesis?

packages/devtools_app/lib/src/landing_screen.dart
```diff
@@ -123,23 +123,36 @@ class _LandingScreenBodyState extends State<LandingScreenBody> {
...
+                    // of hard coding material colors.
+                    borderSide: BorderSide(width: 0.5, color: Colors.grey),
+                  ),
+                ),
+                controller: connectDialogController,
+              ),
+              Padding(
+                padding: const EdgeInsets.all(8.0),
+                child: Text(
+                  '(http://127.0.0.1:12345/auth_code=)',
```
devoncarew, line 150: perhaps, `e.g., http://127.0.0.1:12345/auth_code=...`

jacob314: done.

jacob314:
> DBC - can we left align this and add "e.g. " inside the parenthesis?

done. The left align is a bit less distracting.

jacob314:
Screen Shot 2020-08-12 at 1 22 11 PM

packages/devtools_app/lib/src/landing_screen.dart
```diff
@@ -147,8 +147,8 @@ class _LandingScreenBodyState extends State<LandingScreenBody> {
               Padding(
                 padding: const EdgeInsets.all(8.0),
```
kenzieschmoll, line 148: can you remove the left edge padding so it aligns with the edge of the text field?

jacob314: The padding was intentional to try to make the tip look more linked to the input field but perhaps it just looks ugly.

jacob314: Screen Shot 2020-08-13 at 10 37 37 AM

PR #2275, opened 2020-08-13
Checkbox polish

jacob314:
Fyi @terrylucas 

Cleaning up the NotifierCheckbox so it is more convenient to reuse in the inspector page.
This class now gives us a really terse checkbox fully linked to a ValueNotifier. I recommend we use it for all our UI that needs to link checkboxes to ValueNotifier objects in our controllers.

### DIFF
--- a/packages/devtools_app/lib/src/memory/memory_filter.dart
+++ b/packages/devtools_app/lib/src/memory/memory_filter.dart
@@ -139,24 +139,6 @@ class SnapshotFilterState extends State<SnapshotFilterDialog>
     controller = widget.controller;
 
     cancel();
-
-    // Detect and handle checkboxes state changing.
-    addAutoDisposeListener(controller.filterPrivateClassesListenable, () {
-      setState(() {
-        privateClasses.notifier.value = controller.filterPrivateClasses.value;
-      });
-    });
-    addAutoDisposeListener(controller.filterLibraryNoInstancesListenable, () {
-      setState(() {
-        libraryNoInstances.notifier.value =
-            controller.filterLibraryNoInstances.value;
-      });
-    });
-    addAutoDisposeListener(controller.filterZeroInstancesListenable, () {
-      setState(() {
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/ui/utils.dart
```diff
@@ -7,44 +7,71 @@
...
-/// ValueNotifier associated with a Checkbox widget.
-class CheckboxValueNotifier extends ValueNotifier<bool> {
-  CheckboxValueNotifier(bool value) : super(value);
+class _NotifierCheckboxState extends State<NotifierCheckbox>
+    with AutoDisposeMixin {
+  bool currentValue;
+
+  @override
+  void initState() {
+    // TODO: implement initState
```
kenzieschmoll, line 41: remove todo

packages/devtools_app/lib/src/memory/memory_filter.dart
```diff
@@ -139,24 +139,6 @@ class SnapshotFilterState extends State<SnapshotFilterDialog>
     controller = widget.controller;
 
     cancel();
-
-    // Detect and handle checkboxes state changing.
-    addAutoDisposeListener(controller.filterPrivateClassesListenable, () {
-      setState(() {
-        privateClasses.notifier.value = controller.filterPrivateClasses.value;
```
kenzieschmoll, line 146: are there no other parts of the UI that depend on the value of these notifiers?

packages/devtools_app/lib/src/ui/utils.dart
jacob314: done

packages/devtools_app/lib/src/memory/memory_filter.dart
jacob314: cleaned up the UI a little further so it is clear that this isn't needed.

kenzieschmoll:
lgtm_army

PR #2264, opened 2020-08-11
Update CustomPointerScrollable to latest flutter

kenzieschmoll:
We have copies of Scrollable and ScrollableState. This CL pulls in the latest changes from Flutter and should make our build go green. All changes are copies from Flutter.

PR #2276, opened 2020-08-13
Remove Restoration Framework

goderbauer:
https://github.com/flutter/devtools/pull/2264/ pulled in the restoration framework, whose API are still in flux. Since devtool doesn't seem to use restoration yet, I removed the framework again. Devtools dependency on that API was causing trouble for google3 rolls. Once the restoration framework is stabilized, I'll add this back in.

### DIFF
--- a/packages/devtools_app/lib/src/custom_pointer_scroll_view.dart
+++ b/packages/devtools_app/lib/src/custom_pointer_scroll_view.dart
@@ -9,7 +9,6 @@ import 'dart:ui';
 import 'package:flutter/gestures.dart';
 import 'package:flutter/rendering.dart';
 import 'package:flutter/scheduler.dart';
-import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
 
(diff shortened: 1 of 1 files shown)


devoncarew:
I see this failing on the CI with:

```
  error - The method 'debugIsSerializableForRestoration' isn't defined for the type '_CustomPointerScrollableState' at lib/src/custom_pointer_scroll_view.dart:365:12 - (undefined_method)
```

goderbauer:
@devoncarew I re-added the import that provides the debugIsSerializableForRestoration function. PTAL

kenzieschmoll:
Are our bots running against master still? @devoncarew I expect the CI will fail if so.

goderbauer:
@kenzieschmoll this change should work fine against master.

devoncarew:
> Are our bots running against master still? @devoncarew I expect the CI will fail if so.

Yup - I plan to switch them over after the next dev release of the framework.

PR #2284, opened 2020-08-17
Add autoExpandRoot parameter to TreeTable

peterdjlee:
Add autoExpandRoot parameter to TreeTable after removing it from CodeSizeTable

### DIFF
--- a/packages/devtools_app/lib/src/table.dart
+++ b/packages/devtools_app/lib/src/table.dart
@@ -218,6 +218,7 @@ class TreeTable<T extends TreeNode<T>> extends StatefulWidget {
     @required this.sortColumn,
     @required this.sortDirection,
     this.selectionNotifier,
+    this.autoExpandRoots = false,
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/table.dart
```diff
@@ -209,16 +209,17 @@ class SelectionNotifier<T> extends ValueNotifier<T> {
...
+  TreeTable(
+      {Key key,
+      @required this.columns,
+      @required this.treeColumn,
+      @required this.dataRoots,
+      @required this.keyFactory,
+      @required this.sortColumn,
+      @required this.sortDirection,
+      this.selectionNotifier,
+      this.autoExpandRoot})
```
kenzieschmoll, line 221: trailing comma, and rename to autoExpandRoots

peterdjlee: Done

PR #2286, opened 2020-08-18
Make Treemap a stateful widget and enable tooltip by saving hovered node

peterdjlee:
Make Treemap a stateful widget and enable tooltip by saving hovered node

### DIFF
--- a/packages/devtools_app/lib/src/charts/treemap.dart
+++ b/packages/devtools_app/lib/src/charts/treemap.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/rendering.dart';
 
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -519,16 +542,16 @@ class Treemap extends StatelessWidget {
   /// Selecting this widget will trigger a re-root of the tree
   /// to the associated [TreemapNode].
   ///
-  /// The default value for newRoot is [rootNode].
+  /// The default value for newRoot is [widget.rootNode].
   Widget buildSelectable({@required Widget child, TreemapNode newRoot}) {
-    newRoot ??= rootNode;
+    newRoot ??= widget.rootNode;
     return Tooltip(
```
kenzieschmoll, line 548: if we have a tooltip above, can we remove this tooltip?

peterdjlee: buildSelectable is used for title bars and cells on the first level without any children. In both cases, there is no Tooltip above the widgets since we're not painting them.

Issue #2287, opened 2020-08-18
Google3 URIs are not properly handled in the debug tree view


grouma:
See this example:

<img width="200" alt="Screen Shot 2020-08-18 at 11 18 52 AM" src="https://user-images.githubusercontent.com/1269969/90550358-b2ae1400-e144-11ea-952b-55f9ca1fb359.png">

devoncarew:
Note, the google3 schema urls look something like `Google3://foo/bar`.

PR #2288, opened 2020-08-18
fix an issue handling google3: paths

devoncarew:
- fix an issue handling google3: paths in the debugger scripts view
- fix https://github.com/flutter/devtools/issues/2287


packages/devtools_app/lib/src/debugger/debugger_model.dart
```diff
@@ -278,12 +278,28 @@ class FileNode extends TreeNode<FileNode> {
...
     for (var script in scripts) {
-      var parts = script.uri.split('/');
+      final uri = Uri.parse(script.uri);
+      final scheme = uri.scheme;
+      var parts = uri.path.split('/');
 
-      // Look for and handle dotted package names.
-      if (parts.isNotEmpty && parts.first.contains('.')) {
+      // handle google3:///foo/bar
+      if (parts.first.isEmpty) {
```
grouma, line 286: I don't follow why this check is necessary. Should we not always filter empty paths?

packages/devtools_app/lib/src/debugger/debugger_model.dart
```diff
@@ -278,12 +278,28 @@ class FileNode extends TreeNode<FileNode> {
...
+        parts = parts.where((part) => part.isNotEmpty).toList();
         parts = [
-          ...parts.first.split('.'),
+          '$scheme:${parts.first}',
+          ...parts.sublist(1),
```
grouma, line 290: Why the sublist here?

packages/devtools_app/lib/src/debugger/debugger_model.dart
devoncarew: I'll double check whether this is necessary.

packages/devtools_app/lib/src/debugger/debugger_model.dart
devoncarew: Here, we're concatenating the first part of the file path to the schema. This is so that user's see 'google3:foo' as the top level entry in the scripts list instead of `google3` (w/ lots of entries below it).

I'll add a comment. I may also pull this logic (parsing a uri) out into a new method, separate from the code to create a tree structure from the scripts.

packages/devtools_app/lib/src/debugger/debugger_model.dart
devoncarew: Yup - I checked - this is necessary for the google3 url case (and will be the only one w/ empty path segments).

packages/devtools_app/lib/src/debugger/debugger_model.dart
devoncarew: I added a comment about why this is here (and, refactoring the path segment munging into a separate class for clarity).

devoncarew (commented):
Updated - ptal

grouma (approved):

PR #2289, opened 2020-08-18
Add missing tests for code size page

peterdjlee:
* Add missing tests for code size page


packages/devtools_app/test/code_size_screen_test.dart
```diff
@@ -176,35 +220,88 @@ void main() {
...
+      expect(find.byType(DualFileImportContainer), findsOneWidget);
+      expect(find.byType(FileImportContainer), findsNWidgets(2));
+      expect(find.text(DiffView.importOldInstructions), findsOneWidget);
+      expect(find.text(DiffView.importNewInstructions), findsOneWidget);
+      expect(find.text('No File Selected'), findsNWidgets(2));
+    });
   });
 
-  group('CodeSizeController', () {
+  group('CodeSizeTestController', () {
```
kenzieschmoll, line 294: we can leave this as CodeSizeController

peterdjlee: Do you mean for any instance where we don't need to use delay? So we'd just use it for the case where we need to use the overridden `loadTreeFromJsonFile` with a delay.

kenzieschmoll: I just mean there is no need to rename the testing group - CodeSizeTestController is just an abstraction to make it easier for us to test

peterdjlee: Ah, sorry I didn't catch that at all haha, thanks!

packages/devtools_app/test/code_size_screen_test.dart
```diff
@@ -108,7 +120,11 @@ void main() {
       expect(find.text(SnapshotView.importInstructions), findsOneWidget);
       expect(find.text('No File Selected'), findsOneWidget);
 
-      await loadDataAndPump(tester);
+      codeSizeController.loadTreeFromJsonFile(newV8JsonFile, (error) => {},
+          delayed: true);
```
kenzieschmoll, line 124: why not use loadDataAndPump here?

packages/devtools_app/test/code_size_screen_test.dart
```diff
@@ -122,52 +138,80 @@ void main() {
...
-          name: 'lib/src/code_size/stub_data/new_v8.dart',
-          lastModifiedTime: lastModifiedTime,
-          data: json.decode(newV8),
-        ),
+        oldV8JsonFile,
+        newV8JsonFile,
         (error) => {},
+        delayed: true,
       );
+      await tester.pump(const Duration(milliseconds: 500));
```
kenzieschmoll, line 206: maybe add a similar helper for diff `loadDiffDataAndPump`

packages/devtools_app/test/code_size_screen_test.dart
```diff
@@ -176,35 +220,88 @@ void main() {
...
+        oldV8JsonFile,
+        newV8JsonFile,
+        (error) => {},
+      );
+      await tester.pumpAndSettle();
+
+      await tester.tap(find.byKey(CodeSizeScreen.dropdownKey));
+      await tester.pumpAndSettle();
+
+      await tester.tap(find.text('Increase Only').hitTestable());
```
kenzieschmoll, line 254: is `hitTestable()` necessary?

PR #2292, opened 2020-08-20
More analytics

grouma:
- Merge `analytics/platform.dart` into `analytics/analytics.dart`
- Don't export `constants.dart` from `analytics.dart`
- Create an `analytics_stub` to ensure the application continue to compile for native
- Add analytic calls for each screen


packages/devtools_app/lib/src/app.dart
```diff
@@ -8,6 +8,9 @@ import 'package:pedantic/pedantic.dart';
 import 'package:provider/provider.dart';
 
 import '../devtools.dart' as devtools;
+import 'analytics/analytics_stub.dart'
```
kenzieschmoll, line 11: i kind of like importing `as 'ga'` or `as 'analytics'`. This makes it clearer where we are making analytics calls. Here and elsewhere

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -83,6 +85,7 @@ class _CodeSizeBodyState extends State<CodeSizeBody>
   @override
   void initState() {
     super.initState();
+    screen(_id);
```
kenzieschmoll, line 88: All the other screens call `screen(ScreenClass.id)`. Why did we make code size a top level private field instead of being a static member on the class?

packages/devtools_app/lib/src/analytics/analytics_stub.dart
```diff
@@ -0,0 +1,23 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: can we add more granularity here? I can imagine we'll want to slice data at multiple levels (e.g. screen - feature - action, or screen - feature - subFeature - value).

maybe a generic `track` function with multiple slices? or `trackAction` and `trackValue`?

something like `selectFrame` could then be rolled into the `track` method

packages/devtools_app/lib/src/code_size/code_size_screen.dart
grouma: That was my original approach and I thought I cleaned up all cases. Missed this one though. Will fix.

packages/devtools_app/lib/src/analytics/analytics_stub.dart
grouma: We can easily add more functions or change the APIs in this stub later on. It's private so should be very easy to do.

packages/devtools_app/lib/src/analytics/remote_provider.dart
```diff
@@ -20,18 +23,19 @@ class _RemoteAnalyticsProvider implements AnalyticsProvider {
   final bool _isFirstRun;
 
   @override
-  bool get isGtagsEnabled => analytics.isGtagsEnabled();
+  bool get isGtagsEnabled => _isGtagsEnabled;
+  final bool _isGtagsEnabled;
 
   @override
-  void setAllowAnalytics() => platform.setAllowAnalytics();
+  void setAllowAnalytics() => setAllowAnalytics();
```
jacob314, line 30: unless i'm missing something these two methods would stack overflow.

packages/devtools_app/lib/src/app.dart
jacob314: +1 to importing libs with a lot level fields with a prefix.

packages/devtools_app/lib/src/analytics/remote_provider.dart
grouma: Indeed. Nice catch!

PR #2301, opened 2020-08-26
Fix a bug causing extra evaluation for primitive values

annagrin:
I discovered this bug during my implementation of expression evaluation in webdev (soon to follow,  I am making this change in advance so expression evaluation will work correctly in webdev when checked in):

`response.valueAsStringIsTruncated == false` was evaluated to `false` on primitive JS values that had it set to `null`, causing the devtools to call `toString()` method to extract the string value, which in turn resulted in a crash.

Fixed to check a boolean condition instead.

Also, had to upgrade vm_service version, as the version would resolve but the code wouldn't compile if version 4.1.0 is installed by pub from other projects, like on my machine.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/evaluate.dart
+++ b/packages/devtools_app/lib/src/debugger/evaluate.dart
@@ -172,8 +172,9 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField> {
   }
 
   Future<String> _getBestStringValue(InstanceRef response) async {
-    // Return the string value if present and not truncated.
-    // Note: string value is truncated iff not null and true
+    // Return the string value iff present and not truncated.
+    // Note: string value is truncated iff [valueAsStringIsTruncated]
+    // is not null and is set to true.
     if (response.valueAsString != null &&
         response.valueAsStringIsTruncated != true) {
       return response.valueAsString;
--- a/packages/devtools_app/lib/src/debugger/evaluate.dart
+++ b/packages/devtools_app/lib/src/debugger/evaluate.dart
@@ -174,8 +174,8 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField> {
   Future<String> _getBestStringValue(InstanceRef response) async {
     // Return the string value iff present and not truncated.
     // Note: [valueAsStringIsTruncated] is actually a tri-state:
-    // string value is truncated iff [valueAsStringIsTruncated]
-    // is not null and [valueAsStringIsTruncated] is true.
+    // string value is not truncated iff [valueAsStringIsTruncated]
+    // is null or [valueAsStringIsTruncated] is false.
     if (response.valueAsString != null &&
         response.valueAsStringIsTruncated != true) {
       return response.valueAsString;
--- a/packages/devtools_app/lib/src/debugger/evaluate.dart
+++ b/packages/devtools_app/lib/src/debugger/evaluate.dart
@@ -172,8 +172,7 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField> {
   }
 
   Future<String> _getBestStringValue(InstanceRef response) async {
-    if (response.valueAsString != null &&
-        response.valueAsStringIsTruncated == false) {
+    if (response.valueAsString != null && !response.valueAsStringIsTruncated) {
       return response.valueAsString;
     }
 
@@ -202,7 +201,7 @@ String _valueAsString(InstanceRef ref) {
     return ref.valueAsString;
   }
 
-  if (ref.valueAsStringIsTruncated == true) {
+  if (ref.valueAsStringIsTruncated) {
     return '${ref.valueAsString}...';
   } else {
     return ref.valueAsString;
--- a/packages/devtools_app/lib/src/debugger/evaluate.dart
+++ b/packages/devtools_app/lib/src/debugger/evaluate.dart
@@ -172,7 +172,10 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField> {
   }
 
   Future<String> _getBestStringValue(InstanceRef response) async {
-    if (response.valueAsString != null && !response.valueAsStringIsTruncated) {
+    // Return the string value if present and not truncated.
+    // Note: string value is truncated iff not null and true
+    if (response.valueAsString != null &&
+        response.valueAsStringIsTruncated != true) {
       return response.valueAsString;
     }
 
@@ -201,7 +204,7 @@ String _valueAsString(InstanceRef ref) {
     return ref.valueAsString;
   }
 
-  if (ref.valueAsStringIsTruncated) {
+  if (ref.valueAsStringIsTruncated == true) {
     return '${ref.valueAsString}...';
   } else {
     return ref.valueAsString;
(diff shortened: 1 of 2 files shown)


devoncarew:
Thanks for the PR!

> `response.valueAsStringIsTruncated == false` was evaluated to `false` on primitive JS values that had it set to `null`, causing the devtools to call `toString()` method to extract the string value, which in turn resulted in a crash.
> 
> Fixed to check a boolean condition instead.

Two thoughts:
- we were trying not to expect the `valueAsStringIsTruncated` field to be populated - we've hit issues were the value is null, and then using it where we expect a bool fails. I assume we should only assume the value is truncated if `response.valueAsStringIsTruncated == true` (i.e., not null, and equal to true)? I'll verify the logic in the PR.
- in terms of the crash, the client probably shouldn't call toString() if it has a non-truncated string value, however, it should still be able to call it (the backend should support the toString() call for all object types)

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -172,8 +172,7 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField> {
   }
 
   Future<String> _getBestStringValue(InstanceRef response) async {
-    if (response.valueAsString != null &&
-        response.valueAsStringIsTruncated == false) {
+    if (response.valueAsString != null && !response.valueAsStringIsTruncated) {
```
devoncarew, line 175: I think we might want to switch the logic to `response.valueAsStringIsTruncated != true` (not null and not true). Given that that's slightly convoluted to read in an expression, we might want an extension method to make this more readable (`InstanceRef.valueAsStringCheckTruncated`?). Not sure of a good name there, or if we should instead just clarify things in a comment above the expression.

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -202,7 +201,7 @@ String _valueAsString(InstanceRef ref) {
     return ref.valueAsString;
   }
 
-  if (ref.valueAsStringIsTruncated == true) {
+  if (ref.valueAsStringIsTruncated) {
```
devoncarew, line 204: I think we do want to keep the `ref.valueAsStringIsTruncated == true` check here, to guard against the valueAsStringIsTruncated field being null (we have had exceptions from this in the past).

packages/devtools_app/pubspec.yaml
```diff
@@ -38,7 +38,7 @@ dependencies:
   provider: ^4.0.0
   url_launcher: ^5.0.0
   url_launcher_web: ^0.1.1+6
-  vm_service: ^4.1.0
+  vm_service: ^4.2.0
```
devoncarew, line 41: 👍 thanks!

packages/devtools_app/lib/src/debugger/evaluate.dart
annagrin: >  I assume we should only assume the value is truncated if response.valueAsStringIsTruncated == true (i.e., not null, and equal to true)? I'll verify the logic in the PR.

It seems to be the case with my experiments.

> I think we might want to switch the logic to response.valueAsStringIsTruncated != true (not null and not true). 

I would like to say "null or false" instead, so I think the following would be the simplest  and clear - is this ok with you?

` response.valueAsStringIsTruncated == null  || response.valueAsStringIsTruncated == false `

> in terms of the crash, the client probably shouldn't call toString() if it has a non-truncated string value, however, it should still be able to call it (the backend should support the toString() call for all object types)

I agree, the crash looks like a bug. It happened because the VM proxy (dwds) couldn't find the primitive object by its reference. I am not sure what the reason is exactly, but I suspect that the object was simulated by the debugger and did not actually exist in chrome. We could fix this by cashing fake objects, but it does not look needed for values that have a string value, and the fix is good anyway to eliminate an unnecessary call. We can solve the problem of missing references separately, I will file a bug for webdev if I can recreate it for another case (like calling another functions on that object via invoke or evaluate API.

packages/devtools_app/lib/src/debugger/evaluate.dart
annagrin: sounds good, will revert!

packages/devtools_app/lib/src/debugger/evaluate.dart
annagrin: I changed my mind -  `added response.valueAsStringIsTruncated != true` as you suggested  - as it does mean what I wanted to say, "null or false", and added a clarifying comment.

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -172,8 +172,10 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField> {
   }
 
   Future<String> _getBestStringValue(InstanceRef response) async {
+    // Return the string value if present and not truncated.
+    // Note: string value is truncated iff not null and true
```
DanTup, line 176: typo: iff

"not null and true" looks a little ambiguous - "not null and not true" or "neither null or true" may be clearer.

devoncarew:
Re-running the dart2js bot...

devoncarew:
Hmm, the dart2js bot is consistently failing. I'm re-running it, but it it fails again it might be worth it to update this PR with the latest from master.

packages/devtools_app/lib/src/debugger/evaluate.dart
annagrin: I actually mean '(value != null) && (value == true)'. not sure how to express it the best:) 'Not null and not false' maybe, if it is not too convoluted?

annagrin:
@devoncarew I rebased onto the current master but the dart2js test is still failing, even after retrying. I'd appreciate any ideas!

packages/devtools_app/lib/src/debugger/evaluate.dart
DanTup: Oh sorry, I misunderstood the comment thinking it was referring to the return value inside the `if` but it's describing when the string in the `InstanceRef` is truncated.

In which case, I think

> [valueAsStringIsTruncated] is not null and is set to true

may be intended to say

> [valueAsString] is not null and [valueAsStringIsTruncated] is true

DanTup:
It looks like different test failures each time. The most recent one looks like it hung between launching Chrome and waiting for the DevTools server to register a client was connected (which could mean anything from Chrome failed to load, failed to load the DevTools webpage, or DevTools failed to connect back to the server).

I doubt it's related to this change. I'll see if I can repro on a fork today with more logging to figure out why it's failing (I think I did see this happen on a PR before too, but thought it was pretty infrequent).

DanTup:
I think I've got a fix for a race that occurred in the last failure here:

https://github.com/flutter/devtools/pull/2309

However I think there are other issues too (for example Sentinel isolates in some code that starts the app). I'm still trying to track them down.

DanTup:
I've landed the fix above which should make things better, but I'm still trying to get to the bottom of "Server reuses DevTools instance if not connected to a VM" failing (I think I'm sometimes reproducing this locally now) and the collected isolate issue.

devoncarew (approved):

packages/devtools_app/lib/src/debugger/evaluate.dart
annagrin: no i actually mean [valueAsStringIsTruncated] is not null and [valueAsStringIsTruncated] is true.

annagrin: I actually meant the [valueAsStringIsTruncated] in both cases. I changes it to following to remove the confusion:

    // Note: [valueAsStringIsTruncated] is actually a tri-state: 
    // string value is not truncated iff [valueAsStringIsTruncated]
    // is null or [valueAsStringIsTruncated] is false.

PR #2303, opened 2020-08-27
Use new BreadcrumbNavigator widget in code size screen

kenzieschmoll:
In the user studies, we observed that it was not obvious to users that the breadcrumbs were clickable. This breadcrumb Navigator is similar to an IDE breadcrumb navigator and should feel more familiar to users. The breadcrumbs also look more clickable in this version. 

Screen Shot 2020-08-27 at 2 34 36 PM

### DIFF
--- a/packages/devtools_app/lib/src/charts/treemap.dart
+++ b/packages/devtools_app/lib/src/charts/treemap.dart
@@ -514,15 +514,18 @@ class _TreemapState extends State<Treemap> {
 
   Widget buildBreadcrumbNavigator() {
     final pathFromRoot = widget.rootNode.pathFromRoot();
-    final breadcrumbs = List<Breadcrumb>.generate(pathFromRoot.length, (index) {
-      final node = pathFromRoot[index];
-      return Breadcrumb(
-        text: index < pathFromRoot.length - 1 ? node.name : node.displayText(),
-        isRoot: index == 0,
-        onPressed: () => widget.onRootChangedCallback(node),
-      );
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -512,29 +512,17 @@ class _TreemapState extends State<Treemap> {
...
-              index < pathFromRoot.length - 1
-                  ? pathFromRoot[index].name
-                  : pathFromRoot[index].displayText(),
-            ),
-            newRoot: pathFromRoot[index],
-          );
-        },
-      ),
-    );
+    final breadcrumbs = List<Breadcrumb>.generate(pathFromRoot.length, (index) {
```
jacob314, line 517: nit: there probably won't be enough breadcrumbs to matter but it could be nice to continue to use itemBuilder like the code was previously so you don't build the breadcrumb widgets until they are needed.  you're already using `List.generate` so the api shape seems about identical. I would just have the Breadcrumb navigator take an arbitrary widget as a child so you can specify the ListView here like it was done before.

kenzieschmoll: Switched to a builder constructor so that we still get the benefit of ListView.builder but the consumer does not have to be responsible for the List widget code.

kenzieschmoll:
Landing on red test flake - we may want to disable that test, as it is blocking development pretty bad lately.

PR #2309, opened 2020-09-03
Fix race in "server removes clients that disconnect from the API" test

DanTup:
This test sometimes would hang right at the start:

https://travis-ci.org/github/flutter/devtools/jobs/723601625#L642-L645

```
02:08 +9: /home/travis/build/flutter/devtools/packages/devtools_app/test/integration_tests/server_test.dart: Server (VM Service) server removes clients that disconnect from the API
<== {"event":"server.started","method":"server.started","params":{"host":"127.0.0.1","port":40195,"pid":5530,"protocolVersion":"1.1.0"}}
07:05 +9 -1: /home/travis/build/flutter/devtools/packages/devtools_app/test/integration_tests/server_test.dart: Server (VM Service) server removes clients that disconnect from the API [E]
  TimeoutException after 0:05:00.000000: Test timed out after 5 minutes.
```

I believe it's because we're using a broadcast stream (so it can be listened to throughout the test for expected events) but there's some async work (`await _startApp()`) after the streams are weird up in `setup` which means by the time the test starts listening, the event may have already occurred.

While testing this, I'm still seeing some failures, though I believe they are unrelated (Sentinel isolates in connecting to the VM service).

### DIFF
--- a/packages/devtools_app/test/integration_tests/server_test.dart
+++ b/packages/devtools_app/test/integration_tests/server_test.dart
@@ -23,7 +23,13 @@ DevToolsServerDriver server;
 final Map<String, Completer<Map<String, dynamic>>> completers = {};
 final StreamController<Map<String, dynamic>> eventController =
     StreamController.broadcast();
+
+/// A broadcast stream of events from the server.
+///
+/// Listening for "server.started" events on this stream may be unreliable because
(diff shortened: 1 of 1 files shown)


devoncarew (approved):
👍

PR #2310, opened 2020-09-03
Add more logging when running tests or running the server with --verbose

DanTup:
This adds logging for the Chrome processes spawned in tests, and also by the server (as `server.log` events if in machine mode) when run in `--verbose` mode. Chrome writes `console.log()` output to its `stderr` stream, so this may help track down failures caused by errors in the browser:

```
<== {"event":"server.log","params":{"message":"chrome stderr: \nDevTools listening on ws://127.0.0.1:9223/devtools/browser/281f8c27-2840-4dcb-b350-21682d2c7c8b\n"}}
```

However, it may be a bit spammy if users may use `--verbose` when running the server, so we may want to put this behind another flag (`--very-verbose`? `--trace`?) if the extra output would be unhelpful to them.

### DIFF
--- a/packages/devtools_app/test/integration_tests/server_test.dart
+++ b/packages/devtools_app/test/integration_tests/server_test.dart
@@ -310,7 +310,7 @@ void main() {
         // Spawn our own Chrome process so we can terminate it.
         final devToolsUri =
             'http://${event['params']['host']}:${event['params']['port']}';
-        final chrome = await Chrome.locate().start(url: devToolsUri);
+        final chrome = await _spawnChrome(devToolsUri);
 
         // Wait for DevTools to inform server that it's connected.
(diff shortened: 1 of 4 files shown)


DanTup:
> DevTools listening on 

Turns out this was not a message printed by Dart DevTools as I thought, this is Chrome (and it's Chrome DevTools). So it might not be the case that `console.log` stuff is showing up here. Trying to find if we can enable that though (since it would help with another issue I'm trying to track down!).

devoncarew:
Is this PR obsolete now?

DanTup:
Yup, thanks. We can get the console output, but it involves running with the inspection protocol and getting them that way. The issues I was trying to track down turned out not to be in the client so it didn't end up being useful anyway.

PR #2267, opened 2020-08-11
refactor our travis setup

devoncarew:
- some speculative work to refactor our travis setup

PR #2319, opened 2020-09-07
don't install webdev

devoncarew:
More CI refactoring:

- don't install webdev
- some cleanup around the path, and printing dart and flutter versions
- handle the 'waiting for another flutter command' text when starting tests
- handle the 'waiting for another flutter command' text in /devtools_server_driver.dart
- ensure that we use the `dart` from `flutter/bin/flutter/cache/dart-sdk/bin` and not `flutter/bin/dart` (the dart script from `flutter/bin/` will break everything everywhere due to https://github.com/flutter/devtools/pull/2319#issuecomment-688553860)

### DIFF
--- a/packages/devtools_app/test/integration_tests/app.dart
+++ b/packages/devtools_app/test/integration_tests/app.dart
@@ -23,7 +23,7 @@ void appTests() {
 
   test('can switch pages', () async {
     final DevtoolsManager tools =
-        DevtoolsManager(tabInstance, webdevFixture.baseUri);
+        DevtoolsManager(tabInstance, webBuildFixture.baseUri);
     await tools.start(appFixture);
     await tools.switchPage('logging');
 
@@ -33,9 +33,9 @@ void appTests() {
 
   test('connect dialog displays', () async {
     // start with no port
-    final Uri baseAppUri = webdevFixture.baseUri.resolve('index.html');
+    final Uri baseAppUri = webBuildFixture.baseUri.resolve('index.html');
     final DevtoolsManager tools =
-        DevtoolsManager(tabInstance, webdevFixture.baseUri);
+        DevtoolsManager(tabInstance, webBuildFixture.baseUri);
     await tools.start(
       appFixture,
       overrideUri: baseAppUri,
--- a/packages/devtools_app/test/integration_tests/debugger.dart
+++ b/packages/devtools_app/test/integration_tests/debugger.dart
@@ -25,7 +25,7 @@ void debuggingTests() {
     appFixture = await CliAppFixture.create('test/fixtures/debugging_app.dart');
 
     final DevtoolsManager tools =
-        DevtoolsManager(tabInstance, webdevFixture.baseUri);
+        DevtoolsManager(tabInstance, webBuildFixture.baseUri);
     await tools.start(appFixture);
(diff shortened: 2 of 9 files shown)


devoncarew:
A big issues that we're running into is that the `<flutter-sdk>/bin/dart` script can now print the text `'Waiting for another flutter command to release the startup lock...'`. This messes up many of our testing / cli tools, which try and communicate by printing json to stdout - they no longer emit valid json.

devoncarew:
We still have a flakey test, but I don't think these changes accentuated it.

devoncarew:
I'm going to TBR this, as a portion of https://github.com/flutter/devtools/pull/2267, but reviews after the fact are welcome.

devoncarew:
@jonahwilliams - I'll file an issue on flutter/flutter, but we're seeing issues when using the `dart` instance from <flutter-sdk>/bin (instead of the one from `flutter/bin/flutter/cache/dart-sdk/bin`). The one in the top-level bin/ directory can print out `'Waiting for another flutter command to release the startup lock...'`; this can cause issues with tools that communicated via stdio as that message won't be expected (for example, it won't be formatted as json).

To work around this, we're ensuring that our `PATH` is set up to find the dart in the internal cache of the flutter sdk. We might want to consider different behavior here. Two options that come to mind:

- only take an exclusive lock on `flutter/bin/dart` when we need to do something like update the sdk; don't lock if there's just another instance of `dart` or `flutter` running
- never lock on the `flutter/bin/dart` script; have it fast fail if there's nothing to execute. So, if `flutter/bin/cache/dart-sdk` did not yet exist, `flutter/bin/dart` might fail with a message; in this case, running `flutter` would properly provision the dart sdk. Running `flutter/bin/dart` when the dart sdk already existed would just delegate to it; it wouldn't do something like check a file lock.

I like the properties of the 2nd option - it leaves all the upgrade / sdk provisioning behavior to the `flutter` script. The `dart` script is just a simple pass-though, with largely the same behavior as the dart command that it delegates to.

jonahwilliams:
> To work around this, we're ensuring that our PATH is set up to find the dart in the internal cache of the flutter sdk.

This is bad, you will almost certainly end up using the wrong dart at some point.

> only take an exclusive lock on flutter/bin/dart when we need to do something like update the sdk

Yeah that seems like a reasonable change. TBH I'm not sure why that isn't the current behavior, it might have just been a simplification.

> never lock on the flutter/bin/dart script...

I'm not sure about this one. The dart script still wouldn't be a simple pass through, because it will need all of the same logic to check that its version is up to date. Anyone invoking dart would need to handle the case that the sdk was not yet downloaded, which seems like it could push the complexity out to users.

jonahwilliams:
At any rate, I will look into updating the locking behavior tomorrow to see if there are any issues

devoncarew:
> Anyone invoking dart would need to handle the case that the sdk was not yet downloaded, which seems like it could push the complexity out to users.

It would rely on something from the `flutter` tool being invoked before something from `dart`; I don't think that's an unreasonable simplification (`dart` will only work if `flutter`'s been run once`), but it does have an initial trip hazard.

> At any rate, I will look into updating the locking behavior tomorrow to see if there are any issues

Thanks!

PR #2323, opened 2020-09-09
Add pub_publish.sh script and update readme.

jacob314:
use --force to avoid manually confirming each publish operation.

### DIFF
--- a/tool/README.md
+++ b/tool/README.md
@@ -85,30 +85,9 @@ dart bin/devtools.dart
   generally work, and there are no exceptions in the chrome devtools log
 
 #### Publish the packages
-Paste each of these multiline blocks into your console, and follow the confirmation prompts to upload the packages.
 
 ```shell
-pushd packages/devtools_shared
-pub publish
-
-popd
-pushd packages/devtools_server
-pub publish
-
-popd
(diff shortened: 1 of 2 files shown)


devoncarew:
You may consider converting this to dart and adding to the repo tool (in `//bin/`).

tool/README.md
```diff
@@ -85,30 +85,9 @@ dart bin/devtools.dart
...
-pushd packages/devtools_app
-pub publish
-
-popd
-pushd packages/devtools
-pub publish
-
-popd
-
+./too/pub_publish.sh
```
kenzieschmoll, line 90: ./tool ?

kenzieschmoll (commented):
one question about the behavior: will this still exit on errors, but just force on warnings?

kenzieschmoll (approved):

tool/README.md
jacob314: done

jacob314:
Writing this as a dart script instead of a shell script might obfuscate what is going on a bit

> one question about the behavior: will this still exit on errors, but just force on warnings?

This will still exit on errors. We get warnings because we have tight version constraints and don't update changelogs for all of these packages.

Issue #2324, opened 2020-09-09
Integration tests are too flaky


jacob314:
Integration tests have frequent flakes due to race conditions leading to Sentinel values to sometimes be returned from the VM service.

DanTup:
I tracked this down to this code:

https://github.com/flutter/devtools/blob/69f253114a06dc870ae588ed4cd52650ed1c2ab9/packages/devtools_app/test/support/cli_test_driver.dart#L122-L125

Which calls this:

https://github.com/flutter/devtools/blob/69f253114a06dc870ae588ed4cd52650ed1c2ab9/packages/devtools_app/test/support/cli_test_driver.dart#L144-L147

We call `getVM()` and then call `getIsolate()` for each isolate that is returned. Sometimes we see an (non-system) isolate that appears in the `getVM()` results but then is not available by the time we call `getIsolate()`. I'm not sure what the isolate is (it seems separate to the system isolate that I presume is dartdev) but just ignoring the error allows the test to pass (so whatever the isolate is, it's not the one we care about). @bkonyi - any ideas?

Here's a log of our `getVM()` and `getIsolate()` calls when this happens:

```
    VM: <== {"jsonrpc":"2.0", "result":{"type":"VM","name":"vm","architectureBits":64,"hostCPU":"           Intel(R) Xeon(R) CPU","operatingSystem":"linux","targetCPU":"x64","version":"2.10.0-73.0.dev (dev) (Fri Aug 28 03:14:36 2020 -0700) on \"linux_x64\"","_profilerMode":"Dart","_nativeZoneMemoryUsage":0,"pid":6740,"startTime":1599728685306,"_embedder":"Dart VM","_maxRSS":69230592,"_currentRSS":56193024,"isolates":[],"systemIsolates":[{"type":"@Isolate","id":"isolates\/1232892005587247","name":"vm-service","number":"1232892005587247","isSystemIsolate":true}],"isolateGroups":[],"systemIsolateGroups":[{"type":"@IsolateGroup","id":"isolateGroups\/15112800809996415513","name":"vm-service","number":"15112800809996415513","isSystemIsolateGroup":true},{"type":"@IsolateGroup","id":"isolateGroups\/8991677360922067708","name":"dartdev","number":"8991677360922067708","isSystemIsolateGroup":true}],"_currentMemory":42708992},"id":"9"}
    VM: ==> {"jsonrpc":"2.0","id":"10","method":"getVM","params":{}}
[...]

PR #2328, opened 2020-09-10
Fix flaky tests due to Isolates that appear/vanish at startup

DanTup:
This code assumes all isolates present in a call to `getVM()` are still valid moments later (and `getIsolate()` will not return a Sentinel). This is not currently true - isolates can appear/disappear very quickly at startup. See https://github.com/flutter/devtools/issues/2324#issuecomment-690128227 for more info.

This just ignores these isolates (it turns out they're not the ones we care about) and seems to resolve the flakiness:

Screenshot 2020-09-10 at 11 59 57

If it turns out the VM service is incorrect, we can revert this change once it's fixed (and filters through to the channel we're testing against).

@jacob314

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,10 +22,6 @@ matrix:
     - env: BOT=main CHANNEL=master
     - env: BOT=main CHANNEL=beta
     - env: BOT=main CHANNEL=stable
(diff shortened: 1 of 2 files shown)


jacob314 (commented):
lgtm

PR #2333, opened 2020-09-11
Add search to the Network profiler

kenzieschmoll:
network-search

### DIFF
--- a/packages/devtools_app/lib/src/http/http_request_data.dart
+++ b/packages/devtools_app/lib/src/http/http_request_data.dart
@@ -136,7 +136,10 @@ class HttpRequestData extends NetworkRequest {
   /// around the issue by displaying them as "in-progress". It would be
   /// reasonable to display them as "unknown start time" but that seems like
   /// more complexity than it is worth.
-  bool get isValid => _startEvent != null;
+  // TODO(kenz): figure out how to handle HTTP body events in the network
+  // profiler. For now, mark them as invalid.
+  bool get isValid =>
+      _startEvent != null && !_startEvent.name.contains('HTTP CLIENT response');
 
   /// True if either the request or response contained cookies.
   bool get hasCookies =>
--- a/packages/devtools_app/lib/src/http/http_request_data.dart
+++ b/packages/devtools_app/lib/src/http/http_request_data.dart
@@ -136,8 +136,9 @@ class HttpRequestData extends NetworkRequest {
   /// around the issue by displaying them as "in-progress". It would be
   /// reasonable to display them as "unknown start time" but that seems like
   /// more complexity than it is worth.
-  // TODO(kenz): figure out how to handle HTTP body events in the network
-  // profiler. For now, mark them as invalid.
+  // TODO(kenz): https://github.com/flutter/devtools/issues/2335 - figure out
+  // how to handle HTTP body events in the network profiler. For now, mark them
+  // as invalid.
   bool get isValid =>
       _startEvent != null && !_startEvent.name.contains('HTTP CLIENT response');
 
--- a/packages/devtools_app/lib/src/network/network_screen.dart
+++ b/packages/devtools_app/lib/src/network/network_screen.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/rendering.dart';
 import 'package:provider/provider.dart';
@@ -15,10 +16,13 @@ import '../split.dart';
 import '../table.dart';
 import '../table_data.dart';
 import '../theme.dart';
+import '../ui/search.dart';
 import '../utils.dart';
 import 'network_controller.dart';
 import 'network_model.dart';
 import 'network_request_inspector.dart';
+
+final networkSearchFieldKey = GlobalKey(debugLabel: 'NetworkSearchFieldKey');
 
 class NetworkScreen extends Screen {
   const NetworkScreen()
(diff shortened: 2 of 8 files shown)


packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -136,7 +136,10 @@ class HttpRequestData extends NetworkRequest {
   /// around the issue by displaying them as "in-progress". It would be
   /// reasonable to display them as "unknown start time" but that seems like
   /// more complexity than it is worth.
-  bool get isValid => _startEvent != null;
+  // TODO(kenz): figure out how to handle HTTP body events in the network
+  // profiler. For now, mark them as invalid.
+  bool get isValid =>
+      _startEvent != null && !_startEvent.name.contains('HTTP CLIENT response');
```
jacob314, line 142: Can you file a bug for this?

packages/devtools_app/lib/src/network/network_screen.dart
```diff
@@ -161,6 +165,18 @@ class _NetworkScreenBodyState extends State<NetworkScreenBody>
             _networkController.clear();
           },
         ),
+        const Expanded(child: SizedBox()),
```
jacob314, line 168: can you change the properties on the row instead of adding this degenerate expanded?

packages/devtools_app/lib/src/table.dart
```diff
@@ -673,6 +689,34 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
     sortColumn = widget.sortColumn;
     sortDirection = widget.sortDirection;
     scrollController = ScrollController();
+
+    if (widget.activeSearchMatchNotifier != null) {
+      widget.activeSearchMatchNotifier
+          .addListener(activeSearchListener = _onActiveSearchChange);
+
+      // TODO(kenz): why isn't the listener firing when added via
```
jacob314, line 697: that should be equivalent. Maybe you needed a hot restart for the change to have an effect.
This sort of change should be duplicated in didUpdateWidget as well as initState or there will be issues. Seems like a lot of the logic in the table class has that same issue.

packages/devtools_app/lib/src/table.dart
```diff
@@ -957,6 +1014,36 @@ class _TableRowState<T> extends State<TableRow<T>>
...
+      searchMatches = widget.searchMatchesNotifier.value;
+      addAutoDisposeListener(widget.searchMatchesNotifier, () {
+        final isPreviousMatch = searchMatches.contains(widget.node);
+        final isNewMatch =
+            widget.searchMatchesNotifier.value.contains(widget.node);
+        searchMatches = widget.searchMatchesNotifier.value;
+
+        // We only want to rebuild the row if it the match status has changed.
+        if (isPreviousMatch ^ isNewMatch) {
+          setState(() {});
```
jacob314, line 1028: why not use `!=` instead of `^`?

also can we store whether there is a match in the state object to avoid the empty setState and make it clear what the state we are tracking is?

packages/devtools_app/lib/src/table.dart
```diff
@@ -957,6 +1014,36 @@ class _TableRowState<T> extends State<TableRow<T>>
...
+      });
+    }
+
+    if (widget.activeSearchMatchNotifier != null) {
+      activeSearchMatch = widget.activeSearchMatchNotifier.value;
+      addAutoDisposeListener(widget.activeSearchMatchNotifier, () {
+        final isPreviousActiveSearchMatch = activeSearchMatch == widget.node;
+        final isNewActiveSearchMatch =
+            widget.activeSearchMatchNotifier.value == widget.node;
+        activeSearchMatch = widget.activeSearchMatchNotifier.value;
```
jacob314, line 1039: move this duplicated code into a helper.

packages/devtools_app/lib/src/table.dart
```diff
@@ -719,6 +711,13 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
     }
   }
 
+  void _initSearchListener() {
+    addAutoDisposeListener(
```
jacob314, line 715: nit: why not move the check that activeSearchMatchNotifier is not null into this method?

packages/devtools_app/lib/src/table.dart
```diff
@@ -740,6 +739,10 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
         }
       });
     });
+
+    if (widget.activeSearchMatchNotifier != null) {
```
jacob314, line 743: redundant check for the notifier being not null.

packages/devtools_app/lib/src/table.dart
```diff
@@ -740,6 +739,10 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
         }
       });
     });
+
```
jacob314, line 742: why isn't the selection notifier listener also tracked in the helper called from initState and didUpdateWidget?

jacob314:
lgtm

packages/devtools_app/lib/src/table.dart
kenzieschmoll: done

packages/devtools_app/lib/src/table.dart
kenzieschmoll: removed now that the check is inside _initSearchListener

packages/devtools_app/lib/src/table.dart
kenzieschmoll: I didn't modify anything about the selection listener because it's not in the scope of this PR - will add a note to address

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/2316

PR #2340, opened 2020-09-15
Add filtering to the Network profiler

kenzieschmoll:
Screen Shot 2020-09-15 at 8 02 21 AM

Screen Shot 2020-09-15 at 12 01 57 PM

If a filter is applied, the filter button is toggled:
Screen Shot 2020-09-15 at 8 02 49 AM

At the bottom of the screen, the filter count "x of y" is shown:
Screen Shot 2020-09-15 at 8 49 06 AM

Fixes https://github.com/flutter/devtools/issues/2316

### DIFF
--- a/packages/devtools_app/lib/src/common_widgets.dart
+++ b/packages/devtools_app/lib/src/common_widgets.dart
@@ -575,11 +575,11 @@ class FilterButton extends StatelessWidget {
   }
 }
 
-Widget clearTextFieldButton(VoidCallback onPressed) {
-  return textFieldSuffixButton(Icons.clear, onPressed);
+Widget clearInputButton(VoidCallback onPressed) {
+  return inputDecorationSuffixButton(Icons.clear, onPressed);
 }
 
-Widget textFieldSuffixButton(IconData icon, VoidCallback onPressed) {
+Widget inputDecorationSuffixButton(IconData icon, VoidCallback onPressed) {
   return Container(
     padding: const EdgeInsets.symmetric(horizontal: densePadding),
     width: 24.0,
--- a/packages/devtools_app/lib/src/common_widgets.dart
+++ b/packages/devtools_app/lib/src/common_widgets.dart
@@ -537,6 +537,62 @@ class ExportButton extends StatelessWidget {
   }
 }
 
+class FilterButton extends StatelessWidget {
+  const FilterButton({
+    Key key,
+    @required this.onPressed,
+    @required this.isFilterActive,
+  }) : super(key: key);
+
+  final VoidCallback onPressed;
+
+  final bool isFilterActive;
+
+  @override
+  Widget build(BuildContext context) {
+    final colorScheme = Theme.of(context).colorScheme;
+    return RoundedOutlinedBorder(
+      child: SizedBox(
+        height: defaultButtonHeight,
+        child: Tooltip(
+          message: 'Filter',
+          child: FlatButton(
+            key: key,
+            onPressed: onPressed,
+            color: isFilterActive
+                ? colorScheme.toggleButtonBackgroundColor
+                : Colors.transparent,
+            child: createIcon(
+              Icons.filter_list,
+              color: isFilterActive
+                  ? colorScheme.toggleButtonForegroundColor
+                  : null,
+            ),
+          ),
+        ),
+      ),
+    );
+  }
+}
+
+Widget clearTextFieldButton(VoidCallback onPressed) {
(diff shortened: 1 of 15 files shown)


jacob314:
High level: it could make sense to support a query field where users can type a query like
`uri:foo.com status:403` as well as or instead of providing a field for each possible way of filtering and searching. That could apply to all devtools pages doing search.

packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -537,6 +537,62 @@ class ExportButton extends StatelessWidget {
...
+      ),
+    );
+  }
+}
+
+Widget clearTextFieldButton(VoidCallback onPressed) {
+  return textFieldSuffixButton(Icons.clear, onPressed);
+}
+
+Widget textFieldSuffixButton(IconData icon, VoidCallback onPressed) {
```
jacob314, line 582: why is this a "suffix" button?

packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -537,6 +537,62 @@ class ExportButton extends StatelessWidget {
...
+  return Container(
+    padding: const EdgeInsets.symmetric(horizontal: densePadding),
+    width: 24.0,
+    child: IconButton(
+      padding: const EdgeInsets.all(0.0),
```
jacob314, line 587: nit: would placing the padding in the icon button look different than in the container?

packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -268,12 +281,83 @@ class NetworkController with SearchControllerMixin<NetworkRequest> {
...
+  NetworkFilter({
+    this.method,
+    this.uriSubstring,
+    this.status,
+    this.type,
+    this.showHttp = true,
+    this.showWebSocket = true,
+  });
+
+  static NetworkFilter from(NetworkFilter filter) {
```
jacob314, line 341: nit: make a factory constructor even though that won't change anything.

packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -268,12 +281,83 @@ class NetworkController with SearchControllerMixin<NetworkRequest> {
...
+      method: filter.method,
+      uriSubstring: filter.uriSubstring,
+      status: filter.status,
+      type: filter.type,
+      showHttp: filter.showHttp,
+      showWebSocket: filter.showWebSocket,
+    );
+  }
+
+  String method;
```
jacob314, line 352: can these all be final?

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/common_widgets.dart
kenzieschmoll: It's for the `suffix` parameter of an InputDecoration. Maybe I should rename the method to inputDecorationSuffixButton

packages/devtools_app/lib/src/network/network_controller.dart
kenzieschmoll: done

packages/devtools_app/lib/src/network/network_controller.dart
kenzieschmoll: no because they need to be modified when adjusting the filter in the dialog. We could make a `copyWith` method. Leaving as is for now since I'll likely tweak all this code in a follow up to use a query.

packages/devtools_app/lib/src/common_widgets.dart
kenzieschmoll: the explicit 0 padding is to override the default padding of 4.0 in the IconButton, otherwise the size constraint of 24.0 is not respected.

jacob314:
lgtm

kenzieschmoll:
Updated to use query-based filter as discussed. PTAL

packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -351,11 +351,11 @@ class NetworkFilter {
       if (querySeparatorIndex != -1) {
         final value = part.substring(querySeparatorIndex + 1);
         if (value != '') {
```
jacob314, line 353: it could be nice to have an enum like
QueryFliter { final Iterable<string> keys; }
class so that you can reuse this parsing logic for all pages rather than writing it again on each page.

jacob314 (approved):
lgtm

PR #2344, opened 2020-09-17
Display call graph and dominator tree for code size Diffs

kenzieschmoll:
We generate the call graph for both the old and new snapshot. When a node is selected in the diff tree map:
- we start by looking up the node in the new call graph data
- if we cannot find the node there, we lookup the node in the old call graph
- if we still cannot find the matching node, we default to the root of the new call graph.

Screen Shot 2020-09-17 at 11 01 04 AM
Fixes https://github.com/flutter/devtools/issues/2343

### DIFF
--- a/packages/devtools_app/lib/src/charts/treemap.dart
+++ b/packages/devtools_app/lib/src/charts/treemap.dart
@@ -647,7 +647,16 @@ class TreemapNode extends TreeNode<TreemapNode> {
         // the Dart AOT snapshot.
         return reversedPath.reversed.toList();
       }
-      if (current.name.contains('package:')) {
+
+      // Prevent duplicated package names from being added to the path.
+      // Sometimes the hierarchy can look like this:
+      // package:a
+      //     package: a
+      //     <Type>
+      //     <OneByteString>
+      if ((current.name.contains('package:') ||
+              current.name.contains('dart:')) &&
+          (reversedPath.isEmpty || reversedPath.last != current.name)) {
         reversedPath.add(current.name);
       }
       current = current.parent;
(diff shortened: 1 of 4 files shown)


kenzieschmoll:
cc @cobblest

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -647,7 +647,16 @@ class TreemapNode extends TreeNode<TreemapNode> {
...
       }
-      if (current.name.contains('package:')) {
+
+      // Prevent duplicated package names from being added to the path.
+      // Sometimes the hierarchy can look like this:
+      // package:a
+      //     package: a
+      //     <Type>
+      //     <OneByteString>
+      if ((current.name.contains('package:') ||
```
jacob314, line 657: not a new issue but can we use startsWIth instead of contains to avoid false positives?

packages/devtools_app/lib/src/code_size/code_size_screen.dart
```diff
@@ -396,7 +396,19 @@ class _DiffViewState extends State<DiffView> with AutoDisposeMixin {
               axis: Axis.vertical,
               children: [
                 _buildTreemap(),
-                CodeSizeDiffTable(rootNode: diffRoot),
+                Row(
+                  children: [
+                    Flexible(
```
jacob314, line 401: Why Flexible instead of expanded? In general I think Expanded is a safer default as you won't end up with UI that doesn't fit the full width of the row.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/code_size/code_size_screen.dart
kenzieschmoll: Because each item in the row needs to have a flex value of one

packages/devtools_app/lib/src/charts/treemap.dart
kenzieschmoll: done

PR #2365, opened 2020-09-21
Rename all uses of "code size" to "app size".

kenzieschmoll:
The "code size screen" is actually better described as "app size screen". It supports analyzing the size of your total application - not just the code. 

No functional changes - just renames all uses of "code size" to "app size" where appropriate.


jacob314 (approved):
lgtm

PR #2366, opened 2020-09-21
various updates to the devtools command line UI

devoncarew:
various updates to the devtools command line UI:
- rename some options for clarity
- provide sample values for the help text
- handle issues parsing cli options and print the usage info
- hide some options, and make some others just display for the `-v` mode
- allow `devtools <uri>` to launch and connect to an app
- change to launch a browser by default
- remove a few short option names (the single char versions)

```
Dart DevTools version 0.9.2

usage: devtools <options> [service protocol uri]

Open a DevTools instance in a browser and optionally connect to an existing application.

-h, --help                            Prints help output.
-v, --verbose                         Output more informational messages.
    --host=<host>                     Hostname to serve DevTools on (defaults to localhost).
    --port=<port>                     Port to serve DevTools on; specify 0 to automatically use any available port.
                                      (defaults to "9100")
    --[no-]launch-browser             Launches DevTools in a browser immediately at start.
                                      (defaults to on)
    --machine                         Sets output format to JSON for consumption in tools.
    --record-memory-profile=<file>    Start devtools headlessly and write memory profiling samples to the indicated file.
                                      (defaults to "memory_samples.json")
```


devoncarew:
@DanTup for FYI - some options lost their short option version; I assume this won't impact VS Code.

packages/devtools_server/lib/src/server.dart
```diff
@@ -174,10 +131,9 @@ Future<HttpServer> serveDevToolsWithArgs(
 /// Serves DevTools.
 ///
 /// `handler` is the [shelf.Handler] that the server will use for all requests.
-/// If null, [defaultHandler] will be used.
-/// Defaults to null.
+/// If null, [defaultHandler] will be used. Defaults to null.
+// Note: this method is used by the Flutter CLI.
```
jacob314, line 135: and dwds
https://github.com/dart-lang/webdev/blob/c625f2f9accfac87142cec8bf87c633c42be00b5/dwds/lib/src/servers/devtools.dart#L27

jacob314 (approved):
lgtm

DanTup:
> some options lost their short option version; I assume this won't impact VS Code.

VS Code only uses the full names for all args so this won't affect it.

packages/devtools_server/lib/src/server.dart
```diff
@@ -458,6 +423,119 @@ Future<HttpServer> serveDevTools({
...
+          'available port.',
+    )
+    ..addFlag(
+      argLaunchBrowser,
+      defaultsTo: true,
+      help: 'Launches DevTools in a browser immediately at start.',
+    )
+// TODO: Remove this - prefer that clients use the rest arg.
+    ..addOption(
+      argVmUri,
```
devoncarew, line 462: @terrylucas - where is this option currently used? Somewhere on the flutter engine bots?

packages/devtools_server/lib/src/server.dart
devoncarew: 👍 updated the comment

devoncarew:
The travis failures look like I'm now referencing dart:ffi code (from the shared plugin) in something that we're trying to compile for the web.

devoncarew:
This is building now -

DanTup:
> change to launch a browser by default

@devoncarew it seems like this may trying to launch Chrome when started by VS Code too (and in the case where you don't have Chrome installed, it causes the server to throw and exit). I can add `--no-browser-launch` in VS Code but not sure if IntelliJ would need this too - would it be better to suppress here if running in `--machine` mode or something?

devoncarew:
> > change to launch a browser by default
> 
> @devoncarew it seems like this may trying to launch Chrome when started by VS Code too (and in the case where you don't have Chrome installed, it causes the server to throw and exit). I can add `--no-browser-launch` in VS Code but not sure if IntelliJ would need this too - would it be better to suppress here if running in `--machine` mode or something?

Yes, I'd only intended this to take effect when not running in machine mode. You're seeing this even in machine mode?

DanTup:
> You're seeing this even in machine mode?

Seems like it - see https://github.com/flutter/devtools/issues/2405. I'll change it to default based on the `--machine` flag (and may also prevent it terminating the server if it doesn't find Chrome, even when run from CLI - that doesn't seem desirable). Thanks!

Issue #1034, opened 2019-09-10
Please add a search function to the Performance View


escamoteur:
Hi,

it would be awesome if you could add a search function to locate a specific function on the charts of the different performance views.

Cheers
Thomas

kenzieschmoll:
Thanks for the feature request! Tracking issue for search/filtering tables here: https://github.com/flutter/devtools/issues/785. We'll keep this issue open as well since searching tables won't apply to the CPU flame chart.

kenzieschmoll:
The flame chart supports search. Supporting search in the tree tables (call tree and bottom up) is tracked here: https://github.com/flutter/devtools/issues/2524

PR #2368, opened 2020-09-22
 Implement search for performance page

kenzieschmoll:
Screen Shot 2020-09-22 at 8 14 13 AM

Fixes part of https://github.com/flutter/devtools/issues/1034

### DIFF
--- a/packages/devtools_app/lib/src/charts/flame_chart.dart
+++ b/packages/devtools_app/lib/src/charts/flame_chart.dart
@@ -210,6 +210,25 @@ abstract class FlameChartState<T extends FlameChart, V> extends State<T>
           rows[index].nodes.isEmpty ? sectionSpacing : rowHeightWithPadding,
       computeLength: () => rows.length,
     );
+
+    if (widget.activeSearchMatchNotifier != null) {
+      addAutoDisposeListener(widget.activeSearchMatchNotifier, () async {
+        final activeSearch = widget.activeSearchMatchNotifier.value;
+        if (activeSearch == null) return;
+
+        // Ensure the [activeSearch] is vertically in view.
+        if (!isDataVerticallyInView(activeSearch)) {
+          await scrollVerticallyToData(activeSearch);
+        }
+
+        // TODO(kenz): zoom if the event is less than some min width.
+
+        // Ensure the [activeSearch] is horizontally in view.
+        if (!isDataHorizontallyInView(activeSearch)) {
+          await scrollHorizontallyToData(activeSearch);
+        }
+      });
+    }
   }
 
   @override
@@ -431,6 +450,29 @@ abstract class FlameChartState<T extends FlameChart, V> extends State<T>
       );
     }
   }
+
+  Future<void> scrollVerticallyToData(V data) async {
+    await verticalScrollController.animateTo(
+      // Subtract [2 * rowHeightWithPadding] to give the target scroll event top padding.
+      topYForData(data) - 2 * rowHeightWithPadding,
+      duration: shortDuration,
+      curve: defaultCurve,
+    );
+  }
+
+  Future<void> scrollHorizontallyToData(V data) async {
+    final offset =
+        startXForData(data) + widget.startInset - widget.containerWidth * 0.1;
+    await scrollToX(offset);
+  }
+
+  bool isDataVerticallyInView(V data);
+
+  bool isDataHorizontallyInView(V data);
+
+  double topYForData(V data);
+
+  double startXForData(V data);
 }
 
 class ScrollingFlameChartRow<V> extends StatefulWidget {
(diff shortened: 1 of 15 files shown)


packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -431,6 +450,29 @@ abstract class FlameChartState<T extends FlameChart, V> extends State<T>
       );
     }
   }
+
+  Future<void> scrollVerticallyToData(V data) async {
+    await verticalScrollController.animateTo(
+      // Subtract [2 * rowHeightWithPadding] to give the target scroll event top padding.
+      topYForData(data) - 2 * rowHeightWithPadding,
```
jacob314, line 457: you should take the min of this and zero to avoid a bounce when scrolling to the top. We're also scrolling too far when the item is close to the bottom.

packages/devtools_app/lib/src/performance/performance_screen.dart
```diff
@@ -25,6 +25,9 @@ import '../theme.dart';
 import '../ui/vm_flag_widgets.dart';
 import 'performance_controller.dart';
 
+final performanceSearchFieldKey =
```
jacob314, line 28: why is this global key needed?

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
```diff
@@ -87,6 +89,20 @@ class CpuProfilerController {
     }
   }
 
+  @override
+  List<CpuStackFrame> matchesForSearch(String search) {
+    if (search == null || search.isEmpty) return [];
```
jacob314, line 94: nit:
search?.isEmpty ?? false

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
```diff
@@ -87,6 +89,20 @@ class CpuProfilerController {
...
+    final matches = <CpuStackFrame>[];
+    final caseInsensitiveSearch = search.toLowerCase();
+    final currentStackFrames = _dataNotifier.value.stackFrames.values;
+    for (final frame in currentStackFrames) {
+      if (frame.name.toLowerCase().contains(caseInsensitiveSearch)) {
```
jacob314, line 99: we're adding a lot of toLowerCase calls here and elsewhere. Maybe add caseInsentiveMatch helper to the utils library. At some point it could make sense to do the match a character at a time to avoid the toLowerCase() and that would make it easy to optimize in the future.

packages/devtools_app/lib/src/profiler/cpu_profiler.dart
```diff
@@ -87,6 +97,21 @@ class _CpuProfilerState extends State<CpuProfiler>
               controller: _tabController,
               tabs: CpuProfiler.tabs,
             ),
+            // TODO(kenz): support search for call tree and bottom up tabs as
+            // well. This will require implementing search for tree tables.
+            if (currentTab.key == CpuProfiler.flameChartTab &&
```
jacob314, line 102: conditioning UI based on whether a key is present is a bit surprising. Maybe just pass a boolean indicating whether to include a search field and define a SearchField widget so you can match for it in the tests if that is the only reason you have a key.

packages/devtools_app/lib/src/timeline/timeline_flame_chart.dart
```diff
@@ -222,46 +222,40 @@ class TimelineFlameChartState
...
         eventTopY + rowHeightWithPadding <
             verticalScrollOffset + widget.containerHeight;
   }
 
-  bool _isEventHorizontallyInView(TimelineEvent event) {
-    return visibleTimeRange.contains(event.time.start) &&
-        visibleTimeRange.contains(event.time.end);
+  @override
+  bool isDataHorizontallyInView(TimelineEvent data) {
+    return visibleTimeRange.contains(data.time.start) &&
```
jacob314, line 237: Not new code but I think there is a bug.
This will return false if `data` is a range that starts before the visibleTimeRange and end after the visibleTimeRange. Elsewhere you have the proper range intersection calculation of checking the a.start <= b.end && a.end >= a.start

jacob314 (commented):
Done with a first round of comments.

packages/devtools_app/lib/src/charts/flame_chart.dart
jacob314: hmm. I'm seeing some bounce animations (top of the scrollable flashes blue if using mtv not cupertino style) due to similar bugs in the inspector.

packages/devtools_app/lib/src/performance/performance_screen.dart
jacob314: ok. maybe add a TODO. there is probably an easier way to connect them together.

kenzieschmoll: done

packages/devtools_app/lib/src/charts/flame_chart.dart
kenzieschmoll: added a clamp operation

packages/devtools_app/lib/src/timeline/timeline_flame_chart.dart
kenzieschmoll: fixed

jacob314:
lgtm

PR #2371, opened 2020-09-23
Raster cache

terrylucas:
Collect RasterCache estimates from Flutter engine.
ADB returns all sizes in kilobytes, adjust all ADB sizes to total bytes.


packages/devtools_app/lib/src/memory/memory_protocol.dart
```diff
@@ -117,6 +120,10 @@ class MemoryTracker {
       adbMemoryInfo = AdbMemoryInfo.empty();
     }
 
+    // Query the engine's rasterCache estimate.
+    rasterCache = await _fetchRasterCacheInfo();
+    logger.log('$rasterCache');
```
jacob314, line 125: probably avoid logging this.

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -322,7 +318,39 @@ class ServiceConnectionManager {
...
+    return flutterView['id'];
+  }
+
+  /// Flutter engine returns estimate how much memory is used by layer/picture raster
+  /// cache entries in bytes.
+  ///
+  /// Call to returns JSON payload 'EstimateRasterCacheMemory' with two entries:
+  ///   layerBytes - layer raster cache entries in bytes
+  ///   pictureBytes - picture raster cache entries in bytes
+  Future<Response> getRasterCacheMetrics() async {
```
jacob314, line 330: we should avoid method names that start with get. This one can be `rasterCacheMetrics()` I don't know what to do for `getDisplayRefreshRate` as `DisplayRefreshRate` makes it sound like we are displaying the refresh rate. Maybe `queryDisplayRefreshRate` or similar? Avoiding `get` is a dart style guide suggestion we didn't follow that well as some of this code was ported from Java.

packages/devtools_app/lib/src/service_registrations.dart
```diff
@@ -61,3 +61,7 @@ const flutterMemory = RegisteredServiceDescription._(
 const flutterListViews = '_flutter.listViews';
 
 const displayRefreshRate = '_flutter.getDisplayRefreshRate';
+
+/// Flutter engine returns estimate how much memory is used by layer/picture raster
+/// cache entries in bytes.
+const flutterEngineRasterCache = '_flutter.estimateRasterCacheMemory';
```
jacob314, line 67: nit: maybe keep the word estimate in the name to make it clear that this data may be a bit innacurate.

packages/devtools_server/lib/src/memory_profile.dart
```diff
@@ -184,6 +186,10 @@ class MemoryProfile {
       adbMemoryInfo = AdbMemoryInfo.empty();
     }
 
+    // Query the engine's rasterCache estimate.
+    rasterCache = await _fetchRasterCacheInfo(_selectedIsolate);
+    print('$rasterCache');
```
jacob314, line 191: remove print before commit.

packages/devtools_server/lib/src/memory_profile.dart
```diff
@@ -194,8 +200,64 @@ class MemoryProfile {
...
+    // Each isolate should only have one FlutterView.
+    final flutterView = views.firstWhere(
+      (view) => view['type'] == 'FlutterView',
+      orElse: () => null,
+    );
+
+    if (flutterView == null) {
+      final message =
+          'No Flutter Views to query: ${flutterViewListResponse.json}';
+      print('ERROR: $message');
```
jacob314, line 237: remove print. Also what happens when this is run on a dartvm app? Does this code path get triggered?

packages/devtools_shared/lib/src/memory/adb_memory_info.dart
```diff
@@ -17,17 +17,41 @@ class AdbMemoryInfo {
...
-        json[javaHeapKey] as int,
-        json[nativeHeapKey] as int,
-        json[codeKey] as int,
-        json[stackKey] as int,
-        json[graphicsKey] as int,
-        json[otherKey] as int,
-        json[systemKey] as int,
-        json[totalKey] as int,
-      );
+  /// @param rawData (from service extension) if true then all data returned from
```
jacob314, line 20: In Dart rather than using `@param rawData` write the comments, use `[rawData]`

Instead of having to pass in `rawData`, consider instead turning these all into private fields with public getters that multiply by 1024. That way you wouldn't have to worry about any possible bugs passing in the wrong value for rawData. if you leave rawData, consider changing the name to make it clearer what the effect is. For example `inKilobytes` or similar.

packages/devtools_shared/lib/src/memory/event_sample.dart
```diff
@@ -163,3 +165,35 @@ class EventSample {
...
+
+  static RasterCache parse(Map<String, dynamic> json) =>
+      json == null ? null : RasterCache.fromJson(json);
+
+  int layerBytes;
+  
+  int pictureBytes;
+
+  Map<String, dynamic> toJson() {
+    final json = <String, dynamic>{};
```
jacob314, line 188: why use addAll instead of just returning a map literal directly?

jacob314 (commented):
Done with a first round of comments.

packages/devtools_app/lib/src/memory/memory_protocol.dart
terrylucas: I've added a TODO to remove logging, added for Yuqian to help track issues values not changing.

packages/devtools_app/lib/src/service_manager.dart
terrylucas: There are a few methods that are getters besides these, I'll change all of them in this file e.g., getFlutterVersion, getAdbMemoryInfo, getFlutterViewId.

packages/devtools_app/lib/src/service_registrations.dart
terrylucas: done

packages/devtools_server/lib/src/memory_profile.dart
terrylucas: Done

packages/devtools_server/lib/src/memory_profile.dart
terrylucas: Done.

packages/devtools_shared/lib/src/memory/adb_memory_info.dart
terrylucas: I think this might be better.  Added two constructors for fromJsonInKB and fromJson.

```
  /// All data inside of AdbMemoryInfo is in total bytes. When receiving ADB data
  /// from the service extension (directly from ADB) then the data is in kilobytes.
  /// See the factory constructor fromJsonInKB.
  factory AdbMemoryInfo.fromJson(Map<String, dynamic> json) 

  /// Use when converting data received from the service extension, directly from
  /// ADB. All data received from ADB dumpsys meminfo is in kilobytes must adjust to
  /// total bytes for AdbMemoryInfo data.
  factory AdbMemoryInfo.fromJsonInKB(
    Map<String, dynamic> json,
  ) 

```

packages/devtools_shared/lib/src/memory/event_sample.dart
terrylucas: Should

  Map<String, dynamic> toJson() => {
        'layerBytes': layerBytes,
        'pictureBytes': pictureBytes,
      };

Issue #2335, opened 2020-09-14
Handle HTTP body events in the network and timeline pages


kenzieschmoll:
HTTP body timeline events added in https://github.com/dart-lang/sdk/commit/7d7c88389056b4284018517f07dc900745f58417

maks:
@kenzieschmoll does this mean that the network profiler will get the ability to display request/response details like body contents, headers etc, the way say Chrome devtools does?

kenzieschmoll:
The network profiler already displays request/response headers, but we are working on adding the ability to display the body as well.

kenzieschmoll:
@bkonyi The HTTP body events aren't showing up in the timeline as expected. The trace events are missing the "parentId" argument that is supposed to link the response to the request.

```
{"name":"HTTP CLIENT response of GET","cat":"Dart","tid":15367,"pid":20492,"ts":213468906122,"ph":"b","id":"1","args":{"requestUri":"https://jsonplaceholder.typicode.com/albums/1","statusCode":200,"reasonPhrase":"OK","filterKey":"HTTP/client","isolateId":"isolates/2072390848141751","isolateGroupId":"isolateGroups/616754501538234698"}}
```

Perhaps something is missing in the implementation https://github.com/dart-lang/sdk/commit/7d7c88389056b4284018517f07dc900745f58417#diff-5014705a5d4e165fd43436f8d6015a21

bkonyi:
Yup, looks like it's missing [here](https://github.com/dart-lang/sdk/commit/7d7c88389056b4284018517f07dc900745f58417#diff-9a93a60cb56c68dbbe9ecc961fc5cad9R1197). I'll get a fix out today.

kenzieschmoll:
related: https://github.com/flutter/devtools/issues/1952

SeasonLeee:
Hi, is there any plan of publish the dev tool with http body in the near future? 

(I recommend Flutter as our new APP's main tech... I really don't want to disappoint my team...........)

kenzieschmoll:
The change to support this is here: https://github.com/flutter/devtools/pull/2374. Expecting to land today and publish this week with the latest bits.

softboy99:
Hi,
this seems isn't merged to latest flutter 1.22.1(which includes dart 2.10.1)

kenzieschmoll:
If you switch to the Flutter dev channel, you will be able to use this functionality.

softboy99:
after switch to flutter dev branch, also failed

[...]

PR #2374, opened 2020-09-24
Show HTTP response bodies in the Network profiler

kenzieschmoll:
Screen Shot 2020-09-24 at 12 28 09 PM


kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/2335

jacob314:
If the response body is text, do you see human readable text or a list of numbers?

kenzieschmoll:
Hmm on the few different endpoints I'm working with, it seems to be just numbers. @bkonyi is there some additional decoding that needs to be done?

bkonyi:
It probably needs to be decoded into UTF-8.

kenzieschmoll:
fixed and updated screenshot

packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -988,3 +989,20 @@ class _BreadcrumbPainter extends CustomPainter {
...
+
+  static const encoder = JsonEncoder.withIndent('  ');
+
+  final Map<String, dynamic> json;
+
+  @override
+  Widget build(BuildContext context) {
+    final formattedArgs = encoder.convert(json);
+    return Text(
+      formattedArgs.replaceAll('"', ''),
```
jacob314, line 1004: I would leave in the quotes. Otherwise there is a risk that json like
"Then they said \"hello\"."
will like 
"Then they said \hello\." which is a bit confusing.
You could add a TODO to use a prettier format like YAML. Even though JSON is a bit ugly it is nice in that you can copy and paste it into Dart and it is generally a valid literal.

packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -988,3 +989,20 @@ class _BreadcrumbPainter extends CustomPainter {
...
+      style: const TextStyle(fontFamily: 'RobotoMono'),
```
jacob314, line 1005: don't we already reference the monospace font somewhere in our theme?

packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -100,28 +101,34 @@ class NetworkController with SearchControllerMixin<NetworkRequest> {
...
 
     // Build our list of network requests from the collected events.
     for (final request in httpEvents.entries) {
       final requestId = request.key;
+
+      // Do not handle response events - they are handled as part of the request
+      if (httpRequestIdToResponseId.values.contains(requestId)) continue;
+
+      final responseId = httpRequestIdToResponseId[requestId];
+      final responseEvents = <Map<String, dynamic>>[];
```
jacob314, line 123: prefer Map<String, Object> and add the required casts. that will catch more bugs relative to using dynamic.

packages/devtools_app/lib/src/network/network_request_inspector_views.dart
```diff
@@ -136,6 +136,22 @@ class HttpRequestHeadersView extends StatelessWidget {
...
+  const HttpResponseView(this.data);
+
+  final HttpRequestData data;
+
+  @override
+  Widget build(BuildContext context) {
+    return SingleChildScrollView(
+      child: Text(
+        data.responseBody,
+        style: const TextStyle(fontFamily: 'RobotoMono'),
```
jacob314, line 149: also move this monospace font to a theme.

packages/devtools_app/lib/src/common_widgets.dart
kenzieschmoll: we have `fixedFontStyle`. Used that.

packages/devtools_app/lib/src/network/network_controller.dart
kenzieschmoll: changed to object. What required casts were you referring to? I didn't see any errors when just changing from dynamic to Object

packages/devtools_app/lib/src/network/network_request_inspector_views.dart
kenzieschmoll: used `fixedFontStyle` from theme.dart

packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -999,10 +999,11 @@ class FormattedJson extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    // TODO(kenz): we could consider using a prettier format like YAML
```
jacob314, line 1002: nit: end commends with period.

jacob314 (approved):
lgtm

PR #2375, opened 2020-09-24
Cleanup custom painters to paint based on listening to scroll controllers rather than depending on setState

jacob314:
We were calling setState on the overall FlameChart widget every time the scroll changed which is expensive and defeats some of the lazy logic in the sliver views rendering the more heavyweight content part of the view.
If scroll offsets ever changed due to layout as happens if the offset goes off the edge of the scroll view due to content shrinking, then the current approach would cause rendering artifacts as well as slightly hurting performance as one offset would be used for the list view and another offset would be used by the custom painters.

### DIFF
--- a/packages/devtools_app/lib/src/charts/flame_chart.dart
+++ b/packages/devtools_app/lib/src/charts/flame_chart.dart
@@ -104,11 +104,11 @@ abstract class FlameChartState<T extends FlameChart, V> extends State<T>
 
   double mouseHoverX;
 
-  ScrollController verticalScrollController;
+  ScrollController verticalController;
 
   FixedExtentDelegate verticalExtentDelegate;
 
-  LinkedScrollControllerGroup linkedHorizontalScrollControllerGroup;
+  LinkedScrollControllerGroup horizontalController;
 
   double get maxScrollOffset =>
       widget.containerWidth * (zoomController.value - 1);
@@ -139,7 +139,7 @@ abstract class FlameChartState<T extends FlameChart, V> extends State<T>
       contentWidthWithZoom + widget.startInset + widget.endInset;
 
   TimeRange get visibleTimeRange {
-    final horizontalScrollOffset = linkedHorizontalScrollControllerGroup.offset;
+    final horizontalScrollOffset = horizontalController.offset;
     final startMicros = horizontalScrollOffset < widget.startInset
         ? startTimeOffset
         : startTimeOffset +
@@ -180,8 +180,8 @@ abstract class FlameChartState<T extends FlameChart, V> extends State<T>
     super.initState();
     initFlameChartElements();
 
-    linkedHorizontalScrollControllerGroup = LinkedScrollControllerGroup();
-    verticalScrollController = ScrollController();
+    horizontalController = LinkedScrollControllerGroup();
+    verticalController = ScrollController();
 
     zoomController = AnimationController(
       value: FlameChart.minZoomLevel,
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -1115,37 +1102,48 @@ abstract class FlameChartPainter extends CustomPainter {
...
 
   final double zoom;
 
   final BoxConstraints constraints;
 
-  final double verticalScrollOffset;
+  final ScrollController verticalScroll;
 
-  final double horizontalScrollOffset;
+  final LinkedScrollControllerGroup horizontalScroll;
```
kenzieschmoll, line 1122: nit: let's name these with controller and controllerGroup at the end to make it clear that these are not offsets, or in the case of horizontalScroll, not even a scroll controller.

packages/devtools_app/lib/src/timeline/timeline_flame_chart.dart
```diff
@@ -704,9 +707,9 @@ class SectionLabelPainter extends FlameChartPainter {
   }
 
   @override
-  bool shouldRepaint(SectionLabelPainter oldDelegate) {
-    return verticalScrollOffset != oldDelegate.verticalScrollOffset ||
-        eventGroups != oldDelegate.eventGroups ||
+  bool shouldRepaint(CustomPainter oldDelegate) {
+    return oldDelegate is SectionLabelPainter &&
+            eventGroups != oldDelegate.eventGroups ||
```
kenzieschmoll, line 712: formatting?

packages/devtools_app/lib/src/timeline/timeline_flame_chart.dart
```diff
@@ -834,15 +842,15 @@ class AsyncGuidelinePainter extends FlameChartPainter {
   // TODO(kenz): does this have to return true all the time? Is it cheaper to
   // compare delegates or to just paint?
   @override
-  bool shouldRepaint(AsyncGuidelinePainter oldDelegate) => true;
+  bool shouldRepaint(CustomPainter oldDelegate) => true;
```
kenzieschmoll, line 845: why did this type change?

packages/devtools_app/lib/src/timeline/timeline_flame_chart.dart
```diff
@@ -995,8 +1006,8 @@ class SelectedFrameBracketPainter extends FlameChartPainter {
     this.selectedFrame, {
     @required double zoom,
     @required BoxConstraints constraints,
-    @required double verticalScrollOffset,
-    @required double horizontalScrollOffset,
+    @required ScrollController verticalScroll,
+    @required LinkedScrollControllerGroup horizontalScroll,
```
kenzieschmoll, line 1010: nit: naming. to keep this consistent with other parts of the code, can we add controller and controllerGroup to the names, respectively

packages/devtools_app/lib/src/timeline/timeline_flame_chart.dart
jacob314: You should avoid overriding a dart method with an implementation with a more specific type. Not sure why we don't have a lint to prevent it. This would throw an exception in a debug build if shouldRepaint was called with a CustomPainter of a different type.

packages/devtools_app/lib/src/charts/flame_chart.dart
jacob314: In general you should avoid including types in member names as the language is typed so there is no need to specify it. However,  I do agree that clarifying those are controllers is a good idea.
how about `horizontalController` and `verticalController` so that we avoid overly verbose Java style variable names. Some prior art from flutter is the ScrollController for ListView is just called `controller` not `scrollController` and the `child` property is `child` not `childWidget`.

packages/devtools_app/lib/src/timeline/timeline_flame_chart.dart
jacob314: switched to an if as this logic was actually wrong.

kenzieschmoll:
lgtm_orangatang

kenzieschmoll (approved):

PR #2385, opened 2020-09-29
adjust some text on the landing screen

devoncarew:
- adjust some text on the landing screen

@kenzieschmoll - is the new text correct? Which kinds of files can we load?


kenzieschmoll (approved):
This is still correct. An instruction_sizes file is a subset of a v8 snapshot, so while we support loading both, it is better if a user loads a v8 snapshot. And a v8 snapshot is a subset of an app size analysis file (generated via the --analyze-size flag), so ideally we want to point Flutter users to create the size analysis file. We should link to [documentation](https://flutter.dev/docs/development/tools/devtools/app-size#generating-size-files) in these instructions.

Though thinking big picture - the app size tool may eventually support dart2js and it currently supports snapshots for general Dart applications, so we may want to be careful about being Flutter specific

Issue #2327, opened 2020-09-09
Improve usability of debug file tree


grouma:
The debug file tree can require a significant number of clicks to select a single search result. This happens when the single search result is nested deeply in a folder structure. There are a couple of ways to resolve this issue:

- Show both the tree structure and a small number of flattened search results
- Auto expand folders when there is a single child

kenzieschmoll:
We could also auto expand the path to the matching node

grouma:
That could also work depending on the number of search results.

crtl:
I dont know who have thought it would be a good idea but its astonishing how worse dev tools are getting with each update.
Before you had a simple list with all your project files not grouped in any way.
"WTF" I thought but I worked my way into it.
Now the file tree is organized into a dropdown tree structure with the labels not being clickable but only the arrows which are by the way maybe 4px wide.
Are there any considerations about UX when it comes to developing new features?

devoncarew:
> I dont know who have thought it would be a good idea but its astonishing how worse dev tools are getting with each update.

@crtl, this feedback is not constructive and doesn't meet Flutter's code of conduct for our community (https://github.com/flutter/flutter/blob/master/CODE_OF_CONDUCT.md). We're happy to take constructive feedback on our tools.

grouma:
This was mostly addressed in https://github.com/flutter/devtools/pull/2386. However, the labels are still not clickable. It's somewhat tricky to click the small arrows making for a poor user experience.

grouma:
cc @jacob314

PR #2386, opened 2020-09-29
simplify the debugger's libraries view

devoncarew:
- simplify the debugger's libraries view
- fix https://github.com/flutter/devtools/issues/2327

@grouma - this flattens the Libraries view contents to one level (the top package level and the a flat list of libraries contained in that package). Is something like this what you were thinking of?

Screen Shot 2020-09-29 at 10 59 33 AM


grouma:
This should go a long way in terms of usability and reducing the number of clicks required to open a file.

jacob314 (approved):
lgtm

PR #2392, opened 2020-09-30
Bump vm_service to 5.2.0 and migrate to new httpTimelineLogging apis

kenzieschmoll:
This new API was added in https://github.com/dart-lang/sdk/commit/e85e5b6264edf554c13d328a0af9c38672c4a0c3. This change was made in a backwards compatible manner.


terrylucas (approved):
lgtm, thanks.

packages/devtools_app/lib/src/vm_service_wrapper.dart
```diff
@@ -432,25 +432,38 @@ class VmServiceWrapper implements VmService {
...
-  }
-
-  Future<HttpTimelineLoggingState> getHttpEnableTimelineLogging(
-    String isolateId,
-  ) async {
-    assert(await isHttpTimelineLoggingAvailable(isolateId));
-    return _trackFuture('getHttpEnableTimelineLogging',
-        _vmService.getHttpEnableTimelineLogging(isolateId));
+    if (await isProtocolVersionSupported(
+        supportedVersion: SemanticVersion(major: 5, minor: 2))) {
```
bkonyi, line 436: I don't think this is the right check. You want to check that the [service protocol extension version](https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service_extension.md) is `>= 1.2`, but it looks like you're comparing against the package revision instead.

bkonyi (changes requested):

bkonyi (approved):

packages/devtools_app/lib/src/vm_service_wrapper.dart
```diff
@@ -798,16 +812,42 @@ class VmServiceWrapper implements VmService {
...
   }
 
+  Future<bool> isDartIoVersionSupported({
+    @required SemanticVersion supportedVersion,
+    @required String isolateId,
+  }) async {
+    // We must call [_getDartIOVersion] instead of [getDartIoVersion] here.
+    // Otherwise, we get a NoSuchMethodError on `_call` due to mismatched
+    // arguments.
+    _dartIoVersion ??= await _getDartIOVersion(isolateId);
```
kenzieschmoll, line 844: @munificent @leafpetersen This behavior was surprising. CC @bkonyi 

In this file, VmServiceWrapper implements VmService. [dart_io_extensions.dart](https://github.com/dart-lang/sdk/blob/master/pkg/vm_service/lib/src/dart_io_extensions.dart#L18) declares extension methods DartIOExtension on VmService.

I originally attempted to create a method `VmServiceWrapper.getDartIOVersion` that overrode the extension method `DartIOExtension.getDartIOVersion` - but hit a lint telling me I'm not overriding an inherited method.

I then tried to eliminate the `@override` annotation, and call `VmServiceWrapper.getDartIOVersion` from this line. This is he surprising part. I expected `getDartIOVersion` to call my local method named `getDartIOVersion`, but instead, the call went to `DartIOExtension.getDartIOVersion`. Is this expected behavior?

munificent: Yes, it's expected behavior. Instance methods—either inherited or directly on the class—always win over extensions. Basically any language has to make a choice here and there are trade-offs on either side. Dart follows C# and, I think, other languages in prioritizing instance methods first. The rationale is that if someone with access to the actual class or superclasses put the method there, it's probably the one users of the class are most likely to expect to be calling. Consider:

```dart
import 'foo.dart';

main() {
  var list = [1, 2, 3];
  list.map((c) => c + 1);
}
```

Here, you'd probably be pretty surprised to discover that `map` is not Iterable.map but is actually some extension defined inside 'foo.dart'. We felt prioritizing instance methods was the least surprising choice.

kenzieschmoll: Is `VmServiceWrapper.getDartIOVersion` not the instance method in this case? I think I'm in agreement with you that it is more surprising to call the extension method instead of the instance method, but if I'm not misunderstanding what "instance method" means, that isn't what I observed.

My case was as such:
```
class VmServiceWrapper implements VmService {
  void foo() => print('VmServiceWrapper.foo');

  void someOtherMethod() {
    foo();
  }
}

extension DartIOExtension on VmService {
  void foo() => print('DartIOExtension.foo');
}
```

From inside `someOtherMethod`, the call to `foo()` prints "DartIOExtension.foo" when I would have expected it to print "VmServiceWrapper.foo".

leafpetersen: The behavior you describe is not what I would expect.  Sounds like a bug, do you have a repro?  I can't reproduce it with the code above using the command line VM (after adding a trivial VmService class).

munificent: > Is `VmServiceWrapper.getDartIOVersion` not the instance method in this case?

Oh, sorry, I read your comment wrong. Yeah, your expectation that the instance method wins is the correct one. Could there be something else going on here? Maybe a typo in the instance method name or something like that?

kenzieschmoll: Hmm.. maybe there was something else going on because I can no longer reproduce this. I tried to make a simple example and couldn't repro there either. So false alarm, I guess 👍

Issue #1968, opened 2020-05-15
Network profiler not working after hot restart


ad-on-is:
I have an app that does a network request on startup, without users interaction.

1. Run app in debug mode and open Dart DevTools
2. Go to Network and press Record
3. Hot-Restart app
4. App executes a network-request on startup
5. Network profiler does not show the executed http request

but...

1. Run app in debug mode and open Dart DevTools
2. Go to Network and press Record
3. Press button in app to execute a network-request
4. Network profiler shows the http request

<details><summary>flutter doctor -v</summary>

```
[✓] Flutter (Channel stable, v1.17.0, on Microsoft Windows [Version 10.0.19628.1], locale en-AT)
    • Flutter version 1.17.0 at D:\Projects\Development\Flutter\flutter-sdk-win
    • Framework revision e6b34c2b5c (13 days ago), 2020-05-02 11:39:18 -0700
    • Engine revision 540786dd51
    • Dart version 2.8.1


[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
    • Android SDK at C:\Users\add\AppData\Local\Android\sdk
    • Platform android-29, build-tools 29.0.3
    • Java binary at: C:\Program Files\Android\Android Studio\jre\bin\java
    • Java version OpenJDK Runtime Environment (build 1.8.0_212-release-1586-b04)
    • All Android licenses accepted.

[✓] Android Studio (version 3.6)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin version 45.1.1
    • Dart plugin version 192.8052
    • Java version OpenJDK Runtime Environment (build 1.8.0_212-release-1586-b04)

[✓] VS Code (version 1.45.0)
    • VS Code at C:\Users\add\AppData\Local\Programs\Microsoft VS Code
    • Flutter extension version 3.10.1

[✓] Connected device (1 available)
    • ONEPLUS A6013 • 2739990e • android-arm64 • Android 10 (API 29)
```

</details>

kenzieschmoll:
@ad-on-is if you open the timeline page and click "Refresh" do you see HTTP events from the network request on startup?

@bkonyi IIRC, the network page only pulls traffic from the time the recording is started and beyond - does that sound correct? Is there any reason we shouldn't be recording traffic by default and pulling all network data that is available on the first request (from there after we could only pull data for the most recent polling interval)?

bkonyi:
[...]

Issue #337, opened 2019-02-22
Hot Restart test appears to be causing increased flakes


jacob314:
All tests using hot restart need to be commented out until the underlying source of the flakes can be isolated as a flaky test is a broken test.

jacob314:
The hot restart tests have been commented out for now. Need to figure out how to reduce the # of flakes after adding them back.

kenzieschmoll:
@DanTup is this resolved?

DanTup:
I don't think so. Before I got to this one, I had a flake in another test on a PR. (debug inspector controller) I ran it in a big loop for hours with logging and it never failed (!) and I didn't get back to it. I planned on looking at this one after that.

Maybe I should look at this one first though, sounds like it might be more likely to happen. I'll give have a go at running it in the background today.

DanTup:
Ok, while running these tests again, I hit a flake. It was a timeout in tearDown on this line:

```
await _service.allFuturesCompleted.future;
```

And in the logs:

```
[ERROR:flutter/shell/common/shell.cc(178)] Dart Error: Unhandled exception:
'dart:isolate/runtime/lib/timer_impl.dart': Failed assertion: line 433: '<optimized out>': is not true.
#0      _AssertionError._doThrowNew (dart:core/runtime/lib/errors_patch.dart:40:39)
#1      _AssertionError._throwNew (dart:core/runtime/lib/errors_patch.dart:36:5)
#2      _Timer._cancelWakeup (dart:isolate/runtime/lib/timer_impl.dart:433:12)
#3      _Timer._notifyEventHandler (dart:isolate/runtime/lib/timer_impl.dart:299:7)
#4      _Timer._handleMessage (dart:isolate/runtime/lib/timer_impl.dart:419:5)
#5      _RawReceivePortImpl._handleMessage (dart:isolate/runtime/lib/isolate_patch.dart:171:12)
```

This is our old friend https://github.com/dart-lang/sdk/issues/33838. I think when that error occurs, some callback isn't happening which is leaving an open future.

The test does contain this:

```
      /// After the hot restart some existing calls to the vm service may
      /// timeout and that is ok.
      serviceManager.service.doNotWaitForPendingFuturesBeforeExit();
      print('HRST #6');
```

However that's right before the hot restart. My guess is that we have additional futures being registered afterwards that are failing. I'll see if it's easy to log where each future came from so we can at least tell which one isn't finishing.

[...]

PR #2395, opened 2020-09-30
Restore httpEnableTimelineLogging state after hot restart

kenzieschmoll:
This partially addresses https://github.com/flutter/devtools/issues/1968. To completely fix the issue, we need a socket rpc change like https://github.com/dart-lang/sdk/commit/e85e5b6264edf554c13d328a0af9c38672c4a0c3 @bkonyi

### DIFF
--- a/packages/devtools_app/lib/src/service_manager.dart
+++ b/packages/devtools_app/lib/src/service_manager.dart
@@ -586,28 +586,27 @@ class ServiceExtensionManager {
         await _onFrameEventReceived();
         break;
       case 'Flutter.ServiceExtensionStateChanged':
-        final String name = event.json['extensionData']['extension'].toString();
-        final String valueString =
-            event.json['extensionData']['value'].toString();
-        await _updateServiceExtensionForStateChange(name, valueString);
+        final name = event.json['extensionData']['extension'].toString();
+        final encodedValue = event.json['extensionData']['value'].toString();
+        await _updateServiceExtensionForStateChange(name, encodedValue);
         break;
       case 'HttpTimelineLoggingStateChange':
         final name = extensions.httpEnableTimelineLogging.extension;
-        final valueString = event.json['extensionData']['enabled'].toString();
-        await _updateServiceExtensionForStateChange(name, valueString);
+        final encodedValue = event.json['extensionData']['enabled'].toString();
+        await _updateServiceExtensionForStateChange(name, encodedValue);
     }
   }
 
   Future<void> _updateServiceExtensionForStateChange(
     String name,
-    String valueString,
+    String encodedValue,
   ) async {
     final extension = extensions.serviceExtensionsAllowlist[name];
     if (extension != null) {
-      final dynamic value = _getExtensionValue(name, valueString);
+      final dynamic extensionValue = _getExtensionValue(name, encodedValue);
       final enabled =
           extension is extensions.ToggleableServiceExtensionDescription
-              ? value == extension.enabledValue
+              ? extensionValue == extension.enabledValue
               // For extensions that have more than two states
               // (enabled / disabled), we will always consider them to be
               // enabled with the current value.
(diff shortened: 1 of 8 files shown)


packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -587,42 +587,52 @@ class ServiceExtensionManager {
         break;
       case 'Flutter.ServiceExtensionStateChanged':
         final String name = event.json['extensionData']['extension'].toString();
-        final String valueFromJson =
+        final String valueString =
```
jacob314, line 590: nit: just call this value not valueString. No need to include types in variable names.
Also consider removing the type of the LHS as the type is already inferred by having the toString.

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -587,42 +587,52 @@ class ServiceExtensionManager {
...
             event.json['extensionData']['value'].toString();
+        await _updateServiceExtensionForStateChange(name, valueString);
+        break;
+      case 'HttpTimelineLoggingStateChange':
+        final name = extensions.httpEnableTimelineLogging.extension;
+        final valueString = event.json['extensionData']['enabled'].toString();
+        await _updateServiceExtensionForStateChange(name, valueString);
```
jacob314, line 597: here and elsewhere. valueString -> value

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -587,42 +587,52 @@ class ServiceExtensionManager {
...
+        name,
+        enabled,
+        value,
+        callExtension: false,
+      );
     }
   }
 
-  dynamic _getExtensionValueFromJson(String name, String valueFromJson) {
+  dynamic _getExtensionValue(String name, String valueString) {
```
jacob314, line 625: maybe valueString -> encodedValue here and elsewhere if `value` is confusing.

packages/devtools_app/lib/src/ui/service_extension_widgets.dart
```diff
@@ -421,17 +421,19 @@ mixin _ServiceExtensionMixin<T extends _ServiceExtensionWidget> on State<T> {
...
       await action();
 
-      if (widget.completedText != null) {
+      if (mounted && widget.completedText != null) {
         Notifications.of(context).push(widget.completedText);
       }
     } catch (e, st) {
       log('$e\n$st');
 
       Notifications.of(context).push(widget.describeError(e));
```
jacob314, line 430: shouldn't we also short circuit displaying the notification if the widget is not mounted?

jacob314 (commented):
Can we test this?

packages/devtools_app/lib/src/service_manager.dart
kenzieschmoll: done

packages/devtools_app/lib/src/service_manager.dart
kenzieschmoll: done

packages/devtools_app/lib/src/service_manager.dart
kenzieschmoll: done

packages/devtools_app/lib/src/ui/service_extension_widgets.dart
kenzieschmoll: done

kenzieschmoll:
the [hot restart tests](https://github.com/flutter/devtools/blob/709fa720442af8d219dee391729c5bc4b6ea41a8/packages/devtools_testing/lib/service_manager_test.dart#L207) are currently commented out due to https://github.com/flutter/devtools/issues/337. I could take a look at seeing if they can be re-enabled.

jacob314:
lgtm

kenzieschmoll:
Landing this with a TODO to address the hot restart tests

PR #2399, opened 2020-10-02
Updated devtools_shared dep missing bumped to 0.9.3+1

terrylucas:


terrylucas:
Would like to update all pubspec.yaml files to

`version: 0.9.3+1
`

packages/devtools_server/pubspec.lock
```diff
@@ -243,4 +243,4 @@ packages:
     source: hosted
     version: "0.7.3"
 sdks:
-  dart: ">=2.10.0-110 <=2.11.0-177.0.dev"
+  dart: ">=2.10.0-110 <=2.11.0-161.0.dev"
```
devoncarew, line 246: Not for this PR, but I'm not sure why we so tightly control which SDKs the server can install into.

devoncarew (approved):

packages/devtools_server/pubspec.lock
terrylucas: Me either I was surprised to see this.

devoncarew: Ah! This is the lock file - I thought it was the pubspec. The pubspec sdk constraint is what will matter to users - that one looks good.

kenzieschmoll (commented):
did you use the update_version.sh script to do this? I don't see the change to the devtools.dart file

terrylucas:
> did you use the update_version.sh script to do this? I don't see the change to the devtools.dart file

Not yet wanted to make sure that the +1 was how we wanted to rev.

kenzieschmoll:
+1 sgtm

terrylucas:
Everything is updated.

kenzieschmoll (approved):

PR #2406, opened 2020-10-04
Don't try to start Chrome in --machine mode unless --launch-browser is supplied

### LINKED ISSUE #1936: update_version.sh should handle versions with "+1"
Example:
Current version = 0.2.4+1
Ran script `update_version.sh 0.2.5`

**Expected**
New version = 0.2.5

**Actual**
New version = 0.2.5+1

### LINKED ISSUE #2405: DevTools tries to find Chrome at startup when using --machine option
Running the following
```
/home/kingwill101/snap/flutter/common/flutter/bin/cache/dart-sdk/bin/pub global run devtools --machine --try-ports 10 --allow-embedding
```
Outputs 

```
Unhandled exception:
ProcessException: No such file or directory
  Command: google-chrome http://127.0.0.1:9101
```
Dart version 
```
Dart SDK version: 2.11.0-181.0.dev (dev) (Thu Oct 1 04:39:42 2020 -0700) on "linux_x64"
```
@DanTup

DanTup:
\+ don't crash server if Chrome can't be launched.

Fixes #2405.
Fixes #1936.

@devoncarew @jacob314 

I had to tweak the update_version script to make it work properly (it seemed to not handle the `+` so I ended up with the existing `+3` tagged onto the end of the new version supplied. Also unsure if the change should've gone in the server changelog or devtools (lmk if I should move it to server).

### DIFF
--- a/packages/devtools_server/lib/src/server.dart
+++ b/packages/devtools_server/lib/src/server.dart
@@ -74,7 +74,9 @@ Future<void> _serveDevToolsWithArgs(
 }) async {
   final help = args[argHelp];
   final bool machineMode = args[argMachine];
-  final bool launchBrowser = args[argLaunchBrowser];
+  // launchBrowser defaults based on machine-mode if not explicitly supplied.
+  final bool launchBrowser =
+      args.wasParsed(argLaunchBrowser) ? args[argLaunchBrowser] : !machineMode;
   final bool enableNotifications = args[argEnableNotifications];
   final bool allowEmbedding = args[argAllowEmbedding];
   final port = args[argPort] != null ? int.tryParse(args[argPort]) ?? 0 : 0;
@@ -210,7 +212,11 @@ Future<HttpServer> serveDevTools({
       }).toString();
     }
 
-    await Chrome.start([url]);
+    try {
+      await Chrome.start([url]);
+    } catch (e) {
+      print('Unable to launch Chrome: $e\n');
+    }
   }
 
   if (enableStdinCommands) {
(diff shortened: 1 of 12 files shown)


packages/devtools_server/lib/src/server.dart
```diff
@@ -74,7 +74,9 @@ Future<void> _serveDevToolsWithArgs(
 }) async {
   final help = args[argHelp];
   final bool machineMode = args[argMachine];
-  final bool launchBrowser = args[argLaunchBrowser];
+  // launchBrowser defaults based on machine-mode if not explicitly supplied.
+  final bool launchBrowser =
```
devoncarew, line 78: 👍

devoncarew (approved):
lgtm, and thanks for the fix!

I'm less familiar with the update_version.sh changes; rslgtm there.

DanTup:
Thanks - I'll wait for @jacob314's verdict on the script changes before merging in case I've misunderstood (I'm also not familiar with it, and having to change it felt strange since DevTools releases often have a `+` in them - I may have been using it wrong!).

I tested in VS Code and I do see the Chrome window open (which is confusing, since it's just on the main connect screen) so it might be nice to do a release with this change in once merged (as far as I can tell, there have been no other changes since the last release so it would be just this).

kenzieschmoll:
This also fixes https://github.com/flutter/devtools/issues/1936

DanTup:
Aha, in that case I'm confident in the change to the script. I wasn't sure if I was doing something wrong, but apparently it wasn't just me. Thanks!

PR #2407, opened 2020-10-05
Switch to new routing/pages APIs

### LINKED ISSUE #1725: Permalinks don't work on the flutter:web version of the app
You should be able to refresh the page and stay on the same page in devtools.

DanTup:
This switches away from `onGenerateRoute` and friends and instead returns `Pages` and uses `RouterDelegate`+`RouteInformationParser` to control the pages we display (and persist them in the address bar).

All tests are passing, though I had two make some changes (see my comments below) that may/may not be correct.

I've manually tested all the bits I can think of (including things like disconnecting/reconnecting), though I'm not confident I haven't missed something (for example I'm less familiar with the snapshot pages etc.).

Fixes #1725.

### DIFF
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -133,11 +133,14 @@ class DevToolsAppState extends State<DevToolsApp> {
   }
 
   Widget _buildTabbedPage(
-      BuildContext context, String page, Map<String, String> params) {
-    final appServiceUrl = params['uri'];
+    BuildContext context,
+    String page,
+    Map<String, String> params,
+  ) {
+    final vmServiceUri = params['uri'];
 
     // Always return the landing screen if there's no VM service URI.
-    if (appServiceUrl?.isEmpty ?? true) {
+    if (vmServiceUri?.isEmpty ?? true) {
       return DevToolsScaffold.withChild(
         key: const Key('landing'),
         child: LandingScreenBody(),
@@ -158,7 +161,7 @@ class DevToolsAppState extends State<DevToolsApp> {
     }
     final embed = params['embed'] == 'true';
     return Initializer(
-      url: appServiceUrl,
+      url: vmServiceUri,
       allowConnectionScreenOnDisconnect: !embed,
       builder: (_) {
         final tabs = embed && page != null
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -34,6 +34,7 @@ import 'notifications.dart';
 import 'performance/performance_controller.dart';
 import 'performance/performance_screen.dart';
 import 'preferences.dart';
+import 'routing.dart';
 import 'scaffold.dart';
 import 'screen.dart';
 import 'snapshot_screen.dart';
@@ -42,10 +43,6 @@ import 'timeline/timeline_controller.dart';
 import 'timeline/timeline_screen.dart';
 import 'ui/service_extension_widgets.dart';
 import 'utils.dart';
-
-const homeRoute = '/';
-const snapshotRoute = '/snapshot';
-const appSizeRoute = '/app-size';
 
 // Disabled until VM developer mode functionality is added.
 const showVmDeveloperMode = false;
@@ -102,111 +99,122 @@ class DevToolsAppState extends State<DevToolsApp> {
     _clearCachedRoutes();
   }
 
-  /// Generates routes, separating the path from URL query parameters.
-  Route _generateRoute(RouteSettings settings) {
-    final uri = Uri.parse(settings.name);
-    final path = uri.path.isEmpty ? homeRoute : uri.path;
-    final args = settings.arguments;
-
+  /// Gets the page for a given page/path and args.
+  Page _getPage(BuildContext context, String page, Map<String, String> args) {
     // Provide the appropriate page route.
-    if (routes.containsKey(path)) {
-      WidgetBuilder builder = (context) => routes[path](
-            context,
-            uri.queryParameters,
-            args,
-          );
+    if (pages.containsKey(page)) {
+      Widget widget = pages[page](
+        context,
+        page,
+        args,
+      );
       assert(() {
-        builder = (context) => _AlternateCheckedModeBanner(
-              builder: (context) => routes[path](
-                context,
-                uri.queryParameters,
-                args,
-              ),
-            );
+        widget = _AlternateCheckedModeBanner(
+          builder: (context) => pages[page](
+            context,
+            page,
+            args,
+          ),
+        );
         return true;
       }());
-      return MaterialPageRoute(settings: settings, builder: builder);
+      return MaterialPage(child: widget);
     }
 
     // Return a page not found.
-    return MaterialPageRoute(
-      settings: settings,
-      builder: (BuildContext context) {
-        return DevToolsScaffold.withChild(
-          child: CenteredMessage("'$uri' not found."),
-          ideTheme: ideTheme,
-          analyticsProvider: widget.analyticsProvider,
+    return MaterialPage(
+      child: DevToolsScaffold.withChild(
+        key: const Key('not-found'),
+        child: CenteredMessage("'$page' not found."),
+        ideTheme: ideTheme,
+        analyticsProvider: widget.analyticsProvider,
+      ),
+    );
+  }
+
+  Widget _buildTabbedPage(
+      BuildContext context, String page, Map<String, String> params) {
+    final appServiceUrl = params['uri'];
+
+    // Always return the landing screen if there's no VM service URI.
+    if (appServiceUrl?.isEmpty ?? true) {
+      return DevToolsScaffold.withChild(
+        key: const Key('landing'),
+        child: LandingScreenBody(),
+        ideTheme: ideTheme,
+        analyticsProvider: widget.analyticsProvider,
+        actions: [
+          OpenSettingsAction(),
+          OpenAboutAction(),
+        ],
+      );
+    }
+
+    // TODO(dantup): We should be able simplify this a little, removing params['page']
+    // and only supporting /inspector (etc.) instead of also &page=inspector if
+    // all IDEs switch over to those URLs.
+    if ((page?.length ?? 0) == 0) {
+      page = params['page'];
+    }
+    final embed = params['embed'] == 'true';
+    return Initializer(
+      url: appServiceUrl,
+      allowConnectionScreenOnDisconnect: !embed,
+      builder: (_) {
+        final tabs = embed && page != null
+            ? _visibleScreens().where((p) => p.screenId == page).toList()
+            : _visibleScreens();
+        if (tabs.isEmpty) {
+          return DevToolsScaffold.withChild(
+            child: CenteredMessage(
+                'The "$page" screen is not available for this application.'),
+            ideTheme: ideTheme,
+            analyticsProvider: widget.analyticsProvider,
+          );
+        }
+        return _providedControllers(
+          child: DevToolsScaffold(
+            embed: embed,
+            ideTheme: ideTheme,
+            page: page,
+            tabs: tabs,
+            analyticsProvider: widget.analyticsProvider,
+            actions: [
+              // TODO(https://github.com/flutter/devtools/issues/1941)
+              if (serviceManager.connectedApp.isFlutterAppNow) ...[
+                HotReloadButton(),
+                HotRestartButton(),
+              ],
+              OpenSettingsAction(),
+              OpenAboutAction(),
+            ],
+          ),
         );
       },
     );
   }
 
-  /// The routes that the app exposes.
-  Map<String, UrlParametersBuilder> get routes {
+  /// The pages that the app exposes.
+  Map<String, UrlParametersBuilder> get pages {
     return _routes ??= {
-      homeRoute: (_, params, __) {
-        if (params['uri']?.isNotEmpty ?? false) {
-          final embed = params['embed'] == 'true';
-          final page = params['page'];
-          return Initializer(
-            url: params['uri'],
-            allowConnectionScreenOnDisconnect: !embed,
-            builder: (_) {
-              final tabs = embed && page != null
-                  ? _visibleScreens().where((p) => p.screenId == page).toList()
-                  : _visibleScreens();
-              if (tabs.isEmpty) {
-                return DevToolsScaffold.withChild(
-                  child: CenteredMessage(
-                      'The "$page" screen is not available for this application.'),
-                  ideTheme: ideTheme,
-                  analyticsProvider: widget.analyticsProvider,
-                );
-              }
-              return _providedControllers(
-                child: DevToolsScaffold(
-                  embed: embed,
-                  ideTheme: ideTheme,
-                  initialPage: page,
-                  tabs: tabs,
-                  analyticsProvider: widget.analyticsProvider,
-                  actions: [
-                    // TODO(https://github.com/flutter/devtools/issues/1941)
-                    if (serviceManager.connectedApp.isFlutterAppNow) ...[
-                      HotReloadButton(),
-                      HotRestartButton(),
-                    ],
-                    OpenSettingsAction(),
-                    OpenAboutAction(),
-                  ],
-                ),
-              );
-            },
-          );
-        } else {
-          return DevToolsScaffold.withChild(
-            child: LandingScreenBody(),
-            ideTheme: ideTheme,
-            analyticsProvider: widget.analyticsProvider,
-            actions: [
-              OpenSettingsAction(),
-              OpenAboutAction(),
-            ],
-          );
-        }
-      },
-      snapshotRoute: (_, __, args) {
+      homePageId: _buildTabbedPage,
+      for (final screen in widget.screens)
+        screen.screen.screenId: _buildTabbedPage,
+      snapshotPageId: (_, __, args) {
+        final snapshotArgs = SnapshotArguments.fromArgs(args);
         return DevToolsScaffold.withChild(
+          key: const Key('snapshot'),
           analyticsProvider: widget.analyticsProvider,
           child: _providedControllers(
             offline: true,
-            child: SnapshotScreenBody(args, _screens),
+            child: SnapshotScreenBody(snapshotArgs, _screens),
           ),
           ideTheme: ideTheme,
         );
       },
-      appSizeRoute: (_, __, ___) {
+      appSizePageId: (_, __, ___) {
         return DevToolsScaffold.withChild(
+          key: const Key('appsize'),
           analyticsProvider: widget.analyticsProvider,
           child: _providedControllers(
             child: const AppSizeBody(),
@@ -247,12 +255,13 @@ class DevToolsAppState extends State<DevToolsApp> {
     return ValueListenableBuilder(
       valueListenable: widget.preferences.darkModeTheme,
       builder: (context, value, _) {
-        return MaterialApp(
+        return MaterialApp.router(
           title: 'Dart DevTools',
           debugShowCheckedModeBanner: false,
           theme: themeFor(isDarkTheme: value, ideTheme: ideTheme),
           builder: (context, child) => Notifications(child: child),
-          onGenerateRoute: _generateRoute,
+          routerDelegate: DevToolsRouterDelegate(_getPage),
+          routeInformationParser: DevToolsRouteInformationParser(),
         );
       },
     );
@@ -298,8 +307,8 @@ class DevToolsScreen<C> {
 /// args.
 typedef UrlParametersBuilder = Widget Function(
   BuildContext,
+  String,
   Map<String, String>,
-  SnapshotArguments args,
 );
 
 /// Displays the checked mode banner in the bottom end corner instead of the
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -206,7 +206,7 @@ class DevToolsAppState extends State<DevToolsApp> {
       snapshotPageId: (_, __, args) {
         final snapshotArgs = SnapshotArguments.fromArgs(args);
         return DevToolsScaffold.withChild(
-          key: const Key('snapshot'),
+          key: UniqueKey(),
           analyticsProvider: widget.analyticsProvider,
           child: _providedControllers(
             offline: true,
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -118,12 +118,12 @@ class DevToolsAppState extends State<DevToolsApp> {
         );
         return true;
       }());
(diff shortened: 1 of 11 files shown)


DanTup:
Looks like my changes are dependant on https://github.com/flutter/flutter/pull/66694 so currently fail on CI.

Temporarily updating the CI script to use `master` while working on the tests, but will revert before this is done.

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -479,7 +479,8 @@ class IsolateManager {
   }
 
   void _sendToMessageBus(Event event) {
-    messageBus.addEvent(BusEvent(
+    // TODO(dantup): Some tests don't set up a message bus - is ? here reasonable?
+    messageBus?.addEvent(BusEvent(
```
DanTup, line 483: Is this a reasonable fix? Some tests don't set up a MessageBus and don't have easy access to the `testwidgetsWithContext` function (and are not widget tests).

packages/devtools_app/test/split_test.dart
```diff
@@ -620,12 +620,14 @@ void main() {
...
-        expectEqualSizes(tester.element(find.byKey(_k3)).size,
-            const Size(182.5242718446602, 400));
+        // TODO(dantup): Why did these change? Which care correct? These match
+        // the original widths given a 400x400 SizedBox.
+        expectEqualSizes(
+            tester.element(find.byKey(_k1)).size, const Size(100.0, 400));
+        expectEqualSizes(
+            tester.element(find.byKey(_k2)).size, const Size(176.0, 400));
+        expectEqualSizes(
+            tester.element(find.byKey(_k3)).size, const Size(100.0, 400));
```
DanTup, line 630: Need to understand why these changed (to my eyes the new ones seem reasonable, but I'm not sure why before this change we had the others).

DanTup:
@jacob314 I think this is good to go (with some questions I've added comments/TODOs about). I've tested as much as I understand but there are some areas I'm less familiar with (snapshots, drag/drop, etc.) so any pointers on testing those (or anyone happy to test them before/after landing this) would be appreciated (cc @kenzieschmoll @terrylucas).

@helin24 are there any unreleased changes in Android Studio related to launching/embedding DevTools, or can I just test using the current release version?

helin24:
> @helin24 are there any unreleased changes in Android Studio related to launching/embedding DevTools, or can I just test using the current release version?

There have been some recent changes in the flutter-intellij repository (not yet released to dev channel) - these are setup changes for the embedded browser but should not affect devtools launching and after.

DanTup:
> these are setup changes for the embedded browser but should not affect devtools launching and after.

Great, thanks!

However - it looks like Android Studio activates DevTools from pub when it starts - is there a way to suppress this so I can test the local version (I activated it with `pub global activate --source path .`)? As an aside - I found activation to be pretty slow (I think checking dependency versions) so in VS Code we skip it unless the pub version is higher than the version installed.

kenzieschmoll:
What version of Flutter are you running against? I checked out your branch, and I am on the latest Flutter master. Looks like there are more API changes to `MaterialPage`?

packages/devtools_app/test/split_test.dart
DanTup: Jacob explained this - when the box was resized down to 200, the proportions changed (out of line with what was originally set), and when it was re-increased, it should have maintained this new proportions. This is somehow being lost after my change so I'll revert these and investigate why.

packages/devtools_app/test/split_test.dart
```diff
@@ -677,12 +679,14 @@ void main() {
...
+        // what I would expect (1+3 meet their minimums leaving 2 with what's left).
         expectEqualSizes(
-            tester.element(find.byKey(_k2)).size, const Size(0, 400));
-        expectEqualSizes(tester.element(find.byKey(_k3)).size,
-            const Size(517.3333333333333, 400));
+            tester.element(find.byKey(_k1)).size, const Size(200.0, 400));
+        expectEqualSizes(
+            tester.element(find.byKey(_k2)).size, const Size(176.0, 400));
+        expectEqualSizes(
+            tester.element(find.byKey(_k3)).size, const Size(400.0, 400));
```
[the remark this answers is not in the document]
DanTup: (as above)

kenzieschmoll:
Got it. Switched to dev to test. After a quick test of offline imports, it looks like that flow is broken. To reproduce drag-and-drop a DevTools export file into DevTools from any page (web). You can generate an export file from either the Timeline or Performance pages, just by interacting with a flutter app to load some data and then clicking the "Export" button.

DanTup:
@kenzieschmoll Ugh, I can repro - I'll fix - thanks!

(marking this back as draft, as the split tests are definitely broken and need fixing too).

DanTup:
Ok, I think this is good for a better look.

I've resolved the offline snapshot issue (a typo with the url params) and fixed an additional issue with the Exit button.

I've also resolved the split test issue - the `DevToolsRouterDelegate` was using a new `Key` on every instantiation which was resulting in widgets losing their state in tests (the `wrap` function would create a new one). We're now using a consistent key in tests (I'm not certain if this is the best way though).

If there's anything additional you think I may have overlooked and should test, please let me know!

packages/devtools_app/lib/src/app.dart
```diff
@@ -99,111 +96,122 @@ class DevToolsAppState extends State<DevToolsApp> {
...
-                    OpenAboutAction(),
-                  ],
-                ),
-              );
-            },
-          );
-        } else {
+  Widget _buildTabbedPage(
+      BuildContext context, String page, Map<String, String> params) {
+    final appServiceUrl = params['uri'];
```
kenzieschmoll, line 134: perhaps rename to `vmServiceUri` since that is what it is referred as throughout the code and documentation

packages/devtools_app/lib/src/routing.dart
```diff
@@ -0,0 +1,136 @@
...
+/// Converts between structured DevToolsRouteConfiguration (our internal data
+/// for pages/routing) and RouteInformation (generic data that can be persisted
+/// in the address bar/state objects).
+class DevToolsRouteInformationParser
+    extends RouteInformationParser<DevToolsRouteConfiguration> {
+  @override
+  Future<DevToolsRouteConfiguration> parseRouteInformation(
+      RouteInformation routeInformation) {
+    final uri = Uri.parse(routeInformation.location);
+    final path = uri.path.isNotEmpty ? uri.path.substring(1) : '';
```
kenzieschmoll, line 31: Can you add a comment that explains why we are taking the substring starting at index 1?

packages/devtools_app/lib/src/routing.dart
```diff
@@ -0,0 +1,136 @@
...
+    final configuration = DevToolsRouteConfiguration(path, uri.queryParameters);
+    return SynchronousFuture<DevToolsRouteConfiguration>(configuration);
```
kenzieschmoll, line 33: why SynchronousFuture and not just Future?

packages/devtools_app/lib/src/routing.dart
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  @override
+  RouteInformation restoreRouteInformation(
+      DevToolsRouteConfiguration configuration) {
```
kenzieschmoll, line 38: trailing comma

packages/devtools_app/lib/src/routing.dart
```diff
@@ -0,0 +1,136 @@
...
+    final path = '/${configuration.page ?? ''}';
+    final params =
```
kenzieschmoll, line 40: why not just `configuration.args.isNotEmpty ? configuration.args : null`

packages/devtools_app/lib/src/routing.dart
```diff
@@ -0,0 +1,136 @@
...
+        (configuration.args?.length ?? 0) != 0 ? configuration.args : null;
+    params?.removeWhere((key, value) => value == null);
+    return RouteInformation(
+        location: Uri(path: path, queryParameters: params).toString());
```
kenzieschmoll, line 44: trailing comma

packages/devtools_app/lib/src/routing.dart
```diff
@@ -0,0 +1,136 @@
...
+  DevToolsRouterDelegate(this._getPage, [GlobalKey<NavigatorState> key])
+      : navigatorKey = key ?? GlobalKey<NavigatorState>();
+
+  static DevToolsRouterDelegate of(BuildContext context) =>
+      Router.of(context).routerDelegate as DevToolsRouterDelegate;
+
+  @override
+  final GlobalKey<NavigatorState> navigatorKey;
+  final Page Function(BuildContext, String, Map<String, String>) _getPage;
+  final routes = ListQueue<DevToolsRouteConfiguration>();
```
kenzieschmoll, line 61: new line to separate these

packages/devtools_app/lib/src/routing.dart
```diff
@@ -0,0 +1,136 @@
...
+    final argsChanged = !mapEquals(
+      {...currentConfiguration.args, ...?updateArgs},
+      currentConfiguration.args,
+    );
+    if (!pageChanged && !argsChanged) {
+      return;
+    }
+
+    routes.add(DevToolsRouteConfiguration(
+        page, {...currentConfiguration.args, ...?updateArgs}));
```
kenzieschmoll, line 105: trailing comma

packages/devtools_app/lib/src/routing.dart
```diff
@@ -0,0 +1,136 @@
...
+
+  /// Updates arguments for the current page.
+  ///
+  /// Existing arguments (for example &uri=) will be preserved unless
+  /// overwritten by [updateArgs].
+  void updateArgsIfNotCurrent(Map<String, String> updateArgs) {
+    final argsChanged = !mapEquals(
+      {...currentConfiguration.args, ...?updateArgs},
+      currentConfiguration.args,
+    );
```
kenzieschmoll, line 124: let's put this in a helper since it is used in multiple places `_argsChanged({@required Map<String, String> current, @required Map<String, String> updated})`

packages/devtools_app/lib/src/scaffold.dart
```diff
@@ -260,11 +251,15 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
       // We want to avoid this because the routes underneath the existing
       // '/snapshot' route should remain unchanged while '/snapshot' sits on
       // top.
-      if (ModalRoute.of(context).settings.name == snapshotRoute) {
-        Navigator.popAndPushNamed(context, snapshotRoute, arguments: args);
+      final routerDelegate = DevToolsRouterDelegate.of(context);
+      if (routerDelegate.currentConfiguration.page == snapshotPageId) {
+        routerDelegate
+          ..popPage()
```
kenzieschmoll, line 257: is the pop page necessary here? one way to test would be to drag-and-drop to import a file multiple times in a row without clicking "Exit offline mode"? Then if you click "Exit offline mode", the router should return to the landing screen as expected.

packages/devtools_app/lib/src/service_manager.dart
kenzieschmoll: Can we mock this method instead? I think that would require us to make the method public

packages/devtools_app/lib/src/snapshot_screen.dart
```diff
@@ -69,7 +70,8 @@ class _SnapshotScreenBodyState extends State<SnapshotScreenBody> {
         Row(
           children: [
             ExitOfflineButton(onPressed: () {
-              Navigator.pop(context);
+              DevToolsRouterDelegate.of(context)
+                  .pushPageIfNotCurrent(homePageId, {'screen': null});
```
kenzieschmoll, line 74: With this change, is the snapshot screen still on the navigation stack? I'm not if sure we want a user to be able to go "back" to the offline snapshot they just cleared.

packages/devtools_app/lib/src/snapshot_screen.dart
```diff
@@ -98,5 +100,6 @@ class _SnapshotScreenBodyState extends State<SnapshotScreenBody> {
 
 class SnapshotArguments {
   SnapshotArguments(this.screenId);
+  SnapshotArguments.fromArgs(Map<String, String> args) : this(args['screen']);
```
kenzieschmoll, line 103: add new lines between all three of the lines in this class

packages/devtools_app/lib/src/routing.dart
DanTup: Done! (it's because `path` always starts with `/` but internally we now just use page IDs for our navigation).

packages/devtools_app/lib/src/routing.dart
DanTup: It avoids waiting for the next microtask (the docs for the base class recommend doing this where possible https://github.com/flutter/flutter/blob/d65e2c3a27ca566b2e276725588315830106fc0a/packages/flutter/lib/src/widgets/router.dart#L112-L116).

packages/devtools_app/lib/src/routing.dart
DanTup: Oops, forgot about `isEmpty`/`isNotEmpty`. Have simplified this (though it still needed a null check).

packages/devtools_app/lib/src/routing.dart
DanTup: Done - though since the first arg was always the current args (and calling it was just as verbose as the original code with named args), I made the helper just always compare against the current args.

packages/devtools_app/lib/src/scaffold.dart
DanTup: Based on the comment above, I believe it was to avoid adding more and more snapshot pages to the navigation stack when dragging on many.

I've changed this slightly, adding a "replacePage" method, as the previous one would call `notifyListeners()` twice (once after pop, once after push). The Exit button works correctly, however the back button isn't working right - still working on understanding why.

DanTup:
@kenzieschmoll 

> Can we mock this method instead? I think that would require us to make the method public

For some reason I can't reply to that inline comment (no comment box, but all other comments do?)!). We do have some other tests that set up their own MessageBus which might make more sense (for consistency), though they use `testWidgetsWithContext` that we don't have access to here. Would it be better to try and do something similar (for ex. a `testWithContext` that calls the same `setGlobal` - assuming we can do so from these tests).

packages/devtools_app/lib/src/snapshot_screen.dart
DanTup: I've changed this to the new `replacePage` as above (though it has the same back button issue that needs fixing).

DanTup:
I think I've addressed all of the comments, except for:

- MessageBus TODO
- Back button doesn't handle the pages that are removed from the stack (you can still "back" through them - for ex. snapshot views)

I'm also not sure if I'm seeing "Call stack exceeded" errors more frequently (vs just refreshing/making changes more while working on this), which might be a little concerning.

packages/devtools_app/lib/src/snapshot_screen.dart
```diff
@@ -70,10 +70,10 @@ class _SnapshotScreenBodyState extends State<SnapshotScreenBody> {
         Row(
           children: [
             ExitOfflineButton(onPressed: () {
-              DevToolsRouterDelegate.of(context)
-                  .pushPageIfNotCurrent(homePageId, {'screen': null});
-              reset();
               offlineMode = false;
+              reset();
+              DevToolsRouterDelegate.of(context)
+                  .replaceCurrent(homePageId, {'screen': null});
```
kenzieschmoll, line 76: trailing comma

kenzieschmoll:
> Back button doesn't handle the pages that are removed from the stack (you can still "back" through them - for ex. snapshot views)

Does this only occur when a snapshot has been dragged and dropped on top of a snapshot?

DanTup:
> Does this only occur when a snapshot has been dragged and dropped on top of a snapshot?

That and when you click Exit Offline mode. Anywhere calling the new `replacePage` method (which pops then pushes) doesn't seem to actually affect the Back behaviour. It may be that my understanding of how it's supposed to work is incorrect though (or that it's not complete on dev) - will see if it repros in my other tab sample and do some digging through the docs/comments and try on master.

kenzieschmoll:
I see. IMO this seems like a submit-blocker, or at least should be a quick follow up fix.

DanTup:
Yep, agreed. It repros on master and in my sample app, so I think it's definitely something I've done incorrectly. Definitely should fix before we merge this (because it may indicate I've misunderstood something important 😐).

DanTup:
Ok, I think I found the issue. It seems like the `RouterDelegate`s `popRoute()` method isn't being called on web - (though it is for Android). I've [reproduced it in my small sample](https://github.com/DanTup/tab_route_test/issues/2) and will try to debug there.

In the meantime - I'm wondering if the back button behaviour (when that works) is actually what we want. If a user goes:

Connect screen -> Timeline -> Memory -> Performance

And then clicks Back, do we expect them to go to Memory or back to Connect? In a short list like this, going back to Memory feels correct, however if you'd gone:

Connect screen -> Timeline -> Memory -> Performance -> Memory -> Performance -> Memory -> Performance -> Memory -> Performance -> Memory -> Performance -> Memory -> Performance -> Memory -> Performance

Then now bouncing between tabs as you click Back might feel a bit weird. I'm wondering whether switching tabs should actually replace the current page (while still updating the URL ofc)? (we could also just remove any tab from the earlier history if you revisit it, but that might seem even weirder).

Any opinions? (@devoncarew @jacob314 @kenzieschmoll)

kenzieschmoll:
I think going back to the immediately preceding page feels correct. I can't think of other web examples where the navigation stack is manually shortened. 

The only place I think this does not make sense is for offline imports. To the user that should feel like an operation within the same page, where only the data is changed.

DanTup:
This seems to all work now in `flutter run` mode, but I'm having an issue in release mode - sometimes clicking Back in the browser updates the URL but doesn't re-render anything on the screen. I don't know if its related, but I'm also seeing:

1. CanvasKit errors about fonts:

Screenshot 2020-10-14 at 11 06 45

2. Empty errors printed to the console:

Screenshot 2020-10-14 at 11 06 56

If I build a non-CanvasKit version then I don't seem to be able to reproduce the original issue, so it could be related. Since we're shipping the CanvasKit build though, this seems like a blocker. Anyone seen these before or have any ideas?

DanTup:
(I think a lot of these test failures are unrelated to this change - running `flutter test --platform chrome test/*.dart test/{core,fixtures,support}/` (as the bot does) on `master` results in lots of null-related failures locally too).

kenzieschmoll:
If you sync to head, the build should be green now

DanTup:
Thanks! I've rebased to see what the bots thing, and I'm going to try and get to the bottom of the errors mentioned above tomorrow (I think I have some ideas about at least one of them).

DanTup:
The "Maximum call stack exceeded" error has been raised at https://github.com/flutter/flutter/issues/66122 and also mentions mixins, so I think that's the same thing we're seeing. I've added [notes](https://github.com/flutter/flutter/issues/66122#issuecomment-713498834) with repro instructions (without these changes - it's very easy to trigger). I think that's likely unrelated to the release mode error (seems hot-reload specific).

I've raised https://github.com/flutter/flutter/issues/68696 about the blank errors - getting a proper error message might make it easier to track down the release mode error (which is a show-stopper for merging this).

DanTup:
Latest Flutter master has a fix for the maximum call stack issue. Testing with that, I no longer see that issue. Nor do I see the canvaskit errors, the blank errors in the console nor the tabs not working sometimes when using a release build.

I've also pushed an additional change to re-fix dragging multiple JSON exports over each other (previously `navigateIfNotCurrent` would abort additional drops, as they looked like they were navigating to the same page).

So if the bot is green, I think this is all ready. It's possible there may be some navigations I've missed while testing, so if you can think of any less-used sections I may have missed, let me know (or test and shout with any issues you find).

packages/devtools_app/lib/src/service_manager.dart
DanTup: This change has gone from this PR now during the rebase, as the `?` was already landed to fix another (or perhaps the same) issue in https://github.com/flutter/devtools/pull/2452.

kenzieschmoll:
One other navigation to test is navigating to the app size tool from the connect page. Does this work as expected?

packages/devtools_app/lib/src/routing.dart
```diff
@@ -0,0 +1,161 @@
...
+/// Represents a Page/route for a DevTools screen.
+class DevToolsRouteConfiguration {
+  DevToolsRouteConfiguration(this.page, this.args);
+
+  final String page;
+  final Map<String, String> args;
+}
+
+/// Converts between structured DevToolsRouteConfiguration (our internal data
+/// for pages/routing) and RouteInformation (generic data that can be persisted
```
kenzieschmoll, line 23: nit: put DevToolsRouteConfiguration and RouteInformation in brackets []

DanTup:
> One other navigation to test is navigating to the app size tool from the connect page. Does this work as expected?

It seems to (though I haven't used it before so I'm assuming what I'm seeing is right). Clicking it goes to a page where I can drag on a size json file, and it seems to all render correctly (there's no exit button I can see, but clicking back takes you back to the connection screen).

Is it expected that App Size also appears in the tabs, but when you click on it, you lose the tab bar?

packages/devtools_app/lib/src/routing.dart
```diff
@@ -0,0 +1,161 @@
...
+        PopNavigatorRouterDelegateMixin<DevToolsRouteConfiguration> {
+  DevToolsRouterDelegate(this._getPage, [GlobalKey<NavigatorState> key])
+      : navigatorKey = key ?? GlobalKey<NavigatorState>();
+
+  static DevToolsRouterDelegate of(BuildContext context) =>
+      Router.of(context).routerDelegate as DevToolsRouterDelegate;
+
+  @override
+  final GlobalKey<NavigatorState> navigatorKey;
+  final Page Function(BuildContext, String, Map<String, String>) _getPage;
```
kenzieschmoll, line 66: new line above this member

kenzieschmoll (approved):

packages/devtools_app/lib/src/routing.dart
DanTup: Done!

packages/devtools_app/lib/src/routing.dart
DanTup: Done!

DanTup:
The bot failures seem unrelated to my change, as the bot was green before I pushed 11145e81fdd0011bcb89e1eb0793e6fac0002093 and then went red (and that change is only whitespace/comments). However I can reproduce that failure locally seemingly 100% of the time (even using `master` and the same Flutter version it built against green) - my guess it's maybe a race that the bot mostly wins?

I'll try re-running them on this one anyway.

PR #2413, opened 2020-10-06
Fix some overflow errors and cleanup getDartIOVersion method

kenzieschmoll:


terrylucas (approved):

packages/devtools_app/lib/src/network/network_request_inspector_views.dart
```diff
@@ -272,32 +272,34 @@ class HttpRequestCookiesView extends StatelessWidget {
...
-                context,
-                'Request Cookies',
-                requestCookies,
-                constraints,
-                requestCookiesKey,
-                requestCookies: true,
-              ),
-          ],
+        return SingleChildScrollView(
+          child: Column(
```
jacob314, line 276: Why not a ListView instead of a Column and a SingleChildScrollView?
You can't be getting any of the benefits of the column once it is wrapped in a scroll view because you would get a constraint assert error as soon as any child used a flex factor as the height is unconstrained.

kenzieschmoll: done

PR #2416, opened 2020-10-07
Added initial VM developer mode support

bkonyi:
- Added `VmServicePrivate` extension to `VmServiceWrapper` where all private
  RPC calls required for VM developer options will be implemented. These
  RPCs will throw if invoked while VM developer mode is disabled.

- Added a "Enable VM developer mode" toggle under preferences. Currently
  hidden until VM developer functionality is added.

- Refactored `VmServiceWrapper` to allow for `trackFuture` to be mocked.

image

### DIFF
--- a/packages/devtools_app/lib/src/preferences.dart
+++ b/packages/devtools_app/lib/src/preferences.dart
@@ -25,7 +25,7 @@ class PreferencesController {
 
     // Get the current values and listen for and write back changes.
     String value = await storage.getValue('ui.darkMode');
-    _darkModeTheme.value = value == null || value == 'true';
+    toggleDarkModeTheme(value == null || value == 'true');
     _darkModeTheme.addListener(() {
       storage.setValue('ui.darkMode', '${_darkModeTheme.value}');
     });
--- a/packages/devtools_app/lib/src/preferences.dart
+++ b/packages/devtools_app/lib/src/preferences.dart
@@ -13,8 +13,8 @@ class PreferencesController {
   final ValueNotifier<bool> _darkModeTheme = ValueNotifier(true);
   final ValueNotifier<bool> _vmDeveloperMode = ValueNotifier(false);
 
-  ValueListenable get darkModeTheme => _darkModeTheme;
-  ValueListenable get vmDeveloperModeEnabled => _vmDeveloperMode;
+  ValueListenable<bool> get darkModeTheme => _darkModeTheme;
+  ValueListenable<bool> get vmDeveloperModeEnabled => _vmDeveloperMode;
 
   Future<void> init() async {
     if (storage == null) {
@@ -31,7 +31,7 @@ class PreferencesController {
     });
 
     value = await storage.getValue('ui.vmDeveloperMode');
-    toggleVmDeveloperMode(value == 'false');
+    toggleVmDeveloperMode(value == 'true');
     _vmDeveloperMode.addListener(() {
       storage.setValue('ui.vmDeveloperMode', '${_vmDeveloperMode.value}');
     });
--- a/packages/devtools_app/lib/src/preferences.dart
+++ b/packages/devtools_app/lib/src/preferences.dart
@@ -6,12 +6,15 @@ import 'package:flutter/foundation.dart';
 
 import 'config_specific/logger/logger.dart';
 import 'globals.dart';
+import 'vm_service_wrapper.dart';
 
 /// A controller for global application preferences.
 class PreferencesController {
   final ValueNotifier<bool> _darkModeTheme = ValueNotifier(true);
+  final ValueNotifier<bool> _vmDeveloperMode = ValueNotifier(false);
 
   ValueListenable get darkModeTheme => _darkModeTheme;
+  ValueListenable get vmDeveloperModeEnabled => _vmDeveloperMode;
 
   Future<void> init() async {
     if (storage == null) {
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/vm_service_wrapper.dart
```diff
@@ -10,12 +10,16 @@ import 'package:vm_service/vm_service.dart';
...
+typedef TrackFuture = Future<T> Function<T>(String, Future<T>);
+
 class VmServiceWrapper implements VmService {
   VmServiceWrapper(
     this._vmService,
     this.connectedUri, {
     this.trackFutures = false,
-  });
+  }) {
+    trackFuture = _trackFuture;
```
kenzieschmoll, line 21: why is this necessary? can't we just make the _trackFuture method public and add a visibleForTesting annotation?

packages/devtools_app/test/preferences_controller_test.dart
```diff
@@ -29,5 +29,18 @@ void main() {
...
+
+    test('toggleVmDeveloperMode', () {
+      bool valueChanged = false;
+      final originalValue = controller.vmDeveloperModeEnabled.value;
+
+      controller.vmDeveloperModeEnabled.addListener(() {
+        valueChanged = true;
+      });
+
+      controller.toggleVmDeveloperMode(!controller.darkModeTheme.value);
```
kenzieschmoll, line 41: this should be vmDeveloperModeEnabled, not dartModeTheme

packages/devtools_app/test/vm_service_private_test.dart
```diff
@@ -0,0 +1,43 @@
...
+
+import 'support/mocks.dart';
+
+void main() {
+  FakeServiceManager fakeServiceManager;
+
+  setUp(() {
+    fakeServiceManager = FakeServiceManager();
+    setGlobal(ServiceConnectionManager, fakeServiceManager);
+  });
```
kenzieschmoll, line 16: new line

packages/devtools_app/test/vm_service_private_test.dart
```diff
@@ -0,0 +1,43 @@
...
+      () async {
+    // ignore: unnecessary_cast
+    when(fakeServiceManager.service.trackFuture as Function).thenReturn(
+      <T>(String _, Future<T> future) => future,
+    );
+    when(fakeServiceManager.service
+            .callMethod(argThat(equals('_collectAllGarbage'))))
+        .thenAnswer(
+      (_) => Future.value(Success()),
+    );
```
kenzieschmoll, line 27: can we just override these methods directly in the FakeServiceManager class?

bkonyi: It's not trivial to override `callMethod` since its response varies depending on the method being invoked. We'd have to add a `Response callMethodResult` property to handle this, but that wouldn't work for more complex tests where `callMethod` is called multiple times with different results within a method being tested which is why I opted to use a mock instead of a fake VM service.

However, I'm willing to switch to a fake if you'd prefer.

packages/devtools_app/lib/src/preferences.dart
```diff
@@ -6,12 +6,15 @@ import 'package:flutter/foundation.dart';
...
 import 'globals.dart';
+import 'vm_service_wrapper.dart';
 
 /// A controller for global application preferences.
 class PreferencesController {
   final ValueNotifier<bool> _darkModeTheme = ValueNotifier(true);
+  final ValueNotifier<bool> _vmDeveloperMode = ValueNotifier(false);
 
   ValueListenable get darkModeTheme => _darkModeTheme;
+  ValueListenable get vmDeveloperModeEnabled => _vmDeveloperMode;
```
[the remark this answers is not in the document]
bkonyi: Yeah, I can't see why not. Was just trying to be consistent :-) Done.

packages/devtools_app/lib/src/preferences.dart
```diff
@@ -21,15 +24,27 @@ class PreferencesController {
...
     // Get the current values and listen for and write back changes.
-    final String value = await storage.getValue('ui.darkMode');
+    String value = await storage.getValue('ui.darkMode');
     _darkModeTheme.value = value == null || value == 'true';
     _darkModeTheme.addListener(() {
       storage.setValue('ui.darkMode', '${_darkModeTheme.value}');
     });
+
+    value = await storage.getValue('ui.vmDeveloperMode');
+    toggleVmDeveloperMode(value == 'false');
```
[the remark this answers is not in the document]
bkonyi: Yes, it should. Whoops.

packages/devtools_app/lib/src/vm_service_wrapper.dart
bkonyi: I wasn't sure if we'd want to make `_trackFuture` public, but I guess this effectively does that anyway. Removed and made `_trackFuture` public with `@visibleForTesting`.

packages/devtools_app/test/vm_service_private_test.dart
kenzieschmoll: Ah I see what you are saying. Leaving as is is fine.

kenzieschmoll: I was originally talking about mocking and overriding `VmServicePrivate.collectAllGarbage` but I see now why that wouldn't work

packages/devtools_app/lib/src/preferences.dart
```diff
@@ -21,15 +24,27 @@ class PreferencesController {
     }
 
     // Get the current values and listen for and write back changes.
-    final String value = await storage.getValue('ui.darkMode');
+    String value = await storage.getValue('ui.darkMode');
     _darkModeTheme.value = value == null || value == 'true';
```
kenzieschmoll, line 28: I know it's not your code, but while you're here would you mind changing this to use toggleDarkModeTheme?

kenzieschmoll (approved):

packages/devtools_app/lib/src/preferences.dart
bkonyi: Not a problem!

Issue #2025, opened 2020-06-01
timeline page feedback


devoncarew:
This is separate from https://github.com/flutter/devtools/issues/1964 (though, we could just roll this issue into that one for convenience):

- [x] in the FPS chart, the area above the frame bars should be clickable in order to select that frame
- [x] we may consider fixing the y-axis range for the chart, or only having one of a few set ranges (2 frame times worth - 33ms? 50ms?)
- [x] in the main timeline chart, the row labels scroll away to the left; it would be nice if these were always visible (if the main area scrolled both in x and y, but the left hand y-axis labels only scrolled in the y direction)
- [ ] when entering the page, the layout phase of the main timeline area takes ~300ms to layout

@kenzieschmoll

kenzieschmoll:
closing this issue. I'm not seeing jank loading the Timeline page in release mode. Maybe there have been improvements since the time this issue was filed.

PR #2419, opened 2020-10-08
Make entire frame clickable in timeline flutter frames chart

kenzieschmoll:
Some timeline cleanup. Previously only the colored bars were clickable. Now the whole column containing the two bars is clickable. Addresses part of https://github.com/flutter/devtools/issues/2025.


jacob314 (approved):
lgtm

PR #2421, opened 2020-10-08
Change flutter frames chart y axis scale to 2 * target frame time.

kenzieschmoll:
Addresses part of https://github.com/flutter/devtools/issues/2025. Before, this was variable based on the longest frame time. This gives the chart a consistent scale.

### DIFF
--- a/packages/devtools_app/lib/src/timeline/flutter_frames_chart.dart
+++ b/packages/devtools_app/lib/src/timeline/flutter_frames_chart.dart
@@ -50,6 +50,10 @@ class _FlutterFramesChartState extends State<FlutterFramesChart>
 
   double get availableChartHeight => defaultChartHeight - defaultSpacing;
 
+  /// Milliseconds per pixel value for the y-axis.
+  ///
+  /// This value will result in a y-axis time range spanning two times the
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/timeline/flutter_frames_chart.dart
```diff
@@ -57,8 +51,8 @@ class _FlutterFramesChartState extends State<FlutterFramesChart>
   double get availableChartHeight => defaultChartHeight - defaultSpacing;
 
   double get msPerPx =>
-      widget.longestFrameDurationMs.clamp(minMsForDisplay, maxMsForDisplay) /
-      availableChartHeight;
+      // Multiply by two to reach two times the target frame time.
```
jacob314, line 54: nit: make comment more clearly describe the why not the how. For example something like:
```/// Milliseconds per pixel value to display a time range spanning two times the target frame rate.```

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/timeline/flutter_frames_chart.dart
kenzieschmoll: done

PR #2423, opened 2020-10-09
Don't try to initialize if it isn't enabled

grouma:
Our internal infrastructure always calls `setUpAnalytics` which can cause issues during development. Properly handle when the GTAGs are not enabled.


devoncarew (approved):
I'm curious - when does isGTagsEnabled return false?

grouma:
> I'm curious - when does _isGtagsEnabled return false?

We check for analytics support [here](https://github.com/flutter/devtools/blob/a8438cd926e7bcac24e120d4b3a920b080d74753/packages/devtools_app/lib/src/analytics/remote_provider.dart#L56)

PR #2429, opened 2020-10-14
Allow for Screens to dynamically change visibility of isolate selector

bkonyi:
The VM tools screen will be using a NavigationRail for its subviews and some of the views require the isolate selector whereas others don't. This change updates Screen's `showIsolateSelector` getter to return a `ValueListenable<bool>` instead of `bool`, allowing for Screens to change the visibility of the selector dynamically.


packages/devtools_app/lib/src/debugger/debugger_screen.dart
```diff
@@ -46,7 +47,7 @@ class DebuggerScreen extends Screen {
   String get docPageId => screenId;
 
   @override
-  bool get showIsolateSelector => true;
+  ValueListenable<bool> get showIsolateSelector => ValueNotifier<bool>(true);
```
jacob314, line 50: Returning a new ValueNotifier object each time the getter is called makes me very nervous. It would be easy for a caller to fail to correctly handle the lifecycle for event listeners.

jacob314: I would create a const FixedValueListenable class that implements ValueListenable and just returns a fixed value and never adds or removes listeners. That way you won't have the weirdness of returning a new ValueListenable every time this is queried.

From your pages where showIsolateSelector is not constant, where are you storing the ValueListenable? It will need to be stored on controller or a state object not on a Screen object as Stateless widgets should not have mutable state. As part of a hot reload, all your widgets will be replaced with new widgets so you will lose your state if you store it on the widgets directly as well as violating other Flutter assumptions.

bkonyi: Good point, I didn't even think of that. Up until this point I was keeping the ValueNotifier in the screen, but that's not too hard to move to a controller.

Out of curiosity, do we have a best practices document for DevTools or Flutter that I should read up on? I'm also trying to pull together a getting started guide for VM devs working on the VM tool pages, so it would probably be good to include those references in the guide if they exist.

jacob314: https://flutter.dev/docs/development/data-and-backend/state-mgmt/simple
Is a reasonable starting point Iin DevTools we don't really have much mutable state so we can keep things simpler than if less of the data was read only.

Reasonable page on layouts.
https://flutter.dev/docs/development/ui/layout

packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -492,6 +492,23 @@ class ToolbarAction extends StatelessWidget {
   }
 }
 
+/// A [ValueListenable] that ignores all added listeners and returns a fixed
+/// value.
+class FixedValueListenable<T> extends ValueListenable<T> {
```
jacob314, line 497: I like it! 
Move it to a separate file for value listenable helpers as this isn't a Widget.
Maybe:
`listenable.dart`
I expect at some point we will add a helper for a List that may have its elements change.

jacob314 (approved):
lgtm

PR #2431, opened 2020-10-14
Test bots

DanTup:
This PR does nothing, but should force to run the bots to run, to see if they fail in the same ways I'm seeing locally and on mr PR...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,9 +47,3 @@ script: ./tool/travis.sh
 # See https://github.com/flutter/devtools/pull/398
 # branches:
 #   only: [master]
-
-cache:
- directories:
- - $HOME/.pub-cache # macOS / Linux
- - $APPDATA/Roaming/Pub/Cache # Windows
- - packages/devtools_app/.dart_tool/build


DanTup:
@jacob314 @devoncarew FYI - these are the failures I was seeing. This PR is empty (so the same as the last master build, but with latest Flutter `dev`) but there are a lot of failures (null-safety related).

devoncarew:
Thanks! Just to level set things, can you remove the travis caching from this PR (https://github.com/flutter/devtools/blob/master/.travis.yml#L51)?

DanTup:
@devoncarew doesn't appear to have made any difference.

I don't know if the first part of the error is the interesting bit:

```
Running "flutter pub get" in _flutter_web_build_script...           2.1s
Configuring `build_web_compilers:entrypoint` in target
`devtools_app:devtools_app` but this is not a known Builder
Error creating devtools_app|lib/main.ddc.dill
Error creating kernel summary for module:devtools_app|lib/main.ddc.dill
```

I'm not sure if these are a result of the nullable issues (or perhaps the cause). I'm not very familiar with how this works.

DanTup:
(looks like how these tests run is changing - they've been removed in https://github.com/flutter/devtools/pull/2438 and https://github.com/flutter/devtools/issues/2437 raised about making them work using the new way).

PR #2432, opened 2020-10-14
add text to the Snapshot button

devoncarew:
- add text to the Snapshot button
- don't indent the `Monitor Allocations` and `Reset Accumulators` buttons if there's nothing to their left

This is the primary thing that people need to interact with on this screen (at least initially); having text on it helps to emphasize it.

cc @terrylucas @jacob314

### DIFF
--- a/packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
+++ b/packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
@@ -460,11 +460,11 @@ class HeapTreeViewState extends State<HeapTree>
     return Row(
       children: [
         Tooltip(
-          message: 'Snapshot',
+          message: 'Take snapshot',
           child: OutlineButton(
             key: snapshotButtonKey,
             onPressed: _isSnapshotRunning ? null : _snapshot,
-            child: createIcon(Icons.camera),
+            child: const MaterialIconLabel(Icons.camera, 'Snapshot'),
           ),
         ),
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -538,7 +538,8 @@ class HeapTreeViewState extends State<HeapTree>
                   child: createIcon(Icons.vertical_align_bottom),
                 ),
               ),
-        const SizedBox(width: defaultSpacing),
+
+        if (!controller.showTreemap.value) const SizedBox(width: defaultSpacing),
```
jacob314, line 542: Nit: move this up in the conditional starting on lin 519 that also checks controller.showTreemap.value
That code could be cleaned up by using list comprehensions instead of returning a no-op SizedBox.

devoncarew: 👍 sounds good - that'll be cleaner than all the ternary experssions

devoncarew:
Updated - ptal.

Screen Shot 2020-10-14 at 1 51 23 PM

Screen Shot 2020-10-14 at 1 51 18 PM

jacob314 (approved):
lgtm

devoncarew:
I think we'll need to re-gen the golden masters...

devoncarew:
I'm going to land this as is, and look at addressing the (unrelated) build failures in other PRs.

PR #2435, opened 2020-10-16
Move DevTools server code out of analytics.dart to server.dart

kenzieschmoll:
analytics.dart previously contained code that was not actually related to analytics, but rather the devtools server and the .devtools file local storage.

There are not any functional changes here - just moves.


grouma (approved):

kenzieschmoll:
landing this on red because the failure is a known issue and is tracked on other PRs. This CL is a prerequisite for the devtools survey work.

PR #2436, opened 2020-10-18
fix a 'capacity' typo

devoncarew:
- fix a typo - change `capcity` to `capacity`


devoncarew:
From some of the CI failures:
- we might need to make sure that we're opt-ing out of null safety (set all our sdk constraints to >=2.9.0?)

and from:

```
org-dartlang-app:///packages/devtools_app/src/config_specific/sse/sse_shim.dart:
5:1: Error: Null safe libraries are not allowed to export declarations from of
opt-out libraries.
export '_fake_sse.dart'
^```

It's possible that the synthetic flutter web entrypoint - `org-dartlang-app:///packages/...` - is assumed to be null safe as it doesn't have a language code. It might need to be generated with one - somehow inferred from the host project?

kenzieschmoll (approved):

PR #2440, opened 2020-10-19
Support line segment opacity.

jacob314:


jacob314:
I need this functionality for a followup cl to reuse these classes in the inspector to draw animated lines.

packages/devtools_app/lib/src/timeline/timeline_flame_chart.dart
```diff
@@ -835,7 +835,11 @@ class AsyncGuidelinePainter extends FlameChartPainter {
             (zoomedLine.end.dy - verticalScrollOffset)
                 .clamp(0.0, constraints.maxHeight),
           ),
-          paint,
+          zoomedLine.opacity == 1.0
+              ? defaultPaint
+              : (Paint()
+                ..color = colorScheme.treeGuidelineColor
```
kenzieschmoll, line 841: can you just write `defaultPaint.withOpacity` here instead of creating a new Paint() object?

kenzieschmoll (approved):

packages/devtools_app/lib/src/timeline/timeline_flame_chart.dart
jacob314: I can't quite do that as it will mutate the existing paint object. But I agree creating the new Paint objects here isn't great given the typical case is all the line segments have the same opacity. Updated the code to cache the last opacity used and only update the paint object as needed.

PR #2442, opened 2020-10-20
Drive devtools survey from metadata hosted on flutter.dev

kenzieschmoll:
Screen Shot 2020-11-02 at 3 23 00 PM

Survey data is pulled from here: https://flutter.dev/f/dart-devtools-survey-metadata.json

### DIFF
--- a/packages/devtools_app/lib/src/notifications.dart
+++ b/packages/devtools_app/lib/src/notifications.dart
@@ -103,19 +103,49 @@ class NotificationsState extends State<_NotificationsProvider>
     super.dispose();
   }
 
-  // TODO(peterdjlee): Support clickable links in notifications. See #2268.
-  /// Pushes a notification [message].
+  // TODO(peterdjlee): Support clickable links in notification text. See #2268.
+  /// Pushes a notification [message], and returns whether the notification was
+  /// successfully pushed.
   @override
-  void push(String message) {
+  bool push(
+    String message, {
+    List<NotificationAction> actions,
+    Duration duration = Notifications.defaultDuration,
+    bool allowDuplicates = true,
+  }) {
+    if (!allowDuplicates &&
+        _notifications.isNotEmpty &&
+        _notifications.where((n) => n.message == message).isNotEmpty) {
+      return false;
+    }
     setState(() {
       _notifications.add(
         _Notification(
           message: message,
+          actions: actions,
           remove: _removeNotification,
+          duration: duration,
         ),
       );
       _overlayEntry?.markNeedsBuild();
     });
+    return true;
+  }
+
+  /// Dismisses all notifications with a matching message.
+  void dismiss(String message) {
+    bool didDismiss = false;
+    for (final notification in _notifications) {
+      if (notification.message == message) {
+        _notifications.remove(notification);
+        didDismiss = true;
+      }
+    }
+    if (didDismiss) {
+      setState(() {
+        _overlayEntry?.markNeedsBuild();
+      });
+    }
   }
 
   void _removeNotification(_Notification notification) {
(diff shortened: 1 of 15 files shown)


packages/devtools_app/lib/src/notifications.dart
```diff
@@ -235,19 +267,65 @@ class _NotificationState extends State<_Notification>
...
+      maxLines: 6,
+    );
+  }
+
+  Widget _buildActions() {
+    if (widget.actions.isEmpty) return const SizedBox();
+    return Row(
+      mainAxisAlignment: MainAxisAlignment.end,
+      children: [
+        for (int i = 0; i < widget.actions.length; i++) ...[
```
jacob314, line 299: we should create a joinWith(List<Widget>, Widget separator) helper method or maybe extension method on List<Widget> as we've got this pattern a few places.

packages/devtools_app/lib/src/survey.dart
```diff
@@ -0,0 +1,166 @@
+import 'dart:convert';
+
+import 'dart:io' if (dart.library.html) 'dart:html';
```
jacob314, line 3: this makes me really nervous as a conditional import. dart:html and dart:io are super different. Can we remove this asnd use package:http and similar?

packages/devtools_app/lib/src/survey.dart
```diff
@@ -0,0 +1,166 @@
...
+  /// Duration for which we should show the survey notification.
+  ///
+  /// We use a very long time here to give the appearance of a persistent
+  /// notification. The user will need to interact with the prompt to dismiss
+  /// it.
+  static const notificationDuration = Duration(days: 1);
+
+  /// The amount of time (one day) to wait between checks to the file at
+  /// [metadataUrl].
+  static const checkInterval = Duration(days: 1);
```
jacob314, line 32: why do we bother with this? just loading the file and relying on browser caching should be fine as long as we don't block the devtools UI waiting for the json file to be loaded.

packages/devtools_app/lib/src/survey.dart
```diff
@@ -0,0 +1,166 @@
...
+
+    final currentTimeMs = DateTime.now().millisecondsSinceEpoch;
+    final activeSurveyRange = Range(
+      _cachedSurvey.startDate.millisecondsSinceEpoch,
+      _cachedSurvey.endDate.millisecondsSinceEpoch,
+    );
+    return activeSurveyRange.contains(currentTimeMs);
+  }
+
+  Future<bool> shouldFetchSurveyContent() async {
```
jacob314, line 96: lets remove this method and just rely on browser caching. loading one tiny json file from a server on devtools startup is completely fine if we don't block devtools startup to do it. if we are concerned we can tweak caching settings on the flutter.dev website to keep it cached for a day.

jacob314 (commented):
done with a round of comments.

packages/devtools_app/lib/src/survey.dart
kenzieschmoll: The only reason this is here is for use of the `HttpStatus.ok` const, but we could just as easily write `200`.

jacob314: lets write 200 or maybe package:http does or should define that const.
dart static checking falls over badly for conditional imports like this.

jacob314 (commented):
The logic for the survey shown count will run into trouble the second time we need to show a survey. Won't we use the survey shown count from the previous server id? Maybe change the survey shown count api to take a survey id so we keep track of the # of times each survey id has been shown.

kenzieschmoll:
These values are already stored by the active survey, so if there is a new survey, a new entry will be created in storage and that entry will have a surveyShownCount 0 and surveyActionTaken false

jacob314:
lgtm

PR #2443, opened 2020-10-21
fix an NPE in the memory view

devoncarew:
- fix an NPE in the memory view (found when running w/ asserts on)
- fix a typo in a variable name


devoncarew:
(landing TBR as I'm hitting this locally w/ other work I'm doing)

kenzieschmoll (commented):
lgtm_army

Issue #2444, opened 2020-10-21
convert the memory page to use tabs


devoncarew:
The memory page currently uses a slider to switch between the heap table and the heap treemap. We should convert this to a tab view - it'll be a bit more natural to the user, and will enable us to add more tabs in the future(for things like the image cache, ...).

cc @terrylucas for FYI (and comments); I'm hoping to tackle this shortly, but wanted to give you a heads up

PR #2446, opened 2020-10-21
move the existing memory page content into tabs

devoncarew:
- move the existing memory page content into tabs (`Dart Heap` and `Heap treemap`)
- fix https://github.com/flutter/devtools/issues/2444
- some minor changes to use `ActionButton` in places - it gives us consistent tooltip delay behavior

Screen Shot 2020-10-22 at 2 31 59 PM

Screen Shot 2020-10-22 at 2 32 12 PM

I'm having some trouble with ~4 of the service extension tests - not sure how those failures are related to the changes in this PR.

### DIFF
--- a/packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
+++ b/packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
@@ -360,24 +360,26 @@ class HeapTreeViewState extends State<HeapTree>
       snapshotDisplay = const SizedBox();
     }
 
-    return Column(
-      children: [
-        const SizedBox(height: denseRowSpacing),
-        Row(
-          children: [
-            _buildSnapshotControls(textTheme),
-            const Expanded(child: SizedBox(width: defaultSpacing)),
-            // TODO(peterdjlee): Implement filter and search functionality for treemap.
-            _buildSearchFilterControls(),
-          ],
-        ),
-        const SizedBox(height: denseRowSpacing),
-        Expanded(
-          child: OutlineDecoration(
-            child: buildSnapshotTables(snapshotDisplay),
+    return Padding(
+      padding: const EdgeInsets.only(top: denseRowSpacing),
+      child: Column(
+        children: [
+          Row(
+            children: [
+              _buildSnapshotControls(textTheme),
+              const Expanded(child: SizedBox(width: defaultSpacing)),
+              // TODO(peterdjlee): Implement filter and search functionality for treemap.
+              _buildSearchFilterControls(),
+            ],
           ),
-        ),
-      ],
+          const SizedBox(height: denseRowSpacing),
+          Expanded(
+            child: OutlineDecoration(
+              child: buildSnapshotTables(snapshotDisplay),
+            ),
+          ),
+        ],
+      ),
     );
   }
 
--- a/packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
+++ b/packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
@@ -27,7 +27,6 @@ import 'memory_graph_model.dart';
 import 'memory_instance_tree_view.dart';
 import 'memory_protocol.dart';
 import 'memory_snapshot_models.dart';
-import 'memory_treemap.dart';
 
 final memorySearchFieldKey = GlobalKey(debugLabel: 'MemorySearchFieldKey');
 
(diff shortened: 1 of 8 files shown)


packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -354,23 +354,21 @@ class HeapTreeViewState extends State<HeapTree>
...
-      }
+      snapshotDisplay = MemoryHeapTable();
     } else {
+      // TODO: Have some help text about how to take a snapshot.
       snapshotDisplay = const SizedBox();
     }
 
     return Column(
       children: [
+        const SizedBox(height: denseRowSpacing),
```
jacob314, line 365: Nit: probably cleaner as padding on the column instead of a child of the column. That way the layout will still be what you want if some other tweaks to the Column's properties were made.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -459,83 +457,67 @@ class HeapTreeViewState extends State<HeapTree>
   Widget _buildSnapshotControls(TextTheme textTheme) {
     return Row(
       children: [
-        Tooltip(
-          message: 'Take snapshot',
+        ActionButton(
```
jacob314, line 460: an action button with a child that is an outline button seems surprising. Can we just have an ActionButton?

packages/devtools_app/lib/src/memory/memory_heap_treemap.dart
```diff
@@ -90,22 +93,38 @@ class MemoryTreemapState extends State<MemoryTreemap> with AutoDisposeMixin {
...
+          ),
+        ],
       );
-    } else {
-      return const SizedBox();
     }
+
+    return Column(
+      children: [
+        const SizedBox(height: denseRowSpacing),
```
jacob314, line 111: nit: make this padding on the column

packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -77,25 +78,42 @@ class MemoryScreen extends Screen {
 class MemoryBody extends StatefulWidget {
   const MemoryBody();
 
+  static const ValueKey dartHeapTabKey = ValueKey('Dart Heap');
```
jacob314, line 81: Do we really need keys for these? I don't see the keys used anywhere.

packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -148,6 +166,7 @@ class MemoryBodyState extends State<MemoryBody> with AutoDisposeMixin {
         : '';
 
     _memoryEvents ??= MemoryEventsPane();
```
jacob314, line 168: while we are here we should remove the _memoryChart and _memoryEvents local variables. Caching widgets is most likely an antipattern as it can mess up hot reload and isn't needed for performance.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
devoncarew: done

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
devoncarew: The ActionButton class gets us a tooltip, and sets it up w/ some consistent app defaults around tooltip delay. Besides the tooltip text it just takes a child, which can be anything.

packages/devtools_app/lib/src/memory/memory_heap_treemap.dart
devoncarew: will do

packages/devtools_app/lib/src/memory/memory_screen.dart
devoncarew: I'll remove them.

packages/devtools_app/lib/src/memory/memory_screen.dart
devoncarew: I'll remove and verify that things still look good.

devoncarew (commented):
Thanks! PR updated (I'll still need to fix a few failing tests).

devoncarew:
The test failures - for whatever reason - relate to adding the memory_heap_treemap.dart import to the memory_screen.dart file. Perhaps it's transitively bringing in some dart:io references?

https://github.com/flutter/devtools/pull/2446/files#diff-5f903c7c0ee90b669340f30c88dc445f2aaef84e647c2fec8cca1eaba420207fR24

devoncarew:
Tests fixed w/ some import simplifications.

Issue #2424, opened 2020-10-10
DevTools TimeLine cannot zoom more


AlexFeng007:
My DevTools version 0.9.3+4,
 In the Timeline views, i cannot zoom more clear in using my mouse wheels, but GPU Frame chart is ok, what can i do?

kenzieschmoll:
Do you mean that at some level, you cannot zoom in any further? Or are you seeing some zooming jank where when you zoom in, the view jumps around to another horizontal scroll offset?

And just for clarification, [this](https://flutter.dev/docs/development/tools/devtools/timeline#timeline-events-chart) is the part of the timeline you are trying to zoom correct? This is the "zoomable" chart in the Timeline.

152dyn:
The notes on my computer didn't expand, and they didn't happen on my colleague's computer

![image](https://user-images.githubusercontent.com/19378921/96680743-ac3c5280-13a8-11eb-9bba-42a795ea7142.png)

AlexFeng007:
@kenzieschmoll  Same with my timeline views, just cannot zoom in any further

kenzieschmoll:
What does the chart look like if you zoom all the way out? I suspect the recorded duration is very long, and that the max zoom level is being hit.

152dyn:
@kenzieschmoll 

The minimum

![image](https://user-images.githubusercontent.com/19378921/96857673-d584dd80-1491-11eb-822a-f77e2d45b34c.png)

The biggest

![image](https://user-images.githubusercontent.com/19378921/96857764-f3ead900-1491-11eb-8092-2d3dfdef2fcd.png)

PR #2447, opened 2020-10-22
Remove max zoom level bound for flame charts

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/2424. 

AnimationController's `upperBound` member defaults to 1.0 unless specified, so we need to specify an `upperBound` (`FlameChart.maxZoomLevel`). `AnimationController.unbounded(...)` defaults the `upperBound` member to `double.infinity`, so we are mimicking that behavior here.


grouma (approved):

jacob314 (approved):
lgtm

PR #2448, opened 2020-10-22
Avoid throwing exceptions accessing the server when it is not available.

jacob314:
These exceptions occur every time the app starts up making it annoying to run the app while breaking on exceptions.

### DIFF
--- a/packages/devtools_app/lib/src/survey.dart
+++ b/packages/devtools_app/lib/src/survey.dart
@@ -29,11 +29,11 @@ class SurveyService {
 
   Future<DevToolsSurvey> get activeSurvey async {
     _cachedSurvey ??= await _fetchSurveyContent();
-    if (_cachedSurvey != null) {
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/survey.dart
```diff
@@ -29,11 +29,11 @@ class SurveyService {
 
   Future<DevToolsSurvey> get activeSurvey async {
     _cachedSurvey ??= await _fetchSurveyContent();
-    if (_cachedSurvey != null) {
+    if (_cachedSurvey != null && server.isDevToolsServerAvailable) {
```
kenzieschmoll, line 32: we should add this check before we fetch the survey content and return null if false, because if the server is unavailable we don't need to do anything survey related

jacob314: that is better. done.

kenzieschmoll (approved):

kenzieschmoll:
Landing this to fix noisy console during development 👍

PR #2449, opened 2020-10-22
Pass background color as part of color scheme.

jacob314:


jacob314 (commented):
it is cleaner to use the colorScheme than scaffoldBackgroundColor to access the preferred background color but we weren't updating the colorScheme to reflect the ide background color.

kenzieschmoll (approved):

Issue #2414, opened 2020-10-06
DevTools fails to connect when an isolate is paused


bkonyi:
After connecting to a VM service, DevTools attempts to initialize state based on the responses of certain service extensions. However, this fails for `httpTimelineLoggingEnabled` when invoked on a paused isolate as it runs in-process within `dart:io`, preventing the request from completing until the isolate is resumed.

PR #2450, opened 2020-10-22
Prevent dart.io extensions from being called on paused isolates

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/2414

### DIFF
--- a/packages/devtools_app/lib/src/service_manager.dart
+++ b/packages/devtools_app/lib/src/service_manager.dart
@@ -208,6 +208,7 @@ class ServiceConnectionManager {
     _stateController.add(true);
 
     await _isolateManager._initIsolates(vm.isolates);
+    service.onDebugEvent.listen(_isolateManager._handleDebugEvent);
     service.onIsolateEvent.listen(_isolateManager._handleIsolateEvent);
     service.onExtensionEvent
         .listen(_serviceExtensionManager._handleExtensionEvent);
@@ -461,7 +462,8 @@ class IsolateManager {
           ._maybeAddServiceExtension(event.extensionRPC);
 
       // Check to see if there is a new isolate.
-      if (_selectedIsolate == null && _isFlutterExtension(event.extensionRPC)) {
+      if (_selectedIsolate == null &&
+          extensions.isFlutterExtension(event.extensionRPC)) {
         await _setSelectedIsolate(event.isolate);
       }
     } else if (event.kind == EventKind.kIsolateExit) {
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -795,36 +821,50 @@ class ServiceExtensionManager {
...
-      // https://github.com/dart-lang/sdk/issues/43628.
-      if (name == extensions.httpEnableTimelineLogging.extension) {
-        await _service.forEachIsolate((isolate) async {
-          await _service.httpEnableTimelineLogging(isolate.id, value);
-        });
-      } else {
+    final callExtension = () async {
+      assert(value != null);
+      if (value is bool) {
+        // TODO(kenz): stop special casing http timeline logging once it can be
```
kenzieschmoll, line 827: High level question (see lines 831-833): should we be calling the httpTimeline dart.io extension on all isolates or just the selected isolate like we do for other service extensions? Or on the flip side, should we be calling the other service extensions on all isolates like we are currently doing for the http timeline extension? @bkonyi @jacob314

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -744,37 +756,51 @@ class ServiceExtensionManager {
...
-          return;
-        case int:
-        case double:
-          final num value = num.parse(
-              response.json[name.substring(name.lastIndexOf('.') + 1)]);
-          await _maybeRestoreExtension(name, value);
-          return;
-        default:
-          return;
+    final restore = () async {
```
jacob314, line 759: can't this can just be
`void restore() async {`
?
I don't see anywhere where you depend on equality for the closure.

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -795,36 +821,50 @@ class ServiceExtensionManager {
...
-    if (value is bool) {
-      // TODO(kenz): stop special casing http timeline logging once it can be
-      // called via `callServiceExtension`. See
-      // https://github.com/dart-lang/sdk/issues/43628.
-      if (name == extensions.httpEnableTimelineLogging.extension) {
-        await _service.forEachIsolate((isolate) async {
-          await _service.httpEnableTimelineLogging(isolate.id, value);
-        });
-      } else {
+    final callExtension = () async {
```
jacob314, line 824: same question as for restore

packages/devtools_app/lib/src/service_manager.dart
jacob314: i think it is reasonable to call this for all isolates. The other extensions are flutter specific and only apply to the main isolate.

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -795,36 +821,50 @@ class ServiceExtensionManager {
...
-        args: {name.substring(name.lastIndexOf('.') + 1): value},
-      );
+    };
+
+    final Isolate isolate =
+        await _service.getIsolate(_isolateManager.selectedIsolate.id);
+    // Do not try to call Dart IO extensions for a paused isolate.
+    if (extensions.isDartIoExtension(name) &&
+        isolate.pauseEvent.kind.contains('Pause')) {
+      callbacksOnIsolateResume
```
jacob314, line 863: maybe i'm missing something but why can't we just make these extensions work for paused vm isolates?

kenzieschmoll: From @bkonyi in a conversation we had earlier today:
"Service extensions registered via dart:developer in general shouldn't be called while isolates are paused since they execute Dart code on the isolate"

packages/devtools_app/lib/src/service_manager.dart
kenzieschmoll: you're right. fixed

packages/devtools_app/lib/src/service_manager.dart
jacob314: we don't have to deal with it now but how do we handle the edge case of service extensions that need to be set at or soon after the application starts?

jacob314 (approved):
lgtm

Issue #2454, opened 2020-10-23
consider moving our CI to github actions


devoncarew:
We should consider moving our CI to github actions - the travis jobs seem slow to start lately.

devoncarew:
this is complete

PR #2455, opened 2020-10-23
experiment with github actions

devoncarew:
- WIP PR to experiment with github actions
- some work towards https://github.com/flutter/devtools/issues/2454


devoncarew:
I'm going to TBR this so I can see the results in the 'Actions' tab, but will revert if we don't go forward with this.

PR #2456, opened 2020-10-23
Add setting to opt in/out of analytics

kenzieschmoll:
analytics-setting

### DIFF
--- a/packages/devtools_app/lib/src/analytics/analytics.dart
+++ b/packages/devtools_app/lib/src/analytics/analytics.dart
@@ -168,16 +168,18 @@ class GtagExceptionDevTools extends GtagException {
   external String get flutter_client_id;
 }
 
-ValueNotifier<bool> gaEnabledNotifier = ValueNotifier(false);
+ValueNotifier<bool> _gaEnabledNotifier = ValueNotifier(false);
+
+ValueListenable<bool> get gaEnabledNotifier => _gaEnabledNotifier;
 
 // Exposed function to JS via allowInterop.
-bool gaEnabled() => gaEnabledNotifier.value;
+bool gaEnabled() => _gaEnabledNotifier.value;
 
 /// Request DevTools property value 'enabled' (GA enabled) stored in the file
 /// '~\.devtools'.
 Future<bool> isAnalyticsEnabled() async {
-  gaEnabledNotifier.value = await server.isAnalyticsEnabled();
-  return gaEnabledNotifier.value;
+  _gaEnabledNotifier.value = await server.isAnalyticsEnabled();
+  return _gaEnabledNotifier.value;
 }
 
 /// Set the DevTools property 'enabled' (GA enabled) stored in the file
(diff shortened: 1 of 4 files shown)


kenzieschmoll:
@grouma - just want to verify that this change will still play well with g3

packages/devtools_app/lib/src/app.dart
```diff
@@ -440,6 +440,11 @@ class SettingsDialog extends StatelessWidget {
             listenable: preferences.darkModeTheme,
             toggle: preferences.toggleDarkModeTheme,
           ),
+          _buildOption(
```
jacob314, line 443: It would be nice to make both these checkboxes use NotifierCheckbox which is a little cleaner as there is not a separate toggle method and listenable.

packages/devtools_app/lib/src/analytics/analytics.dart
```diff
@@ -167,24 +168,24 @@ class GtagExceptionDevTools extends GtagException {
...
-bool _gaEnabled;
+ValueNotifier<bool> gaEnabledNotifier = ValueNotifier(false);
 
 // Exposed function to JS via allowInterop.
-bool gaEnabled() => _gaEnabled;
+bool gaEnabled() => gaEnabledNotifier.value;
 
 /// Request DevTools property value 'enabled' (GA enabled) stored in the file
 /// '~\.devtools'.
 Future<bool> isAnalyticsEnabled() async {
```
jacob314, line 178: remove these two methods and trigger sending events to the server based on listening for changes to the gaEnabledNotifier. That way the notifier and the server can't get out of sync. Right now someone could modify gaEnabledNotifier and then the server wouldn't record the value.

grouma:
> @grouma - just want to verify that this change will still play well with g3

I don't think this will impact g3 but I'm wondering if we should make the option configurable and pass in the configuration through `main.dart`.

packages/devtools_app/lib/src/app.dart
kenzieschmoll: I agree we should find a way to merge the functionality here. One thing this has that NotifierCheckbox does not is that the checkbox label is also clickable. Additionally, some of these options have toggle callbacks that do more than just update the notifier, so replacement won't be trivial. I will add a todo.

kenzieschmoll:
> > @grouma - just want to verify that this change will still play well with g3
> 
> I don't think this will impact g3 but I'm wondering if we should make the option configurable and pass in the configuration through main.dart.

Ah - as in disallow opt-out for g3 users? That makes sense.

packages/devtools_app/lib/src/analytics/analytics.dart
kenzieschmoll: I took a different approach that accomplishes the same ends. I made the notifier private and added a public listenable getter. So the notifier value can only be modified in the context of these methods.

jacob314 (approved):
lgtm

Issue #2305, opened 2020-08-29
Automatic port number fallback


xster:
```bash 
pub global run devtools
Unhandled exception:
SocketException: Failed to create server socket (OS Error: Address already in use, errno = 48), address = 127.0.0.1, port = 9100
#0      serveDevTools (package:devtools_server/src/server.dart:235:5)
<asynchronous suspension>
#1      serveDevToolsWithArgs (package:devtools_server/src/server.dart:156:10)
#2      main (file:///Users/xster/.pub-cache/hosted/pub.dartlang.org/devtools-0.9.0/bin/devtools.dart:9:3)
#3      _startIsolate.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:299:32)
#4      _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:168:12)
```

We can probably show a simpler error message and keep cycling random numbers to try different ports? 

Also not sure why it failed to create the socket. `lsof -i` doesn't seem to show other processes using port 9100.

kenzieschmoll:
I thought we already attempt to increase the port number until one works? Maybe we have regressed.

PR #2458, opened 2020-10-23
Increase default try ports to 10 when launching DevTools

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/2305

10 is fairly arbitrary. We can modify as needed.


jacob314 (approved):
Fyi @helin24 here is a case where we try multiple different ports. I think the original motivation for this work was for chromeos where we needed to support port forwarding.

PR #2459, opened 2020-10-23
Use font size from IDE in theme (WIP)

helin24:
This passes font size from URL parameters and scales all text to the input size. I assumed based on https://api.flutter.dev/flutter/material/TextTheme-class.html that size 14 is default for body text, and visually it looks like if I pass in `?fontSize=14` things are unchanged. 

Examples:

`?fontSize=10`
Screen Shot 2020-10-28 at 2 20 54 PM
Screen Shot 2020-10-28 at 2 56 21 PM

`?fontSize=20
Screen Shot 2020-10-28 at 2 21 23 PM
Screen Shot 2020-10-28 at 2 55 58 PM

### DIFF
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -28,7 +28,7 @@ class CodeView extends StatefulWidget {
     this.onSelected,
   }) : super(key: key);
 
-  static const rowHeight = 20.0;
+  static double get rowHeight => 20.0 * fontSizeFactorFromTheme();
   static const assumedCharacterWidth = 16.0;
 
   final DebuggerController controller;
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -29,7 +29,7 @@ class CodeView extends StatefulWidget {
   }) : super(key: key);
 
   static double get rowHeight => 20.0 * fontSizeFactorFromTheme();
-  static const assumedCharacterWidth = 16.0;
+  static double get assumedCharacterWidth => 16.0 * fontSizeFactorFromTheme();
 
   final DebuggerController controller;
   final ScriptRef scriptRef;
--- a/packages/devtools_app/lib/src/inspector/inspector_tree.dart
+++ b/packages/devtools_app/lib/src/inspector/inspector_tree.dart
@@ -44,6 +44,14 @@ const double iconPadding = 5.0;
 const double chartLineStrokeWidth = 1.0;
 const double columnWidth = 16.0;
 
+double verticalPadding() {
+  return 10.0 * fontSizeFactorFromTheme();
+}
+
+double rowHeight() {
+  return 24.0 * fontSizeFactorFromTheme();
+}
+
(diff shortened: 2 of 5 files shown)


kenzieschmoll:
> Line heights, e.g. inspector_tree.dart seems to use rowHeight = 24.0; - maybe there's something related to font size scaling that I could access from theme to fix? Not sure.

Instead of using rowHeight consts like 24.0, could you do `fontSize + 8.0` or something like that?

packages/devtools_app/lib/src/theme.dart
```diff
@@ -9,20 +9,43 @@ import 'common_widgets.dart';
...
 import 'ui/theme.dart';
 
+const DEFAULT_FONT_SIZE = 14.0;
+
 /// Constructs the light or dark theme for the app taking into account
 /// IDE-supplied theming.
 ThemeData themeFor({
   @required bool isDarkTheme,
   @required IdeTheme ideTheme,
+  BuildContext context,
```
kenzieschmoll, line 19: instead of passing in the BuildContext, let's pass in the theme, since it seems we are only using context to access `Theme.of(context)`

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -28,7 +28,7 @@ class CodeView extends StatefulWidget {
     this.onSelected,
   }) : super(key: key);
 
-  static const rowHeight = 20.0;
+  static double get rowHeight => 20.0 * fontSizeFactorFromTheme();
   static const assumedCharacterWidth = 16.0;
```
jacob314, line 32: this const also needs to change.

packages/devtools_app/lib/src/inspector/inspector_tree.dart
```diff
@@ -42,8 +43,14 @@ extension InspectorColorScheme on ColorScheme {
...
 const double chartLineStrokeWidth = 1.0;
 const double columnWidth = 16.0;
-const double verticalPadding = 10.0;
-const double rowHeight = 24.0;
+
+double verticalPadding() {
+  return 10.0 * fontSizeFactorFromTheme();
+}
+
+double rowHeight() {
```
jacob314, line 51: these should be getters not methods. 
Only use a method for something that is conceptually a getter if the operation performed is very slow.

packages/devtools_app/lib/src/theme.dart
```diff
@@ -10,13 +10,18 @@ import 'config_specific/ide_theme/ide_theme.dart';
 import 'ui/theme.dart';
 
 const DEFAULT_FONT_SIZE = 14.0;
+var fontSizeFactor = 1.0;
```
jacob314, line 13: this data should not be stored as a top level field like this as that could easily cause problems due to being fragile.
For example, now the themeFor method has a surprising side effect.
It is better to use the font sizes froom the existing textTheme part of a theme.

helin24: I was hoping to use the theme to find the font size, but it looked like this would mean passing it through a lot of logic for generating inspector rows. Is there any way around doing that?

helin24:
Closing in favor of https://github.com/flutter/devtools/pull/3054

PR #2460, opened 2020-10-25
set up more github actions CI jobs

devoncarew:
- set up more github actions CI jobs

WIP - I'll likely re-factor the config file a bit.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -19,7 +19,7 @@ env:
 jobs:
 
   main:
-    name: Main (${{ matrix.channel }} channel)
+    name: main (${{ matrix.channel }} channel)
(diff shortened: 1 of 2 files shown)


DanTup:
The test_ddc failure seems to be the same on Travis, so I don't think that's CI-related:

https://travis-ci.org/github/flutter/devtools/jobs/740100188#L489

Some of the failures I checked for the other (for ex. "toggle numeric service extension") don't seem like the environment should matter, so my first thought would be a race somewhere. I'd probably try temporarily committing a delay before the failing expectations and see if they start passing.

devoncarew (commented):
@kenzieschmoll, this is ready for review (although I had to disable two of the bots).

kenzieschmoll (approved):

PR #2462, opened 2020-10-26
Add average FPS info to the timeline

kenzieschmoll:
Screen Shot 2020-10-27 at 3 26 23 PM

### DIFF
--- a/packages/devtools_app/lib/src/timeline/flutter_frames_chart.dart
+++ b/packages/devtools_app/lib/src/timeline/flutter_frames_chart.dart
@@ -214,7 +214,8 @@ class _FlutterFramesChartState extends State<FlutterFramesChart>
         (sum, frame) =>
             sum + math.max(frame.uiDurationMs, frame.rasterDurationMs));
     final avgFrameTime = sumFrameTimesMs / widget.frames.length;
-    final avgFps = (1 / avgFrameTime * 1000).round();
+    final avgFps =
+        math.min(widget.displayRefreshRate, 1 / avgFrameTime * 1000).round();
     return Text(
       '$avgFps FPS (average)',
       maxLines: 2,
--- a/packages/devtools_app/lib/src/timeline/flutter_frames_chart.dart
+++ b/packages/devtools_app/lib/src/timeline/flutter_frames_chart.dart
@@ -219,8 +219,7 @@ class _FlutterFramesChartState extends State<FlutterFramesChart>
           ),
     );
     final avgFrameTime = sumFrameTimesMs / widget.frames.length;
-    final avgFps =
-        math.min(widget.displayRefreshRate, 1 / avgFrameTime * 1000).round();
+    final avgFps = (1 / avgFrameTime * 1000).round();
     return Text(
       '$avgFps FPS (average)',
       maxLines: 2,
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/timeline/flutter_frames_chart.dart
```diff
@@ -198,6 +207,19 @@ class _FlutterFramesChartState extends State<FlutterFramesChart>
       ],
     );
   }
+
+  Widget _buildAverageFps() {
+    final double sumFrameTimesMs = widget.frames.fold(
```
jacob314, line 212: Fyi @InMatrix  @cobblest 
Will this estimate of average FPS confuse users?
This calculation could report over 60FPS even when the app really rendered at under 60fps and could report far higher than 60FPS for some cases where the app only rendered at 60fps.
Some ideas:
1. Cap the reported frame rate to the target frame rate for the device so we never report above 60FPS when we were only trying to render at 60fps.
2. Change from FPS to "Average time to render frame" or similar so there is less of a claim that this was the frame rate achieved.

InMatrix: My intuition is leaning towards option #1. Running at 60fps is an idea that seems to be synonymous to being performant for better or worse. I think developers would want to keep an eye on whether the app renders at 60fps or not rather than specific frame rates. Unless the device can render at 90fps/120fps, for example, specific numbers above 60fps don't seem to be useful.

kenzieschmoll: Capped the FPS at the target rate 👍

packages/devtools_app/lib/src/timeline/flutter_frames_chart.dart
```diff
@@ -198,6 +207,20 @@ class _FlutterFramesChartState extends State<FlutterFramesChart>
...
+        0.0,
+        (sum, frame) =>
+            sum + math.max(frame.uiDurationMs, frame.rasterDurationMs));
```
jacob314, line 215: if you take the max of
math.max(frame.uiDurationMs, frame.rasterDurationMs, 1000/averageFrameTime)
you will never pretend a frame ran faster than the target frame time. That will also ensure we never report above the target frame rate and ensure that we don't claim you hit 60fps even though you had one frame that took 400ms and then a bunch of frames that were very quick.

kenzieschmoll: What about the scenario where you have 1000 frames that run in "16.6 ms" (adjusted for this calculation) and a single frame that runs in 18 ms. Then the "average" FPS would be slower than 60 even though one frame barely missed the budget. Perhaps we should remove the word "average" from the label if we are going to adjust times in the calculation?

jacob314: If one frame missed the budget I'm happy if the frame rate is slightly bellow 60fps and the app didn't hit 60fps.
Another way to think about it is
One frame at 18ms and one frame at 14ms is a lot worse than
Two frames both at 16ms.

kenzieschmoll: I see. Addressed your comments.

jacob314 (approved):
lgtm

PR #2466, opened 2020-10-26
Call out failed network requests in the network profiler

kenzieschmoll:
Screen Shot 2020-10-26 at 3 23 37 PM

Fixes https://github.com/flutter/devtools/issues/2353

### DIFF
--- a/packages/devtools_app/lib/src/http/http_request_data.dart
+++ b/packages/devtools_app/lib/src/http/http_request_data.dart
@@ -192,6 +192,23 @@ class HttpRequestData extends NetworkRequest {
     return connectionInfo != null ? connectionInfo[_localPortKey] : null;
   }
 
+  @override
+  bool get didFail {
+    if (status == null) return false;
+    try {
+      final code = int.parse(status);
+      // Status codes 400-499 are client errors and 500-599 are server errors.
+      if (code >= 400) {
+        return true;
+      }
+    } on Exception catch (_) {
+      if (status == 'Error') {
+        return true;
+      }
+    }
+    return false;
+  }
+
   /// True if the HTTP request hasn't completed yet, determined by the lack of
   /// an end event.
   bool get inProgress => _endEvent == null;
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/network/network_model.dart
```diff
@@ -83,6 +85,10 @@ class WebSocket extends NetworkRequest {
   @override
   int get port => _socket.port;
 
+  // TODO(kenz): what determines a web socket request failure?
+  @override
+  bool get didFail => false;
```
kenzieschmoll, line 90: @bkonyi do we get information from web socket requests that signal success or failure? I'm not seeing anything that would tell us this in the `SocketStatistic` class

bkonyi: I don't think the socket profiling covers `WebSocket`, just the `dart:io` `Socket` classes.

kenzieschmoll: The WS examples in the screenshot are created via `IOWebSocketChannel.connect('ws://echo.websocket.org').sink.add('example message')`, and those are pulled from the socket profile.

bkonyi: Oh I thought they were pulled from the HTTP profiler. I'll have to take a look at the web socket implementation first, but I'll get back to you on this. I'm 99% sure the web socket implementation is built on top of `Socket` at some level, so I doubt there's any special information about success / failure for web sockets that isn't already available for sockets.

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -192,6 +192,23 @@ class HttpRequestData extends NetworkRequest {
...
+  bool get didFail {
+    if (status == null) return false;
+    try {
+      final code = int.parse(status);
+      // Status codes 400-499 are client errors and 500-599 are server errors.
+      if (code >= 400) {
+        return true;
+      }
+    } on Exception catch (_) {
+      if (status == 'Error') {
```
jacob314, line 205: handle this before the int.parse. DevTools is good at avoiding throwing exceptions in normal operation cases and it would be nice to keep it that way.

jacob314 (approved):
lgtm

PR #2470, opened 2020-10-29
Ensure app-size has unique URLs for standalone/embedded

### LINKED ISSUE #2468: [new navigator] - when navigating to the app size tab, the stand alone app size page is rendered
navigating to the app size tab should render the app size page in the context of the entire devtools app (e.g. tab bar, footer, buttons in top right should all still be present). 

Instead, navigating to the app size tab navigates to the stand alone app size route that is used when navigating to the app size tool from the landing screen

@DanTup

DanTup:
Fixes #2468.

@kenzieschmoll

### DIFF
--- a/packages/devtools_app/lib/src/app_size/app_size_screen.dart
+++ b/packages/devtools_app/lib/src/app_size/app_size_screen.dart
@@ -40,7 +40,7 @@ class AppSizeScreen extends Screen {
   ///
   /// This must be different to the top-level appSizePageId which is also used
   /// in routing when to ensure they have unique URLs.
-  static const id = 'appsize';
+  static const id = 'app-size';
 
   @visibleForTesting
   static const dropdownKey = Key('Diff Tree Type Dropdown');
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/app_size/app_size_screen.dart
```diff
@@ -35,13 +35,19 @@ class AppSizeScreen extends Screen {
 
   static const analysisTabKey = Key('Analysis Tab');
   static const diffTabKey = Key('Diff Tab');
-  static const id = 'app-size';
+
+  /// The ID (used in routing) for the tabbed app-size page.
+  ///
+  /// This must be different to the top-level appSizePageId which is also used
+  /// in routing when to ensure they have unique URLs.
+  static const id = 'appsize';
```
kenzieschmoll, line 43: does this break the documentation link (bottom right of devtools)? If so we should change the other id

DanTup: Yeah, it did :(

I've swapped them around (another option would've been to set `docId`, though this seemed simpler). Thanks!

kenzieschmoll (approved):

DanTup:
Looks like Travis is on vacation. The bots were green for the run before that last commit, so if you're keen to get the fix it's probably safe to land this (famous last words..).

Issue #2391, opened 2020-09-30
Source is not displayed in Timeline when profiling app


sceee:
I am executing a flutter app in profile mode using the following VS Code config:^
```
{
            "name": "Profile App",
            "program": "lib/main.dart",
            "request": "launch",
            "type": "dart",
            "flutterMode": "profile"
}
```

Then, when having a look at the Timeline in Dart DevTools, I unfortunately do not see the source:
![image](https://user-images.githubusercontent.com/11340487/94692441-d0b98780-0332-11eb-95fb-dd8bab279bcf.png)

It's empty for the complete stack and it's empty in the browser as well as in the embedded VS Code view.

I'm actually not sure if the app is executed correctly in profile mode as it seems I cannot see any of my own code in the stack. But this could also be because I am not yet very experienced in using the profiler in DevTools. I just execute the app using the above mentioned profile launch config and then open the DevTools/Timeline, make some action in the app and then click refresh.

Versions are the following:
* Device info from DevTools: 
![image](https://user-images.githubusercontent.com/11340487/94692851-550c0a80-0333-11eb-8dcf-cda1dbfbcc56.png)
* Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel stable, 1.20.4, on Microsoft Windows [Version 10.0.18363.1082], locale de-DE)
[√] Android toolchain - develop for Android devices (Android SDK version 29.0.2)
[√] Android Studio (version 4.0)
[√] VS Code (version 1.49.2)
[√] Connected device (2 available)
* DevTools version 0.9.2

kenzieschmoll:
We don't currently have file URIs in profile mode - this is due to https://github.com/dart-lang/sdk/issues/37553. I will ping that issue.

kenzieschmoll:
Closing this issue - the source column has been removed from the CPU profiler for flutter apps (see https://github.com/flutter/devtools/pull/2471). It can be added back once https://github.com/dart-lang/sdk/issues/37553 is fixed.

PR #2471, opened 2020-10-29
Remove Source column from cpu profiler for flutter apps

kenzieschmoll:
We do not have stack frame source information in profile mode, and we recommend that users profile flutter apps in profile mode. Users have filed issues about the "missing" source information (see https://github.com/flutter/devtools/issues/2391), but we do not have this information to show (see https://github.com/dart-lang/sdk/issues/37553)


bkonyi (approved):
LGTM

PR #2473, opened 2020-10-30
Prep for 0.9.4 release and bump vm_service to 5.4.0

kenzieschmoll:


packages/devtools_shared/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: Package of shared structures between devtools_app and devtools_serv
 
 # Note: this version should only be updated by running tools/update_version.sh
 # that updates all versions of packages from packages/devtools.
-version: 0.9.3+4
+version: 0.9.4
```
jacob314, line 6: while you are here can you update the intl dependency?

jacob314 (approved):
lgtm

packages/devtools_shared/pubspec.yaml
kenzieschmoll: I can't bump the intl dependency to `0.17.0-nullsafety.1` unless mp_chart bump's its intl dep to `0.17.0-nullsafety.1` and is published. @terrylucas is working on this. I'll hold off on landing this until mp_chart is published and we can depend on the new version.

PR #2481, opened 2020-11-02
Restore socket profiling state after hot restart

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/1968

This CL also enables network recording by default once the network page is opened and changes the buttons from record/stop to pause/resume.

### DIFF
--- a/packages/devtools_app/lib/src/network/network_controller.dart
+++ b/packages/devtools_app/lib/src/network/network_controller.dart
@@ -189,7 +189,7 @@ class NetworkController with SearchControllerMixin<NetworkRequest> {
   void _updatePollingState(bool recording) {
     if (recording) {
       _pollingTimer ??= Timer.periodic(
-        const Duration(milliseconds: 500),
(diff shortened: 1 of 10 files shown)


packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -191,24 +186,10 @@ class NetworkController with SearchControllerMixin<NetworkRequest> {
...
-    _recordingNotifier.value = state;
-  }
-
-  void updatePollingState(bool recording) {
-    if (recording && _countActiveClients > 0) {
+  void _updatePollingState(bool recording) {
+    if (recording) {
       _pollingTimer ??= Timer.periodic(
-        const Duration(milliseconds: 500),
+        const Duration(milliseconds: 2000),
```
jacob314, line 192: can you comment why this is now 2 seconds?

jacob314 (approved):
lgtm

Issue #1982, opened 2020-05-18
DevTools Logging Page: UX Punch List 2020


raison00:
##
## Update Overall UX 
- [x] Add header for 'details' section
 - [x] Use consistent borders matching other DevTools pages
- [x] Create consistent search / filter bindings matching other DevTools pages
- [x] Use DevTools standard table
 - [x] ~Unify Search & Filter Expectations~ 
 - [x] **Update/Add unified color themes**
 - - [x] Rows need alternate colors 
 - - [x] Distinguish color on hover `#36c6f4`

#### Current ( DevTools 0.2.5 Logging Page)
![image](https://user-images.githubusercontent.com/8105802/82244228-c3f14580-98f5-11ea-9e8d-ea6ed782f11e.png)

![image](https://user-images.githubusercontent.com/8105802/82244084-8096d700-98f5-11ea-9afd-e79b6198b64b.png)

:sparkles:

devoncarew:
@raison00, are 'Create consistent search / filter bindings matching other DevTools pages' and  'Unify Search & Filter Expectations' the same item?

And, wrt the 'Search' and 'Filter' nomenclature, have you thought about what to name these fields? I believe we have a search/filter text filed in the device info dialog, the logging page, the debugger scripts view, and the memory snapshot view. We'll likely get search boxes in the timeline page and possibly the inspector page.

Here are some thoughts, but please add your own:
- we call the field 'Filter' when it is reducing the number of items shown
- where space is available, when the filter is non-empty, we should text near it indicating the number of matches (`showing 7 or 423 items`)
- where the field is matching items but not hiding any from display, we call the field 'Search'
- we might consider (where space permits), indicate the number of matches ('4 matches', or if we can toggle through the selections, '2 of 4 matches')

raison00:
We're on the same page;  I have been looking at the holistic approach and familiar conventions.
I will update/expand/etc the various pages/requirements.  

To parse it down:
**Filtering** takes an existing full list and removes items based on criteria that matches/doesn't match.

**Search** takes a blank slate and adds to it based on criteria that matches/doesn't match.

devoncarew:
> **Search** takes a blank slate and adds to it based on criteria that matches/doesn't match.

Yes, or possibly, advances through the matches in the content being searched.

raison00:
The Punch List has been updated for Filter / Search for DevTools.  The outlier is the Memory Profiler, and will be addressed separately.

[...]

PR #2493, opened 2020-11-06
Add search and filter to the logging page; refactor filter code

kenzieschmoll:
This adds a search and filter experience consistent with what is on other pages of DevTools (timeline, network, performance), and refactors some of the filter code to be reusable across devtools.

Screen Shot 2020-11-05 at 3 25 45 PM

Screen Shot 2020-11-06 at 4 06 42 PM

Addresses part of https://github.com/flutter/devtools/issues/1982

### DIFF
--- a/packages/devtools_app/lib/src/logging/logging_controller.dart
+++ b/packages/devtools_app/lib/src/logging/logging_controller.dart
@@ -165,12 +165,12 @@ class LoggingController
 
   static const kindFilterId = 'logging-kind-filter';
 
-  final _filterArgs = [
-    FilterArgument(id: kindFilterId, keys: ['kind', 'k']),
-  ];
+  final _filterArgs = {
+    kindFilterId: FilterArgument(keys: ['kind', 'k']),
+  };
 
   @override
-  List<FilterArgument> get filterArgs => _filterArgs;
+  Map<String, FilterArgument> get filterArgs => _filterArgs;
 
   /// Listen on a stream and track the stream subscription for automatic
   /// disposal if the dispose method is called.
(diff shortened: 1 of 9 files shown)


packages/devtools_app/lib/src/logging/logging_controller.dart
```diff
@@ -635,6 +609,58 @@ class LoggingController {
...
+    return matches;
+  }
+
+  @override
+  void filterData(QueryFilter filter) {
+    if (filter == null) {
+      filteredData.value = List.from(data);
+    } else {
+      filteredData.value = data.where((log) {
+        for (final arg in filter.filterArguments) {
```
jacob314, line 637: maybe process the filterArugments into a `Map<filterId,  List<Filter>>` so this logic is a little cleaner and slightly more efficient for the case of a really long log.

packages/devtools_app/lib/src/logging/logging_controller.dart
```diff
@@ -635,6 +609,58 @@ class LoggingController {
...
+        if (filter.substrings.isNotEmpty) {
+          for (final substring in filter.substrings) {
+            final caseInsensitiveSubstring = substring.toLowerCase();
+            final matchesKind = log.kind != null &&
+                log.kind.toLowerCase().contains(caseInsensitiveSubstring);
+            final matchesSummary = log.summary != null &&
+                log.summary.toLowerCase().contains(caseInsensitiveSubstring);
+            final matchesDetails = log.details != null &&
+                log.summary.toLowerCase().contains(caseInsensitiveSubstring);
+            if (matchesKind || matchesSummary || matchesDetails) {
```
jacob314, line 653: nit: why not just return immediately at the point each of these is true rather than computing them all and then returning?

packages/devtools_app/test/logging_controller_test.dart
```diff
@@ -59,36 +69,70 @@ void main() {
...
+      addGcData('gc1');
+      addGcData('gc2');
 
-      expect(controller.data, hasLength(3));
-      expect(controller.filteredData, hasLength(3));
+      expect(controller.data, hasLength(5));
+      expect(controller.filteredData.value, hasLength(5));
 
-      controller.filterText = 'abc';
+      controller.filterData(QueryFilter.parse('abc', controller.filterArgs));
```
jacob314, line 111: All the tests for QueryFilter look like they take a single argument. Should we have some tests with multiple arguments? In particular, multiple negative arguments seems relevant.

kenzieschmoll: added more tests for logging_controller_test and network_controller_test.

jacob314 (approved):
lgtm

escamoteur:
Awesome work!

PR #2498, opened 2020-11-10
New chart

terrylucas:
New charting for DevTools to replace usage of MP_FlutterChart. Below is the README.md (also included in the PR).

## Overview

DevTools has used a number of existing charting packages with various issues e.g.,
* rich set of charts, more than DevTools would produce
* complex abstraction to support rich set of charts
* not extensible stuff
* not flexible enough without hacky code to augment
* not Dart or Flutter compliant

This charting subsystem plots [timeseries data](https://en.wikipedia.org/wiki/Time_series).  Two basic charts are supported:
1. line chart to display non-numeric values e.g., events
1. scatter chats with or without shading from 0 to their Y coordinate.

## The Data
A chart consist of one or more traces. A trace is a collection of data to be plotted in the same chart on the same temporal X-axis (time).  For example, a chart could have 4 traces:
1. current capacity of the heap (free/used) at every sample period (time)
1. amount of heap used at every sample period (time)
1. number of objects in the heap at every sample period (time)
1. when a GC (garbage collection) occurred.

In all cases, the above 4 traces share the same X-Axis time scale.  However, the first two traces have numeric data a real value in total bytes, the third is number of objects and the last is when a GC occurred.  There are a few ways to display these seemly disparate pieces of data.

## Charts

First, create a StatefulWidget to contain the chart e.g.,
```
class MyChart extends StatefulWidget {
  const MyChart({Key key}) : super(key: key);

  @override
  State<StatefulWidget> createState() => MyChartState(ChartController());
}

class MyChartState extends State<MyChart> {
  MyChartState(this.controller);

  ChartController controller;

}
```
Then, override the State's initState and build methods in MyChartState e.g.,

```
  @override
  void initState() {
    super.initState();

    // If you're creating a line chart then fix the Y-range. For a scatter chart
    // don't create a fixed range, then the chart will autmatically compute the
    // min/max range and rescale the axis.
    controller.setFixedYRange(0.4, 2.4);            // Line chart fixed.

    // Create the traces to hold each set of data points e.g., temperature
    // collected every hour as well as barametric pressure collected every hour.
    setupTraces();
  }

  @override
    final ballChart = Chart(controller, 'Balls Chart');

    return Padding(
      padding: const EdgeInsets.all(0.0),
      child: ballChart,
    );
  }
```
## Create a Trace
Create each trace, in this example we're creating a red circle and blue ball to appear at particular points in time. The Y coordinate of a datum (for a line chart) is not significant other than where to place the event in time.  For a scatter chart, the datum's Y coordinate would be significant e.g., temperature. For both the line and scatter charts the X coordinate is the time. Important note a trace contains the specific plotting characteristers e.g., color, shape, connected line, etc.

This implies that even though different datum could be contained in a single trace each datum would need to hold its rendering characteristics when a datum is rendered in the trace. Considering that a trace may contain many thousands (tens of thousands) of pieces of data that would create unnessary datum overhead as well as more complex painting inside the canvas code when rendering different datum in a trace. The internal chart rendering instead, blasts each trace's data with the same trace's monolithic rendering characterstics.
```
  void setupTraces() {
    // Green Circle
    greenTraceIndex = controller.createTrace(
      ChartType.symbol,
      PaintCharacteristics(
        color: Colors.green,
        strokeWidth: 4,
        diameter: 6,
        fixedMinY: 0.4,
        fixedMaxY: 2.4,
      ),
      name: 'Green Circle',
    );

    // Small Blue Ball
    blueTraceIndex = controller.createTrace(
      ChartType.symbol,
      PaintCharacteristics(
        color: Colors.blue,
        symbol: ChartSymbol.disc,
        diameter: 4,
        fixedMinY: 0.4,
        fixedMaxY: 2.4,
      ),
      name: 'Blue Ball',
    );

    // Red Circle
    redTraceIndex = controller.createTrace(
      ChartType.symbol,
      PaintCharacteristics(
        color: Colors.red,
        strokeWidth: 4,
        diameter: 6,
        fixedMinY: 0.4,
        fixedMaxY: 2.4,
      ),
      name: 'Red Circle',
    );
  }
```
## Adding Data
Create a datum (see Data class) and add the data to a trace (adddatum). In the below example, a datum is created first and added to the green trace (green circle), every second a datum is added to the blue trace (blue ball) then after 30 seconds the last datum is added to the red trace (red circle).
```
  static const greenPosition = 0.4;    // Starting symbol position
  static const bluePosition = 1.4;     // Every second symbol position
  static const redPosition = 2.4;      // Stoping symbol position

  var previousTime = DateTime.now();
  final startDatum = Data(previousTime.millisecondsSinceEpoch, greenPosition);

  // Start the heartbeat.
  controller.timestamps.add(startDatum.timestamp);
  
  // Add the first real datum.
  controller.trace(greenTraceIndex).addDatum(startDatum);

  Timer.periodic(
    const Duration(seconds: 1),
    (Timer timer) {
      if (controller.trace(blueTraceIndex).data.length < 30) {
        final currentTime = DateTime.now();
        final  = currentTime.millisecondsSinceEpoch;

        // Once a second heartbeat.
        controller.timestamps.add(timestamp);

        // Add the blue ball.
        final datum = Data(timestamp, bluePosition);
        controller.trace(blueTraceIndex).addDatum(datum);

        previousTime = currentTime;
      } else {
        controller.timestamps.add(previousTime.millisecondsSinceEpoch);
        final stopDatum = Data(
          previousTime.millisecondsSinceEpoch,
          redPosition,
        );
        // Last datum is the red circle.
        controller.trace(redTraceIndex).addDatum(stopDatum);

        timer.cancel();
      }
    },
  );
```
Each call to addDatum will 
1. notifies the chart that new data has arrived.
1. compute the Y-axis scale (if not fixed)
1. chart the data
1. update the X-axis.

 One important piece of information is adding a heartbeat. If a live chart is needed where the timeline (X axis) moves in a linear fashion requires adding a heart beat (at the grandularity requested of the X axis) a tickstamp is added to the ChartController timestamps field on every tick e.g.,

```
            controller.timestamps.add(currentTime.millisecondsSinceEpoch)
```
The heartbeat allows the data to be replayed as if the data collection is live, at the same rate of experiencing the collecting of the live data.


packages/devtools_app/lib/src/charts/README.md
```diff
@@ -0,0 +1,157 @@
+## DevTools Charting Package
```
jacob314, line 1: This isn't a "package". Maybe note here that this README.md exists because we might make it a package in the future.

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -0,0 +1,588 @@
...
+  ChartState createState() => ChartState();
+}
+
+class ChartState extends State<Chart> with AutoDisposeMixin {
+  ChartController controller;
+
+  @override
+  void initState() {
+    super.initState();
+    controller = widget.controller;
```
jacob314, line 32: don't cache the controller at all or update the cached version when the widget changes.

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -0,0 +1,588 @@
...
+  }
+
+  @override
+  void dispose() {
+    controller.dispose();
+    super.dispose();
+  }
+
+  @override
+  void didChangeDependencies() {
```
jacob314, line 42: this should be didUpdateWidget not didUpdateDependencies. It seems like the state is changing based on the widget not some inherited widget.

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -0,0 +1,588 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+      padding: const EdgeInsets.symmetric(horizontal: 5, vertical: 10),
+      child: LayoutBuilder(
+        // Inner container
+        builder: (_, constraints) => Container(
+          width: constraints.widthConstraints().maxWidth,
+          height: constraints.widthConstraints().maxHeight,
+          color: Colors.black,
```
jacob314, line 63: why black? This should be from the theme or we shouldn't specify a background.

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -0,0 +1,588 @@
...
+
+  // TODO(terry): Use drawText?
+  void drawTitle(Canvas canvas, Size size, String title) {
+    final tp = createText(title, 1.5);
+    tp.paint(canvas, Offset(size.width / 2 - tp.width / 2, 0));
+  }
+
+  void drawAxes(Canvas canvas, Size size, Paint axis) {
+    // Top line of chart.
+    canvas.drawLine(
```
jacob314, line 319: I'm seeing a lot of +xCanvasChart and +yCanvasChart here
you can 
write
https://api.flutter.dev/flutter/dart-ui/Canvas/translate.html
`canvas.translate(xCanvasChart, yCanvasChart)`
to simplify this math.

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -0,0 +1,588 @@
...
+
+  /// Separated out from drawAxis because we don't know range until plotted.
+  // TODO(terry): Consider combining into drawAxis function.
+  void drawYTicks(Canvas canvas, ChartController chartController, Paint axis) {
+    final xCoord = chartController.xCanvasChart;
+
+    final yScale = chartController.yScale;
+
+    for (var labelIndex = yScale.labelTicks; labelIndex >= 0; labelIndex--) {
+      final unit = pow(10, yScale?.labelUnitExponent).toInt();
```
jacob314, line 370: maybe make this .floor() to make it clear that you intentionally want to round down?

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -0,0 +1,588 @@
...
+      final y = labelIndex * unit;
+      final yCoord = chartController.yPositonToYCanvasCoord(y);
+
+      final labelName = constructLabel(
+        labelIndex.toInt(),
+        yScale.labelUnitExponent.toInt(),
+      );
+
+      drawText('$labelName', canvas, 5, yCoord);
```
jacob314, line 379: @pq do we have a lint for unneccssary string comprehensions? Would seem easy to have a lint that realized that
`'$labelName'` could just be `labelName` particularly once we have null safety so there isn't a risk the code is needed to ensure we have the string `'null`'

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -0,0 +1,588 @@
...
+      case 3:
+        labelValue = labelValue * pow(10, unitExponent);
+        break;
+      // 10K, 100K
+      case 4:
+      case 5:
+        labelValue = labelValue * pow(10, unitExponent - 4);
+        unit = 'K';
+        break;
+      // 1M, 10M, 100M
```
jacob314, line 404: Nice comments!

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -0,0 +1,588 @@
...
+        unit = 'B';
+        break;
+      // 1T, 10T, 100T
+      case 12:
+      case 13:
+      case 14:
+        labelValue = labelValue * pow(10, unitExponent - 12);
+        unit = 'T';
+        break;
+    }
```
jacob314, line 425: is there prior art here? we might want to show 3000T instead of showing 3000000000000000 but probably won't matter in practice

packages/devtools_app/lib/src/charts/chart.dart
pq: https://dart-lang.github.io/linter/lints/unnecessary_string_interpolations.html 👍

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -0,0 +1,588 @@
...
+      Offset(xTickCoord, xAxisYCoord),
+      Offset(xTickCoord, xAxisYCoord + (shortTick ? 2 : 6)),
+      axis,
+    );
+
+    if (displayTime) {
+      final tp = createText(prettyTimestamp(timestamp), 1);
+      tp.paint(
+        canvas,
+        Offset(xTickCoord - 30, xAxisYCoord + 15 - tp.height / 2),
```
jacob314, line 453: I'm guessing 15 and -30 should be in terms of the font size .
Does it make sense to center the x coordinate instead of using -30?

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -0,0 +1,588 @@
...
+  }
+
+  void drawText(String key, Canvas canvas, double x, double y) {
+    final tp = createText(key, 1);
+    tp.paint(canvas, Offset(x, y - tp.height / 2));
+  }
+
+  TextPainter createText(String key, double scale) {
+    final span = TextSpan(
+      style: TextStyle(color: Colors.grey[600]),
```
jacob314, line 465: does this need to be themeable?

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -0,0 +1,588 @@
...
+  ) {
+    final paint = Paint()
+      ..style = PaintingStyle.stroke
+      ..strokeWidth = characteristics.strokeWidth
+      ..color = characteristics.color;
+
+    canvas.drawLine(Offset(x, y), Offset(x + tickWidth, y), paint);
+  }
+
+  void drawFillArea(
```
jacob314, line 538: comment on what kind of area this is filling.

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -0,0 +1,588 @@
...
+    double y0,
+    double y0Bottom,
+    double x1,
+    double y1,
+    double y1Bottom,
+  ) {
+    final paint = Paint()
+      ..style = PaintingStyle.fill
+      ..strokeWidth = characteristics.strokeWidth
+      ..color = characteristics.color.withAlpha(100);
```
jacob314, line 551: withAlpha(100) isn't needed.

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -0,0 +1,588 @@
...
+      ..lineTo(x0, y0Bottom);
+    fillArea.close();
+    canvas.drawPath(fillArea, paint);
+
+    fillArea.reset();
+  }
+
+  @override
+  bool shouldRepaint(ChartPainter oldDelegate) =>
+      true; // percentage != oldDelegate.percentage;
```
jacob314, line 567: implement this method or add a TODO

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -0,0 +1,588 @@
...
+  /// Return the largest Y value in a particular trace if traceIndex is passed or
+  /// all traces if traceIndex is not passed in.
+  double maxValue({int traceIndex}) {
+    var maxValue = 0.0;
+    if (traceIndex == null) {
+      for (var index = 0; index < chartController.traces.length; index++) {
+        final trace = chartController.traces[index];
+        final traceMax =
+            trace.data.reduce((curr, next) => curr.y > next.y ? curr : next).y;
+        if (traceMax > maxValue) maxValue = traceMax;
```
jacob314, line 578: nit: this can just be a `min`

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -0,0 +1,588 @@
...
+      }
+    } else {
+      final trace = chartController.traces[traceIndex];
+      maxValue =
+          trace.data.reduce((curr, next) => curr.y > next.y ? curr : next).y;
```
jacob314, line 583: add a helper to compute the max value from a trace.

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -0,0 +1,449 @@
...
+
+/// Zoom level of chart
+enum displayDuration { live, fiveMinutes, fifteenMinutes, all }
+
+/// Given a durationLabel return the displayDuration.
+displayDuration durationToDisplayDuration(
+  displayDuration currentZoomDuration,
+  Duration duration,
+) {
+  if (currentZoomDuration == displayDuration.live) {
```
jacob314, line 21: comment why the values in the tests don't align with the enum entry names.

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -0,0 +1,449 @@
...
+  } else if (currentZoomDuration == displayDuration.fiveMinutes) {
+    if (duration.inMinutes == 3) {
+      return displayDuration.fiveMinutes;
+    }
+  }
+
+  return null;
+}
+
+String prettyTimestamp(int timestamp) {
```
jacob314, line 38: not chart specific. Move this with our other time based helper utils.
@kenzieschmoll  we should also align with how we display times in the timeline page.

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -0,0 +1,449 @@
...
+}
+
+/// Indexes into xAxisLabeledTimestamps.
+const leftLabelIndex = 0;
+const centerLabelIndex = 1;
+const rightLabelIndex = 2;
+
+/*
+                                          xPaddingRight
+                                          |
```
jacob314, line 50: what is the difference between xPaddingRight and rightPadding?

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -0,0 +1,449 @@
...
+  final timestampDT = DateTime.fromMillisecondsSinceEpoch(timestamp);
+  return intl.DateFormat.Hms().format(timestampDT); // HH:mm:ss
+}
+
+/// Indexes into xAxisLabeledTimestamps.
+const leftLabelIndex = 0;
+const centerLabelIndex = 1;
+const rightLabelIndex = 2;
+
+/*
```
jacob314, line 48: Make this a `///` comment for the class and add a couple sentences to go along with the nice diagram.

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -0,0 +1,449 @@
...
+  final topPadding = 20.0;
+  // TODO(terry): Compute dynamically based on Y-axis lables text width.
+  final leftPadding = 50.0;
+  // Computed minimum right padding.
+  int rightPadding = 25;
+  // TODO(terry): Compute dynamically based on X-axis lables text height.
+  final bottomPadding = 40.0;
+
+  // TODO(terry): Consider computation of each X-axis tick width (scaled).
+  double tickWidth = 10.0;
```
jacob314, line 90: should these be private with getters?

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -0,0 +1,449 @@
...
+
+  int visibleTicks;
+
+  // TODO(terry): For now three labels.  Need better mechanism, some number of labels
+  //              based on x-axis zoom factor and default unit to display for labels e.g.,
+  //              live take the duration (visible) and divide by some nice unit e.g., 10/20
+  //              seconds, 5 minute show 1 minute units, 15 minutes show 5 minute units, etc.
```
jacob314, line 97: should also adjust the # of labels based on how wide the window is.

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -0,0 +1,449 @@
...
+  set rightLabelTimestamp(int timestamp) {
+    _xAxisLabeledTimestamps[rightLabelIndex] = timestamp;
+  }
+
+  final bool displayXLabels;
+  final bool displayYLabels;
+
+  Duration durationLabel;
+
+  // TODO(terry): Duration based on x-axis zoom factor (live,5 min,15 min,etc).
```
jacob314, line 128: nit: missing space after comma

packages/devtools_app/lib/src/charts/chart_controller.dart
jacob314: maybe also add a comment by the left, center, and right label code to make it clear we know it should be rewritten.

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -0,0 +1,449 @@
...
+    // Left-side out of visible range. Slide center and right labels to
+    // left-most and center. Need to eventually recompute right-most when
+    // duration gap is met.
+    leftLabelTimestamp = centerLabelTimestamp;
+    centerLabelTimestamp = rightLabelTimestamp;
+    rightLabelTimestamp = null;
+  }
+
+  // xCanvas coord for plotting data.
+  double xCanvasChart;
```
jacob314, line 218: one convention from Flutter is to generally use Offset instead of having pairs of variables with names that start with x and y.

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -0,0 +1,449 @@
...
+    final visibleCount = numberOfVisibleXAxisTicks;
+    if (totalTimestamps < visibleCount) {
+      final startIndex = visibleCount - totalTimestamps;
+      indexOffset += (startIndex * tickWidth).toInt();
+    }
+
+    return indexOffset;
+  }
+
+  // -1 implies not visible.
```
jacob314, line 438: Make this a proper doc comment.

packages/devtools_app/lib/src/charts/chart_trace.dart
```diff
@@ -0,0 +1,213 @@
...
+class Data {
+  Data(this.timestamp, this.y);
+
+  final int timestamp;
+  final double y;
+}
+
+class PaintCharacteristics {
+  PaintCharacteristics({
+    this.color = Colors.black,
```
jacob314, line 20: defaulting to Colors.black is somewhat at odds with using a color scheme. perhaps make color required?

packages/devtools_app/lib/src/charts/chart_trace.dart
```diff
@@ -0,0 +1,213 @@
...
+    if (datum.y > controller?.yMaxValue) controller?.yMaxValue = datum.y;
+
+    final traceIndex = controller.traceIndex(this);
+
+    // New data has arrived notify listeners this data needs to be plotted.
+    controller.traceChanged.value = TraceNotifier(traceIndex, data.length - 1);
+  }
+}
+
+class TraceNotifier {
```
jacob314, line 94: document what traceIndex and dataIndex are.

packages/devtools_app/lib/src/charts/chart_trace.dart
```diff
@@ -0,0 +1,213 @@
...
+  TraceNotifier(this.traceIndex, this.dataIndex);
+
+  int traceIndex;
```
jacob314, line 97: make these final.

packages/devtools_app/lib/src/charts/chart_trace.dart
```diff
@@ -0,0 +1,213 @@
...
+
+enum ChartSymbol {
+  ring,
+  disc,
+  square,
+  triangle,
+  dashedLine,
+}
+
+double log10(num x) => log(x) / ln10;
```
jacob314, line 114: probably move this to a math utils class.

packages/devtools_app/lib/src/charts/chart_trace.dart
```diff
@@ -0,0 +1,213 @@
...
+
+  final double minPoint, maxPoint;
+
+  final double maxTicks;
+
+  double _computedMin, _computedMax;
+
+  double _tickSpacing;
+
+  double get tickSpacing => _tickSpacing;
```
jacob314, line 129: flutter contention: put field associated with a getter right after the getter.

packages/devtools_app/lib/src/charts/chart_trace.dart
```diff
@@ -0,0 +1,213 @@
...
+    }
+  }
+
+  double _calculateMin() {
+    final floored = (minPoint / _tickSpacing).floor();
+    return floored * _tickSpacing;
+  }
+
+  double _calculateMax() {
+    final ceiled = (maxPoint / _tickSpacing).ceil();
```
jacob314, line 165: nit remove variable ceiled that doesn't clarify anything. The question I have is why the max is intentionally past the maxPoint. A quick comment indicating why that is intentional could be helpful.

packages/devtools_app/lib/src/charts/chart_trace.dart
```diff
@@ -0,0 +1,213 @@
...
+  Map _exponentFraction(double range) {
+    if (range == 0) return {};
+
+    final exponent = (log10(range)).floor().toDouble();
+    final fraction = range / pow(10, exponent);
+
+    return {'exponent': exponent, 'fraction': fraction};
+  }
+
+  double _niceNum(double range, bool round) {
```
jacob314, line 178: Add a comment explaining what this is trying to do and also move it to a utils library.

packages/devtools_app/test/chart_test.dart
```diff
@@ -0,0 +1,931 @@
+import 'dart:io';
+
+@TestOn('vm')
```
jacob314, line 3: why is this test on the vm only?

packages/devtools_app/test/chart_test.dart
```diff
@@ -0,0 +1,931 @@
...
+import 'package:flutter/widgets.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+import 'support/wrappers.dart';
+
+void main() {
+  const windowSize = Size(2225.0, 1000.0);
+
+  final nowDT = DateTime.now();
+  final isUSLocalePST = Platform.localeName == 'en_US' &&
```
jacob314, line 19: can we force the  Platform and time zone instead of having the tests quietly skip if not run in that zone? E.g. it would be a shame if these tests were skipped on the bots.

packages/devtools_app/test/chart_test.dart
```diff
@@ -0,0 +1,931 @@
...
+        expect(yScale.labelTicks, equals(7));
+        for (var labelIndex = yScale.labelTicks;
+            labelIndex >= 0;
+            labelIndex--) {
+          final labelName = ChartPainter.constructLabel(
+            labelIndex.toInt(),
+            yScale.labelUnitExponent.toInt(),
+          );
+
+          switch (labelIndex.toInt()) {
```
jacob314, line 220: nit: instead of writing a switch write
expectedLabels = ['0', '100M',...];
expect(labelName, expectedLabel[labelIndex.toInt()]);

packages/devtools_app/test/chart_test.dart
```diff
@@ -0,0 +1,931 @@
...
+          // Validate the x-axis labels.
+          expect(controller.leftLabelTimestamp, isNotNull);
+          expect(controller.centerLabelTimestamp, isNotNull);
+          expect(controller.rightLabelTimestamp, isNotNull);
+          expect(controller.leftLabelTimestamp, equals(1595682514617));
+          expect(controller.centerLabelTimestamp, equals(1595682534381));
+          expect(controller.rightLabelTimestamp, equals(1595682556248));
+
+          // Only validate if run in a known country, language and timezone.
+          if (isUSLocalePST) {
```
jacob314, line 319: remove this if after hard coding the locale and time zone assuming that is easy.

jacob314: won't the golden image tests also have an issue if run in the wrong time zone?

jacob314 (commented):
done with a round of comments.

packages/devtools_app/lib/src/charts/README.md
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Maybe I'm confused.  This is the how I thought we hooked up first time things.  didUpdateDependencies is guaranteed to be called immediately after initState I use to to connect up listeners.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: No reason, must have something I had played with originally and didn't remove.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Right it should be specified (that would be a passed in theme).

terrylucas: Fixed

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Fixed.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Was missing default case, for now I'll return e notation if we need values over trillion then I'll add those too.

        unit = 'e+$unitExponent';

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Added a few more words. :-)

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Yep done.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Yep, I'll add a TODO only theming is on traces Chart needs characteristics too to pull from the theme.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: I plan to add a light grid below (paper of the chart) that show's movement/point of reference with relation to the x-axis moving (probably only a few vertical lines).

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: return false for now.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart_controller.dart
terrylucas: Its old code originally was using enum then migrated to a Duration no need for enums.

packages/devtools_app/lib/src/charts/chart_controller.dart
terrylucas: They're both used to give space to the labels on the X-axis e.g.,
rightPadding is a constant entire Chart padding on right-side it's a constant (or could be settable) - might be better to used edget inset.  Good to add a TODO here.
xPaddingRight is a computed position of the right-side of the X-axis gives us room to display a label.

packages/devtools_app/lib/src/charts/chart_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart_controller.dart
terrylucas: got idea tickWidth should be private with getter.

Done.

packages/devtools_app/lib/src/charts/chart_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart_trace.dart
terrylucas: Good idea.

packages/devtools_app/lib/src/charts/chart_trace.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart_trace.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart_trace.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart_trace.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart_trace.dart
terrylucas: Done.

packages/devtools_app/test/chart_test.dart
terrylucas: Copied from another test.

packages/devtools_app/test/chart_test.dart
terrylucas: Done.

packages/devtools_app/test/chart_test.dart
terrylucas: I don't think the golden image tests do because the text areas are opaque.

packages/devtools_app/test/chart_test.dart
terrylucas: I'll take a look but the tests ran everywhere except one bot.

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -7,11 +7,26 @@ import 'dart:ui';
 
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
+import 'package:intl/intl.dart' as intl;
 
 import '../auto_dispose_mixin.dart';
 import 'chart_controller.dart';
 import 'chart_trace.dart';
 
+/// Helper function to:
```
jacob314, line 16: nit: make this a proper doc comment. First line should summarize what the method does.
E.g.
`/// Perform some draw operations on a canvas after applying translate.`

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -7,11 +7,26 @@ import 'dart:ui';
...
+///   1. save canvas state
+///   2. translate coordinates
+///   3. draw to the canvas with respect to translation coordinates
+///   4. restore canvas state back to the saved state.
+void translate(Canvas canvas, double x, double y, VoidCallback drawCode) {
```
jacob314, line 21: maybe make 
drawCode take the `canvas` as a parameter rather than being a void callback?

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -20,35 +35,43 @@ class Chart extends StatefulWidget {
   final ChartController controller;
 
   @override
-  ChartState createState() => ChartState();
+  ChartState createState() => ChartState(controller);
 }
 
 class ChartState extends State<Chart> with AutoDisposeMixin {
-  ChartController controller;
+  ChartState(this.controller);
```
jacob314, line 42: State objects should have zero argument constructors. Instead of doing this you should write

`ChartController get controller => widget.controller;`

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -20,35 +35,43 @@ class Chart extends StatefulWidget {
   final ChartController controller;
 
   @override
-  ChartState createState() => ChartState();
+  ChartState createState() => ChartState(controller);
```
jacob314, line 38: Needs to be zero argument. State objects can be reused so this is very dangerous to do.

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -20,35 +35,43 @@ class Chart extends StatefulWidget {
...
+  }
 
-    addAutoDisposeListener(controller.traceChanged, () {
-      setState(() {
-        controller.computeChartArea();
-      });
-    });
+  @override
+  void dispose() {
+    controller.dispose();
```
jacob314, line 73: disposing the controller from the State object seems wrong. Controllers should have lifecycles independent of what widget is rendering them. Why do we need this?

packages/devtools_app/lib/src/charts/chart.dart
jacob314, line 21: maybe call this drawTranslated?

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -275,18 +255,23 @@ class ChartController extends DisposableController
 
   String title;
 
-  displayDuration zoomDuration = displayDuration.live;
+  /// zoomDuration values of:
+  ///     null implies all (default)
+  ///     Duration() imples live
+  ///     Duration(minutes: 5) implies 5 minute interval
+  ///     Duration(minutes: 15) implies 15 minute interval
+  Duration _zoomDuration = const Duration();
```
jacob314, line 263: Seems like live is the default not null.

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -275,18 +255,23 @@ class ChartController extends DisposableController
...
+
+  Duration get zoomDuration => _zoomDuration;
 
   static const oneMinuteInMs = 1000 * 60;
 
-  void setXAxisZoomDuration(Duration duration) {
+  set zoomDuration(Duration duration) {
```
jacob314, line 269: seems surprising that this method can only be called with certain very specific duration values. Comment indicating what the values are or make it more robust.

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -322,20 +307,21 @@ class ChartController extends DisposableController
...
-          : displayDuration.fifteenMinutes;
+      _zoomDuration = duration;
     }
 
     // All tick labels need to be recompted.
     recomputeLabels(refresh: true);
   }
 
   void computeChartArea() {
+    // Check if ready to start coputations?
```
jacob314, line 322: typo in word computations.

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -355,15 +341,16 @@ class ChartController extends DisposableController
     }
   }
 
-  double xPositonToYCanvasCoord(double y) {
-    final zeroPosition = xCanvasChart + canvasChartWidth;
-    return zeroPosition - yScale?.tickFromValue(y);
-  }
+  double get zeroXPosition => xCanvasChart + canvasChartWidth;
```
jacob314, line 344: do we really need these to be relative to xCanvasChart now that we are using a translate?

packages/devtools_app/lib/src/charts/chart_trace.dart
```diff
@@ -175,6 +174,12 @@ class AxisScale {
     return {'exponent': exponent, 'fraction': fraction};
   }
 
+  /// Produce a rounded or first number of the Y-axis scale and its
```
jacob314, line 177: I don't understand this comment. What does "The goal
  /// is to show high water mark (over 7) flips to 10's unit and
  /// it's exponent."
mean?

jacob314 (commented):
Done with the next round of comments. Close to LGTM. Main concern is lifecycle for the controller.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: typedef DrawCodeCallback = void Function(Canvas canvas);
void translate(Canvas canvas, double x, double y, DrawCodeCallback drawCode) 

Done.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: I had this originally, I thought you didn't like this?  I'll put it back to what I had done.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Back to original.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: We don't nor the dispose. Removed.

Also, I have a test app that I use noticed there was a problem with state/controller fixed.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart_controller.dart
terrylucas: Oops comment matches reality.

packages/devtools_app/lib/src/charts/chart_controller.dart
terrylucas: Changed to accept any duration.

packages/devtools_app/lib/src/charts/chart_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart_controller.dart
terrylucas: zeroXPosition is no longer used.

packages/devtools_app/lib/src/charts/chart_trace.dart
terrylucas: /// Produce a whole number for the Y-axis scale and its unit using
  /// the exponent. Goal is to compute the range of values for min, max
  /// and exponent (our unit of measurement e.g., K, M, B, etc.). The axis
  /// labels are displayed in 1s, 10s and 100s e.g., 10M, 50M, 100M or
  /// 2000, 10K, 20K, 30K.
  /// 
  /// @param round if false, chunks the whole number to keep more available
  /// space above the max Y value highpoint to handle future bigger data
  /// values w/o having to rescale too quickly e.g., over 3e+3 displays:
  ///    10K, 20K, 30K, 40K, 50K
  /// This allows new data points >30K to be plotted w/o having to
  /// re-layout new Y-axis scale.

jacob314 (approved):
lgtm

PR #2499, opened 2020-11-10
Add VM Tools screen with initial VM and Isolate statistics tabs

bkonyi:
**VM Developer Mode setting**

Screen Shot 2020-11-10 at 12 13 59 PM

**VM Statistics tab under VM Tools**
Screen Shot 2020-11-10 at 12 14 28 PM

**Isolate Statistics tab under VM Tools**
Screen Shot 2020-11-10 at 12 14 33 PM

**Isolate Statistics tab with `ReceivePort` allocation location displayed**
Screen Shot 2020-11-16 at 4 41 35 PM

Looking for a general architectural / style review from @kenzieschmoll and @jacob314. @rmacnak-google and @a-siva, can you provide feedback on the content?

### DIFF
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -44,15 +44,13 @@ import 'timeline/timeline_controller.dart';
 import 'timeline/timeline_screen.dart';
 import 'ui/service_extension_widgets.dart';
 import 'utils.dart';
+import 'vm_developer/vm_developer_tools_screen.dart';
 
 // Disabled until VM developer mode functionality is added.
 const showVmDeveloperMode = false;
 
 /// Whether this DevTools build is external.
 bool isExternalBuild = true;
-
-// Disabled until VM developer mode functionality is added.
-const showVmDeveloperMode = false;
 
 /// Top-level configuration for the app.
 @immutable
@@ -174,35 +172,42 @@ class DevToolsAppState extends State<DevToolsApp> {
       url: vmServiceUri,
       allowConnectionScreenOnDisconnect: !embed,
       builder: (_) {
-        final tabs = _visibleScreens()
-            .where((p) => embed && page != null ? p.screenId == page : true)
-            .where((p) => !hide.contains(p.screenId))
-            .toList();
-        if (tabs.isEmpty) {
-          return DevToolsScaffold.withChild(
-            child: CenteredMessage(
-                'The "$page" screen is not available for this application.'),
-            ideTheme: ideTheme,
-            analyticsProvider: widget.analyticsProvider,
-          );
-        }
-        return _providedControllers(
-          child: DevToolsScaffold(
-            embed: embed,
-            ideTheme: ideTheme,
-            page: page,
-            tabs: tabs,
-            analyticsProvider: widget.analyticsProvider,
-            actions: [
-              // TODO(https://github.com/flutter/devtools/issues/1941)
-              if (serviceManager.connectedApp.isFlutterAppNow) ...[
-                HotReloadButton(),
-                HotRestartButton(),
-              ],
-              OpenSettingsAction(),
-              OpenAboutAction(),
-            ],
-          ),
+        // Rebuild when VM developer mode is enabled or disabled to update the
+        // visible screens list.
+        return ValueListenableBuilder(
+          valueListenable: widget.preferences.vmDeveloperModeEnabled,
+          builder: (context, value, _) {
+            final tabs = _visibleScreens()
+                .where((p) => embed && page != null ? p.screenId == page : true)
+                .where((p) => !hide.contains(p.screenId))
+                .toList();
+            if (tabs.isEmpty) {
+              return DevToolsScaffold.withChild(
+                child: CenteredMessage(
+                    'The "$page" screen is not available for this application.'),
+                ideTheme: ideTheme,
+                analyticsProvider: widget.analyticsProvider,
+              );
+            }
+            return _providedControllers(
+              child: DevToolsScaffold(
+                embed: embed,
+                ideTheme: ideTheme,
+                page: page,
+                tabs: tabs,
+                analyticsProvider: widget.analyticsProvider,
+                actions: [
+                  // TODO(https://github.com/flutter/devtools/issues/1941)
+                  if (serviceManager.connectedApp.isFlutterAppNow) ...[
+                    HotReloadButton(),
+                    HotRestartButton(),
+                  ],
+                  OpenSettingsAction(),
+                  OpenAboutAction(),
+                ],
+              ),
+            );
+          },
         );
       },
     );
@@ -266,16 +271,21 @@ class DevToolsAppState extends State<DevToolsApp> {
 
   @override
   Widget build(BuildContext context) {
-    return ValueListenableBuilder(
-      valueListenable: widget.preferences.darkModeTheme,
+    return ValueListenableBuilder<bool>(
+      valueListenable: widget.preferences.vmDeveloperModeEnabled,
       builder: (context, value, _) {
-        return MaterialApp.router(
-          title: 'Dart DevTools',
-          debugShowCheckedModeBanner: false,
-          theme: themeFor(isDarkTheme: value, ideTheme: ideTheme),
-          builder: (context, child) => Notifications(child: child),
-          routerDelegate: DevToolsRouterDelegate(_getPage),
-          routeInformationParser: DevToolsRouteInformationParser(),
+        return ValueListenableBuilder<bool>(
+          valueListenable: widget.preferences.darkModeTheme,
+          builder: (context, value, _) {
+            return MaterialApp.router(
+              title: 'Dart DevTools',
+              debugShowCheckedModeBanner: false,
+              theme: themeFor(isDarkTheme: value, ideTheme: ideTheme),
+              builder: (context, child) => Notifications(child: child),
+              routerDelegate: DevToolsRouterDelegate(_getPage),
+              routeInformationParser: DevToolsRouteInformationParser(),
+            );
+          },
         );
       },
     );
@@ -452,8 +462,6 @@ class SettingsDialog extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final preferences = DevToolsApp.of(context).preferences;
-    // Disabled until VM developer mode functionality is added.
-    const showVmDeveloperMode = false;
     return DevToolsDialog(
       title: dialogTitleText(Theme.of(context), 'Settings'),
       content: Column(
@@ -471,12 +479,11 @@ class SettingsDialog extends StatelessWidget {
               listenable: ga.gaEnabledNotifier,
               toggle: ga.setAnalyticsEnabled,
             ),
-          if (showVmDeveloperMode)
-            _buildOption(
-              label: const Text('Enable VM developer mode'),
-              listenable: preferences.vmDeveloperModeEnabled,
-              toggle: preferences.toggleVmDeveloperMode,
-            ),
+          _buildOption(
+            label: const Text('Enable VM developer mode'),
+            listenable: preferences.vmDeveloperModeEnabled,
+            toggle: preferences.toggleVmDeveloperMode,
+          ),
         ],
       ),
       actions: [
@@ -549,6 +556,10 @@ List<DevToolsScreen> get defaultScreens => <DevToolsScreen>[
         const AppSizeScreen(),
         createController: () => AppSizeController(),
       ),
+      DevToolsScreen<VMDeveloperToolsScreen>(
+        const VMDeveloperToolsScreen(),
+        createController: () => null,
+      )
 // Uncomment to see a sample implementation of a conditional screen.
 //      DevToolsScreen<ExampleController>(
 //        const ExampleConditionalScreen(),
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -44,8 +44,8 @@ import 'timeline/timeline_controller.dart';
 import 'timeline/timeline_screen.dart';
 import 'ui/service_extension_widgets.dart';
 import 'utils.dart';
+import 'vm_developer/vm_developer_tools_controller.dart';
 import 'vm_developer/vm_developer_tools_screen.dart';
-import 'vm_developer/vm_developer_tools_screen_controller.dart';
 
 // Disabled until VM developer mode functionality is added.
 const showVmDeveloperMode = false;
@@ -115,8 +115,6 @@ class DevToolsAppState extends State<DevToolsApp> {
 
     _vmDeveloperModeEnabled = preferences.vmDeveloperModeEnabled.value;
     preferences.vmDeveloperModeEnabled.addListener(() {
-      print(
-          'developer mode enabled: ${preferences.vmDeveloperModeEnabled.value}');
       setState(() {
         _vmDeveloperModeEnabled = preferences.vmDeveloperModeEnabled.value;
       });
@@ -569,10 +567,10 @@ List<DevToolsScreen> get defaultScreens => <DevToolsScreen>[
         const AppSizeScreen(),
         createController: () => AppSizeController(),
       ),
-      DevToolsScreen<VMDeveloperToolsScreenController>(
+      DevToolsScreen<VMDeveloperToolsController>(
         const VMDeveloperToolsScreen(),
-        createController: () => VMDeveloperToolsScreenController(),
-      )
+        createController: () => VMDeveloperToolsController(),
+      ),
 // Uncomment to see a sample implementation of a conditional screen.
 //      DevToolsScreen<ExampleController>(
 //        const ExampleConditionalScreen(),
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -51,6 +51,9 @@ const showVmDeveloperMode = false;
 /// Whether this DevTools build is external.
 bool isExternalBuild = true;
 
+// Disabled until VM developer mode functionality is added.
+const showVmDeveloperMode = false;
+
 /// Top-level configuration for the app.
 @immutable
 class DevToolsApp extends StatefulWidget {
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -45,6 +45,7 @@ import 'timeline/timeline_screen.dart';
 import 'ui/service_extension_widgets.dart';
 import 'utils.dart';
 import 'vm_developer/vm_developer_tools_screen.dart';
+import 'vm_developer/vm_developer_tools_screen_controller.dart';
 
 // Disabled until VM developer mode functionality is added.
 const showVmDeveloperMode = false;
@@ -87,6 +88,12 @@ class DevToolsAppState extends State<DevToolsApp> {
   PreferencesController get preferences => widget.preferences;
   IdeTheme get ideTheme => widget.ideTheme;
 
+  bool get isDarkThemeEnabled => _isDarkThemeEnabled;
+  bool _isDarkThemeEnabled;
+
+  bool get vmDeveloperModeEnabled => _vmDeveloperModeEnabled;
+  bool _vmDeveloperModeEnabled;
+
   @override
   void initState() {
     super.initState();
@@ -96,6 +103,22 @@ class DevToolsAppState extends State<DevToolsApp> {
     serviceManager.isolateManager.onSelectedIsolateChanged.listen((_) {
       setState(() {
         _clearCachedRoutes();
+      });
+    });
+
+    _isDarkThemeEnabled = preferences.darkModeTheme.value;
+    preferences.darkModeTheme.addListener(() {
+      setState(() {
+        _isDarkThemeEnabled = preferences.darkModeTheme.value;
+      });
+    });
+
+    _vmDeveloperModeEnabled = preferences.vmDeveloperModeEnabled.value;
+    preferences.vmDeveloperModeEnabled.addListener(() {
+      print(
+          'developer mode enabled: ${preferences.vmDeveloperModeEnabled.value}');
+      setState(() {
+        _vmDeveloperModeEnabled = preferences.vmDeveloperModeEnabled.value;
       });
     });
   }
@@ -172,11 +195,11 @@ class DevToolsAppState extends State<DevToolsApp> {
       url: vmServiceUri,
       allowConnectionScreenOnDisconnect: !embed,
       builder: (_) {
-        // Rebuild when VM developer mode is enabled or disabled to update the
-        // visible screens list.
-        return ValueListenableBuilder(
-          valueListenable: widget.preferences.vmDeveloperModeEnabled,
-          builder: (context, value, _) {
+        // Force regeneration of visible screens when VM developer mode is
+        // enabled.
+        return ValueListenableBuilder<bool>(
+          valueListenable: preferences.vmDeveloperModeEnabled,
+          builder: (_, __, ___) {
             final tabs = _visibleScreens()
                 .where((p) => embed && page != null ? p.screenId == page : true)
                 .where((p) => !hide.contains(p.screenId))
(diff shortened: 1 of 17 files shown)


packages/devtools_app/lib/src/app.dart
```diff
@@ -263,16 +271,21 @@ class DevToolsAppState extends State<DevToolsApp> {
 
   @override
   Widget build(BuildContext context) {
-    return ValueListenableBuilder(
-      valueListenable: widget.preferences.darkModeTheme,
+    return ValueListenableBuilder<bool>(
```
kenzieschmoll, line 274: instead of nesting ValueListenableBuilders here and above, let's just add a listener in initState that listens to the value listenable and stores the value in DevToolsAppState class. You can do that for both the vmDeveloperModeEnabled and darkModeTheme value listenables

packages/devtools_app/lib/src/screen.dart
```diff
@@ -164,5 +171,12 @@ bool shouldShowScreen(Screen screen) {
       return false;
     }
   }
+  if (screen.requiresVmDeveloperMode) {
+    // TODO(bkonyi): this might be a bit brittle. Should have a global getter
+    // somewhere for 'vmDeveloperModeEnabled'.
+    if (!VmServicePrivate.enablePrivateRpcs) {
+      return false;
+    }
```
kenzieschmoll, line 179: we could potentially make PreferencesController a global (see globals.dart and `setGlobal`), calling `setGlobal` at the end of `PreferencesController.init()`

packages/devtools_app/lib/src/app.dart
bkonyi: Done.

packages/devtools_app/lib/src/screen.dart
bkonyi: Good call, didn't realize that was a thing. Done.

a-siva:
Some comments about content in the screens

- in the initial VM screen the VM version is displayed and it states embedder is Flutter, maybe it should also include the version of the embedder (maybe Embedder: Flutter 1.24.x.(dev/beta/stable)
- what is the significance of the Number column in the isolates pane
- the new/old total sizes format (6.0 MB of 16.0 MB probably needs some explanation, use Vs allocated? not sure if the current live value is significant here)
- in the third and fourth screens what is the section "Service Extensions (57)" with a list of names ?

bkonyi:
@a-siva 

> Some comments about content in the screens
> 
> * in the initial VM screen the VM version is displayed and it states embedder is Flutter, maybe it should also include the version of the embedder (maybe Embedder: Flutter 1.24.x.(dev/beta/stable)

This string is provided by the embedder, so we should file an issue against Flutter for this.

> * what is the significance of the Number column in the isolates pane

I'm not sure what you're referring to here. Can you please clarify?

> * the new/old total sizes format (6.0 MB of 16.0 MB probably needs some explanation, use Vs allocated? not sure if the current live value is significant here)

We might be able to add a `(?)` tooltip here. I'll do some digging.

> * in the third and fourth screens what is the section "Service Extensions (57)" with a list of names ?

This is the list of service extensions registered with the isolate. The same list is currently displayed under the isolate summary in Observatory.

a-siva:
> > * what is the significance of the Number column in the isolates pane
> 
> I'm not sure what you're referring to here. Can you please clarify?
> 

In screen 2 under the Isolates pane there are the following columns
Name, Number, NewSpace, OldSpace, Heap

I am referring to that number, I presume it is the 'isolate id'

bkonyi:
> > > * what is the significance of the Number column in the isolates pane
> > 
> > 
> > I'm not sure what you're referring to here. Can you please clarify?
> 
> In screen 2 under the Isolates pane there are the following columns
> Name, Number, NewSpace, OldSpace, Heap
> 
> I am referring to that number, I presume it is the 'isolate id'

Oh, yes, that's the isolate ID. That's the name of the property, but ID would probably be a better column name.

packages/devtools_app/lib/src/app.dart
```diff
@@ -97,6 +105,22 @@ class DevToolsAppState extends State<DevToolsApp> {
...
+    _isDarkThemeEnabled = preferences.darkModeTheme.value;
+    preferences.darkModeTheme.addListener(() {
+      setState(() {
+        _isDarkThemeEnabled = preferences.darkModeTheme.value;
+      });
+    });
+
+    _vmDeveloperModeEnabled = preferences.vmDeveloperModeEnabled.value;
+    preferences.vmDeveloperModeEnabled.addListener(() {
+      print(
```
kenzieschmoll, line 118: nit: remove print statement

packages/devtools_app/lib/src/app.dart
```diff
@@ -171,35 +195,42 @@ class DevToolsAppState extends State<DevToolsApp> {
...
-                HotReloadButton(),
-                HotRestartButton(),
-              ],
-              OpenSettingsAction(),
-              OpenAboutAction(),
-            ],
-          ),
+        // Force regeneration of visible screens when VM developer mode is
+        // enabled.
+        return ValueListenableBuilder<bool>(
```
kenzieschmoll, line 200: can you remove this value listenable builder now? if the routes need to be regenerated, try calling clearCachedRoutes in setState call from the vmDeveloperModeEnabled listener above

packages/devtools_app/lib/src/app.dart
```diff
@@ -544,6 +569,10 @@ List<DevToolsScreen> get defaultScreens => <DevToolsScreen>[
         const AppSizeScreen(),
         createController: () => AppSizeController(),
       ),
+      DevToolsScreen<VMDeveloperToolsScreenController>(
+        const VMDeveloperToolsScreen(),
+        createController: () => VMDeveloperToolsScreenController(),
+      )
```
kenzieschmoll, line 575: nit: trailing comma

packages/devtools_app/lib/src/app.dart
```diff
@@ -544,6 +569,10 @@ List<DevToolsScreen> get defaultScreens => <DevToolsScreen>[
         const AppSizeScreen(),
         createController: () => AppSizeController(),
       ),
+      DevToolsScreen<VMDeveloperToolsScreenController>(
+        const VMDeveloperToolsScreen(),
+        createController: () => VMDeveloperToolsScreenController(),
```
kenzieschmoll, line 574: nit: rename to VMDeveloperToolsController

packages/devtools_app/lib/src/utils.dart
```diff
@@ -71,6 +71,9 @@ String prettyPrintBytes(
   int mbFractionDigits = 1,
   bool includeUnit = false,
 }) {
+  if (bytes == null) {
+    return null;
+  }
```
kenzieschmoll, line 76: should we assert that bytes != null here instead? or perhaps just return the empty string? why is this method called on a null value?

packages/devtools_app/lib/src/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -0,0 +1,121 @@
...
+    this.table,
+  });
+
+  final String title;
+  final List<MapEntry> rowKeyValues;
+  final Widget table;
+
+  @override
+  Widget build(BuildContext context) {
+    return Flexible(
```
kenzieschmoll, line 33: Can you remove the flexible and declare it upstream? It is more intuitive for `Flexible` widgets to be declared as part of their parent Flex widget (row, column, etc.), that way the flex value can be modified and the flex layout of the parent is more clear from the build method

packages/devtools_app/lib/src/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -0,0 +1,121 @@
...
+              for (final row in rowKeyValues)
+                Row(
+                  mainAxisAlignment: MainAxisAlignment.spaceBetween,
+                  children: [
+                    Text(
+                      '${row.key.toString()}:',
+                      style: fixedFontStyle(context),
+                    ),
+                    Text(
+                      row?.value?.toString() ?? '',
```
kenzieschmoll, line 78: perhaps use '--' instead of the empty string

packages/devtools_app/lib/src/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -0,0 +1,121 @@
...
+      ),
+    );
+  }
+}
+
+class VMInfoList extends StatelessWidget {
+  const VMInfoList({
+    @required this.title,
+    this.rowKeyValues,
+    this.table,
```
kenzieschmoll, line 49: are there ever any cases where both of these will be specified? if not let's start by adding an assert `(assert (rowKeyValue == null) != (table == null))`

packages/devtools_app/lib/src/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -0,0 +1,121 @@
...
+    return [
+      for (int i = 0; i < rows.length; ++i)
+        _buildAlternatingRow(context, i, rows[i]),
+    ];
+  }
+
+  Widget _buildAlternatingRow(BuildContext context, int index, Widget row) {
+    return SizedBox(
+      height: defaultRowHeight,
+      child: Material(
```
kenzieschmoll, line 100: instead of using SizedBox, Material, and Padding, just use Container

packages/devtools_app/lib/src/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -0,0 +1,121 @@
...
+          context,
+        ),
+        child: Padding(
+          padding: const EdgeInsets.symmetric(
+            horizontal: defaultSpacing,
+          ),
+          child: Column(
+            mainAxisAlignment: MainAxisAlignment.center,
+            children: [
+              Flexible(
```
kenzieschmoll, line 112: you may be able to remove both of these widgets as well and just have row be a direct child of Container

packages/devtools_app/lib/src/vm_developer/vm_developer_tools_screen_controller.dart
```diff
@@ -0,0 +1,21 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+
+import 'vm_developer_tools_screen.dart';
+
+class VMDeveloperToolsScreenController {
+  ValueListenable<int> get selectedIndex => _selectedIndex;
+  final ValueNotifier<int> _selectedIndex = ValueNotifier<int>(0);
```
kenzieschmoll, line 11: remove type LHS here and below

packages/devtools_app/lib/src/vm_developer/vm_developer_tools_screen_controller.dart
```diff
@@ -0,0 +1,21 @@
...
+
+  void setSelectedIndex(int index) {
```
kenzieschmoll, line 13: rename to selectIndex

packages/devtools_app/lib/src/vm_developer/vm_service_private_extensions.dart
```diff
@@ -0,0 +1,93 @@
...
+/// A wrapper around [VM] which allows for access to VM internal fields.
+class VMPrivateView {
+  VMPrivateView._(this.vm);
+
+  String get embedder => vm.json['_embedder'];
+  String get profilerMode => vm.json['_profilerMode'];
+  int get currentMemory => vm.json['_currentMemory'];
+  int get currentRSS => vm.json['_currentRSS'];
+  int get maxRSS => vm.json['_maxRSS'];
+  int get nativeZoneMemoryUsage => vm.json['_nativeZoneMemoryUsage'];
```
kenzieschmoll, line 30: is there a reason these can't be directly in VMPrivateViewExtension?

packages/devtools_app/lib/src/vm_developer/vm_service_private_extensions.dart
```diff
@@ -0,0 +1,93 @@
...
+/// An internal representation of a thread running within an isolate.
+class Thread {
+  const Thread({
+    @required this.id,
+    @required this.kind,
+    @required this.zoneHighWatermark,
+    @required this.zoneCapacity,
+  });
+
+  static Thread parse(Map<String, dynamic> json) {
```
kenzieschmoll, line 80: use a factory constructor instead

packages/devtools_app/lib/src/vm_developer/vm_statistics_view.dart
```diff
@@ -0,0 +1,295 @@
...
+        return Column(
+          children: [
+            Row(
+              children: [
+                RefreshButton(
+                  onPressed: controller.refresh,
+                ),
+              ],
+            ),
+            Flexible(
```
kenzieschmoll, line 47: use expanded

packages/devtools_app/lib/src/vm_developer/vm_statistics_view.dart
```diff
@@ -0,0 +1,295 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder(
+      valueListenable: controller.refreshing,
+      builder: (context, refreshing, _) {
+        return Column(
+          children: [
+            Row(
+              children: [
+                RefreshButton(
```
kenzieschmoll, line 42: does this need to be in a row? if you take it out of the row, you can tweak the column alignment to put the refresh button where you want it

packages/devtools_app/lib/src/vm_developer/vm_statistics_view.dart
```diff
@@ -0,0 +1,295 @@
...
+  IsolatesPreviewWidget({
+    @required this.controller,
+    this.systemIsolates = false,
+  });
+
+  final ColumnData<Isolate> name = _IsolateNameColumn();
+  final ColumnData<Isolate> number = _IsolateNumberColumn();
+  final ColumnData<Isolate> newSpace = _IsolateNewSpaceColumn();
+  final ColumnData<Isolate> oldSpace = _IsolateOldSpaceColumn();
+  final ColumnData<Isolate> heap = _IsolateHeapColumn();
```
kenzieschmoll, line 263: remove LHS types

kenzieschmoll (commented):
done with first review

packages/devtools_app/lib/src/app.dart
bkonyi: No, that doesn't seem to work.

packages/devtools_app/lib/src/utils.dart
bkonyi: An example is when we first create the VM statistics view and we build the page before we've gotten any data from the service. Returning null here saves us from having to do null checks in multiple places (at least 4 instances in `vm_statistics_view.dart`). I'm happy to change the approach if you have any suggestions.

packages/devtools_app/lib/src/vm_developer/vm_developer_common_widgets.dart
bkonyi: Sure, I'll give that a shot. This code was mostly pulled from the common table code so I didn't look at it too closely.

packages/devtools_app/lib/src/vm_developer/vm_service_private_extensions.dart
```diff
@@ -0,0 +1,93 @@
...
+
+/// NOTE: this file contains extensions to classes provided by
+/// `package:vm_service` in order to expose VM internal fields in a controlled
+/// fashion. Objects and extensions in this class should not be used outside of
+/// the `vm_developer` directory.
+
+/// Convenience extension to create a [VMPrivateView] from a [VM].
+extension VMPrivateViewExtension on VM {
+  VMPrivateView toPrivateView() {
+    return this == null ? null : VMPrivateView._(this);
```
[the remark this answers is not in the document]
bkonyi: Oddly enough, yes, it is. You can call extension methods on `null` receivers with no ill-effect assuming that extension method doesn't actually access anything. It's very similar to being able to do this in C++:

```c++
Foo* foo = nullptr;
// Totally fine if no locals are accessed in PrintHello
foo->PrintHello();
```

packages/devtools_app/lib/src/vm_developer/vm_service_private_extensions.dart
bkonyi: I thought there was, but apparently that works just fine... Good suggestion; it really cleans things up!

packages/devtools_app/lib/src/vm_developer/vm_statistics_view.dart
bkonyi: The problem I run into when removing the row is that the button decides it wants to take up the full width of the page. Not sure if there's another way around that?

kenzieschmoll: Ah interesting. Can't think of a a better solution off the top of my head.

packages/devtools_app/lib/src/vm_developer/vm_developer_common_widgets.dart
bkonyi: Yes, see `IsolateMemoryStatisticsWidget`.

packages/devtools_app/lib/src/vm_developer/vm_developer_common_widgets.dart
bkonyi: Sure, will do.

packages/devtools_app/lib/src/vm_developer/vm_developer_common_widgets.dart
bkonyi: Yup, that seems to work. Done.

packages/devtools_app/lib/src/vm_developer/isolate_statistics_view.dart
```diff
@@ -0,0 +1,530 @@
...
+  @override
+  Widget build(BuildContext context) => IsolateStatisticsViewBody();
+}
+
+class IsolateStatisticsViewBody extends StatelessWidget {
+  final IsolateStatisticsViewController controller =
+      IsolateStatisticsViewController();
+
+  @override
+  Widget build(BuildContext context) {
```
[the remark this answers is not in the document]
bkonyi: Done.

bkonyi:
Done addressing first round of comments. PTAL.

packages/devtools_app/lib/src/vm_developer/isolate_statistics_view.dart
```diff
@@ -0,0 +1,544 @@
...
+
+  final name = _TagColumn();
+  final percentage = _PercentageColumn();
+
+  List<ColumnData<VMTag>> get columns => [name, percentage];
+  final IsolateStatisticsViewController controller;
+
+  @override
+  Widget build(BuildContext context) {
+    return Flexible(
```
kenzieschmoll, line 245: instead of making the Flexible part of this widget, put the flexible directly inside the flex widget (row / column) where it is used. This makes the code more flexible (no pun intended) because as is, the TagStatisticsWidget can only be used inside flex widgets like Row or Column, but this is not evident unless you dig into the build method. Same comment for similar widgets like GeneralIsolateStatisticsWidget and ServiceExtensionsWidget. 

Placing the Flexible widgets directly in the Row or Column also makes the flex layout of the row more readable, as you can see what the flex values are directly in the flex widget where they are being used.

packages/devtools_app/lib/src/vm_developer/isolate_statistics_view.dart
```diff
@@ -0,0 +1,544 @@
...
+  final InstanceRef stackTrace;
+  final frame = _StackTraceViewerFrameColumn();
+
+  @override
+  Widget build(BuildContext context) {
+    final List<String> lines = stackTrace?.allocationLocation?.valueAsString
+        ?.split('\n')
+        ?.where((e) => e.isNotEmpty)
+        ?.toList();
+    return Flexible(
```
kenzieschmoll, line 411: same comment here

packages/devtools_app/lib/src/vm_developer/isolate_statistics_view.dart
```diff
@@ -0,0 +1,544 @@
...
+                    },
+                  ),
+                ),
+                StackTraceViewerWidget(
+                  stackTrace: selectedPort.value,
+                ),
+              ], initialFractions: const [
+                0.3,
+                0.7,
+              ]),
```
kenzieschmoll, line 496: trailing comma

packages/devtools_app/lib/src/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -0,0 +1,108 @@
...
+                      child: SelectableText(
+                        row?.value?.toString() ?? '--',
+                        style: fixedFontStyle(context),
+                      ),
+                    ),
+                  ],
+                )
+            ],
+          ),
+        if (table != null) table,
```
kenzieschmoll, line 86: if every instance of table needs to be wrapped in a `Flexible`, wrap table in a `Flexible` here instead of wrapping each `table` parameter in a `Flexible` at the declaration of `VMInfoList`

packages/devtools_app/lib/src/vm_developer/vm_developer_tools_screen.dart
```diff
@@ -0,0 +1,120 @@
...
+                onDestinationSelected: controller.selectIndex,
+                destinations: [
+                  for (final view in VMDeveloperToolsScreenBody.views)
+                    NavigationRailDestination(
+                      label: Text(view.title),
+                      icon: Icon(view.icon),
+                    )
+                ],
+              ),
+            Flexible(
```
kenzieschmoll, line 106: use expanded

packages/devtools_app/lib/src/vm_developer/vm_statistics_view.dart
```diff
@@ -0,0 +1,304 @@
...
+///   - Current memory consumption
+class GeneralVMStatisticsWidget extends StatelessWidget {
+  const GeneralVMStatisticsWidget({@required this.controller});
+
+  final VMStatisticsViewController controller;
+
+  @override
+  Widget build(BuildContext context) {
+    final vm = controller.vm;
+    return Flexible(
```
kenzieschmoll, line 115: same comment here about moving Flexible to wrap GeneralVMStatisticsWidget where it is declared instead of here

packages/devtools_app/lib/src/vm_developer/vm_statistics_view.dart
```diff
@@ -0,0 +1,304 @@
...
+///   - Zone allocator memory utilization
+class ProcessStatisticsWidget extends StatelessWidget {
+  const ProcessStatisticsWidget({@required this.controller});
+
+  final VMStatisticsViewController controller;
+
+  @override
+  Widget build(BuildContext context) {
+    final vm = controller.vm;
+    return Flexible(
```
kenzieschmoll, line 159: ditto

packages/devtools_app/lib/src/vm_developer/vm_statistics_view.dart
```diff
@@ -0,0 +1,304 @@
...
+
+  final VMStatisticsViewController controller;
+  final bool systemIsolates;
+
+  @override
+  Widget build(BuildContext context) {
+    final title = systemIsolates ? 'System Isolates' : 'Isolates';
+    final isolates =
+        systemIsolates ? controller.systemIsolates : controller.isolates;
+    return Flexible(
```
kenzieschmoll, line 288: same - not trying to beat a dead horse, but just pointing out all the spots so you don't have to hunt :)

kenzieschmoll (approved):
lgtm_army
 once comments are addressed

packages/devtools_app/lib/src/vm_developer/vm_developer_common_widgets.dart
bkonyi: Unfortunately we need to use `Expanded` sometimes, so I think we're stuck with adding `Flexible` at each declaration site.

kenzieschmoll:
lgtm_orangatang

PR #2502, opened 2020-11-16
Url and query param parsing cleanup

kenzieschmoll:
Cleanup and refactoring in preparation for loading app size files from server.


jacob314 (approved):
lgtm

Issue #2505, opened 2020-11-16
P2: [devtools_app] Make the "Import File" button works.


tugorez:
![image](https://user-images.githubusercontent.com/7821057/99320677-c8110800-2831-11eb-9a7e-8322e47dec84.png)

tugorez:
I will start writing the PR once https://github.com/flutter/plugins/pull/3140 is merged.

kenzieschmoll:
closing as a dupe of https://github.com/flutter/devtools/issues/2149

Issue #2149, opened 2020-07-01
P2: Add file picker to DevTools


kenzieschmoll:
P1 target is the web - but we would like the picker to work for desktop too since that is where we develop.

This picker is needed for the Code Size page under development. The picker will also be used to load snapshots into DevTools for the pages that currently support it (timeline, performance).

The current need is to pick .json files, but we may need to support more file types in the future (likely .zip).

kenzieschmoll:
In terms of Desktop support - even if we were able to just do Mac, that would be helpful for us.

kenzieschmoll:
@ditman

ditman:
@kenzieschmoll there's currently no support for desktop at all in the plugin (and I've never done any Mac development! :P)

As a backup, have you considered using https://pub.dev/packages/file_picker ? That one seems to support more platforms, and the only thing that you need to consider when dealing on native vs web is that native returns File objects from `dart:io`, and the web version returns File objects from `dart:html`.

ditman:
_Maybe @stuartmorgan can add desktop support to the current image_picker with https://github.com/google/flutter-desktop-embedding/tree/master/plugins/file_chooser?_

stuartmorgan:
> As a backup, have you considered using https://pub.dev/packages/file_picker ? That one seems to support more platforms

It doesn't support any of the Fluter desktop embeddings, so would only help you for web.

> Maybe @stuartmorgan can add desktop support to the current image_picker with https://github.com/google/flutter-desktop-embedding/tree/master/plugins/file_chooser?

I'm confused. The bug is talking about opening json and zip files; what does that have to do with image_picker?

(Image picker functionality could be approximated on macOS using file_chooser, but there are better ways to pick images. file_chooser is poorly suited to implementing image_picker as is, and it would be much better to add an actual macOS implementation to image_picker.)

> OR MAYBE we can add web to the file_chooser plugin?

The goal is to eliminate FDE; I am strongly opposed to expanding its scope to web.


I think the first step here is to back up and have a Flutter-level discussion about how file picking should work:
1. Should it be part of the framework itself?
2. If not, should it be a 1P plugin?

If the answer to both is no, then the question for this issue becomes: which existing 3P plugin should be used for dev tools, and what's involved. (E.g., if you want to use file_picker, should we contribute desktop implementations assuming they would be accepted, or host federated implementations somewhere if not?)

[...]

PR #2506, opened 2020-11-16
[devtools_app] Integrate the file_selector plugin

tugorez:
Fix https://github.com/flutter/devtools/issues/2505


kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/2149

tugorez:
@kenzieschmoll I added support for web, linux and macOS. I manually tested and the web integration works. 
It should be the same for linux and macOS: Can you help me to try those platforms integration?

packages/devtools_app/lib/src/app_size/file_import_container.dart
```diff
@@ -138,18 +140,35 @@ class _FileImportContainerState extends State<FileImportContainer> {
...
+            _handleImportedFile(devToolsJsonFile);
+          },
+          child: const MaterialIconLabel(Icons.file_upload, 'Import File'),
+        ),
+      ],
+    );
+  }
+
+  // Only show the import button in linux, macOS and web.
+  bool get _shouldShowImportButton =>
```
kenzieschmoll, line 168: can this just be true? what will happen if someone tries to build devtools and run on windows? `defaultTargetPlatform == TargetPlatform.macOS` is false when I tested on my mac.

kenzieschmoll:
Tested on MacOs and it appears that only directories are selectable when finder is launched. See this screenshot with all the .json files disabled.
Screen Shot 2021-01-14 at 9 37 03 AM

packages/devtools_app/lib/src/app_size/file_import_container.dart
tugorez: > can this just be true? what will happen if someone tries to build devtools and run on windows?

If someone presses the import button in a platform that is not supported by file_selector then an [Unimplemented error][1] will be thrown.

> defaultTargetPlatform == TargetPlatform.macOS is false when I tested on my mac.
 
Oh then I should find another way to discover the platform the project is running on. 
 
[1]: https://github.com/flutter/plugins/blob/master/packages/file_selector/file_selector_platform_interface/lib/src/platform_interface/file_selector_interface.dart#L55

tugorez:
> Tested on MacOs and it appears that only directories are selectable when finder is launched. See this screenshot with all the .json files disabled.
> Screen Shot 2021-01-14 at 9 37 03 AM

@stuartmorgan do you know why this might be happening? I'm providing [this XTypeGroup][1] to the plugin.

[1]: https://github.com/tugorez/devtools/blob/file_selector/packages/devtools_app/lib/src/app_size/file_import_container.dart#L150

packages/devtools_app/lib/src/app_size/file_import_container.dart
stuartmorgan: Is there some reason you're not using `file_selector_windows`?

stuartmorgan:
> I'm providing [this XTypeGroup](https://github.com/tugorez/devtools/blob/file_selector/packages/devtools_app/lib/src/app_size/file_import_container.dart#L150) to the plugin.

You want `XTypeGroup(extensions: ['json'])`. The extension of 'foo.json' is `json`, not `.json`; `.` is a separator.

I'm surprised it worked on web. Maybe there's fixup happening there?

packages/devtools_app/lib/src/app_size/file_import_container.dart
tugorez: > Is there some reason you're not using file_selector_windows?

That's a good point. Let me add it and remove the platform validation. @kenzieschmoll does that sounds good to you?

tugorez:
> > I'm providing [this XTypeGroup](https://github.com/tugorez/devtools/blob/file_selector/packages/devtools_app/lib/src/app_size/file_import_container.dart#L150) to the plugin.
> 
> You want `XTypeGroup(extensions: ['json'])`. The extension of 'foo.json' is `json`, not `.json`; `.` is a separator.
> 
> I'm surprised it worked on web. Maybe there's fixup happening there?

Oh you are right. Prepending a "." to the extensions is a [web only thing][1]. It should be fine to change ".json" to "json" as the "normalization" should be handled by [this][2].

[1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#unique_file_type_specifiers
[2]: https://github.com/flutter/plugins/blob/master/packages/file_selector/file_selector_web/lib/src/utils.dart#L37

packages/devtools_app/lib/src/app_size/file_import_container.dart
```diff
@@ -138,18 +140,35 @@ class _FileImportContainerState extends State<FileImportContainer> {
     );
```
kenzieschmoll, line 140: not your fault, but can you make the Text widget in _buildImportedFileDisplay ellipsize instead of wrap to a second line?

tugorez:
> > > I'm providing [this XTypeGroup](https://github.com/tugorez/devtools/blob/file_selector/packages/devtools_app/lib/src/app_size/file_import_container.dart#L150) to the plugin.
> > 
> > 
> > You want `XTypeGroup(extensions: ['json'])`. The extension of 'foo.json' is `json`, not `.json`; `.` is a separator.
> > I'm surprised it worked on web. Maybe there's fixup happening there?
> 
> Oh you are right. Prepending a "." to the extensions is a [web only thing](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#unique_file_type_specifiers). It should be fine to change ".json" to "json" as the "normalization" should be handled by [this](https://github.com/flutter/plugins/blob/master/packages/file_selector/file_selector_web/lib/src/utils.dart#L37).

I changed the extension, @kenzieschmoll PTAL :)

tugorez:
cc @ditman

kenzieschmoll (commented):
tested manually on Mac and it works great! I assume linux would work too - maybe @jacob314 or @terrylucas could do a quick check since you already have a linux dev environment set up?

packages/devtools_app/lib/src/app_size/file_import_container.dart
kenzieschmoll: making it work on all platforms sgtm 👍

kenzieschmoll: end users won't be hit by this anyway - they use DevTools on web. We mostly develop on Mac and Linux and none of us to my knowledge develop on windows.

tugorez: I just added windows and removed the platform validation. Thanks to both :)!

packages/devtools_app/lib/src/app_size/file_import_container.dart
```diff
@@ -128,29 +129,36 @@ class _FileImportContainerState extends State<FileImportContainer> {
...
   }
 
-  Widget _buildImportedFileDisplay() {
-    return Text(
-      importedFile?.path ?? 'No File Selected',
-      style: TextStyle(
-        color: Theme.of(context).textTheme.headline1.color,
-      ),
-      textAlign: TextAlign.left,
+  Widget _buildImportedFileDisplay() => Text(
```
kenzieschmoll, line 132: can you add `overflow: TextOverflow.ellipsis,` to this text widget? thanks.

packages/devtools_app/lib/src/app_size/file_import_container.dart
```diff
@@ -128,29 +129,36 @@ class _FileImportContainerState extends State<FileImportContainer> {
...
+
+  Widget _buildImportButton() {
+    return Row(
+      mainAxisAlignment: MainAxisAlignment.center,
+      children: [
+        OutlinedButton(
+          onPressed: () async {
+            final acceptedTypeGroups = [
+              XTypeGroup(extensions: ['json']),
+            ];
```
kenzieschmoll, line 146: can you make `List<String> acceptedTypeGroups` an optional parameter in the `FileImportContainer` constructor that defaults to `['json']`, and use that here instead?

packages/devtools_app/pubspec.yaml
```diff
@@ -11,23 +11,27 @@ version: 0.9.6
 homepage: https://github.com/flutter/devtools
 
 environment:
-  sdk: '>=2.10.0 <3.0.0'
+  sdk: ">=2.10.0 <3.0.0"
```
kenzieschmoll, line 14: any reason for changing from ' to " ? if not, can we leave as single quotes?

packages/devtools_app/pubspec.yaml
```diff
@@ -113,12 +116,12 @@ flutter:
...
-    path: ../devtools_server  #OVERRIDE_FOR_DEVELOPMENT
-  devtools_shared:  #OVERRIDE_FOR_DEVELOPMENT
-    path: ../devtools_shared  #OVERRIDE_FOR_DEVELOPMENT
-  devtools_testing:  #OVERRIDE_FOR_DEVELOPMENT
-    path: ../devtools_testing  #OVERRIDE_FOR_DEVELOPMENT
+    path: ../devtools_server #OVERRIDE_FOR_DEVELOPMENT
+  devtools_shared: #OVERRIDE_FOR_DEVELOPMENT
+    path: ../devtools_shared #OVERRIDE_FOR_DEVELOPMENT
+  devtools_testing: #OVERRIDE_FOR_DEVELOPMENT
+    path: ../devtools_testing #OVERRIDE_FOR_DEVELOPMENT
```
kenzieschmoll, line 127: please revert these white space changes as well. Thank you.

packages/devtools_app/lib/src/app_size/file_import_container.dart
```diff
@@ -174,6 +175,21 @@ class _FileImportContainerState extends State<FileImportContainer> {
     );
   }
 
+  void _importFile() async {
+    final acceptedTypeGroups = [
+      XTypeGroup(extensions: ['json'])
+    ];
```
ditman, line 181: Echoing kenz's earlier review:

> can you make `List<String> acceptedTypeGroups` an optional parameter in the `FileImportContainer` constructor that defaults to `['json']`, and use that here instead?

`acceptedTypeGroups` should be defined in line ~20 of this file, and be an optional named parameter to FileImportContainer, which defaults to the current XTypeGroup of json, so they can use it in other places of the App :)

packages/devtools_app/lib/src/app_size/file_import_container.dart
```diff
@@ -35,6 +38,9 @@ class FileImportContainer extends StatefulWidget {
 
   final void Function(String error) onError;
 
+  // The file's extensions where we are going to get the data from.
```
kenzieschmoll, line 41: nit: use /// for dart doc. And maybe "the valid file extensions for this file selector"?

kenzieschmoll (approved):
One nit - then LGTM. Thanks for your work on this!

kenzieschmoll (approved):

ditman:
👏 bravo!

PR #2509, opened 2020-11-17
expose information about oversized images

devoncarew:
expose information about oversized images:
- add a toggle for the `ext.flutter. invertOversizedImages` service extension; this is on the Inspector page right now - it seems more immediately something people would use while looking at their app that a memory related toggle. Happy to move to to wherever is most useful however
- process the new `Flutter.ImageSizesForFrame` events; these include information about how much more memory an image is using than it needs
- handle displaying Flutter.Error event information on the logging page. Previously we had a todo here to render using the Inspector mechanisms. Since this was not implemented however, the user just saw a blank details area. This instead renders as JSON; we can improve this going forward.

@terrylucas and @jacob314 for review and comments

@dnfield - the default value for `debugImageOverheadAllowance` seems super twitchy. I think we should likely dial it back so that the flutter image errors trigger less often. I _believe_ it's currently on overage of 1024 bytes. Perhaps some value like 2x memory usage?

Screen Shot 2020-11-17 at 1 41 45 PM

Screen Shot 2020-11-17 at 1 42 17 PM

Screen Shot 2020-11-17 at 1 43 24 PM

Screen Shot 2020-11-17 at 1 44 01 PM

### DIFF
--- a/packages/devtools_app/lib/src/console.dart
+++ b/packages/devtools_app/lib/src/console.dart
@@ -53,6 +53,45 @@ class Console extends StatelessWidget {
   }
 }
 
+/// Renders a _ConsoleText widget with ConsoleControls overlaid on the
+/// top-right corner.
+class ConsoleUsingTextWidget extends StatelessWidget {
+  const ConsoleUsingTextWidget({
+    this.controls,
+    @required this.lines,
+    this.title,
+  }) : super();
+
+  final Widget title;
+  final List<Widget> controls;
+  final List<String> lines;
+
+  String get textContent => lines.join('\n');
+
+  @override
+  Widget build(BuildContext context) {
+    return Column(
+      children: [
+        if (title != null) title,
+        Expanded(
+          child: Material(
+            child: Stack(
+              fit: StackFit.expand,
+              children: [
+                _ConsoleText(text: textContent),
+                if (controls != null && controls.isNotEmpty)
+                  _ConsoleControls(
+                    controls: controls,
+                  ),
+              ],
+            ),
+          ),
+        ),
+      ],
+    );
+  }
+}
+
 /// Renders a top-right aligned ButtonBar wrapping a List of IconButtons
 /// (`controls`).
 class _ConsoleControls extends StatelessWidget {
@@ -70,6 +109,30 @@ class _ConsoleControls extends StatelessWidget {
         buttonPadding: EdgeInsets.zero,
         alignment: MainAxisAlignment.end,
         children: controls,
+      ),
+    );
+  }
+}
+
+class _ConsoleText extends StatelessWidget {
+  const _ConsoleText({@required this.text});
+
+  final String text;
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+    final textStyle =
+        theme.textTheme.bodyText2.copyWith(fontFamily: 'RobotoMono');
+
+    return Padding(
+      padding: const EdgeInsets.all(denseSpacing),
+      child: SingleChildScrollView(
+        child: Text(
+          text,
+          style: textStyle,
+          softWrap: true,
+        ),
       ),
     );
   }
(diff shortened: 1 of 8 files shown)


dnfield:
I'd be fine with changing the default, but I think using a multiplier is going to be misleading.

For example, if you are displaying a 48x48 icon with an image that's 100x100, you may not care, but if you're displaying a full screen image that's taking 1.9x more memory you should probably care.

devoncarew:
> I'd be fine with changing the default, but I think using a multiplier is going to be misleading.
> 
> For example, if you are displaying a 48x48 icon with an image that's 100x100, you may not care, but if you're displaying a full screen image that's taking 1.9x more memory you should probably care.

Would a fixed amount of inefficient memory usage work? 128kb / image?

devoncarew:
> DBC based on screenshot - the icon for the service extension toggle button should be changed from the debug banner icon

I did take a quick look through our existing icons in `icons/` but nothing jumped out at me; I'll take a look again.

packages/devtools_app/lib/src/console.dart
```diff
@@ -53,6 +53,45 @@ class Console extends StatelessWidget {
...
+  Widget build(BuildContext context) {
+    return Column(
+      children: [
+        if (title != null) title,
+        Expanded(
+          child: Material(
+            child: Stack(
+              fit: StackFit.expand,
+              children: [
+                _ConsoleText(text: textContent),
```
jacob314, line 81: why isn't this a list view where each element is a line?

dnfield:
128 is probably fine. We can always adjust if it's too aggressive or not aggressive enough.

jacob314:
> 128 is probably fine. We can always adjust if it's too aggressive or not aggressive enough.
128K is probably fine for now.
It is more complex but maybe the threshold we want is the total wasted KB for the current screen. If you are rendering 10 images on the screen each one of which is wasting 50KB that is about as bad a single image wasting 500KB.

packages/devtools_app/lib/src/logging/logging_controller.dart
```diff
@@ -338,6 +339,18 @@ class LoggingController {
         e.timestamp,
         summary: frameInfoText,
       ));
+    } else if (e.extensionKind == ImageSizesForFrame.eventName) {
+      final List<ImageSizesForFrame> images =
```
jacob314, line 343: remove types that don't add value.

packages/devtools_app/lib/src/service_extensions.dart
```diff
@@ -93,6 +93,18 @@ final debugAllowBanner = ToggleableServiceExtensionDescription<bool>._(
   gaItem: ga.debugBanner,
 );
 
+final invertOversizedImages = ToggleableServiceExtensionDescription<bool>._(
```
jacob314, line 96: It isn't great but material design has an invert colors icon.
https://material.io/resources/icons/?icon=invert_colors&style=baseline

packages/devtools_app/lib/src/console.dart
devoncarew: This was in order to get line wrapping behavior in the logging page's detail area. The flutter error's sent for image overage had very long lines and some of the error message was lost.

I can play around w/ how we display json in that view instead - I think the long line had some newlines embedded in it that we weren't respecting (that were json escaped). I don't think the details area gets very long generally, so I wasn't worried about virtualizing the display.

jacob314: Ok. Virtualizing a list view in Flutter is trivial so it should just be the same lines of code but I agree the performance won't matter in this case unless we are displaying a really large block of text.

devoncarew:
I have some updates to this PR in-flight, but for the location of the toggle button:

> add a toggle for the ext.flutter. invertOversizedImages service extension; this is on the Inspector page right now - it seems more immediately something people would use while looking at their app that a memory related toggle.

It's on the inspector page currently, and I don't see a good spot on the memory page for it. I think a better long term home for it would be:
- for us to add an 'Images' tab to the memory page
- for one of the available toggles in that new tab to be the 'oversized images' toggle

jacob314:
I think an images tab in the memory page would be a good place for this.  I'd like the images tab to have this toggle along with a full list of all images currently displayed in the app.

packages/devtools_app/lib/src/service_extensions.dart
devoncarew: I was able to find a reasonable 'image' icon in the set of material ones.

devoncarew:
This is updated w/ the latest changes now - ptal.

jacob314 (approved):
lgtm

PR #2510, opened 2020-11-17
Support loading app size files from query parameters and local storage

kenzieschmoll:


packages/devtools_app/lib/src/app_size/app_size_screen.dart
```diff
@@ -93,6 +99,43 @@ class _AppSizeBodyState extends State<AppSizeBody>
...
+      preLoadingData = true;
+      final primaryAppSizeFile = await server.requestPrimaryAppSizeFile(
+          queryParams[primaryAppSizeFilePropertyName]);
+      DevToolsJsonFile secondaryAppSizeFile;
+      if (queryParams.containsKey(secondaryAppSizeFilePropertyName)) {
+        secondaryAppSizeFile = await server.requestSecondaryAppSizeFile(
+            queryParams[secondaryAppSizeFilePropertyName]);
+      }
+
+      if (primaryAppSizeFile != null) {
```
jacob314, line 114: should we throw an exception `if primaryAppSizeFile == null`

packages/devtools_app/lib/src/app_size/app_size_screen.dart
```diff
@@ -93,6 +99,43 @@ class _AppSizeBodyState extends State<AppSizeBody>
...
+        preLoadingData = true;
+        if (secondaryAppSizeFile != null) {
+          controller.loadDiffTreeFromJsonFiles(
+            oldFile: secondaryAppSizeFile,
+            newFile: primaryAppSizeFile,
+            onError: (error) => Notifications.of(context).push(error),
```
jacob314, line 120: @pq is this the kind of error case you were seeing where callbacks were issued with a potentially invalid context?
I think we should check whether the context is still valid before pushing the notification. If the page has already been removed from the app then there is no need to notify.

packages/devtools_app/lib/src/app_size/app_size_screen.dart
```diff
@@ -93,6 +99,43 @@ class _AppSizeBodyState extends State<AppSizeBody>
...
+          );
+          _tabController.animateTo(tabs.indexOf(diffTab));
+        } else {
+          controller.loadTreeFromJsonFile(
+            primaryAppSizeFile,
+            (error) => Notifications.of(context).push(error),
```
jacob314, line 126: same possible issue about possibly pushing notifications to an disposed context.

packages/devtools_app/lib/src/app_size/app_size_screen.dart
```diff
@@ -93,6 +99,43 @@ class _AppSizeBodyState extends State<AppSizeBody>
...
+          );
+          _tabController.animateTo(tabs.indexOf(analysisTab));
+        }
+      }
+    }
+    if (preLoadingData) {
+      setState(() {
+        preLoadingData = false;
```
jacob314, line 134: i'm not spotting any awaits between preload start and end. Are there some unawaited futures?

packages/devtools_app/lib/src/app_size/app_size_screen.dart
```diff
@@ -93,6 +99,43 @@ class _AppSizeBodyState extends State<AppSizeBody>
...
+      final primaryAppSizeFile = await server.requestPrimaryAppSizeFile(
+          queryParams[primaryAppSizeFilePropertyName]);
+      DevToolsJsonFile secondaryAppSizeFile;
+      if (queryParams.containsKey(secondaryAppSizeFilePropertyName)) {
+        secondaryAppSizeFile = await server.requestSecondaryAppSizeFile(
+            queryParams[secondaryAppSizeFilePropertyName]);
+      }
+
+      if (primaryAppSizeFile != null) {
+        preLoadingData = true;
```
jacob314, line 115: should `preLoadingData = true;` be wrapped in setState?

packages/devtools_app/lib/src/config_specific/server/_server_web.dart
```diff
@@ -250,6 +250,49 @@ Future<void> resetDevToolsFile() async {
   }
 }
 
+Future<DevToolsJsonFile> requestPrimaryAppSizeFile(String path) async {
+  if (isDevToolsServerAvailable) {
+    final resp = await _request(
+      '$apiGetPrimaryAppSizeFile?$primaryAppSizeFilePropertyName=$path',
```
jacob314, line 256: Specify these as real query parameters instead of building the uri manually for more robust escaping. For example, what would happen if `path` contained an `&`

packages/devtools_app/lib/src/config_specific/server/_server_web.dart
```diff
@@ -250,6 +250,49 @@ Future<void> resetDevToolsFile() async {
...
+    }
+  }
+  return null;
+}
+
+DevToolsJsonFile _devToolsJsonFileFromAppSizeResponse(
+  HttpRequest resp,
+  String filePath,
+) {
+  final _json = json.decode(resp.response);
```
jacob314, line 285: Making a variable private is a little odd. Maybe call this `data` or map instead of `_json`? Also is _json of type dynamic?

packages/devtools_server/lib/src/file_system.dart
```diff
@@ -0,0 +1,34 @@
...
+  static File file(String pathFromHomeDir) {
+    final file = File('${userHomeDir()}/$pathFromHomeDir');
+    if (!file.existsSync()) {
+      return null;
+    }
+    return file;
+  }
+
+  static String fileAsJson(String pathFromHomeDir) {
+    final _file = file(pathFromHomeDir);
```
jacob314, line 23: nit: don't use privates to avoid naming conflicts. Maybe it is a sign the method name is not good if it also makes a good variable name as methods should typically be verbs while variable names should be nouns.

packages/devtools_app/lib/src/config_specific/server/_server_web.dart
```diff
@@ -250,6 +250,49 @@ Future<void> resetDevToolsFile() async {
   }
 }
 
+Future<DevToolsJsonFile> requestPrimaryAppSizeFile(String path) async {
```
jacob314, line 253: these two methods are basically the same. Can we just have a single generic method that fetches files from the server. It would work for the appSize case and future cases where other files also need to be fetched from the server.

packages/devtools_server/lib/src/server_api.dart
```diff
@@ -148,6 +149,32 @@ class ServerApi {
           request,
           json.encode(_devToolsUsage.surveyShownCount),
         );
+      case apiGetPrimaryAppSizeFile:
```
jacob314, line 152: these two cases lookidentical and should probably be merge to a single method that is just about fetching cached files rather than something appSize specific.

packages/devtools_server/lib/src/server.dart
```diff
@@ -554,10 +555,10 @@ File _devToolsBackup;
 
 bool backupAndCreateDevToolsStore() {
   assert(_devToolsBackup == null);
-  final devToolsStore = File('${DevToolsUsage.userHomeDir()}/.devtools');
+  final devToolsStore = File('${LocalFileSystem.userHomeDir()}/.devtools');
```
jacob314, line 558: rather than exposing userHomeDIr which is a terrifying thing to allow access to paths relative to, what if we expose only
`devtoolsRootDir()` which can be `userHomeDir()/.devtools`
That will help avoid cases where these server APIs could be abused to access arbitrary files on the user's machine.

packages/devtools_server/lib/src/server_api.dart
jacob314: We also need to avoid serving arbitrary files on the users machine which could lead to security vulnerabilities. serving only files under the `.devtools` directory and prohibiting paths that resolve to a directory outside of .devtools would make this safe.
For example we need to avoid accepting a file path like
`././mypasswords.txt`
as that would resolve to a directory not under `.devtools`

packages/devtools_server/lib/src/server.dart
jacob314: We'll still need userHomeDir to access other files but we should never use it for file serving logic due to security concerns.

packages/devtools_server/lib/src/server_api.dart
kenzieschmoll: App size files may not be in a .devtools directory though. If a user runs the --analyze-size command on their app, the app size files would likely live in their app's directory or wherever else on their machine they choose to specify as output location (for example, when I have run this command, I usually send the files to my Downloads folder).

One thing we could do is change the flutter_tools code to save them not to the flutter app's base directory by default, but instead to a .devtools folder. Right now, we only have a .devtools file where we store client id data, analytics opt-in, survey info, etc., but we could modify this so that we have .devtools folder with the .devtools file inside of it. If we did something like this, would we have the same security concern for accessing files at `<user-home-dir>/.devtools/`?

packages/devtools_app/lib/src/config_specific/server/_server_web.dart
kenzieschmoll: fixed

packages/devtools_app/lib/src/app_size/app_size_screen.dart
kenzieschmoll: checked if mounted

packages/devtools_app/lib/src/app_size/app_size_screen.dart
kenzieschmoll: the awaits happen right after setting preLoadingData to true

packages/devtools_app/lib/src/config_specific/server/_server_web.dart
kenzieschmoll: refactored to create generic requestFile method

packages/devtools_server/lib/src/file_system.dart
kenzieschmoll: fixed

packages/devtools_app/lib/src/config_specific/server/_server_web.dart
kenzieschmoll: The HTTP request fails when the path is encoded. 
e.g.
`http://localhost:9101/api/getTestAppSizeFile?appSizeTest=Downloads%2Fapk-size-analysis-encryption-app-new.json 40`

kenzieschmoll: Never mind - this was a naming issue, not due to encoding 👍

packages/devtools_server/lib/src/server_api.dart
kenzieschmoll: Changed to using ~/.devtools/ directory. Made the change backwards compatible so existing .devtools files will be moved inside of the .devtools/ directory if present

packages/devtools_server/lib/src/file_system.dart
```diff
@@ -3,24 +3,61 @@ import 'dart:io';
...
         Platform.operatingSystem == 'windows' ? 'APPDATA' : 'HOME';
     final String value = Platform.environment[envKey];
     return value == null ? '.' : value;
   }
 
-  static File fileFromPath(String pathFromHomeDir) {
-    final file = File('${userHomeDir()}/$pathFromHomeDir');
+  /// Returns the path to the DevTools storage directory.
+  static String devToolsDir() {
+    return path.join(_userHomeDir(), '.devtools');
```
jacob314, line 18: oh shoot. For safety we should pick a slightly different name (e,g. 
.dartdevtools for the directory.
otherwise we will cause problems if someone later opens an older version of devtools that assumes .devtools is a file would probably fail. Having the maybeMigrate method is still good as we want that to save your old opt-in setting.

packages/devtools_server/lib/src/file_system.dart
```diff
@@ -3,24 +3,61 @@ import 'dart:io';
...
+    Directory('${LocalFileSystem.devToolsDir()}').createSync();
+  }
+
+  /// Returns a DevTools file from the given path.
+  ///
+  /// Only files withing ~/.devtools/ can be accessed.
+  static File devToolsFileFromPath(String pathFromDevToolsDir) {
+    ensureDevToolsDirectory();
+    final file = File(path.join(devToolsDir(), pathFromDevToolsDir));
     if (!file.existsSync()) {
```
jacob314, line 50: also ensure the path isn't outside of the devtools dir.
e.g.
pathFromDevToolsDir better not be
../../../yourpassword.txt

packages/devtools_server/lib/src/file_system.dart
```diff
@@ -43,8 +43,13 @@ class LocalFileSystem {
 
   /// Returns a DevTools file from the given path.
   ///
-  /// Only files withing ~/.devtools/ can be accessed.
+  /// Only files within ~/.devtools/ can be accessed.
   static File devToolsFileFromPath(String pathFromDevToolsDir) {
+    if (pathFromDevToolsDir.contains('../')) {
```
jacob314, line 48: more robust would be to use 
https://api.dart.dev/stable/2.10.4/dart-io/FileSystemEntity/resolveSymbolicLinks.html
to resolve all . and ../
you could also fail if there is any `..` in the path.

packages/devtools_server/lib/src/file_system.dart
```diff
@@ -43,8 +43,13 @@ class LocalFileSystem {
...
+      // The passed in path should not be able to walk up the directory tree
+      // outside of the ~/.dart-devtools/ directory.
```
jacob314, line 50: isn't this flutter-devtools not dart-devtools?

jacob314 (approved):
lgtm

packages/devtools_app/test/app_size_screen_test.dart
```diff
@@ -372,12 +375,12 @@ void main() {
...
 
       appSizeController.loadTreeFromJsonFile(
-        DevToolsJsonFile(
+        jsonFile: DevToolsJsonFile(
           name: 'unsupported_file.json',
           lastModifiedTime: lastModifiedTime,
           data: unsupportedFile,
         ),
-        (error) => Notifications.of(buildContext).push(error),
+        onError: (error) => Notifications.of(buildContext).push(error),
```
jacob314, line 383: looks like this one is missing a check for mounted

kenzieschmoll: this is a callback in a test, so I left it because we can't check `if (mounted)` here

packages/devtools_app/lib/src/app_size/app_size_screen.dart
kenzieschmoll: added a todo to check for this - going to land and polish in follow ups

packages/devtools_app/lib/src/app_size/app_size_screen.dart
kenzieschmoll: added a todo to add better error handling. need to get some of this code landed to build implementation in google3.

packages/devtools_server/lib/src/file_system.dart
kenzieschmoll: done

packages/devtools_server/lib/src/file_system.dart
kenzieschmoll: done

packages/devtools_server/lib/src/file_system.dart
kenzieschmoll: returned null if .. is present

packages/devtools_server/lib/src/file_system.dart
kenzieschmoll: done

PR #2515, opened 2020-11-19
Fix null error in service manager

kenzieschmoll:


devoncarew (approved):

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -656,7 +656,7 @@ class ServiceExtensionManager extends Disposer {
   Future<void> _handleDebugEvent(Event event) async {
     if (event.kind == EventKind.kResume) {
       final isolateId = event.isolate.id;
-      final callbacks = _callbacksOnIsolateResume[isolateId];
+      final callbacks = _callbacksOnIsolateResume[isolateId] ?? [];
```
jacob314, line 659: does this indicate we have resumed an isolate we didn't know about?

jacob314:
lgtm

PR #2518, opened 2020-11-26
add padding between columns in tables

devoncarew:
- add padding between columns in tables

This adds `borderPadding` padding (~2px) between columns in a table - this prevents adjacent cell content from running into each other. This does not increase the width of cells, but insets each cell a small amount (a cell the requests 20px will instead get 18px for its content).

If this lands it will likely supersede https://github.com/flutter/devtools/pull/2087.

cc @jacob314 @kenzieschmoll 

Screen Shot 2020-11-24 at 4 28 45 PM

### DIFF
--- a/packages/devtools_app/lib/src/table.dart
+++ b/packages/devtools_app/lib/src/table.dart
@@ -1,3 +1,7 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
 import 'dart:math';
 
 import 'package:flutter/foundation.dart';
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/table.dart
```diff
@@ -1250,8 +1254,9 @@ class _TableRowState<T> extends State<TableRow<T>>
         );
       }
 
-      content = SizedBox(
+      content = Container(
         width: columnWidth,
+        padding: const EdgeInsets.only(right: borderPadding),
```
jacob314, line 1259: having right padding for the last cell in the table seems undesirable. Let me see if I can bring back that old CL I did to try to fix the padding on the table the right way.

devoncarew: > having right padding for the last cell in the table seems undesirable

In practice - as long as a cell doesn't draw a different background color - this works pretty well (and is strictly better than no padding).

> Let me see if I can bring back that old CL I did to try to fix the padding on the table the right way.

Sure - happy either way.

devoncarew:
(dropping to a draft PR for now)

PR #2526, opened 2020-12-02
Add padding between columns and add minWidth for flexible columns.

jacob314:


packages/devtools_app/lib/src/app_size/app_size_table.dart
```diff
@@ -77,7 +77,8 @@ class _NameColumn extends TreeColumnData<TreemapNode> {
 }
 
 class _SizeColumn extends ColumnData<TreemapNode> {
-  _SizeColumn() : super('Size', alignment: ColumnAlignment.right);
+  _SizeColumn()
+      : super('Size', alignment: ColumnAlignment.right, fixedWidthPx: 100.0);
```
jacob314, line 81: we were overriding fields instead of using the existing parameter in the constructor. Cleaned that up so that we can make the constructor parameter required. Fyi @terrylucas.

packages/devtools_app/lib/src/table.dart
```diff
@@ -117,18 +119,36 @@ class FlatTableState<T> extends State<FlatTable<T>>
     sortData(widget.sortColumn, widget.sortDirection);
   }
 
-  List<double> _computeColumnWidths(double maxWidth) {
+  @visibleForTesting
+  List<double> computeColumnWidths(double maxWidth) {
```
jacob314, line 123: This method is the core algorithm change.

packages/devtools_app/test/table_test.dart
```diff
@@ -179,6 +184,114 @@ void main() {
...
+        final FlatTableState<TestData> state =
+            tester.state(find.byWidget(table));
+        final columnWidths = state.computeColumnWidths(200.0);
+        expect(columnWidths.length, equals(3));
+        expect(columnWidths[0], equals(300.0)); // Fixed width column.
+        expect(columnWidths[1], equals(400.0)); // Fixed width column.
+        expect(columnWidths[2], equals(0.0)); // Variable width column.
+      }
+
+      // TODO(jacobr): add a golden image test.
```
jacob314, line 241: I added some golden tests but the dimensions of the SizedBox aren't being respected in the golden screenshot and I don't have time to figure out why.

packages/devtools_app/lib/src/device_dialog.dart
```diff
@@ -254,7 +254,7 @@ class _NameColumn extends ColumnData<_DialogFlag> {
 }
 
 class _DescriptionColumn extends ColumnData<_DialogFlag> {
-  _DescriptionColumn() : super.wide('Description');
+  _DescriptionColumn() : super.wide('Description', minWidthPx: 100);
```
jacob314, line 257: probably not useful to shrink the description down to zero width so give it a min width.

packages/devtools_app/lib/src/network/network_screen.dart
```diff
@@ -310,7 +310,7 @@ class NetworkRequestsTable extends StatelessWidget {
 
 class UriColumn extends ColumnData<NetworkRequest>
     implements ColumnRenderer<NetworkRequest> {
-  UriColumn() : super.wide('Uri');
+  UriColumn() : super.wide('Uri', minWidthPx: 50);
```
jacob314, line 313: Fyi @kenzieschmoll gave this one a minWidth because completely hiding the Uri column when there is not enough room like we were doing is probably exactly the opposite of what we want. 
A larger min width could also be reasonable. Open to suggestions.

jacob314:
This fixes the issue where the Uri column was hidden by default when the network page was shown in slightly narrow window as well as padding between rows.
Fixed the two issues together as providing min-widths for flexible columns means I don't have to care about edge cases when flexible columns shrink to width zero leaving only multiple spans of adjacent padding.
Screen Shot 2020-12-01 at 7 26 53 PM

Screen Shot 2020-12-01 at 8 21 23 PM

packages/devtools_app/lib/src/network/network_screen.dart
terrylucas: Could we use a ... in the middle of URI with a hover over to show to the complete URI?

terrylucas:
lgtm, maybe each column could use similar code functionality like FractionallySizedBox widget uses, then we could use a percentage instead of a min or fixed pixel width?  Not now, but maybe later in other things too but tables are the most common.

packages/devtools_app/lib/src/app_size/app_size_table.dart
kenzieschmoll, line 81: nit: trailing comma in all of the modified super calls

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -1202,7 +1202,7 @@ class _LibraryRefColumn extends TreeColumnData<Reference> {
 
 class _ClassOrInstanceCountColumn extends ColumnData<Reference> {
   _ClassOrInstanceCountColumn()
-      : super('Count', alignment: ColumnAlignment.right);
+      : super('Count', alignment: ColumnAlignment.right, fixedWidthPx: 75.0);
```
kenzieschmoll, line 1205: nit: trailing comma in the modified lines in this file too

jacob314:
Making the constraints for Columns more similar to other flutter constraints would be nice but would be a lot of work so I'm afraid to do it right now. I'm also not sure whether we would want to matched FractionallySizedBox or Flex layout semantics.
For now I think we are generally ok as we typically only have one column that really needs to have expanded size and the others are all nice at fixed sizes but I can be convinced if we have cases where we would really like more than one flexibly sized column.
> lgtm, maybe each column could use similar code functionality like FractionallySizedBox widget uses, then we could use a percentage instead of a min or fixed pixel width? Not now, but maybe later in other things too but tables would be the most common.

packages/devtools_app/lib/src/network/network_screen.dart
jacob314: if we had a good generic text widget like that then yes that would probably be useful.

packages/devtools_app/lib/src/table.dart
```diff
@@ -117,18 +119,36 @@ class FlatTableState<T> extends State<FlatTable<T>>
...
+    final available = max(0, maxWidth - allocatedWidth);
 
     final widths = <double>[];
     for (ColumnData<T> column in widget.columns) {
-      final width = column.fixedWidthPx ?? (available / unconstrainedCount);
+      double width = column.fixedWidthPx;
+      if (width == null) {
+        width = available / unconstrainedCount;
+        if (column.minWidthPx != null) {
+          width += column.minWidthPx;
```
kenzieschmoll, line 149: should this be `width = math.max(width, column.minWidthPx)`? Otherwise we could be adding minWidthPx to the divided space for unconstrained columns.

packages/devtools_app/lib/src/table.dart
```diff
@@ -117,18 +119,36 @@ class FlatTableState<T> extends State<FlatTable<T>>
...
+        }
+      }
       widths.add(width);
```
kenzieschmoll, line 152: at the end of this for loop, should `width` be subtracted from `available`? Otherwise the available space could be inaccurate for the remaining unconstrained columns

packages/devtools_app/lib/src/table_data.dart
```diff
@@ -328,17 +328,21 @@ abstract class ColumnData<T> {
...
     this.alignment = ColumnAlignment.left,
-    this.fixedWidthPx,
-  });
+    @required this.fixedWidthPx,
+  })  : assert(fixedWidthPx != null),
+        minWidthPx = null;
 
-  ColumnData.wide(this.title, {this.alignment = ColumnAlignment.left})
+  ColumnData.wide(this.title,
+      {this.minWidthPx, this.alignment = ColumnAlignment.left})
```
kenzieschmoll, line 336: trailing comma

packages/devtools_app/test/table_test.dart
```diff
@@ -39,6 +39,11 @@ void main() {
       );
       await tester.pumpWidget(wrap(table));
       expect(find.byWidget(table), findsOneWidget);
+      // Make sure the total wi
```
kenzieschmoll, line 42: incomplete comment?

packages/devtools_app/test/table_test.dart
```diff
@@ -179,6 +184,114 @@ void main() {
...
+        SizedBox(
+          width: 200.0,
+          height: 200.0,
+          child: table,
+        ),
+      ));
+      {
+        final FlatTableState<TestData> state =
+            tester.state(find.byWidget(table));
+        final columnWidths = state.computeColumnWidths(200.0);
```
kenzieschmoll, line 288: why isn't the max width of 200.0 respected here? the sum of the below column widths are greater than 200?

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
jacob314: done

packages/devtools_app/lib/src/app_size/app_size_table.dart
jacob314: done

packages/devtools_app/lib/src/table.dart
jacob314: Good point. I was afraid that would lead to edge case issues calculating the available width but it is no worse than what I had for the edge cases.

jacob314: actually I am wrong. It is tricky. There is an extra edge case if we do it that way. The problem is when some columns are at their minimum constraints and some aren't. The option you suggested will result in a bug where sometimes we don't actually use the available width or we use too much width. We'll need a 3rd pass to do it correctly. Shouldn't be hard, let me go ahead and add that.

packages/devtools_app/lib/src/table.dart
jacob314: no because the number of unconstrained columns is not changed. Changing the # of unconstrained columns would be even stranger as the the relative sizes of columns would depend on the order of the columns.

packages/devtools_app/lib/src/table_data.dart
jacob314: done

packages/devtools_app/test/table_test.dart
jacob314: removed

packages/devtools_app/test/table_test.dart
jacob314: the max width is not respected if you have fixed width columns that sum up to more than the available width. @terrylucas this is one motivation for supporting fractional width columns as well for cases where we have opinions on the relative widths of columns rather than just fixed and min widths.

jacob314:
I've run out of time. I'll land as is and do a follow up for prettier handling of multiple variable sized columns and some trailing commas. What I've got locally scares me about landing right now as the logic is more complex and I need to start driving.

PR #2531, opened 2020-12-03
Fix analysis failures from e8a23e2

bkonyi:


kenzieschmoll (commented):
are these all really supposed to be wide? there is also an option to specify a fixed width

bkonyi:
> are these all really supposed to be wide? there is also an option to specify a fixed width or a min width

This basically matches the original behavior. Whether or not this is right thing to do longer term is another question, but I just wanted to get the bots green again.

kenzieschmoll (approved):

PR #2539, opened 2020-12-08
Better x-axis labeling

terrylucas:
Updated x-axis labels to be better and not sparse based on duration interval and amount of collected data.
Traces added for collected Raster metrics to VM memory chart.


packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -61,9 +56,15 @@ class ChartController extends DisposableController
     this.displayXLabels = true,
     this.displayYLabels = true,
     this.name,
+    List<int> sharedLabelsTimestamps,
```
jacob314, line 59: nit: rename to labelTimestamps or sharedLabelTimestamps

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -387,17 +291,127 @@ class ChartController extends DisposableController
...
+    }
+
+    if (labelTimestamps.isEmpty) {
+      labelTimestamps.add(timestamps.last);
+    } else {
+      // Check left label is it out of range?
+      final leftEdge = leftMostVisibleTimestampIndex;
+      if (labelTimestamps.first < timestamps[leftEdge]) {
+        // Label is outside of visible range, remove the left label.
+        labelTimestamps.removeAt(0);
```
jacob314, line 374: why is only at most one timestamp out of range?

packages/devtools_app/lib/src/memory/memory_vm_chart.dart
```diff
@@ -150,12 +164,11 @@ class MemoryVMChartState extends State<MemoryVMChart> with AutoDisposeMixin {
 
   // TODO(terry): Move colors to theme?
   final capacityColor = Colors.grey[400];
-  //static const usedColor = Color.fromARGB(255, 51, 181, 229);
-  static const usedColor = Color(0xff33b5e5); // HoloBlue
+  static const externalColor = Color(0xff42a5f5); // Color.blue[400]
```
jacob314, line 167: Nit: if this is Color.blue[400] why not just make this be
static final externalColor = Color.blue[400]
?

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/memory/memory_vm_chart.dart
terrylucas: static final capacityColor = Colors.grey[400];
  static final externalColor = Colors.blue[400];

Done.

packages/devtools_app/lib/src/charts/chart_controller.dart
terrylucas: I guess theoretically this could happen, if the amount samples collected over the fastest label interval of 15-20 seconds was somehow not painted during that 15-20 sec.  It could happen, but I don't think that's likely, but generally 1 or 2 chart ticks a repaint happens and the live feed left-most label slowly move out of the visible chart range.

terrylucas: I was just thinking that when I introduce panning the chart than this could happen.  I'll add a TODO to fix this for panning.

packages/devtools_app/lib/src/charts/chart_controller.dart
terrylucas: Done.

Issue #2507, opened 2020-11-17
Support frame paging


grouma:
The latest VM service protocol supports a limit for the [getStack](https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#getstack) call. DevTools can use this to support frame paging and improve the perceived performance of stepping.

PR #2545, opened 2020-12-09
Support truncated frames

grouma:
Use the new optional `limit` parameter to `getStack` to display the first 10 frames by default. This improves the perceived stepping performance. When the frames are truncated, display a new `SHOW ALL` button which will collect all frames. Also rework the logic to no longer have a race condition when a user steps multiple times before the full stack value is returned.

Screen Shot 2020-12-10 at 11 56 31 AM

Closes https://github.com/flutter/devtools/issues/2507

### DIFF
--- a/packages/devtools_app/lib/src/debugger/call_stack.dart
+++ b/packages/devtools_app/lib/src/debugger/call_stack.dart
@@ -34,23 +34,30 @@ class _CallStackState extends State<CallStack> {
   @override
   Widget build(BuildContext context) {
     return ValueListenableBuilder<List<StackFrameAndSourcePosition>>(
-      valueListenable: controller.stackFramesWithLocation,
-      builder: (context, stackFrames, _) {
-        return ValueListenableBuilder<StackFrameAndSourcePosition>(
-          valueListenable: controller.selectedStackFrame,
-          builder: (context, selectedFrame, _) {
-            return ListView.builder(
-              itemCount: stackFrames.length,
-              itemExtent: defaultListItemHeight,
-              itemBuilder: (_, index) {
-                final frame = stackFrames[index];
-                return _buildStackFrame(frame, frame == selectedFrame);
+        valueListenable: controller.stackFramesWithLocation,
+        builder: (context, stackFrames, _) {
+          return Column(children: [
+            Expanded(
+                child: ValueListenableBuilder<StackFrameAndSourcePosition>(
+              valueListenable: controller.selectedStackFrame,
+              builder: (context, selectedFrame, _) {
+                return ListView.builder(
+                  itemCount: stackFrames.length,
+                  itemExtent: defaultListItemHeight,
+                  itemBuilder: (_, index) {
+                    final frame = stackFrames[index];
+                    return _buildStackFrame(frame, frame == selectedFrame);
+                  },
+                );
               },
-            );
-          },
-        );
-      },
-    );
+            )),
+            if (controller.hasTruncatedFrames)
+              FlatButton(
+                onPressed: () => controller.getFullStack(),
+                child: const Text('SHOW ALL'),
+              )
+          ]);
+        });
   }
 
   Widget _buildStackFrame(
--- a/packages/devtools_app/lib/src/debugger/call_stack.dart
+++ b/packages/devtools_app/lib/src/debugger/call_stack.dart
@@ -33,11 +33,11 @@ class _CallStackState extends State<CallStack> {
 
   @override
   Widget build(BuildContext context) {
-    return ValueListenableBuilder<List<StackFrameAndSourcePosition>>(
-        valueListenable: controller.stackFramesWithLocation,
-        builder: (context, stackFrames, _) {
-          return Column(children: [
-            Expanded(
+    return Column(children: [
+      ValueListenableBuilder<List<StackFrameAndSourcePosition>>(
+          valueListenable: controller.stackFramesWithLocation,
+          builder: (context, stackFrames, _) {
+            return Expanded(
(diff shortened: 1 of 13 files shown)


packages/devtools_app/lib/src/debugger/call_stack.dart
```diff
@@ -33,24 +33,37 @@ class _CallStackState extends State<CallStack> {
...
+                );
               },
-            );
-          },
-        );
-      },
-    );
+            ));
+          }),
+      ValueListenableBuilder<bool>(
```
jacob314, line 55: it would be nice to avoid having a SHOW ALL button and instead load the stack frames when they are needed. All you have to do to make that happen is trigger the load when a frame that isn't available yet is requested in the itemBuilder callback.

grouma: I think that's a fine improvement but should be left as a follow up PR. Note that it can take several seconds to compute the full stack so we'll want to give some indication to the user when it is truncated, and when we are loading more frames.

devoncarew: Ah, this was the feedback I thought of when I saw the screen shot above. This would definitely be a useful improvement.

In terms of displaying progress, perhaps something like a circular progress indicator in the top right of the call stack (in the area where we display actions for other views like the breakpoints view).

devoncarew (commented):
Very cool to see this PR - it should be a good performance improvement for the web backends.

packages/devtools_app/lib/src/debugger/call_stack.dart
```diff
@@ -33,24 +33,37 @@ class _CallStackState extends State<CallStack> {
...
+          valueListenable: controller.hasTruncatedFrames,
+          builder: (_, hasTruncatedFrames, __) {
+            if (hasTruncatedFrames) {
+              return FlatButton(
+                onPressed: () => controller.getFullStack(),
+                child: const Text('SHOW ALL'),
+              );
+            }
+            return Container(height: 0, width: 0);
```
kenzieschmoll, line 64: nit: use `SizedBox()` instead

packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -585,33 +592,55 @@ class DebuggerController extends DisposableController
         [pauseEvent.topFrame],
         reportedException: pauseEvent?.exception,
       );
-      _stackFramesWithLocation.value = [
-        await _createStackFrameWithLocation(tempFrames.first),
-      ];
+      _populateFrameInfo(
+          [await _createStackFrameWithLocation(tempFrames.first)],
+          truncated: true);
```
kenzieschmoll, line 597: nit: trailing comma

packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -585,33 +592,55 @@ class DebuggerController extends DisposableController
...
 
-    _stackFramesWithLocation.value =
-        await Future.wait(frames.map(_createStackFrameWithLocation));
+    return _StackInfo(
+        await Future.wait(frames.map(_createStackFrameWithLocation)),
+        stack.truncated);
+  }
+
+  void _populateFrameInfo(List<StackFrameAndSourcePosition> frames,
+      {bool truncated}) {
```
kenzieschmoll, line 625: nit: trailing comma

packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -585,33 +592,55 @@ class DebuggerController extends DisposableController
...
+    truncated ??= false;
```
kenzieschmoll, line 626: instead of doing this, just move the default value into the method signature

packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -585,33 +592,55 @@ class DebuggerController extends DisposableController
...
       asyncCausalFrames: stack.asyncCausalFrames,
-      reportedException: pauseEvent?.exception,
+      reportedException: _lastEvent?.exception,
     );
 
-    _stackFramesWithLocation.value =
-        await Future.wait(frames.map(_createStackFrameWithLocation));
+    return _StackInfo(
+        await Future.wait(frames.map(_createStackFrameWithLocation)),
+        stack.truncated);
```
kenzieschmoll, line 621: trailing comma

kenzieschmoll (approved):
lgtm with some nits

packages/devtools_app/lib/src/debugger/debugger_controller.dart
grouma: Until NNBD that's actually not a good pattern. Users can still explicitly pass in null which will cause issues.

packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -585,33 +592,55 @@ class DebuggerController extends DisposableController
         [pauseEvent.topFrame],
         reportedException: pauseEvent?.exception,
       );
-      _stackFramesWithLocation.value = [
-        await _createStackFrameWithLocation(tempFrames.first),
-      ];
+      _populateFrameInfo(
+          [await _createStackFrameWithLocation(tempFrames.first)],
```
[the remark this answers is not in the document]
grouma: Yeah this can go either way. Calculating 10 frames should almost always be nearly instant. Do you think I should remove this?

jacob314 (approved):
lgtm
Please file a bug to make lazy loading of stack frames automatic.

Issue #2475, opened 2020-10-30
DevTools url has fragment and query parameters in the wrong order


kenzieschmoll:
Current: `http://127.0.0.1:9100/#/timeline?ide=IntelliJ-IDEA&uri=http%253A%252F%252F127.0.0.1%253A59289%252FolleV5zVaVs%253D`

Expected: `http://127.0.0.1:9100/?ide=IntelliJ-IDEA&uri=http%253A%252F%252F127.0.0.1%253A59289%252FolleV5zVaVs%253D#/timeline`

In the current form, Uri.parse gets the incorrect values for `fragment` and `queryParameters`

kenzieschmoll:
FYI @DanTup

DanTup:
Although it's a bit unintuitive, this is expected. It's not how a URL normally would be formed but in Flutter we only get the fragment part, and we had a convention of putting a query string in it for pass our data around.

Since the routing changes, we handle the URI parsing in `DevToolsRouteInformationParser.parseRouteInformation` where `routeInformation.location` will be just the fragment, eg.:

When at page: `http://localhost:50593/#/?hide=debugger&ide=VSCode&theme=dark&uri=http%3A%2F%2F127.0.0.1%3A50288%2FERZc7YGNhyo%3D%2F`
`routeInformation.fragment` is: `/?hide=debugger&ide=VSCode&theme=dark&uri=http%3A%2F%2F127.0.0.1%3A50288%2FERZc7YGNhyo%3D%2F`

Which we then parse into a URI to get `path` (`/`) and `queryParameters`.

Have you found code that is parsing the URI directly? If so, we should update it to use the params from the routing (since this should also work better for non-web too).

(As an aside, while looking at this I noticed `hide=debugger` isn't working so I'm just looking into that - I don't think it's a result of this though as I can see it in `params`).

kenzieschmoll:
Now there is another '?' character in the url that is causing url parsing issues when trying to show the DevTools survey
e.g. 'http://127.0.0.1:9103/?#/?'

related: https://github.com/flutter/website/pull/6696

DanTup:
@kenzieschmoll are you looking at this or did you want me to? If me, how do you end up with the bad URL? Is the goal to fix that, or just handle it better when it happens?

kenzieschmoll:
[...]

PR #2546, opened 2020-12-10
Add `appSizeBase` and `appSizeTest` args to the devtools server

kenzieschmoll:
This will allow specifying `appSizeBase` and `appSizeTest` file paths as args to the server (e.g. `--appSizeBase old.json --appSizeTest new.json`)

### DIFF
--- a/packages/devtools_server/lib/src/server.dart
+++ b/packages/devtools_server/lib/src/server.dart
@@ -30,6 +30,8 @@ const argHelp = 'help';
 const argVmUri = 'vm-uri';
 const argEnableNotifications = 'enable-notifications';
 const argAllowEmbedding = 'allow-embedding';
+const argAppSizeBase = 'appSizeBase';
+const argAppSizeTest = 'appSizeTest';
 const argHeadlessMode = 'headless';
 const argDebugMode = 'debug';
 const argLaunchBrowser = 'launch-browser';
@@ -90,6 +92,8 @@ Future<void> _serveDevToolsWithArgs(
       : defaultTryPorts;
   final bool verboseMode = args[argVerbose];
   final String hostname = args[argHost];
+  final String appSizeBase = args[argAppSizeBase];
+  final String appSizeTest = args[argAppSizeTest];
 
   if (help) {
     print('Dart DevTools version ${await _getVersion()}');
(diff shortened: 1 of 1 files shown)


packages/devtools_server/lib/src/server.dart
```diff
@@ -208,12 +216,27 @@ Future<HttpServer> serveDevTools({
...
+        .replace(queryParameters: queryParameters)
+        .toString();
+
+    // If app size parameters are present, open to the standalone `appsize`
+    // page, regardless if there is a vm service uri specified. We only check
+    // for the presence of [appSizeBase] here because [appSizeTest] may or may
+    // not be specified (it should only be present for diffs). If [appSizeTest]
+    // is present without [appSizeBase], we will ignore the parameter.
+    if (appSizeBase != null) {
+      final startQueryParamIndex = url.indexOf('?');
```
jacob314, line 234: use uri parsing instead of indexof

jacob314 (approved):
lgtm

packages/devtools_server/lib/src/server.dart
kenzieschmoll: when I tried to add `/#/appsize` using `Uri.replace(path: '/#/appsize)`, the `#` is encoded and the url does load properly.

jacob314: I think you want to use fragment to change the fragment property rather than path.
https://api.dart.dev/stable/2.10.4/dart-core/Uri/fragment.html

kenzieschmoll: this also doesn't work because the fragment is _after_ the query parameters (`http://127.0.0.1:9102?appSizeBase=old.json&appSizeTest=new.json#appsize`), and devtools expects the fragment id to be _before_ the query parameters (`http://127.0.0.1:9102/#/appsize?appSizeBase=old.json&appSizeTest=new.json`)

jacob314: where in devtools are we requiring the fragment to be before the query parameters? that seems like a bug.

kenzieschmoll: see https://github.com/flutter/devtools/issues/2475 for some input from danny

PR #2548, opened 2020-12-10
Fix failing lints

grouma:
Looks like we were broken by some new deprecated members.


kenzieschmoll (approved):

packages/devtools_app/lib/src/memory/memory_graph_model.dart
```diff
@@ -120,7 +120,7 @@ HeapGraph convertHeapGraph(
 
   // Construct all the classes in the snapshot.
   final List<HeapGraphClassLive> classes =
-      List<HeapGraphClassLive>(graph.classes.length);
+      List<HeapGraphClassLive>.filled(graph.classes.length, null);
```
jacob314, line 123: this is a case where we should use `generate` as there is a for loop right bellow where we fill in the values for each class.

PR #2552, opened 2020-12-11
Fix formatting and cleanup handling of variable width columns.

jacob314:
Variable width columns that can have a width larger than their minWidth now always have exactly the same width.

### DIFF
--- a/packages/devtools_app/lib/src/table.dart
+++ b/packages/devtools_app/lib/src/table.dart
@@ -164,7 +164,7 @@ class FlatTableState<T> extends State<FlatTable<T>>
       for (var column in sortedColumns) {
         if (column.fixedWidthPx == null && column.minWidthPx != null) {
           // Width of this column if it was not clamped to its min width.
-          // We add column.minWidthPx to the avaialable width because
+          // We add column.minWidthPx to the available width because
           // available is currently not considering the space reserved for this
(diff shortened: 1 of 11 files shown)


packages/devtools_app/lib/src/table.dart
```diff
@@ -126,31 +126,80 @@ class FlatTableState<T> extends State<FlatTable<T>>
...
+      // be treated as unconstrained as their minWidth constraint is satisfied
+      // by the width given to all unconstrained columns.
+      // We use a greedy algorithm to accurately compute this by iterating
+      // through the columns from the smallest minWidth to largest minWidth
+      // incrementally adding columns where the minWidth constraint can be
+      // satisfied using the width given to unconstrained columns.
+      for (var column in sortedColumns) {
+        if (column.fixedWidthPx == null && column.minWidthPx != null) {
+          // Width of this column if it was not clamped to its min width.
+          // We add column.minWidthPx to the avaialable width because
```
kenzieschmoll, line 167: nit: typo available

kenzieschmoll (approved):

packages/devtools_app/lib/src/table.dart
jacob314: done

PR #2560, opened 2020-12-17
Cleanup in logging_controller and remove some unnecessary code.

kenzieschmoll:
Use existing AutoDispose code in logging_controller.dart and remove `preferences` from `DevToolsApp` constructor. `preferences` is a global var as of some recent code changes.

### DIFF
--- a/packages/devtools_app/lib/src/logging/logging_controller.dart
+++ b/packages/devtools_app/lib/src/logging/logging_controller.dart
@@ -13,6 +13,7 @@ import 'package:meta/meta.dart';
 import 'package:path/path.dart' as path;
 import 'package:vm_service/vm_service.dart';
 
+import '../auto_dispose.dart';
 import '../config_specific/logger/logger.dart' as logger;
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/logging/logging_controller.dart
```diff
@@ -354,6 +353,7 @@ class LoggingController
         summary: '${changedInfo.extension}: ${changedInfo.value}',
       ));
     } else if (e.extensionKind == 'Flutter.Error') {
+      print('logging error event in loggign controller');
```
devoncarew, line 356: Should we use a logger here?

Also, sp: `loggign`

devoncarew (approved):
lgtm though we should probably remove the use of a raw `print` call

packages/devtools_app/lib/src/logging/logging_controller.dart
kenzieschmoll: accidentally included. removed.

PR #2561, opened 2020-12-17
Add ErrorBadgeManager and UI for badging Screen tabs

kenzieschmoll:
This CL adds the skeleton for adding error-badging to devtools tabs. 

This is a checkpoint with the base support, and error handling for appropriate screens will be done in follow up PRs. The result will look like this:
Screen Shot 2020-12-15 at 4 35 23 PM
Screen Shot 2020-12-15 at 4 35 27 PM

### DIFF
--- a/packages/devtools_app/lib/src/error_badge_manager.dart
+++ b/packages/devtools_app/lib/src/error_badge_manager.dart
@@ -1,0 +1,49 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+import 'package:vm_service/vm_service.dart';
+
+import 'auto_dispose.dart';
+import 'globals.dart';
+import 'service_extensions.dart' as extensions;
+import 'vm_service_wrapper.dart';
+
+class ErrorBadgeManager extends DisposableController
+    with AutoDisposeControllerMixin {
+  // TODO(kenz): populate this map with screens as support is added
+  // (e.g. { LoggingScreen.id: ValueNotifier<int>(0) }
+  final _activeErrorCounts = <String, ValueNotifier<int>>{};
+
+  void vmServiceOpened(VmServiceWrapper service) {
+    // Ensure structured errors are enabled.
+    serviceManager.serviceExtensionManager.setServiceExtensionState(
+      extensions.structuredErrors.extension,
+      true,
+      true,
+    );
+
+    // Log Flutter extension events.
+    autoDispose(service.onExtensionEvent.listen(_handleExtensionEvent));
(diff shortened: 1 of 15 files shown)


devoncarew:
As a drive by comment on the look, did you try badging the lower right of the tab text? It might be alright to overlap the tab title (it looks good as is, but I was curious about how other options might look).

packages/devtools_app/lib/src/error_badge_manager.dart
```diff
@@ -0,0 +1,49 @@
...
+    // Log stderr events.
+    autoDispose(service.onStderrEvent.listen(_handleStdErr));
+  }
+
+  void _handleExtensionEvent(Event e) async {
+    // TODO(kenz): handle Flutter.Error extension events.
+  }
+
+  void _handleStdErr(Event e) {
+    // TODO(kenz): handle stderr events
```
jacob314, line 39: might want to highlight the logging page when these show up?

packages/devtools_app/lib/src/error_badge_manager.dart
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  ValueListenable<int> errorCountNotifier(String screenId) {
+    return _activeErrorCounts[screenId] ?? ValueNotifier<int>(0);
```
jacob314, line 43: I think ben added a class that is a ValueNotifier with a const Value. You'll want that for the case where the value is zero to make it a little clearer what is going on.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/error_badge_manager.dart
kenzieschmoll: @bkonyi does this sounds familiar? I looked through the vm developer code but didn't catch what jacob is referencing.

packages/devtools_app/lib/src/error_badge_manager.dart
kenzieschmoll: Yeah we should. Will handle this in a follow up, along with the `Flutter.error` events.

kenzieschmoll:
> As a drive by comment on the look, did you try badging the lower right of the tab text? It might be alright to overlap the tab title (it looks good as is, but I was curious about how other options might look).

This is what it would look like in the bottom left. IMO I think the top right is a little better because it doesn't crowd the blue selection indicator at the bottom of the tab
Screen Shot 2020-12-18 at 10 11 12 AM

packages/devtools_app/lib/src/error_badge_manager.dart
bkonyi: Yup, you're looking for `FixedValueListenable` in `devtools_app/lib/src/listenable.dart`.

Issue #2497, opened 2020-11-09
P1: Consider merging functionality from Timeline and Performance pages


kenzieschmoll:
@cobblest is running a UX study that should help us determine some direction for this item

cobblest:
Research Brief: https://docs.google.com/document/d/1MndVORQw4jHN8ob-VL2GrvnyxZ1EvAsR89lQJrr50Ag/edit#heading=h.x9snb54sjlu9

kenzieschmoll:
Based on findings from the UX studies, the following tasks should be completed to address this issue:
[ ]
[ ]
[ ]

Issue #1674, opened 2020-02-26
Devtools cached overly aggressively


jacob314:
Repro steps:
```
$ pub global activate devtools 0.1.13
$ pub global run devtools -p 1234
Access http://127.0.0.1:1234/ and confirm the the UI works and "DevTools version 0.1.13." is shown in the console

$ pub global activate devtools 0.1.15
$ pub global run devtools -p 1234
Access http://127.0.0.1:1234/, observe "Loading..." label stuck on the screen. "DevTools version 0.1.13." is shown in the console.
```
Apparently main.dart.js gets cached persistently and survives version upgrades/downgrades and gets into weird states due to version mismatches between different files/ajax APIs.

Suggested fix:
adding a cache busting query parameter to main.dart.js or changing our server configuration.

devoncarew:
Bumping this to the following milestone, but do people have ideas for how to address this?

I'd made some changes to the cache control headers in our devtools server (https://github.com/flutter/devtools/pull/1993), but either those didn't take effect or it wasn't a complete solution.

jacob314:
Let me try reducing the cache time from 300s to 0s. Currently you can get caching issues for the 300s while the data is still cached. The service worker added for flutter web may also be able to help mitigate any performance impact from reducing caching from 300s to 0s.

kenzieschmoll:
Is this issue still happening?

DanTup:
I think there are still some caching issues - I was just testing a release build locally and got errors for 404s on "widget.json". I couldn't find any references in the code (was removed in https://github.com/flutter/devtools/pull/2433/files). Reloading the page made no difference, but clicking the "Clear site data" button in Chrome devtools -> Application seems to have fixed it. In hindsight I should've checked where the widgets.json reference was before fixing it, since I'm not sure exactly what may have been cached for this to occur.

kenzieschmoll:
I'm also still seeing caching issues. Sometimes a hard refresh won't even fix it. I've been opening DevTools in an incognito window to work around it.

DanTup:
Yep, that's what I saw - though this button seems to fix it:

![Screenshot 2020-10-21 at 17 37 50](https://user-images.githubusercontent.com/1078012/96750751-3e058900-13c4-11eb-946d-12b5b67d2c8c.png)

That unregisters service workers, and I noticed there's a lot of caching stuff in the Flutter service worker - though I'm not sure if that's the cause.

DanTup:
[...]

PR #2562, opened 2020-12-21
Improvements to performance tooling from UXR

kenzieschmoll:
Addresses part of https://github.com/flutter/devtools/issues/2497:
- [x] move performance tab next to timeline tab
- [x] rename "event details" to CPU profiler on timeline page
- [x] rename performance page to "CPU profiler"
- [x] move bottom up profiler tab to position 0

### DIFF
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -32,8 +32,8 @@ import 'memory/memory_screen.dart';
 import 'network/network_controller.dart';
 import 'network/network_screen.dart';
 import 'notifications.dart';
-import 'performance/performance_controller.dart';
-import 'performance/performance_screen.dart';
+import 'profiler/profiler_screen.dart';
+import 'profiler/profiler_screen_controller.dart';
 import 'routing.dart';
 import 'scaffold.dart';
 import 'screen.dart';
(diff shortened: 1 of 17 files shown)


jacob314 (approved):
lgtm

jacob314 (commented):
Fyi @DanTup in case you are directly launching links to the timeline page from vscode.

DanTup:
> Fyi @DanTup in case you are directly launching links to the timeline page from vscode.

I am, thanks! I'll need to handle this based on the version. The current live version is 0.9.5+1. Is it safe for me to assume that >= 0.9.6 would include this, and anything < 0.9.5+1 should use the old URLs?

kenzieschmoll:
yes that is correct. See also https://github.com/flutter/devtools/pull/2563, which renames timeline to performance. so >= 0.9.6

"timeline" --> "performance"
"performance" --> "cpu-profiler"

DanTup:
Got it! I've got a fix ready to ship in a patch, though I'm a little worried about caching causing issues - I added some notes here:

https://github.com/flutter/devtools/issues/1674#issuecomment-754609493

If users see the same thing I did, they might end up getting errors because the old JS still loads and we use the new page IDs.

DanTup:
I've published a new Dart-Code with the new page IDs (selected based on the version). I'll send a PR that removes the service worker here shortly.

If you let me know when DevTools is released, I'll re-test using the published version. Thanks!

PR #2563, opened 2020-12-22
Rename Timeline to Performance and populate CPU profile on frame selection

kenzieschmoll:
Only functional change is populating a the CPU profile section of the timeline page (now performance page) when a flutter frame is selected from above. The remaining changes are renames from "timeline" to "performance". 

Fixes https://github.com/flutter/devtools/issues/2497


packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -966,6 +966,7 @@ class ServiceExtensionManager extends Disposer {
       }
     }
 
+    if (mainIsolate == null) return;
```
kenzieschmoll, line 969: @jacob314 this seemed to fix the flakes we were seeing. Do you see any problem with this change?

jacob314: seems reasonable as it would crash on the next line anyway without the return.

jacob314 (approved):
lgtm

PR #2564, opened 2020-12-22
Handle cases where the inspector controller is opened while service extensions are being registered more robustly.

jacob314:


packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -1016,6 +1030,28 @@ class ServiceExtensionManager extends Disposer {
...
 
+  Future<bool> waitForServiceExtensionAvailable(String name) {
+    if (isServiceExtensionAvailable(name)) return Future.value(true);
+
+    Completer<bool> createCompleter() {
+      // Listen for when the service extension is added and use it.
+      final completer = Completer<bool>();
+      final listenable = hasServiceExtension(name);
+      VoidCallback listener;
+      listener = () {
```
kenzieschmoll, line 1041: why not combine these two lines?

kenzieschmoll: oh nvm i see now - you are using it in the closure.

kenzieschmoll (approved):

PR #2565, opened 2020-12-22
Add scrollbar to flutter frames chart

kenzieschmoll:
Screen Shot 2020-12-22 at 10 48 44 AM

### DIFF
--- a/packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
+++ b/packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
@@ -42,6 +42,8 @@ class _FlutterFramesChartState extends State<FlutterFramesChart>
 
   static const outlineBorderWidth = 1.0;
 
+  static const scrollBarOffset = 10.0;
+
   PerformanceController _controller;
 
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
```diff
@@ -106,18 +108,21 @@ class _FlutterFramesChartState extends State<FlutterFramesChart>
         right: denseSpacing,
         bottom: defaultSpacing,
       ),
-      height: defaultChartHeight,
+      height: defaultChartHeight + scrollBarOffset,
```
jacob314, line 111: nit: maybe move scrollBarOffset to a util file so we remember to use the same const everywhere we need non-overlapping scrollbars?

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
kenzieschmoll: done

Issue #173, opened 2019-01-27
P1: Badge tabs with red dot / count for errors


devoncarew:
When the logging page is not active, and we receive framework error messages, we should badge the 'Logging' page with the number of framework error messages received. This will indicate to the user that there are messages in the logging view that they may want to see. When the logging view next becomes active, the badge count on that page can be cleared out.

<img width="67" alt="screen shot 2019-01-26 at 6 04 49 pm" src="https://user-images.githubusercontent.com/1269969/51795257-13f24900-2195-11e9-8016-f35349b3a8cc.png">

<img width="144" alt="screen shot 2019-01-26 at 6 05 33 pm" src="https://user-images.githubusercontent.com/1269969/51795258-18b6fd00-2195-11e9-910a-0866a34b5004.png">

kenzieschmoll:
Modifying this issue to track all cases where we should be badging tabs with errors:
e.g. overflow errors in inspector tab, log exceptions, etc

we should also support more detailed badging within these views (i.e. badge the specific row in the inspector that has the error, badge the specific log entries with exceptions/errors)

kenzieschmoll:
Copying from the Dart DevTools 2.0 PRD:

> When running DevTools, if there is an error on any of the tabs, e.g. an exception in the log or a failed network operation or a layout error, provide a red dot in the appropriate tabs. When the user clicks on a tab w/ an error on it, show a red dot on the appropriate element on that tab, e.g. the log entry with the error, the widget w/ the layout error, the failed network communication, etc.

kenzieschmoll:
Badging the memory tab for leaks is dependent upon us being able to definitively detect a leak (see https://github.com/flutter/devtools/issues/1437). Added a comment to that issue to note that we should badge the memory tab when we have the proper signal.

This issue can be closed out once the performance tab badging CL lands.

PR #2567, opened 2020-12-22
Badge the Network tab when we receive failed network requests.

kenzieschmoll:
Screen Shot 2020-12-22 at 2 40 25 PM
Part of https://github.com/flutter/devtools/issues/173

### DIFF
--- a/packages/devtools_app/lib/src/network/network_controller.dart
+++ b/packages/devtools_app/lib/src/network/network_controller.dart
@@ -320,7 +320,7 @@ class NetworkController
   void filterData(QueryFilter filter) {
     serviceManager.errorBadgeManager.clearErrors(NetworkScreen.id);
     if (filter == null) {
-      final requests = List.from(_requests.value.requests);
+      final requests = List<NetworkRequest>.from(_requests.value.requests);
       requests.forEach(_checkForError);
       filteredData.value = requests;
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -343,26 +347,45 @@ class NetworkController
             final caseInsensitiveSubstring = substring.toLowerCase();
             final matchesUri =
                 r.uri.toLowerCase().contains(caseInsensitiveSubstring);
-            if (matchesUri) return true;
+            if (matchesUri) {
```
jacob314, line 350: these four blocks all look the same.
Maybe a helper that takes `bool checkMatch(r, textToMatch, caseInsensitiveSubstring)`

jacob314: alternately just run
filteredData.value.forEach(checkForError)
after building up the filteredData. That seems a bit easier to get all the cases right.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/network/network_controller.dart
kenzieschmoll: done

Issue #2559, opened 2020-12-17
Can't use 0.9.3+4 and installing newer versions of dartdev tool fails


JulianBissekkou:
The current version 0.9.3+4 is not able to scroll enough on the timeline which makes it hard to debug performance issues.
I saw that this issue is already fixed in this PR https://github.com/flutter/devtools/pull/2447, and I think you shipped the fix in 0.9.4.

I am installing 0.9.4 with this command : `pub global activate devtools 0.9.4`
which produces this error:
```
Because devtools >=0.9.4 <0.9.5 depends on devtools_server 0.9.4 which depends on intl ^0.17.0-nullsafety.1, devtools >=0.9.4 <0.9.5 requires intl ^0.17.0-nullsafety.1.
So, because intl >=0.17.0-nullsafety requires SDK version >=2.11.0-180.0.dev <3.0.0 and pub global activate depends on devtools 0.9.4, version solving failed.
```

What is the issue here? 🤔 

Might be related to this issue:
https://github.com/flutter/devtools/issues/2512

kenzieschmoll:
try `pub global activate devtools` without specifying a version number. Also can you add the output of `flutter --version`? Thanks

JulianBissekkou:
I already tried `pub global activate devtools` and `flutter pub global activate devtools`. This is the output:

```
Package devtools is currently active at version 0.9.3+4.
Resolving dependencies... (1.8s)
... 
Precompiling executables... (2.5s)
Precompiled devtools:devtools.
Installed executable devtools.
Activated devtools 0.9.3+4.
```


Here are the flutter and dart version outputs:
`flutter --version`:

```
Flutter 1.22.5 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 7891006299 (8 days ago) • 2020-12-10 11:54:40 -0800
Engine • revision ae90085a84
Tools • Dart 2.10.4
```

`pub --version`:
```
Pub 2.10.4
```

This might be interesting as well:
`pub global list`
```
aqueduct 3.2.0
coverage 0.12.4
devtools 0.9.3+4
intl_utils 1.9.0
stagehand 3.3.7
webdev 2.5.4
```

kenzieschmoll:
Can you try deactivating and reactivating the package?
`pub global deactivate devtools`
`pub global activate devtools`

JulianBissekkou:
Yes, but it didn't work. 

Here the output:
```
➜  ~ pub global deactivate devtools
[...]

PR #2571, opened 2020-12-28
Remove dependencies on nullsafe versions of packages from devtools.

jacob314:
devtools_server needs to be compatible with old Dart versions.
Fixes https://github.com/flutter/devtools/issues/2559


jacob314:
Fyi @terrylucas. We need to avoid any dependencies on nullsafe packages to pub global activate devtools with stable flutter.

helin24 (approved):

kenzieschmoll (commented):
LGTM

PR #2576, opened 2020-12-29
Badge inspector tab for structured inspector errors (Flutter.error)

kenzieschmoll:
Screen Shot 2020-12-29 at 11 13 11 AM

Part of https://github.com/flutter/devtools/issues/173

### DIFF
--- a/packages/devtools_app/lib/src/error_badge_manager.dart
+++ b/packages/devtools_app/lib/src/error_badge_manager.dart
@@ -7,6 +7,7 @@ import 'package:vm_service/vm_service.dart';
 
 import 'auto_dispose.dart';
 import 'globals.dart';
+import 'inspector/inspector_screen.dart';
 import 'listenable.dart';
 import 'logging/logging_screen.dart';
 import 'network/network_screen.dart';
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/error_badge_manager.dart
```diff
@@ -36,9 +38,14 @@ class ErrorBadgeManager extends DisposableController
   }
 
   void _handleExtensionEvent(Event e) async {
-    // TODO(jacobr): badge inspector for appropriate errors.
     if (e.extensionKind == 'Flutter.Error') {
       incrementBadgeCount(LoggingScreen.id);
+
+      final json = e.extensionData.data;
+      final objectId = json['objectId'] as String;
+      if (objectId?.contains('inspector-') ?? false) {
```
kenzieschmoll, line 46: will this check include things we don't want to badge the inspector for? @jacob314

jacob314: All object ids sent with errors will start with `inspector-` currently so this won't really check for much. I think just assuming any case with an objectId is an inspector error is fine for now. We could add a TODO to verify that the objectId is referring to an Element or Widget (may require another round trip unless the json also contains the object type which it may.

jacob314 (approved):
lgtm

PR #2577, opened 2020-12-29
Remove spurious timeout from treemap_test.

jacob314:


packages/devtools_app/test/treemap_test.dart
```diff
@@ -168,8 +168,6 @@ void main() {
           find.byKey(treemapKey),
           matchesGoldenFile('goldens/treemap_sizes.png'),
         );
-        // Await delay for golden comparison.
-        await tester.pumpAndSettle(const Duration(seconds: 2));
```
kenzieschmoll, line 172: did you verify locally that the test still completes as expected without this? this test doesn't run on the bots because the mac bots are disabled, so I want to make sure that this has the intended results. IIRC this await was necessary because without it, the test wouldn't fail even if it should have.

jacob314: High level: should we switch all these tests to compute goldens only for linux as that is what our build bots can run or do we expect to be able to support macos on our build bots soon? I've verified that await isn't needed but the goldens were all out of date and there was another test in this suite that was skipped on macOs for no reason.
If a pump and settle is ever needed in a test after all other logic in the test has completed that means we have significant bugs in our widget code (e.g. futures awaited after a widget has closed).

kenzieschmoll: I'm actually not sure on the status of this now that we are on github actions. @devoncarew can you add anything regarding how different platforms (macos, windows, linux) are tested with github actions?

kenzieschmoll:
lgtm_army

Issue #1314, opened 2019-11-07
Inspector throwing RenderFlex overflow on initial load


kenzieschmoll:
Seen on both flutter desktop and flutter web:
```
flutter: ══╡ EXCEPTION CAUGHT BY RENDERING LIBRARY ╞═════════════════════════════════════════════════════════
flutter: The following assertion was thrown during layout:
flutter: A RenderFlex overflowed by 7130 pixels on the right.
flutter: 
flutter: The relevant error-causing widget was:
flutter:   Row
flutter:   file:///Users/kenzieschmoll/develop/devtools/packages/devtools_app/lib/src/inspector/flutter/inspector_tree_flutter.dart:536:18
flutter: 
flutter: The overflowing RenderFlex has an orientation of Axis.horizontal.
flutter: The edge of the RenderFlex that is overflowing has been marked in the rendering with a yellow and
flutter: black striped pattern. This is usually caused by the contents being too big for the RenderFlex.
flutter: Consider applying a flex factor (e.g. using an Expanded widget) to force the children of the
flutter: RenderFlex to fit within the available space instead of being sized to their natural size.
flutter: This is considered an error condition because it indicates that there is content that cannot be
flutter: seen. If the content is legitimately bigger than the available space, consider clipping it with a
flutter: ClipRect widget before putting it in the flex, or using a scrollable container rather than a Flex,
flutter: like a ListView.
flutter: The specific RenderFlex in question is: RenderFlex#6a870 OVERFLOWING:
flutter:   needs compositing
flutter:   creator: Row ← ClipRect ← Padding ← CustomPaint ← InspectorRowContent ← DefaultTextStyle ←
flutter:     AnimatedDefaultTextStyle ← _InkFeatures-[GlobalKey#ee7c3 ink renderer] ←
flutter:     NotificationListener<LayoutChangedNotification> ← PhysicalModel ← AnimatedPhysicalModel ← Material
flutter:     ← ⋯
flutter:   parentData: <none> (can use size)
flutter:   constraints: BoxConstraints(w=2080.0, h=24.0)
flutter:   size: Size(2080.0, 24.0)
flutter:   direction: horizontal
flutter:   mainAxisAlignment: start
flutter:   mainAxisSize: min
flutter:   crossAxisAlignment: center
flutter:   textDirection: ltr
flutter:   verticalDirection: down
flutter:   textBaseline: alphabetic
[...]

PR #2579, opened 2020-12-30
Inspector light weight integration test support.

jacob314:
Adds golden image tests for the inspector and light weight integration tests using the integration test running with real async.
Also fixes https://github.com/flutter/devtools/issues/1314 by using RichText instead of multiple Text widgets to avoid overflow issues.


kenzieschmoll (approved):

packages/devtools_app/lib/src/inspector/diagnostics.dart
```diff
@@ -38,38 +38,54 @@ class DiagnosticsNodeDescription extends StatelessWidget {
...
-        output.add(Text(match.group(1), style: textStyle));
-        output.add(Text(match.group(3), style: textStyle));
+        yield TextSpan(text: match.group(1), style: textStyle);
+        yield TextSpan(text: match.group(3), style: textStyle);
         return;
       }
     }
     if (description?.isNotEmpty == true) {
-      output.add(Text(description, style: textStyle));
+      yield TextSpan(text: description, style: textStyle);
```
kenzieschmoll, line 67: TIL that `yield` is a thing.

PR #2580, opened 2020-12-30
Badge performance tab when UI jank is detected and add a setting to enable/disable this functionality.

kenzieschmoll:
Screen Shot 2020-12-30 at 10 36 52 AM

Added a setting to the dialog to enable/disable this feature.
Screen Shot 2020-12-30 at 10 42 24 AM

Also changed the icon to better reflect what the dialog encompasses:
Screen Shot 2020-12-30 at 10 43 45 AM

Part of https://github.com/flutter/devtools/issues/173.

### DIFF
--- a/packages/devtools_app/lib/src/performance/performance_controller.dart
+++ b/packages/devtools_app/lib/src/performance/performance_controller.dart
@@ -85,6 +85,9 @@ class PerformanceController
       _httpTimelineLoggingEnabled;
   final _httpTimelineLoggingEnabled = ValueNotifier<bool>(false);
 
+  ValueListenable<bool> get badgeTabForJankyFrames => _badgeTabForJankyFrames;
+  final _badgeTabForJankyFrames = ValueNotifier<bool>(true);
+
   // TODO(kenz): switch to use VmFlagManager-like pattern once
   // https://github.com/dart-lang/sdk/issues/41822 is fixed.
   /// Recorded timeline stream values.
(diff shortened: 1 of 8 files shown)


packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -85,6 +85,9 @@ class PerformanceController
       _httpTimelineLoggingEnabled;
   final _httpTimelineLoggingEnabled = ValueNotifier<bool>(false);
 
+  ValueListenable<bool> get badgeTabForJankyFrames => _badgeTabForJankyFrames;
+  final _badgeTabForJankyFrames = ValueNotifier<bool>(true);
```
jacob314, line 89: should we default this to true in profile builds and false in debug builds?

packages/devtools_app/lib/src/performance/performance_screen.dart
```diff
@@ -308,17 +309,32 @@ class TimelineConfigurationsDialog extends StatelessWidget {
...
+            ...dialogSubHeader(theme, 'Advanced Recorded Timeline Streams'),
             ..._advancedStreams(theme),
+            const SizedBox(height: denseSpacing),
+            ...dialogSubHeader(theme, 'Additional Settings'),
+            Row(
+              children: [
+                NotifierCheckbox(notifier: controller.badgeTabForJankyFrames),
+                Text(
+                  'Badge Performance tab when Flutter UI jank is detected',
+                  // Specify font size to match style of other RichText widgets
```
jacob314, line 332: should we just use RichText instead of text to avoid this?
Hard coding font sizes will make it hard for us to later support custom font sizes for DevTools which I believe @helin24 was working on.

helin24: Yes I'd like to add custom font sizes! (mainly for the intellij embedded browser) That change is stuck waiting for a refactor of some controllers though

kenzieschmoll: done

packages/devtools_app/lib/src/performance/performance_controller.dart
kenzieschmoll: done

jacob314 (approved):

Issue #1927, opened 2020-05-06
P2: Support syntax highlighting in the debugger code view


kenzieschmoll:
Originally from https://github.com/flutter/devtools/issues/1648

PR #2583, opened 2021-01-05
Added syntax highlighting to CodeView

bkonyi:
**Dark theme:**
Screen Shot 2021-01-05 at 11 42 51 AM

**Light theme:**
Screen Shot 2021-01-05 at 11 43 08 AM

Fixes https://github.com/flutter/devtools/issues/1927

### DIFF
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -213,9 +213,7 @@ class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
     final theme = Theme.of(context);
 
     final lines = <TextSpan>[];
-    final style = theme.textTheme.bodyText2.copyWith(
-      fontFamily: 'RobotoMono',
-    );
+    final style = fixedFontStyle(context);
 
     // Ensure the syntax highlighter has been initialized.
     // TODO(bkonyi): process source for highlighting on a separate thread.
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -57,7 +57,7 @@ class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
   void initState() {
     super.initState();
 
-    SyntaxHighlighter.initialize().then((_) => _initScriptInfo());
+    _initScriptInfo();
     verticalController = LinkedScrollControllerGroup();
     gutterController = verticalController.addAndGet();
     textController = verticalController.addAndGet();
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -213,9 +213,13 @@ class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
     final theme = Theme.of(context);
 
     final lines = <TextSpan>[];
+    final style = theme.textTheme.bodyText2.copyWith(
+      fontFamily: 'RobotoMono',
+    );
 
     // Ensure the syntax highlighter has been initialized.
-    if (highlighter != null) {
+    // TODO(bkonyi): process source for highlighting on a separate thread.
+    if (script.source.length < 500000 && highlighter != null) {
       final highlighted = highlighter.highlight(context);
 
       // Look for [TextSpan]s which only contain '\n' to manually break the
@@ -226,9 +230,7 @@ class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
         if (span.toPlainText() == '\n') {
           lines.add(
             TextSpan(
-              style: theme.textTheme.bodyText2.copyWith(
-                fontFamily: 'RobotoMono',
-              ),
+              style: style,
               children: currentLine,
             ),
           );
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -42,21 +43,21 @@ class CodeView extends StatefulWidget {
...
 
   ScriptRef get scriptRef => widget.scriptRef;
 
   @override
   void initState() {
     super.initState();
 
-    _initScriptInfo();
-
+    SyntaxHighlighter.initialize().then((_) => _initScriptInfo());
```
devoncarew, line 60: It looks like the initialization happens asynchronously; I assume it wouldn't be ideal if the highlighter instance is created before the SyntaxHighlighter is inited. Perhaps move the async initialization elsewhere? Into some async debugger page initialization, or, into some other async initialization for the app>

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -66,8 +67,10 @@ class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
   }
 
   void _parseScriptLines() {
-    // Parse the source into lines.
-    lines = script.source?.split('\n') ?? [];
+    // Create a new SyntaxHighlighter with the script's source in preparation
+    // for building the code view.
+    highlighter = SyntaxHighlighter(source: script?.source ?? '');
```
devoncarew, line 72: Do we pay the cost of parsing the source here? I assume that we highlight the whole file and don't do something like lazily highlight sections.

If we highlight the whole file, we might want a cut-out for very large files. dart:html for instance is (from memory) 1.5MB. I'd assume we wouldn't be able to highlight that performantly. You might want to do a quick check for very large files and add an arbitrary cut-off for large files.

devoncarew (commented):
I added two drive by comments -

packages/devtools_app/lib/src/debugger/codeview.dart
bkonyi: No, we incur that cost at [line 219](https://github.com/flutter/devtools/pull/2583/files#diff-486524c0f38b9588774d5a6c3942a7d646cbd82e8398047ed0a0a541b7b84e1dR219) since we can't generate the `TextSpan`s without a `BuildContext`. You're correct: we do highlight the whole file at this point since doing anything lazy here would probably require a complete overhaul of `CodeView`.

I just looked into seeing if there was performance issues analyzing `dart:html`, but it looks like it's not included in the library list when inspecting DevTools on web (is that expected?). A cut-off sounds like a reasonable idea, but we should probably figure out what "very large" should be defined as.

packages/devtools_app/lib/src/debugger/codeview.dart
bkonyi: Good point. It looks like the only good place to do this is before we call `runApp`, so I've gone ahead and done that there.

packages/devtools_app/lib/src/debugger/codeview.dart
devoncarew: Ah, `dart:html` would likely only be visible if debugging a flutter web / dart web app.

You can try `package:flutter/src/material/icons.dart`; I'm not sure how it compares to dart:html, but it is ~17k lines line / 1.1MB.

bkonyi: That's what I was doing. I used DevTools on web to self-inspect and didn't see `dart:html`. That seems strange to me.

Anyway, I went ahead and opened the icons file using DevTools for desktop and it did take a few seconds to load (I'm sure it'll be much worse on web). I'll use the character count of that file to come up with some sort of upper limit.

bkonyi: I gave some more thought about how to parse the script asynchronously so we could avoid introducing an upper limit but I think we'd need to push that work onto another thread, which isn't exactly trivial at this point with no isolate support on web. I've set an upper limit of 500,000 characters for now and added a TODO to revisit this later, potentially when Flutter Web has `compute` support.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -210,6 +212,50 @@ class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
   Widget buildCodeArea(BuildContext context) {
     final theme = Theme.of(context);
 
+    final lines = <TextSpan>[];
+    final style = theme.textTheme.bodyText2.copyWith(
+      fontFamily: 'RobotoMono',
+    );
```
kenzieschmoll, line 218: use `fixedFontStyle` from theme.dart instead. Feel free to add an optional param for fontSize that defaults to 13.0 if you need a different size.

packages/devtools_app/lib/src/debugger/syntax_highlighter.dart
```diff
@@ -0,0 +1,300 @@
...
+import 'package:flutter/material.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter/services.dart' show rootBundle;
+
+import '../theme.dart';
+import 'span_parser.dart';
+
+class SyntaxHighlighter {
+  SyntaxHighlighter({
+    this.source,
```
kenzieschmoll, line 18: no need to use trailing comma if it fits on one line

kenzieschmoll (approved):

PR #2585, opened 2021-01-07
Make devtools title reflect app type.

jacob314:


jacob314 (commented):
This CL does not handle offline snapshots. We should store the app type with the snapshot to resolve that issue.

packages/devtools_app/web/index.html
```diff
@@ -12,7 +12,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
 
-    <title>Dart DevTools</title>
+    <title></title>
```
devoncarew, line 15: DBC: Should we pre-populate w/ something neutral like 'DevTools'?

packages/devtools_app/lib/src/globals.dart
```diff
@@ -31,6 +31,13 @@ SurveyService get surveyService => globals[SurveyService];
 
 PreferencesController get preferences => globals[PreferencesController];
 
+String generateDevToolsTitle() {
+  if (!serviceManager.hasConnection) return '';
+  final connectedApp = serviceManager.connectedApp;
+  if (!connectedApp.appTypeKnown) return '';
+  return connectedApp.isFlutterAppNow ? 'Flutter DevTools' : 'Dart DevTools';
```
devoncarew, line 38: `DevTools • Flutter` and `DevTools • Dart`?

`DevTools / Flutter` and `DevTools / Dart`?

`DevTools for Flutter` and `DevTools for Dart`?

devoncarew (commented):
Two thoughts for title ideas -

packages/devtools_app/lib/src/globals.dart
kenzieschmoll: I'm in favor of `Flutter DevTools` and `Dart DevTools`. Since those are similar to `Chrome DevTools`, I think we can leverage some familiarity from our users.

packages/devtools_app/lib/src/scaffold.dart
```diff
@@ -301,15 +302,19 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
...
           handleDrop: _importController.importData,
-          child: Scaffold(
-            appBar: widget.embed ? null : _buildAppBar(),
-            body: TabBarView(
-              physics: defaultTabBarViewPhysics,
-              controller: _tabController,
-              children: tabBodies,
+          child: Title(
+            title: title,
+            color: Colors.blue,
```
kenzieschmoll, line 307: Where is this blue color? can you add a screenshot to the PR description?

packages/devtools_app/lib/src/globals.dart
jacob314: `DevTools • Flutter`  would be nice as we can display `DevTools` first and then only add the type of app being debugged once we know it. Flickering from `DevTools` to `Flutter DevTools` would be more distracting.
On the other hand I like `Flutter DevTools` and `Dart DevTools` better independent of the UI flickering constraints and as long as 95% of users never see the flicker it is probably fine.

packages/devtools_app/lib/src/scaffold.dart
jacob314: The blue color is a hack to make this prototype work. The color only matters on android. Switched to mirror what WidgetsApp does and added a comment.

Btw,  the title matters to change the title displayed in the web browser tab.

packages/devtools_app/web/index.html
jacob314: @kenzieschmoll any thoughts on this? I left this blank to minimize flicker. Now you just go from the URL to Flutter DevTools once the app loads.
If we pre-populate we will go from 
URL -> DevTools -> Flutter DevTools
If we decide to display the name as
`DevTools • Flutter` then we should definitely pre-populate.

kenzieschmoll: I am in favor of minimizing the flicker. May be a little overkill, but also could be simple with Flutter animations - what if we added a fade out animation on "DevTools" when connecting to an app. That would make the change look more intentional.

jacob314: There isn't really that much noticeable flicker. Here are a couple demos where we show a generic connect message and avoid displaying an app title until the connection is established. We don't show tabs until the connection is established so this is generally consistent.
Large GIF (1034x556)

open_connected

kenzieschmoll: LGTM this looks good

kenzieschmoll (approved):
LGTM - are we landing now or waiting until 2.0 to reflect the "rename"

kenzieschmoll:
@csells - are you okay with us landing this to be included in Friday's release? We'll want this for the recording.

PR #2586, opened 2021-01-07
Remove use of Flutter service worker

DanTup:
Setting `--pwa-strategy=none` just writes an empty service worker file and didn't seem to unregister it, so I wasn't sure if that would cause any issues and added code to explicitly unregister it (I don't know if this is necessary, but it seemed safest).

@jacob314


jacob314 (approved):
lgtm

PR #2592, opened 2021-01-11
Migrate to new material buttons

kenzieschmoll:
See flutter.dev/go/material-button-migration-guide

No functional changes - just migrating away from deprecated classes


HansMuller (approved):
NICE
LGTM

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -653,25 +653,25 @@ class FilterButton extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final colorScheme = Theme.of(context).colorScheme;
+    final theme = Theme.of(context);
     return RoundedOutlinedBorder(
       child: SizedBox(
-        height: defaultButtonHeight,
+        height: theme.buttonTheme.height,
```
HansMuller, line 659: Please don't use ButtonTheme, it's obsolete. The component defaults are documented but are not readily available via the widgets or themes. The default height for TextButton is 64 per minimumSize in https://api.flutter.dev/flutter/material/TextButton/defaultStyleOf.html.

packages/devtools_app/lib/src/theme.dart
```diff
@@ -25,45 +25,62 @@ ThemeData themeFor({
...
     accentColor: devtoolsBlue[400],
     backgroundColor: devtoolsGrey[600],
-    canvasColor: background,
     toggleableActiveColor: devtoolsBlue[400],
-    selectedRowColor: devtoolsBlue[600],
-    buttonTheme: theme.buttonTheme.copyWith(minWidth: buttonMinWidth),
+    canvasColor: background,
     scaffoldBackgroundColor: background,
     colorScheme: theme.colorScheme.copyWith(background: background),
+    buttonTheme: theme.buttonTheme.copyWith(
```
HansMuller, line 70: Specifying the buttonTheme shouldn't be needed. To change the minimum size of TextButtons, specify minimumSize for TextButtonTheme below.

packages/devtools_app/lib/src/common_widgets.dart
kenzieschmoll: removed all uses of buttonTheme

packages/devtools_app/lib/src/theme.dart
kenzieschmoll: done. Is there a way to set the default button height? I only saw a way to set the min, so had to enforce button height by wrapping certain buttons in SizedBoxes throughout devtools

HansMuller: You're right, currently there's only support for specifying a minimumSize. Using SizedBox is OK for forcing a particular height.  If you think we should add a fixedSize ButtonStyle parameter to ButtonStyle and the button themes, please file an issue.  It seems reasonable to support that, and it's true that we used to.

kenzieschmoll: filed https://github.com/flutter/flutter/issues/73741

PR #2593, opened 2021-01-11
Add timeline grid to CPU Profiler flame chart

kenzieschmoll:
Screen Shot 2021-01-11 at 9 50 34 AM

TODO in a follow up CL: snap timestamps to round values


jacob314 (approved):
lgtm
It would be nice to create a golden image test for this that uses canned json data.

PR #2599, opened 2021-01-12
Add tooltips to CPU profiler column titles.

kenzieschmoll:
Call Tree Total and Self time
Screen Shot 2021-01-12 at 1 21 54 PM
Screen Shot 2021-01-12 at 1 21 48 PM

Bottom up Total and Self time
Screen Shot 2021-01-12 at 1 21 40 PM
Screen Shot 2021-01-12 at 1 21 34 PM

Definitions taken straight from public docs: https://flutter.dev/docs/development/tools/devtools/cpu-profiler


jacob314 (approved):
lgtm

Issue #2480, opened 2020-11-02
Network tab missing requests


sm2017:
I am using DevTools to inspect network, I figured out that it not shown all requests, I am sure that I clear filters and it must show all requests, but when my app send multiple request in sub-second, network inspection only shows some of them, not all

I am using latest version of DevTools, dart, flutter, android studio, and SDKs
I only tested with simulators and in windows 10

kenzieschmoll:
A few things to check:

1) if you add print logging to your app, are the expected number of requests being sent?
2) can you see http events in the DevTools timeline, and if so, are there the expected number of requests there?
3) http logging is only enabled once you connect devtools to your app and either navigate to the timeline page or start recording from the network page. Are you connected to DevTools and recording on the network page before the requests are sent from your app?

Can you please paste the output of `flutter --version` here as well as the devtools version you are using? Thanks

kenzieschmoll:
CC @bkonyi

sm2017:
@kenzieschmoll 
1. Yes
 2. I see, but no all of them, when in sub-seconds we have multiple request, inspection lost somes
3. Yes, I connected before sending request, I also stop/started app and devtools, I think there is a bug in devtools

For `flutter --version `, I will send later, currently I have no access to my PC, but I upgrade to latest version a week a go

sm2017:
@kenzieschmoll the output of `flutter --version` is here

```
Flutter 1.22.2 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 84f3d28555 (3 weeks ago) • 2020-10-15 16:26:19 -0700
Engine • revision b8752bbfff
Tools • Dart 2.10.2
```

bkonyi:
I'll dig into this a bit deeper before EOY.

Issue #2398, opened 2020-10-02
Unable to inspect response bodies on Network page


gustavogamino:
As described at the [Flutter 1.22 announcement - Updated Network Page in DevTools](https://medium.com/flutter/announcing-flutter-1-22-44f146009e5f#10ac), we should be able to inspect network calls after getting the newest flutter version using `flutter upgrade`, but I'm not able to locate/use the feature.

I'm currently running devtools 0.9.2, so I tried to force the update to the latest release by running `pub global activate devtools 0.9.3` but it fails trying to resolve dependencies:
```shell
Because devtools >=0.9.3 depends on devtools_shared 0.9.3 which doesn't match any versions, devtools >=0.9.3 is forbidden.
So, because pub global activate depends on devtools 0.9.3, version solving failed.
pub finished with exit code 1
```
You can check the same error on pub.dev here https://pub.dev/documentation/devtools/0.9.3/log.txt

So what I tried instead was to clone this repo and build from `v0.9.3` tag and `master` branch. In both cases the build passes fine, but again the feature doesn't seem to be there. Every request on the Network page never completes, so no `Response` tab appears and `Header` sections are empty.

While running the app in debug mode one of the exceptions shown was this one:
```
Error: Assertion failed: file:///.../devtools/packages/devtools_app/lib/src/http/http_request_data.dart:62:16
requestEndEvent == null
is not true
    at Object.throw_ [as throw] (http://localhost:63874/dart_sdk.js:4333:11)
    at Object.assertFailed (http://localhost:63874/dart_sdk.js:4276:15)
    at Function.fromTimeline (http://localhost:63874/packages/devtools_app/src/ui/gtags.dart.lib.js:68892:48)
    at network_controller.NetworkController.new.processNetworkTrafficHelper
    (http://localhost:63874/packages/devtools_app/src/ui/gtags.dart.lib.js:69224:61)
    at network_controller.NetworkController.new.processNetworkTraffic
    (http://localhost:63874/packages/devtools_app/src/ui/gtags.dart.lib.js:69254:36)
    at network_service.NetworkService.new.refreshNetworkData
    (http://localhost:63874/packages/devtools_app/src/ui/gtags.dart.lib.js:68569:32)
    at refreshNetworkData.next (<anonymous>)
    at http://localhost:63874/dart_sdk.js:37669:33
    at _RootZone.runUnary (http://localhost:63874/dart_sdk.js:37523:58)
[...]

PR #2602, opened 2021-01-13
Correctly process large HTTP responses in network profiler

bkonyi:
The network profiler was assuming that a response body would only ever
consist of a single chunk. This was resulting in us trying to decode a
partial response, causing a FormatException to be thrown and halting
processing of HTTP events.

This change stitches together all of the response body events for a
given response before trying to decode it. In addition, the
FormatException is caught and ignored in order to allow for non-UTF8
responses.

Fixes both https://github.com/flutter/devtools/issues/2398 and https://github.com/flutter/devtools/issues/2480

### DIFF
--- a/packages/devtools_app/lib/src/http/http_request_data.dart
+++ b/packages/devtools_app/lib/src/http/http_request_data.dart
@@ -82,7 +82,9 @@ class HttpRequestData extends NetworkRequest {
     }
 
     try {
-      responseBody = utf8.decode(encodedData);
+      if (encodedData.isNotEmpty) {
+        responseBody = utf8.decode(encodedData);
+      }
     } on FormatException {
       // Non-UTF8 response.
     }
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -69,18 +69,24 @@ class HttpRequestData extends NetworkRequest {
...
-        responseBody = utf8.decode(encodedData);
-        break;
+        encodedData.addAll((traceEvent.args['data'] as List).cast<int>());
       }
     }
 
+    try {
+      responseBody = utf8.decode(encodedData);
+    } on FormatException {
+      // Non-UTF8 response.
```
kenzieschmoll, line 87: Should we do something to handle non UTF8 responses so that the user has some indication that we did receive a response? What is an example of a non-utf8 response?

bkonyi: An image is an example of a non-UTF8 response. If we try and decode that, it's certain to blow up.

kenzieschmoll (approved):

PR #2605, opened 2021-01-13
Events and extension events in event pane, added chart selection/hover card.

terrylucas:
Added event and extension events to the event pane.  The event pane is divided into four sections:

image

The memory chart Event Pane appears as:

image

Events are displayed as a single event e.g., 

image

or aggregate events e.g.,

image

Aggregate events, as the name implies, collects all the events nearest a particular timestamp (tick) and displays the events to the x-axis' closest tick.

In addition to Flutter events users can post their own events using the dart:developer package postEvent method.  Add the following import:

`import 'dart:developer' as developer;
`

and method:

```
  void devToolsPostEvent(String eventName, Map<String, Object> eventData) {
    developer.postEvent('DevTools.Event_$eventName', eventData);
  }
```

to your application, then calls to **devToolsPostEvent** e.g.,

```
  Widget recordLoadedImage(ImageChunkEvent imageChunkEvent, String imageUrl) {
 
    // Record the event in the memory event pane.
    devToolsPostEvent('MyFirstApp', { 'method': 'recordLoadedImage', 'param': imageUrl });

    if (imageChunkEvent == null) return null;

    ...

  }

```

Will record when the event occurred in the memory history chart e.g.,

image

A lot of Flutter and application custom events (MyFirstApp) were processed in the event pane.  Clicking on a point in the chart will display the values and events collected at a particular timestamp e.g.,

image

Clicking on this timestamp causes the hover card to appear:

image

At timestamp 13:57:41 there were 2 events:

image

Expanding the "2 Events" tile displays the two events:

image

image

The two events are:

1. MyFirstApp generated when the following method was called.

    devToolsPostEvent('MyFirstApp', { 'method': 'recordLoadedImage', 'param': imageUrl });
2. Flutter.ImageSizesForFrame
 
Additionally, the data collected (RSS, capacity, used, external, etc.) are also displayed in the hover card.

Clicking or pressing the escape key will close the hover card.  If the Android Memory chart is visible then those values, collected at that timestamp, would also be displayed in the hover card e.g.,

image


jacob314:
There are a couple of minor analysis errors to fix to help the bots go green
```
   info - Unused import: 'dart:convert' at lib/src/memory/memory_screen.dart:5:8 - (unused_import)
          Try removing the import directive.
   info - DO use curly braces for all flow control structures at lib/src/memory/memory_screen.dart:1160:7 - (curly_braces_in_flow_control_structures)
```

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -398,6 +454,19 @@ class ChartPainter extends CustomPainter {
     }
   }
 
+  void drawSelection(Canvas canvas, double x, y) {
```
jacob314, line 457: missing type for y.

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -398,6 +454,19 @@ class ChartPainter extends CustomPainter {
...
+    final paint = Paint()
+      ..strokeWidth = 2.0
+      ..color = colorScheme.hoverSelectionBarColor;
+
+    // Draw the yellow vertical selection bar.
```
jacob314, line 462: nit: remove comment that this is yellow.

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -525,6 +594,76 @@ class ChartPainter extends CustomPainter {
...
+        final path = symbolPathToDraw.shift(
+          Offset(
+            x - characteristics.width / 2,
+            y - characteristics.height / 2,
+          ),
+        );
+        canvas.drawPath(path, paint);
+        break;
+      default:
+        logger.log(
```
jacob314, line 660: maybe add an assert here.

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -208,6 +208,15 @@ class ChartController extends DisposableController
     _title = value;
   }
 
+  final ValueNotifier<TapNotifier> _tapNotifier =
+      ValueNotifier<TapNotifier>(TapNotifier.empty());
+
+  ValueNotifier<TapNotifier> get tapNotifier => _tapNotifier;
+
+  void setTapNotifier(TapNotifier tap) {
```
jacob314, line 216: rather than having a getter and a setter here, why not just make tapNotifier public?

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -556,4 +565,53 @@ class ChartController extends DisposableController
...
+      timestampedIndex = index + firstVisibleIndex;
+    }
+
+    // If index is left side, negative, that has yet to have any data
+    // return the first timestamp.
+    return timestampedIndex >= 0 ? timestampedIndex : 0;
+  }
+}
+
+/// When tap occurs in a chart compute the timestamp and index clicked.
```
jacob314, line 593: TapNotifier seems like a confusing name as this doesn't do any notifying.
Seems more like a TapLocation.

Word compute also seems wrong as the timestamp is passed in not computed.

packages/devtools_app/lib/src/charts/chart_trace.dart
```diff
@@ -33,13 +44,23 @@ class PaintCharacteristics {
   /// Will assert if new data point is more than max specified.
   double fixedMaxY;
 
+  /// Primary color.
   Color color;
 
+  /// Used for DataCount, if count > 1 use this color.
+  Color colorAggregate;
```
jacob314, line 51: Nit: maybe rewrite as 
/// Color to use if count > 1.
///
/// See [DataCount.count].

packages/devtools_app/lib/src/memory/memory_events_pane.dart
```diff
@@ -127,6 +137,9 @@ class MemoryEventsPaneState extends State<MemoryEventsPane>
   /// Controller for managing memory collection.
   MemoryController _memoryController;
 
+  /// Flutter events and user custom events.
+  static const extensionEvent = 3.7;
```
jacob314, line 141: comment on what  3.7 and 2.4 mean here and why they were chosen.

packages/devtools_app/lib/src/memory/memory_protocol.dart
```diff
@@ -248,6 +249,22 @@ class MemoryTracker {
     }
   }
 
+  /// Grab current EventSample, if any extensionEvents add those to this
+  /// EventSample and clear the outstanding extension events.
+  EventSample pullCloneClear(MemoryTimeline memoryTimeline, int time) {
```
jacob314, line 254: Can you give this a clearer name that indicates the task done not how it is implemented?

packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -167,6 +175,76 @@ class MemoryBodyState extends State<MemoryBody>
...
+        final tapData = vmChartController.tapNotifier.value;
+        final index = tapData.index;
+        final timestamp = tapData.timestamp;
+
+        final copied = TapNotifier.copy(vmChartController.tapNotifier.value);
+        eventChartController.setTapNotifier(copied);
+        androidChartController.setTapNotifier(copied);
+
+        final allValues = ChartsValues(this, index, timestamp);
+        if (isDebugging) {
```
jacob314, line 216: why not use debugLogger instead of isDebugging and print?

packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -181,53 +259,66 @@ class MemoryBodyState extends State<MemoryBody>
...
-        ),
-        const SizedBox(width: defaultSpacing),
-        Expanded(
-          child: TabBarView(
-            physics: defaultTabBarViewPhysics,
-            controller: tabController,
+    return RawKeyboardListener(
+      focusNode: FocusNode(),
+      onKey: (RawKeyEvent event) {
+        if (event.isKeyPressed(LogicalKeyboardKey.escape)) {
```
jacob314, line 265: Can we integrate into the general Flutter focus system instead of listening for escape ourselves?

packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -651,3 +1099,140 @@ class MemoryBodyState extends State<MemoryBody>
...
+      if (eventInfo.allocationAccumulator.isStart) {
+        results[monitorStartDisplayName] = true;
+      }
+      if (eventInfo.allocationAccumulator.isReset) {
+        results[monitorResetDisplayName] = true;
+      }
+    }
+
+    if (eventInfo.hasExtensionEvents) {
+      final List<Map<String, Object>> events = [];
```
jacob314, line 1194: nit:
`final events = <Map<String, Object>>[];`

packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -651,3 +1099,140 @@ class MemoryBodyState extends State<MemoryBody>
...
+            {
+              'name': event.eventKind,
+              'custom': {
+                'name': event.customEventName,
+                'data': event.data,
+              },
+            },
+          );
+        } else {
+          events.add(
```
jacob314, line 1207: nit: this all fits on one line. feel free to remove spurious trailing commas.

packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -651,3 +1099,140 @@ class MemoryBodyState extends State<MemoryBody>
...
+  void _fetchData(
+    ChartController chartController,
+    Map<String, Object> results,
+  ) {
+    for (var trace in chartController.traces) {
+      final theData = trace.data[index];
+      final yValue = theData.y;
+
+      // Convert enum'd string e.g., 'TraceName.capacity' to Map key'capacity', etc.
+      results[trace.name.split('.').last] = yValue;
```
jacob314, line 1229: So that we avoid collisions with surprising rendering:
```dart
final key = trace.name.split('.').last;
assert(!results.containsKey(key));
results[key] = yValue;
```

packages/devtools_app/lib/src/memory/memory_timeline.dart
```diff
@@ -85,12 +85,17 @@ class MemoryTimeline {
...
+
+  ValueListenable<Event> get eventNotifier => _eventFiredNotifier;
+
   /// Notifies that a new Heap sample has been added to the timeline.
   final _sampleAddedNotifier = ValueNotifier<HeapSample>(null);
 
   ValueListenable<HeapSample> get sampleAddedNotifier => _sampleAddedNotifier;
 
-  /// List of events awaiting to be posted to HeapSample.
+  /// List of events (DevTools) awaiting to be posted to HeapSample.
```
jacob314, line 98: remove `(DevTools)`?

packages/devtools_app/lib/src/memory/memory_timeline.dart
```diff
@@ -121,25 +126,35 @@ class MemoryTimeline {
...
+      events: extensionEvents,
     ));
   }
 
   void addGCEvent() {
     final timestamp = DateTime.now().millisecondsSinceEpoch;
-    postEventSample(EventSample.gcEvent(timestamp));
+    postEventSample(EventSample.gcEvent(
+      timestamp,
+      events: extensionEvents,
```
jacob314, line 137: why are we batching up extension events and only posting them when there is another event? it would seem we would want to add each as it occurs with the correct timestamp in stead of pretending they occur at the nearest gc or other event?

packages/devtools_app/lib/src/memory/memory_vm_chart.dart
```diff
@@ -164,11 +164,11 @@ class MemoryVMChartState extends State<MemoryVMChart> with AutoDisposeMixin {
 
   // TODO(terry): Move colors to theme?
   static final capacityColor = Colors.grey[400];
-  static final externalColor = Colors.blue[400];
-  static const rasterLayerColor = Color(0xff99cc00); // HoloGreenLight
-  static const rasterPictureColor = Color(0xffff4444); // HoloRedLight
-  static const rssColor = Color(0xffffbb33); // HoloOrangeLight
-  static const usedColor = Color(0xff33b5e5); // HoloBlue
+  static const usedColor = Color(0xff33b5e5);
```
jacob314, line 167: Would be good to get some uxr feedback on these new colors.

packages/devtools_app/lib/src/theme.dart
```diff
@@ -201,6 +201,45 @@ extension DevToolsColorScheme on ColorScheme {
...
+  Color get hoverBackgroundColor => const Color.fromARGB(150, 255, 255, 141);
+
+  // Bar color for current selection (hover).
+  Color get hoverSelectionBarColor =>
+      isLight ? Colors.lime[600] : Colors.yellowAccent;
+
+  // Title of the hover card.
+  TextStyle get hoverTitleTextStyle => const TextStyle(
+        color: Colors.black,
+        fontWeight: FontWeight.w600,
```
jacob314, line 215: I would use bold not semibold to reduce the # of different colors we use.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -404,6 +404,10 @@ class JsonUtils {
   static bool hasJsonData(String data) {
     return data != null && data.isNotEmpty && data != 'null';
   }
+
+  /// Pretty print a JSON payload.
+  static String prettyPrint(Map<String, Object> json) =>
```
jacob314, line 409: this might be a nice extension method on `Map<String, Object>`

packages/devtools_app/lib/src/utils.dart
```diff
@@ -979,13 +983,23 @@ class DevToolsFile<T> {
   final T data;
 }
 
-/// Logging to debug console only in debug runs.
-void debugLogger(String message) {
+bool get isDebugging {
```
jacob314, line 986: remove this and use the 
assert(() {

  return true;
}());

pattern used by Flutter.

packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -167,6 +175,76 @@ class MemoryBodyState extends State<MemoryBody>
       });
     });
 
+    addAutoDisposeListener(eventChartController.tapNotifier, () {
+      if (!eventChartController.tapNotifier.value.isEmpty &&
```
jacob314, line 179: is there a reason we add these isEmpty checks rather than checking whether the value is null?

jacob314 (commented):
Done with a first round of review comments.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart.dart
terrylucas: Done.

packages/devtools_app/lib/src/charts/chart_controller.dart
terrylucas: Yep, now that I have both.

packages/devtools_app/lib/src/charts/chart_trace.dart
terrylucas: Done.
```
  /// Color to use if count > 1.
  ///
  /// See [DataAggregate.count].

```

packages/devtools_app/lib/src/memory/memory_events_pane.dart
terrylucas:
```
  /// Note: The event pane is a fixed size chart (y-axis does not scale). The
  ///       Y-axis fixed range is (visibleVmEvent to extensionEvent) e.g.,
  /// 
  ///                         ____________________
  ///         extensionEvent -|            * (3.7)
  ///                         |         * (2.4)
  ///                         |      * (1.4)
  ///         visibleVmEvent -|__ *  (0.4)
  /// 
  ///       The *s in the above chart are plotted at each y position (3.7, 2.4, 1.4, 0.4).
  ///       Their y-position is such that the symbols won't overlap.
  /// TODO(terry): Consider a better solution percent location in the Y-axis
  ///              might be a better interface.
  
  /// Flutter events and user custom events.
  static const extensionEvent = 3.7;

  /// Event to display in the event pane (User initiated GC, snapshot,
  /// automatic snapshot, etc.)
  static const visibleEvent = 2.4;

  /// Monitor events Y axis.
  static const visibleMonitorEvent = 1.4;

  /// VM's GCs are displayed in a smaller glyph and closer to the heap graph.
  static const visibleVmEvent = 0.4;
```

packages/devtools_app/lib/src/memory/memory_protocol.dart
terrylucas:
```
  /// Many extension events could arrive between memory collection ticks, those
  /// events need to be associated with a particular memory tick (timestamp). This
  /// routine collects those new events received that are closest to a tick
  /// (time parameter)).
  /// @returns copy of events to associate with an existing HeapSample tick
  /// (contained in the EventSample). See [processEventSample] it computes the
  /// events to aggregate to an existing HeapSample or delay associating those
  /// events until the next HeapSample (tick) received see [_recalculate].
  EventSample pullClone(MemoryTimeline memoryTimeline, int time) {

```

packages/devtools_app/lib/src/memory/memory_screen.dart
terrylucas: Wanted an isDebugging (that only shows up if turned on).  I've added a field just for that with a test that ensures isDebugging is never checked in as true.

packages/devtools_app/lib/src/memory/memory_screen.dart
terrylucas: Maybe but after playing with Focus, FocusNode, and FocusScope doesn't seem as straight forward because I have no TextField.  Maybe I could create a hidden text field but that seems ugly too. I'll ad a TODO.

packages/devtools_app/lib/src/memory/memory_screen.dart
terrylucas: done.

packages/devtools_app/lib/src/memory/memory_screen.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_screen.dart
terrylucas: Good idea.

packages/devtools_app/lib/src/memory/memory_timeline.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_timeline.dart
terrylucas: Because the x-axis is driven by the the ticks returned from the VM.  if we start synthesizing other timestamps then the x-axis will appear stretched as extension events arrive. Possibly we could plot x-axis based on a fix interval and layer the events but then I suspect we'd have the same issue some events would have to be aggregated.

packages/devtools_app/lib/src/memory/memory_vm_chart.dart
terrylucas: I presume you talking about the raster cache colors only?
I'll create an issue for this https://github.com/flutter/devtools/issues/2616

packages/devtools_app/lib/src/theme.dart
terrylucas: Done.

packages/devtools_app/lib/src/utils.dart
terrylucas: Done.

packages/devtools_app/lib/src/utils.dart
terrylucas: Good idea, done.

packages/devtools_app/lib/src/memory/memory_screen.dart
terrylucas: isEmpty can be replaced with null and the correct null checks and ?. 
Done.

terrylucas:
PTAL

jacob314 (approved):
lgtm

kenzieschmoll (approved):
LGTM on button changes

PR #2617, opened 2021-01-20
Support latest null safe vm_service

grouma:


kenzieschmoll:
whoops - sorry Gary I didn't push up my latest version bump to 0.9.6+1 yesterday. Could you add that entry to the changelog below yours?
```
## 0.9.6+1  
* Fallback to port 0 if we cannot connect the DevTools server to ports 9100-9109 #2600 
```

kenzieschmoll (approved):

PR #2622, opened 2021-01-22
dumpsys meminfo is a heavy action(trigger native GC), only fetch when android memory chart visible

uppet:
This makes the performance more accurate when the android part of the memory data is not needed.

### DIFF
--- a/packages/devtools_app/lib/src/memory/memory_protocol.dart
+++ b/packages/devtools_app/lib/src/memory/memory_protocol.dart
@@ -106,7 +106,9 @@ class MemoryTracker {
 
     // Polls for current Android meminfo using:
     //    > adb shell dumpsys meminfo -d <package_name>
-    if (hasConnection && serviceManager.vm.operatingSystem == 'android') {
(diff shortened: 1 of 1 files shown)


devoncarew:
Thanks for the PR! Adding @terrylucas and @kenzieschmoll for review.

terrylucas:
This is problematic.  Opening up the android memory chart will have empty data from (start) till Android memory is opened.

A better solution is to have an option to collection Android statistics instead of whether or not the chart is visible.  In fact, soon the extra Android Memory chart button will be removed and the android data will be overlayed with the Dart collected data and event pane.

terrylucas:
I'm adding an option to enable Android Memory Collection to preferences will add your change but checking the state of the preferences.

terrylucas:
This [PR](https://github.com/flutter/devtools/pull/2624) should solve your problem.

uppet:
> This [PR](https://github.com/flutter/devtools/pull/2624) should solve your problem.

Okay, this is a better solution.

terrylucas:
Here's the new [PR](https://github.com/flutter/devtools/pull/2637)

PR #2624, opened 2021-01-22
Added collecting of Android memory stats as a setting.

terrylucas:
image

By default collecting Android memory stats is off.

### DIFF
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -487,6 +487,11 @@ class SettingsDialog extends StatelessWidget {
             listenable: preferences.vmDeveloperModeEnabled,
             toggle: preferences.toggleVmDeveloperMode,
           ),
+          _buildOption(
+            label: const Text('Enable Android Memory Collection'),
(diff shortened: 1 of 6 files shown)


kenzieschmoll:
This seems like a setting that should be part of the memory page directly - not a top level setting for the DevTools app. Look at the performance page for an example of page-level settings.

terrylucas:
Didn't realize performance had settings.  As we add more settings probably want to have a tabbed settings at the top-level for each page otherwise its clunky.  Was hoping to help the person that wanted to disable android memory stats collection easily using an already existing settings.  Guess I'll need to add a real settings page.

kenzieschmoll:
I don't think it would be much more work to put this in the memory page - you can re-use the existing settings UI from the Performance page and just add a value listenable to the memory controller. Look at `badgeTabForJankyFrames` in the performance controller and follow how that is used with a `NotifierCheckbox` in the performance screen. We have an existing `DevToolsDialog` class that is used there as well. 

This change wouldn't be released until next friday anyway, so no rush to land quickly 👍

PR #2629, opened 2021-01-26
Listen for logs with event history in logging page and error badge manager

kenzieschmoll:
This uses the vm service extensions from dds: https://github.com/dart-lang/sdk/blob/master/pkg/dds/lib/vm_service_extensions.dart to pull event history from a stream when subscribing a listener. Before these service extensions were added, event history would only be sent to the _first_ subscriber of an event stream. So when we subscribe from different parts of devtools, not all subscribers were getting event history.

Added TODOs throughout other parts of the code where we listen to event streams - tagging owners for input on whether we should listen with history in these parts as well.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/debugger_controller.dart
+++ b/packages/devtools_app/lib/src/debugger/debugger_controller.dart
@@ -16,6 +16,7 @@ import '../auto_dispose.dart';
 import '../core/message_bus.dart';
 import '../globals.dart';
 import '../utils.dart';
+import '../vm_service_wrapper.dart';
 import 'debugger_model.dart';
 
 // TODO(devoncarew): Add some delayed resume value notifiers (to be used to
@@ -38,7 +39,9 @@ class DebuggerController extends DisposableController
         .listen(switchToIsolate));
     autoDispose(_service.onDebugEvent.listen(_handleDebugEvent));
     autoDispose(_service.onIsolateEvent.listen(_handleIsolateEvent));
+    // TODO(kenz): do we want to listen with event history here?
     autoDispose(_service.onStdoutEvent.listen(_handleStdoutEvent));
+    // TODO(kenz): do we want to listen with event history here?
     autoDispose(_service.onStderrEvent.listen(_handleStderrEvent));
 
     _scriptHistoryListener = () {
@@ -47,7 +50,7 @@ class DebuggerController extends DisposableController
     scriptsHistory.addListener(_scriptHistoryListener);
   }
 
-  VmService get _service => serviceManager.service;
+  VmServiceWrapper get _service => serviceManager.service;
 
   final ScriptCache _scriptCache = ScriptCache();
 
--- a/packages/devtools_app/lib/src/inspector/inspector_service.dart
+++ b/packages/devtools_app/lib/src/inspector/inspector_service.dart
@@ -68,6 +68,7 @@ class InspectorService extends DisposableController
     this.inspectorLibrary,
     this.supportedServiceMethods,
   ) : clients = {} {
+    // TODO(kenz): do we want to listen with event history here?
     autoDispose(
         vmService.onExtensionEvent.listen(onExtensionVmServiceRecieved));
     autoDispose(vmService.onDebugEvent.listen(onDebugVmServiceReceived));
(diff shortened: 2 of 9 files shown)


packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -38,7 +39,9 @@ class DebuggerController extends DisposableController
         .listen(switchToIsolate));
     autoDispose(_service.onDebugEvent.listen(_handleDebugEvent));
     autoDispose(_service.onIsolateEvent.listen(_handleIsolateEvent));
+    // TODO(kenz): do we want to listen with event history here?
     autoDispose(_service.onStdoutEvent.listen(_handleStdoutEvent));
+    // TODO(kenz): do we want to listen with event history here?
     autoDispose(_service.onStderrEvent.listen(_handleStderrEvent));
```
kenzieschmoll, line 45: @devoncarew @grouma - any input on whether we should listen to event history here? I would think listening with history would mean that the console would populate with stdout and stderr event that occurred before a user opens the DevTools debugger.

packages/devtools_app/lib/src/inspector/inspector_service.dart
```diff
@@ -68,6 +68,7 @@ class InspectorService extends DisposableController
     this.inspectorLibrary,
     this.supportedServiceMethods,
   ) : clients = {} {
+    // TODO(kenz): do we want to listen with event history here?
```
kenzieschmoll, line 71: @jacob314  - any input on whether we should listen to event history here?

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -567,6 +567,7 @@ class MemoryController extends DisposableController
     _memoryTracker.start();
 
     // Log Flutter extension events.
+    // TODO(kenz): do we want to listen with event history here?
```
kenzieschmoll, line 570: @terrylucas - any input on whether we should listen to event history here?

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -1092,6 +1092,7 @@ class ServiceExtensionManager extends Disposer {
     cancel();
     _connectedApp = connectedApp;
     _service = service;
+    // TODO(kenz): do we want to listen with event history here?
```
kenzieschmoll, line 1095: @jacob314 - any input on whether we should listen to event history here?

packages/devtools_app/lib/src/debugger/debugger_controller.dart
grouma: From the web perspective we don't capture log messages until debugging is initiated so for consistency we wouldn't want event history. However I can see an argument to also display them if you have the capability.

jacob314: We should try to change that for the web. I'm pretty sure the chrome devtools protocol gives you an api to fetch all historic messages.  The old messages may only show up as plain text (helps them avoid memory leaks).

packages/devtools_app/lib/src/inspector/inspector_service.dart
jacob314: no need to listen here as the inspector uses a separate api to get the current selection.

packages/devtools_app/lib/src/memory/memory_controller.dart
jacob314: unless we were able to get historic events about total memory usage there is probably no need to.

packages/devtools_app/lib/src/service_manager.dart
jacob314: We don't. If we did it could cause bugs particularly if some of these events are from before the last hot reload.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -567,6 +567,8 @@ class MemoryController extends DisposableController
     _memoryTracker.start();
 
     // Log Flutter extension events.
+    // Note: We do not need to listen to event history here because we do not
+    // have matching historical data about total memory usage.
```
terrylucas, line 571: That's correct. Today we could start collecting memory from the beginning of time - if the app to profile was started with `--start-paused`. When DevTools is up:
1. flip to the memory page
2. flip to debugger
3. press GO

I'm wondering should we have a simpler way like a setting to "start collect memory statistics on DevTools attach"?  Doesn't effect anything today just a question if anyone has a suggestion.

Issue #2503, opened 2020-11-16
P1: Surface layout errors in Inspector


kenzieschmoll:
When inspecting a screen with errors (e.g. renderflex overflow), surface these errors in the inspector. We could highlight the widget names in red and provide a way to navigate to the problem widget.

kenzieschmoll:
@jacob314 do we have information about which widgets triggered errors? do we get this information on a vm-service stream that?

DanTup:
@kenzieschmoll @jacob314 if the same error is thrown multiple times, the URL with the `inspectorRef` is the same:

> &inspectorRef=inspector-0

However that ID doesn't seem to occur anywhere else in the structured error (all `objectId`s in the errors properties are different to this). It seems like we need to be able to de-dupe these to make the error badge count accurate (eg. it goes up by two when I run my sample overflow app, even though the inspect URL is the same) and only list any errored widgets once.

DanTup:
Is something like this what you had in mind?

![Screenshot 2021-01-26 at 15 06 04](https://user-images.githubusercontent.com/1078012/105862746-0c661480-5fe8-11eb-90bf-0798d88803ff.png)

I'm not sure if we have enough info to highlight the highlight the item in the tree in some way, though I figured showing the error list would be useful in case the errored item is collapsed/offscreen.

I have a few issues though:

- It's not clear when an error becomes stale - they build up in the list and clicking them does nothing (eg. after you'd reloaded). We could clear them on reload/restart, but that's probably just masking the issue (since they could presumably be invalidated by a user navigating around the app)
- I can't figure out how to give the errors an obvious mouse pointer to make it clear they're clickable (the tree has pointer cursors, but I couldn't find where that's done)

kenzieschmoll:
In the future we could add the inspector ref Id as a field to the structured error json. But for the first implementation we will need to get that info from the parsed uri. 

I agree we should de-dupe the errors. We could do this in the error badge manager by storing a set of the active error causing inspector refs in the tree. When the inspector tree is refreshed, we could clear this list and the badge count, or maybe a better solution would be to clear the errors that are no longer in the tree and keep the ones that are.

In the error badge manager code, we badge the inspector for more errors than just the overflow errors. Perhaps to start we should just badge errors for the overflow errors (the ones we have added DevToolsDeepLinkProperty for) and keep things simple for the MVP.

DanTup:
> for the first implementation we will need to get that info from the parsed uri

[...]

PR #2630, opened 2021-01-26
Render inspectable widget errors below the inspector tree

DanTup:
WIP towards #2503.

- [x] Figure out why there's sometimes a red-error-screen of a failed assert when clicking the "Refresh Tree" button
- [x] Ensure error navigator updates when manually selecting nodes
- [ ] ~~Support errors for nodes that aren't in the summary tree~~ `inspectorRef` in errors appears to always be in the summary tree
- [x] Check how this works in embedded views - https://youtu.be/mKc7kX2B7W0
- [x] Tests


packages/devtools_app/lib/src/error_badge_manager.dart
```diff
@@ -43,16 +44,52 @@ class ErrorBadgeManager extends DisposableController
...
+    if (devToolsUrlNode == null) {
+      return null;
+    }
+
+    var inspectWidgetUrl = Uri.tryParse(devToolsUrlNode['value'] as String);
+    if (inspectWidgetUrl == null) {
+      return null;
+    }
+
+    // Handle when querystring is in the fragement.
```
kenzieschmoll, line 81: See `devToolsQueryParams` in utils.dart - you may be able to use this here.

packages/devtools_app/lib/src/error_badge_manager.dart
```diff
@@ -23,6 +23,7 @@ class ErrorBadgeManager extends DisposableController
     NetworkScreen.id: ValueNotifier<int>(0),
     LoggingScreen.id: ValueNotifier<int>(0),
   };
+  final _activeErroredWidgets = ValueNotifier<List<InspectableWidgetError>>([]);
```
kenzieschmoll, line 26: rename to `_activeErrors` and make it a `Map<String, ValueNotifier<List<DevToolsError>>>` where `DevToolsError` is some base class that takes a message and a unique id. `InspectableWidgetError` can be an extension of this class. 

This will allow the functionality you are building to be extended to other screens for which we'll want to track unique errors.

packages/devtools_app/lib/src/error_badge_manager.dart
```diff
@@ -65,15 +102,43 @@ class ErrorBadgeManager extends DisposableController
     notifier.value = currentCount + 1;
   }
 
+  void appendInspectorErroredWidget(InspectableWidgetError error) {
+    var currentErrors = _activeErroredWidgets.value;
+    if (!currentErrors.any((e) => e.inspectorRef == error.inspectorRef)) {
+      currentErrors = [...currentErrors, error];
```
kenzieschmoll, line 108: would it be better to use a set instead of a list?

packages/devtools_app/lib/src/error_badge_manager.dart
```diff
@@ -65,15 +102,43 @@ class ErrorBadgeManager extends DisposableController
...
+      _activeErroredWidgets.value = currentErrors;
+      _errorCountNotifier(InspectorScreen.id).value = currentErrors.length;
+    }
+  }
+
   ValueListenable<int> errorCountNotifier(String screenId) {
     return _errorCountNotifier(screenId) ?? const FixedValueListenable<int>(0);
   }
 
+  ValueListenable<List<InspectableWidgetError>> erroredWidgetNotifier() {
```
kenzieschmoll, line 118: if you can implement my suggestion on `_activeErroredWidgets` above, this should take a string `screenId` and return the valueListenable for the screen

packages/devtools_app/lib/src/inspector/inspector_controller.dart
```diff
@@ -350,16 +350,19 @@ class InspectorController extends DisposableController
...
 
   @override
-  Future<void> onForceRefresh() {
+  Future<void> onForceRefresh() async {
     assert(!_disposed);
     if (!visibleToUser || _disposed) {
       return Future.value();
     }
-    recomputeTreeRoot(null, null, false);
+    // TODO(dantup): Is it safe to add this await here?
```
kenzieschmoll, line 358: Not sure - that would be a question for @jacob314

packages/devtools_app/lib/src/inspector/inspector_controller.dart
```diff
@@ -613,14 +623,11 @@ class InspectorController extends DisposableController
...
 
-    if (firstFrame) {
-      final queryParams = loadQueryParams();
-      if (queryParams.containsKey(inspectorRefQueryParam)) {
-        await group.setSelectionInspector(
-          InspectorInstanceRef(queryParams[inspectorRefQueryParam]),
-          false,
-        );
-      }
+    if (inspectorRef != null) {
```
kenzieschmoll, line 626: What is the motivation behind moving the query param check out of `updateSelectionFromService`? `updateSelectionFromService` is also called from `onInspectorSelectionChanged` - is it safe that we are no longer checking the query params from that call site?

packages/devtools_app/lib/src/inspector/inspector_controller.dart
```diff
@@ -915,4 +922,17 @@ class InspectorController extends DisposableController
       onLayoutExplorerSupported,
     );
   }
+
+  Stream<String> _collectValidInspectorRefs(
```
kenzieschmoll, line 926: is there anyway we could create a set of all these ids at the point that we compute the tree? That way we wouldn't have to traverse the tree twice and we would have have quick lookup time for seeing if the set contains one of the error-causing inspectorRefs

@jacob314 for some input here.

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -93,6 +94,46 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+    List<InspectableWidgetError> errors,
+  ) {
+    return ListView(
+      padding: const EdgeInsets.all(8),
+      children: <Widget>[
+        for (final error in errors)
+          Padding(
+            padding: const EdgeInsets.only(bottom: 8),
+            // TODO(dantup): How to make the cursor a hand pointer?
+            child: GestureDetector(
```
kenzieschmoll, line 108: try using an InkWell widget instead of a gesture detector

packages/devtools_app/lib/src/error_badge_manager.dart
```diff
@@ -43,16 +44,52 @@ class ErrorBadgeManager extends DisposableController
...
-      if (objectId?.contains('inspector-') ?? false) {
-        incrementBadgeCount(InspectorScreen.id);
+      final inspectableError = _extractInspectableError(e);
+      if (inspectableError != null) {
+        appendInspectorErroredWidget(inspectableError);
       }
     }
   }
 
+  InspectableWidgetError _extractInspectableError(Event error) {
```
jacob314, line 54: Use the existing RemoteDiagnosticsNode class to parse these objects instead of writing a custom json parser here.

packages/devtools_app/lib/src/error_badge_manager.dart
jacob314, line 108: why aren't we just adding error to currentErrors?

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -104,6 +145,26 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
         isSummaryTree: true,
       ),
     );
+
+    final treeColumn = ValueListenableBuilder<List<InspectableWidgetError>>(
```
kenzieschmoll, line 149: how does this look for the narrow width version of the inspector? We may need some UX input on how to best display this information - especially since this change would likely be part of the demo I am recording soon. @InMatrix @cobblest

packages/devtools_app/lib/src/error_badge_manager.dart
kenzieschmoll: I think it's because a ValueNotifier with a list won't trigger set state if the lists content changes.

packages/devtools_app/lib/src/inspector/inspector_controller.dart
jacob314: this is way too slow because you are awaiting fetching children one at a time possibly for the entire tree.
The computed tree already includes the ids and has a set similar to this.
The challenge is that will only be ids from the summary tree so make sure you are only needing to compare against a node id from the summary tree.
For the detail tree we don't load the whole tree as it is huge.

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -93,6 +94,46 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+              onTap: () => inspectorController.updateSelectionFromService(
+                  firstFrame: false, inspectorRef: error.inspectorRef),
+              child: Row(
+                children: [
+                  Padding(
+                    padding: const EdgeInsets.only(right: 8),
+                    child: Container(
+                      padding: const EdgeInsets.symmetric(horizontal: 3.0),
+                      decoration: BoxDecoration(
+                        color: const Color.fromARGB(0xff, 0xF4, 0x43, 0x36),
```
jacob314, line 118: nit: move all colors to constants next to color colors from our "theme"

packages/devtools_app/lib/src/inspector/inspector_screen.dart
InMatrix: I just took a look at the screenshot in #2503. I'm curious about the intended workflow for the user. How can someone go from seeing the badge of error count on the inspector tab to locating the error-causing widget?

cobblest: I had the same question. From the screenshot it might be difficult for users to find the error-causing widget. It would be great if we could highlight the problematic widget in the widget tree like the prototype here: https://docs.google.com/presentation/d/1Ocgvw9RP_mL657ACSmGVs0RRiPGuacGbAK97KAtE0eM/edit#slide=id.g7edbfe90b0_2_47

packages/devtools_app/lib/src/error_badge_manager.dart
DanTup: Thanks! Done in 6f779c11

packages/devtools_app/lib/src/error_badge_manager.dart
DanTup: Done in 6f779c11, although because we can't cast `ValueNotifier <List<DevToolsError>>` to `ValueNotifier <List<InspectableWidgetError>>` we have to call `.cast()` on the list when rendering the errors. I'm not sure if there's a way around that.

packages/devtools_app/lib/src/error_badge_manager.dart
DanTup: To avoid the `any` enumerating through it, or something else? I think using a set would also require enumerating (since we're checking for the same ID rather than full same object), and both require creating a new object because the notifier only fires if `.value` changes.

Changing it to a `Map<String, DevToolsError>` where the ID is also the key makes this easier (using `containsKey`) so I've done that (in 6f779c11ea3b3f9617548fed1b7e3b4e8f51982b), but lmk if this doesn't seem like a good change.

packages/devtools_app/lib/src/error_badge_manager.dart
DanTup: Done! (6f779c11ea3b3f9617548fed1b7e3b4e8f51982b)

packages/devtools_app/lib/src/inspector/inspector_controller.dart
DanTup: It was just because I wanted to provide a different value for clicking the errors and it felt a little awkward adding a parameter to allow you to supply it, but then reading it from the querystring based on the value of a different parameter.

> is it safe that we are no longer checking the query params from that call site?

The query string code was inside an `if (firstFrame)` so it wouldn't have executed via that code path.

(happy to change this if you prefer it the other way - presumably we'd always prefer the parameter if one was provided over the querystring?)

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: Aha, thanks! (Done in 6f779c1)

packages/devtools_app/lib/src/error_badge_manager.dart
DanTup: Yep! I did try to call `notifyListeners()` directly, but here's a warning saying it's not allowed outside of the Flutter package. This code has changed slightly (it's now a map not a list), but still has to do this - if you know of a better way, let me know!

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: @kenzieschmoll 

> how does this look for the narrow width version of the inspector?

Looks like this:

Screenshot 2021-01-27 at 17 37 28

I'm a little unsure if this is how we wanted to render errors (I was after some feedback) - I assumed we need a list somewhere, because the errored widgets may be offscreen (or perhaps collapsed?). I think the widgets should still be badged something like the mockup above.

@InMatrix @cobblest

> I just took a look at the screenshot in #2503. I'm curious about the intended workflow for the user. How can someone go from seeing the badge of error count on the inspector tab to locating the error-causing widget?

The number in the badge on the Inspector tab corresponds to the number of errors rendered in the bottom left of that screenshot. You can click on them to jump to the relevant widget in the tree. The widgets should also be marked visibily in the tree, though that's not done here yet.

packages/devtools_app/lib/src/inspector/inspector_controller.dart
DanTup: Ah, didn't realise this code was used in both trees. I can guard it with a check of `isSummaryTree` (my understanding is that the widgets are all in the summary tree and this should be fine?).

> The computed tree already includes the ids and has a set similar to this.

Where is the computed tree? It looked like these calls to `children` were all bailing out early because they already had the data, but I don't know if that's guaranteed or not (or if that's dependant on the `await` I added to the `recomputeTreeRoot` call?).

packages/devtools_app/lib/src/error_badge_manager.dart
DanTup: Done in a0f58d79, although I'm passing `null` as the `inspectorService` (it wasn't clear how I can get one here, and it didn't seem to be required) and I'm not sure if that's reasonable?

packages/devtools_app/lib/src/inspector/inspector_screen.dart
InMatrix: Thanks for explaining the interaction. It sounds good to me. Looking forward to trying it out live.

raison00: i've created some mocks to put the prototype into better context @cobblest posted earlier: 

Note:  Using the error color in the screenshot. **Final error color should match established color scheme.** This is FPO.

- Design 1A has a marker on the Row and the Overflow error.  
_Intention:_  if there's an issue at the Row level, the Row is badged when collapsed.  When open, the badged marker identifies the overflow error and the Row level badge is not visible. This marker appears when the Row is collapsed and the error has not been resolved.
- Design 1 + 6 have the difference of the badge marker by the overflow error.  
_Intention:_  Using the badge markers to navigate to the overflow error quickly without needing to scroll within the tree.  This helps visualize the user experience without a marker with an error.  Marker will help surface the error if the overflow widget is nested within a deep tree.


devtools-le-error-0a1

devtools-le-error-01
devtools-le-error-02
devtools-le-error-03
devtools-le-error-04
devtools-le-error-05
devtools-le-error-06

kenzieschmoll:
I like 1 or 6 of all the mocks above. In combination with that, what if we had a little overlay at the top right corner of the summary tree:
Screen Shot 2021-01-27 at 2 58 02 PM
This area will almost always be open, due to the nature of the tree structure starting on the left side and the first node always being named root. When errors are present, this overlay would appear and look something like
`Errors: 1/2    <     >`
Where the carets are actually buttons that allow a user to navigate to the previous / next error in the tree.

raison00:
Here's some updated mocks of the error navigation with carets that function as actual buttons.  1 and 6 are more developed, the others were only updated with slight variations of the error count/carat navigation.

The Round Circle:  Color alone is not  the error indicator; the circle indication helps low vision users and more.

There are a few options to navigate and discover errors:

- the error count and carat navigation at the top right of the tree
- the error console will have a badge of the issue and its number
- the error in the tree will have a badge of the issue and its number

The orangish color has been updated to a red in these mocks, but the exact color expectation should use the theme color for dark/light.   

aerror-nav-devtools-le-error-tree-01


2error-nav-devtools-le-error-tree-06

aerror-nav-devtools-le-error-tree-02
aerror-nav-devtools-le-error-tree-03
aerror-nav-devtools-le-error-tree-04
aerror-nav-devtools-le-error-tree-05

DanTup:
I've made some changes based on the above, this is how it looks:

https://youtu.be/EYv3y2lWF08

A red background on all nodes with errors seemed a bit confusing (they looked like selected nodes) so I just left a border (and made the background red when selected) - how does that seem?

I haven't done anything on the details tree yet, nor made it select/unselect the "current" error if you manually click around the tree.

kenzieschmoll:
I like the red outlines and the look within the widget tree. I also thing the "error navigator" looks good. I'm not sold on the extra splitter under the summary tree, just because when the inspector is has a narrow width, this feels a little off. If we have the "error navigator" at the top of the summary tree, I don't think we'd even need the extra section with the full error. One thing we could do is to give the error dots / errored widgets a tooltip that describes the error. This would still allow the user to see the exact error text, and then we wouldn't have multiple ways to accomplish the same thing (navigating through errors in the tree).

packages/devtools_app/lib/src/inspector/inspector_controller.dart
kenzieschmoll: How will we handle errors for widgets that are not in the summary tree? I think this could be a pretty common occurrence, right @jacob314?

DanTup: Ah, I wasn't sure if that was possible. Right now IDs that aren't in the summary tree will just be pruned out. We'd need to be able to tell which ones are still valid that aren't in the summary tree to support this.

Is there an error I can generate that has the deep link that wouldn't show up in the summary tree?

DanTup:
>  If we have the "error navigator" at the top of the summary tree, I don't think we'd even need the extra section with the full error
>
> One thing we could do is to give the error dots / errored widgets a tooltip that describes the error

Sounds reasonable to me. I was concerned if we only had the dots they might not be discoverable if offscreen/collapsed, but I think you're right - now there's another way to navigate through the errors (and hopefully users won't have large numbers) it's probably unnecessary.

InMatrix:
> I've made some changes based on the above, this is how it looks:
> https://youtu.be/EYv3y2lWF08

I love it. Thanks for the quick turnaround, @DanTup and @raison00! 

> One thing we could do is to give the error dots / errored widgets a tooltip that describes the error

+1. I think there is value to provide a way to access relevant error messages within the Inspector page. It would be awkward to have the user go back and forth between DevTools and the IDE to have the full context about an error.

packages/devtools_app/lib/src/error_badge_manager.dart
jacob314: I believe InspectorRef should work as a key type in a Map and that way it is clearer what the types are.

packages/devtools_app/lib/src/inspector/inspector_controller.dart
jacob314: it is fine to add an await here. It is not ok to incrementally fetch the tree one level at a time. Fetching the tree needs to be more of a batch operation to avoid excessive round trips over the VM Service.

packages/devtools_app/lib/src/inspector/inspector_controller.dart
jacob314: updateSelectionFromService is the breadcrumb for how you want to implement this efficiently. That code path has the same challenges of you need to update the location in the summary tree that is closest to the current selection and update the details selection. The code to solve it is ugly and convoluted so it is much easier to make sure you use the same code path than make another code path that does the same thing.
Basically what happens is given an inspector ref it computes the first parent of the inspector ref that is in the summary tree and selects that.
If the node is already in the detail tree it selects that node, otherwise it reloads the detail tree rooted appropriately to include the node you need from the details tree.

packages/devtools_app/lib/src/inspector/inspector_controller.dart
DanTup: This is now using `hasDiagnosticsValue` on the controller, which checks a local map that has these.

DanTup:
Here's how it looks with the error list removed, tooltips on the red dots and the navigator updating as you manually move around the nodes:

https://youtu.be/XE1064QSZh8

I removed the numbers on the dots now the error list is gone, since they looked like counts (like you had 2 errors on the second one) and seemed less useful without the list (though if they seemed useful they're easy to add back).

I still need to do some testing with errors that only appear in the details tree (I suspect right now they just won't show up and may need some work, but I haven't been able to produce a sample for that yet).

jacob314:
I think we need to make the red dots behave more like "unread" indicators. Once you have selected both errors the red dot on the tab should go away.

DanTup:
@jacob314 

> I think we need to make the red dots behave more like "unread" indicators. Once you have selected both errors the red dot on the tab should go away.

Should this apply only to the count (eg. it's possible to have no "unread" errors on the tab, but still see the red dots/backgrounds on the tree)?

Otherwise, I think we'd end up removing the background/dot immediately when you click on it, and if you're using the navigator it would look strange (eg. when you move to error 1/2 the count would immediately decrease to 1 and you'd no longer be on an active error).

kenzieschmoll:
> Should this apply only to the count (eg. it's possible to have no "unread" errors on the tab, but still see the red dots/backgrounds on the tree)?

Yes. If errors are present in the tree, the error navigator and red dots in the tree should remain. But the badge on the tab can go away.

DanTup:
I've pushed changes that mark errors as "read" when they're visited so that the badge count can be independent of the errors being shown.

@jacob314 are there existing tests that render the summary tree? I need to add tests for the error indicators in the tree, but couldn't find anything suitable to use as a template.

packages/devtools_app/lib/src/error_badge_manager.dart
DanTup: ^^ this is still outstanding. It seems to work, but I'm not sure it's valid to pass `null` here (and if not, what the correct way to get a reference to an `inspectorService` here would be.

packages/devtools_app/lib/src/error_badge_manager.dart
```diff
@@ -23,6 +28,12 @@ class ErrorBadgeManager extends DisposableController
     NetworkScreen.id: ValueNotifier<int>(0),
     LoggingScreen.id: ValueNotifier<int>(0),
   };
+  final _activeErrors = <String,
+      ValueNotifier<LinkedHashMap<InspectorInstanceRef, DevToolsError>>>{
```
kenzieschmoll, line 32: we have the problem here that InspectorInstanceRef is specific to the inspector, so this `_activeErrors` won't apply to other pages, like the logging view or performance view for example. It may take some extra work to generalize this, so I'm fine with you changing this to `_activeInspectorErrors` of type `ValueNotifier<LinkedHashMap<InspectorInstanceRef, DevToolsError>>` for now. Can you also add a todo explaining that we should generalize this to be compatible with errors for other screens?

packages/devtools_app/lib/src/error_badge_manager.dart
```diff
@@ -44,16 +55,42 @@ class ErrorBadgeManager extends DisposableController
...
+      orElse: () => null,
+    );
+    if (devToolsUrlNode == null) {
+      return null;
+    }
+
+    final queryParams =
+        devToolsQueryParams(devToolsUrlNode.getStringMember('value'));
+    final inspectorRef =
+        queryParams != null ? queryParams['inspectorRef'] : null;
```
kenzieschmoll, line 89: Will it be possible to get the inspector ref for errors that do not have a deep link? @jacob314 @DanTup 

As is this would only work for overflow errors (the only errors with a deep link).

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -160,6 +168,42 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
     );
   }
 
+  Widget _buildSummaryTreeColumn() {
+    final errors = _errors
+        ?.map((key, value) => MapEntry(key, value as InspectableWidgetError));
+    final errorList = errors?.values?.toList() ?? [];
+
+    return Container(
```
kenzieschmoll, line 176: use `OutlineDecoration` from common_widgets.dart

packages/devtools_app/lib/src/error_badge_manager.dart
DanTup: How about we change the key back from `InspectorInstanceRef` to just `String`? I think the rest of the code is fairly generic, it's just this `key` that is a wrapper over the string.

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: Done! (f97c0d2b)

packages/devtools_app/lib/src/error_badge_manager.dart
kenzieschmoll: sgtm

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: This code was removed, and now uses a shared existing color.

packages/devtools_app/lib/src/error_badge_manager.dart
DanTup: Done! (b2c3652e)

packages/devtools_app/lib/src/error_badge_manager.dart
DanTup: When I first looked, the `inspectorRef` in the deep link didn't seem to be repeated anywhere else in the structured error, so I wasn't sure. If there is a way to get it (or add it - I'm not very familiar with what these IDs are or how they'd made), then I think this should be able to tweak this to work for others too.

packages/devtools_app/lib/src/error_badge_manager.dart
kenzieschmoll: @jacob314? ☝️

packages/devtools_app/lib/src/error_badge_manager.dart
kenzieschmoll: IIRC from when I was adding the deep link to overflow errors, the widget needed to be "registered" in a sense on the inspector service, which will happen when the inspector is initialized with all its widgets unless manually registered from Flutter (like in the case of widgets related to overflow errors). @jacob314 can add more context here.

packages/devtools_app/lib/src/error_badge_manager.dart
```diff
@@ -66,15 +101,84 @@ class ErrorBadgeManager extends DisposableController
...
+        _activeErrors[screenId]?.value ?? const <String, DevToolsError>{};
+    _errorCountNotifier(screenId).value =
+        errors.values.where((err) => !err.read).length;
+  }
+
   ValueListenable<int> errorCountNotifier(String screenId) {
     return _errorCountNotifier(screenId) ?? const FixedValueListenable<int>(0);
   }
 
+  ValueListenable<LinkedHashMap<String, DevToolsError>> erroredWidgetNotifier(
```
kenzieschmoll, line 129: let's rename this to erroredElementNotifier, because "widget" is specific to the inspector screen, whereas the logging screen for example would be looking for errored "logs".  "element" seems like a good generic name.

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -278,6 +319,40 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+    }
+    // Update the selected index for the error navigator.
+    if (_selectedErrorIndex != errorIndex) {
+      setState(() => _selectedErrorIndex = errorIndex);
+    }
+    // Additionally, mark this error as "read" so that it isn't counted by the badge
+    // (but is still highlighted as an error).
+    if (errorIndex != null) {
+      serviceManager.errorBadgeManager
+          .markErrorAsRead(InspectorScreen.id, _errors[inspectorRef]);
```
kenzieschmoll, line 347: I think we want to mark all errors as read as soon as we navigate to the inspector screen. Then just depend on the error-navigator UI to guide users to errored widgets.

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -296,3 +371,72 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
     });
   }
 }
+
+class ErrorNavigator extends StatelessWidget {
+  const ErrorNavigator(
+      {Key key,
+      @required this.selectedErrorIndex,
+      @required this.errorCount,
+      @required this.onSelectedErrorIndexChanged})
```
kenzieschmoll, line 380: nit: trailing comma

packages/devtools_app/lib/src/inspector/inspector_tree_flutter.dart
```diff
@@ -27,13 +29,17 @@ class _InspectorTreeRowWidget extends StatefulWidget {
...
     @required this.inspectorTreeState,
+    this.errorText,
   }) : super(key: key);
 
   final _InspectorTreeState inspectorTreeState;
 
   InspectorTreeNode get node => row.node;
   final InspectorTreeRow row;
 
+  /// A description of any error for this row. null if there is no error.
```
kenzieschmoll, line 40: nit: dartdoc - new line after first sentence

```
/// A description of any error for this row.
///
/// This will be null if there is no error.
```

packages/devtools_app/lib/src/inspector/inspector_tree_flutter.dart
```diff
@@ -415,10 +424,14 @@ class _InspectorTreeState extends State<InspectorTree>
                     (context, index) {
                       final InspectorTreeRow row =
                           controller.root?.getRow(index);
+                      final inspectorRef = row.node.diagnostic?.valueRef?.id;
                       return _InspectorTreeRowWidget(
                         key: PageStorageKey(row?.node),
                         inspectorTreeState: this,
                         row: row,
+                        errorText: widget.widgetErrors != null
```
kenzieschmoll, line 432: should we also be checking here that `inspectorRef != null`?

packages/devtools_app/lib/src/inspector/inspector_tree_flutter.dart
```diff
@@ -520,13 +533,17 @@ class InspectorRowContent extends StatelessWidget {
...
+    this.errorText,
   });
 
   final InspectorTreeRow row;
   final InspectorTreeControllerFlutter controller;
   final VoidCallback onToggle;
   final Animation<double> expandArrowAnimation;
 
+  /// A description of any error for this row. null if there is no error.
+  final String errorText;
```
kenzieschmoll, line 545: same comment as above about dart doc and naming

packages/devtools_app/lib/src/inspector/inspector_tree_flutter.dart
```diff
@@ -536,14 +553,16 @@ class InspectorRowContent extends StatelessWidget {
       return const SizedBox();
     }
     Color backgroundColor;
-    if (row.isSelected || row.node == controller.hover) {
-      backgroundColor = row.isSelected
-          ? colorScheme.selectedRowBackgroundColor
-          : colorScheme.hoverColor;
+    if (row.isSelected) {
+      backgroundColor = errorText != null
```
kenzieschmoll, line 557: add a getter `hasError => errorText != null` and use that for all the errorText null checks

packages/devtools_app/lib/src/inspector/inspector_tree_flutter.dart
```diff
@@ -591,5 +613,52 @@ class InspectorRowContent extends StatelessWidget {
...
+      message: errorText,
+      child: Padding(
+        padding: EdgeInsets.only(
+          left: indent,
+          top: (rowHeight - defaultIconSize) / 2,
+        ),
+        child: SizedBox(
+          width: defaultIconSize,
+          height: defaultIconSize,
+          child: Container(
```
kenzieschmoll, line 654: instead of wrapping this container in a SizedBox and Padding, just use the `width`, `height`, and `padding` properties on `Container`

packages/devtools_app/lib/src/inspector/inspector_tree_flutter.dart
```diff
@@ -591,5 +613,52 @@ class InspectorRowContent extends StatelessWidget {
...
+
+    // Wrap with error tooltip/marker if there is an error for this node's widget.
+    return errorText == null
+        ? rowWidget
+        : Stack(
+            children: [
+              rowWidget,
+              ErrorIndicator(
+                errorText: errorText,
+                indent: currentX - rowHeight * 2,
```
kenzieschmoll, line 625: why are we subtracting `rowHeight * 2`? if this is for the width of the error dot, instead subtract some const for that specifically instead of repurposing the `rowHeight` const. You could add some const to the ErrorIndicator class `leadingOffset` or similar.

packages/devtools_app/test/inspector_error_navigator_test.dart
```diff
@@ -0,0 +1,81 @@
...
+
+import 'support/wrappers.dart';
+
+void main() {
+  group('Inspector Error Navigator', () {
+    Future<void> testNavigate(WidgetTester tester,
+        {IconData tapIcon,
+        int errorCount,
+        int startIndex,
+        int expectedIndex}) async {
```
kenzieschmoll, line 17: nit: trailing comma

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -278,6 +319,40 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
           //);
         });
       }
+
+      addAutoDisposeListener(
+          inspectorController.selectedNode, _selectedNodeChanged);
```
kenzieschmoll, line 324: Instead of calling setState on the whole inspector screen, make ErrorNavigator a stateful widget that listens for changes to `inspectorController.selectedNode`. That way we won't have to rebuild the entire tree to update the error navigator

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -296,3 +371,72 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+
+  @override
+  Widget build(BuildContext context) {
+    final label = selectedErrorIndex != null
+        ? 'Error ${selectedErrorIndex + 1}/$errorCount'
+        : 'Errors: $errorCount';
+    return Container(
+      color: devtoolsError,
+      child: Padding(
+        padding: const EdgeInsets.symmetric(horizontal: 12, vertical: 6),
```
kenzieschmoll, line 395: instead of using 12 and 6 for const padding values throughout this code, can we use the padding and spacing consts from theme.dart? defaultSpacing, denseSpacing, densePadding, etc.

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -296,3 +371,72 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+          children: [
+            Padding(
+              padding: const EdgeInsets.only(right: 6),
+              child: Text(label),
+            ),
+            IconButton(
+              padding: EdgeInsets.zero,
+              constraints: const BoxConstraints(),
+              splashRadius: Material.defaultSplashRadius / 2,
+              icon: const Icon(Icons.chevron_left),
```
kenzieschmoll, line 406: can we use `Icons.keyboard_arrow_up` and `Icons.keyboard_arrow_down` to match the navigators for Search in other parts of DevTools?
Screen Shot 2021-02-08 at 9 25 39 AM

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -296,3 +371,72 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+        padding: const EdgeInsets.symmetric(horizontal: 12, vertical: 6),
+        child: Row(
+          children: [
+            Padding(
+              padding: const EdgeInsets.only(right: 6),
+              child: Text(label),
+            ),
+            IconButton(
+              padding: EdgeInsets.zero,
+              constraints: const BoxConstraints(),
```
kenzieschmoll, line 404: why the empty box constraints?

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -296,3 +371,72 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+      color: devtoolsError,
+      child: Padding(
+        padding: const EdgeInsets.symmetric(horizontal: 12, vertical: 6),
+        child: Row(
+          children: [
+            Padding(
+              padding: const EdgeInsets.only(right: 6),
+              child: Text(label),
+            ),
+            IconButton(
```
kenzieschmoll, line 402: do you need to add `iconSize: defaultIconSize` here?

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -296,3 +371,72 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+              padding: EdgeInsets.zero,
+              constraints: const BoxConstraints(),
+              splashRadius: Material.defaultSplashRadius / 2,
```
kenzieschmoll, line 405: let's use `splashRadius: defaultIconSize` to match the search navigator

https://github.com/flutter/devtools/blob/ba53577e5da552501c882b4987354c93bae84d5d/packages/devtools_app/lib/src/common_widgets.dart#L782

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -160,6 +168,39 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+
+    return OutlineDecoration(
+      child: Stack(
+        children: [
+          InspectorTree(
+            controller: summaryTreeController,
+            isSummaryTree: true,
+            widgetErrors: errors,
+          ),
+          if (errors.isNotEmpty)
```
kenzieschmoll, line 184: should this be checking for `errorList.isNotEmpty` to make sure nothing gets out of sync?

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -296,3 +371,72 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+              onPressed: _nextError,
+            ),
+          ],
+        ),
+      ),
+    );
+  }
+
+  void _previousError() {
+    var newIndex = errorCount == 0
```
kenzieschmoll, line 423: errorCount should never be 0 in this widget, right? We don't build the ErrorNavigator is the errorsList is empty

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -296,3 +371,72 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+
+class ErrorNavigator extends StatelessWidget {
+  const ErrorNavigator(
+      {Key key,
+      @required this.selectedErrorIndex,
+      @required this.errorCount,
+      @required this.onSelectedErrorIndexChanged})
+      : super(key: key);
+
+  final int selectedErrorIndex;
```
kenzieschmoll, line 383: use a ValueNotifier for `selectedIndex`. See `matchIndex` and the `nextMatch` and `previousMatch` methods here for a similar example: https://github.com/flutter/devtools/blob/ba53577e5da552501c882b4987354c93bae84d5d/packages/devtools_app/lib/src/ui/search.dart#L54

packages/devtools_app/lib/src/error_badge_manager.dart
jacob314: There will be multiple inspector refs for errors without a deep link but we shouldn't be showing them as we don't know what they reference. For example, if there are inspector refs for each widget up the parent tree, you don't want to highlight them all in red. Showing widgets in the tree in red is distracting so we want to be careful we don't create a bunch of false positives and users will start ignoring this feature.

packages/devtools_app/lib/src/error_badge_manager.dart
jacob314: Element has a specific name in Flutter so I would avoid using it unless you want to imply something is a Flutter Element.

packages/devtools_app/lib/src/error_badge_manager.dart
jacob314: Passing a null inspectorService is ok. Add a TODO, I'm about to add inspectorService to the serviceManager so it is always available if you have a Flutter app.

jacob314:
lgtm from me once Kenzie's comments are addressed.
lgtm

packages/devtools_app/lib/src/error_badge_manager.dart
kenzieschmoll: Maybe erroredItemsForPage instead then

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: I think that's different to what was noted at https://github.com/flutter/devtools/pull/2630#issuecomment-771168346. It's easy enough to change, but I wanted to double-check!

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: Without this, the IconButtons seem to force the container to be quite large:

Screenshot 2021-02-08 at 18 56 38

I'm not certain this is the correct fix though?

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: They're quite small using `defaultIconSize` and felt hard to click. What do you think?

`defaultIconSize`:

Screenshot 2021-02-08 at 18 57 31

Without:

Screenshot 2021-02-08 at 18 57 51

packages/devtools_app/lib/src/inspector/inspector_screen.dart
kenzieschmoll: discussed with jacob offline - let's mark errors as read when you visit the screen. I think there could be iterations we could do in a follow up CL to mark the individual error indicators a little differently once each errored widget has been visited, but that can be done in a later CL. Could be good to add a TODO somewhere in the error navigator code to indicate errors as "visited".

packages/devtools_app/lib/src/inspector/inspector_screen.dart
kenzieschmoll: interesting I agree without looks better

packages/devtools_app/lib/src/inspector/inspector_screen.dart
kenzieschmoll: yeah that's not great either. fine to leave as is - not worth blocking commit over 👍

packages/devtools_app/lib/src/error_badge_manager.dart
DanTup: Done in e74b35db

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: Done in e5ed5548 and 5a0f9afb. I made the error indicator circles become just an outline once an error is read (rather than removing the "read" code or leaving it unused). I'll post a video of how it currently looks below for review.

packages/devtools_app/lib/src/inspector/inspector_tree_flutter.dart
DanTup: Done, though changed this back to `error` and made it the whole `DevToolsError` so that it can be rendered differently based on whether it's `read`.

packages/devtools_app/lib/src/inspector/inspector_tree_flutter.dart
DanTup: Done for width/height, though not padding as that causes the `BoxDecoration` to render in the wrong place (we need it to be inset by the padding, but if the padding is on the Container, the BoxDecoration doesn't consider it).

packages/devtools_app/lib/src/inspector/inspector_tree_flutter.dart
DanTup: It's to account for the other widgets (expand/collapse and guide lines). I've added a `static const double indicatorOffset = -(defaultIconSize * 2 + denseSpacing);` for this and a comment explaining.

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: Yeah, I guess this was overly defensive - removed now!

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: Done! (41863a30)

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: This list is gone now, as the ErrorNavigator takes the map since it handles the selectedNodeChanged code.

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: I've done this in 67382098, although the ValueNotifier is in the screen's state and I'm not sure if that's correct (in the example above it's on a Controller). Let me know if this should be moved somewhere else.

DanTup:
Here's how things currently look:

https://youtu.be/zhvQTamXMtk

The Inspector tab badge is removed when you navigate to it, and the "seen" errors are now marked (showing just an outline instead of a full red dot).

packages/devtools_app/lib/src/error_badge_manager.dart
DanTup: Added in 257f2cbe

DanTup:
I have two outstanding issues:

1. Refreshing the page results in the error navigator showing an incorrect count because errors are recorded that are no longer valid - we currently only filter the errors when you click the Refresh Tree button. Presumably we need to also filter the errors on first load (eg. if a bunch of older log events are processed). I'm unsure where that's best done.

2. I'm periodically seeing the below error in my debug console, although I'm not sure why (it sometimes happens when changing the selected node). I'm sure I've introduced it, but it's happening in code I'm not familiar with and I don't understand what it's doing. 
 
```
TypeError: Cannot read property 'serviceExtensionManager' of null
    at inspector_service.ObjectGroup.new.invokeServiceMethodDaemonParams (http://localhost:55648/packages/devtools_app/src/info/info_controller.dart.lib.js:112187:47)
    at invokeServiceMethodDaemonParams.next (<anonymous>)
    at runBody (http://localhost:55648/dart_sdk.js:39177:34)
    at Object._async [as async] (http://localhost:55648/dart_sdk.js:39208:7)
    at inspector_service.ObjectGroup.new.invokeServiceMethodDaemonParams (http://localhost:55648/packages/devtools_app/src/info/info_controller.dart.lib.js:112185:20)
    at inspector_service.ObjectGroup.new.invokeServiceMethodDaemonArg (http://localhost:55648/packages/devtools_app/src/info/info_controller.dart.lib.js:112168:19)
    at inspector_service.ObjectGroup.new.invokeServiceMethodDaemonInspectorRef (http://localhost:55648/packages/devtools_app/src/info/info_controller.dart.lib.js:112196:19)
    at inspector_service.ObjectGroup.new.invokeServiceMethodReturningNodeInspectorRef (http://localhost:55648/packages/devtools_app/src/info/info_controller.dart.lib.js:112314:53)
    at inspector_service.ObjectGroup.new.getSelection (http://localhost:55648/packages/devtools_app/src/info/info_controller.dart.lib.js:112363:40)
    at getSelection.next (<anonymous>)
    at runBody (http://localhost:55648/dart_sdk.js:39177:34)
    at Object._async [as async] (http://localhost:55648/dart_sdk.js:39208:7)
    at inspector_service.ObjectGroup.new.getSelection (http://localhost:55648/packages/devtools_app/src/info/info_controller.dart.lib.js:112354:20)
```

packages/devtools_app/lib/src/error_badge_manager.dart
```diff
@@ -144,21 +138,33 @@ class ErrorBadgeManager extends DisposableController
...
+    final newValue =
         Map.fromEntries(errors.value.entries.where((e) => isValid(e.key)));
-    _updateErrorCount(screenId);
+    if (newValue.length != oldCount) {
+      errors.value = newValue;
+    }
   }
 
   void markErrorAsRead(String screenId, DevToolsError error) {
+    clearErrors(screenId);
```
kenzieschmoll, line 151: do we need to do this here? shouldn't this be the responsibility of the screen?

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -70,7 +72,7 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
 
   LinkedHashMap<String, DevToolsError> _errors =
       LinkedHashMap<String, DevToolsError>();
-  int _selectedErrorIndex;
+  final ValueNotifier<int> _selectedErrorIndex = ValueNotifier<int>(null);
```
kenzieschmoll, line 75: let's move this to inspector controller and expose a public ValueListenable getter that returns `_selectedErrorIndex`. You can expose a public method `selectError(int index)` that performs what is in `_onSelectedErrorIndexChanged` below.

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -372,45 +350,95 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+  const ErrorNavigator({
+    Key key,
+    @required this.errors,
+    @required this.selectedNode,
+    @required this.selectedErrorIndex,
+  }) : super(key: key);
+
+  final LinkedHashMap<String, InspectableWidgetError> errors;
+  final ValueListenable<InspectorTreeNode> selectedNode;
+  final ValueNotifier<int> selectedErrorIndex;
```
kenzieschmoll, line 363: nit: new line between these members

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -372,45 +350,95 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+  }
+
+  void _selectedNodeChanged() {
+    final node = widget.selectedNode.value;
+    final inspectorRef = node?.diagnostic?.valueRef?.id;
+    // Check whether the node that was just selected has any errors associated
+    // with it.
+    var errorIndex = inspectorRef != null
+        ? widget.errors.keys.toList().indexOf(inspectorRef)
+        : null;
```
kenzieschmoll, line 398: why not -1 instead of null?

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -372,45 +350,95 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+    if (errorIndex == -1) {
+      errorIndex = null;
+    }
+
+    setState(() => widget.selectedErrorIndex.value = errorIndex);
```
kenzieschmoll, line 403: should we only call setState if the value has changed? I think we should avoid calling set state if widget.selectedErrorIndex.value was already null and we are re-assigning to null.

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -372,45 +350,95 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+
+    // Additionally, mark this error as "read" and clear the badge.
+    if (errorIndex != null) {
+      serviceManager.errorBadgeManager.clearErrors(InspectorScreen.id);
```
kenzieschmoll, line 407: I don't think we should do this here. We should allow the screen to clear its errors once it is opened.

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -372,45 +350,95 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+          .markErrorAsRead(InspectorScreen.id, widget.errors[inspectorRef]);
+    }
+  }
 
   @override
   Widget build(BuildContext context) {
-    final label = selectedErrorIndex != null
-        ? 'Error ${selectedErrorIndex + 1}/$errorCount'
-        : 'Errors: $errorCount';
+    final errorIndex = widget.selectedErrorIndex.value;
```
kenzieschmoll, line 415: if the value of `selectedErrorIndex` is updated outside of this widget, the error navigator UI would not update. Instead, wrap the Container in a ValueListenableBuilder that listens to `inspectorController.selectedErrorIndex`

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -372,45 +350,95 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+    final label = errorIndex != null
+        ? 'Error ${errorIndex + 1}/${widget.errors.length}'
+        : 'Errors: ${widget.errors.length}';
     return Container(
       color: devtoolsError,
       child: Padding(
-        padding: const EdgeInsets.symmetric(horizontal: 12, vertical: 6),
+        padding: const EdgeInsets.symmetric(
+            horizontal: defaultSpacing, vertical: denseSpacing),
```
kenzieschmoll, line 423: nit: trailing comma

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -420,23 +448,21 @@ class ErrorNavigator extends StatelessWidget {
...
-        ? null
-        : selectedErrorIndex == null
-            ? 0
-            : (selectedErrorIndex + 1) % errorCount;
-    onSelectedErrorIndexChanged(newIndex);
+    final errorIndex = widget.selectedErrorIndex.value;
+    final newIndex =
+        errorIndex == null ? 0 : (errorIndex + 1) % widget.errors.length;
+
+    setState(() => widget.selectedErrorIndex.value = newIndex);
```
kenzieschmoll, line 466: instead of calling set state here for previous and next, call a method on the controller `selectError(int index)` (referenced in another comment earlier) that sets the value of the notifier. Then this widget will rebuild automatically because it will be wrapped in a ValueListenableBuilder

kenzieschmoll (commented):
This is looking great! A few more comments to move some code into controllers

packages/devtools_app/lib/src/error_badge_manager.dart
DanTup: This was an error - it was already being called before this function. I've removed in 04143cd0!

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: I don't know if I've understood correctly, but I think I've done this in eee1fd77. I ended up moving quite a bit of code into there, and ErrorNavigator became stateless again. Let me know if I've misunderstood/gone too far!

I did wonder if we should move the error stuff into another class that hangs off InspectorController too, as it's quite large and this seems to have made it larger!

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: The honest answer is that -1 made the line short enough that it unwrapped and seemed much less readable than this (and both did the same thing, since the -1 from `indexOf()` is also then converted to null below. Happy to change if you prefer though!

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: This code is gone now, and it's based on ValueListenableBuilders.

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: This was to cover the case where you're already on the Inspector screen and you're reloading, and the badge increases. It felt expensive to be re-filtering the list on every error event, so I thought it may be nice to clear the badge if you then click on an error node (this felt like acknowledgement of the errors).

I wasn't entirely sure about it though, so happy to change (or remove)?

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: This was done as part of eee1fd77.

packages/devtools_app/lib/src/inspector/inspector_screen.dart
DanTup: Done in eee1fd77!

DanTup:
I think I've addressed (or responded to) all comments above now, let me know if I missed anything. I also added a golden test to verify the indicators/navigator are displaying (and updating as you navigate):

Screenshot 2021-02-10 at 19 47 44

The two issues in https://github.com/flutter/devtools/pull/2630#issuecomment-775916113 are still outstanding.

packages/devtools_app/lib/src/inspector/inspector_controller.dart
```diff
@@ -737,9 +741,52 @@ class InspectorController extends DisposableController
...
+        : null;
+    if (errorIndex == -1) {
+      errorIndex = null;
+    }
+
+    _selectedErrorIndex.value = errorIndex;
+
+    // Additionally, mark this error as "read" and clear the badge.
+    if (errorIndex != null) {
+      serviceManager.errorBadgeManager.clearErrors(InspectorScreen.id);
```
kenzieschmoll, line 770: Can you add a comment explaining why this clearErrors is needed?

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -178,32 +157,37 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+                        child: ErrorNavigator(
+                          errors: inspectableErrors,
+                          errorIndex: selectedErrorIndex,
+                          onSelectError: inspectorController.selectErrorByIndex,
+                        ),
+                      ),
+                    )
+                ],
+              );
+            }),
```
kenzieschmoll, line 189: nit: trailing comma

kenzieschmoll (approved):
LGTM with a couple nits

packages/devtools_app/lib/src/inspector/inspector_controller.dart
DanTup: Done! (46d17542)

kenzieschmoll:
lgtm_ship_it

DanTup:
@kenzieschmoll the two issues noted [here](https://github.com/flutter/devtools/pull/2630#issuecomment-775916113) are still outstanding and I'm not sure how best to address.

kenzieschmoll:
> Refreshing the page results in the error navigator showing an incorrect count because errors are recorded that are no longer valid - we currently only filter the errors when you click the Refresh Tree button. Presumably we need to also filter the errors on first load (eg. if a bunch of older log events are processed). I'm unsure where that's best done.

Could you try making the filter call in `_handleConnectionStart`? You can add a method to InspectorController that does this so that the filter call is not duplicated. @jacob314 if you have a better idea of where to do this.

> I'm periodically seeing the below error in my debug console, although I'm not sure why (it sometimes happens when changing the selected node). I'm sure I've introduced it, but it's happening in code I'm not familiar with and I don't understand what it's doing.

It looks like this is the line where the error is being thrown. Can you reproduce this error without your changes? https://github.com/flutter/devtools/blob/231a75f311d7d8de56d9445a0a4093b229a8abcb/packages/devtools_app/lib/src/inspector/inspector_service.dart#L746

DanTup:
> > Refreshing the page results in the error navigator showing an incorrect count because errors are recorded that are no longer valid - we currently only filter the errors when you click the Refresh Tree button. Presumably we need to also filter the errors on first load (eg. if a bunch of older log events are processed). I'm unsure where that's best done.
> 
> Could you try making the filter call in _handleConnectionStart? You can add a method to InspectorController that does this so that the filter call is not duplicated. @jacob314 if you have a better idea of where to do this.

I actually couldn't repro this anymore, however refreshing on the Inspector page would bring the badge back as it processed the errors, so I added a badge clear to the `_handleConnectionStart` method that seems to address this (4ed270d4).

> It looks like this is the line where the error is being thrown. Can you reproduce this error without your changes? Tagging @jacob314 here too in case he has any ideas or familiarity.

Actually, it turns out I can. I was using the same overflow_errors.dart file from this branch, but back on master. I ran the app from the terminal and hot reloaded a few times, and then moving around the nodes in the tree triggered the same error. I'll file an issue, it's probably not related to this.

@kenzieschmoll I might not be around much later today, so if you're happy with the final change above, feel free to merge this (and ping me or revert if it seems to introduce any issues).

PR #2632, opened 2021-01-27
These are no new features refactors to get the LayoutExplorer code

jacob314:
Added a base class for the Layout visualization widgets and moved files around for better sharing as a lot of the Flex code is handy for non flex widgets.


packages/devtools_app/lib/src/inspector/inspector_data_models.dart
```diff
@@ -195,37 +196,21 @@ class LayoutProperties {
 
   String describeHeight() => 'h=${toStringAsFixed(size.height)}';
 
+  // TODO(jacobr): need to also consider offset from parent.
   bool get isOverflowWidth {
-    return width > parent.width + overflowEpsilon;
+    return parent != null && width > parent.width + overflowEpsilon;
```
jacob314, line 201: Add extra error checking to avoid crashing.

packages/devtools_app/lib/src/inspector/inspector_data_models.dart
```diff
@@ -195,37 +196,21 @@ class LayoutProperties {
...
   }
 
+  // TODO(jacobr): need to also consider offset from parent.
   bool get isOverflowHeight {
-    return height > parent.height + overflowEpsilon;
+    return parent != null && height > parent.height + overflowEpsilon;
```
jacob314, line 206: Add extra error checking to avoid crashing.

kenzieschmoll (approved):

Issue #1883, opened 2020-04-28
P1: Visualize box constraints from render objects in Layout Explorer


InMatrix:
Layout Explorer currently only visualizes layout information for FlexBox-based widgets such as Row or Column. It would be useful to visualize the box constraints data for other types of widgets through their respective render objects. The data is already available as a tree node in the detail view of the widget inspector, but it's hard to find. 

In short, the box constraints visualization should at least communicate these three pieces of data:
* offset
* constraints
* size

In [this example](https://dartpad.dev/28b9cad2a19769ecafe91ecf57016ba5), the Column widget has an offset of (0.0, 56.0), constraints of (0.0<=w<=859.0, 0.0<=h<=687.0), and the size of (360.0, 687.0). As mentioned, this data is available in text from the inspector: 

![image](https://user-images.githubusercontent.com/348942/80433906-2e8e0300-88ad-11ea-8bec-07e9206157bd.png)
 
Visualizing a widget's offset, constraints, and size in the same diagram should make it easier to tell whether the widget was trying to be as large as possible or just large enough to fit its child widget(s). 

Cc: @raison00

kenzieschmoll:
@jacob314 as part of this change, can you please also make the Layout Explorer tab first in the details section?

PR #2633, opened 2021-01-27
Layout explorer

jacob314:


jacob314:
Fixes https://github.com/flutter/devtools/issues/1883

This mimics the Flex layout explorer but for Box layout widgets.

One trick is the parent of the Box layout widget may be counter intuitive because while the Flex child of a widget is usually unsurprising, the Box parent of a widget is quite surprising as it may be multiple widgets up in the tree because not all widgets have RenderObjects.
Similarly, unconstrained sizes are less of a big deal if you have a box rather than flex layout so we do not highlight them in red.
Currently I show the render object names for each widget to try to clarify what is going on those names may confuse. We can add tooltips explaining the names that are render objects.
This looks pretty good for padding widgets.
Screen Shot 2021-01-27 at 2 12 39 PM
Notice in this case the parent isn't the GalleryApp as the GalleryApp has exactly the same RenderObject as the ModelBinding. 
Screen Shot 2021-01-27 at 2 36 41 PM

jacob314:
Fyi @albertusdev it was fun having a chance to work with this code again. I was pleased by how much of the existing code could be reused to visualize box layouts.

kenzieschmoll:
DBC: as part of this change, can you also make the layout explorer the primary tab in the details tree?

jacob314:
The other option we could add as a quick followup is to show the layout of the child of the widget for the common case where it has only one child. That would make cases like Padding look nice where you would like to see the widgets impact on its child's layout as well.
The code is structured so I think that would be possible although we might run out of space.

jacob314:
Made the layout explorer the default tab as part of this change and moved it to be the first tab. Not sure that is wise so leaving it as a separate commit in this CL so we can easily pull it if we decide against that.
One idea is to add a list of just the properties of the selected widget in a simple list view bellow the layout explorer so you don't have to go the layout explorer just to see properties.
Crazy idea is to show the properties directly within the box for a widget in the layout explorer. That would save space.

packages/devtools_app/lib/src/inspector/layout_explorer/ui/free_space.dart
```diff
@@ -150,10 +152,15 @@ class PaddingVisualizerWidget extends StatelessWidget {
...
-        height: renderProperties.height,
+        width: safeDouble(renderProperties.width),
+        height: safeDouble(renderProperties.height),
         child: horizontal ? widthWidget : heightWidget,
       ),
     );
   }
 }
+
+double safeDouble(double value) {
```
kenzieschmoll, line 163: we may already have something like this in utils.dart. if not, can you move this method there?

packages/devtools_app/lib/src/inspector/layout_explorer/ui/widget_constraints.dart
```diff
@@ -39,17 +43,17 @@ class VisualizeWidthAndHeightWithConstraints extends StatelessWidget {
...
-                  properties.isOverflowHeight)
+              if (properties.constraints != null) ...[
+                if (!showOverflowHeight) const TextSpan(text: '\n'),
+                TextSpan(
+                  text: ' (${properties.describeHeightConstraints()})',
+                  style: properties.constraints.hasBoundedHeight ||
+                          !warnIfUnconstrained
+                      ? null
+                      : TextStyle(color: colorScheme.unconstrainedColor),
+                )
```
kenzieschmoll, line 54: trailing comma

packages/devtools_app/lib/src/inspector/layout_explorer/ui/widget_constraints.dart
```diff
@@ -97,13 +101,16 @@ class VisualizeWidthAndHeightWithConstraints extends StatelessWidget {
...
-          ),
+          if (properties.constraints != null) ...[
+            if (!showChildrenWidthsSum) const TextSpan(text: '\n'),
+            TextSpan(
+              text: '(${properties.describeWidthConstraints()})',
+              style:
+                  properties.constraints.hasBoundedWidth || !warnIfUnconstrained
+                      ? null
+                      : TextStyle(color: colorScheme.unconstrainedColor),
+            )
```
kenzieschmoll, line 112: nit: trailing comma

packages/devtools_app/lib/src/inspector/layout_explorer/box/box.dart
```diff
@@ -0,0 +1,428 @@
...
+  const BoxLayoutExplorerWidget(
+    InspectorController inspectorController, {
+    Key key,
+  }) : super(inspectorController, key: key);
+
+  static bool shouldDisplay(RemoteDiagnosticsNode node) {
+    // Pretend this layout explorer is always available. This layout explorer
+    // will gracefully fall back to an error message if the required properties
+    // are not needed.
+    // TODO(jacobr) this method could be made asynchronous we could
```
kenzieschmoll, line 32: incomplete comment?

packages/devtools_app/lib/src/inspector/layout_explorer/box/box.dart
```diff
@@ -0,0 +1,428 @@
...
+  }
+
+  @override
+  bool shouldDisplay(RemoteDiagnosticsNode node) {
+    return BoxLayoutExplorerWidget.shouldDisplay(selectedNode);
+  }
+
+  @override
+  AnimatedLayoutProperties computeAnimatedProperties(
+      LayoutProperties nextProperties) {
```
kenzieschmoll, line 56: nit: trailing comma

packages/devtools_app/lib/src/inspector/layout_explorer/box/box.dart
```diff
@@ -0,0 +1,428 @@
...
+  /// On the other hand, a non-null size indicates that the minFractions
+  /// constraints should be obeyed. This is needed to ensure that negative sizes
+  /// are visualized reasonably.
+  // The minFractions aren't exactly obeyed but they are approximated in a way
+  // that keeps this algorithm simple and has the nice property that an initial
+  // value much smaller than the minSize results in a slightly smaller value
+  // than a value that is almost minSize.
+  // In the most extreme case an item will get not minFraction but will instead
+  // get the slightly smaller value of minFraction / (1 + minFraction)
+  // which is close enough for the simple values we need this for.
```
kenzieschmoll, line 114: dart doc? ///

packages/devtools_app/lib/src/inspector/layout_explorer/box/box.dart
```diff
@@ -0,0 +1,428 @@
...
+  static List<double> minFractionLayout({
+    @required double availableSize,
+    @required List<double> sizes,
+    @required List<double> minFractions,
+  }) {
+    assert(sizes.length == minFractions.length);
+    final length = sizes.length;
+    double total = 0.0;
+    total += 1.0; // Add a tiny amount to avoid divide by zero bugs.
```
kenzieschmoll, line 123: can we just initialize to 1.0 instead?

packages/devtools_app/lib/src/inspector/layout_explorer/box/box.dart
```diff
@@ -0,0 +1,428 @@
...
+          sizes: widths,
+          minFractions: minFractions,
+        );
+        final displayHeights = minFractionLayout(
+          availableSize: availableHeight,
+          sizes: heights,
+          minFractions: minFractions,
+        );
+        final widgetWidth = displayWidths[1];
+        final widgetHeight = displayHeights[1];
```
kenzieschmoll, line 205: why index 1?

packages/devtools_app/lib/src/inspector/layout_explorer/box/box.dart
```diff
@@ -0,0 +1,428 @@
...
+        // has a different render object.
+        title: describeBoxName(parentProperties),
+        largeTitle: true,
+        layoutProperties: parentProperties,
+        isSelected: false,
+        child: VisualizeWidthAndHeightWithConstraints(
+          properties: parentProperties,
+          warnIfUnconstrained: false,
+          child: Padding(
+            padding: const EdgeInsets.all(8.0),
```
kenzieschmoll, line 320: use defaultSpacing const (or whichever one is 8.0 in our theme consts)

packages/devtools_app/lib/src/inspector/layout_explorer/ui/free_space.dart
jacob314: done

packages/devtools_app/lib/src/inspector/layout_explorer/ui/widget_constraints.dart
jacob314: done

packages/devtools_app/lib/src/inspector/layout_explorer/box/box.dart
jacob314: done

packages/devtools_app/lib/src/inspector/layout_explorer/box/box.dart
jacob314: done

packages/devtools_app/lib/src/inspector/layout_explorer/box/box.dart
jacob314: done

packages/devtools_app/lib/src/inspector/layout_explorer/box/box.dart
jacob314: done

packages/devtools_app/lib/src/inspector/layout_explorer/box/box.dart
jacob314: commented on what the 3 elements in the widths and heights array are.

packages/devtools_app/lib/src/inspector/layout_explorer/box/box.dart
jacob314: @pq has there ever been discussion of a lint to warn when a raw value is used where a const with the same value could be used?
It would be tricky to get right without false positives. For example you would probably need to see that EdgeInsets.all(denseSpacing) existed in the rest of the code base. This is probably more one of those lints that should be a code suggestion as I write code. As soon as I wrote 8 it should suggest the fix to the similar value used to define EdgeInsets.

pq: As it happens: https://github.com/dart-lang/linter/pull/2429 👍

kenzieschmoll (approved):

albertusdev:
Whoa! Exciting work and plans ahead for Layout Explorer 😍 
Thank you @jacob314! 🚀

Issue #2420, opened 2020-10-08
In recent versions of devtools. Internet access is required to load and so use dev tools.


elansys-kc:
If you open devtools without being connected to the internet then devtools fails to load and the following message is shown in google chromes console.

unpkg.com/canvaskit-wasm@0.17.2/bin/canvaskit.js:1 Failed to load resource: net::ERR_INTERNET_DISCONNECTED

Opening this issue for tracking as requested by @yjbanov in the following issue:
https://github.com/flutter/devtools/issues/2125

elansys-kc:
This is also true of Android Studio. Without internet you get DevTools failed and even closing android studio and re-opening without  internet access prevents use of dev tools, just like the browser.

elansys-kc:
Installing devtools such as described here:
[devtools](https://flutter.dev/docs/development/tools/devtools/cli)

Also still requires unpkg.com access in the browser and android studio still hangs with "installing devtools".

jacob314:
Fyi @helin24 to resolve the Android Studio issue. Seems like it is time for a more offline friendly solution to run devtools than `pub global activate`

dnfield:
I think the problem here is that it needs some way to bundle the CanvasKit WASM blob witht he package rather than pulling it from a CDN.

This probably would require upstream changes in the web engine to accomodate. @yjbanov

dnfield:
If you set `FLUTTER_WEB_CANVASKIT_URL` as a compiler define, you can override the URL of the wasm blob.

I think it'd work if you just included the blob as an asset and overrode that define to point t the asset URL.

dnfield:
https://github.com/flutter/engine/blob/master/lib/web_ui/lib/src/engine/canvaskit/initialization.dart#L51-L79

kevlar700:
Has this regressed?

kevlar700:
It seems that devtools via android studio on the dev channel is useable offline today but still not for web making testing navigation to a local server via ethernet quite tedious

jacob314:
The canvas kit files are now usable offline. The problem is the Android Studio process to activate devtools. We are working to fix that issue and streamline how DevTools is deployed in general by bundling it with the Dart SDK.

kevlar700:
I'm not sure why fonts.gstatic.com for Roboto is required either when I have my own fonts bundled?

jacob314:
Are you talking about your Flutter Web app or Dart DevTools?

kevlar700:
[...]

PR #2634, opened 2021-01-28
bundle canvaskit when releasing

dnfield:
This would fix https://github.com/flutter/devtools/issues/2420

This approach works, but has a couple downsides:

- Someone has to manually update the `CANVASKIT_URL` when it gets bumped in the engine (or when DevTools wants to use a different build anyway).
- This is not necessarily the same version as used in local development

I did it this way to avoid:

- Checking in binary blobs/large minified JS files to this repo
- Having `assets` entries in the pubspec that get populated by a script

I'm not strongly committed to doing it this way. This also lacks automated testing (I've tested it out locally and it works). If there's somewhere that this build script is tested I could update that test, or we could do something like build the app, serve it up and make sure we can load it in a (headless?) browser without any HTTP errors from the server.

At any rate, providing the dart defines like this and packaging up the WASM blobs in some shape or form is the right way to make it all available offline.

### DIFF
--- a/tool/build_release.sh
+++ b/tool/build_release.sh
@@ -10,6 +10,9 @@ set -ex #echo on
 # This URL should be updated to keep in sync with the version from the engine.
 # See https://github.com/flutter/engine/blob/353efcdf3c0ec0ecf0275d55e4a22329397f899f/lib/web_ui/lib/src/engine/canvaskit/initialization.dart#L51-L79,
 # but compare with the code in master for getting the current version.
+# A better solution would be to either upstream this functionality into the flutter_tools,
+# (https://github.com/flutter/flutter/issues/74936), or to read this from a manifest 
+# provided (https://github.com/flutter/flutter/issues/74934).
 function download_canvaskit() {
   local canvaskit_url=https://unpkg.com/canvaskit-wasm@0.22.0/bin
 
--- a/tool/build_release.sh
+++ b/tool/build_release.sh
@@ -10,15 +10,41 @@ pushd packages/devtools_app
 
 rm -rf build
 rm -rf ../devtools/build
+
+# This avoids requiring an internet connection for CanvasKit at runtime.
+# This URL should be updated to keep in sync with the version from the engine.
+# See https://github.com/flutter/engine/blob/353efcdf3c0ec0ecf0275d55e4a22329397f899f/lib/web_ui/lib/src/engine/canvaskit/initialization.dart#L51-L79,
+# but compare with the code in master for getting the current version.
+mkdir -p build/web/assets/canvaskit/profiling
+
+CANVASKIT_URL=https://unpkg.com/canvaskit-wasm@0.22.0/bin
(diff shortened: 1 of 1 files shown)


tool/build_release.sh
```diff
@@ -10,15 +10,42 @@ pushd packages/devtools_app
...
 # Build a profile build rather than a release build to avoid minification
 # as code size doesn't matter very much for us as minification makes some
 # crashes harder to debug. For example, https://github.com/flutter/devtools/issues/2125
 
-flutter build web --pwa-strategy=none --profile --dart-define=FLUTTER_WEB_USE_EXPERIMENTAL_CANVAS_TEXT=true --no-tree-shake-icons
+flutter build web \
+  --pwa-strategy=none \
+  --profile \
+  --dart-define=FLUTTER_WEB_USE_EXPERIMENTAL_CANVAS_TEXT=true \
+  --dart-define=FLUTTER_WEB_CANVASKIT_URL=assets/canvaskit/ \
```
dnfield, line 37: this one and L46 are the actual change to this command.

tool/build_release.sh
```diff
@@ -10,15 +10,41 @@ pushd packages/devtools_app
 
 rm -rf build
 rm -rf ../devtools/build
+
+# This avoids requiring an internet connection for CanvasKit at runtime.
+# This URL should be updated to keep in sync with the version from the engine.
+# See https://github.com/flutter/engine/blob/353efcdf3c0ec0ecf0275d55e4a22329397f899f/lib/web_ui/lib/src/engine/canvaskit/initialization.dart#L51-L79,
+# but compare with the code in master for getting the current version.
```
jacob314, line 17: We need a way to check that automatically or we will soon have bugs due to using the wrong canvaskit version.
I'd be fine with something as crude as grepping that https://github.com/flutter/engine/lib/web_ui/lib/src/engine/canvaskit/initialization.dart 
contains exactly https://unpkg.com/canvaskit-wasm@0.22.0/bin/
and failing if it doesn't.
It might also make sense to rewrite this release script in Dart now that it is getting more complex.

kenzieschmoll: I'm agree jacob that we need to include an automated check in tandem with this change.

dnfield: It'd probably be good if the web engine just included a manifest somewhere of what files/URLs it does/might access at runtime. Checking with some web folks about that.

dnfield: Added a grep for it in the script. This is .. pretty hacky. It'd be a lot better to have a manifest of some sort. But I tried with an invalid canvaskit URL and it works. It will break if the initialization code gets moved around, or Flutter changes internal cache stuff.

I think it's probably ok until some better option is available, since this is just a build script run by developers who hopefully know what's going on.

kenzieschmoll: Can you file a bug to move to a better solution and link to it in comments here?

dnfield: Added a couple bugs.

kenzieschmoll (approved):

PR #2635, opened 2021-01-28
Added tracking when a class is constructed (new).

terrylucas:
Sometimes it is difficult to know which code is leaking or causing too much memory to be allocated. Allocation monitoring was created to see the delta of objects between two points in time.  However, even this useful tool can be hard to pinpoint who or what is causing the memory explosion.

To monitor a particular allocation first start the allocation monitoring by pressing

image

Using the Search field, type in the name of a class to track e.g., ObjectWithUniqueId

image

Adding one of your known classes in your application being profiled in this example "ObjectWithUniqueId" is typed in

image

Selecting the class to the tracking will scroll the class into view and enable the checkbox will add the class to the list of classes to track:

image

now interact with the application your profiling to track the class(es) that might be constructed.

now press the monitor button image again

At the bottom on the monitor pane (right side) are the track classes, expanding the tracked class will display all instances of  classes that were constructed (between when the class to added to tracking list and the monitor button was pressed). In the below example, 14 instances where constructed.

image

Expanding on when (timestamp) the construction occurred will display the call stack of the top 4 entries when the constructor was called:

image

Clicking on "More ..." will display the entire call stack.

image

### DIFF
--- a/case_study/memory_leaks/images_1/lib/main.dart
+++ b/case_study/memory_leaks/images_1/lib/main.dart
@@ -38,16 +38,16 @@ class MyHomePage extends StatefulWidget {
 
 int globalObjectId = 0;
 
-class Tracker {
-  Tracker()
+class ObjectWithUniqueId {
+  ObjectWithUniqueId()
       : now = DateTime.now(),
-        trackerId = globalObjectId++;
+        uniqueId = globalObjectId++;
 
   DateTime now = DateTime.now();
-  int trackerId = globalObjectId++;
+  int uniqueId = globalObjectId++;
 
   @override
-  String toString() => 'Collected @ $now, id=$trackerId';
+  String toString() => 'Collected @ $now, id=$uniqueId';
 }
 
 class MyHomePageState extends State<MyHomePage>
@@ -62,12 +62,12 @@ class MyHomePageState extends State<MyHomePage>
     _tabController = TabController(length: tabs.length, vsync: this);
   }
 
-  final trackers = <Tracker>[];
+  final objects = <ObjectWithUniqueId>[];
 
   void devToolsPostEvent(String eventName, Map<String, Object> eventData) {
     developer.postEvent('DevTools.Event_$eventName', eventData);
 
-    trackers.add(Tracker());
+    objects.add(ObjectWithUniqueId());
   }
 
   Widget recordLoadedImage(ImageChunkEvent imageChunkEvent, String imageUrl) {
--- a/case_study/memory_leaks/images_1/lib/main.dart
+++ b/case_study/memory_leaks/images_1/lib/main.dart
@@ -36,6 +36,20 @@ class MyHomePage extends StatefulWidget {
   MyHomePageState createState() => MyHomePageState();
 }
 
+int globalObjectId = 0;
+
+class Tracker {
+  Tracker()
+      : now = DateTime.now(),
+        trackerId = globalObjectId++;
+
+  DateTime now = DateTime.now();
+  int trackerId = globalObjectId++;
+
+  @override
+  String toString() => 'Collected @ ${now}, id=$trackerId';
+}
+
 class MyHomePageState extends State<MyHomePage>
     with SingleTickerProviderStateMixin {
   TabController _tabController;
@@ -48,8 +62,12 @@ class MyHomePageState extends State<MyHomePage>
     _tabController = TabController(length: tabs.length, vsync: this);
   }
 
+  final trackers = <Tracker>[];
+
   void devToolsPostEvent(String eventName, Map<String, Object> eventData) {
     developer.postEvent('DevTools.Event_$eventName', eventData);
+
+    trackers.add(Tracker());
   }
 
   Widget recordLoadedImage(ImageChunkEvent imageChunkEvent, String imageUrl) {
--- a/case_study/memory_leaks/images_1/lib/main.dart
+++ b/case_study/memory_leaks/images_1/lib/main.dart
@@ -47,7 +47,7 @@ class Tracker {
   int trackerId = globalObjectId++;
 
   @override
-  String toString() => 'Collected @ ${now}, id=$trackerId';
+  String toString() => 'Collected @ $now, id=$trackerId';
 }
 
 class MyHomePageState extends State<MyHomePage>
--- a/packages/devtools_app/lib/src/memory/memory_allocation_table_data.dart
+++ b/packages/devtools_app/lib/src/memory/memory_allocation_table_data.dart
@@ -9,6 +9,28 @@ import 'memory_protocol.dart';
 
 const defaultNumberFieldWidth = 100.0;
 
+class FieldTrace extends ColumnData<ClassHeapDetailStats> {
+  FieldTrace() : super('✔', fixedWidthPx: 50.0);
+
+  @override
+  int getValue(ClassHeapDetailStats dataObject) =>
+      dataObject.isStacktraced ? 1 : 0;
+
+  @override
+  String getDisplayValue(ClassHeapDetailStats dataObject) =>
+        dataObject.isStacktraced ? '✔' : ' '; // U+2714 ✔ HEAVY CHECK MARK
+
+  @override
+  bool get supportsSorting => true;
+
+  @override
+  int compare(ClassHeapDetailStats a, ClassHeapDetailStats b) {
+    final Comparable valueA = getValue(a);
+    final Comparable valueB = getValue(b);
+    return valueA.compareTo(valueB);
+  }
+}
+
 class FieldClassName extends ColumnData<ClassHeapDetailStats> {
   FieldClassName() : super('Class', fixedWidthPx: 200.0);
 
--- a/packages/devtools_app/lib/src/memory/memory_allocation_table_data.dart
+++ b/packages/devtools_app/lib/src/memory/memory_allocation_table_data.dart
@@ -4,21 +4,29 @@
 
 import 'dart:math';
 
+import 'package:flutter/material.dart';
+import 'package:provider/provider.dart';
+
+import '../table.dart';
 import '../table_data.dart';
+import 'memory_controller.dart';
 import 'memory_protocol.dart';
 
 const defaultNumberFieldWidth = 100.0;
 
-class FieldTrace extends ColumnData<ClassHeapDetailStats> {
-  FieldTrace() : super('✔', fixedWidthPx: 50.0);
+class FieldTrack extends ColumnData<ClassHeapDetailStats>
+    implements ColumnRenderer<ClassHeapDetailStats> {
+  FieldTrack()
+      : super(
+          'Track',
+          titleTooltip: 'Track Class Allocations',
+          fixedWidthPx: 55.0,
+          alignment: ColumnAlignment.left,
+        );
 
   @override
   int getValue(ClassHeapDetailStats dataObject) =>
       dataObject.isStacktraced ? 1 : 0;
-
-  @override
-  String getDisplayValue(ClassHeapDetailStats dataObject) =>
-        dataObject.isStacktraced ? '✔' : ' '; // U+2714 ✔ HEAVY CHECK MARK
 
   @override
   bool get supportsSorting => true;
@@ -28,6 +36,24 @@ class FieldTrace extends ColumnData<ClassHeapDetailStats> {
     final Comparable valueA = getValue(a);
     final Comparable valueB = getValue(b);
     return valueA.compareTo(valueB);
+  }
+
+  @override
+  Widget build(
+    BuildContext context,
+    ClassHeapDetailStats item, {
+    bool isRowSelected = false,
+  }) {
+    final controller = Provider.of<MemoryController>(context);
+
+    return Checkbox(
+      value: item.isStacktraced,
+      onChanged: (value) {
+        item.isStacktraced = value;
+        controller.setTracking(item.classRef, value);
+        controller.changeStackTraces();
+      },
+    );
   }
 }
 
--- a/packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
+++ b/packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
@@ -4,10 +4,13 @@
 
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
+import 'package:vm_service/vm_service.dart';
 
 import '../auto_dispose_mixin.dart';
+import '../split.dart';
 import '../table.dart';
 import '../table_data.dart';
+import '../theme.dart';
 import '../utils.dart';
 import 'memory_allocation_table_data.dart';
 import 'memory_controller.dart';
@@ -25,12 +28,15 @@ class AllocationTableViewState extends State<AllocationTableView>
 
   final List<ColumnData<ClassHeapDetailStats>> columns = [];
 
+  var samplesProcessed = <ClassRef, CpuSamples>{};
+
   @override
   void initState() {
     super.initState();
 
     // Setup the columns.
     columns.addAll([
+      FieldTrace(),
       FieldClassName(),
       FieldInstanceCountColumn(),
       FieldInstanceDeltaColumn(),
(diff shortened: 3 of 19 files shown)


jacob314:
High level UX feedback:
I would expect most of the time I want to track a class after noticing something troubling about the objects allocated of the class.
Perhaps add a column of "track" checkboxes similar to the checkbox icon you currently show?
Do you have an example problem app with a problem that can be solved with the tracker data? That could help us iterate on the UI with some concrete examples in mind. Having it be a separate table makes me nervous particularly if a user is tracking allocations for multiple different classes but wants to drill down into allocations for just one of them.
I would suggest we instead integrate it with the existing allocation tracker display now and in the future with the snapshot display once we can associate allocations with snapshots. For example, from the allocation tracker table I would want to see all allocations for the class I selected.
The stack trace for an allocation means nothing to me unless I know what class was allocated. Maybe I'm missing where that is in the screenshots.

case_study/memory_leaks/images_1/lib/main.dart
```diff
@@ -36,6 +36,20 @@ class MyHomePage extends StatefulWidget {
   MyHomePageState createState() => MyHomePageState();
 }
 
+int globalObjectId = 0;
+
+class Tracker {
```
jacob314, line 41: nit: can we use a different name as this sounds like the name of the feature it is testing which makes the UI confusing. Maybe just something simple like
ObjectWithUniqueId.
I think my confusion in previous comments was from seeing the word Tracker in the table.

packages/devtools_app/lib/src/memory/memory_allocation_table_data.dart
```diff
@@ -9,6 +9,28 @@ import 'memory_protocol.dart';
 
 const defaultNumberFieldWidth = 100.0;
 
+class FieldTrace extends ColumnData<ClassHeapDetailStats> {
```
jacob314, line 12: lets use a real checkbox and make these clickable.

packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
```diff
@@ -58,28 +64,166 @@ class AllocationTableViewState extends State<AllocationTableView>
         controller.computeAllLibraries(rebuild: true);
       });
     });
+
+    addAutoDisposeListener(controller.updateClassStackTraces, () {
+      setState(() {});
+    });
+
+    addAutoDisposeListener(controller.treeChangedNotifier, () {
+      if (controller.isTreeChanged) {
```
jacob314, line 73: why does this fire if the tree hasn't changed?

packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
```diff
@@ -58,28 +64,166 @@ class AllocationTableViewState extends State<AllocationTableView>
...
+        setState(() {});
+      }
+    });
+  }
+
+  /// Construct all objects tracked and their stacktraces.
+  List<Widget> allTrackedCreations(ClassRef classRef) {
```
jacob314, line 80: This should use ListView or preferably our TreeTable class instead. As written this will kill Flutter Web performance once there are a lot of allocations.
If we need to add some features to the TreeTable I think that would be world be worth it. For example, having an option to truncate to only show the first N (4) children by default and support expanding to show the rest of the content.

packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
```diff
@@ -58,28 +64,166 @@ class AllocationTableViewState extends State<AllocationTableView>
...
+      );
+    } else {
+      return null;
+    }
+  }
+
+  Widget trackedClassesList() {
+    return controller.trackAllocations.isEmpty
+        ? const Padding(
+            padding: EdgeInsets.only(top: 15),
```
jacob314, line 175: use the existing size constants. there is a normalSpacing that is 16 and small that is 8. Using those will keep the UI a little more consistent.

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -101,6 +101,125 @@ ChartInterval chartInterval(String displayName) {
...
 }
 
+class AllocationStackTrace {
+  AllocationStackTrace(CpuSample sample, {List<ProfileFunction> functions}) {
+    _computeStacktrace(sample, functions: functions);
+  }
+
+  int timestamp;
+
+  var showFullStacktrace = false;
```
jacob314, line 111: var -> bool

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -101,6 +101,125 @@ ChartInterval chartInterval(String displayName) {
...
+
+  int get stackDepth => stacktrace.length;
+
+  static const NoClasses = 0;
+  static const DartClasses = 1;
+  static const FlutterClasses = 2;
+
+  static const DartFlutterClasses = DartClasses | FlutterClasses;
+
+  bool hideDartClasses(int hideClasses) =>
```
jacob314, line 125: Fyi @kenzieschmoll I think you have wanted to filter stack traces from the Dart core libraries and Flutter as well. @terrylucas is this method used currently? I suspect implementing it will be hard as the stacktraces currently lack the info needed.
One other item to hide and show is native frames.
This is something we should filter consistently across DevTools.

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -126,6 +245,14 @@ class MemoryController extends DisposableController
   static const libraryRootNode = '___LIBRARY___';
   static const classRootNode = '___CLASSES___';
 
+  final _updateClassStackTraces = ValueNotifier(0);
+
+  ValueListenable<int> get updateClassStackTraces => _updateClassStackTraces;
+
+  void changeStackTraces() {
```
jacob314, line 252: maybe add a comment on what this does?

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -661,11 +664,68 @@ class HeapTreeViewState extends State<HeapTree>
     selectFromSearchField(controller, foundName);
   }
 
+  Widget trackerWidget(MemoryController controller) {
```
jacob314, line 667: Is there a use case where we need this instead of manually selecting classes?

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -661,11 +664,68 @@ class HeapTreeViewState extends State<HeapTree>
...
+                  // TODO(terry): Use search auto-complete.
+                  debugLogger('Class $trackingName not found');
+                } else {
+                  final foundClass = controller.monitorAllocations.firstWhere(
+                    (element) => element.classRef.name == trackingName,
+                    orElse: () => null,
+                  );
+                  foundClass.isStacktraced = true;
+                  controller.setTracking(foundClass.classRef, true);
+                  textController.clear();
```
jacob314, line 706: we need to display a toast at least the first time this is done that the results only show up for new allocations.

packages/devtools_app/lib/src/vm_service_wrapper.dart
```diff
@@ -221,6 +221,44 @@ class VmServiceWrapper implements VmService {
...
+  }) async {
+    final Map<String, dynamic> args = {};
+    if (classId != null) {
+      args['classId'] = classId;
+    }
+    if (enable != null) {
+      args['enable'] = enable ? true : false;
+    }
+
+    final response = await trackFuture(
```
jacob314, line 238: Fyi @bkonyi we need to make these or similar APIs public instead of using a private API.

packages/devtools_app/lib/src/vm_service_wrapper.dart
```diff
@@ -221,6 +221,44 @@ class VmServiceWrapper implements VmService {
...
+    String isolateId, {
+    String classId,
+  }) async {
+    final Map<String, dynamic> args = {};
+    if (classId != null) {
+      args['classId'] = classId;
+    }
+    final response = await trackFuture(
+      '_getAllocationSamples',
+      callMethod('_getAllocationSamples', isolateId: isolateId, args: args),
```
jacob314, line 256: @bkonyi same issue with this API

terrylucas:
> High level UX feedback:
> I would expect most of the time I want to track a class after noticing something troubling about the objects allocated of the class.
> Perhaps add a column of "track" checkboxes similar to the checkbox icon you currently show?
> Do you have an example problem app with a problem that can be solved with the tracker data? That could help us iterate on the UI with some concrete examples in mind. Having it be a separate table makes me nervous particularly if a user is tracking allocations for multiple different classes but wants to drill down into allocations for just one of them.
> I would suggest we instead integrate it with the existing allocation tracker display now and in the future with the snapshot display once we can associate allocations with snapshots. For example, from the allocation tracker table I would want to see all allocations for the class I selected.
> The stack trace for an allocation means nothing to me unless I know what class was allocated. Maybe I'm missing where that is in the screenshots.

Checkboxes looked like a bit of work to add in TableRow, I do think it's a better UI.  However the scenario you're suggesting is how it works.  I think the confusion might be that I added a Tracker to the case study to test where find when the user's class e.g., Tracker as found.

case_study/memory_leaks/images_1/lib/main.dart
terrylucas: Done.

jacob314:
> > High level UX feedback:
> > I would expect most of the time I want to track a class after noticing something troubling about the objects allocated of the class.
> > Perhaps add a column of "track" checkboxes similar to the checkbox icon you currently show?
> > Do you have an example problem app with a problem that can be solved with the tracker data? That could help us iterate on the UI with some concrete examples in mind. Having it be a separate table makes me nervous particularly if a user is tracking allocations for multiple different classes but wants to drill down into allocations for just one of them.
> > I would suggest we instead integrate it with the existing allocation tracker display now and in the future with the snapshot display once we can associate allocations with snapshots. For example, from the allocation tracker table I would want to see all allocations for the class I selected.
> > The stack trace for an allocation means nothing to me unless I know what class was allocated. Maybe I'm missing where that is in the screenshots.
> 
> Checkboxes looked like a bit of work to add in TableRow, I do think it's a better UI. However the scenario you're suggesting is how it works. I think the confusion might be that I added a Tracker to the case study to test where to find the user's class e.g., Tracker as found.

You should be able to add checkboxes to a table row by specifying a builder for the ui for each entry in the row. If that is hard for the current table impl, we should clean it up. The table used in the logging view shows some examples of this.

packages/devtools_app/lib/src/vm_service_wrapper.dart
```diff
@@ -213,7 +213,7 @@ class VmServiceWrapper implements VmService {
     if (await isProtocolVersionSupported(
         supportedVersion: SemanticVersion(major: 3, minor: 18))) {
       return trackFuture(
-        'getAllocationProfile',
+        '_getAllocationProfile',
```
bkonyi, line 216: Why is this being changed to `_getAllocationProfile`? The current string references the existing public RPC.

packages/devtools_app/lib/src/vm_service_wrapper.dart
bkonyi: Yeah, I'd strongly prefer that we avoid using private methods outside of VM developer mode. I'm working on a public implementation now, so it would be great if we can wait on landing this PR until we have that in place. I'm much more comfortable having a DevTools out in the wild that checks to see if there's support for `setTraceClassAllocation` and falls back to `_setTraceClassAllocation` than just blindly calling `_setTraceClassAllocation`, which will likely be renamed and break with newer service targets.

bkonyi (changes requested):
I'm working on making `_setTraceClassAllocation` public, so please wait until support has landed in `package:vm_service` before landing.

I'll be checking this in.  We wont we publishing devtools for a number of weeks so once you checkin new vm_service I'll update to use published name.  Will not have version.

packages/devtools_app/lib/src/vm_service_wrapper.dart
bkonyi: I also have a public implementation in the works for this.

packages/devtools_app/lib/src/memory/memory_allocation_table_data.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
terrylucas: Yeah I've re-written using TreeTable can't nest treeTable in another treetable but I'll look at how to change Table to support that.

packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
terrylucas: Re-wrote using TreeTable not needed.

packages/devtools_app/lib/src/memory/memory_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: Re-wrote and eliminated using TreeTable and auto-complete search.

packages/devtools_app/lib/src/vm_service_wrapper.dart
terrylucas: okay, let me know when its ready.

packages/devtools_app/lib/src/vm_service_wrapper.dart
terrylucas: okay, ditto let me know when ready.

packages/devtools_app/lib/src/vm_service_wrapper.dart
terrylucas: It was accidentally changed - shouldn't have been.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
```diff
@@ -0,0 +1,259 @@
...
+
+    // Add the entire call stack as leaf nodes.
+    final totalCallStack = item.stacktrace.length;
+    for (var index = leafNodesCount - 1; index < totalCallStack; index++) {
+      allocation.addChild(TrackerCall(item.stacktrace[index]));
+    }
+    allocation.expandCascading();
+  }
+
+  void createTrackerTreeSample() {
```
jacob314, line 200: this looks like test code move it to a test directory file.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
```diff
@@ -0,0 +1,259 @@
...
+    assert(false, 'Unknown dataObject');
+  }
+
+  @override
+  int compare(Tracker a, Tracker b) {
+    Comparable valueA = getValue(a);
+    Comparable valueB = getValue(b);
+    if (a is TrackerCall) {
+      // Sort is always by child index for the stacktrace, call stack should not reorder.
+      valueA = a.children.indexOf(a);
```
jacob314, line 86: this seems odd. why does the children of a contain a? Add a comment explaining more what is going on.
Naively seems like we are just returning -1 a lot.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
terrylucas: Yes it is odd in many ways.  Returning -1 happens to not sort (which is what I wanted).  Also,  `valueB = b.children.indexOf(a)` is wrong too.  Must have been a brain fart.

The two lines should have been:

```
      valueA = a.parent.children.indexOf(a);
      valueB = b.parent.children.indexOf(b);
```

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_controller.dart
terrylucas: This is currently controlled by the filter dialog in memory profiler.  Should this filtering be a a higher level (top-level setting)?  Then all stack traces, classes displayed in snapshots, allocation monitoring, etc. would use the same filtering.

I'll add an issue https://github.com/flutter/devtools/issues/2685

packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
terrylucas: It only fires if the tree has changed true or false.  However, the tree never resets back to false once fired (value == true). So the test is meaningless - removed.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: Added TODO.  We have a message bus for toast but I don't see a toast to display when event is received. Do we have a toast in DevTools (I thought we did maybe it never got ported from HTML to Flutter).

bkonyi:
Can we make sure this isn't published in any release until [this CL](https://dart-review.googlesource.com/c/sdk/+/182666) lands? This change as-is will break with newer VM service implementations once it lands.

terrylucas:
Yes, this will not be published for a few weeks (2/24).  We should have a
new VMService by then - right?

bkonyi:
Yes, we should. I'm still waiting on a review for the CL above but I think Ryan is occupied with something else at the moment. I'll push harder for a review if it's not ready today.

PR #2637, opened 2021-01-28
Added settings dialog for memory page.

terrylucas:
Add ability to disable automatic ADB collection of android memory statistics.  Collecting Android memory metrics will not resemble normal operations while metrics are collected.  ADB metric collection can cause lots of GCs in the Android environment.

image

### DIFF
--- a/packages/devtools_app/lib/src/memory/memory_screen.dart
+++ b/packages/devtools_app/lib/src/memory/memory_screen.dart
@@ -22,6 +22,7 @@ import '../octicons.dart';
 import '../screen.dart';
 import '../theme.dart';
 import '../ui/label.dart';
+import '../ui/utils.dart';
 import '../utils.dart';
 
 import 'memory_android_chart.dart' as android;
@@ -136,6 +137,9 @@ class MemoryBodyState extends State<MemoryBody>
 
   OverlayEntry hoverOverlayEntry;
   OverlayEntry legendOverlayEntry;
+
+  /// Updated when the MemoryController's _androidCollectionEnabled ValueNotifier changes.
+  bool isAndroidCollection = MemoryController.androidADBDefault;
 
   @override
   void initState() {
@@ -286,9 +290,11 @@ class MemoryBodyState extends State<MemoryBody>
     });
 
     addAutoDisposeListener(controller.androidCollectionEnabled, () {
-      final isAndroidCollection = controller.androidCollectionEnabled.value;
+      isAndroidCollection = controller.androidCollectionEnabled.value;
       setState(() {
         if (!isAndroidCollection && controller.isAndroidChartVisible) {
+          // If we're no longer collecting android stats then hide the
+          // chart and disable the Android Memory button.
           controller.toggleAndroidChartVisibility();
         }
       });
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -282,6 +286,15 @@ class MemoryBodyState extends State<MemoryBody>
       }
     });
 
+    addAutoDisposeListener(controller.androidCollectionEnabled, () {
+      setState(() {
+        if (!controller.androidCollectionEnabled.value &&
+            controller.isAndroidChartVisible) {
+          controller.toggleAndroidChartVisibility();
```
kenzieschmoll, line 293: what if the chart was previously not visible, and the checkbox was toggled to true? do we also need to toggle chart visibility in this case?

packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -282,6 +286,15 @@ class MemoryBodyState extends State<MemoryBody>
       }
     });
 
+    addAutoDisposeListener(controller.androidCollectionEnabled, () {
+      setState(() {
```
kenzieschmoll, line 290: can we add a local var `androidCollectionEnabled` that tracks the value of `controller.androidCollectionEnabled` and is updated in `setState()`? then we can use that local var below.

packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -1420,3 +1453,46 @@ class ChartsValues {
     results[adbGraphicsJsonName] = androidData.graphics;
   }
 }
+
+class MemoryConfigurationsDialog extends StatelessWidget {
+  const MemoryConfigurationsDialog(this.controller);
+
+  static const dialogWidth = 700.0;
```
kenzieschmoll, line 1460: does this match the other settings dialogs? we should use a consistent width. If there isn't already a defaultDialogWidth const, feel free to create one and place it where our other default const values live

packages/devtools_app/lib/src/memory/memory_screen.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_screen.dart
terrylucas: This will enable/disable the Android Memory button if on.  I don't think to display a new chart when enabled makes sense.  However, when android collection is off I do hide the android chart - because it's useless if there is no data.

I'm not sure flipping on is necessary as the data is record and when clicked on the all collected data is available.

packages/devtools_app/lib/src/memory/memory_screen.dart
terrylucas: Yes it matches exactly, although I'm not convinced I need that size.  But a common is probably a good idea.

packages/devtools_app/lib/src/memory/memory_screen.dart
terrylucas: Done.

terrylucas:
PTAL, thanks.

packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -287,9 +286,9 @@ class MemoryBodyState extends State<MemoryBody>
     });
 
     addAutoDisposeListener(controller.androidCollectionEnabled, () {
+      final isAndroidCollection = controller.androidCollectionEnabled.value;
```
kenzieschmoll, line 289: Sorry if I was unclear - I meant local to this class, not the callback context. Then you'll be able to use that local variable in the build method below.

kenzieschmoll (approved):
LGTM after one comment is fixed

packages/devtools_app/lib/src/memory/memory_screen.dart
terrylucas: Done.

PR #2640, opened 2021-01-28
Run flutter format.

jacob314:
Looks like with the switch to github actions we lost the check that ran the formatter on every commit.
Fyi @devoncarew


devoncarew:
Hmm, it's still there (https://github.com/flutter/devtools/blob/master/tool/bots.sh#L92), but might no longer be working.

The results and failure should show up in the 'main' bot.

devoncarew:
Yeah, I suspect we need a fix to bots.sh: https://github.com/flutter/devtools/runs/1787276940?check_suite_focus=true#step:3:353.

kenzieschmoll (approved):

Issue #2642, opened 2021-01-28
P0: Exception thrown from layout explorer


kenzieschmoll:
Interestingly, I can run this app from command line, go to the layout explorer, and select a box widget with no issues. However, if I run the same app from intelliJ, and then connect it to my running Devtools instance, I get this error. @jacob314 

Marking this P0 because we should ideally get a fix in before the release (it also affects the tooling demo).

```
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
The following JSNoSuchMethodError was thrown building LayoutExplorerTab(state:
_LayoutExplorerTabState#96f9d):
NoSuchMethodError: invalid member on null: 'inspectorService'

The relevant error-causing widget was:
  LayoutExplorerTab
  file:///Users/kenzieschmoll/develop/devtools/packages/devtools_app/lib/src/inspector/inspector_screen_details_
  tab.dart:65:9

When the exception was thrown, this was the stack:
packages/devtools_app/src/inspector/layout_explorer/ui/layout_explorer_widget.dart 254:41
[_updateObjectGroupManager]
packages/devtools_app/src/inspector/layout_explorer/ui/layout_explorer_widget.dart 150:5   didUpdateWidget
packages/flutter/src/widgets/framework.dart 4696:57                                        update
packages/flutter/src/widgets/framework.dart 3299:14                                        updateChild
packages/flutter/src/widgets/framework.dart 4534:16                                        performRebuild
packages/flutter/src/widgets/framework.dart 4681:11                                        performRebuild
packages/flutter/src/widgets/framework.dart 4203:5                                         rebuild
packages/flutter/src/widgets/framework.dart 4713:5                                         update
packages/flutter/src/widgets/framework.dart 3299:14                                        updateChild
packages/flutter/src/widgets/framework.dart 4534:16                                        performRebuild
packages/flutter/src/widgets/framework.dart 4203:5                                         rebuild
packages/flutter/src/widgets/framework.dart 4590:5                                         update
packages/flutter/src/widgets/framework.dart 3299:14                                        updateChild
packages/flutter/src/widgets/framework.dart 5995:14                                        update
packages/flutter/src/widgets/framework.dart 3299:14                                        updateChild
packages/flutter/src/widgets/framework.dart 5995:14                                        update
packages/flutter/src/widgets/framework.dart 3299:14                                        updateChild
packages/flutter/src/widgets/framework.dart 4534:16                                        performRebuild
[...]

PR #2644, opened 2021-01-29
Make layout explorer the default tab.

jacob314:
Fix some lifecycle bugs in the inspector and layout explorer code so this is ok.
Fixed a lifecycle bug in the service manager as well.

### DIFF
--- a/packages/devtools_app/lib/src/inspector/inspector_controller.dart
+++ b/packages/devtools_app/lib/src/inspector/inspector_controller.dart
@@ -149,22 +149,6 @@ class InspectorController extends DisposableController
     }
   }
 
-  final List<VoidCallback> _selectionListeners = [];
-
-  void addSelectionListener(VoidCallback listener) {
-    _selectionListeners.add(listener);
-  }
-
-  void removeSelectionListener(VoidCallback listener) {
-    _selectionListeners.remove(listener);
-  }
-
-  void notifySelectionListeners() {
-    for (var notifyListener in _selectionListeners) {
-      notifyListener();
-    }
-  }
-
   int _clientCount = 0;
 
   /// Maximum frame rate to refresh the inspector at to avoid taxing the
@@ -224,7 +208,8 @@ class InspectorController extends DisposableController
 
   bool programaticSelectionChangeInProgress = false;
 
-  InspectorTreeNode selectedNode;
+  ValueListenable<InspectorTreeNode> get selectedNode => _selectedNode;
+  final ValueNotifier<InspectorTreeNode> _selectedNode = ValueNotifier(null);
 
   InspectorTreeNode lastExpanded;
 
@@ -241,7 +226,8 @@ class InspectorController extends DisposableController
 
   bool get detailsSubtree => parent != null;
 
-  RemoteDiagnosticsNode get selectedDiagnostic => selectedNode?.diagnostic;
+  RemoteDiagnosticsNode get selectedDiagnostic =>
+      selectedNode.value?.diagnostic;
 
   FlutterTreeType getTreeType() {
     return treeType;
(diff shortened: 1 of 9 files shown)


packages/devtools_app/lib/src/inspector/inspector_controller.dart
```diff
@@ -224,7 +208,8 @@ class InspectorController extends DisposableController
 
   bool programaticSelectionChangeInProgress = false;
 
-  InspectorTreeNode selectedNode;
+  ValueListenable<InspectorTreeNode> get selectedNode => _selectedNode;
```
jacob314, line 211: use ValueListenable pattern instead so we can auto dispose.

packages/devtools_app/lib/src/inspector/inspector_tree_flutter.dart
```diff
@@ -77,17 +77,20 @@ class _InspectorTreeRowState extends State<_InspectorTreeRowWidget>
...
-  InspectorControllerClient get client => _client;
-  InspectorControllerClient _client;
+  final Set<InspectorControllerClient> _clients = {};
 
-  set client(InspectorControllerClient value) {
-    if (_client == value) return;
-    // Do not set a new client if there is still an old client.
-    assert(value == null || _client == null);
-    _client = value;
+  void addClient(InspectorControllerClient value) {
```
jacob314, line 82: due to order disposed methods were called the previous checks in this method were too strict and caused issues.

packages/devtools_app/lib/src/inspector/layout_explorer/layout_explorer.dart
```diff
@@ -30,9 +32,11 @@ class _LayoutExplorerTabState extends State<LayoutExplorerTab>
     if (BoxLayoutExplorerWidget.shouldDisplay(node)) {
       return BoxLayoutExplorerWidget(controller);
     }
-    return const Center(
+    return Center(
       child: Text(
-        'Currently, Layout Explorer only supports Box and Flex-based widgets.',
+        node != null
+            ? 'Currently, Layout Explorer only supports Box and Flex-based widgets.'
+            : 'Select a widget to view its layout.',
```
jacob314, line 39: We could also leave the layout explorer window completely blank when no widget is selected. @InMatrix @cobblest any thoughts?

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -755,21 +755,15 @@ class ServiceExtensionManager extends Disposer {
...
           return;
         }
-        for (String extension in isolate.extensionRPCs) {
-          await _maybeAddServiceExtension(extension);
-          if (isolateRef != _mainIsolate.value) {
-            // Isolate has changed again.
-            return;
-          }
-        }
+        await Future.wait([
```
jacob314, line 758: Without this we had super confusing bugs as our app state would reach an async gap with only some of the service extensions added. Now they are all added at the same block.

jacob314:
Fixes https://github.com/flutter/devtools/issues/2642

packages/devtools_app/lib/src/inspector/layout_explorer/layout_explorer.dart
kenzieschmoll: Or should we always select the root if no widget is selected?

InMatrix: Leave the layout explorer window there could help discovery. As an enhancement to the current message, can we suggest a widget to inspect? For example, the second line can be modified to 'Try it out: select the first `Row` widget to view its layout.`  [Go button].

kenzieschmoll (approved):

packages/devtools_app/lib/src/inspector/layout_explorer/layout_explorer.dart
jacob314: I like the suggestion of finding the first Row widget. Just need to know what alternate message to show if there are no Row/Column widgets in the app.

InMatrix: How about linking to online documentation in that case? We could also consider embedding a short video demo in that space.

PR #2645, opened 2021-01-29
Button cleanup

kenzieschmoll:
We were using buttons of different styles and heights throughout DevTools. This PR cleans that up.

### DIFF
--- a/packages/devtools_app/lib/src/common_widgets.dart
+++ b/packages/devtools_app/lib/src/common_widgets.dart
@@ -103,7 +103,7 @@ class IconLabelButton extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return OutlinedButton(
+    return FixedHeightOutlinedButton(
       onPressed: onPressed,
       child: MaterialIconLabel(
         icon,
@@ -196,13 +196,11 @@ class RecordButton extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return OutlinedButton(
+    return IconLabelButton(
       onPressed: recording ? null : onPressed,
-      child: MaterialIconLabel(
-        Icons.fiber_manual_record,
-        labelOverride ?? 'Record',
-        includeTextWidth: includeTextWidth,
-      ),
+      icon: Icons.fiber_manual_record,
+      label: labelOverride ?? 'Record',
+      includeTextWidth: includeTextWidth,
     );
   }
 }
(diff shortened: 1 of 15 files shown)


packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -229,12 +227,111 @@ class StopRecordingButton extends StatelessWidget {
...
+  }
+}
+
+class SettingsOutlinedButton extends StatelessWidget {
+  const SettingsOutlinedButton({
+    @required this.onPressed,
+    @required this.tooltip,
+  });
+
+  final void Function() onPressed;
```
jacob314, line 245: nit: `void Function()` to VoidCallback?

packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -229,12 +227,111 @@ class StopRecordingButton extends StatelessWidget {
...
+
+  final ButtonStyle style;
+
+  final double width;
+
+  final String tooltip;
+
+  @override
+  Widget build(BuildContext context) {
+    final button = Container(
```
jacob314, line 290: nit: Container --> SizedBox

packages/devtools_app/lib/src/dialogs.dart
```diff
@@ -104,3 +104,21 @@ class DialogApplyButton extends StatelessWidget {
     );
   }
 }
+
+class DialogTextButton extends StatelessWidget {
+  const DialogTextButton({this.onPressed, this.child});
+
+  final void Function() onPressed;
```
jacob314, line 111: VoidCallback

packages/devtools_app/lib/src/dialogs.dart
```diff
@@ -104,3 +104,21 @@ class DialogApplyButton extends StatelessWidget {
...
+  final Widget child;
+
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+        height: defaultButtonHeight,
+        child: TextButton(
+          onPressed: onPressed,
+          child: child,
+        ));
```
jacob314, line 122: nit: trailing comma

packages/devtools_app/lib/src/landing_screen.dart
```diff
@@ -154,12 +154,14 @@ class _LandingScreenBodyState extends State<LandingScreenBody> {
...
         ),
-        const Padding(
-          padding: EdgeInsets.only(left: 20.0),
-        ),
-        ElevatedButton(
-          child: const Text('Connect'),
-          onPressed: connectDebounce.invoke,
+        const SizedBox(width: defaultSpacing),
+        Padding(
+          // Padding to center the connect button with the `TextField`.
```
jacob314, line 159: if the font size is different will we lose this centering?
Can you use mainAxisAlignment instead to make things look good? or maybe crossAxisAlignment? I'd have to use the layout explorer to remember which.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/landing_screen.dart
kenzieschmoll: rearranged some of these widgets so we could use axis alignment instead.

PR #2647, opened 2021-01-29
Select the first widget after the [root] widget.

jacob314:


jacob314:
Kenzie's idea of selecting the root widget or actually the first child of the [root] widget.
This will actually be nice once we start filtering to only show a build method at a time as this makes the root build method be expanded.
This matching what chrome devtools does and is a bit less work than displaying UI to find the first Row.

jacob314:
@InMatrix this matches what Chrome DevTools does and is easier than the alternatives to implement.

packages/devtools_app/lib/src/inspector/inspector_controller.dart
```diff
@@ -620,6 +620,26 @@ class InspectorController extends DisposableController
     try {
       final RemoteDiagnosticsNode newSelection = await pendingSelectionFuture;
       if (group.disposed) return;
+      if (newSelection == null && firstFrame) {
```
kenzieschmoll, line 623: does this code play well with the selection from query parameters above? i.e. can a widget still be selected from a query parameter on load or will the root widget be selected instead.

jacob314: yep. it only triggers if the selection is null. If query parameters are a busted selection it will mean the root is selected.

kenzieschmoll (approved):

Issue #2649, opened 2021-01-30
DevTools 0.9.7 crashing when used with Flutter 1.22


bkonyi:
Now DevTools are not even running after the update.
![image](https://user-images.githubusercontent.com/6624048/106348186-4c7c0e80-62ea-11eb-86f4-d586e4b13148.png)

_Originally posted by @sunnykinger in https://github.com/flutter/devtools/issues/2398#issuecomment-770160929_

PR #2650, opened 2021-01-30
Handle missing DDS instance gracefully

bkonyi:
Flutter stable currently doesn't start a DDS instance, so calling `getDartDevelopmentServiceVersion()` will result in a method not found error being returned. Handle this case and always return false from `_isDdsVersionSupportedNow` if DDS is not present.

Fixes https://github.com/flutter/devtools/issues/2649.


devoncarew (commented):
rslgtm (through someone more familiar w/ the code may want to review later)

PR #2665, opened 2021-02-02
Support selection within the TextView and fix bug showing pause location.

jacob314:
Fyi @bkony, looks like syntax highlighting broken this. For some reason, the fixed width font isn't really exactly fixed width once you have a bunch of separate TextSpans.
I'm sure there is a better way to do this but this was quick and easy.

Also added single line text selection as it is better than nothing.


jacob314:
Fyi @grouma I think you were seeing this issue.

grouma:
I was experiencing something similar for our internal build. It wasn't clear to me if it was a font loading issue or this one though.

kenzieschmoll (approved):

PR #2666, opened 2021-02-02
Expandable console support

jacob314:


jacob314 (commented):
We should rewrite how we are testing the debugger to use Flutter integration tests similar to what has been done for the inspector tests. That will let us use golden image tests to verify the UI state.

jacob314:
Fyi @grouma, @annagrin

kenzieschmoll:
can you add a screenshot or gif to your PR description? thanks

kenzieschmoll (approved):

packages/devtools_app/lib/src/tree.dart
```diff
@@ -13,10 +13,27 @@ class TreeView<T extends TreeNode<T>> extends StatefulWidget {
...
+  /// has unbounded constraints in the [scrollDirection], then [shrinkWrap] must
+  /// be true.
+  ///
+  /// Shrink wrapping the content of the scroll view is significantly more
+  /// expensive than expanding to the maximum allowed size because the content
+  /// can expand and contract during scrolling, which means the size of the
+  /// scroll view needs to be recomputed whenever the scroll position changes.
+  ///
+  /// Defaults to false.
+  final bool shrinkWrap;
```
grouma, line 35: Is this a copied comment? It doesn't seem appropriate here?

grouma (approved):

jacob314:
Here is a screenshot.
Screen Shot 2021-02-02 at 8 51 24 AM

jacob314:
One note is the console has a mix of fonts. We could make everything monospace if we want.

packages/devtools_app/lib/src/tree.dart
jacob314: Yeah this is the boilerplate Flutter comment on what shrinkWrap does. I have now swapped to the TLDR of when it is needed.

jacob314:
I think we should make all expandable variables in the debugger monospace font for consistency. Tweaking that.

jacob314:
Please review the follow up changes since the last review. I needed to make a small refactor to resolve fighting between SelectableText widgets and InkWell widgets. Also fixed some tests that were broken and made JSON dumps fully selectable.

kenzieschmoll (approved):

Issue #2547, opened 2020-12-10
Migrate off ValueListenableProvider as it is deprecated


grouma:
See https://github.com/flutter/devtools/pull/2548

Issue #2377, opened 2020-09-25
P2: show when app is paused from pages other than the debugger page


devoncarew:
We should show when the app is paused from pages other than the debugger page.

One option for this is to:

- move the 'device info' status line item from the lower right portion of the DevTools' UI to an icon item in the upper right of the screen
- add a run state / app status item to the right hand side of the status line ('running' / 'paused' / 'paused at an exception'), and include some simple controls to resume the app if paused
- include a badge on the 'Debugger' tab to indicate when the app is paused (a (P) badge icon, or, a breakpoint icon)

PR #2676, opened 2021-02-04
Add floating debugger controls when app is paused

kenzieschmoll:
This will show up when an app is paused (except when the debugger screen is the current screen)

Screen Shot 2021-02-04 at 10 53 37 AM

Fixes https://github.com/flutter/devtools/issues/2377
Fixes https://github.com/flutter/devtools/issues/2547

### DIFF
--- a/packages/devtools_app/lib/src/common_widgets.dart
+++ b/packages/devtools_app/lib/src/common_widgets.dart
@@ -568,8 +568,8 @@ class Badge extends StatelessWidget {
 
 /// A widget, commonly used for icon buttons, that provides a tooltip with a
 /// common delay before the tooltip is shown.
-class ActionButton extends StatelessWidget {
-  const ActionButton({
+class DevToolsTooltip extends StatelessWidget {
+  const DevToolsTooltip({
     @required this.tooltip,
     @required this.child,
   });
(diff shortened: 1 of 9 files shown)


packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -271,6 +271,7 @@ class DebuggerController extends DisposableController
     if (isolate.pauseEvent != null &&
         isolate.pauseEvent.kind != EventKind.kResume) {
       _lastEvent = isolate.pauseEvent;
+      print('pausing from switchToIsolate');
```
bkonyi, line 274: Debug print?

kenzieschmoll: whoops. fixed.

packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -789,6 +789,39 @@ Widget inputDecorationSuffixButton(IconData icon, VoidCallback onPressed) {
...
+  @override
+  Widget build(BuildContext context) {
+    final color = borderColor ?? Theme.of(context).focusColor;
+    final childrenWithOutlines = <Widget>[];
+    for (int i = 0; i < children.length; i++) {
+      childrenWithOutlines.addAll([
+        Container(
+          decoration: BoxDecoration(
+            border: Border(
+              left: i == 0 ? BorderSide(color: color) : BorderSide.none,
```
jacob314, line 808: why is only the left side special cased and not the right side.?

kenzieschmoll: We do this to make sure there are not any double borders. The right side of the previous element will serve as the left border for the next element (for all except the first since the first does not have a previous element)

packages/devtools_app/lib/src/debugger/debugger_screen.dart
```diff
@@ -370,3 +370,73 @@ class _DebuggerStatusState extends State<DebuggerStatus> with AutoDisposeMixin {
     return 'paused$reason$fileName $pos';
   }
 }
+
+class FloatingDebuggerControls extends StatelessWidget {
+  @visibleForTesting
+  static const debuggerControlsKey = Key('debugger controls');
```
jacob314, line 376: general nit: why not find the buttons by their text instead of adding keys that are only used for testing?

kenzieschmoll: done.

packages/devtools_app/lib/src/debugger/debugger_screen.dart
```diff
@@ -370,3 +370,73 @@ class _DebuggerStatusState extends State<DebuggerStatus> with AutoDisposeMixin {
...
+  @visibleForTesting
+  static const debuggerControlsStepKey = Key('debugger controls - step');
+
+  @override
+  Widget build(BuildContext context) {
+    final controller = Provider.of<DebuggerController>(context);
+    return ValueListenableBuilder<bool>(
+      valueListenable: controller.isPaused,
+      builder: (context, paused, _) {
+        if (!paused) return const SizedBox();
```
jacob314, line 381: nit: rather than having this become a sized box when not paused consider making the ui unclickable when not paused and providing an animated transition showing and hiding the ui.

jacob314 (approved):
lgtm
with a couple minor nits.

packages/devtools_app/lib/src/debugger/debugger_screen.dart
kenzieschmoll: good idea. done - see pr description for gif

PR #2680, opened 2021-02-05
Fix lifecycle bug in network page

kenzieschmoll:
In `didChangeDependencies` we have several calls to `addAutoDisposeListener`, provided by a mixin. In `dispose` we called `_networkController?.stopRecording()`, which causes one of these listeners to fire, before we called `super.dispose()`. `super.dispose()` will dispose of the listeners because dispose will be called on the mixin, so we need to call `super.dispose()` first, before we do anything else that may cause some of these listeners to fire (the listeners call `setState` so this created a lifecycle bug where we were calling `setState` on an already-disposed state.

@pq is there a way we can warn users of this sort of thing with a lint?

### DIFF
--- a/packages/devtools_app/lib/src/network/network_screen.dart
+++ b/packages/devtools_app/lib/src/network/network_screen.dart
@@ -114,8 +114,6 @@ class _NetworkScreenBodyState extends State<NetworkScreenBody>
     with AutoDisposeMixin, SearchFieldMixin<NetworkScreenBody> {
   NetworkController _networkController;
 
-  bool recording;
-
   NetworkRequests requests;
 
   List<NetworkRequest> filteredRequests;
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/network/network_screen.dart
```diff
@@ -158,8 +158,8 @@ class _NetworkScreenBodyState extends State<NetworkScreenBody>
 
   @override
   void dispose() {
-    _networkController?.stopRecording();
     super.dispose();
+    _networkController?.stopRecording();
```
jacob314, line 162: calls to super dispose should always be first. Lets try to fix this a different way.

kenzieschmoll: Cancelled listeners in `AutoDisposeMixin` in `deactivate` instead of `dispose`. This ensures the listeners are cancelled before we do anything in `dispose` that may trigger listeners we added in `didChangeDependencies` via `AutoDisposeMixin.addAutoDisposeListener`

pq:
> @pq is there a way we can warn users of this sort of thing with a lint?

If we can define the desired semantics, for sure.  @jacob314 and I have chatted a bit about related analyses.  Something like this would be great.

If you have specific ideas or just want to bike-shed, feel free to [open an issue on the linter](https://github.com/dart-lang/linter/issues/new/choose).  Thanks!

jacob314 (approved):
lgtm

PR #2681, opened 2021-02-05
Rev sse version

grouma:
Version `3.7.0` of `package:sse` now buffers requests before the connection is established. `onOpen` is now deprecated. This fixes an issue where we were trying to complete the service upon reconnects.


jacob314 (approved):
lgtm

Issue #2682, opened 2021-02-06
Step in, step over and step out don't work after a refresh


grouma:
Steps to reproduce:

1) Set a breakpoint
2) Hit the breakpoint in your running application
3) Refresh DevTools
4) Notice that the state is properly set but only resume is enabled

kenzieschmoll:
Sounds like the state of `DebuggerController.isPaused` isn't getting set properly on refresh

grouma:
I believe the pause state is correct since you can select resume. It's the rest of the functionality that isn't correctly updated.

kenzieschmoll:
Ah you're right. This is the problem: `_hasFrames` is set to false: https://github.com/flutter/devtools/blob/4f76dfe77e478a7e5a67d0318d51b652a5a1728c/packages/devtools_app/lib/src/debugger/debugger_controller.dart#L101 when the controls are built https://github.com/flutter/devtools/blob/ccc6c60750dd9e195127600d6a5888e1e014fe90/packages/devtools_app/lib/src/debugger/controls.dart#L28

PR #2683, opened 2021-02-08
Fix bug with debugger stepping buttons state

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/2682 and cleans up some state management in the debugger controls

### DIFF
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -112,13 +112,11 @@ class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
 
   @override
   void dispose() {
-    super.dispose();
-
     gutterController.dispose();
     textController.dispose();
-
     widget.controller.scriptLocation
         .removeListener(_handleScriptLocationChanged);
+    super.dispose();
   }
 
   void _initScriptInfo() {
--- a/packages/devtools_app/lib/src/debugger/controls.dart
+++ b/packages/devtools_app/lib/src/debugger/controls.dart
@@ -3,114 +3,138 @@
 // found in the LICENSE file.
 
 import 'package:flutter/material.dart' hide Stack;
+import 'package:provider/provider.dart';
 import 'package:vm_service/vm_service.dart';
 
+import '../auto_dispose_mixin.dart';
 import '../common_widgets.dart';
 import '../theme.dart';
 import '../ui/label.dart';
 import 'debugger_controller.dart';
 import 'scripts.dart';
 
-class DebuggingControls extends StatelessWidget {
-  const DebuggingControls({Key key, @required this.controller})
-      : super(key: key);
+class DebuggingControls extends StatefulWidget {
+  const DebuggingControls({Key key}) : super(key: key);
 
-  final DebuggerController controller;
+  @override
+  _DebuggingControlsState createState() => _DebuggingControlsState();
+}
+
+class _DebuggingControlsState extends State<DebuggingControls>
+    with AutoDisposeMixin {
+  DebuggerController controller;
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+    controller = Provider.of<DebuggerController>(context);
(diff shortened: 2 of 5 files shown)


packages/devtools_app/lib/src/debugger/controls.dart
```diff
@@ -3,114 +3,138 @@
...
+    addAutoDisposeListener(controller.resuming);
+    addAutoDisposeListener(controller.stackFramesWithLocation);
+  }
 
   @override
   Widget build(BuildContext context) {
+    final isPaused = controller.isPaused.value;
+    final resuming = controller.resuming.value;
+    final hasStackFrames = controller.stackFramesWithLocation.value.isNotEmpty;
+    final canStep = isPaused && !resuming && hasStackFrames;
```
grouma, line 41: Why do you need stack frames to be able to step?

kenzieschmoll: before we were checking for `hasFrames`, but the state wasn't getting propagated to the buttons, so I left the requirement of needing frames. Maybe you don't need frames at all to be able to step? that would probably be a question for you.

Looking at the code, the step calls call the resume method on the vm service and just pass an isolate id as well as a StepOption, so maybe frames aren't needed. In what scenario would we have 0 stack frames while paused? and in that case if there is not a top stack frame, what code is there to step through?

devoncarew: Yeah, from memory, there are times you can be paused but don't have any frames. You still want to be able to resume execution in those cases.

kenzieschmoll: > Yeah, from memory, there are times you can be paused but don't have any frames. You still want to be able to resume execution in those cases.

To be clear, Resume would still be enabled in this case - as the code is, though, the stepping buttons would not. Do we want to be able to step in this case as well as resume?

devoncarew: Ah, no. Just having resume enabled would be sufficient. If we don't have an actual location - at least one frame - then stepping doesn't make sense.

kenzieschmoll: Okay - in that case, this code achieves that
- when paused with a top frame, resume and step buttons are enabled
- when paused without a top frame, resume is enabled but step buttons are disabled

grouma: You can pause your application at any time so on the web side you can be paused in some random JS / SDK function. We won't show frames if paused in one of these locations but nothing prevents stepping from working. For the web at least, we continue stepping until you get to the next known Dart location.

I can see the argument to not support stepping in these scenarios however.

kenzieschmoll: I think this CL is fine to leave as is for now, as it mimics the previous intended behavior and handling the paused with 0 frames case seems like a larger issue we should think through. Filed https://github.com/flutter/devtools/issues/2687 to discuss and identify steps forward.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -112,13 +112,11 @@ class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
...
   @override
   void dispose() {
-    super.dispose();
-
     gutterController.dispose();
     textController.dispose();
-
     widget.controller.scriptLocation
         .removeListener(_handleScriptLocationChanged);
+    super.dispose();
```
jacob314, line 119: @pq is there a lint we should be opting into to catch calling super first instead of last?

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/debugger/codeview.dart
pq: No.  We had one (now deprecated) for enforcing calling super last in constructor initializers (https://dart-lang.github.io/linter/lints/super_goes_last.html).  Rolling one for dispose calls is an interesting idea.  Especially if we can generalize somehow.

Issue #2697, opened 2021-02-11
Selection in layout explorer does not scroll into view.


jacob314:
Repro: click between children in a column/row. They do not scroll into view on selection change. This is confusing particularly when multiple children have the same title.
![scroll_layout_explorer](https://user-images.githubusercontent.com/1226812/107600055-6e716d00-6bd7-11eb-98ec-348564357f3c.gif)

PR #2698, opened 2021-02-11
Flex layout polish.

jacob314:
Tweak layout fractions and ensure the selected widget is visible even when view is a bit less tall.
Fix https://github.com/flutter/devtools/issues/2697 where the selected child in the flex layout wouldn't scroll into view.
Fix bug with scrollbars in flex layout. It is safer to manually manage the scroll controller than to hope the right thing will happen.
fix_scroll_explorer

### DIFF
--- a/packages/devtools_app/lib/src/inspector/layout_explorer/flex/flex.dart
+++ b/packages/devtools_app/lib/src/inspector/layout_explorer/flex/flex.dart
@@ -10,6 +10,7 @@ import 'package:flutter/rendering.dart';
 import 'package:flutter/widgets.dart';
 
 import '../../../math_utils.dart';
+import '../../../theme.dart';
 import '../../diagnostics_node.dart';
 import '../../inspector_controller.dart';
 import '../../inspector_data_models.dart';
@@ -38,6 +39,8 @@ class FlexLayoutExplorerWidget extends LayoutExplorerWidget {
 
 class _FlexLayoutExplorerWidgetState extends LayoutExplorerWidgetState<
     FlexLayoutExplorerWidget, FlexLayoutProperties> {
+  final scrollController = ScrollController();
+
   Axis get direction => properties.direction;
 
   Color horizontalColor(ColorScheme colorScheme) =>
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/inspector/layout_explorer/flex/flex.dart
```diff
@@ -101,104 +104,6 @@ class _FlexLayoutExplorerWidgetState extends LayoutExplorerWidgetState<
     });
   }
 
-  Widget _visualizeFlex(BuildContext context) {
```
jacob314, line 104: Fyi @scheglov the extract widget refactor almost handled this case perfectly.  It did have trouble adding multiple named parameters with the same name (for example properties) that should have all been one parameter. A reference to this also caused problems and I had to manually add that parameter.

packages/devtools_app/lib/src/inspector/layout_explorer/flex/flex.dart
```diff
@@ -452,6 +364,149 @@ class _FlexLayoutExplorerWidgetState extends LayoutExplorerWidgetState<
...
+  final ScrollController scrollController;
+  final Axis direction;
+  final _FlexLayoutExplorerWidgetState state;
+
+  @override
+  _VisualizeFlexChildrenState createState() => _VisualizeFlexChildrenState();
+}
+
+class _VisualizeFlexChildrenState extends State<VisualizeFlexChildren> {
+  LayoutProperties lastHighlighted;
```
jacob314, line 390: This is the only change to this stateful widget relative to the original code.

packages/devtools_app/lib/src/inspector/layout_explorer/flex/flex.dart
```diff
@@ -452,6 +364,149 @@ class _FlexLayoutExplorerWidgetState extends LayoutExplorerWidgetState<
...
+  static final selectedChildKey = GlobalKey(debugLabel: 'selectedChild');
+
+  @override
+  Widget build(BuildContext context) {
+    if (lastHighlighted != widget.highlighted) {
```
jacob314, line 395: begin of new code to add the highlighted widget.

packages/devtools_app/lib/src/inspector/layout_explorer/flex/flex.dart
```diff
@@ -452,6 +364,149 @@ class _FlexLayoutExplorerWidgetState extends LayoutExplorerWidgetState<
...
+            widget.scrollController.position.ensureVisible(
+              selectedRenderObject,
+              alignment: 0.5,
+              duration: defaultDuration,
+            );
+          }
+        });
+      }
+    }
+    if (!widget.properties.hasChildren) {
```
jacob314, line 412: The rest of this code is just refactored from before.

packages/devtools_app/lib/src/inspector/layout_explorer/flex/flex.dart
```diff
@@ -452,6 +364,149 @@ class _FlexLayoutExplorerWidgetState extends LayoutExplorerWidgetState<
...
+        final freeSpacesWidgets = [
+          for (var renderProperties in [...mainAxisSpaces, ...crossAxisSpaces])
+            FreeSpaceVisualizerWidget(renderProperties),
+        ];
+        return Scrollbar(
+          isAlwaysShown: true,
+          controller: widget.scrollController,
+          child: SingleChildScrollView(
+            scrollDirection: widget.properties.direction,
+            controller: widget.scrollController,
```
jacob314, line 477: except for this line to specify the scrollController so that ScrollView doesn't get confused and scrollbars appear on the first frame and are interactive.

packages/devtools_app/lib/src/inspector/layout_explorer/flex/flex.dart
```diff
@@ -452,6 +364,149 @@ class _FlexLayoutExplorerWidgetState extends LayoutExplorerWidgetState<
   }
 }
 
+class VisualizeFlexChildren extends StatefulWidget {
+  const VisualizeFlexChildren({
+    Key key,
+    @required this.state,
```
kenzieschmoll, line 370: I don't think this is new code, but it seems strange that we are passing a state around. Is this needed? from what I can tell it isn't used anywhere in this widget.

kenzieschmoll (approved):
LGTM with one cleanup request

packages/devtools_app/lib/src/inspector/layout_explorer/flex/flex.dart
jacob314: it is needed and I agree this is an anti-pattern. I looked for a cheap way to avoid it for this CL but it would have been more of a refactor than I wanted to add here. Filed https://github.com/flutter/devtools/issues/2701 to track.

PR #2708, opened 2021-02-17
Fix noisy assertion error in timeline processing code

kenzieschmoll:
A child can have a different id than its parent if the parent-child relationship was specified via the "parentId" arg. This assertion was firing in part because of what the assertion was checking, and also because we were not returning from the recursion once the event was added. This PR cleans both those issues up and adds a test to verify that the assertion does not fire for cases like this.

### DIFF
--- a/packages/devtools_app/lib/src/performance/performance_model.dart
+++ b/packages/devtools_app/lib/src/performance/performance_model.dart
@@ -938,7 +938,7 @@ class AsyncTimelineEvent extends TimelineEvent {
   /// The return value will be used to stop the recursion early.
   bool endAsyncEvent(TraceEventWrapper eventWrapper) {
     assert(
-      parentId == null ? asyncId == eventWrapper.event.id : true,
+      parentId != null || asyncId == eventWrapper.event.id,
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/performance/performance_model.dart
```diff
@@ -932,23 +932,29 @@ class AsyncTimelineEvent extends TimelineEvent {
...
 
-  void endAsyncEvent(TraceEventWrapper eventWrapper) {
+  /// End the async event with [eventWrapper] and return whether or not the
+  /// end event was successfully added.
+  ///
+  /// The return value will be used to stop the recursion early.
+  bool endAsyncEvent(TraceEventWrapper eventWrapper) {
     assert(
-      asyncId == eventWrapper.event.id,
+      parentId == null ? asyncId == eventWrapper.event.id : true,
```
jacob314, line 941: nit: simpler to read as `parentId != null || asyncId == eventWrapper.event.id`

jacob314 (approved):
lgtm

PR #2710, opened 2021-02-17
Request focus from flame chart keyboard listener

kenzieschmoll:
Fixes a bug where the keyboard navigation in the flame chart was inconsistently focused, and therefore did not always work.


jacob314 (approved):
lgtm

Issue #2664, opened 2021-02-02
Call stack "view all" is not sticky


Hixie:
The call stack sometimes truncates itself and shows a "view all" button. When you press it, the list goes truncates again when you step forward in the code.

Ideally this list would be sticky. An alternative if this is a performance optimization would be to only compute as many lines as is necessary to fill the call stack window, using a lazy list.

jacob314:
Using a lazy list is easy enough polish and will make the UI less annoying to use.
Fyi @grouma @kenzieschmoll

grouma:
One other alternative is to put the `SHOW ALL` button in the list itself. So users scroll to the bottom and click the button and the button is replace with additional frames keeping the list sticky.

PR #2711, opened 2021-02-17
auto-populate call frames

devoncarew:
- auto-populate call frames
- fix https://github.com/flutter/devtools/issues/2664

This removes the 'show all' button from the call frames area and instead auto-populates them. We don't have quite enough info from the vm service protocol to implement a lazy list (the ability to request a set of frames at an offset and knowing the total number of frames for a stack). Instead we request 12 frames, show those once we have the info, and request all the frames in the background. 12 frames roughly corresponds to the number of visible frames.

This PR also removes the code to initially just show one in frame in the UI and populate a small set of frames in the background. That feature was added when retrieving web frames could be very slow. Web frame population is faster now, and, just showing one for a brief time causes flashing on the screen.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/debugger_controller.dart
+++ b/packages/devtools_app/lib/src/debugger/debugger_controller.dart
@@ -633,8 +633,11 @@ class DebuggerController extends DisposableController
 
     // We populate the first 12 frames; this ~roughly corresponds to the number
     // of visible stack frames.
-    _getStackOperation =
-        CancelableOperation.fromFuture(_getStackInfo(limit: 12));
+    const initialFrameRequestCount = 12;
+
+    _getStackOperation = CancelableOperation.fromFuture(_getStackInfo(
+      limit: initialFrameRequestCount,
+    ));
     final stackInfo = await _getStackOperation.value;
     _populateFrameInfo(
       stackInfo.frames,
--- a/packages/devtools_app/lib/src/debugger/debugger_controller.dart
+++ b/packages/devtools_app/lib/src/debugger/debugger_controller.dart
@@ -31,8 +31,10 @@ final _log = DebugTimingLogger('debugger', mute: true);
 /// well.
 class ConsoleLine {
   factory ConsoleLine.text(String text) => TextConsoleLine(text);
+
   factory ConsoleLine.variable(Variable variable) =>
       VariableConsoleLine(variable);
+
   ConsoleLine._();
 }
 
@@ -258,7 +260,7 @@ class DebuggerController extends DisposableController
       _breakpoints.value = [];
       _breakpointsWithLocation.value = [];
       await _getStackOperation?.cancel();
-      _populateFrameInfo([]);
+      _populateFrameInfo([], truncated: false);
       return;
     }
 
@@ -612,6 +614,7 @@ class DebuggerController extends DisposableController
   }
 
   final _hasTruncatedFrames = ValueNotifier<bool>(false);
+
   ValueListenable<bool> get hasTruncatedFrames => _hasTruncatedFrames;
 
   CancelableOperation<_StackInfo> _getStackOperation;
@@ -624,31 +627,24 @@ class DebuggerController extends DisposableController
 
     // Perform an early exit if we're not paused.
     if (!paused) {
-      _populateFrameInfo([]);
+      _populateFrameInfo([], truncated: false);
       return;
     }
 
-    // First, notify based on the single 'pauseEvent.topFrame' frame.
-    if (pauseEvent?.topFrame != null) {
-      final tempFrames = _framesForCallStack(
-        [pauseEvent.topFrame],
-        reportedException: pauseEvent?.exception,
-      );
-      _populateFrameInfo(
-        [await _createStackFrameWithLocation(tempFrames.first)],
-        truncated: true,
-      );
-      _log.log('created first frame');
-    }
-
-    // We populate the first 10 frames to match the behavior in VS Code.
+    // We populate the first 12 frames; this ~roughly corresponds to the number
+    // of visible stack frames.
     _getStackOperation =
-        CancelableOperation.fromFuture(_getStackInfo(limit: 10));
+        CancelableOperation.fromFuture(_getStackInfo(limit: 12));
     final stackInfo = await _getStackOperation.value;
     _populateFrameInfo(
       stackInfo.frames,
       truncated: stackInfo.truncated ?? false,
     );
+
+    // In the background, populate the rest of the frames.
+    if (stackInfo.truncated) {
+      unawaited(_getFullStack());
+    }
   }
 
   Future<_StackInfo> _getStackInfo({int limit}) async {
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -624,31 +627,24 @@ class DebuggerController extends DisposableController
...
-      );
-      _log.log('created first frame');
-    }
-
-    // We populate the first 10 frames to match the behavior in VS Code.
+    // We populate the first 12 frames; this ~roughly corresponds to the number
+    // of visible stack frames.
     _getStackOperation =
-        CancelableOperation.fromFuture(_getStackInfo(limit: 10));
+        CancelableOperation.fromFuture(_getStackInfo(limit: 12));
```
jacob314, line 637: nit: make this a constant so you can stick your comment next to the constant.

packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -624,31 +627,24 @@ class DebuggerController extends DisposableController
...
     final stackInfo = await _getStackOperation.value;
     _populateFrameInfo(
       stackInfo.frames,
       truncated: stackInfo.truncated ?? false,
     );
+
+    // In the background, populate the rest of the frames.
+    if (stackInfo.truncated) {
+      unawaited(_getFullStack());
```
jacob314, line 646: I think we need to rate limited requests to _getFullStack. You can use the RateLimiter class to make sure that only 1 request is running at a time. Otherwise if you step quickly you might end up with 20 pending stack trace requests which could bog down flutter web. Would be nice if the VM Service had a way for us to cancel requests or detect that a getFullStack request should be ignored if execution resumed since when the request was issued.

packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -624,31 +627,27 @@ class DebuggerController extends DisposableController
...
-      );
-      _populateFrameInfo(
-        [await _createStackFrameWithLocation(tempFrames.first)],
-        truncated: true,
-      );
-      _log.log('created first frame');
-    }
+    // We populate the first 12 frames; this ~roughly corresponds to the number
+    // of visible stack frames.
+    const initialFrameRequestCount = 12;
```
kenzieschmoll, line 636: is 12 enough if devtools is opened on a large screen (4k monitor for example)?

kenzieschmoll:
One edge case to consider: if we populate the first 12 frames, then scroll down in the list immediately (before the request for the full stack finishes), does the scroll offset of the call stack list go back to 0?

packages/devtools_app/lib/src/debugger/debugger_controller.dart
jacob314: If we want to make this smarter you can add a LayoutBuilder to the debugger UI and send a hint to the controller with the # of frames that will fit in the area each time the layout changes.

devoncarew: > is 12 enough if devtools is opened on a large screen (4k monitor for example)? what is the added cost to make this larger than 12?

It probably wouldn't be large enough; the user would see some blank space at the bottom of the call stack which would fill in after the rest of the frames were available - they'd see a flash of content filling in at the bottom.

We could arbitrarily make it larger - populating a few more frames would only slow things down proportionally. We had used 10 here before. Any fixed value would be somewhat arbitrary. 12? 20?

> If we want to make this smarter you can add a LayoutBuilder to the debugger UI and send a hint to the controller with the # of frames that will fit in the area each time the layout changes.

This sounds reasonable, though we'd have to measure the height of a row here. I'd like to avoid more complexity here however. I think something like https://github.com/dart-lang/sdk/issues/43827 would be a good end state. We'd be able to just request all the frames when paused w/o having to limit to the first n frames.

packages/devtools_app/lib/src/debugger/debugger_controller.dart
devoncarew: If we rate limit the requests here, we wouldn't be sending the request for the latest (current) set of frames? I'm trying to think of how we'd actually use this.

jacob314: Here is how you do what you want with `RateLimiter`. It is more of a scheduler than a true RateLimiter as it doesn't ignore requests, it just ensures that the rate limit is respected.
We should probably add a wrapper class on top of RateLimiter that adds the extra logic to toss results that are known to be stale as a newer result was requested.
```dart
int nextTimeStampRequestId = 0;
myRateLimiter = RateLimiter(1.0, () async {
  final requestId = nextTimeStampRequestId++;
  var stackTrace = await fetchStackTrace();
  if (requestId != nextTimeStampRequestId) {
    // Stale request... ignore. Also ignore if we are no longer paused, etc.
    return;
  }
  // show the StackTrace in the ui.
  stackTraceValueNotifier.value = stackTrace;
}

call
stackTrace. scheduleRequest() any time you need a full stacktrace for a new frame.  The rate limiter will ensure you don't have more than one simultaneous request and you don't violate the rate limiters frame rate.
```

packages/devtools_app/lib/src/debugger/debugger_controller.dart
jacob314: Agreed it doesn't make sense to implement as dart-lang/sdk#43827  should be cheap to implement and a better solution.

devoncarew:
> One edge case to consider: if we populate the first 12 frames, then scroll down in the list immediately (before the request for the full stack finishes), does the scroll offset of the call stack list go back to 0?

I think it might. Note that we had a similar problem before this PR - we call reset the currently selected stack frame when we get a new set of frame info - either the first 12 frames, or later, the entire list.

packages/devtools_app/lib/src/debugger/debugger_controller.dart
devoncarew: I think I'd prefer to land the PR as is, w/o adding additional code to rate limit requests, and instead work on making populating frame info more lazy. Still, happy to add some rate limiting here if you think it's net useful.

grouma: Note that in `package:dwds` we ensure the expensive work is only done once so rate limiting is not super necessary for the web workflow.

kenzieschmoll:
lgtm_army

packages/devtools_app/lib/src/debugger/debugger_controller.dart
jacob314: do you cancel pending requests in dwds if they become obsolete (e.g. stack frame changed)? that would mean this doesn't need to be rate limited.

jacob314 (approved):
lgtm_dog

packages/devtools_app/lib/src/debugger/debugger_controller.dart
grouma: We do not cancel requests. That could present a problem.

devoncarew: I think we should just move to requesting all frames, and requesting them with a parameter so the variables aren't populated. There can then be another method (getObject(), getStack(), ...), to request more info for a specific frame. That should make everything fast and keep various implementations ~simple.

Issue #2598, opened 2021-01-12
Resizing Flutter Inspector causes CanvasKit crash


johnpryan:
I often see a dialog saying that CanvasKit has encountered errors. This usually happens when I resize the panel.

<img width="823" alt="Screen Shot 2021-01-12 at 10 49 01 AM" src="https://user-images.githubusercontent.com/1145719/104358862-dc852000-54c3-11eb-8034-a33559e764cd.png">

Related to https://github.com/flutter/flutter-intellij/issues/5169 and https://github.com/flutter/flutter/issues/71668

PR #2713, opened 2021-02-18
Delete code to fallback to the dart:html version of the app.

jacob314:
Leave the index_fallback.html as an escape hatch to debug if users hit issues or IDEs need to fallback.


jacob314:
Fixes https://github.com/flutter/devtools/issues/2598 unless we are really crashing canvaskit.

packages/devtools_app/lib/main.dart
```diff
@@ -9,50 +9,10 @@ import 'src/analytics/stub_provider.dart'
 import 'src/app.dart';
 import 'src/config_specific/framework_initialize/framework_initialize.dart';
 import 'src/config_specific/ide_theme/ide_theme.dart';
-import 'src/config_specific/load_fallback_app/load_fallback_app.dart';
 import 'src/debugger/syntax_highlighter.dart';
 import 'src/preferences.dart';
 
 void main() async {
-  final defaultOnError = FlutterError.onError;
```
kenzieschmoll, line 17: Can you please also submit a CL with this change to the internal main.dart file?

kenzieschmoll (approved):

packages/devtools_app/lib/main.dart
jacob314: good news is as the files it depends on were deleted you should get a compile error as part of the next roll. I can do the next roll into internal and fix as part of that.

PR #2717, opened 2021-02-18
Remove inspector error indicators + render error message inline

DanTup:
(Sorry for the bad GIF framerate 😬)

Feb-18-2021 18-33-54


jacob314:
Fyi @cobblest.

packages/devtools_app/lib/src/inspector/diagnostics.dart
```diff
@@ -209,14 +215,37 @@ class DiagnosticsNodeDescription extends StatelessWidget {
             textStyle.merge(inspector_text_styles.regularBold(colorScheme));
       }
 
-      children.add(Flexible(
+      children.add(Expanded(
```
jacob314, line 218: This should be Flexible. You should be able to instead tweak the properties on the Row widget to get the effect you want instead of hacking it by setting Flex to 2.

packages/devtools_app/lib/src/inspector/diagnostics.dart
```diff
@@ -209,14 +215,37 @@ class DiagnosticsNodeDescription extends StatelessWidget {
...
           isProperty: false,
         ),
       ));
+
+      if (errorText != null) {
+        children.add(
+          Expanded(
+            // Errors tend to be longer than widget classes, so allow them
+            // to take up a larger proportion of the space.
+            flex: 2,
```
jacob314, line 232: Would be better to let the Row layout itself know that the Description is flexible so does not need to use the full width it is offered.

packages/devtools_app/lib/src/inspector/diagnostics.dart
```diff
@@ -209,14 +215,37 @@ class DiagnosticsNodeDescription extends StatelessWidget {
...
+            child: Align(
```
jacob314, line 233: this use of align isn't idealYou should instead be using the flex layout itself which has properties for aligning children. That may fix the issues you were seeing that made using Flexible not work.

cobblest:
> Fyi @cobblest.

LGTM!

packages/devtools_app/lib/src/inspector/diagnostics.dart
DanTup: Which properties might help? I tried a bunch of things I could think of, but struggled to get anything that behaved better than this.

Setting `mainAxisAlignment` to `spaceBetween` or `end` resulted in the main labels also being moved around:

Screenshot 2021-02-19 at 12 29 40

Setting both children to `Flexible` and not setting `flex` at all results in this space at the end of the row:

Screenshot 2021-02-19 at 12 32 42

Setting `fit: FlexFit.tight,` on the description kinda works, but results in the description always being as wide as the error, resulting in a lot of unnecessary truncation:

Screenshot 2021-02-19 at 12 34 37

I don't fully entirely why the error text is truncating when there's enough space for it. I would've expected truncation only when all items can't be fit in naturally.

kenzieschmoll: did you try changing this `Expanded` the `Flexible` and removing the `Align` widget below?

DanTup: Yeah, that was the middle screenshot (sorry, didn't mention I removed `Align`). Here's what the code looked like:

```dart
      children.add(Flexible(
        child: buildDescription(
          diagnostic.description,
          textStyle,
          colorScheme,
          isProperty: false,
        ),
      ));

      if (errorText != null) {
        children.add(
          Flexible(
            child: Text(
              errorText,
              textAlign: TextAlign.right,
              overflow: TextOverflow.ellipsis,
              // When the node is selected, the backgorund will be an error
              // color so don't render the text the same color.
              style: isSelected
                  ? inspector_text_styles.regular
                  : inspector_text_styles.error(colorScheme),
            ),
          ),
        );
      }
    }

    return Row(mainAxisSize: MainAxisSize.min, children: children);
```

It leaves a large space on the right. I can see that space in the layout explorer, though I can't figure out how to get rid of it 😞

kenzieschmoll: I forked your branch and did some experimenting. The trick was adding another row so that we could use `MainAxisAlignment.spaceBetween` to align the widget description and the error. The following configuration (starting replacement at this line) should give you what you need:

```
      final diagnosticDescription = buildDescription(
        diagnostic.description,
        textStyle,
        colorScheme,
        isProperty: false,
      );

      if (errorText != null) {
        children.add(
          Expanded(
            child: Row(
              mainAxisAlignment: MainAxisAlignment.spaceBetween,
              children: [
                diagnosticDescription,
                Flexible(
                  child: RichText(
                    textAlign: TextAlign.right,
                    overflow: TextOverflow.ellipsis,
                    text: TextSpan(
                      text: errorText,
                      // When the node is selected, the background will be an error
                      // color so don't render the text the same color.
                      style: isSelected
                          ? inspector_text_styles.regular
                          : inspector_text_styles.error(colorScheme),
                    ),
                  ),
                ),
              ],
            ),
          ),
        );
      } else {
        children.add(Expanded(
          child: diagnosticDescription,
        ));
      }
```

jacob314: Great job finding a way to make it work! I suspect you can solve it with 1 row instead of two nested rows by using mainAxisSize on the row.
Fyi @InMatrix, @cobblest  this would be a great flex layout usability study test case.

kenzieschmoll: I thought so too until I started to play with the code. Tweaking mainAxisSize did not fix the problem. The row size was fine, it was the size of the children that were off. Wrapping the error text in Expanded did not extend that text widget to the end of the row as we would expect. There was a large gap of space to the right. It seemed it had something to do with having multiple flexible widgets in the row. This is a strange case, everything I thought would work to fix the flex layout did not work as I expected it would.

jacob314: The big problem here is that the solution is 2 code changes to the Flex widget away from the original code so no single change feels like it makes progress.

DanTup: Aha! Thanks! I wasn't sure if nesting a Row was a reasonable thing to do. I've updated as above (and flattened a bit, inc. moving the error label to a function).

DanTup:
@jacob314 @kenzieschmoll I think this may be good now :-)

packages/devtools_app/lib/src/inspector/diagnostics.dart
DanTup: Apparently I missed some comments here..

@jacob314 is the current code alright, or do you think it should still be possible with 1 row?

kenzieschmoll: If you want to go ahead and land this, I'm fine with adding a TODO to try to simplify the widget structure. This does feel like an odd case that needs a little further investigation, but I don't thinks its submit-blocking.

jacob314:
lgtm
Lets land now and offline investigate whether there is a way to solve this without nesting flex widgets.

PR #2719, opened 2021-02-19
Support latest VMService version.

jacob314:


jacob314:
@bkonyi just pushed a new vm service version and we are broken without this change on pub upgrade.

jacob314:
TBR

packages/devtools_app/lib/src/vm_service_wrapper.dart
```diff
@@ -900,6 +878,24 @@ class VmServiceWrapper implements VmService {
       trackFuture('getRetainingPath',
           _vmService.getRetainingPath(isolateId, targetId, limit));
 
+  @override
+  Future<CpuSamples> getAllocationTraces(String isolateId,
```
terrylucas, line 882: The old API calls to getAllocationSamples needs to be renamed to use getAllocationTraces and the old entry removed.  I'll create another PR to do this.

terrylucas (commented):
LGTM, thanks.

packages/devtools_app/lib/src/vm_service_wrapper.dart
kenzieschmoll: commented on https://github.com/flutter/devtools/pull/2720/files#r579327782 - we probably need to call getAllocationSamples here for older versions of the protocol

PR #2720, opened 2021-02-19
Use new VM API for allocation trace.

terrylucas:


terrylucas:
TBR

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -620,9 +620,9 @@ class MemoryController extends DisposableController
   }
 
   /// Track where/when a particular class is allocated (constructor new'd).
-  Future<CpuSamples> getAllocationSamples(ClassRef ref) async {
+  Future<CpuSamples> getAllocationTraces(ClassRef ref) async {
     if (!await isIsolateLive(_isolateId)) return null;
-    final returnObject = await serviceManager.service.getAllocationSamples(
+    final returnObject = await serviceManager.service.getAllocationTraces(
```
kenzieschmoll, line 625: We need to add a wrapper method for this in vm_service_wrapper so that we can gate the version on when this RPC became available and so that we can track the future

kenzieschmoll: Oh I just saw https://github.com/flutter/devtools/pull/2719 so adding the wrapper method is done. However, we should still be gating the service version. Probably using getAllocationSamples for older vm service versions and using getAllocationTraces for newer ones.

bkonyi: Yes, this should definitely be conditional on the service version. The private RPC for older versions, the public for newer ones.

terrylucas: DevTools never used it before (never published) so it doesn't matter from DevTools.

jacob314: I agree we don't need to be backwards compatible but we need to make sure the memory page doesn't fail in confusing ways if the VM of the app being debugged is old. I would be fine with the memory page having very strict version constraints for all the new functionality as we are trying to iterate on it with users on tip of trunk Flutter before we worry about other users.

PR #2728, opened 2021-02-24
Fixed snapshot semantics.

terrylucas:
- Fixed supporting both snapshot versions (older is 1-based class index and VM Service protocol version 3.43.0  is 0-based index).
- Fixed Null type to not display as ObjectReference (expandable).
- Fixed predefined were not correctly showing as scalar. 
- Fixed duplicating entries while expanding and collapsing.
- Added a vertical splitter between snapshot tree table and the right-side inspector/monitor/analysis view.


jacob314 (approved):
lgtm

PR #2732, opened 2021-02-24
Fix lifecycle management issues with disconnecting and reconnecting to apps

kenzieschmoll:


packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -1191,3 +1201,11 @@ class VmServiceCapabilities {
   bool get supportsGetScripts =>
       version.major > 3 || (version.major == 3 && version.minor >= 12);
 }
+
+class ConnectedState {
+  const ConnectedState(this.connected, {this.manual = false});
+
+  final bool connected;
+
+  final bool manual;
```
jacob314, line 1210: nit: manual isn't the mount obvious name.  Add a comment or maybe change it to 
`intentionalDisconnect` or something ugly but clear.

packages/devtools_app/lib/src/device_dialog.dart
```diff
@@ -31,6 +31,8 @@ class DeviceDialog extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    if (!serviceManager.connectedState.value.connected) return const SizedBox();
```
jacob314, line 34: nit: are we really showing the device dialog when not connected?
Consider instead just checking whether serviceManager.vm is null rather than the connected state so it is clear you are just avoiding showing content when that content isn't available.

jacob314:
lgtm
This is great!

packages/devtools_app/lib/src/device_dialog.dart
kenzieschmoll: strangely enough, yes the dialog build method gets hit one last time when we disconnect the app. I spend a while trying to debug, but ended up just adding this check to be safe.

PR #2734, opened 2021-02-25
Fix Memory panel hover overlay leak.

uppet:
switching out of Memory panel when hover overlay is visible will trigger UI leak.

override dispose() to call hideHover() in time.


uppet:
The test_dart2js  failed test seems to be a false negative.  Could someone please rerun the bot check?

terrylucas:
Thanks Joyer.  I'll re-run the bots should pass sometimes this error pops up erratically.

terrylucas (approved):

PR #2736, opened 2021-02-25
Fix another lifecycle issue

kenzieschmoll:
This covers the case where you manually change the URL when DevTools is run on flutter web. Whenever we navigate to the landing screen, we should disconnect from any connected app.


packages/devtools_app/lib/src/app.dart
```diff
@@ -162,6 +162,8 @@ class DevToolsAppState extends State<DevToolsApp> {
 
     // Always return the landing screen if there's no VM service URI.
     if (vmServiceUri?.isEmpty ?? true) {
+      // Ensure we disconnect from any previously connected apps.
+      serviceManager.manuallyDisconnect();
```
jacob314, line 166: building a widget shouldn't have side effects like this. Is there a different fix?

kenzieschmoll: Moved disconnect calls into router. Fixing tests, then will land so I can get this into the dev release.

packages/devtools_app/lib/src/scaffold.dart
```diff
@@ -327,6 +327,10 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
                   children: tabBodies,
                 ),
                 if (serviceManager.hasConnection &&
+                    DevToolsRouterDelegate.of(context)
```
jacob314, line 330: Add a TODO to clean this up. Seems odd we have to bug the router for this.

jacob314 (approved):
lgtm

kenzieschmoll:
Landing on red. Test failure is from memory screenshot tests failing with 0.00% diff...

PR #2739, opened 2021-02-25
Move codicon.ttf file as part of publish script to include it in build.

kenzieschmoll:
Verified this works. published at 2.0.0-dev.2

### DIFF
--- a/tool/publish.sh
+++ b/tool/publish.sh
@@ -11,6 +11,9 @@ pushd packages/devtools_app
 flutter pub get
 popd
 
+cp third_party/packages/codicon/lib/font/codicon.ttf packages/devtools_app/fonts/codicon.ttf
+perl -pi -e 's/packages\/codicon\/font\/codicon.ttf/fonts\/codicon.ttf/g' packages/devtools_app/pubspec.yaml
+
 tool/build_release.sh
 
 pushd packages/devtools
--- a/packages/devtools/pubspec.yaml
+++ b/packages/devtools/pubspec.yaml
@@ -8,7 +8,7 @@ description: A suite of web-based performance tooling for Dart and Flutter.
 # package remaining purely as a container for the compiled copy of the devtools
 # app. That ensures that version constraints for the devtools_app do not impact
 # this package.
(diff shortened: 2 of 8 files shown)


tool/publish.sh
```diff
@@ -11,6 +11,9 @@ pushd packages/devtools_app
 flutter pub get
 popd
 
+cp third_party/packages/codicon/lib/font/codicon.ttf packages/devtools_app/fonts/codicon.ttf
+perl -pi -e 's/packages\/codicon\/font\/codicon.ttf/fonts\/codicon.ttf/g' packages/devtools_app/pubspec.yaml
```
devoncarew, line 15: This will re-write the pubspec file in place but will leave behind a modified file (it doesn't convert it back after the publish).

Do we want to copy the font file to the `packages/devtools_app/fonts` as part of some other step (like a packages get step or some other setup) and just always reference it from there in the pubspec?

devoncarew (approved):
One comment; not a blocker for landing since it would be better to have a valid reference to the font file, but we should think if there are any possible improvements.

tool/publish.sh
kenzieschmoll: We do other in place edits below in this publish script, also leaving behind modified files (pubspecs, gitignore). 

> Do we want to copy the font file to the packages/devtools_app/fonts as part of some other step (like a packages get step or some other setup) and just always reference it from there in the pubspec?

Not sure I understand this suggestion. What would a packages get step be? and what would be different about the pubspec edit to devtools_app/pubspec.yaml?

devoncarew: > What would a packages get step be? and what would be different about the pubspec edit to devtools_app/pubspec.yaml?

I meant as something like a `./bin/repo_tool pub-get` step or similar - something that would need to be run for developers to init the repo locally. If the pubspec file always pointed to fonts/codicon.ttf we wouldn't have to make as edit during the release script.

> We do other in place edits below in this publish script, also leaving behind modified files (pubspecs, gitignore).

Ah, ok, I was unaware.

PR #2740, opened 2021-02-26
Memory page UX polish

kenzieschmoll:
- allocation icons are now the same size as other icons (refactored the MaterialIconLabel utility for this)
- normalize padding around buttons and tables
- replace expand all / collapse all icon buttons with the Expand All / Collapse All buttons used elsewhere in DevTools
- selecting a row in the accumulator table toggles the checkbox for that row (@terrylucas is there another use for selecting a row? I found myself expecting the checkbox would change when I selected a row)
- general cleanup

### DIFF
--- a/packages/devtools_app/lib/src/common_widgets.dart
+++ b/packages/devtools_app/lib/src/common_widgets.dart
@@ -105,8 +105,8 @@ class IconLabelButton extends StatelessWidget {
     return FixedHeightOutlinedButton(
       onPressed: onPressed,
       child: MaterialIconLabel(
-        icon,
-        label,
+        label: label,
+        iconData: icon,
         includeTextWidth: includeTextWidth,
       ),
     );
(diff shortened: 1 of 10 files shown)


packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -280,6 +280,35 @@ class HelpButton extends StatelessWidget {
   }
 }
 
+class ExpandAllButton extends StatelessWidget {
+  const ExpandAllButton({Key key, @required this.onPressed}) : super(key: key);
+
+  final VoidCallback onPressed;
+
+  @override
+  Widget build(BuildContext context) {
```
terrylucas, line 289: nit: =>

Not sure what we like to do here.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -430,9 +439,10 @@ class HeapTreeViewState extends State<HeapTree>
                 Column(
                   children: [
                     _buildAllocationsControls(themeData),
-                    Expanded(
+                    const SizedBox(height: denseRowSpacing),
```
terrylucas, line 442: Shouldn't the analyzer/linter have mentioned that a const was needed?  Surprised didn't see that in latest.

terrylucas (approved):
LGTM, of course update-goldens will be needed.

packages/devtools_app/lib/src/common_widgets.dart
kenzieschmoll: if it's more than a one line return, we should use brackets.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
kenzieschmoll: this is a const? When I don't use a const where I could be, I get an analyzer warning and a squiggley in intellij.

terrylucas: I used VSCode.  Should be the same?

kenzieschmoll: Are you saying you got an analysis error for the code I wrote? what was the analysis error?

PR #2745, opened 2021-02-26
Flame chart scrolling polish

kenzieschmoll:
This makes it so you cannot scroll beyond the bounds of the scroll view. Without this, there is a distracting bounce back effect.


jacob314 (approved):
lgtm
Would be great to apply this to the Table views as well. The bounce is annoying on the logging view.

kenzieschmoll:
AFAICT there is not bounce on the table views? I just tried to reproduce with the logging table and I can't get it to bounce

PR #2746, opened 2021-02-26
Evaluation HoverCard

grouma:
- Create new `HoverCard` class to handle absolutely positioned overlays that automatically remove themself
- Update `CodeView` to support hovering over source code and display an evaluation result

Screen Shot 2021-02-26 at 10 51 19 AM


kenzieschmoll:
DBC: Is this the card placement we want? I would think we'd want the card to be centered above the mouse position or below. I think Chrome devtools places the hover card above. IntelliJ places it below like this: 
Screen Shot 2021-02-26 at 12 54 18 PM

grouma:
> DBC: Is this the card placement we want? I would think we'd want the card to be centered above the mouse position or below. I think Chrome devtools places the hover card above. IntelliJ places it below like this:

I'm open to changing it. I was making this consistent with the current hover logic in the memory page. Let me know if you prefer consistency or the Chrome logic.

kenzieschmoll:
>>DBC: Is this the card placement we want? I would think we'd want the card to be centered above the mouse position or below. I think Chrome devtools places the hover card above. IntelliJ places it below like this:

> I'm open to changing it. I was making this consistent with the current hover logic in the memory page. Let me know if you prefer consistency or the Chrome logic.

Yes let's be consistent with chrome devtools or with the IDEs for debugging purposes.

grouma:
> Yes let's be consistent with chrome devtools or with the IDEs for debugging purposes.

Done.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -39,10 +43,14 @@ class CodeView extends StatefulWidget {
...
 
   @override
-  _CodeViewState createState() => _CodeViewState();
+  _CodeViewState createState() => _CodeViewState(controller);
 }
 
 class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
+  _CodeViewState(this.debuggerController);
+
+  final DebuggerController debuggerController;
```
kenzieschmoll, line 52: access controller via widget.controller instead of passing it into the state object

kenzieschmoll: And actually, since CodeViewState is a stateful widget, we can access the controller from Provider in didChangeDependencies. See https://github.com/flutter/devtools/blob/4456f8124ff2a553f0207ed3010ab663b7c23b16/packages/devtools_app/lib/src/debugger/debugger_screen.dart#L91 for an example of that

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -538,31 +549,96 @@ class Lines extends StatelessWidget {
...
     );
   }
 }
 
-class LineItem extends StatelessWidget {
+class LineItem extends StatefulWidget {
   const LineItem({
     Key key,
     @required this.lineContents,
+    @required this.debuggerController,
```
kenzieschmoll, line 564: access debugger controller from didChangeDependencies instead of requiring it to be a parameter for this stateful widget (see above comment for example)

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -538,31 +549,96 @@ class Lines extends StatelessWidget {
...
     this.pausedFrame,
   }) : super(key: key);
 
+  static const _hoverDelay = Duration(milliseconds: 500);
```
kenzieschmoll, line 568: we already have a const for this in common_widgets (tooltipWait)

kenzieschmoll: actually, seeing how this is used, using tooltipWait may not be as clear. fine to leave.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -538,31 +549,96 @@ class Lines extends StatelessWidget {
...
+
+  void _onHover(PointerHoverEvent event, BuildContext context) {
+    _showTimer?.cancel();
+    _removeTimer?.cancel();
+    if (!widget.debuggerController.isPaused.value) return;
+    _showTimer = Timer(LineItem._hoverDelay, () async {
+      final theme = Theme.of(context);
+      _hoverCard?.remove();
+      final word = wordForHover(
+          event.localPosition.dx, widget.lineContents, theme.fixedFontStyle);
```
kenzieschmoll, line 603: nit: trailing comma

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -538,31 +549,96 @@ class Lines extends StatelessWidget {
...
+        try {
+          final response =
+              await widget.debuggerController.evalAtCurrentFrame(word);
+          final variable = Variable.fromRef(response);
+          await widget.debuggerController.buildVariablesTree(variable);
+          _hoverCard = HoverCard(
+              contents: Material(
+                  child: ExpandableVariable(
+                      debuggerController: widget.debuggerController,
+                      variable: ValueNotifier(variable))),
```
kenzieschmoll, line 614: trailing commas

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -538,31 +549,96 @@ class Lines extends StatelessWidget {
...
+              event: event,
+              width: 250,
```
kenzieschmoll, line 616: put this into a const

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -538,31 +549,96 @@ class Lines extends StatelessWidget {
...
+              title: word,
+              context: context);
```
kenzieschmoll, line 618: trailing comma

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -590,8 +666,8 @@ class LineItem extends StatelessWidget {
               // functionality exists to render the selection handles properly.
               Opacity(
                 opacity: 0,
-                child:
-                    RichText(text: truncateTextSpan(lineContents, column - 1)),
+                child: RichText(
+                    text: truncateTextSpan(widget.lineContents, column - 1)),
```
kenzieschmoll, line 670: trailing comma

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -606,22 +682,28 @@ class LineItem extends StatelessWidget {
...
-            maxLines: 1,
-          ),
+          MouseRegion(
+              onExit: (_) => _onHoverExit(),
+              onHover: (e) => _onHover(e, context),
+              child: SelectableText.rich(
+                widget.lineContents,
+                scrollPhysics: const NeverScrollableScrollPhysics(),
+                maxLines: 1,
+              )),
```
kenzieschmoll, line 692: trailing comma

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -606,22 +682,28 @@ class LineItem extends StatelessWidget {
...
-        maxLines: 1,
-      );
+      child = MouseRegion(
+          onExit: (_) => _onHoverExit(),
+          onHover: (e) => _onHover(e, context),
+          child: SelectableText.rich(
+            widget.lineContents,
+            scrollPhysics: const NeverScrollableScrollPhysics(),
+            maxLines: 1,
+          ));
```
kenzieschmoll, line 703: tc

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -606,22 +682,28 @@ class LineItem extends StatelessWidget {
...
+              )),
         ],
       );
     } else {
-      child = SelectableText.rich(
-        lineContents,
-        scrollPhysics: const NeverScrollableScrollPhysics(),
-        maxLines: 1,
-      );
+      child = MouseRegion(
```
kenzieschmoll, line 696: this is duplicated code. Can you pull out into a local var and use in both places?

packages/devtools_app/lib/src/debugger/variables.dart
```diff
@@ -72,16 +72,19 @@ class VariablesList extends StatelessWidget {
...
+  final DebuggerController debuggerController;
 
   @override
   Widget build(BuildContext context) {
     return ValueListenableBuilder<Variable>(
       valueListenable: variable,
       builder: (context, variable, _) {
-        final controller = Provider.of<DebuggerController>(context);
+        final controller =
+            debuggerController ?? Provider.of<DebuggerController>(context);
```
kenzieschmoll, line 87: Is this because we don't have a debugger controller available in the context of the hover card overlay?

packages/devtools_app/lib/src/debugger/hover.dart
```diff
@@ -0,0 +1,153 @@
...
+  if (index == -1) return word;
+  word = line.children[index].toPlainText();
+  word = _mergedLinkedWords(word, index, line);
+  return word;
+}
+
+/// Merges words linked by a `.`.
+///
+/// For example, hovering over `bar` in `foo.bar.baz` would return
+/// `foo.bar.baz`.
```
kenzieschmoll, line 25: nit: should this be foo.bar instead of foo.bar.baz?

packages/devtools_app/lib/src/debugger/hover.dart
```diff
@@ -0,0 +1,153 @@
...
+String _mergedLinkedWords(String word, int index, TextSpan line) {
+  var left = index - 1;
```
kenzieschmoll, line 27: how is the index = 1 case handled? If we are hovering over bar in foo.bar, don't we want to evaluate foo.bar? This looks like we only merge the words if there are two previous values. We should be able to change this to chain previous values one at a time instead of 2 at a time.

packages/devtools_app/lib/src/debugger/hover.dart
```diff
@@ -0,0 +1,153 @@
...
+          top: position.dy + _hoverYOffset,
+          child: MouseRegion(
+              onExit: (_) {
+                remove();
+              },
+              onEnter: (_) {
+                _hasMouseEntered = true;
+              },
+              child: Container(
+                padding: const EdgeInsets.fromLTRB(8, 8, 8, 8),
```
kenzieschmoll, line 105: use `const EdgeInsets.all(denseSpacing);`

packages/devtools_app/lib/src/debugger/hover.dart
```diff
@@ -0,0 +1,153 @@
...
+                decoration: BoxDecoration(
+                  color: colorScheme.defaultBackgroundColor,
+                  border: Border.all(
+                    color: focusColor,
+                    width: _hoverCardBorderWidth,
+                  ),
+                  borderRadius: BorderRadius.circular(10.0),
```
kenzieschmoll, line 112: put this into a const or use the `defaultBorderRadius` used elsewhere in devtools

packages/devtools_app/lib/src/debugger/hover.dart
```diff
@@ -0,0 +1,153 @@
...
+                ),
+                width: width,
+                child: Column(
+                  crossAxisAlignment: CrossAxisAlignment.start,
+                  children: [
+                    Container(
```
kenzieschmoll, line 118: use SizedBox instead since width is the only property

packages/devtools_app/lib/src/debugger/hover.dart
```diff
@@ -0,0 +1,153 @@
...
+                      width: width,
+                      child: Text(
+                        title,
```
kenzieschmoll, line 121: what if the title is long? should this ellipsize?

packages/devtools_app/lib/src/debugger/hover.dart
```diff
@@ -0,0 +1,153 @@
...
+                        style: hoverHeading,
+                        textAlign: TextAlign.center,
+                      ),
+                    ),
+                    Divider(color: colorScheme.hoverTextStyle.color),
+                    contents,
+                  ],
+                ),
+              )));
```
kenzieschmoll, line 130: trailing commas

packages/devtools_app/lib/src/debugger/hover.dart
```diff
@@ -0,0 +1,153 @@
...
+    });
+    overlayState.insert(_overlayEntry);
+  }
+
+  OverlayEntry _overlayEntry;
+  bool _isRemoved = false;
+  bool _hasMouseEntered = false;
```
kenzieschmoll, line 137: new line between class members

kenzieschmoll (commented):
Done with first round of comments

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -39,10 +43,14 @@ class CodeView extends StatefulWidget {
   final void Function(ScriptRef scriptRef, int line) onSelected;
 
   @override
-  _CodeViewState createState() => _CodeViewState();
+  _CodeViewState createState() => _CodeViewState(controller);
```
jacob314, line 46: https://dart-lang.github.io/linter/lints/no_logic_in_create_state.html
will avoid this issue without code review. lets add the lint to devtools.

packages/devtools_app/lib/src/debugger/codeview.dart
jacob314: why not just lookup the Provider in the build method?

packages/devtools_app/lib/src/debugger/codeview.dart
jacob314: why didUpdateDependencies instead of the build method?

packages/devtools_app/lib/src/debugger/variables.dart
jacob314: correct. Overlay doesn't play great with provider. We might be able to resolve if we moved our providers to the absolute root of the tree but that would cause problems for the snapshot app case.

packages/devtools_app/lib/src/debugger/hover.dart
```diff
@@ -0,0 +1,153 @@
...
+///
+/// For example, hovering over `bar` in `foo.bar.baz` would return
+/// `foo.bar.baz`.
+String _mergedLinkedWords(String word, int index, TextSpan line) {
+  var left = index - 1;
+  while (left > 1) {
+    final prev = line.children[left].toPlainText();
+    final prevprev = line.children[left - 1].toPlainText();
+    if (prev == '.') {
+      word = '$prevprev$prev$word';
```
jacob314, line 32: nit: one way to write this logic a little bit more simply is to create a
List<String> that is the components of the line.
Then you can just track the the start and end indexes and use String.join to get the output string when you are done.

packages/devtools_app/lib/src/debugger/hover.dart
grouma: No thi sis correct.

packages/devtools_app/lib/src/debugger/hover.dart
grouma: We want to evaluate all words linked by a `.`. This is consistent with Chrome's behavior.

index in this case would point to `bar` and `foo` would be merged. There are test cases covering this.

packages/devtools_app/lib/src/debugger/hover.dart
kenzieschmoll: can you explain that rationale to me? If a user hovers over `bar`, wouldn't they want to look at the value of `bar`, not `baz`?

grouma: I thought Chrome's behavior was to treat all things linked by a `.` as a single "block".

I just tested and that doesn't seem to be the case:
Screen Shot 2021-03-02 at 3 16 52 PM

I'll update the logic.

grouma:
This should be ready to review again. PTAL.

packages/devtools_app/lib/src/debugger/hover.dart
grouma: I updated the logic and the tests. If we hover over `bar` in `foo.bar.baz` we want to evaluate `foo.bar` to be consistent with Chrome. There will always be at least two items because one of the items will be `.`. Otherwise it won't be linked.

kenzieschmoll:
lgtm_orangatang

kenzieschmoll (approved):

PR #2748, opened 2021-02-26
Add support for serving a custom DevTools build

bkonyi:
The Dart SDK will use a pinned version pre-built version of DevTools. This means that DDS requires a way to provide a path to the pinned build to devtools_server in order to correctly serve the application.

### DIFF
--- a/packages/devtools_server/lib/src/external_handlers.dart
+++ b/packages/devtools_server/lib/src/external_handlers.dart
@@ -27,18 +27,23 @@ import 'server_api.dart';
 /// DevTools project.
 Future<shelf.Handler> defaultHandler(
   ClientManager clients, {
+  String customDevToolsPath,
   bool debugMode = false,
 }) async {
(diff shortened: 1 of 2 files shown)


jacob314:
will there be possible version skew between the devtools_server and devtools_app in this mode. That should be fine but we should document that you cannot depend on the app and server to be the exact same version anymore if that is the plan.

jacob314:
lgtm

bkonyi:
> will there be possible version skew between the devtools_server and devtools_app in this mode. That should be fine but we should document that you cannot depend on the app and server to be the exact same version anymore if that is the plan.

Nope! The SDK will pull in `devtools_server` and `devtools_shared` alongside the `devtools_app` build output, all from the same revision (see the [CIPD update script](https://github.com/dart-lang/sdk/blob/master/third_party/devtools/update.sh) in `$DART_SDK/third_party/devtools/`).

PR #2751, opened 2021-02-28
Bubble and reset

terrylucas:
- Fixed event pane symbol displayed when a monitor allocation reset occurs.
- Fixed symbols for light mode as well.

- Monitor track displays (light and dark mode):

  image
image

- Monitor reset displays (light and dark mode):

  image
image

- Added a animated bubble that appears to the left of the timestamp displayed when an allocation monitor happens.  On initial change detection the bubble grows like a balloon then settles back to a bubble.  After 5 seconds then the bubble will disappear.

   image


packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -640,7 +649,78 @@ class HeapTreeViewState extends State<HeapTree>
...
+      _animation.reverse();
+    }
+  }
+
+  static const _updateCircleRadius = 8.0;
+
+  Timer removeUpdateBubble;
+
+  Widget displayTimestampUpdateBubble() {
+    final themeData = Theme.of(context);
```
jacob314, line 679: nit: load theme data right when it is needed rather from this widget's context.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -669,16 +749,60 @@ class HeapTreeViewState extends State<HeapTree>
...
+          ),
         ),
       ],
     );
   }
 
+  double textWidgetWidth(String message, {TextStyle style}) {
+    // self-defined constraint
+    const constraints = BoxConstraints(
+      maxWidth: 800.0, // maxwidth calculated
```
jacob314, line 792: comment seems incomplete. where is this calculated from?

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -669,16 +749,60 @@ class HeapTreeViewState extends State<HeapTree>
...
+    );
+    final richTextWidget = Text.rich(TextSpan(text: message), style: style)
+        .build(context) as RichText;
+    final renderObject = richTextWidget.createRenderObject(context);
```
jacob314, line 796: @grouma do you have some pointers to a better way to perform text measurement. Running the widget build pipeline is clever but a bit heavy weight. 
This is fine to leave for now but add a TODO.  This might cause some subtle bugs. For example, makes me nervous that we are passing our context to the createRenderObject call.

jacob314 (commented):
what is going on with the empty file goldens? probably remove them from the CL or add tests and generate real goldens.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: Good idea, don't think it changes anything.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: The message is not > 800 pixels in this usage.  However, I did add a TODO and a check to log a warning if this occurs.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: Done. Let me know Gary if there is a lightweight computation?

terrylucas:
> what is going on with the empty file goldens? probably remove them from the CL or add tests and generate real goldens.

terrylucas:
PTAL updated.

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
```diff
@@ -786,11 +784,16 @@ class HeapTreeViewState extends State<HeapTree>
...
   double textWidgetWidth(String message, {TextStyle style}) {
-    // self-defined constraint
+    // Longest message must fit in this width.
     const constraints = BoxConstraints(
-      maxWidth: 800.0, // maxwidth calculated
+      maxWidth: maxWidth,
     );
+
+    // TODO(terry): Is there a better (less heavyweight) way of computing text
+    //              width then using the widget pipeline?
```
jacob314, line 796: nit: then -> than

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/memory/memory_heap_tree_view.dart
terrylucas: Done.

PR #2762, opened 2021-03-05
Add option to load offline file from landing screen

kenzieschmoll:
Screen Shot 2021-03-05 at 9 31 57 AM

This PR also includes some cleanup and bug fixes for offline mode.

### DIFF
--- a/packages/devtools_app/lib/src/file_import.dart
+++ b/packages/devtools_app/lib/src/file_import.dart
@@ -6,12 +6,12 @@ import 'dart:convert';
 import 'package:flutter/material.dart';
 import 'package:file_selector/file_selector.dart';
 
-import '../common_widgets.dart';
-import '../config_specific/drag_and_drop/drag_and_drop.dart';
-import '../notifications.dart';
-import '../theme.dart';
-import '../ui/label.dart';
-import '../utils.dart';
+import 'common_widgets.dart';
+import 'config_specific/drag_and_drop/drag_and_drop.dart';
+import 'notifications.dart';
+import 'theme.dart';
+import 'ui/label.dart';
+import 'utils.dart';
 
 class FileImportContainer extends StatefulWidget {
   const FileImportContainer({
@@ -125,7 +125,7 @@ class _FileImportContainerState extends State<FileImportContainer> {
           ),
         ),
         const SizedBox(width: denseSpacing),
-        _buildImportButton(),
+        FileImportButton(onPressed: _importFile),
         // Horizontal spacer with flex value of 1.
         const Flexible(
           child: SizedBox(height: rowHeight),
@@ -142,14 +142,6 @@ class _FileImportContainerState extends State<FileImportContainer> {
         color: Theme.of(context).textTheme.headline1.color,
       ),
       textAlign: TextAlign.left,
-    );
-  }
-
-  Widget _buildImportButton() {
-    return IconLabelButton(
-      onPressed: _importFile,
-      icon: Icons.file_upload,
-      label: 'Import File',
     );
   }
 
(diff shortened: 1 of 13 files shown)


packages/devtools_app/lib/src/file_import.dart
```diff
@@ -204,6 +190,43 @@ class _FileImportContainerState extends State<FileImportContainer> {
...
+  final acceptedTypeGroups = [XTypeGroup(extensions: acceptedTypes)];
+  final file = await openFile(acceptedTypeGroups: acceptedTypeGroups);
+  final data = jsonDecode(await file.readAsString());
+  final lastModifiedTime = await file.lastModified();
+  final devToolsJsonFile = DevToolsJsonFile(
+    data: data,
+    name: file.name,
+    lastModifiedTime: lastModifiedTime,
+  );
+  importHandler(devToolsJsonFile);
```
jacob314, line 206: we be both async and have a callback? Simpler to just return the json from this method and then call the import handler after awaiting.

packages/devtools_app/lib/src/landing_screen.dart
```diff
@@ -67,6 +73,39 @@ class _LandingScreenBodyState extends State<LandingScreenBody> {
...
+    return _buildSection(
+      title: 'Load DevTools Data',
+      child: Column(
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          _subtitleText(
+              'Import a data file to use DevTools without an app connection.'),
+          const SizedBox(height: denseRowSpacing),
+          _captionText(
+              'At this time, DevTools only supports importing files that were originally'
```
jacob314, line 86: What users is this caption for?

packages/devtools_app/lib/src/performance/performance_model.dart
```diff
@@ -293,14 +293,14 @@ class OfflinePerformanceData extends PerformanceData {
                 .cast<String, dynamic>())
         : null;
 
-    final double displayRefreshRate =
+    final num displayRefreshRate =
```
jacob314, line 296: rather than making this a num prefer to convert the json to a double. generally use num only where really necessary as it is easier to just reason about data consistently being int or double.

packages/devtools_app/lib/src/landing_screen.dart
kenzieschmoll: This is to warn users about the files that we currently support. for example, we don't support loading an export from chrome://tracing or perfetto yet

packages/devtools_app/lib/src/scaffold.dart
```diff
@@ -37,7 +37,7 @@ class DevToolsScaffold extends StatefulWidget {
     Key key,
     @required this.tabs,
     @required this.analyticsProvider,
-    this.title,
+    this.title = 'DevTools for Flutter & Dart',
```
jacob314, line 40: this is a little of an ugly default value. Are multiple different screens using it?

kenzieschmoll: The landing screen was already using it. The snapshot screen (for offline imports) and app size screen had no title, so adding this makes them the same as the landing screen.

packages/devtools_app/lib/src/file_import.dart
kenzieschmoll: done

packages/devtools_app/lib/src/performance/performance_model.dart
kenzieschmoll: done

packages/devtools_app/lib/src/landing_screen.dart
jacob314: Maybe mention those specifically? You could even mention we plan to support them.

jacob314:
lgtm

PR #2706, opened 2021-02-15
Add tab for inspecting the state of Providers 

rrousselGit:

PR #2773, opened 2021-03-06
Add provider_app fixture to devtool_testing

rrousselGit:
This adds a fixture using `provider` in various situations

related to #2706


jacob314 (commented):
lgtm once copyright notices are added.
lgtm

PR #2783, opened 2021-03-10
Show correct mouse cursor for splitters

Timmmm:
This changes the cursor from the grab hand to a proper resize double-arrow. I've tested this change in my app (I stole `split.dart` from here since there isn't an official one) and it works there. I haven't actually built the dev tools.


kenzieschmoll:
LGTM

PR #2784, opened 2021-03-11
Improve file picker UX

grouma:
Mirror the file picker UX in Chrome which shows file results in two lines, the first line is simply the file name, the second line is the entire path potentially truncated. This helps for very long files as it's difficult to distinguish between truncated file names.

Screen Shot 2021-03-10 at 4 10 35 PM

### DIFF
--- a/packages/devtools_app/lib/src/debugger/debugger_model.dart
+++ b/packages/devtools_app/lib/src/debugger/debugger_model.dart
@@ -294,6 +294,13 @@ class FileNode extends TreeNode<FileNode> {
 
   bool get hasScript => scriptRef != null;
 
+  String _fileName = '';
+
+  /// Returns the name of the file.
+  ///
+  /// May be empty.
+  String get fileName => _fileName;
+
   /// Given a flat list of service protocol scripts, return a tree of scripts
   /// representing the best hierarchical grouping.
   static List<FileNode> createRootsFrom(List<ScriptRef> scripts) {
@@ -310,6 +317,7 @@ class FileNode extends TreeNode<FileNode> {
       }
 
       node.scriptRef = script;
+      node._fileName = ScriptRefUtils.fileName(script);
     }
 
     // Clear out the _childrenAsMap map.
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/debugger/debugger_model.dart
```diff
@@ -353,6 +361,9 @@ class FileNode extends TreeNode<FileNode> {
 }
 
 class ScriptRefUtils {
+  static String fileName(ScriptRef scriptRef) =>
+      Uri.parse(scriptRef.uri).path.split('/').last;
```
kenzieschmoll, line 365: would it be more performant if we cached this value on the ScriptRef rather than computing this each time?

packages/devtools_app/lib/src/debugger/scripts.dart
```diff
@@ -148,10 +150,36 @@ class ScriptPickerState extends State<ScriptPicker> {
               ),
               const SizedBox(width: densePadding),
               Expanded(
-                child: Text(
-                  node.name,
-                  maxLines: 1,
-                  overflow: TextOverflow.ellipsis,
+                child: Row(
+                  children: [
+                    Expanded(
```
kenzieschmoll, line 155: we can probably remove this Row and Expanded since Expanded is the only child of Row

packages/devtools_app/lib/src/debugger/scripts.dart
```diff
@@ -148,10 +150,36 @@ class ScriptPickerState extends State<ScriptPicker> {
...
+                        mainAxisAlignment: MainAxisAlignment.center,
+                        crossAxisAlignment: CrossAxisAlignment.start,
+                        children: [
+                          if (!node.hasScript)
+                            Text(
+                              node.name,
+                              maxLines: 1,
+                              overflow: TextOverflow.ellipsis,
+                            ),
+                          if (node.hasScript) ...[
```
kenzieschmoll, line 166: you should be able to use an inline else here instead 

```
if (condition) 
  Text(...)
else
  ...[...],
```

packages/devtools_app/lib/src/debugger/debugger_model.dart
grouma: It's cached where it's used in the `FileNode`. This is simply a helper method on top of it.

grouma:
This should be good. PTAL.

kenzieschmoll (approved):

PR #2785, opened 2021-03-11
Improve file history picker UX

grouma:
Related to https://github.com/flutter/devtools/pull/2784

Mirror the file picker UX in Chrome which shows file results in two lines, the first line is simply the file name, the second line is the entire path potentially truncated. This helps for very long files as it's difficult to distinguish between truncated file names, especially in the history workflow as we do not have a tooltip.

Screen Shot 2021-03-10 at 5 10 15 PM


packages/devtools_app/lib/src/debugger/debugger_model.dart
```diff
@@ -353,6 +353,9 @@ class FileNode extends TreeNode<FileNode> {
 }
 
 class ScriptRefUtils {
+  static String fileName(ScriptRef scriptRef) =>
```
kenzieschmoll, line 356: wasn't this landed in your last PR?

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -383,11 +383,22 @@ class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
...
+            Text(
+              ScriptRefUtils.fileName(scriptRef),
+              maxLines: 1,
+              overflow: TextOverflow.ellipsis,
+            ),
+            Text(
+              scriptRef.uri,
+              overflow: TextOverflow.ellipsis,
+              maxLines: 1,
+              style: const TextStyle(fontSize: 14.0, color: Colors.grey),
```
kenzieschmoll, line 399: can you used theme.subtleTextStyle here instead?

packages/devtools_app/lib/src/debugger/debugger_model.dart
grouma: Yeh. Just need to merge master.

packages/devtools_app/lib/src/debugger/codeview.dart
grouma: Done.

kenzieschmoll (approved):

packages/devtools_app/lib/src/debugger/codeview.dart
kenzieschmoll: also not sure if i missed this in your last PR, but if we are changing the color manually to gray there too, can we use subtle instead?

grouma: I'll update it in this PR and land it here.

PR #2795, opened 2021-03-12
Fixed location of exported memory stat JSON file.

terrylucas:
- Changed location of 'memory_log_YYYYMMDD_HH_MM' files for Flutter Desktop MacOS moved from Downloads to $TMPDIR (Mac system temp directory).
- Added notification/logging if version mismatch loading exported memory JSON file (not converted).
- Added notification displaying the name and location of the exported file.


packages/devtools_shared/lib/src/memory/memory_json.dart
```diff
@@ -38,10 +38,12 @@ abstract class MemoryJson<T> implements DecodeEncode<T> {
     if (payloadVersion != version) {
       // TODO(terry): Convert Payload TBD - only one version today.
       // TODO(terry): Notify user the file is being converted.
+      print('ERROR: Unable to convert JSON memory file payload version=$payloadVersion.');
```
jacob314, line 41: change to a log message

jacob314 (approved):
lgtm

packages/devtools_shared/lib/src/memory/memory_json.dart
terrylucas: I've added a TODO, I'll create another PR that moves all of the config_specific/logger code into devtools_shared if we think it's important to handle this one print

Issue #1974, opened 2020-05-16
DevTools Memory Page: UX Punch List 2020


raison00:
##
## Update Overall UX 
 - [ ] Update Time Range Slider for optimized layout
 - [x] Add legend to right side of Dart Memory Heap graph
 - [x] Add legend to right side of Android Memory graph
- [ ] Show Android data by default when connected to an Android app
 - [ ] Remove Android Memory CTA if not connected to an Android app
 - [ ] Move Android Memory CTA: visible on conditional: if available, display button to top of page
- [ ] Make Export Memory button consistent with other export buttons (ie: match the Timeline page)
- [ ] Update the processing spinner on the Snapshot table, it should be centered vertically. 
- [x] Make tab for Heat Map instead of toggle/slider: convert to tab view with two tabs (one for Heat Map  and one for Snapshot table)
- [ ] Make Search bar consistent from Heat Map & Snapshot
- [x] Consider switching from Heat Map to Tree Map 
**Note:** [UX UI Color Updates for Tree Map](https://github.com/flutter/devtools/issues/2139)
- [ ] Make a small thumbnail of the Heat Map/Tree Map in the right of the timeline to give quick visual context 
- [ ] Gain better insight into the non-Dart Heap Memory
- - [ ] Visualize size and ability to surface non-Dart Heap Objects
- - [ ] Add visualization and raise awareness of where memory resources are allotted

##
- [ ] Unify Search & Filter Expectations 
- - [ ] Keep Search bar in Snapshot table view
- - [ ] Update bottom bar filter based on search/hide 
- - [ ] When search / filter is applied, show message same as logging page (IE: “showing 300 of 400”)
- - [ ] Specify Filter and visualize it (split screen: snapshot table on left, tree map on right) with capabilities to drill into each instance (with pagination) and maintain the visualization tree mapping context
-  - [ ] Can we combine search & filter? 
- - [ ] How does search work with - filters?
- - [ ] Advanced Search / Filter? UXR Testing to determine best practices to adopt

##
-  [ ] **Match global UX of DevTools pages**

- - [ ] Add orders as consistent with DevTools pages
- - [ ] Update Charts on page to align with DevTools (theme changes aren't always respected on current page)
- - [ ] Consider options for visualization and Snapshot information at the same time
- - [ ] Update spacing, padding of data entries
- - [ ] Add Header to bottom area to define section
- - [x] Use DevTools standard table
- - [x] Clean-up the design, less _busy_
[...]

PR #2799, opened 2021-03-15
Remove Android Memory CTA if not connected to an Android app and spinner centering

alexlindroos:
#1974 Are these changes still valid?

Goldens should be probably updated, but if I update them they are conflicting.


alexlindroos:
cc @terrylucas @kenzieschmoll

terrylucas (approved):
Sorry for the delay for investigating and fixing why goldens were failing in a general way on the bots.  They're fixed now.

Issue #2761, opened 2021-03-05
Timeline event zoom to frame is inaccurate when Raster finishes before UI


kenzieschmoll:
Expected: timeline zooms to fit entire frame into view
<img width="1217" alt="Screen Shot 2021-03-05 at 8 48 53 AM" src="https://user-images.githubusercontent.com/43759233/110146821-e0c90d80-7d8f-11eb-90d0-85371174fd37.png">

Actual: timeline uses Raster event end as end of frame event when UI thread activity finishes later
<img width="1219" alt="Screen Shot 2021-03-05 at 8 48 44 AM" src="https://user-images.githubusercontent.com/43759233/110146843-e58dc180-7d8f-11eb-819a-fbb1764e99c1.png">

[dart_devtools_2021_3_5-1614962908914315.json.zip](https://github.com/flutter/devtools/files/6091974/dart_devtools_2021_3_5-1614962908914315.json.zip)

PR #2802, opened 2021-03-15
Fix frame timing issue in performance page

kenzieschmoll:
fixes https://github.com/flutter/devtools/issues/2761

### DIFF
--- a/packages/devtools_app/lib/src/performance/performance_model.dart
+++ b/packages/devtools_app/lib/src/performance/performance_model.dart
@@ -441,15 +441,21 @@ class FlutterFrame {
     type ??= event?.type;
     if (type == TimelineEventType.ui) {
       time.start = event?.time?.start;
+      // If [rasterEventFlow] has already completed, set the end time for this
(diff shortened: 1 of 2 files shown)


jacob314 (approved):
lgtm

packages/devtools_app/lib/src/performance/performance_model.dart
```diff
@@ -441,9 +441,23 @@ class FlutterFrame {
     type ??= event?.type;
     if (type == TimelineEventType.ui) {
       time.start = event?.time?.start;
+      if (rasterEventFlow != null) {
+        time.end = event?.time?.end;
+      }
     }
     if (type == TimelineEventType.raster) {
-      time.end = event?.time?.end;
+      // If there is not a [uiEventFlow] for this frame, there will either never
```
jacob314, line 449: nit: this comment is a little hard to read due to be written as almost a double negative.

kenzieschmoll: fixed.

Issue #1081, opened 2019-09-23
Memory legend renames.


terrylucas:
1) explain more in the doc? Need to review I think it is explained.  I think a possible solution was for get info circle with ? near the legends that tells what it is.

2) Capacity = Allocated
3) Used = Dart/Flutter
4) External = Flutter engine/rendering

cobblest:
It will also be great if we could show the complete legend list when user open the Memory page. Right now they need to scroll down to see the whole list and all the legend options.

cobblest:
@terrylucas could we update these, especially the Used - Dart renaming?

PR #2806, opened 2021-03-16
Memory legends rename

alexlindroos:
#1081 Renaming of the memory legends items


devoncarew:
cc @kenzieschmoll @terrylucas

packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -683,11 +685,14 @@ class MemoryBodyState extends State<MemoryBody>
...
-        child: Row(
-          children: rowChildren,
-        ));
+    return Expanded(
+      child: Container(
+          padding: const EdgeInsets.fromLTRB(10, 0, 0, 2),
+          child: Row(
+            mainAxisAlignment: MainAxisAlignment.spaceBetween,
+            children: rowChildren,
+          )),
```
kenzieschmoll, line 694: nit: not your fault, but please add a trailing comma between these two parens here

kenzieschmoll:
Thanks for the PR! Could you please add a screenshot to the PR description of what the new legend looks like?

packages/devtools_app/lib/src/memory/memory_screen.dart
alexlindroos: Thanks, fixed!

alexlindroos:
Hi @kenzieschmoll,

Sure thing, here you have:

Screenshot 2021-03-17 at 9 31 54
Default view

Screenshot 2021-03-17 at 9 32 12
Android Memory View

packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -969,9 +975,9 @@ class MemoryBodyState extends State<MemoryBody>
 
   List<Widget> displayVmDataInHover(ChartsValues chartsValues) {
     const rssDisplay = 'RSS';
-    const capacityDisplay = 'Capacity';
-    const usedDisplay = 'Used';
-    const externalDisplay = 'External';
+    const capacityDisplay = 'Allocated';
+    const usedDisplay = 'Dart/Flutter';
+    const externalDisplay = 'Flutter engine/rendering';
```
terrylucas, line 980: This now seems too long.  I wonder if a more accurate name might be "Dart/Flutter Native" as this more accurate.  It contains native objects that a Dart embedder would manage with a Dart wrapper .

packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -1217,9 +1223,9 @@ class MemoryBodyState extends State<MemoryBody>
                 padding: const EdgeInsets.fromLTRB(5, 0, 0, 4),
                 child: Text('Memory Legend', style: legendHeading),
               ),
-              legendRow(name1: 'Capacity', image1: capacityLegend),
-              legendRow(name1: 'Used', image1: usedLegend),
-              legendRow(name1: 'External', image1: externalLegend),
+              legendRow(name1: 'Allocated', image1: capacityLegend),
```
terrylucas, line 1226: All of these names should be pulled from one const instead of two of the same strings.

packages/devtools_app/test/chart_test.dart
```diff
@@ -57,7 +57,7 @@ void main() {
             symbol: ChartSymbol.disc,
             diameter: 1.5,
           ),
-          name: 'External',
+          name: 'Flutter engine/rendering',
```
terrylucas, line 60: Same here, and other places where each trace is created, should reference the one const name.

terrylucas (changes requested):
There are more changes that need to be made.  These are problems that existed in the code that I didn't notice until this review - sorry.

Please update the legend and hover images in the PR after this so we can see that there is enough border around these overlays.

Also, show a  hover card screenshot with the settings 'Display Data in Units' turned off too.

alexlindroos:
@terrylucas 

Screenshot 2021-03-22 at 11 10 31
Screenshot 2021-03-22 at 11 10 55
Screenshot 2021-03-22 at 11 11 08

packages/devtools_app/lib/src/memory/memory_screen.dart
terrylucas: Alex can you change this label "Flutter engine/rendering" to "Dart/Flutter Native"? Plus update the screen shots?

I think after this the PR will be fine.  I can commit and follow up with new goldens so the bots turn green.

alexlindroos: Yes, I changed those now. For me Dart/Flutter Native seems too close to Dart/Flutter, so it might get confusing for some developers. Yet I also agree that Flutter engine/rendering is too long string.

alexlindroos:
Creating a new PR

PR #2807, opened 2021-03-16
Add EvalOnDartLibrary utilities

rrousselGit:
The `Disposable` interface separates `EvalOnDartLibrary` from `ObjectGroup` which is specific to the inspector service.

`asyncEval` allows optionally using `await` inside an expression, in which case the expression will not resolve until the future awaited completes.


packages/devtools_app/lib/src/eval_on_dart_library.dart
```diff
@@ -1,11 +1,13 @@
-// Copyright 2018 The Chromium Authors. All rights reserved.
```
jacob314, line 1: leave existing copyright dates unchanged. only new files get new copyright dates.

packages/devtools_app/lib/src/eval_on_dart_library.dart
```diff
@@ -156,25 +178,229 @@ class EvalOnDartLibrary {
...
+          scope: scope,
+          exception: result,
+        );
+      }
+    } catch (err, stack) {
+      _handleError(err, stack);
+      rethrow;
+    }
+
+    // We throw if the request was canceled instead of returning `null` because
```
jacob314, line 391: are you sure this is the case? You should get an InstanceRef with a null value not null. I think you might have been seeing null getting returned when eval was cancelled.

packages/devtools_app/lib/src/eval_on_dart_library.dart
```diff
@@ -14,6 +16,22 @@ import 'globals.dart';
 import 'inspector/inspector_service.dart';
 import 'vm_service_wrapper.dart';
 
+abstract class Disposable {
+  bool get disposed;
+
+  void dispose();
+}
+
+class IsAlive extends Disposable {
```
jacob314, line 25: I'm not seeing a need for two classes in this CL. Just make this the default implementation of Disposable and annotate dispose with mustCallSuper(). In dart someone can implement a class if they need to use it as an interface.

packages/devtools_app/lib/src/eval_on_dart_library.dart
```diff
@@ -95,28 +114,31 @@ class EvalOnDartLibrary {
...
+    @required Disposable isAlive,
     Map<String, String> scope,
   }) {
     return addRequest(isAlive, () => _eval(expression, scope: scope));
   }
 
+  Future<LibraryRef> _waitForLibraryRef() async {
+    while (true) {
+      final libraryRef = await _libraryRef.future;
+      if (_libraryRef.isCompleted) {
```
jacob314, line 126: this is a good race condition to check for but you should check for it by checking whether the isolateRef matches the isolateRef when the method started.

rrousselGit: I didn't write this code. It was an "extract as method" from `eval`

packages/devtools_app/lib/src/eval_on_dart_library.dart
```diff
@@ -156,25 +178,229 @@ class EvalOnDartLibrary {
...
+  }) async {
+    final ref = await safeEval(expression, isAlive: isAlive, scope: scope);
+    if (ref == null) return null;
+
+    return getInstance(ref, isAlive);
+  }
+
+  static int _nextAsyncEvalId = 0;
+
+  EvalOnDartLibrary _dartDeveloperEvalCache;
```
jacob314, line 227: you should dispose these caches if they are non null in the dispose method for this eval.

packages/devtools_app/lib/src/eval_on_dart_library.dart
```diff
@@ -156,25 +178,229 @@ class EvalOnDartLibrary {
...
+
+        return await service.evaluate(
+          isolateId,
+          libraryRef.id,
+          expression,
+          scope: scope,
+        );
+      });
+
+      if (result == null) return null;
```
jacob314, line 363: this case is where the EvalOnDartLibrary was disposed while the request was in progress. Do you want to throw an error here or only if it was disposed before the request started?

jacob314 (commented):
Thanks for adding all this great functionality to eval!
A few minor nits.

packages/devtools_app/lib/src/eval_on_dart_library.dart
rrousselGit: Upon testing it, it appears that `eval('null')` does return an InstanceRef.

But `eval('[null]')` followed by a `getInstance` and reading `instance.elements.first` returns `null` (and I believe the issue is the same for maps)

So it appears to be inconsistent

jacob314:
lgtm

packages/devtools_app/lib/src/eval_on_dart_library.dart
jacob314: feel free to just make this a TODO for me to fix.

packages/devtools_app/lib/src/eval_on_dart_library.dart
```diff
@@ -156,25 +178,229 @@ class EvalOnDartLibrary {
...
+    @required Disposable isAlive,
+    Map<String, String> scope,
+  }) async {
+    final id = _nextAsyncEvalId++;
+
+    // start awaiting the event before starting the evaluation, in case the
+    // event is received before the eval function completes.
+    final future = serviceManager.service.onExtensionEvent.firstWhere((event) {
+      return event.extensionKind == 'future_completed' &&
+          event.extensionData.data['id'] == id;
```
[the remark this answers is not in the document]
rrousselGit: Done

packages/devtools_app/lib/src/eval_on_dart_library.dart
rrousselGit: I've removed the null checks and updated the comment

rrousselGit:
I have fixed the different nits so this should be good to go I believe

packages/devtools_app/lib/src/eval_on_dart_library.dart
jacob314: Sounds good. Lets file a bug to get the VM Service consistent on this point. I think you have discovered a bug in the VM Service api or a micro-optimization that makes the api less consistent.

jacob314 (approved):
lgtm

PR #2808, opened 2021-03-16
Fix some flame chart scrolling bugs

kenzieschmoll:


packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -384,6 +381,10 @@ abstract class FlameChartState<T extends FlameChart, V> extends State<T>
             FlameChart.maxZoomLevel,
           );
           await zoomTo(newZoomLevel, jump: true);
+          if (newZoomLevel == FlameChart.minZoomLevel &&
+              horizontalController.offset != 0.0) {
```
jacob314, line 385: What if there is so much content that a scroll offset other than zero is still valid at the min zoom level?

packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -415,8 +416,7 @@ abstract class FlameChartState<T extends FlameChart, V> extends State<T>
       // to call this that guarantees the scroll controller offsets will be
       // updated for the new zoom level and layout size
       // https://github.com/flutter/devtools/issues/2012.
-      horizontalController.jumpTo(
-          newScrollOffset.clamp(FlameChart.minScrollOffset, maxScrollOffset));
+      scrollToX(newScrollOffset, jump: true);
```
jacob314, line 419: scrolllToX doesn't seem safe to call after a zoom as the scrollbar will still have the old scroll extent for a frame if I remember correctly. Perhaps we need a custom scroll controller class that lets us update its extent immediately based on the new size to avoid all these annoying edge cases?

packages/devtools_app/lib/src/charts/flame_chart.dart
kenzieschmoll: That is not currently how the flame_chart is implemented. At a zoom level of 1.0 (min zoom level), all of the content is rendered in view, so 0.0 is the only scroll offset we would want.

packages/devtools_app/lib/src/charts/flame_chart.dart
kenzieschmoll: This change is a no-op to use the helper method instead of calling jumpTo and manually clamping the offset. We do have to scroll though in order to keep the mouse centered on the position of the scroll.

I'd have to dig in a little deeper to remember the timing of everything, but that sounds reasonable to do in a future improvement.

jacob314 (approved):

jacob314:
lgtm_dog

PR #2809, opened 2021-03-16
Fix bugs with DevTools title and move title code to separate file

kenzieschmoll:
There were bugs with the title not updating after an app connected. This fixes those and moves the title code to a new file. Moving title code to separate file should make it simple for us to do custom titles in g3

### DIFF
--- a/packages/devtools_app/lib/src/scaffold.dart
+++ b/packages/devtools_app/lib/src/scaffold.dart
@@ -12,6 +12,7 @@ import 'package:provider/provider.dart';
 import 'analytics/prompt.dart';
 import 'analytics/provider.dart';
 import 'app.dart';
+import 'auto_dispose_mixin.dart';
 import 'banner_messages.dart';
 import 'common_widgets.dart';
 import 'config_specific/drag_and_drop/drag_and_drop.dart';
@@ -25,6 +26,7 @@ import 'routing.dart';
 import 'screen.dart';
 import 'status_line.dart';
 import 'theme.dart';
+import 'title.dart';
 
 /// Scaffolding for a screen and navigation in the DevTools App.
 ///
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/scaffold.dart
```diff
@@ -137,6 +135,9 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
         frameworkController.onConnectVmEvent.listen(_connectVm);
     _showPageSubscription =
         frameworkController.onShowPageId.listen(_showPageById);
+
+    cancel();
```
jacob314, line 139: cancel in initState seems very strange. why do we need this?

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -275,6 +276,8 @@ class ServiceConnectionManager {
     vm = null;
     sdkVersion = null;
     connectedApp = null;
+    generateDevToolsTitle();
```
jacob314, line 279: I would either move the devtools title functionality directly into service_manager or add logic to title so it is just listening for events on the service manager.

kenzieschmoll: That would make it challenging to modify titles inside google3. 

We also need to know what the connected app type is, which is initialized in the ConnectedApp class. This class doesn't currently dispatch any events. We could change it so that it does, but I'm not sure what extra value that would add. We only initialize connected app values on vm service opened, and we clear out the connected app here in vm service closed. I suppose we could add a method to ConnectedApp `detach` or similar, that calls `generateDevToolsTitle` but that feels like a lateral change without much effect.

packages/devtools_app/lib/src/scaffold.dart
kenzieschmoll: Ah. copied this pattern from what we do in other places where we call `addAutoDisposeListener`, but now I realize those are calls in didChangeDependencies where the listener could be added twice. removed.

jacob314 (approved):
lgtm

Issue #2797, opened 2021-03-13
Trim parentheses from hover spans 


grouma:
The current debugger hover logic reuses the syntax highlighting parser logic for splitting spans to hover over. It looks like this logic may include parentheses in the span. We should implement a simple heuristic to remove these spans so that hover is more reliable. Related: https://github.com/flutter/devtools/issues/2796

annagrin:
Another example (with `=>` getting into the span):

![image](https://user-images.githubusercontent.com/12737689/111013223-7eee4200-8353-11eb-9c06-83b6da502529.png)

grouma:
I was able to reproduce this with several `=>` examples. Since we piggy back off the syntax grammar I don't think it makes sense to add a bunch of heuristics on top of it.

@bkonyi is this something you can look into fixing? Can we update the grammar such that parens, spaces and fat arrows cause new spans?

devoncarew:
I suspect that the syntax grammar might not be optimized to create tokens that work well with expression evaluation. How is it (the grammar) being used currently? Would you just want to identify the character position under the cursor and then work backwards and forwards from there, looking for valid dart identifier chars?

grouma:
> I suspect that the syntax grammar might not be optimized to create tokens that work well with expression evaluation. How is it (the grammar) being used currently? 

The grammar produces a tree of spans.

> Would you just want to identify the character position under the cursor and then work backwards and forwards from there, looking for valid dart identifier chars?

Currently we find the underlying span that we are hovered over then potentially merge spans due to `.`'s. That works well for the most part because the spans generally contain just the information we need. I'm hesitant to go with the character approach as I imagine we'll end up something very similar to the grammar which seems redundant in my mind.

bkonyi:
> I was able to reproduce this with several `=>` examples. Since we piggy back off the syntax grammar I don't think it makes sense to add a bunch of heuristics on top of it.
> 
> @bkonyi is this something you can look into fixing? Can we update the grammar such that parens, spaces and fat arrows cause new spans?

In theory we could, (I'd have to take a closer look at the grammar)  but it's possible it could have an impact on the performance of the syntax highlighter if we break things into smaller spans. If we're only interested in tokens that can be inspected, Devon's proposal might make more sense than hacking the textmate grammar to produce more fine-grained spans.

PR #2810, opened 2021-03-16
Improve hover logic

grouma:
Closes https://github.com/flutter/devtools/issues/2797

This gets us in a better place but isn't perfect. If you hover inside a quoted string then the entire string isn't considered a single entity for evaluation. Today at HEAD, if the string doesn't contain interpolation, the entire string is considered for evaluation. However that's not very useful as the result is obvious. I think it makes sense to move forward with this change that improves the overall hover experience for useful cases.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/hover.dart
+++ b/packages/devtools_app/lib/src/debugger/hover.dart
@@ -55,7 +55,8 @@ String wordForHover(double dx, TextSpan line) {
 /// located.
 int _hoverIndexFor(double dx, TextSpan line) {
   int hoverIndex = -1;
-  for (var i = 0; i < line.toPlainText().length; i++) {
+  final length = line.toPlainText().length;
+  for (var i = 0; i < length; i++) {
     final painter = TextPainter(
       text: truncateTextSpan(line, i + 1),
       textDirection: TextDirection.ltr,
--- a/packages/devtools_app/lib/src/debugger/hover.dart
+++ b/packages/devtools_app/lib/src/debugger/hover.dart
@@ -7,33 +7,27 @@ import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 
 import '../theme.dart';
+import '../ui/utils.dart';
 
 /// Regex for valid Dart identifiers.
 final _identifier = RegExp(r'^[a-zA-Z0-9]|_|\$');
 
 /// Returns the word in the [line] for the provided hover [dx] offset given
 /// the [line]'s [textStyle].
-String wordForHover(double dx, TextSpan line, TextStyle textStyle) {
-  final painter = TextPainter(
-    text: TextSpan(children: [line], style: textStyle),
-    textDirection: TextDirection.ltr,
-  )..layout();
+String wordForHover(double dx, TextSpan line) {
+  String word = '';
+  final hoverIndex = _hoverIndexFor(dx, line);
   final lineText = line.toPlainText();
-  final hoverIndex = ((dx / painter.width) * (lineText.length - 1)).round();
-
-  // Use a psuedo StringBuffer for performance.
-  final word = <String>[];
-
   if (hoverIndex >= 0 && hoverIndex < lineText.length) {
     final hoverChar = lineText[hoverIndex];
-    word.add(hoverChar);
+    word = '$word$hoverChar';
     if (_identifier.hasMatch(hoverChar) || hoverChar == '.') {
       // Merge trailing valid identifiers.
       int charIndex = hoverIndex + 1;
       while (charIndex < lineText.length) {
         final character = lineText[charIndex];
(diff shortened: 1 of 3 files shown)


devoncarew:
It may not be practical or desired, but one other possibility is to use `package:analyzer` to tokenize the file. We'd have to guess less about being in things like string interpolations. From memory, there is an API to return tokens from a string input.

This PR is probably the better solution but we might want to track the idea of using the analyzer in an issue.

packages/devtools_app/lib/src/debugger/hover.dart
```diff
@@ -8,52 +8,53 @@ import 'package:flutter/material.dart';
...
+        }
+        charIndex++;
+      }
+
+      // Merge preceding characters including those linked by a `.`.
+      charIndex = hoverIndex - 1;
+      while (charIndex >= 0) {
+        final character = lineText[charIndex];
+        if (_identifier.hasMatch(character) || character == '.') {
+          word.insert(0, character);
```
jacob314, line 48: nit: inserting at the start of an List isn't performant string concat is more likely to be fast for this case due to perhaps using Ropes under the hood.

packages/devtools_app/lib/src/debugger/hover.dart
```diff
@@ -8,52 +8,53 @@ import 'package:flutter/material.dart';
...
-  word = line.children[index].toPlainText();
-  word = _mergedLinkedWords(word, index, line);
-  return word;
-}
+  final painter = TextPainter(
+    text: TextSpan(children: [line], style: textStyle),
+    textDirection: TextDirection.ltr,
+  )..layout();
+  final lineText = line.toPlainText();
+  final hoverIndex = ((dx / painter.width) * (lineText.length - 1)).round();
```
jacob314, line 22: try to avoid assuming letters have fixed width. We already know that isn't accurate for our currently styled text (unclear exactly why). Better to instead measure the length up until each point so we are exactly right at the hover location.

packages/devtools_app/test/hover_test.dart
```diff
@@ -9,12 +9,15 @@ import 'package:test/test.dart';
 const _defaultStyle = TextStyle();
 const _textSpan = TextSpan(children: [
   TextSpan(text: 'hello'),
```
jacob314, line 11: bonus points: make this text use a variable width font to ensure it is robust.

packages/devtools_app/lib/src/debugger/hover.dart
grouma: Discussed offline. Using a slightly different approach.

packages/devtools_app/test/hover_test.dart
grouma: Done.

packages/devtools_app/lib/src/debugger/hover.dart
grouma: Changed it to be simply string based. Surprised that Dart's StringBuffer doesn't have an insert API.

grouma:
@jacob314 PTAL.

packages/devtools_app/lib/src/debugger/hover.dart
jacob314: its hard for StringBuffer to insert efficiently without switching to Ropes.

packages/devtools_app/lib/src/debugger/hover.dart
```diff
@@ -7,53 +7,65 @@ import 'package:flutter/gestures.dart';
...
+/// Returns the index in the Textspan's plainText for which the hover offset is
 /// located.
-int _hoverIndex(double dx, TextSpan line, TextStyle textStyle) {
-  int index = 0;
-  var cumulativeWidth = 0.0;
-  while (index < line.children.length) {
-    final span = line.children[index];
+int _hoverIndexFor(double dx, TextSpan line) {
+  int hoverIndex = -1;
+  for (var i = 0; i < line.toPlainText().length; i++) {
```
jacob314, line 58: nit: cache length as computing it is expensive.
`for (var i = 0, length = line.toPlainText().length; i++)`

packages/devtools_app/test/hover_test.dart
```diff
@@ -6,33 +6,35 @@ import 'package:devtools_app/src/debugger/hover.dart';
 import 'package:flutter/widgets.dart';
 import 'package:test/test.dart';
 
-const _defaultStyle = TextStyle();
 const _textSpan = TextSpan(children: [
   TextSpan(text: 'hello'),
+  TextSpan(text: ' '),
```
jacob314, line 11: add some textStyles (e.g. bold or font sizes) now that you support it correctly so that your test verifies you are really doing the right thing.

jacob314 (approved):
lgtm

PR #2813, opened 2021-03-17
Fixed Goldens working on bots.

terrylucas:
Golden masters will correctly work on the bots for all memory tests.  There were a number of issues:

- Fixed timing problem not waiting long enough for screen to settle before screenshot.
- Having exported files in /tmp directory can adjust screen ever so slightly (drop-down widens). Added check in memory_screen_test to fail if any exported files.
- Auto-complete TextField can have different enable colors on different OS and platform (desktop vs/ web) added constant color for our search TextField.

Any failure on the bot was impossible to determine what didn't match on the screen between the test image versus the master image.  The testImage is the image of the test running (bot) and masterImage is the checked in golden master.  Added a a github artifact "test-image-failures" (Danny thanks for the pointers).  Any golden matching failures on the bot for test_ddc and test_dart2js will display "image failures" artifacts (expanded below).

image

When the runs are complete (test_ddc and test_dart2js) then the "Artifacts" drop-down will appear in the top-right area (see above).  Clicking on the Artifact displays:

image

Clicking on the test-image-failures will download the zip file of the failures.  The important files for each golden files are:

```<golden_filename>```**_masterImage.png** ``` - PNG file to compare (committed in github).```
```<golden_filename>```**_testImage.png**```  - PNG of screenshot of running test on bot.```
```<golden_filename>```**_maskedDiff.png**``` - masked file (color adjusted).```
```<golden_filename>```**_isolatedDiff.png**``` - differences detected.```

For example:

**masterImage**
image

**testImage**
image

**isolatedDiff** _Show just the differences top area and orange rectangle in middle right-side (color different from master and test images)_

image

### CODE REVIEW

File: .github/workflows/build.yaml [hunk 1]
```diff
@@ -63,12 +63,20 @@ jobs:
...
         env:
           BOT: ${{ matrix.bot }}
           PLATFORM: vm
         run: ./tool/bots.sh
 
+      - name: image failures
+        uses: actions/upload-artifact@v1
+        if: failure() # Only if failure then failures directory exists.
+        with:
+          name: test-image-failures-${{ matrix.os }} # Name for the artifact
```
DanTup, line 77: The "matrix.os" here is currently being replaced with an empty string (as there's no `os` in the `matrix` config). It would be removed, although if it may be run on multiple OSes in the future, perhaps it's worth leaving.

DanTup (commented):
The GH Actions bits look good to me (though I added a comment about an unset variable in the name). I'm less familiar with the other changes though, so best checked by Jacob.

packages/devtools_app/test/memory_screen_test.dart
```diff
@@ -368,12 +370,40 @@ void main() {
       await tester.pumpAndSettle(_twoSeconds);
     }
 
+    Future<void> checkGolden(String goldenFilename, {Key key}) async {
+      // Await delay for golden comparison.
+      await tester.pumpAndSettle(const Duration(seconds: 2));
```
jacob314, line 375: nit: be consistent about using pumpAndSettle and pumpAndSettleTwoSeconds.

packages/devtools_app/test/memory_screen_test.dart
```diff
@@ -368,12 +370,40 @@ void main() {
...
 
+    // Any exported memory files in the /tmp or $TMPDIR or /var/folders/Downloads will cause
+    // the memory page's drop-down widget width to be wider than normal (longest exported file
+    // name). For goldens don't generate snapshots as they we be slightly different on the bots.
     expect(
-        find.byKey(HeapTreeViewState.dartHeapAnalysisTabKey), findsOneWidget);
+      controller.memoryLog.offlineFiles().isEmpty,
+      isTrue,
+      reason:
+          '\n\n=========================================================================\n'
```
jacob314, line 397: nit: little odd to start the reason with `\n\n`

packages/devtools_app/test/memory_screen_test.dart
```diff
@@ -604,68 +619,53 @@ void main() {
...
+    await checkGolden(
+      'goldens/allocation_dropdown_hilight_line_3_golden.png',
+      key: searchAutoCompleteKey,
     );
-    // Await delay for golden comparison.
-    await pumpAndSettleTwoSeconds();
 
     // SecondClass hilighted.
     await downArrow(autoCompletes4.indexOf('SecondClass'));
     await pumpAndSettleOneSecond();
```
jacob314, line 642: you are now waiting 2+1 =3 seconds for some of these check golden cases. I would remove the pumpAndSettleOneSecond calls or remove the two second delay from check goldens and just add the delay you need where you need it.

packages/devtools_app/lib/src/theme.dart
```diff
@@ -339,6 +339,12 @@ const wideSearchTextWidth = 400.0;
 const defaultSearchTextWidth = 200.0;
 const defaultTextFieldHeight = 36.0;
 
+/// Default color of cursor and color used by search's TextField.
+Color searchColor = Colors.lightBlueAccent[400];
```
jacob314, line 343: Comment on where this is from. Does this match the default on a specific platform? Comment why we don't just use the default so someone viewing the theme understands why it makes sense to customize this to improve our testability and consistency.

jacob314 (approved):
lgtm_dog2

packages/devtools_app/test/memory_screen_test.dart
terrylucas: Oops, was messing with duration and forgot reverting. Done.

packages/devtools_app/test/memory_screen_test.dart
terrylucas: Did it this was because its the very first failure if it happens a lot of text above and below kind of obscuring so I wanted to make it stand out as its bad.  Can certainly remove.  If there's a strong opinion I'll remove.

packages/devtools_app/test/memory_screen_test.dart
terrylucas: They weren't needed I think this was part of early testing. They're gone.

packages/devtools_app/lib/src/theme.dart
terrylucas: Added comment on why we need a consistent color for golden snapshot compares.

After looking carefully at the theme-ing code the default is Color.blue (blue[500]) checked with a color picker to confirm.

Issue #2779, opened 2021-03-10
CPU profiler "Total Time" column is sometimes truncated


DanTup:
While doing some profiling of the analysis server, the Total Time column is truncated hiding the percentages (even on an ultrawide screen):

![Screenshot 2021-03-10 at 11 10 34](https://user-images.githubusercontent.com/1078012/110620755-51f23180-8191-11eb-8d27-1c57a1fa745d.png)

As far as I can tell there's no way to resize this column, nor a tooltip showing the full contents of the cell.

kenzieschmoll:
Danny can you click "Export" and email the file to me? that way I can repro easily and take a look. thanks.

DanTup:
Sorry for the delay! Here's a new capture:

[dart_devtools_2021_3_15-1615818858678000.json.txt](https://github.com/flutter/devtools/files/6142101/dart_devtools_2021_3_15-1615818858678000.json.txt)

It's the Call Tree table that shows it:

![Screenshot 2021-03-15 at 14 35 16](https://user-images.githubusercontent.com/1078012/111170292-ad5e5e00-859b-11eb-9aa1-e8c65a4e49e6.png)

Issue #2047, opened 2020-06-05
Use flexed / dynamic widths for table columns


kenzieschmoll:
When we use fixed widths for table columns, it can lead to a lot of unused space:
![Screen Shot 2020-06-05 at 2 17 34 PM](https://user-images.githubusercontent.com/43759233/83923682-9ac71680-a737-11ea-947a-dfe5d2e8ada1.png)

kenzieschmoll:
in our existing column widget calculation code, we should look into using the heuristic of measuring the longest text for a column instead of using `TreeColumnWidthExtension._defaultColumnWidth`

PR #2814, opened 2021-03-18
Increase size of total time column in CPU profiler

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/2779

The bigger issue here is that we don't have flexible column widths, but that is a larger fish to fry: https://github.com/flutter/devtools/issues/2047


jacob314 (approved):
lgtm

PR #2818, opened 2021-03-18
bump shelf proxy version to 1.0.0

kenzieschmoll:


kenzieschmoll:
This is blocked by `dds` bumping shelf version to 1.0.0 and publishing a new `dds`. @bkonyi 

Once that happens, this CL should be updated with the new `dds` version and devtools should be published again. The DevTools publish for 2.0.0+2 should be a cherry pick of this commit into the v2.0.0+1 tag.

Issue #2796, opened 2021-03-13
Hover over shows compilation errors


annagrin:
**Issue:**

Hover over variables shows compilation errors:

![image](https://user-images.githubusercontent.com/12737689/111010956-d76e1100-834c-11eb-97ec-be5618713d94.png)

**Compilation error from expression compiler**

```
[FINEST] ExpressionCompilerService: Failed to compile " galleryInjector)": org-dartlang-debug:synthetic_debug_expression:1:17: Error: Expected one expression, but found additional input.
 galleryInjector)
```

**Suggestion**

We currently hide compilation errors on hover over in VSCode and IntelliJ in order to keep the popups meaningful, should we match that behavior in devtools? We do show compilation errors in expression compilation box and in watch window.

annagrin:
@grouma  - should we match the behavior in other UIs here?

grouma:
This should be the behavior in DevTools. When the corresponding request throws and error we silently ignore it: https://github.com/flutter/devtools/blob/a05ebe4c1e70b9feafca844a7e7e5674fc7662dc/packages/devtools_app/lib/src/debugger/codeview.dart#L629

My guess is that `package:dwds` is not returning the correct error response.

annagrin:
This result is a valid result from dwds.evaluateInFrame - it does not throw and returns an `ErrorRef` which is treated differently dependent on the context (in VSCode, at least):
- in hover over - do not show
- in expression evaluation box or watch window - show.

grouma:
Thanks for the details. I'll take a look next week.

grouma:
@annagrin I'm having difficulty reproducing this. Can you provide repro steps, maybe with some edited code [here](https://github.com/dart-lang/webdev/blob/master/example/web/main.dart)?

annagrin:
@grouma I am having problems using the latest devtools with that example, some constraint solving fails. My repro is in google3 and it requires https://critique-ng.corp.google.com/cl/362633754 to work for now. We can also wait until all SDK and dwds changes propagate to flutter for an easy repro.

PR #2831, opened 2021-03-23
Ignore errors for hover

grouma:
Closes https://github.com/flutter/devtools/issues/2796

### DIFF
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -610,7 +610,7 @@ class _LineItemState extends State<LineItem> {
       if (word != '') {
         try {
           final response = await _debuggerController.evalAtCurrentFrame(word);
-          if (response is ErrorRef) return;
+          if (response is! InstanceRef) return;
           final variable = Variable.fromRef(response);
           await _debuggerController.buildVariablesTree(variable);
           _hoverCard = HoverCard(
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -610,6 +610,7 @@ class _LineItemState extends State<LineItem> {
       if (word != '') {
         try {
           final response = await _debuggerController.evalAtCurrentFrame(word);
+          if (response.json['kind'] == 'error') return;
```
jacob314, line 613: I would have thought you could use the type of the response.

grouma: Yeah that's what I thought too but the type is `Response`.

grouma: Found a better alternative.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -610,6 +610,7 @@ class _LineItemState extends State<LineItem> {
       if (word != '') {
         try {
           final response = await _debuggerController.evalAtCurrentFrame(word);
+          if (response is ErrorRef) return;
```
jacob314, line 613: not that it will happen, but it should be `is! InstanceRef` for robustness.

jacob314 (approved):
lgtm

grouma:
Merging on red as this should not impact tests and we want to roll this internally.

PR #2837, opened 2021-03-24
Use widgets for flame chart group labels instead of custom painters

kenzieschmoll:
This will allow us to do things like make the labels clickable and implement expandable and collapsible threads in the timeline flame chart

### DIFF
--- a/packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
+++ b/packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
@@ -1,7 +1,6 @@
 // Copyright 2019 The Chromium Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
-import 'dart:collection';
 import 'dart:math' as math;
 import 'dart:ui';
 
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
```diff
@@ -471,18 +479,75 @@ class TimelineFlameChartState
...
+        eventGroups.values.toList().indexOf(group),
+        colorScheme,
+      );
+      final backgroundWithOpacity = Color.fromRGBO(
+        backgroundColor.red,
+        backgroundColor.green,
+        backgroundColor.blue,
+        0.85,
+      );
+      children.addAll([
```
jacob314, line 517: why are topSpacer and bottomSpacer separate widgets instead of just padding on the container?

jacob314 (approved):
lgtm

jacob314 (commented):
lgtm

packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
kenzieschmoll: fixed.

PR #2839, opened 2021-03-24
Update network profiler to support dart:io HTTP profiling service extensions

bkonyi:


packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -361,3 +448,164 @@ class HttpRequestData extends NetworkRequest {
...
+    int timelineMicrosBase,
+    this._request,
+  ) : super(timelineMicrosBase) {
+    if (_request.isResponseComplete) {
+      serviceManager.service
+          .getHttpProfileRequest(
+            _request.isolateId,
+            _request.id,
+          )
+          .then((value) => _request = value);
```
kenzieschmoll, line 463: call getFullRequestData here instead of duplicating the code

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -361,3 +448,164 @@ class HttpRequestData extends NetworkRequest {
...
+      if (!_request.isResponseComplete) return null;
+      if (_responseBody != null) return _responseBody;
+      _responseBody = utf8.decode(fullRequest.responseBody);
+      if (contentType.contains('json')) {
+        print('formatting json');
+        _responseBody = const JsonEncoder.withIndent('  ').convert(json.decode(_responseBody));
+      }
+      return _responseBody;
+    } on FormatException {
+      return '<binary data>'; //fullRequest.responseBody.toString();
```
kenzieschmoll, line 595: remove comment?

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -361,3 +448,164 @@ class HttpRequestData extends NetworkRequest {
...
+    if (_request is! HttpProfileRequest) {
+      return null;
+    }
+    final fullRequest = _request as HttpProfileRequest;
+    try {
+      if (!_request.isResponseComplete) return null;
+      if (_responseBody != null) return _responseBody;
+      _responseBody = utf8.decode(fullRequest.responseBody);
+      if (contentType.contains('json')) {
+        print('formatting json');
```
kenzieschmoll, line 590: remove print?

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -361,3 +448,164 @@ class HttpRequestData extends NetworkRequest {
...
+        _responseBody = const JsonEncoder.withIndent('  ').convert(json.decode(_responseBody));
```
kenzieschmoll, line 591: Let's use `FormattedJson.encoder` to make sure we use the same encoder throughout DevTools. Can you also add a TODO that says consider returning a `Map<String, dynamic>` from responseBody so the `FormattedJson` widget can be used directly in the UI?

packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -171,6 +209,22 @@ class NetworkController
       currentValues.add(webSocket);
     }
 
+    if (timeline != null) {
+      _processTimeline(
```
kenzieschmoll, line 213: in what case will `timeline != null`? is this for backwards compatibility?

packages/devtools_app/lib/src/network/network_request_inspector_views.dart
```diff
@@ -450,6 +451,25 @@ class NetworkRequestOverviewView extends StatelessWidget {
...
+        child: Tooltip(
+          waitDuration: tooltipWait,
+          message: 'Response - ${msText(duration)}',
+          child: Container(
+            height: _timingGraphHeight,
+            color: _nextColor(),
+          ),
+        ),
+      ),
+    );
```
kenzieschmoll, line 472: this is duplicated above. put into a helper that takes a duration and a label.

packages/devtools_app/lib/src/network/network_service.dart
```diff
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:devtools_app/src/version.dart';
```
kenzieschmoll, line 5: nit: use relative imports

packages/devtools_app/lib/src/vm_service_wrapper.dart
```diff
@@ -503,6 +503,40 @@ class VmServiceWrapper implements VmService {
...
   }
 
+  /// The `getHttpProfile` RPC is used to retrieve HTTP profiling information
+  /// for requests made via `dart:io`'s `HttpClient`.
+  ///
+  /// The returned [HttpProfile] will only include requests issued after
+  /// [httpTimelineLogging] has been enabled or after the last
+  /// [clearHttpProfile] invocation.
+  Future<HttpProfile> getHttpProfile(String isolateId, {int updatedSince}) {
+    return trackFuture(
```
kenzieschmoll, line 513: don't we need to add `isProtocolVersionSupported` checks around all of these to make sure we don't try to call this method on an old version of the vm_service?

packages/devtools_app/lib/src/network/network_service.dart
```diff
@@ -39,18 +40,43 @@ class NetworkService {
...
+    assert(serviceManager.service != null);
+    if (serviceManager.service == null) return [];
+
+    final requests = <HttpProfileRequest>[];
+    await serviceManager.service.forEachIsolate((isolate) async {
+      final request = await serviceManager.service.getHttpProfile(
+        isolate.id,
+        updatedSince: networkController.lastRefreshMicros,
+      );
+      requests.addAll(request.requests);
```
kenzieschmoll, line 75: can we just return request.requests instead of making a copy of the list?

packages/devtools_app/lib/src/http/http_request_data.dart
bkonyi: Will do. I meant to remove this change for this PR and land it separately. WDYT?

packages/devtools_app/lib/src/network/network_controller.dart
bkonyi: Yes, this is for backward compatibility. When we start shipping a pinned DevTools with the SDK we can tear this all out.

packages/devtools_app/lib/src/network/network_service.dart
bkonyi: I'll blame VSCode for this... :-)

packages/devtools_app/lib/src/network/network_service.dart
bkonyi: No, since we need to get profiling data from each isolate, which each returns their own `request`.

packages/devtools_app/lib/src/vm_service_wrapper.dart
bkonyi: Good call, yes we do.

packages/devtools_app/lib/src/network/network_service.dart
kenzieschmoll: oops - didn't see `forEachIsolate`. sg.

packages/devtools_app/lib/src/http/http_request_data.dart
kenzieschmoll: remove the json formatting change? Sure, you can land as a follow up if you want.

packages/devtools_app/lib/src/http/http_request_data.dart
bkonyi: Not sure we'd be able to return `Map<String, dynamic>` in the general case since a simple string response is completely valid.

packages/devtools_app/lib/src/network/network_request_inspector_views.dart
```diff
@@ -420,6 +420,7 @@ class NetworkRequestOverviewView extends StatelessWidget {
     final _colors = [
       searchMatchColor,
       mainRasterColor,
+      mainAsyncColor,
```
bkonyi, line 423: This is something else I noticed. When we have only two colors available here and we encounter one really short event, there's no visible difference between the events on either side of the effectively invisible event. Having more colors here should help alleviate that particular issue, but I'm not sure if there's any particular reason for only using two in the first place?

kenzieschmoll: I figured that's what this is for - I've noticed that issue in the past. Adding a 3rd color sgtm

packages/devtools_app/lib/src/vm_service_wrapper.dart
```diff
@@ -503,13 +503,22 @@ class VmServiceWrapper implements VmService {
...
+
   /// The `getHttpProfile` RPC is used to retrieve HTTP profiling information
   /// for requests made via `dart:io`'s `HttpClient`.
   ///
   /// The returned [HttpProfile] will only include requests issued after
   /// [httpTimelineLogging] has been enabled or after the last
   /// [clearHttpProfile] invocation.
-  Future<HttpProfile> getHttpProfile(String isolateId, {int updatedSince}) {
+  Future<HttpProfile> getHttpProfile(String isolateId,
+      {int updatedSince}) async {
```
kenzieschmoll, line 520: nit: trailing comma

packages/devtools_app/lib/src/vm_service_wrapper.dart
```diff
@@ -520,7 +529,8 @@ class VmServiceWrapper implements VmService {
   }
 
   Future<HttpProfileRequest> getHttpProfileRequest(
-      String isolateId, int id) {
+      String isolateId, int id) async {
```
kenzieschmoll, line 532: nit: trailing comma

packages/devtools_app/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -2,6 +2,8 @@
 //  Generated file. Do not edit.
 //
 
+// clang-format off
```
kenzieschmoll, line 5: revert this change

kenzieschmoll (approved):

kenzieschmoll:
lgtm_army

kenzieschmoll (approved):

Issue #2297, opened 2020-08-21
Add 'clear' function to file import container


kenzieschmoll:
In a user study, it was not obvious to the user that a new file could be dragged into the file import container to replace the old file. The user refreshed the page to clear the selected file text. We should add a clear option 'X' to the selected file container.

We have similar function for the search bar in the Timeline flame chart.

PR #2843, opened 2021-03-26
Clear function for FileContainer

alexlindroos:
@kenzieschmoll 

#2297 Is this what we are trying to achieve?

Screenshot 2021-03-26 at 13 51 59

Screenshot 2021-03-26 at 13 51 43


kenzieschmoll:
Instead of adding an extra button, let's add an 'X' button to the end of the file name area. I'd wrap [this](https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/file_import.dart#L138) in a Row and add the X button to the end. Similar to what the clear function looks like for search fields in DevTools:
Screen Shot 2021-03-29 at 9 55 49 AM

We actually have a helper method `clearInputButton` in common_widgets.dart that you may be able to re-use or at least use as an example. This button should only be visible when a file is already selected.

PR #2846, opened 2021-03-30
Reworked UX for tracking call stack.

terrylucas:
Changed the UI on how call stacks for allocation tracking.  Moved the call stack from a nested tree to the right as a scroll-able list. Also, display the source file as a subtitle of the list.

image


jacob314:
High level: there is a lot of whitespace between stack frames. I would use less whitespace and add Separator widgets between each pair of file name and path.
May be something like what chrome devtools does?
Screen Shot 2021-03-30 at 1 39 19 PM

packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
```diff
@@ -229,13 +226,15 @@ class AllocationTableViewState extends State<AllocationTableView>
           controller.searchMatchMonitorAllocationsNotifier,
     );
 
+    final ScrollController scroller = ScrollController();
```
jacob314, line 229: nit: call `scrollController` not `scroller` for consistency with other code in DevTools.

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -1030,6 +1030,36 @@ class MemoryController extends DisposableController
     return serviceManager?.vm?.operatingSystem == 'android';
   }
 
+  /// All library dependencies for the app.
+  List<LibraryRef> get activeLibraries =>
```
jacob314, line 1034: nit: remove this helper

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -1030,6 +1030,36 @@ class MemoryController extends DisposableController
...
+      serviceManager.isolateManager.selectedIsolateLibraries;
+
+  /// Source file name as returned from allocation's stacktrace.
+  /// Map source URI
+  ///    packages/flutter/src/widgets/image.dart
+  /// would map to
+  ///    package:flutter/lib/src/widgets/image.dart
+  // TODO(terry): Review with Ben pathing doesn't quite work the
+  //              source file has the lib/ maybe a LibraryRef could
+  //              be returned if in a library otherwise file:/// as today?
```
jacob314, line 1044: you mean in a package not in a library.

kenzieschmoll:
High level comment - should we use a vertical splitter to separate tracked instances from the allocation call stack or force the view to be split 50/50? My concern is that the count column will be hidden and not discoverable. Or we could make the "Tracking" column have a smaller width so that the "Count" column is more likely to be visible

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
```diff
@@ -31,14 +33,13 @@ class TrackerClass extends Tracker {
...
+  TrackerAllocation(this.timestamp, this.stacktrace)
+      : super._internal(name: null);
 
   final int timestamp;
 
-  bool displayAll = false;
-
-  /// If displayAll is false then depth of callstack is:
-  final callDepthDisplay = 4;
+  /// Souces and callstack for an instance.
```
kenzieschmoll, line 41: typo: sources

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
```diff
@@ -160,78 +151,142 @@ class TreeTracker {
...
-    // Add the entire call stack as leaf nodes.
-    final totalCallStack = item.stacktrace.length;
-    for (var index = leafNodesCount - 1; index < totalCallStack; index++) {
-      allocation.addChild(TrackerCall(item.stacktrace[index]));
-    }
-    allocation.expandCascading();
+  Widget _helpStacktraceScreen(BuildContext context) => Row(
+        mainAxisAlignment: MainAxisAlignment.center,
+        children: const [
+          Text('Select an instance of a tracked class to'
```
kenzieschmoll, line 187: does this overflow if the container size is very narrow? or does it just break to new lines?

packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
```diff
@@ -229,13 +226,15 @@ class AllocationTableViewState extends State<AllocationTableView>
...
+
     return Split(
       initialFractions: const [0.8, 0.2],
       minSizes: const [200, 0],
       axis: Axis.vertical,
       children: [
         controller.allocationsFieldsTable,
-        trackerData.createTrackingTable(context, controller),
+        trackerData.createTrackingTable(context, controller, scroller),
```
kenzieschmoll, line 237: instead of having this be a method on TreeTracker, create a widget class TrackingTable and use that here

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
```diff
@@ -160,78 +151,142 @@ class TreeTracker {
...
-
-    // Show the entire call stack for this instance.
-    allocation.displayAll = true;
-
-    // Re-construct the callstack for this instance.
-    final leafNodesCount = allocation.children.length;
-
-    assert(allocation.children.last is TrackerMore);
-    allocation.removeLastChild(); // Remove "More ..."
+  Widget _helpTrackScreen(BuildContext context) {
```
kenzieschmoll, line 164: rename to AllocationTrackingInstructions and make this a widget class instead of a method

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -1030,6 +1030,36 @@ class MemoryController extends DisposableController
...
+  static const packageName = '/packages/';
+  String displayAsPackage(String sourceName) {
+    final packagesIndex = sourceName.indexOf(packageName);
+    final restToMatchStartIndex = packagesIndex + packageName.length;
+    if (packagesIndex >= 0) {
+      final match = activeLibraries.firstWhere(
```
jacob314, line 1050: this seems slow and error prone. Can we just continue to display the full file path for now? Longer term we need to get an API in DDS that maps file paths to dart uris and vice versa.
Add a reference to 
https://github.com/dart-lang/sdk/issues/45530
to fix this code once there is a DDS api for this.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
```diff
@@ -160,78 +151,142 @@ class TreeTracker {
...
+    );
+  }
 
-    // Add the entire call stack as leaf nodes.
-    final totalCallStack = item.stacktrace.length;
-    for (var index = leafNodesCount - 1; index < totalCallStack; index++) {
-      allocation.addChild(TrackerCall(item.stacktrace[index]));
-    }
-    allocation.expandCascading();
+  Widget _helpStacktraceScreen(BuildContext context) => Row(
```
kenzieschmoll, line 184: rename to AllocationCallStackInstructions and make this a widget class instead of a method

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
```diff
@@ -160,78 +151,142 @@ class TreeTracker {
...
+            controller: scroller,
+            child: ListView.builder(
+              controller: scroller,
+              itemCount: callstackLength,
+              itemBuilder: (context, index) {
+                return ListTile(
+                  contentPadding: EdgeInsets.zero,
+                  dense: true,
+                  title: Text('#$index ${stackTrace[index]}'),
+                  //subtitle: Text(controller.displayAsPackage(sources[index])),
```
kenzieschmoll, line 235: remove

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
```diff
@@ -160,78 +151,142 @@ class TreeTracker {
...
-    allocation.expandCascading();
+  Widget _helpStacktraceScreen(BuildContext context) => Row(
+        mainAxisAlignment: MainAxisAlignment.center,
+        children: const [
+          Text('Select an instance of a tracked class to'
+              ' view its allocation stack trace.'),
+        ],
+      );
+
+  Widget displaySelectedStackTrace(
```
kenzieschmoll, line 192: Same here. Make this a widget class instead of a method.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
```diff
@@ -160,78 +151,142 @@ class TreeTracker {
...
-    final totalCallStack = item.stacktrace.length;
-    for (var index = leafNodesCount - 1; index < totalCallStack; index++) {
-      allocation.addChild(TrackerCall(item.stacktrace[index]));
-    }
-    allocation.expandCascading();
+  Widget _helpStacktraceScreen(BuildContext context) => Row(
+        mainAxisAlignment: MainAxisAlignment.center,
+        children: const [
+          Text('Select an instance of a tracked class to'
+              ' view its allocation stack trace.'),
```
jacob314, line 188: Nit: add trailing comma after `'` to make this a little easier to read.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
```diff
@@ -160,78 +151,142 @@ class TreeTracker {
...
+    final callstackLength = stackTrace.length;
+    assert(sources.length == callstackLength);
+
+    final titleBackground = titleSolidBackgroundColor(Theme.of(context));
+
+    return Column(
+      children: [
+        SizedBox(
+          width: double.infinity,
+          height: defaultRowHeight + 4, // From table.dart
```
jacob314, line 211: see high level comment about too much spacing.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
```diff
@@ -160,78 +151,142 @@ class TreeTracker {
...
+              )),
+        ),
+        Expanded(
+          child: Scrollbar(
+            isAlwaysShown: true,
+            controller: scroller,
+            child: ListView.builder(
+              controller: scroller,
+              itemCount: callstackLength,
+              itemBuilder: (context, index) {
```
jacob314, line 230: there is a named parameter to ListView to add separators between items. probably use it to add a line between each stack frame

terrylucas:
> High level: there is a lot of whitespace between stack frames. I would use less whitespace and add Separator widgets between each pair of file name and path.
> May be something like what chrome devtools does?
> Screen Shot 2021-03-30 at 1 39 19 PM

Tried using ListTile but it's impossible to remove all the spacing (dense true and padding zero) maybe for touch gestures? Instead went to a Column with two left align Text children.
image

terrylucas:
> High level: there is a lot of whitespace between stack frames. I would use less whitespace and add Separator widgets between each pair of file name and path.
> May be something like what chrome devtools does?
> Screen Shot 2021-03-30 at 1 39 19 PM

Done

image

packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_controller.dart
terrylucas: Packages are libraryRefs in the vm service packages are exposed in selectedIsolateLibraries. But I'll change the name in the comment to say packages.

terrylucas:
```
  // TODO(terry): Review with Ben pathing doesn't quite work the source
  //              file has the lib/ maybe a LibraryRef could be returned
  //              if in a package today all packages are file:///?
```

terrylucas: Done.

terrylucas:
> High level comment - should we use a vertical splitter to separate tracked instances from the allocation call stack or force the view to be split 50/50? My concern is that the count column will be hidden and not discoverable. Or we could make the "Tracking" column have a smaller width so that the "Count" column is more likely to be visible

I think 50/50 will now work (now that I've fixed the size of the call stack displayed not using ListTile seems better).

image

packages/devtools_app/lib/src/memory/memory_controller.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
terrylucas: Wrapped in a Column unlikely to but this will ensure.  I think there are other places in DevTools where there is other help text that is in a Row which would be a problem with wrapping.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
terrylucas: Done.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
terrylucas: Already removed.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
terrylucas: This is the title bar to match the Table's header height.  The contents of the list have been compressed by eliminating ListTile.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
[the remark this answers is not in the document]
terrylucas: I must have beat you and Kenzie to this is definitely gone.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
terrylucas: I used List.separator.

packages/devtools_app/lib/src/memory/memory_allocation_table_view.dart
terrylucas: I'll put TODOs for this and the other function to be a class and move to a new file outside of MemoryTrackerModel

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
terrylucas: Added TODO will fix post PR.

terrylucas:
PTAL updated.

terrylucas:
> New screenshots look better although seems like there is now a little too little whitespace between the lines.
> Did you mean to push another version of this pull request with updated code?

I did add 4 more pixels (had it at 8) its at 12 now.  I think it looks better the default was 16 which was too much.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
```diff
@@ -248,7 +248,7 @@ class TreeTracker {
                 ],
               ),
               separatorBuilder: (context, index) =>
-                  Divider(color: devtoolsGrey[400], height: 8),
```
jacob314: nit: make 12 a constant in the list of devtools spacing constants. Could be defaultSpacerHeight or similar.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
```diff
@@ -224,18 +226,29 @@ class TreeTracker {
...
-                  contentPadding: EdgeInsets.zero,
-                  dense: true,
-                  title: Text('#$index ${stackTrace[index]}'),
-                  //subtitle: Text(controller.displayAsPackage(sources[index])),
-                  subtitle: Text(sources[index]),
-                );
-              },
+              itemBuilder: (context, index) => Column(
+                children: [
+                  Align(
```
jacob314, line 237: you can tweak the Flex properties on the Column instead of using Align here.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
terrylucas: Done.

PR #2851, opened 2021-04-01
Add tab for inspecting the state of package:provider

rrousselGit:


packages/devtools_app/lib/src/provider/provider_list.dart
```diff
@@ -0,0 +1,213 @@
+// Copyright 2021 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
```
jacob314, line 8: how much functionality are you using from riverpod? Unless it is extensive it could make sense to avoid the dependency for now.

packages/devtools_app/lib/src/provider/provider_screen.dart
```diff
@@ -0,0 +1,115 @@
...
+  provider.Provider.of<BannerMessagesController>(
+    context,
+    listen: false,
+  ).addMessage(
+    const ProviderUnknownErrorBanner(screenId: ProviderScreen.id)
+        .build(context),
+  );
+}
+
+class _SplitBorder extends StatelessWidget {
```
jacob314, line 101: Can you attach an updated screenshot to the CL?
I'm interested in seeing how this widget looks as part of the UI.

packages/devtools_app/pubspec.yaml
```diff
@@ -123,12 +121,12 @@ flutter:
...
+# The '#OVERRIDE_FOR_DEVELOPMENT' lines are stripped out when we publish.
 # All overriden dependencies are published together so there is no harm
 # in treating them like they are part of a mono-repo while developing.
   devtools_server: #OVERRIDE_FOR_DEVELOPMENT
-    path: ../devtools_server  #OVERRIDE_FOR_DEVELOPMENT
-  devtools_shared:  #OVERRIDE_FOR_DEVELOPMENT
-    path: ../devtools_shared  #OVERRIDE_FOR_DEVELOPMENT
-  devtools_testing:  #OVERRIDE_FOR_DEVELOPMENT
-    path: ../devtools_testing  #OVERRIDE_FOR_DEVELOPMENT
+    path: ../devtools_server #OVERRIDE_FOR_DEVELOPMENT
```
jacob314, line 128: We'll need to verify that the 
too/publish.sh script still strips these lines out. Otherwise this cleanup is fine.

packages/devtools_app/lib/src/provider/provider_screen.dart
rrousselGit: Screenshot 2021-04-01 at 03 43 03

packages/devtools_app/test/support/utils.dart
```diff
@@ -201,3 +201,49 @@ void _mockFlutterAssets() {
     return Future<ByteData>.value(encoded.buffer.asByteData());
   });
 }
+
+/// Load fonts used by the devtool for golden-tests to use them
```
jacob314, line 205: Thanks for adding this!

packages/devtools_app/lib/src/provider/instance_viewer/instance_providers.dart
```diff
@@ -0,0 +1,492 @@
...
+  AutoDisposeProviderReference ref,
+  InstancePath path,
+) async {
+  return path.pathToProperty.isNotEmpty
+      ? await ref.watch(rawInstanceProvider(path.parent).future)
+      : null;
+}
+
+/// Public properties first, then sort alphabetically
+int _sortFieldsByName(ObjectField a, ObjectField b) {
```
jacob314, line 178: Move this to utils and apply it to the debugger as well to keep field order consistent.

packages/devtools_app/lib/src/provider/instance_viewer/instance_providers.dart
```diff
@@ -0,0 +1,492 @@
+// Copyright 2021 The Chromium Authors. All rights reserved.
```
jacob314, line 1: Add a TODO that we should merge this code with the basic debugger variable view. This code is in many ways better than the existing debugger view. Having two inconsistent ways of showing debugger object views will be a bit confusing for users.

packages/devtools_app/lib/src/provider/provider_list.dart
rrousselGit: It's used extensively
With it, the implementation of the instance viewer is fully declarative, and it takes care of the process of canceling pending evaluations whenever possible.

jacob314: Interesting. Can you point to the pattern that ensures evaluations are cancelled?

packages/devtools_app/lib/src/provider/provider_screen.dart
```diff
@@ -0,0 +1,115 @@
...
+  );
+
+  return instance is AsyncError;
+});
+
+class ProviderScreen extends Screen {
+  const ProviderScreen()
+      : super.conditional(
+          id: id,
+          requiresLibrary: 'package:provider/',
```
jacob314, line 35: I assume this doesn't work on all versions of package:provider. What versions does it work on and what message do we show if it is not supported?
You could use EvalOnDartLibrary to verify the provider library has a required method only available on versions you support.

packages/devtools_app/lib/src/provider/provider_screen.dart
jacob314: Can you expand out a list and a map in the screenshot and show a screenshot where the hashcodes are fixed or omitted if they are -1.
Hashcodes should also be truncated 5 char hex to match what Flutter does in its debug print.

jacob314: I don't see the split border in the screenshot. I would have expected to see a border that was the focus color.

packages/devtools_app/lib/src/provider/provider_screen.dart
rrousselGit: The version supported is ^5.0.0

There is a banner that shows if any evaluation related to `package:provider` throws an error, which suggests to use ^5.0.0

packages/devtools_app/lib/src/provider/provider_list.dart
rrousselGit: It's the "autoDispose" functionality on providers (which `package:provider` sadly cannot support)
https://riverpod.dev/docs/concepts/modifiers/auto_dispose

It keeps track of the number of listeners on the associated provider (be it widgets or other providers (such as when chaining providers)). And when all listeners are removed, the provider state will automatically be destroyed.

Then providers can dispose a dispose method to cancel pending requests. That's the "Disposable" object which all providers use:

```dart
AutoDisposeProvider((ref) {
  final isAlive = Disposable();
  ref.onDispose(isAlive.dispose);
  
  
  // when the provider is destroyed, `isAlive` will be disposed, canceling the pending requests.
  await eval.eval('...', isAlive: isAlive);
})
```


This combines nicely with the collapse functionality too.
All object properties are associated with a unique provider. So when collapsing a property, its listener is removed, its state destroyed and pending requests canceled.

rrousselGit: This architecture also automatically deals with loading and error handling.

It relies on the fact that using `await` throws when the awaited future fails, purposefully letting the exception bubble. Then the provider will catch the exception, and use a union-type to force the UI to deal with loading/errors.

That's the `when(loading: ..., error: ..., data: ...)` you can see inside widgets.

packages/devtools_app/lib/src/provider/provider_screen.dart
rrousselGit: The split border is the thin grey border

But maybe `focusColor` is not the correct variable to use

rrousselGit: Screenshot 2021-04-01 at 04 35 19

nt4f04uNd:
a view from the side: provider is not an official package, wouldn't it make more sense to make the devtools extension system first and then build the provider devtools around that?

packages/devtools_app/lib/src/provider/instance_viewer/instance_providers.dart
rrousselGit: Done

rrousselGit:
Should I skip goldens? Since they generate 0.01% diff, probably due to environment differences.

rrousselGit:
Hum interesting that this time they passed

kenzieschmoll:
coming in kind of late here but I thought we agreed to not use `package:freezed` because that isn't a package we want to have rolled into g3 (the preferred solution in g3 is using built_value)? @jacob314 @rrousselGit

rrousselGit:
After discussing with @jacob314 , we agreed to remove the dependency and commit the generated file

Those `@freezed` are fake annotations, defined as `const freezed = Object()`.
They exist to easily re-generate the files

kenzieschmoll:
Will the .freezed files ever need  to be regenerated? Perhaps we could include a comment at the top of the .freezed files that describes how to regenerate these files if necessary

rrousselGit:
There is one here: https://github.com/flutter/devtools/blob/5e81d8202709caa3ca8988ef2522aab7ee06483b/packages/devtools_app/lib/src/provider/instance_viewer/instance_details.dart#L16

packages/devtools_app/lib/src/provider/provider_screen.dart
```diff
@@ -0,0 +1,115 @@
...
+                Expanded(
+                  child: _SplitBorder(
+                    child: InstanceViewer(
+                      rootPath: InstancePath.fromProviderId(selectedProviderId),
+                    ),
+                  ),
+                ),
+              ] else
+                const Expanded(
+                  child: _SplitBorder(child: SizedBox.expand()),
```
kenzieschmoll, line 81: Then if there is not a provider selected, the title should be '[No provider selected]' or similar. This will match what we do in the performance page:

Screen Shot 2021-04-14 at 12 58 22 PM
Screen Shot 2021-04-14 at 1 01 28 PM

packages/devtools_app/lib/src/provider/provider_list.dart
```diff
@@ -0,0 +1,213 @@
...
+        isSelected ? colorScheme.selectedRowBackgroundColor : null;
+
+    return GestureDetector(
+      behavior: HitTestBehavior.opaque,
+      onTap: () => context.read(selectedProviderIdProvider).state = providerId,
+      child: Container(
+        color: backgroundColor,
+        padding: _tilePadding,
+        child: state.when(
+          loading: () => const CenteredCircularProgressIndicator(),
```
kenzieschmoll, line 206: can we get rid of these spinners? It looks like there is a single spinner for the whole container, and then there is a spinner for each row. It created a kind of strange multi-spinner effect that we probably don't want:
Screen Shot 2021-04-14 at 12 52 31 PM

kenzieschmoll (commented):
can we sort the list of providers alphabetically? If you are looking for a specific provider, it can be challenging to find what you are looking for at first glance
Screen Shot 2021-04-14 at 1 45 01 PM

kenzieschmoll:
Looking good! a few UI requests and then LGTM

packages/devtools_app/lib/src/provider/provider_screen.dart
```diff
@@ -0,0 +1,115 @@
...
+});
+
+class ProviderScreen extends Screen {
+  const ProviderScreen()
+      : super.conditional(
+          id: id,
+          requiresLibrary: 'package:provider/',
+          title: 'Provider',
+          requiresDebugBuild: true,
+          icon: Icons.palette,
```
kenzieschmoll, line 38: feel free to change the icon - not sure if this once was listed in the conditional screens example by default.

jacob314:
lgtm
once the few polish comments @kenzieschmoll mentioned are addressed.

packages/devtools_app/lib/src/provider/provider_screen.dart
rrousselGit: Done, switched to attach_file

packages/devtools_app/lib/src/provider/provider_list.dart
rrousselGit: Should be good now

jacob314 (approved):
lgtm
As soon as the tests pass lets land this.

rrousselGit:
This should be good to go now

PR #2854, opened 2021-04-01
Add ListValueNotifier class

kenzieschmoll:
When you have a `ValueNotifier` that is a list, modifying that list does not notify listeners. We have gotten around this limitation in the past by copying the list and resetting the value of the change notifier (e.g. `notifier.value = List.from(notifier.value)`). This is extra work that is eliminated by the `ListValueNotifier` class, which provides more flexibility with ValueNotifiers for lists.

This PR adds a class `ListValueNotifier` that exposes methods to modify elements in the list and notify listeners. `notifyListeners` is protected, so it cannot be called by classes that are not subclasses of `ChangeNotifier`


kenzieschmoll:
I am using this class in a larger PR - breaking up to land this util class separately.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1069,3 +1069,54 @@ String prettyTimestamp(
...
+
+/// A [ChangeNotifier] that holds a list of data.
+///
+/// When [value] is replaced with something that is not equal to the old
+/// value as evaluated by the equality operator ==, this class notifies its
+/// listeners.
+///
+/// This class also exposes methods `addAndNotify` and `addAllAndNotify` that
+/// will notify listeners upon the list [value] being modified.
+class ListValueNotifier<T extends List<V>, V> extends ChangeNotifier
```
jacob314, line 1081: I would recommend removing T. and list make the value be List<V>

kenzieschmoll: restructured PTAL

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1069,3 +1069,54 @@ String prettyTimestamp(
...
+  set value(T newValue) {
+    if (_value == newValue) {
+      return;
+    }
+    _value = newValue;
+    notifyListeners();
+  }
+
+  void addAndNotify(V item) {
+    _value.add(item);
```
jacob314: nit: why not just call these `add` and `addAll`? The other ValueListenable apis don't include the words "andNotify"?

kenzieschmoll: done

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1069,3 +1069,54 @@ String prettyTimestamp(
...
+  T _value;
+  set value(T newValue) {
+    if (_value == newValue) {
+      return;
+    }
+    _value = newValue;
+    notifyListeners();
+  }
+
+  void addAndNotify(V item) {
```
jacob314: my concern with this API is it will break any downstream Widget code that checks
widget.oldList == widget.newList
as the value is the same but is mutated. The way around this is to create cheap views of  the underlying List objects with the correct # of elements.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1069,3 +1069,54 @@ String prettyTimestamp(
...
+
+  /// The current value stored in this notifier.
+  ///
+  /// When the value is replaced with something that is not equal to the old
+  /// value as evaluated by the equality operator ==, this class notifies its
+  /// listeners.
+  @override
+  T get value => _value;
+  T _value;
+  set value(T newValue) {
```
jacob314: probably avoid providing a set value method here as it will be tricky to make work efficiently with the other apis. You might instead provide a clear() method.

jacob314: or add 
```dart
void initialize(List<V>  elements) {
}
```
to make it clear the the actual value returned will be a view of the list rather than an identical object.

kenzieschmoll: this was removed before your review - may want to peek at the updated bits. I changed the class to extend ValueNotifier instead of ChangeNotifier, so this copied override was removed.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1069,3 +1069,34 @@ String prettyTimestamp(
...
+  void add(T item) {
+    value.add(item);
+    notifyListeners();
+  }
+
+  void addAll(List<T> itemList) {
+    value.addAll(itemList);
+    notifyListeners();
+  }
+
```
terrylucas, line 1094: clear() would be useful (especially when user clear all all collected memory/snapshot data) as would removeAt() might be too other removes I'd add if needed later.

packages/devtools_app/test/utils_test.dart
```diff
@@ -911,6 +911,57 @@ void main() {
...
+        notifier.value = [1, 2];
+        expect(didNotify, isTrue);
+        expect(notifier.value, equals([1, 2]));
+
+        didNotify = false;
+        notifier.forceNotify();
+        expect(didNotify, isTrue);
+        expect(notifier.value, equals([1, 2]));
+      });
+    });
```
terrylucas, line 964: Maybe a clear if added to base class and remove/forceNotify?

terrylucas:
Exciting to see (I'll have to use once checked in).

packages/devtools_app/lib/src/utils.dart
kenzieschmoll: This would be the same problem for a plain `List<object>`, correct? dart:collection provides collection equality functions that can be used if two collections should be compared for content.

packages/devtools_app/test/utils_test.dart
kenzieschmoll: added clear. forceNotify is still useful for the case where a list's contents were updated and add or addAll could not be used

packages/devtools_app/lib/src/utils.dart
kenzieschmoll: added clear

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1069,3 +1069,225 @@ String prettyTimestamp(
...
+    _currentList = ImmutableList(_rawList);
+  }
+
+  final List<T> _rawList;
+
+  ImmutableList<T> _currentList;
+
+  @override
+  List<T> get value => _currentList;
+  set value(List<T> newValue) {
```
jacob314, line 1091: implement https://api.flutter.dev/flutter/foundation/ValueNotifier-class.html instead of implementing the ValueListenable API and then throwing.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1069,3 +1069,225 @@ String prettyTimestamp(
...
+
+  void addAll(List<T> itemList, {bool notify = true}) {
+    _rawList.addAll(itemList);
+    _currentList = ImmutableList(_rawList);
+    if (notify) {
+      notifyListeners();
+    }
+  }
+
+  bool remove(T element, {bool notify = true}) {
```
jacob314, line 1114: these remove methods are wrong. They will mess up the previous immutable values. Remove these methods as there isn't an easy way to support them efficiently. The right way to do it would be to copy the _rawList object but no point unless you have a use case.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1069,3 +1069,225 @@ String prettyTimestamp(
...
+  T removeAt(int index, {bool notify = true}) {
+    final element = _rawList.removeAt(index);
+    _currentList = ImmutableList(_rawList);
+    if (notify) {
+      notifyListeners();
+    }
+    return element;
+  }
+
+  void removeWhere(bool test(T element), {bool notify = true}) {
```
jacob314, line 1132: also wrong

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1069,3 +1069,225 @@ String prettyTimestamp(
...
+    _rawList.removeWhere(test);
+    _currentList = ImmutableList(_rawList);
+    if (notify) {
+      notifyListeners();
+    }
+  }
+
+  void insert(int index, T element, {bool notify = true}) {
```
jacob314, line 1140: also wrong/ remove

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1069,3 +1069,225 @@ String prettyTimestamp(
...
+
+  void insertAll(int index, Iterable<T> iterable, {bool notify = true}) {
+    _rawList.insertAll(index, iterable);
+    _currentList = ImmutableList(_rawList);
+    if (notify) {
+      notifyListeners();
+    }
+  }
+
+  void clear({bool notify = true}) {
```
jacob314, line 1156: this needs to set
_rawList = [];
instead of clearing. Otherwise you are breaking existing immutable values.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1069,3 +1069,225 @@ String prettyTimestamp(
...
+    _rawList.clear();
+    _currentList = ImmutableList(_rawList);
+    if (notify) {
+      notifyListeners();
+    }
+  }
+
+  void sort({int Function(T a, T b) compare, bool notify = true}) {
```
jacob314, line 1164: this is wrong. remove.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1069,3 +1069,225 @@ String prettyTimestamp(
...
+    if (notify) {
+      notifyListeners();
+    }
+  }
+
+  /// Forces notifying listeners with the current value.
+  ///
+  /// This should only be used if the list [value] cannot be modified with
+  /// [add] or [addAll].
+  void forceNotify() {
```
jacob314, line 1176: remove forceNotify as it isn't needed.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1069,3 +1069,225 @@ String prettyTimestamp(
...
+  ImmutableList(this._rawList) : length = _rawList.length;
+
+  final List<T> _rawList;
+
+  @override
+  int length;
+
+  @override
+  T operator [](int index) {
+    if (_rawList.length != length) {
```
jacob314, line 1195: Remove this case. you shouldn't throw just because the raw list added elements.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1069,3 +1069,225 @@ String prettyTimestamp(
...
+    } else if (index >= 0 && index < length) {
+      return _rawList[index];
+    } else {
+      throw Exception('Index out of range [0-${length - 1}]: $index');
+    }
+  }
+
+  @override
+  void operator []=(int index, T value) {
+    throw Exception('Cannot modify the content of _ImmutableList');
```
jacob314, line 1206: messages say _ImmutableList while code says ImmutableList.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1069,3 +1069,159 @@ String prettyTimestamp(
...
+  }
+
+  List<T> _rawList;
+
+  ImmutableList<T> _currentList;
+
+  @override
+  List<T> get value => _currentList;
+
+  void add(T item, {bool notify = true}) {
```
jacob314, line 1092: Add a comment that `notify=false` should rarely be used or remove it.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/utils.dart
kenzieschmoll: removed notify param

PR #2857, opened 2021-04-02
Fixed UI overflow and chart refresh cleanup.

terrylucas:
- Cleaned up some overflows which can happen.
- Cleaned up chart refresh to use a ValueNotifier.
- Fixed tests to refresh charts when data changes for golden comparisons.


packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -893,8 +896,8 @@ class MemoryBodyState extends State<MemoryBody>
   //                final displayData = encoder.convert(data);
   String longValueToShort(String longValue) {
     var value = longValue;
-    if (longValue.length > 35) {
-      final firstPart = longValue.substring(0, 10);
+    if (longValue.length > 34) {
+      final firstPart = longValue.substring(0, 9);
```
kenzieschmoll, line 900: can you put 34 and 9 into named consts so we know what they are for?

terrylucas: Done.

PR #2863, opened 2021-04-06
Fix spurious new lines

grouma:
We correctly remove trailing new lines based off the last element. However we had a stale reference for this logic which caused issues.


jacob314 (approved):
Can you add a unit test that that fails with this bug and passes with it fixed?
lgtm

Issue #2823, opened 2021-03-20
Header takes up too much space when the window is not maximized


grouma:
The tabs are rendered below the title which takes up a ton of space when the window is not maximized:
![W2c88u9McpGVkfH](https://user-images.githubusercontent.com/1840773/111852752-6c38b780-88d5-11eb-891a-2d7b9b891b75.png)

There is still plenty of space to keep the tabs on the same line so I'm not sure why this is happening.

grouma:
cc @kenzieschmoll

kenzieschmoll:
Our logic for whether to use the stacked scaffold (title and tabs) is hard coded to a specific px size. We should instead calculate using a layout builder or by an estimate based off of the number of tabs.

PR #2658, opened 2021-02-01
Use ButtonStyle.fixedSize in DevTools theme

kenzieschmoll:
Cleanup. No functional changes.

PR #2864, opened 2021-04-06
Improve scaffold UX for density

grouma:
- Cut the scaffold height in half
- Update the `isNarrow` logic so that is responsive to the number of tabs
- Fix an issue where the padding was invalid while `isNarrow`

Wide (but not full screen):
Screen Shot 2021-04-05 at 8 05 42 PM

Narrow:
Screen Shot 2021-04-05 at 8 23 27 PM

Closes https://github.com/flutter/devtools/issues/2823


jacob314:
The density of the toolbar and the buttons now look a bit inconsistent. I would recommend also densifying the other buttons so that the page looks consistent.

kenzieschmoll:
I just merged master into https://github.com/flutter/devtools/pull/2658, so hopefully the bots will be green now and we can land. Gary, once that lands, you'll have an easy place in the theme data to update fixed size for buttons.

packages/devtools_app/lib/src/scaffold.dart
```diff
@@ -375,8 +369,8 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
         tabs: [for (var screen in widget.tabs) screen.buildTab(context)],
       );
       preferredSize = isNarrow
-          ? const Size.fromHeight(kToolbarHeight + 40.0)
-          : const Size.fromHeight(kToolbarHeight);
+          ? const Size.fromHeight(kToolbarHeight / 2.0 + 40.0)
```
kenzieschmoll, line 372: instead of dividing by 2 everywhere, can we add a const to the top of this file that is kToolbarHeight / 2 and use that new const everywhere we currently have `kToolbarHeight / 2.0`

packages/devtools_app/lib/src/scaffold.dart
```diff
@@ -402,8 +396,14 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
     final appBar = AppBar(
       // Turn off the appbar's back button.
       automaticallyImplyLeading: false,
-      title: Text(title),
+      title: Text(
+        title,
+        style: const TextStyle(
+          fontSize: 14,
```
kenzieschmoll, line 402: instead of setting the font size, can we use one of the text styles from TextTheme? headline2, headline3, etc.

grouma: Done.

packages/devtools_app/lib/src/scaffold.dart
grouma: Done.

grouma:
Simply halving the `defaultButtonHeight` made things looked too cramped. I settled on `24`. See here:
Screen Shot 2021-04-06 at 10 16 13 AM

kenzieschmoll:
I almost think 24 for the toolbar size is too small. It feels a little crammed. 24.0 for the button height feels a little crammed too unless we shrink the font and icons too.

If you are wanting to get the toolbar fix in, maybe we can meet somewhere in the middle to start and iterate later. Perhaps we can try changing both the toolbar height and the default button height to 32.0, and include your fix to measure the tab widths.

packages/devtools_app/lib/src/scaffold.dart
```diff
@@ -402,8 +396,12 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
     final appBar = AppBar(
       // Turn off the appbar's back button.
       automaticallyImplyLeading: false,
-      title: Text(title),
+      title: Text(
+        title,
+        style: Theme.of(context).textTheme.subtitle1,
+      ),
       centerTitle: false,
+      toolbarHeight: kToolbarHeight / 2.0,
```
kenzieschmoll, line 404: use `defaultToolbarHeight` here too

grouma: Done.

terrylucas (approved):
LGTM - will follow up with new goldens.

PR #2872, opened 2021-04-07
Support passing keypresses up to IDEs to enable shortcut keys when embedded DevTools has focus

### LINKED ISSUE #2775: There is no way to exit widget inspector by keyboard in vscode
## Steps to reproduce
1. Debug any flutter project.
2. Open widget inspector in vscode.
3. Select any widget by click
4. Press Cmd+1, Cmd+Shift+P, F1, Cmd+W, etc.

## Expected behavior
Focus first editor group or open command pallete, close widget inspector.

## Actual behavior
Widget inspector consumes all keypress and nothing happens.

DanTup:
The `origin` for the VS Code web-view appears to include a random hash, so we have to pass `*` for `postMessage`, so this also checks that DevTools is connected to a VM Service and also that the user agent contains Electron. @jacob314 let me know if you think this is not sufficient (we could perhaps ping the server and see if it was started in embedded mode too).

@helin24 if you're embedding inside an iframe (probably not, though in VS Code it's required) in IntelliJ and also need to handle keypresses, we may need to add an additional user agent check.

Fixes #2775 (with an equiv change in VS Code).

### DIFF
--- a/packages/devtools_app/lib/src/config_specific/framework_initialize/_framework_initialize_web.dart
+++ b/packages/devtools_app/lib/src/config_specific/framework_initialize/_framework_initialize_web.dart
@@ -29,6 +29,8 @@ Future<String> initializePlatform() async {
   // handle in the app. This is a workaround for
   // https://github.com/flutter/flutter/issues/58119.
   window.onKeyDown.listen((event) {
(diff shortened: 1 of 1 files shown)


jacob314 (approved):
lgtm

PR #2873, opened 2021-04-08
Fix a couple bugs with flame chart styling and zoom.

kenzieschmoll:
Styling changes: Sections of the flame chart did not have a consistent height because odd-indexed sections appeared to have both padding on the top and bottom. This was due to the fact that we colored all spacer rows the same instead of making them take on the color of the section they belonged to. This PR fixes that.

Zoom bugs: when zooming the flame chart in as far as it will go, you could reach the point where each interval represented less than 1 microsecond. At this point, all the timestamps in the flame chart would go to 0.0ms because of rounding and the fact that microseconds are as fine grained as we have data for. Setting the max zoom level solves this issue.

### DIFF
--- a/packages/devtools_app/lib/src/charts/flame_chart.dart
+++ b/packages/devtools_app/lib/src/charts/flame_chart.dart
@@ -660,10 +660,10 @@ class ScrollingFlameChartRowState<V> extends State<ScrollingFlameChartRow>
   @override
   Widget build(BuildContext context) {
     if (nodes.isEmpty) {
-      return Container(
+      return EmptyFlameChartRow(
         height: sectionSpacing,
         width: widget.width,
-        color: widget.backgroundColor,
+        backgroundColor: widget.backgroundColor,
       );
     }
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -282,14 +289,36 @@ abstract class FlameChartState<T extends FlameChart, V> extends State<T>
       customPointerSignalHandler: _handlePointerSignal,
       childrenDelegate: SliverChildBuilderDelegate(
         (context, index) {
+          final nodes = rows[index].nodes;
+          Color rowBackgroundColor = Colors.transparent;
```
jacob314, line 293: nit. remove color tpe from LHS.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/charts/flame_chart.dart
kenzieschmoll: done

PR #2875, opened 2021-04-08
Support copying file in the debugger

grouma:
This comes from an internal request. Add  a simple copy button to copy the file's name. 

Screen Shot 2021-04-08 at 11 38 59 AM

### DIFF
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -7,6 +7,7 @@ import 'dart:math' as math;
 
 import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:provider/provider.dart';
 import 'package:vm_service/vm_service.dart' hide Stack;
 
@@ -347,6 +348,11 @@ class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
                   scriptRef?.uri ?? ' ',
                   style: theme.textTheme.subtitle2,
                 ),
+              ),
+              ToolbarAction(
+                icon: Icons.copy,
+                onPressed: () =>
+                    Clipboard.setData(ClipboardData(text: scriptRef?.uri)),
               ),
               const SizedBox(width: denseSpacing),
               PopupMenuButton<ScriptRef>(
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -348,6 +349,11 @@ class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
                   style: theme.textTheme.subtitle2,
                 ),
               ),
+              ToolbarAction(
```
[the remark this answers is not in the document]
kenzieschmoll: +1

kenzieschmoll (approved):
lgtm_ship_it

jacob314:
Why are we adding a toolbar icon instead of using SelectableText?

kenzieschmoll:
> Why are we adding a toolbar icon instead of using SelectableText?

Maybe we should do both. I do like the copy to clipboard button, but I also like the idea of having the text be selectable too.

grouma:
> > Why are we adding a toolbar icon instead of using SelectableText?
> 
> Maybe we should do both. I do like the copy to clipboard button, but I also like the idea of having the text be selectable too.

@jacob314 and I discussed offline. I think I'm going to go with a tripple dot icon which has a dropdown copy option. Eventually we'll add a bunch more actions like open in code search etc.

grouma:
I also found the chevron icon confusing so I updated the icon to be history and update the tooltip:

Screen Shot 2021-04-08 at 2 10 12 PM

kenzieschmoll:
Should we get some UX input on changing the caret to the history icon? IntelliJ, for example, uses a caret for their dropdown menu of files:
Screen Shot 2021-04-08 at 4 13 42 PM

at first glance, the history icon looks like a refresh button to me

grouma:
> uses a caret for their dropdown menu of files

 The caret made sense when it was the only dropdown. Sitting next to the triple dot icon makes it not vey informative. I find the history icon more descriptive. Regardless the tooltip text helps a lot.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -349,8 +350,36 @@ class _CodeViewState extends State<CodeView> with AutoDisposeMixin {
                 ),
               ),
               const SizedBox(width: denseSpacing),
+              PopupMenuButton<_ScriptPopupMenuOptions>(
```
kenzieschmoll, line 353: nit: this build method is getting pretty long. can you break these items into two helpers `_buildScriptOptionsPopupMenu` and `_buildScriptHistoryPopupMenu`

kenzieschmoll:
lgtm_army

packages/devtools_app/lib/src/debugger/codeview.dart
kenzieschmoll: better yet, two stateless widgets

jacob314:
lgtm

PR #2877, opened 2021-04-08
Auto expand search results

grouma:
Another simple feature request from an internal user. We only auto expand while searching as I feel it looks odd otherwise.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/scripts.dart
+++ b/packages/devtools_app/lib/src/debugger/scripts.dart
@@ -42,6 +42,8 @@ class ScriptPickerState extends State<ScriptPicker> {
   // TODO(devoncarew): How to retain the filter text state?
   final _filterController = TextEditingController();
 
+  final _maxAutoExpandChildCount = 3;
+
   List<ObjRef> _items = [];
   List<ObjRef> _filteredItems = [];
   List<FileNode> _rootScriptNodes;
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/debugger/scripts.dart
```diff
@@ -114,6 +116,13 @@ class ScriptPickerState extends State<ScriptPicker> {
           if (!_isLoading)
             Expanded(
               child: TreeView<FileNode>(
+                onTraverse: (node) {
+                  // Auto expand children when there are minimal search results.
+                  if (_filterController.text.isNotEmpty &&
+                      node.children.length <= _maxAutoExpandChildCount) {
+                    node.expand();
```
kenzieschmoll, line 123: what about this scenario:
- a user searches "foo" and there are three results - all three nodes are expanded
- then they backspace so the search query is "fo", and now there are 5 results

At this point, I would think the 3 results that also matched "foo" would be expanded while the 2 results that only match "fo" would be collapsed, which doesn't seem like what we want. Is this what happens? If so, we'll need to make sure previously expanded nodes are collapsed when the search query changes

grouma: We construct a new tree for each query so previous nodes are collapsed correctly.

kenzieschmoll (approved):

PR #2878, opened 2021-04-08
Add previous/next event in thread buttons to the Timeline

kenzieschmoll:
When the timeline has a lot of events, it can be difficult to see where events are because the width of each event is so small. Because of how much time can pass between events on any thread, it can also be challenging to know when the next event happens. These previous and next buttons help with this and let a user jump to the previous/next event on any thread.

Screen Shot 2021-04-08 at 1 31 41 PM

Previous buttons are shorter when a group is small, as to not overlap with the group label:
Screen Shot 2021-04-08 at 1 32 06 PM

@jacob314 @InMatrix

### DIFF
--- a/packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
+++ b/packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
@@ -298,6 +298,7 @@ class TimelineFlameChartState
       compare: (TimelineEvent a, TimelineEvent b) =>
           a.time.start.compareTo(b.time.start),
     );
+  }
 
   Future<void> _viewPreviousEventInGroup(TimelineEventGroup group) async {
     final firstInViewIndex = _indexOfFirstEventInView(group);
--- a/packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
+++ b/packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
@@ -14,7 +14,9 @@ import '../charts/flame_chart.dart';
 import '../common_widgets.dart';
 import '../flutter_widgets/linked_scroll_controller.dart';
 import '../geometry.dart';
+import '../notifications.dart';
 import '../theme.dart';
+import '../trace_event.dart';
 import '../ui/colors.dart';
 import '../ui/search.dart';
 import '../utils.dart';
@@ -218,6 +220,10 @@ class TimelineFlameChartState
 
   ScrollController _groupLabelScrollController;
 
+  ScrollController _previousInGroupButtonsScrollController;
+
+  ScrollController _nextInGroupButtonsScrollController;
+
   @override
   int get rowOffsetForTopPadding => TimelineFlameChart.rowOffsetForTopPadding;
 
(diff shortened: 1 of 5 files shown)


jacob314:
I think the arrows would be a little clearer if they matched the entire height of the block they are associated with rather than being 1 row high. Currently I thought they were associated with specific rows rather than the entire block.

packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
```diff
@@ -275,6 +284,77 @@ class TimelineFlameChartState
...
+    if (firstInViewIndex > 0) {
+      final event = group.sortedEventRoots[firstInViewIndex - 1];
+      await zoomToTimeRange(
+        startMicros: event.time.start.inMicroseconds,
+        durationMicros: event.time.duration.inMicroseconds,
+      );
+      await scrollHorizontallyToData(event);
+      return;
+    }
+    // This notification should not be shown if we are correctly disabling the
```
jacob314, line 310: if -> as
?

packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
```diff
@@ -548,6 +626,105 @@ class TimelineFlameChartState
...
+        Theme.of(context).colorScheme,
+      );
+      final backgroundWithOpacity = Color.fromRGBO(
+        backgroundColor.red,
+        backgroundColor.green,
+        backgroundColor.blue,
+        0.85,
+      );
+
+      prevInThreadChildren.add(PreviousInThreadButton(
```
jacob314, line 661: why isn't this using a lazy list view?

packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
```diff
@@ -1012,6 +1189,224 @@ class SelectedFrameBracketPainter extends FlameChartPainter {
       );
 }
 
+class PreviousInThreadButton extends StatefulWidget {
```
jacob314, line 1192: PreviousInThreadButton and NextInThreadButton are basically the same. Merge into one class.

packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
```diff
@@ -1012,6 +1189,224 @@ class SelectedFrameBracketPainter extends FlameChartPainter {
...
+    final topPadding =
+        !isNext && group.displayDepth == 2 ? topPaddingForMediumGroups : 0.0;
+    return Container(
+      margin: EdgeInsets.only(
+        top: topSpacer,
+        bottom: bottomSpacer,
+      ),
+      padding: EdgeInsets.only(
+        top: topPadding,
+        bottom: useSmallButton ? 2.0 : 0.0,
```
jacob314, line 1347: make 2.0 a constant in the theme?

packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
```diff
@@ -1012,6 +1189,224 @@ class SelectedFrameBracketPainter extends FlameChartPainter {
...
+
+  final Color backgroundColor;
+
+  final String tooltip;
+
+  final IconData icon;
+
+  final VoidCallback onPressed;
+
+  static const smallButtonHeight = 20.0;
```
jacob314, line 1384: move all button heights to the theme

jacob314 (approved):
lgtm

InMatrix:
Sorry, late to the party. This looks like a useful addition. I was thinking, maybe we can allow the user to hit the arrow keys on their keyboard to go to the next or previous events as well, once the thread is highlighted.

Issue #2882, opened 2021-04-09
Search bar in flame chart hogs focus


kenzieschmoll:
This prevents functions like w-a-s-d navigation from being enabled.

PR #2883, opened 2021-04-09
Fix focus management in timeline flame chart

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/2882.

Focus is now handled as expected. The flame chart now has focus by default. It releases focus to the search field when the search field is clicked, and takes focus back once the flame chart area is clicked.


jacob314 (approved):
lgtm

Issue #2879, opened 2021-04-09
Flame chart painters rendering at different horizontal scroll offsets


kenzieschmoll:
To fix this, we should have one listener to the horizontal offset notifier, and pass the offset into all the painters that need it. This will ensure that all painters are using the same horizontal scroll offset.

PR #2884, opened 2021-04-09
Use a single scroll offset for all flame chart painters instead of having them listen individually to notifiers

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/2879


jacob314 (approved):
lgtm

PR #2885, opened 2021-04-09
Use a different copy icon

grouma:
Use this icon for consistency with other usages inside Dart DevTools.


packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -752,7 +752,7 @@ class ScriptPopupMenu extends StatelessWidget {
                 Text('Copy filename',
                     style: Theme.of(context).regularTextStyle),
                 const Icon(
-                  Icons.copy,
+                  Icons.file_copy,
```
kenzieschmoll, line 755: can you use `Icons.content_copy` instead? this will match what we already use in the debugger console.

PR #2887, opened 2021-04-09
Add framework for internal features and add debugger menu options hook.

kenzieschmoll:
Verified that this code works as intended inside g3.


jacob314 (approved):
lgtm_dog

grouma (approved):

PR #2898, opened 2021-04-14
Fix "Count" text getting cut off when sorting

alexlindroos:
@kenzieschmoll @terrylucas 

Hi, here is a fix for "Count" text getting cut off in tracking memory allocations when sorting by count.

Screenshot 2021-04-14 at 12 09 38


packages/devtools_app/lib/src/memory/memory_tracker_model.dart
```diff
@@ -97,7 +97,7 @@ class _TrackerCountColumn extends ColumnData<Tracker> {
   _TrackerCountColumn()
       : super(
           'Count',
-          alignment: ColumnAlignment.right,
+          alignment: ColumnAlignment.left,
```
terrylucas, line 100: Is this correct?

```
0
10
200
5000
```

not

```
   0
  10
 200
5000
```

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
```diff
@@ -112,7 +112,7 @@ class _TrackerCountColumn extends ColumnData<Tracker> {
   }
 
   @override
-  double get fixedWidthPx => 50.0;
+  double get fixedWidthPx => 300.0;
```
terrylucas, line 115: This should match the above constructor 100  value.  If set to 100 then the width is wide enough.

packages/devtools_app/lib/src/memory/memory_tracker_model.dart
alexlindroos: I changed it, because I thought there is more space on right. Other parameters uses .right, so changing it back to make it more consistent

terrylucas (approved):
LGTM, thanks

terrylucas:
Thanks Alex I'll merge your PR and update the goldens.

Issue #2844, opened 2021-03-29
Support jump to line from the sources panel in the debugger.


jacob314:
We should match the keyboard shortcut for jump to line users are familiar with.
The default in VSCode is control-g 
Command + / + g is also reasonable on mac.
There may be a conflict with contrl-g / command-g in web browsers

grouma:
Alternatively we could add an option in the newly created drop down menu. I think that's more discoverable.

jacob314:
Is there any other editor that puts jump to line on that menu? VSCode, IntelliJ, and Chrome DevTools don't. I'm open to putting the item in that menu short term to improve discoverability for the keyboard shortcut but long term it would be good to keep that menu clean and focused on operations related to manipulating the files open in DevTools rather than manipulating a file.

grouma:
So it looks like Chrome uses the general search box for this functionality. Users should enter `:<line-number>`. I think at first we can have the drop down option then eventually when we support general text search we can merge the functionality like in Chrome.

PR #2902, opened 2021-04-14
Go To Line Number Option

grouma:
Add a dialog that accepts a line number. Upon enter the code will scroll to that position (or to the end if the number is larger than the max number of lines). Also wire up a keyboard shortcut to open the dialog.

UI:
Screen Shot 2021-04-14 at 2 10 14 PM

Closes https://github.com/flutter/devtools/issues/2844


kenzieschmoll:
does this handle bad input like letters and or negative numbers?

grouma:
> does this handle bad input like letters and or negative numbers?

The input is restricted so you can't even type non digit numbers.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -808,12 +809,67 @@ class ScriptPopupMenuOption {
...
+      content: Column(
+        mainAxisSize: MainAxisSize.min,
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          TextField(
+            autofocus: true,
+            onSubmitted: (value) {
+              if (value.isNotEmpty) {
+                Navigator.of(context, rootNavigator: true).pop('dialog');
+                final line = int.parse(value);
```
kenzieschmoll, line 853: we should fail gracefully if int.parse fails or if the value is a negative number.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -808,12 +809,67 @@ class ScriptPopupMenuOption {
...
+                _debuggerController.showScriptLocation(
+                  ScriptLocation(
+                    _debuggerController.scriptLocation.value.scriptRef,
+                    location: SourcePosition(line: line, column: 0),
```
kenzieschmoll, line 857: perhaps we should support `line:column`. Would be pretty easy to add and would match what IntelliJ supports for go to

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -808,12 +809,67 @@ class ScriptPopupMenuOption {
...
+                  ),
+                );
+              }
+            },
+            decoration: const InputDecoration(labelText: 'Line Number'),
+            keyboardType: TextInputType.number,
+            inputFormatters: <TextInputFormatter>[
+              FilteringTextInputFormatter.digitsOnly
```
kenzieschmoll, line 865: I see this now - looks like this should handle bad input.

kenzieschmoll (approved):

packages/devtools_app/lib/src/debugger/codeview.dart
grouma: We only scroll the line into view we don't adjust your cursor so supporting column doesn't make sense. We'd have to rework the code view to support that.

packages/devtools_app/lib/src/debugger/codeview.dart
grouma: It can't be a negative number due to the limited input.

kenzieschmoll:
FYI you're probably blocked on https://github.com/flutter/devtools/pull/2903

PR #2905, opened 2021-04-14
Fix issue where first build of the VM Tools status bar would cause a null pointer exception

bkonyi:
`VMDeveloperToolsScreen` exposes a `ValueListenable` provided by the `VMDeveloperToolsController` in order to control the visibility of the isolate selector. Given that `VMDeveloperToolsScreen.showIsolateSelector` can be invoked before the static controller in `_VMDeveloperToolsScreenState` is initialized, a null pointer exception was being thrown when first building the VM Tools screen.

This change updates `DevToolsScreen` to take either a `createController` callback or a pre-instantiated `controller`, allowing for the same instance of `VMDeveloperToolsController` to be used by `VMDeveloperToolsScreen` while also being exposed via `Provider`.

### DIFF
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -273,7 +273,7 @@ class DevToolsAppState extends State<DevToolsApp> {
   Widget _providedControllers({@required Widget child, bool offline = false}) {
     final _providers = widget.screens
         .where((s) =>
-            s.createController != null && (offline ? s.supportsOffline : true))
+            s.providesController && (offline ? s.supportsOffline : true))
         .map((s) => s.controllerProvider)
         .toList();
 
@@ -303,7 +303,8 @@ class DevToolsAppState extends State<DevToolsApp> {
 class DevToolsScreen<C> {
   const DevToolsScreen(
     this.screen, {
-    @required this.createController,
+    this.createController,
+    this.controller,
     this.supportsOffline = false,
   });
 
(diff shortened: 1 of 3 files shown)


bkonyi:
I'm not 100% sure this is the right approach here, but it does fix the issue. The other approach I tried involved statically initializing the controller in `_VMDeveloperToolsScreenState` and not initializing the controller via `Provider`, which also worked.

kenzieschmoll:
discussed offline - going to try a different approach that doesn't require adding the `ensureStateInitialized` method

bkonyi:
Updated with new approach (and was able to drop the static controller in `_VMDeveloperToolsScreenState`, hurrah!). PTAL.

packages/devtools_app/lib/src/app.dart
```diff
@@ -315,17 +316,35 @@ class DevToolsScreen<C> {
...
+
   /// Whether this screen has implemented offline support.
   ///
   /// Defaults to false.
   final bool supportsOffline;
 
   Provider<C> get controllerProvider {
-    assert(createController != null);
+    assert((createController != null && controller == null) ||
+        (createController == null && controller != null));
```
kenzieschmoll, line 344: Actually, I think we should allow both to be null based on the comment above `providesController`

kenzieschmoll: so this can just be `assert(createController == null || controller == null)` and instead of doing this here, do this at the constructor.

packages/devtools_app/lib/src/status_line.dart
```diff
@@ -47,6 +47,9 @@ class StatusLine extends StatelessWidget {
 
     // Optionally display an isolate selector.
     if (currentScreen != null) {
+      // Ensure that the current screen's state is initialized. In particular,
+      // make sure that the screen's controller is available before calling
+      // `currentScreen.showIsolateSelector`.
```
kenzieschmoll, line 52: remove

packages/devtools_app/lib/src/app.dart
bkonyi: Should we be calling `controllerProvider` when there's no controller provided? In that case there's nothing to create a `Provider` for.

kenzieschmoll (approved):
approved with a few comments

packages/devtools_app/lib/src/app.dart
kenzieschmoll: good point. We can leave this as is then. I still think adding `assert(createController == null || controller == null)` at the constructor level would be a good idea so it is clear from the start that you should not be providing both a `controller` and `createController` param.

bkonyi: Done.

kenzieschmoll (approved):

kenzieschmoll:
Red bots caused by https://github.com/flutter/devtools/issues/2916

bkonyi:
Landing on red because flaky bots are frustrating.

PR #2909, opened 2021-04-15
WIP: Re-register extensions with new isolate id after hot restart

annagrin:


packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -809,7 +813,7 @@ class ServiceExtensionManager extends Disposer {
   }
 
   Future<void> _maybeAddServiceExtension(String name) async {
-    if (_firstFrameEventReceived || !isUnsafeBeforeFirstFlutterFrame(name)) {
+    if (_firstFrameEventReceived /*|| !isUnsafeBeforeFirstFlutterFrame(name)*/) {
```
kenzieschmoll, line 816: why remove the check to `isUnsafeBeforeFirstFlutterFrame`?

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -650,6 +651,9 @@ class ServiceExtensionManager extends Disposer {
   ConnectedApp _connectedApp;
 
   Future<void> _handleIsolateEvent(Event event) async {
+    if (event.kind == EventKind.kIsolateExit) {
```
kenzieschmoll, line 654: will we receive this event for more isolates than the main isolate if a user is using more than one?

annagrin: Yes, definitely. I am not sure how to fix the actual issue...

packages/devtools_app/lib/src/service_manager.dart
annagrin: This change is more for illustrating the problem than fixing it - it makes devtools work if paused on start and then hot-restarted. 

If the user connects devtools to a paused application and hot-restarts the app, `_maybeAddServiceExtension` does not wait for the isolate to finish restarting and proceeds to re-register the extensions with an old isolate id.

I am just forcing the `else` clause to execute here instead of `then` and add registrations as pending after isolate exit.

Does it seem like a bug in devtools, or is this something that dwds doing incorrectly?

jacob314: seems like a bug in DevTools.

annagrin:
Closing this as the failures seem to all in dwds.

Issue #2906, opened 2021-04-14
GC Events should be rendered in their respective threads in the Timeline flame chart


kenzieschmoll:
If all GC events are rendered in a collective "GC" event group, the duration events can overlap. Duration events are synchronous so there should not be overlap of two different event roots. This is only happening visually because we currently group all GC events from all threads. Rendering GC events in their respective thread will avoid this issue and will provide a more accurate picture of when a specific thread is experiencing a GC.

PR #2917, opened 2021-04-16
Render GC and non-ui/non-raster events in their respective thread groups

kenzieschmoll:
Example of a GC event that should have been rendered in the UI thread:
Screen Shot 2021-04-16 at 12 51 03 PM

Example of an event that was previous rendered in a broad category of "unknown", but now is rendered under its respective thread (this is how other trace viewers render events on different threads):
Screen Shot 2021-04-16 at 2 05 01 PM

Fixes https://github.com/flutter/devtools/issues/2906


jacob314 (approved):
lgtm

Issue #2766, opened 2021-03-05
When a frame is selected there's no way to unselect it


Hixie:
It would be great if, after selecting a frame, you could unselect it, e.g. by clicking anywhere that doesn't have a frame in the top view or generally clicking anywhere in the bottom view. That way the nice rounded corner highlights would go away when you're no longer focusing on one frame.

kenzieschmoll:
as of #2921 you can unselect a frame by clicking the already selected frame again. Unselecting based off of other clicks in the charts may cause some undesired behavior when scrolling the charts around. This adds the option for users to remove the selection, but does not force them into unselecting when they are navigating around charts.

PR #2921, opened 2021-04-19
Add ability to unselect frame in the performance page

kenzieschmoll:
unselect-frame
Fixes https://github.com/flutter/devtools/issues/2766

### DIFF
--- a/packages/devtools_app/lib/src/performance/performance_controller.dart
+++ b/packages/devtools_app/lib/src/performance/performance_controller.dart
@@ -191,7 +191,7 @@ class PerformanceController
   ValueListenable<double> get displayRefreshRate => _displayRefreshRate;
   final _displayRefreshRate = ValueNotifier<double>(defaultRefreshRate);
 
-  Future<void> selectFrame(FlutterFrame frame) async {
+  Future<void> toggleSelectedFrame(FlutterFrame frame) async {
     if (frame == null || data == null) {
       return;
     }
--- a/packages/devtools_app/lib/src/performance/performance_controller.dart
+++ b/packages/devtools_app/lib/src/performance/performance_controller.dart
@@ -192,9 +192,17 @@ class PerformanceController
   final _displayRefreshRate = ValueNotifier<double>(defaultRefreshRate);
 
   Future<void> selectFrame(FlutterFrame frame) async {
-    if (frame == null || data == null || data.selectedFrame == frame) {
+    if (frame == null || data == null) {
       return;
     }
+
+    // Unselect [frame] if is already selected.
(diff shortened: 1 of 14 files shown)


packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -192,9 +192,17 @@ class PerformanceController
   final _displayRefreshRate = ValueNotifier<double>(defaultRefreshRate);
 
   Future<void> selectFrame(FlutterFrame frame) async {
```
jacob314, line 194: method name is now misleading. Maybe it is now "clickFrame" or "toggleFrame"

packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
```diff
@@ -1036,7 +1036,12 @@ class SelectedFrameBracketPainter extends FlameChartPainter {
 
   @override
   void paint(Canvas canvas, Size size) {
-    if (selectedFrame == null) return;
+    if (selectedFrame == null) {
+      canvas
```
jacob314, line 1040: why this save and restore? that should be a no-op.

kenzieschmoll: you cant call `restore` on Canvas without first calling `save`. It throws an exception if you do

packages/devtools_app/lib/src/performance/performance_controller.dart
kenzieschmoll: changed

packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
jacob314: right but why are you doing a save + restore with nothing in the middle?
That seem like writing:
```dart
foo += 1;
foo -= 1;
```

kenzieschmoll: this is to ensure that the selected frame painter (the blue brackets that surround the frame's timeline events) disappears when a frame is unselected. We essentially need to repaint an empty scene for this custom painter.

jacob314 (approved):
lgtm
with the save+restore removed.

packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
kenzieschmoll: hmm - maybe this was a reloading issue (or operator error) when I saw this behavior earlier. Was able to remove the canvas calls and still get the desired effect.

Issue #2603, opened 2021-01-13
Decode and display images in the network profiler for HTTP/HTTPS responses


kenzieschmoll:

bkonyi:
As discussed offline, we've hit a limitation with the current timeline based HTTP profiling due to the 32KiB ring buffer being much too small to record HTTP response data. In theory, this could be worked around by using an endless buffer, but that's not something we'll want to do by default due to memory implications.

We'll likely want to move to a hybrid timeline/RPC based profiling approach, where we still log HTTP events to the timeline but don't include body data. These events would also be available via a service extension, in addition to request/response bodies for each HTTP request/response.

This will involve a decent amount of work, so it's unlikely this can be done before 2.0.

bkonyi:
Blocked on https://github.com/dart-lang/sdk/issues/44950.

PR #2922, opened 2021-04-20
Add network profiler response image preview

bkonyi:
Preview:

image

Fixes https://github.com/flutter/devtools/issues/2603

### DIFF
--- a/packages/devtools_app/lib/src/http/http_request_data.dart
+++ b/packages/devtools_app/lib/src/http/http_request_data.dart
@@ -102,7 +102,21 @@ abstract class HttpRequestData extends NetworkRequest {
     // "text/html" is the MIME for both .html and .htm, but since .htm comes
     // first alphabetically, `extensionFromMime` returns "htm". Other cases are
     // more unintuitive such as "text/plain" returning "conf".
-    return extensionFromMime(mime);
+    return _extensionFromMime(mime);
+  }
+
+  /// Extracts the extension from [mime], with overrides for shortened
+  /// extenstions of common types (e.g., jpe -> jpeg).
+  String _extensionFromMime(String mime) {
+
+    final extension = extensionFromMime(mime);
+    if (extension == 'jpe') {
+      return 'jpeg';
+    }
+    if (extension == 'htm') {
+      return 'html';
+    }
+    return extension;
   }
 
   /// Whether the request is safe to display in the UI.
--- a/packages/devtools_app/lib/src/http/http_request_data.dart
+++ b/packages/devtools_app/lib/src/http/http_request_data.dart
@@ -98,10 +98,6 @@ abstract class HttpRequestData extends NetworkRequest {
     if (mime.endsWith(']')) {
       mime = mime.substring(0, mime.length - 1);
     }
-    // TODO(kenz): consider special casing some extensions. For example,
-    // "text/html" is the MIME for both .html and .htm, but since .htm comes
-    // first alphabetically, `extensionFromMime` returns "htm". Other cases are
-    // more unintuitive such as "text/plain" returning "conf".
     return _extensionFromMime(mime);
   }
 
@@ -114,6 +110,10 @@ abstract class HttpRequestData extends NetworkRequest {
     }
     if (extension == 'htm') {
       return 'html';
+    }
+    // text/plain -> conf
+    if (extension == 'conf') {
+      return 'txt'; 
     }
     return extension;
   }
--- a/packages/devtools_app/lib/src/http/http_request_data.dart
+++ b/packages/devtools_app/lib/src/http/http_request_data.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 import 'dart:convert';
+import 'dart:typed_data';
 
 import 'package:flutter/foundation.dart';
 import 'package:mime/mime.dart';
@@ -593,6 +594,12 @@ class DartIOHttpRequestData extends HttpRequestData {
     } on FormatException {
       return '<binary data>';
     }
+  }
+
+  Uint8List get encodedResponse {
+    if (!_request.isResponseComplete) return null;
+    final fullRequest = _request as HttpProfileRequest;
+    return fullRequest.responseBody;
   }
 
   String _responseBody;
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/network/network_request_inspector_views.dart
```diff
@@ -151,6 +160,92 @@ class HttpResponseView extends StatelessWidget {
...
 
+class ImageResponseView extends StatelessWidget {
+  const ImageResponseView(this.data);
+
+  final DartIOHttpRequestData data;
+
+  @override
+  Widget build(BuildContext context) {
+    final img = image.decodeImage(data.encodedResponse);
+    // Format should be of the form: [image/$format]
```
kenzieschmoll, line 171: can you use quotes in this comment instead? I was (may still be) confused on whether the brackets are part of the expected form value

packages/devtools_app/lib/src/network/network_request_inspector_views.dart
```diff
@@ -151,6 +160,92 @@ class HttpResponseView extends StatelessWidget {
...
+    String key,
+    dynamic value,
+    BoxConstraints constraints,
+  ) {
+    return Container(
+      width: constraints.minWidth,
+      padding: _rowPadding,
+      child: Row(
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
```
kenzieschmoll, line 231: if you use `mainAxisSize: MainAxisSize.min` I think you should be able to get rid of the constraints that are passed around everywhere and remove the wrapping container

packages/devtools_app/lib/src/network/network_request_inspector_views.dart
```diff
@@ -151,6 +160,92 @@ class HttpResponseView extends StatelessWidget {
...
+          );
+        }),
+      ],
+    );
+  }
+
+  Widget _buildRow(
+    BuildContext context,
+    String key,
+    dynamic value,
```
kenzieschmoll, line 223: can this be a String instead of dynamic?

packages/devtools_app/lib/src/network/network_request_inspector_views.dart
```diff
@@ -151,6 +160,92 @@ class HttpResponseView extends StatelessWidget {
...
+        children: [
+          Text(
+            '$key: ',
+            style: Theme.of(context).textTheme.subtitle2,
+          ),
+          Expanded(
+            child: Text(
+              value,
+              overflow: TextOverflow.ellipsis,
+              maxLines: 5,
```
kenzieschmoll, line 240: Is this necessary? the values passed to _buildRow above seem relatively small and even if they weren't, would we want to cut the data off?

bkonyi: Probably not. Not sure how this got there in the first place... I just popped this old change off my stash, so my memory's a bit fuzzy.

packages/devtools_app/lib/src/network/network_request_inspector_views.dart
bkonyi: The brackets were part of the expected format (e.g., "[image/$format]"), but I've just removed it.

bkonyi:
Addressed comments. Looks like we were processing the mime incorrectly before, so now `package:mime` is giving us `jpe` for `image/jpeg` instead of `jpg` (see below). Just figured it'd be worth pointing out.

Screen Shot 2021-04-20 at 10 30 40 AM

kenzieschmoll:
> now package:mime is giving us jpe for image/jpeg instead of jpg

package:mime has multiple abbreviations for some types. See all the options for image/jpeg here: https://github.com/dart-lang/mime/blob/master/lib/src/default_extension_map.dart#L363-L365.

packages/devtools_app/lib/src/network/network_request_inspector_views.dart
```diff
@@ -151,6 +160,86 @@ class HttpResponseView extends StatelessWidget {
...
+              padding: const EdgeInsets.all(
+                denseSpacing,
+              ),
+              child: Image.memory(
+                data.encodedResponse,
+              ),
+            ),
+          ],
+        ),
+        LayoutBuilder(builder: (context, constraints) {
```
kenzieschmoll, line 187: we aren't using the constraints anymore so we should be able to remove this layout builder

packages/devtools_app/lib/src/network/network_request_inspector_views.dart
```diff
@@ -151,6 +160,86 @@ class HttpResponseView extends StatelessWidget {
...
+      ],
+    );
+  }
+
+  Widget _buildRow(
+    BuildContext context,
+    String key,
+    String value,
+  ) {
+    return Container(
```
kenzieschmoll, line 221: change to Padding widget since padding is the only property being used from Container

bkonyi:
> > now package:mime is giving us jpe for image/jpeg instead of jpg
> 
> package:mime has multiple abbreviations for some types. See all the options for image/jpeg here: https://github.com/dart-lang/mime/blob/master/lib/src/default_extension_map.dart#L363-L365.

Right, but since there's no option there to actually select `jpeg`, do we care enough to work around it? `package:mime` will always return `jpe` for `image/jpeg` when using `extensionFromMime`.

kenzieschmoll:
There's a TODO there to consider a workaround for html as well - extensionFromMime returns htm. I'm fine with us adding a  workaround for common types we expect users to encounter.

bkonyi:
> There's a TODO there to consider a workaround for html as well - extensionFromMime returns htm. I'm fine with us adding a workaround for common types we expect users to encounter.

Done. Not sure where the TODO is though.

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -94,11 +95,28 @@ abstract class HttpRequestData extends NetworkRequest {
     if (mime.startsWith('[')) {
       mime = mime.substring(1);
     }
+    if (mime.endsWith(']')) {
+      mime = mime.substring(0, mime.length - 1);
+    }
     // TODO(kenz): consider special casing some extensions. For example,
```
kenzieschmoll, line 101: TODO I'm referring to is right here :)

kenzieschmoll (approved):

packages/devtools_app/lib/src/http/http_request_data.dart
bkonyi: I'm blind... :-) Added handling for `text/plain` while I'm here.

kenzieschmoll:
lgtm_army

bkonyi:
Merging on flaky red.

PR #2931, opened 2021-04-22
Add search in file functionality to the debugger code view.

kenzieschmoll:
cmd+f / ctrl+f opens the search bar. Pressing escape or clicking the 'x' button hides the search bar and clears search results.
debugger-search

### DIFF
--- a/packages/devtools_app/lib/src/debugger/debugger_controller.dart
+++ b/packages/devtools_app/lib/src/debugger/debugger_controller.dart
@@ -155,6 +155,8 @@ class DebuggerController extends DisposableController
     return cachedScript;
   }
 
+  /// Parses the current script into executable lines and prepares the script
+  /// for syntax highlighting.
   Future<void> _parseCurrentScript() async {
     // Return early if the current script has not changed.
     if (_currentParsedScript.value?.script?.id == _currentScriptRef.value.id)
--- a/packages/devtools_app/lib/src/debugger/debugger_controller.dart
+++ b/packages/devtools_app/lib/src/debugger/debugger_controller.dart
@@ -111,9 +111,10 @@ class DebuggerController extends DisposableController
 
   ValueListenable<ScriptRef> get currentScriptRef => _currentScriptRef;
 
-  final _currentParsedScript = ValueNotifier<ParsedScript>(null);
+  @visibleForTesting
+  final parsedScript = ValueNotifier<ParsedScript>(null);
 
-  ValueListenable<ParsedScript> get currentParsedScript => _currentParsedScript;
+  ValueListenable<ParsedScript> get currentParsedScript => parsedScript;
 
   final _showSearchInFileField = ValueNotifier<bool>(false);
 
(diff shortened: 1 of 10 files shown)


packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -127,12 +138,56 @@ class DebuggerController extends DisposableController
...
 
+  Future<Script> getScriptForRef(ScriptRef ref) async {
+    final cachedScript = getScriptCached(ref);
+    if (cachedScript == null && ref != null) {
+      return await getScript(ref);
+    }
+    return cachedScript;
+  }
+
+  Future<void> parseCurrentScript() async {
```
grouma, line 158: This looks like it can be private. We should also document what it's doing.

packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -1057,6 +1112,37 @@ class DebuggerController extends DisposableController
...
+
+  void toggleSearchInFileVisibility(bool visible) {
+    _showSearchInFileField.value = visible;
+    if (!visible) {
+      resetSearch();
+    }
+  }
+
+  @override
+  List<SourceToken> matchesForSearch(String search) {
```
grouma, line 1124: Can we adjust this to search across lines?

packages/devtools_app/lib/src/debugger/debugger_screen.dart
```diff
@@ -273,6 +284,17 @@ final LogicalKeySet goToLineNumberKeySet = LogicalKeySet(
   LogicalKeyboardKey.keyG,
 );
 
+final LogicalKeySet searchInFileKeySet = LogicalKeySet(
```
grouma, line 287: We don't have to do in this PR but it's probably worth breaking all of the keybindings out into a single file so that they are easy to find.

grouma (commented):
Some minor comments. This is awesome!

packages/devtools_app/lib/src/debugger/debugger_controller.dart
kenzieschmoll: you mean like search for a query like 'hello world' where 'hello' is on line 1 and 'world' is on line 2?

grouma: The latter. `hello` on line 1 and `world` on line two. I thought Chrome supported this but that doesn't seem to be the case. This is probably fine regardless.

kenzieschmoll: The IDEs don't support this type of search either, so I think it's on par if we leave as is.

packages/devtools_app/lib/src/debugger/debugger_screen.dart
kenzieschmoll: added a todo for this

grouma (approved):

packages/devtools_app/lib/src/debugger/debugger_controller.dart
kenzieschmoll: renamed and added dart doc

kenzieschmoll:
Added tests for search matching logic.

kenzieschmoll:
tests are passing except for flakes like 0.00% golden image diffs unrelated to this PR. Landing.

PR #2936, opened 2021-04-23
Improve debug hover experience

grouma:
- Significantly reduce the delay for hover (this feels more in line with Chrome)
- Remember the previous word so moving your mouse over the same word doesn't flash a new hover card


kenzieschmoll (approved):

grouma:
Merging on red since the flakes are unrelated.

Issue #2932, opened 2021-04-22
Logs can get out of order when they have the same timestamp


kenzieschmoll:
<img width="443" alt="Screen Shot 2021-04-22 at 3 19 13 PM" src="https://user-images.githubusercontent.com/43759233/115791934-3a0bf100-a37e-11eb-9c5d-5b23f01c3bf5.png">

we should do a secondary sort by event name so that the `#N` would determine the secondary ordering after timestamp. This makes the search iteration jumpy too because navigating through search results visits the logs in the order of `#N`

PR #2940, opened 2021-04-26
Add a secondary sort column to tables and fix logging page sorting bug.

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/2932, where logs are not ordered properly.


packages/devtools_app/lib/src/logging/logging_screen.dart
```diff
@@ -413,6 +415,20 @@ class _MessageColumn extends ColumnData<LogData>
...
       dataObject.summary ?? dataObject.details;
 
+  @override
+  int compare(LogData a, LogData b) {
+    final String valueA = getValue(a);
+    final String valueB = getValue(b);
+    // Matches frame descriptions (e.g. '#12  11.4ms ')
+    final regex = RegExp(r'#\d+\s+\d+.\d+ms\s*');
+    if (valueA.startsWith(regex) && valueB.startsWith(regex)) {
+      final frameNumberA = valueA.substring(1, valueA.indexOf(' '));
```
jacob314, line 425: safer to have a match in the regexp using `(` and `)` so the part you match in the reg-exp is exactly the # you expected.

packages/devtools_app/lib/src/logging/logging_screen.dart
```diff
@@ -413,6 +415,20 @@ class _MessageColumn extends ColumnData<LogData>
   String getValue(LogData dataObject) =>
       dataObject.summary ?? dataObject.details;
 
+  @override
+  int compare(LogData a, LogData b) {
+    final String valueA = getValue(a);
+    final String valueB = getValue(b);
+    // Matches frame descriptions (e.g. '#12  11.4ms ')
+    final regex = RegExp(r'#\d+\s+\d+.\d+ms\s*');
+    if (valueA.startsWith(regex) && valueB.startsWith(regex)) {
```
jacob314, line 424: How do you want to compare if one value is a number and the other isn't? Would it be better to sort all numbers before or after all non-numbers?

packages/devtools_app/lib/src/table.dart
```diff
@@ -1414,6 +1499,16 @@ abstract class SortableTable<T> {
...
-  return column.compare(a, b) * _compareFactor(direction);
+int _compareData<T>(
+  T a,
+  T b,
+  ColumnData column,
+  SortDirection direction, {
+  ColumnData secondarySortColumn,
+}) {
+  var compare = column.compare(a, b) * _compareFactor(direction);
+  if (compare == 0 && secondarySortColumn != null) {
```
jacob314, line 1510: tiny nit: maybe add
if (compare != 0) return compare;
that way of writing compares with multiple fields is cleaner if you added a 3rd field later.

packages/devtools_app/test/logging_screen_test.dart
```diff
@@ -312,6 +312,33 @@ void main() {
         });
       });
     });
+
+    group('MessageColumn', () {
```
jacob314, line 316: maybe add a simple test for the secondarySortColumn functionality?

Also add a test with the following data
```
     final c = LogData('flutter.frame', '#10 3.6ms', 1);
     final d = LogData('flutter.frame', '#10foo 3.6ms', 1);
     final e = LogData('flutter.frame', '#9  3.6ms ', 1);
     final f = LogData('flutter.frame', '#9foo  3.6ms ', 1);
     to see what happens where there is data that is almost a id but isn't.

File: packages/devtools_app/lib/src/logging/logging_screen.dart (hunk 2 above)
Comment (kenzieschmoll) [reply packages/devtools_app/lib/src/logging/logging_screen.dart:424]: I think the default string sorting would be fine in this case, which is covered by the end return `valueA.compareTo(valueB)`

File: packages/devtools_app/lib/src/logging/logging_screen.dart (hunk 1 above)
Comment (kenzieschmoll) [reply packages/devtools_app/lib/src/logging/logging_screen.dart]: done

File: packages/devtools_app/lib/src/logging/logging_screen.dart (hunk 2 above)
Comment (jacob314) [reply packages/devtools_app/lib/src/logging/logging_screen.dart:424]: I think that won't quite work because the sort won't be a consistent ordering.
Imagine you have
```
#2
#2NOTNUM
#1
#1NOTNUM
#10
#10NOTNUM
```
And see what you get when you sort it.

kenzieschmoll:
> lgtm once the sort is fixed so that non-numbers and numbers aren't interleaved. that makes it easier to get the comparator right.

The default string comparator handles this already. Symbols (like #) before numbers and both before letters.

packages/devtools_app/lib/src/logging/logging_screen.dart
kenzieschmoll: Ah whoops - github didn't refresh before I made my last comment. I see now. Let's put the frame numbers before non numbers.

kenzieschmoll: Fixed the sort.

PR #2943, opened 2021-04-26
Add "Load all CPU samples" button to the CPU profiler

kenzieschmoll:
Screen Shot 2021-04-26 at 3 39 41 PM

This will allow for profiling app start, since clicking this button pulls all the available CPU samples the VM has access to.

This CL also refactors some of the cpu profiler code to use widgets instead of helper methods


packages/devtools_app/lib/src/profiler/profiler_screen.dart
```diff
@@ -209,54 +262,42 @@ class _ProfilerScreenBodyState extends State<ProfilerScreenBody>
...
+  static const _secondaryControlsMinIncludeTextWidth = 1100.0;
+  final ProfilerScreenController controller;
 
-  Widget _buildSecondaryControls() {
+  @override
+  Widget build(BuildContext context) {
     return Row(
       mainAxisAlignment: MainAxisAlignment.end,
       children: [
+        RefreshButton(
```
jacob314, line 277: Fyi @InMatrix. I'm not sure users will understand what "Load all CPU samples" does.  Any ideas for a detailed tooltip and/or an alternate name that makes it clearer?
This is basically the refresh button from the Performance page but in addition to the existing record and pause buttons.  What we are trying to say is.
"The VM already has N seconds of recorded CPU profile samples in a buffer of recent activity. Display that data on this page?"

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/profiler/profiler_screen.dart
InMatrix: Do we expect this button to be used for debugging slow app start only? If that's the case, maybe giving it a name that reflects the user's goal, something like "Load samples from app start." 

> "The VM already has N seconds of recorded CPU profile samples in a buffer of recent activity. Display that data on this page?"

I wonder if the data would be misleading if the earliest available sample doesn't go all the way back to the app's start. For example, if the user clicks this button 10 minutes after they started their app, is there any guarantee that they'll still have samples from the start? Can we show timestamps relative to the app start time?

kenzieschmoll: > if the user clicks this button 10 minutes after they started their app, is there any guarantee that they'll still have samples from the start?

No, there is not. And if the user was actually interacting with their app, it's almost certain that we will not have samples from app start. 

> Can we show timestamps relative to the app start time?
This would be possible if we had a way to access the timestamp of app start. We could probably get close by taking the timestamp of the first timeline event we have, but that wouldn't be entirely accurate. @bkonyi do you know how we could get the timestamp of app start?

What if the button said "Load all available data" and then the tooltip said something more detailed like, "Load all CPU profile samples available in the VM sample buffer. This will include CPU samples from app start if this button is clicked immediately after your app finishes startup."

bkonyi: What's sufficient for a "timestamp" in this case? [getVMTimelineMicros](https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#getvmtimelinemicros) returns the approximate time the VM has been running in microseconds.

InMatrix: Is there a use case for loading all available data after the data from app start is gone? If this button won't be useful in that situation, it's probably better to disable that and show a tooltip explaining the intended workflow.

kenzieschmoll: > What's sufficient for a "timestamp" in this case? getVMTimelineMicros returns the approximate time the VM has been running in microseconds.

@bkonyi is this 0 based? I was unaware that is the case.

> Is there a use case for loading all available data after the data from app start is gone?

@InMatrix I could see a case where a user experienced a performance problem and wants to be able to profile what happened without having to reproduce the issue and record it under a record/stop workflow. Maybe it would be useful to have different intervals the user can select from. Like "load data from last 30 seconds" / minute / all available. Not sure what the UI would look like for that. Maybe a selector with choices and a paired button that says "Load"

InMatrix: I'm concerned that the user might have trouble associating this feature with the goal of debugging start-up time, which is the motivating use case of this PR. None of the proposed labels suggest the user will get the data from the first few seconds of their profiling session. My preference is to support that specific use case for now by providing a button to load data from the first 30 seconds and disable that button once that data is no longer available with a tooltip explanation. We can add support for other use cases once we have an opportunity to examine them more closely.

bkonyi: > @bkonyi is this 0 based? I was unaware that is the case.

Ah, I was mistaken, it's a monotonically increasing clock that starts at an arbitrary value determined by the OS. Anyway, the [service `VM`](https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#vm) object has a `startTime` property that should work.

kenzieschmoll: > My preference is to support that specific use case for now by providing a button to load data from the first 30 seconds and disable that button once that data is no longer available with a tooltip explanation.

@bkonyi would it be possible for us to know when data from app start is no longer available without having to poll for cpu samples and check the start time? This doesn't seem like a viable option as it would be fairly expensive.

bkonyi: No, there's no way to check without trying to grab at least some samples (e.g., samples starting from the time of application startup + some small window).

bkonyi: Also, there's a `--endless-trace-buffer` flag for `flutter run` that should be used when keeping all samples is necessary, but this shouldn't be used in situations where the application will be long running.

kenzieschmoll: Perhaps a compromise solution could be to add a button "Load profile from app start" or similar. We could pull a profile for app start + 30 seconds in one of three scenarios: 1) the user clicks the button, 2) the user opens devtools after the app start + 30 second mark has passed, and in this case we pull the profile and store it immediately upon the user opening the page, 3) the user opens devtools before the app start + 30 second mark, in which case we pull the app start profile at the first occurence of a) the user clicking the button, or b) the app start + 30 second mark has passed.

If we cached the profile data, then the app start profile would be available for loading as long as DevTools is left open and connected to the app. In the scenarios where the user did not explicitly click the "load profile from app start" button, we could pull the data and not process it to save the CPU load.

InMatrix: > If we cached the profile data, then the app start profile would be available for loading as long as DevTools is left open and connected to the app.

I think this is key. We need to control the workflow of profiling and ensure the proper order of events. To debug start-up time, DevTools needs to be ready to capture data _before_ the user starts the app in profile mode. We could consider enforcing this workflow by making it really explicit. For example, let's imagine a button called "Profile Start-up Time". Clicking it will request the user to restart the app in profile mode or do that for the user (with a heads-up). Upon restarting the app, DevTools will auto reconnect to the app and grab the CPU samples from the first N seconds. Some coordination between DevTools and Flutter tool might be required to make this workflow seamless. But does it sound remotely possible?

kenzieschmoll: > To debug start-up time, DevTools needs to be ready to capture data before the user starts the app in profile mode. 

This isn't entirely true. DevTools pulls data in a backward looking fashion - the VM is prepared to collect data as soon as the app starts, and DevTools can pull this data from the VM. We can specify a start and an end time on our request, so pulling the first 30 seconds is easy.

> Clicking it will request the user to restart the app in profile mode or do that for the user (with a heads-up). Upon restarting the app, DevTools will auto reconnect to the app and grab the CPU samples from the first N seconds. Some coordination between DevTools and Flutter tool might be required to make this workflow seamless

I think this would be possible. However, this seems like a different problem of automating a "Profile Start-up Time" process. I think this could be useful, definitely. But the solution I described _almost_ gets us to the same solution with a much simpler implementation cost. The main difference with what I suggested is that A) we can't guarantee that the startup CPU samples will be available if the user interacted with their app a long time before ever connecting to DevTools (but we could mitigate this with a warning and instructions in the profiler), and B) that we can't enforce a profile build runtime (though we already warn for this, and could add an additional warning if we wanted).

Maybe we should start with the behind-the-scenes pulling and caching the profile solution that I suggested, verify usefulness of the feature first, and then improve upon it with a more automated solution like you described? I do like the name of the button being "Profile app start up" or similar.

InMatrix: > Maybe we should start with the behind-the-scenes pulling and caching the profile solution that I suggested, verify usefulness of the feature first, and then improve upon it with a more automated solution like you described?

Trying out the pulling and caching idea sounds good to me. I see two related problems we need to solve: 1) the label of the button needs to be descriptive of what the user tries to do. 2) DevTools need to have consistent access to the CPU samples from the app start or have a user-in-the-loop process to generate such samples. It can become easier to solving 1) when 2) is solved.

kenzieschmoll: @bkonyi would it be possible to add code to dds that stores the profile from app start? Then DevTools could request the cached app start profile from DDS, no matter when the user actually opens DevTools. DDS knows the start time of the app, and could request the start up profiler when start up completes (maybe there is a way we can know the exact time of this?) or when some arbitrary number of seconds goes by (30?). And in the case that DevTools requests the app start up profile from DDS before either of these two events happen, DDS could return whatever data is available. I think this would let us avoid having to work with flutter tools to provide a good user experience for profiling app start.

bkonyi: It's possible, but there might be some performance implications if DDS is polling for CPU samples by default. Grabbing _only_ the samples from the first 30 seconds seems arbitrary, but buffering all samples would cause DDS to consume more and more memory until it eventually crashes, so I'm not sure where to draw the line.

I'm open to the idea of using DDS to manage some buffer of CPU samples which can store more elements than the VM's ring buffer can, but we'd probably want to pull together a doc to evaluate alternatives and outline design criteria before committing to anything.

kenzieschmoll: > Grabbing only the samples from the first 30 seconds seems arbitrary

this is what I am proposing. I'll throw together a 1-pager. Thanks.

jacob314: The overall approach sounds great. Instead of using 30 seconds, perhaps we could just surface profile samples until the first frame rendered and make this a `startup profile`. That could also give a clearer startup profile if the app happened to run an animation on startup after the first frame rendered and would mean we only need a second or two of profile samples unless an app has horrible performance.

kenzieschmoll: As a temporary solution while we finalize a more robust design using DDS, I'd like to change the button text to "Profile app startup", and manually request a profile in DevTools from time 0 to the timestamp of the Flutter.FirstFrame event. This runs the risk that samples in that timeframe will not be available, and if that happens, we can show a description for why that is. When we move to a more robust solution, this problem will go away, but this could give users something to work with for now.

For Dart CLI apps, we can request the first 30 seconds or just not show the button.

kenzieschmoll: Running into another problem here where the `Flutter.Frame` and `Flutter.FirstFrame` event timestamps don't match the timestamps that the VM uses for timeline events and CPU samples. @bkonyi @jonahwilliams ideas for ways around this?

jonahwilliams: Is this another case where having frame number of the CPU traces themselves would solve the problem?

kenzieschmoll: If we had a 0-based or 1-based frame numbering system, then we could look for the 0th or 1st frame. If we couldn't guarantee what the first frame number would be, then this wouldn't work, but if we can, then that would solve it.

kenzieschmoll:
I'm going to land this PR as a temporary solution to this use case. Once @bkonyi has completed the necessary work in DDS I will work on the improved implementation for profiling app start.

PR #2947, opened 2021-04-27
Add support for system isolates to isolate selector while in VM developer mode

bkonyi:
- Removed 'Isolate' from the isolate selector, replaced with 'forking' icon for isolates and 'gear' for system isolates.
- Added tooltip for the isolate selector: 'Selected Isolate'.
- Added 'No source available' path for CodeView.
- Disable various debugger controls for system isolates.

**Tooltip**
Screen Shot 2021-04-27 at 12 33 58 PM

**Isolates + system isolates with icons**
Screen Shot 2021-04-27 at 12 34 07 PM

**System isolate with source available**
Screen Shot 2021-04-27 at 12 34 15 PM

**System isolate with no source available**
Screen Shot 2021-04-27 at 12 34 20 PM

### DIFF
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -194,42 +194,44 @@ class _CodeViewState extends State<CodeView>
 
     // Ensure the syntax highlighter has been initialized.
     // TODO(bkonyi): process source for highlighting on a separate thread.
-    if (parsedScript.script.source.length < 500000 &&
-        parsedScript.highlighter != null) {
-      final highlighted = parsedScript.highlighter.highlight(context);
+    if (parsedScript.script.source != null) {
+      if (parsedScript.script.source.length < 500000 &&
+          parsedScript.highlighter != null) {
+        final highlighted = parsedScript.highlighter.highlight(context);
 
-      // Look for [TextSpan]s which only contain '\n' to manually break the
-      // output from the syntax highlighter into individual lines.
-      var currentLine = <TextSpan>[];
-      highlighted.visitChildren((span) {
-        currentLine.add(span);
-        if (span.toPlainText() == '\n') {
-          lines.add(
-            TextSpan(
-              style: theme.fixedFontStyle,
-              children: currentLine,
-            ),
-          );
-          currentLine = <TextSpan>[];
-        }
-        return true;
-      });
-      lines.add(
-        TextSpan(
-          style: theme.fixedFontStyle,
-          children: currentLine,
-        ),
-      );
-    } else {
-      lines.addAll(
-        [
-          for (final line in parsedScript.script.source.split('\n'))
-            TextSpan(
-              style: theme.fixedFontStyle,
-              text: line,
-            ),
-        ],
-      );
+        // Look for [TextSpan]s which only contain '\n' to manually break the
+        // output from the syntax highlighter into individual lines.
+        var currentLine = <TextSpan>[];
+        highlighted.visitChildren((span) {
+          currentLine.add(span);
+          if (span.toPlainText() == '\n') {
+            lines.add(
+              TextSpan(
+                style: theme.fixedFontStyle,
+                children: currentLine,
+              ),
+            );
+            currentLine = <TextSpan>[];
+          }
+          return true;
+        });
+        lines.add(
+          TextSpan(
+            style: theme.fixedFontStyle,
+            children: currentLine,
+          ),
+        );
+      } else {
+        lines.addAll(
+          [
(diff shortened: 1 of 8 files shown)


packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -245,62 +247,74 @@ class _CodeViewState extends State<CodeView>
...
-                                activeSearchMatchNotifier:
-                                    widget.controller.activeSearchMatch,
-                              );
-                            },
-                          ),
-                        ),
-                      ],
-                    );
-                  },
+          (lines.isNotEmpty)
```
kenzieschmoll, line 250: instead of using a ternary you can just use inline if / else
```
if (condition)
  widget
else
  widget,
```

packages/devtools_app/lib/src/debugger/controls.dart
```diff
@@ -60,20 +61,26 @@ class _DebuggingControlsState extends State<DebuggingControls>
     @required bool isPaused,
     @required bool resuming,
   }) {
+    final isSystemIsolate = controller.isolateRef.isSystemIsolate;
```
kenzieschmoll, line 64: make this a getter in this class since it is repeated

packages/devtools_app/lib/src/debugger/debugger_screen.dart
```diff
@@ -107,6 +108,13 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
   }
 
   Future<void> _toggleBreakpoint(ScriptRef script, int line) async {
+    // The VM doesn't support debugging for system isolates and will crash on
+    // a failed assert in debug mode. Disable the toggle breakpoint
+    // functionality for system isolates.
```
kenzieschmoll, line 113: in this case, we should also make sure we aren't drawing the little gray dots on executable lines in the gutter for system isolate scripts.

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -441,6 +446,24 @@ class IsolateManager extends Disposer {
...
+    addAutoDisposeListener(preferences.vmDeveloperModeEnabled, () async {
+      final vmDeveloperModeEnabled = preferences.vmDeveloperModeEnabled.value;
+      final vm = await serviceManager.service.getVM();
+      final isolates = [
+        ...vm.isolates,
+        if (vmDeveloperModeEnabled) ...vm.systemIsolates,
+      ];
+      if (selectedIsolate.isSystemIsolate && !vmDeveloperModeEnabled) {
+        selectIsolate(_isolates.first.id);
+      }
```
kenzieschmoll, line 461: does this also need to be called outside of the listener callback before _initIsolates(isolates) on l:464?

bkonyi: Nope, this should only be necessary when we remove system isolates from the list of available isolates. This just makes sure that we change the selected isolate to one of the remaining isolates if we currently have a system isolate selected when VM developer mode is disabled.

packages/devtools_app/lib/src/debugger/controls.dart
bkonyi: Done.

packages/devtools_app/lib/src/debugger/codeview.dart
bkonyi: I tried that before, but the trailing `,` after `widget` was giving me an error (I blame you for conditioning me to put trailing commas after everything!). Seems to work now :-)

packages/devtools_app/lib/src/debugger/debugger_screen.dart
bkonyi: Done. Also went ahead and forced the mouse cursor to the default instead of the pointing hand when there's nothing to interact with in the gutter.

kenzieschmoll (approved):

packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -250,6 +250,7 @@ class DebuggerController extends DisposableController
   ValueListenable<List<ConsoleLine>> get stdio => _stdio;
 
   IsolateRef isolateRef;
+  bool get isSystemIsolate => isolateRef.isSystemIsolate;
```
kenzieschmoll, line 253: should this be isolateRef?.isSystemIsolate in case isolateRef is null?

bkonyi: Doesn't hurt to be safe. When are we migrating to null safety again...? :-)

Issue #2953, opened 2021-04-28
Evaluation error when running a profile mode app


kenzieschmoll:
We should not be attempting to make eval calls in profile mode
```
flutter: [ERROR]: Unrecognized error: evaluate: (113) Expression compilation error
Debugger is disabled in AOT mode. - Platform.isAndroid
flutter: [ERROR]: #0      new _OutstandingRequest (package:vm_service/src/vm_service.dart:1606:45)
#1      VmService._call (package:vm_service/src/vm_service.dart:2078:21)
#2      VmService.evaluate (package:vm_service/src/vm_service.dart:1764:7)
#3      VmServiceWrapper.evaluate (package:devtools_app/src/vm_service_wrapper.dart:186:20)
#4      EvalOnDartLibrary._eval (package:devtools_app/src/eval_on_dart_library.dart:145:36)
<asynchronous suspension>
#5      EvalOnDartLibrary.addRequest.wrappedRequest (package:devtools_app/src/eval_on_dart_library.dart:438:30)
<asynchronous suspension>
```

PR #2954, opened 2021-04-28
Do not log eval error for _connectedToProfileBuild check

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/2953. This error is expected because we are using eval to determine whether or not the connected app is a profile build.


jacob314 (approved):
lgtm

PR #2956, opened 2021-04-28
Add source paths to CPU profile bottom up and call tree tables

kenzieschmoll:
revert changes from https://github.com/flutter/devtools/commit/229fee6fbb39c42c4d207dc526d7d0e6687f3323 since https://github.com/dart-lang/sdk/issues/37553 is fixed. Related to https://github.com/flutter/devtools/issues/2391.

### DIFF
--- a/packages/devtools_app/lib/src/profiler/cpu_profile_columns.dart
+++ b/packages/devtools_app/lib/src/profiler/cpu_profile_columns.dart
@@ -70,8 +70,18 @@ class TotalTimeColumn extends ColumnData<CpuStackFrame> {
 class MethodNameColumn extends TreeColumnData<CpuStackFrame> {
   MethodNameColumn() : super('Method');
 
+  static const maxMethodNameLength = 75;
+
   @override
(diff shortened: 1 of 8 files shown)


packages/devtools_app/lib/src/profiler/cpu_profile_columns.dart
```diff
@@ -70,9 +70,19 @@ class TotalTimeColumn extends ColumnData<CpuStackFrame> {
...
   MethodNameColumn() : super('Method');
 
+  static const maxMethodNameLength = 75;
+
   @override
   dynamic getValue(CpuStackFrame dataObject) => dataObject.name;
 
+  @override
+  String getDisplayValue(CpuStackFrame dataObject) {
+    if (dataObject.name.length > maxMethodNameLength) {
```
jacob314, line 80: instead of truncating manually, we should let Flutter determine where to truncate the method name using the actual layout.

jacob314 (approved):
lgtm
Can you file a bug to eliminate the manual string truncation from the profiler?

PR #2957, opened 2021-04-29
Refactor Debugger history view into HistoryViewport widget

bkonyi:
Will allow for future re-use in VM tools (e.g., object inspector).

### DIFF
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -257,7 +257,7 @@ class _CodeViewState extends State<CodeView>
           enabled: widget.controller.scriptsHistory.hasScripts,
         ),
       ],
-      buildContents: (context, script) {
+      contentBuilder: (context, script) {
         if (lines.isNotEmpty) {
           return DefaultTextStyle(
             style: theme.fixedFontStyle,
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -18,6 +18,7 @@ import '../config_specific/logger/logger.dart';
 import '../dialogs.dart';
 import '../flutter_widgets/linked_scroll_controller.dart';
 import '../globals.dart';
+import '../history_viewport.dart';
 import '../theme.dart';
 import '../ui/colors.dart';
 import '../ui/search.dart';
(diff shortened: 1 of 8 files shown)


packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -243,81 +244,93 @@ class _CodeViewState extends State<CodeView>
...
-                                widget.controller.breakpointsWithLocation,
-                            builder: (context, breakpoints, _) {
-                              return Gutter(
-                                gutterWidth: gutterWidth,
-                                scrollController: gutterController,
-                                lineCount: lines.length,
+    return HistoryViewport(
+      history: widget.controller.scriptsHistory,
+      generateTitle: (script) => script.uri,
+      buildControls: (context) {
```
kenzieschmoll, line 250: instead of passing in buildControls and buildContents as parameters, can we just build these in the HistoryViewport widget itself?

bkonyi: `buildContents` needs to be a builder as we need to rebuild the content whenever the currently selected historical data changes. `buildControls` is there since this is supposed to be generalized to be used with any historical data and it wouldn't make sense to have script specific controls for non-script history (e.g., navigation history for VM service objects).

bkonyi: `buildControls` could take a `Row` instead of a `Row Function(BuildContext)` though.

packages/devtools_app/lib/src/history_viewport.dart
```diff
@@ -0,0 +1,90 @@
...
+  const HistoryViewport({
+    @required this.history,
+    @required this.buildContents,
+    this.buildControls,
+    this.generateTitle,
+  });
+
+  final HistoryManager<T> history;
+  final String Function(T) generateTitle;
+  final List<Widget> Function(BuildContext) buildControls;
```
kenzieschmoll, line 36: instead of making this a builder, just make this a `List<Widget> controls` that is passed in.

packages/devtools_app/lib/src/history_manager.dart
```diff
@@ -0,0 +1,87 @@
+// Copyright 2021. The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+
+class HistoryManager<T> extends ChangeNotifier
```
kenzieschmoll, line 8: can you rename this to HistoryNotifier to make it clear that this is a notifier?

kenzieschmoll: actually on second thought, it seems strange that this class is a notifier at all? Why can't we just have this class be a normal dart class that has a value notifiers for the data other parts of the UI are subscribing to? then all the methods in here can update those notifiers accordingly

packages/devtools_app/lib/src/history_viewport.dart
```diff
@@ -0,0 +1,90 @@
...
+/// [buildContents] is invoked with the currently selected historical data
+/// when building the contents of the viewport.
+///
+/// If [buildControls] is provided, the returned [List<Widget>] will be
+/// inserted at the end of the viewport title bar.
+///
+/// If [generateTitle] is provided, the title string will be set to the
+/// returned value. If not provided, the title will be empty.
+class HistoryViewport<T> extends StatelessWidget {
+  const HistoryViewport({
```
kenzieschmoll, line 27: we should be able to make this class more declarative instead of using builders. Instead of using builders and generators, just pass in `String title, List<Widget> controls, Widget child`. Sound like the problem is that there is some state that is not getting updated when the selected script changes? If that's the case, then I think the proper way to make sure this widget rebuilds is to wrap it in a ValueListenable builder where this widget is used, or to make this widget stateful and add listeners to appropriate notifiers in initState

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -243,81 +244,91 @@ class _CodeViewState extends State<CodeView>
...
+        ScriptPopupMenu(widget.controller),
+        ScriptHistoryPopupMenu(
+          itemBuilder: _buildScriptMenuFromHistory,
+          onSelected: (scriptRef) {
+            widget.controller.showScriptLocation(ScriptLocation(scriptRef));
+          },
+          enabled: widget.controller.scriptsHistory.hasScripts,
+        ),
+      ],
+      buildContents: (context, script) {
```
kenzieschmoll, line 260: super-nit: can you rename to `contentBuilder` to match naming conventions for other builders in flutter

packages/devtools_app/lib/src/history_manager.dart
```diff
@@ -0,0 +1,78 @@
...
+  void push(T value) {
+    _history.add(value);
+    _historyIndex = _history.length - 1;
+    _current.value = _history[_historyIndex];
+  }
+
+  /// The currently selected historical item.
+  ///
+  /// Returns null if there is no historical items.
+  ValueListenable<T> get current => _current;
```
kenzieschmoll, line 77: nit: move this getter to be above `_current`

packages/devtools_app/lib/src/history_manager.dart
```diff
@@ -0,0 +1,78 @@
...
+
+  /// Returns true if there is a next historical item available on the stack.
+  bool get hasNext {
+    return _history.isNotEmpty && _historyIndex < _history.length - 1;
+  }
+
+  /// Return the next historical item on the stack.
+  ///
+  /// Throws [StateError] if there's no next item available.
+  T moveForward() {
```
kenzieschmoll, line 27: I don't think these methods need to return anything. Doesn't look like the return values are used anywhere, nor should they be really because we want `current` to be the source of data for parts of the code that care about the `_current.value`. Can we change the return to void for `moveForward`, `moveBack`, and `pop`?

packages/devtools_app/lib/src/history_viewport.dart
```diff
@@ -0,0 +1,94 @@
...
+    final theme = Theme.of(context);
+    return OutlineDecoration(
+      child: Column(
+        children: [
+          _buildTitle(context, theme),
+          ValueListenableBuilder(
+              valueListenable: history.current,
+              builder: (context, current, _) {
+                return buildContents(context, current);
+              }),
```
kenzieschmoll, line 50: trailing comma

packages/devtools_app/lib/src/history_viewport.dart
```diff
@@ -0,0 +1,94 @@
...
+        ],
+      ),
+    );
+  }
+
+  Widget _buildTitle(BuildContext context, ThemeData theme) {
+    return ValueListenableBuilder(
+      valueListenable: history.current,
```
kenzieschmoll, line 58: instead of wrapping in a ValueListenableBuilder here and above around `buildContents`, just use one ValueListenableBuilder that wraps the surrounding Column and pass the value current into the _buildTitle method

kenzieschmoll (approved):
LGTM after a few comments are resolved. Nice refactor 👍

bkonyi:
Landing on flaky failures.

Issue #2941, opened 2021-04-26
Search box on flame graph page freezes when searching in large profile


jonahwilliams:
For the following profile (renamed to .txt to defeat githubs upload check), type `v` in the search box and observe the page become unresponsive.

[dart_devtools_2021_4_26-1619463922190000.json.txt](https://github.com/flutter/devtools/files/6379640/dart_devtools_2021_4_26-1619463922190000.json.txt)

PR #2959, opened 2021-04-29
Improve search and build performance in flame charts

kenzieschmoll:
Improve flame chart search performance by **~27x** by caching the values isSearchMatch and isActiveSearch. We were performing very expensive calls to `List.contains` in the process of building each flame chart node. 

Before:
Screen Shot 2021-04-29 at 9 17 36 AM

After:
Screen Shot 2021-04-29 at 9 36 53 AM

Fixes https://github.com/flutter/devtools/issues/2941

### DIFF
--- a/packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
+++ b/packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
@@ -6,6 +6,7 @@ import 'dart:convert';
 
 import 'package:meta/meta.dart';
 
+import '../charts/flame_chart.dart';
 import '../trace_event.dart';
 import '../trees.dart';
 import '../ui/search.dart';
@@ -158,7 +159,10 @@ class CpuProfileMetaData {
 }
 
 class CpuStackFrame extends TreeNode<CpuStackFrame>
-    with SearchableDataMixin, SearchableTreeDataMixin<CpuStackFrame> {
+    with
+        DataSearchStateMixin,
+        TreeDataSearchStateMixin<CpuStackFrame>,
+        FlameChartDataMixin {
   CpuStackFrame({
     @required this.id,
     @required this.name,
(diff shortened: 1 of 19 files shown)


packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
```diff
@@ -209,6 +211,10 @@ class CpuStackFrame extends TreeNode<CpuStackFrame> {
 
   Duration _selfTime;
 
+  @override
+  String get tooltip =>
+      '$name - ${msText(totalTime)}${url != null ? ' - $url' : ''}';
```
jacob314, line 216: nit: rewrite as
[name, msText(totalTime), if (url != null) url].join(' - ')

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -454,3 +457,14 @@ class SearchNavigationControls extends StatelessWidget {
     );
   }
 }
+
+mixin SearchableDataMixin {
+  bool isSearchMatch = false;
+  bool isActiveSearchMatch = false;
+  String get tooltip => '';
```
jacob314, line 464: Why does this mixin have a tooltip? That seems unrelated to the rest of what the mixin is doing.
Also consider renaming this 
SearchStateMixin or similar as this mixin seems to be the state of whether something is a search match. That name suggestion isn't great so feel free to ignore or coome up with something better.

kenzieschmoll: I did this to avoid having to add another Mixin, but you're right it doesn't really fit here. I was hoping to be able to reuse this tooltip in the table code in a follow up PR, but for now I created a mixin `FlameChartDataMixin` that has the tooltip getter, and added that to the data models we use in flame charts. I can generalize in the follow up PR where I address the table search perf.

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart

jacob314 (approved):
lgtm

PR #2962, opened 2021-05-01
Add richer auto-complete for use in an expression evaluator.

terrylucas:
Auto-complete e.g.,

https://user-images.githubusercontent.com/2055873/116821024-e68f6500-ab2c-11eb-9218-503f830d8cf3.mov

When the pop-up auto-complete list is visible the following keys have special actions (even though the caret is in the TextField):
- Up/Down arrows highlight a line in the list and cycles around the list.
- Enter or Tab keys selects the highlighted line.
- Escape closes the pop-up auto-complete list.
- Tap/Click on an item in the pop-up auto-complete list selects the item.

The static method `activeEdtingParts` in `search.dart` will help to decide the part of an expression (word) that should be used to auto-complete. An active word is a variable name:

`[_, a..z, A..Z] [_, a..z, A..Z, 0..9]+
`

And a dot (period) signals that a field is being typed.

Fixes/Changes in this PR:
- Re-plumbed auto-complete to support a robust version to work with an expression evaluator.
- Auto-complete refreshed too much, subclass the auto-complete list (in the overlay). Now only the auto-complete list is updated not the TextField too.
- Support TextEditingValue (text and selection) within the TextField when replacing a word from the auto-complete list.
- Supports searching both snapshots (single word) and expression evaluator e.g., multiple words and field processing (dot operator).
- If tracking in auto-complete is enabled, the pop-up will appear immediately to the right-most character in the TextField.
- Fixed some deprecated flutter names (analyzer error).
- Fixed goldens images could fail if the date of golden run changed. Updated to use a fixed date/time (in the test) to guarantee the date/time displayed in the auto-tracking line.
- Fixed raw keys handling for Enter/Tab working correctly on all platforms (Mac/Linux).
- Added more tests.


packages/devtools_app/lib/src/app.dart
```diff
@@ -48,6 +48,18 @@ import 'utils.dart';
 import 'vm_developer/vm_developer_tools_controller.dart';
 import 'vm_developer/vm_developer_tools_screen.dart';
 
+// Assign to true to use a sample implementation of a conditional screen.
+// WARNING: Do not check in this file with enableExample set to true.
+const enableSampleScreen = false;
```
jacob314, line 53: nit: call this debugEnableSampleScreen to match the flutter convention for top level fields that are debug toggles.

packages/devtools_app/lib/src/app.dart
```diff
@@ -613,15 +625,10 @@ List<DevToolsScreen> get defaultScreens {
...
-// Uncomment to see a sample implementation of a conditional screen.
-//      DevToolsScreen<ExampleController>(
-//        const ExampleConditionalScreen(),
-//        createController: () => ExampleController(),
-//        supportsOffline: true,
-//      ),
   ];
+
+  // Show the sample DevTools screen.
+  if (enableSampleScreen) screen.add(sampleConditionScreen);
```
jacob314, line 631: only ever include this screen in debug + profile builds.

packages/devtools_app/lib/src/example/conditional_screen.dart
```diff
@@ -59,6 +72,164 @@ class _ExampleConditionalScreenBodyState
...
+        maxWidth: false,
+      );
+    });
+
+    addAutoDisposeListener(controller.selectTheSearchNotifier, _handleSearch);
+
+    addAutoDisposeListener(controller.searchNotifier, _handleSearch);
+  }
+
+  List<String> _nameScopeMatches(
```
jacob314, line 91: putting examples of more functionality in the example screen is a great idea!

packages/devtools_app/lib/src/app.dart
```diff
@@ -613,15 +625,10 @@ List<DevToolsScreen> get defaultScreens {
       VMDeveloperToolsScreen(controller: vmDeveloperToolsController),
       controller: vmDeveloperToolsController,
     ),
-    DevToolsScreen<void>(
```
jacob314, line 616: merge conflict. Looks like you accidentally removed the provider screen.

packages/devtools_app/lib/src/example/conditional_screen.dart
```diff
@@ -82,10 +282,50 @@ class _ExampleConditionalScreenBodyState
             color: Colors.grey[50],
             child: const CenteredCircularProgressIndicator(),
           ),
+        evaluator,
       ],
     );
   }
 
+  void highlightDropdown(bool directionDown) {
```
jacob314, line 290: nit: this method seems like functionality that could be built into the auto-complete behavior. I don't see anything different about this case than a default implementation.

jacob314 (approved):

packages/devtools_app/lib/src/app.dart
terrylucas: Done

packages/devtools_app/lib/src/app.dart
terrylucas: Good idea - Done.

packages/devtools_app/lib/src/app.dart
terrylucas: Good catch I didn't notice that. Fixed.

packages/devtools_app/lib/src/example/conditional_screen.dart
terrylucas: Good point, moved to built in default behavior.

PR #2967, opened 2021-05-04
Fixed grep from double escaping.

terrylucas:
Unable to find canvas kit version because string to find was

`"defaultValue: \'\''https://unpkg.com/canvaskit-wasm@0.25.1/bin/"`

instead of

`"defaultValue: 'https://unpkg.com/canvaskit-wasm@0.25.1/bin/'"`


jacob314 (approved):
lgtm

PR #2969, opened 2021-05-04
Refactor CPU profiler screen controls into their own widgets

kenzieschmoll:
no functional changes, just breaking some code out into their own widgets.


jacob314 (approved):
lgtm

PR #2970, opened 2021-05-04
Temporarily disable the provider screen.

terrylucas:


kenzieschmoll:
We can't import `freezed_annotation` because we don't want that package in google3. Let's disable the provider screen and figure out a way around that for the next milestone. That was the original plan, so we may need to dig a bit more to make sure this dep is not necessary.

kenzieschmoll:
LGTM on pub_publish.sh changes

terrylucas:
Okay removed the dependency and temporarily disabled the provider screen.

kenzieschmoll (approved):

jacob314 (approved):
lgtm

PR #2975, opened 2021-05-04
Use fake freeze_annotation instead of 3rd party package.

terrylucas:


jacob314 (commented):
It was a bug that freezed_annotation was used. All the code was supposed to reference fake_freezed_anotation. You can remove the TODO.

jacob314:
lgtm

PR #2976, opened 2021-05-04
Refactor ProfileGranularityDropdown to be a stateless widget

kenzieschmoll:
no functional changes - just a cleanup

### DIFF
--- a/packages/devtools_app/lib/src/profiler/cpu_profile_service.dart
+++ b/packages/devtools_app/lib/src/profiler/cpu_profile_service.dart
@@ -7,11 +7,12 @@ import 'package:meta/meta.dart';
 
 import '../globals.dart';
 import '../vm_flags.dart' as vm_flags;
+import '../vm_service_wrapper.dart';
 import 'cpu_profile_model.dart';
 
 /// Manages interactions between the Cpu Profiler and the VmService.
-abstract class CpuProfilerService {
-  static Future<CpuProfileData> getCpuProfile({
+extension CpuProfilerExtension on VmServiceWrapper {
+  Future<CpuProfileData> getCpuProfile({
     @required int startMicros,
     @required int extentMicros,
   }) async {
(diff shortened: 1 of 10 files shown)


packages/devtools_app/lib/src/ui/vm_flag_widgets.dart
```diff
@@ -14,24 +14,18 @@ import '../profiler/profile_granularity.dart';
 /// DropdownButton that controls the value of the 'profile_period' vm flag.
 ///
 /// This flag controls the rate at which the vm samples the CPU call stack.
-class ProfileGranularityDropdown extends StatefulWidget {
-  const ProfileGranularityDropdown(this.screenId);
+class ProfileGranularityDropdown extends StatelessWidget {
+  ProfileGranularityDropdown(this.screenId)
+      : profilerService = CpuProfilerService();
```
jacob314, line 19: creating a mutable profiler service in the constructor of a stateful widget seems like a bug. You don't want a new profiler service every time the widget rebuilds. Can you hang the service off the service manager, use a provider, or a stateful widget?

kenzieschmoll: Looking back through the original code, the methods on CpuProfilerService probably could have / should have been written as static. All they really are are helpers that interact with the vm service. Cleaned this up - PTAL

packages/devtools_app/lib/src/profiler/cpu_profile_service.dart
```diff
@@ -23,23 +22,16 @@ class CpuProfilerService {
...
-      serviceManager.vmFlagManager.flag(vm_flags.profilePeriod);
-
-  Future clearCpuSamples() {
+  static Future clearCpuSamples() {
     return serviceManager.service
         .clearCpuSamples(serviceManager.isolateManager.selectedIsolate.id);
   }
 
-  Future<dynamic> setProfilePeriod(String value) {
+  static Future<dynamic> setProfilePeriod(String value) {
```
jacob314, line 30: nit: make these extension methods on serviceManager.service?
I don't think we have any services implemented as static methods so this is a little confusing and hides the fact that you are depending on the current serviceManager.service.

jacob314 (approved):
lgtm

Issue #2972, opened 2021-05-04
Burndown list to enable the provider page by default


jacob314:
- [x] Eliminate hangs opening providers https://github.com/flutter/devtools/issues/2971
- [x] Normalize colors and style with debugger object view. For the most part the provider UX is superior. However displying the number of elements (fields) for objects seems questionable. https://github.com/flutter/devtools/issues/2987
- [x] Fix value editing UI which doesn't have enough whitespace. (only repros on web)
- [x] Provider screen shows spinning icon for some apps that never goes away https://github.com/flutter/devtools/issues/2980
- [x] Provider screen fails to update after a hot restart. https://github.com/flutter/devtools/issues/2977

jacob314:
Fyi @rrousselGit

rrousselGit:
> Evaluate whether a refresh button is needed or the expected behavior is that providers will be displayed when available.

The current behavior is, if there is no selected provider but there is one provider in the list, the provider is selected. 

I handled all scenarios where this could happen. Such as when the selected provider is unmounted in the inspected ap


So unless there are no providers in the application, one provider will be selected.

jacob314:
> > Evaluate whether a refresh button is needed or the expected behavior is that providers will be displayed when available.
> 
> The current behavior is, if there is no selected provider but there is one provider in the list, the provider is selected.
> 
> I handled all scenarios where this could happen. Such as when the selected provider is unmounted in the inspected ap
> 
> So unless there are no providers in the application, one provider will be selected.

I played around and in general updating the providers works beautifully. It appears there is a race condition the first time the provider screen is opened for an app. Added a bug with details on how to repro. Seems very reliable on the simple provider sample apps.

rrousselGit:
Could you expand on what the following is about?

> Normalize colors and style with debugger object view. For the most part the provider UX is superior. However displying the number of elements (fields) for objects seems questionable.

jacob314:
> Could you expand on what the following is about?
> 
> > Normalize colors and style with debugger object view. For the most part the provider UX is superior. However displying the number of elements (fields) for objects seems questionable.
Broke this out into a bug.
https://github.com/flutter/devtools/issues/2987

kenzieschmoll:
AFAICT, all these issues have been resolved. Please reopen if that is incorrect.

PR #2985, opened 2021-05-06
Fix editing TextField padding

rrousselGit:
This should solve https://github.com/flutter/devtools/issues/2972:

> Fix value editing UI which doesn't have enough whitespace. (only repros on web)

It also adds support for "tap escape to stop editing"


kenzieschmoll (approved):

kenzieschmoll:
looks like the bots are red because of an unused import

rrousselGit:
This should be good now

PR #2986, opened 2021-05-06
Throw RPCError when invoking getSourceReport in profile mode

bkonyi:
There was a bug in the VM service where invoking getSourceReport on an application running in profile mode (Dart AOT) would cause the VM to crash (resolved in [this CL](https://dart-review.googlesource.com/c/sdk/+/197241)).

This change adds a guard in vm_service_wrapper.dart to manually check for a profile mode run before invoking the RPC, throwing the correct RPCError that the VM service package would throw if the VM didn't crash.

### DIFF
--- a/packages/devtools_app/lib/src/vm_service_wrapper.dart
+++ b/packages/devtools_app/lib/src/vm_service_wrapper.dart
@@ -8,6 +8,7 @@ import 'package:dds/vm_service_extensions.dart';
 import 'package:meta/meta.dart';
 import 'package:vm_service/vm_service.dart';
 
+import 'globals.dart';
 import 'profiler/cpu_profile_model.dart';
 import 'version.dart';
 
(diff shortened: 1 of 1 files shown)


bkonyi:
https://github.com/flutter/devtools/pull/2931 is the likely candidate for triggering the error, which is why we hadn't encountered this issue previously even though there was a bug in the service for ~2+ years.

kenzieschmoll (approved):
this is probably important enough to do a cherry pick release

bkonyi:
I agree, this is impacting quite a few users.

jacob314 (commented):
How broken is the debugger page with this change? I'm a bit concerned it will now be throwing a lot of exceptions. We should probably at least dd a  try-catch block to the debugger.
https://github.com/flutter/devtools/blob/c619cb8a9b0c69545cc2664e2dad9971cc22f910/packages/devtools_app/lib/src/debugger/debugger_controller.dart#L1113

bkonyi:
> How broken is the debugger page with this change? I'm a bit concerned it will now be throwing a lot of exceptions. We should probably at least dd a try-catch block to the debugger.
> https://github.com/flutter/devtools/blob/c619cb8a9b0c69545cc2664e2dad9971cc22f910/packages/devtools_app/lib/src/debugger/debugger_controller.dart#L1113

There's a try-catch up the call tree. This is the exact exception that would have been thrown by `package:vm_service` assuming the bug in the VM service causing the crash wasn't there.

Issue #2964, opened 2021-05-03
Support UserTags in the DevTools CPU profiler


kenzieschmoll:
Users should be able to group samples in the profiler: https://api.flutter.dev/flutter/dart-developer/UserTag-class.html

Issue #947, opened 2019-08-22
CPU profiler start and end times should be able to be configured in user code


kenzieschmoll:
In order to profile a single method or sequence of events, a user should be able to specify start and end times in their code.

Maybe there is something in the dart:developer API that the user can use to do this. Then, when recording a CPU profile, we can pull samples from the interval specified by the user-triggered events.

This mode could be enabled via a setting "explicit profile interval" or something more clever.

cc @lambdabaa

escamoteur:
May I add I think a time interval might not be the best solution but just a **start/stop** function. Also not only for CPU profile but also Performance and a Future Method table.  
For Memory to **make snapshot** from the program and a command to  **reset the accumulator.**

PR #2988, opened 2021-05-06
Support filtering CPU profiles by UserTags

kenzieschmoll:
If a user specifies [UserTag](https://api.dart.dev/stable/2.12.4/dart-developer/UserTag/UserTag.html)s in their code, those tags will show up in the DevTools CPU profiler as filter options.
filter-by-user-tag
Fixes https://github.com/flutter/devtools/issues/2964
Fixes https://github.com/flutter/devtools/issues/947

### DIFF
--- a/packages/devtools_app/lib/src/performance/performance_controller.dart
+++ b/packages/devtools_app/lib/src/performance/performance_controller.dart
@@ -418,8 +418,9 @@ class PerformanceController
     }
 
     if (offlinePerformanceData.cpuProfileData != null) {
-      cpuProfilerController
-          .loadProcessedData(offlinePerformanceData.cpuProfileData);
+      cpuProfilerController.loadProcessedData(
+        offlinePerformanceData.cpuProfileData,
+      );
     }
   }
 
--- a/packages/devtools_app/lib/src/performance/performance_controller.dart
+++ b/packages/devtools_app/lib/src/performance/performance_controller.dart
@@ -418,7 +418,8 @@ class PerformanceController
     }
 
     if (offlinePerformanceData.cpuProfileData != null) {
-      cpuProfilerController.loadData(offlinePerformanceData.cpuProfileData);
+      cpuProfilerController
+          .loadProcessedData(offlinePerformanceData.cpuProfileData);
     }
   }
 
--- a/packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
+++ b/packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
@@ -143,7 +143,6 @@ class CpuProfilerController with SearchControllerMixin<CpuStackFrame> {
     _processingNotifier.value = true;
 
     try {
-      assert(false);
       _dataNotifier.value = await processDataForTag(tag);
     } catch (e) {
       // In the event of an error, reset the data to the original CPU profile.
--- a/packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
+++ b/packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
@@ -15,6 +15,8 @@ import 'cpu_profile_service.dart';
 import 'cpu_profile_transformer.dart';
 
 class CpuProfilerController with SearchControllerMixin<CpuStackFrame> {
+  static const userTagNone = 'none';
+
   /// Data for the initial value and reset value of [_dataNotifier].
   ///
   /// When this data is the value of [_dataNotifier], the CPU profiler is in a
@@ -33,6 +35,14 @@ class CpuProfilerController with SearchControllerMixin<CpuStackFrame> {
   ValueListenable get selectedCpuStackFrameNotifier =>
       _selectedCpuStackFrameNotifier;
   final _selectedCpuStackFrameNotifier = ValueNotifier<CpuStackFrame>(null);
+
+  /// Notifies that a user tag filter is set on the CPU profile flame chart.
+  ValueListenable<String> get userTagFilter => _userTagFilter;
+  final _userTagFilter = ValueNotifier<String>(userTagNone);
+
+  final _dataByTag = <String, CpuProfileData>{};
+
+  Iterable<String> get userTags => _dataByTag[userTagNone]?.userTags ?? [];
 
   int selectedProfilerTabIndex = 0;
 
@@ -85,6 +95,7 @@ class CpuProfilerController with SearchControllerMixin<CpuStackFrame> {
     try {
       await transformer.processData(cpuProfileData, processId: processId);
       _dataNotifier.value = cpuProfileData;
+      _dataByTag[userTagNone] = cpuProfileData;
       refreshSearchMatches();
       _processingNotifier.value = false;
     } on AssertionError catch (_) {
@@ -115,9 +126,32 @@ class CpuProfilerController with SearchControllerMixin<CpuStackFrame> {
     return matches;
   }
 
-  void loadData(CpuProfileData data) {
+  void loadProcessedData(CpuProfileData data) {
     assert(data.processed);
     _dataNotifier.value = data;
+    _dataByTag[userTagNone] = data;
+  }
+
+  Future<void> loadDataWithTag(String tag) async {
+    _userTagFilter.value = tag;
+
+    _dataNotifier.value = null;
+    _processingNotifier.value = true;
+
+    _dataNotifier.value = await processDataForTag(tag);
+    _processingNotifier.value = false;
+  }
+
+  Future<CpuProfileData> processDataForTag(String tag) async {
+    final fullData = _dataByTag[userTagNone];
+    final data = _dataByTag.putIfAbsent(
+      tag,
+      () => fullData.dataForUserTag(tag),
+    );
+    if (!data.processed) {
+      await transformer.processData(data);
+    }
+    return data;
   }
 
   void selectCpuStackFrame(CpuStackFrame stackFrame) {
@@ -134,6 +168,7 @@ class CpuProfilerController with SearchControllerMixin<CpuStackFrame> {
   void reset() {
     _selectedCpuStackFrameNotifier.value = null;
     _dataNotifier.value = baseStateCpuProfileData;
+    _dataByTag.clear();
     _processingNotifier.value = false;
     transformer.reset();
     resetSearch();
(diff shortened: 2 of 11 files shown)


packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
```diff
@@ -15,6 +15,8 @@ import 'cpu_profile_service.dart';
 import 'cpu_profile_transformer.dart';
 
 class CpuProfilerController with SearchControllerMixin<CpuStackFrame> {
+  static const userTagNone = 'none';
```
jacob314, line 18: Comment whether the word 'none' has a magic meaning or is just a user friendly name for no tag.

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
```diff
@@ -34,6 +36,20 @@ class CpuProfilerController with SearchControllerMixin<CpuStackFrame> {
...
 
+  /// Notifies that a user tag filter is set on the CPU profile flame chart.
+  ValueListenable<String> get userTagFilter => _userTagFilter;
+  final _userTagFilter = ValueNotifier<String>(userTagNone);
+
+  final _dataByTag = <String, CpuProfileData>{};
+
+  Iterable<String> get userTags => _dataByTag[userTagNone]?.userTags ?? [];
+
+  int selectedProfilerTabIndex = 0;
```
jacob314, line 47: why isn't this a value listenable?

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
```diff
@@ -109,9 +126,32 @@ class CpuProfilerController with SearchControllerMixin<CpuStackFrame> {
...
+    _dataByTag[userTagNone] = data;
+  }
+
+  Future<void> loadDataWithTag(String tag) async {
+    _userTagFilter.value = tag;
+
+    _dataNotifier.value = null;
+    _processingNotifier.value = true;
+
+    _dataNotifier.value = await processDataForTag(tag);
```
jacob314, line 141: wrap with a try+finally block in case an exception is thrown. otherwise you can get stuck processing.

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
```diff
@@ -181,6 +245,24 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
...
 
+  Iterable<String> get userTags => _userTagSampleCount.keys;
+
+  /// Maps a user tag to the number of CPU samples associated with it.
+  ///
+  /// A single [CpuStackFrame] can have multiple tags because a single object
+  /// can be part of multiple samples.
+  final _userTagSampleCount = <String, int>{};
+
+  void incrementTag(String userTag, {int increment = 1}) {
```
jacob314, line 256: nit: incrementTag --> incrementTagSampleCount

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
kenzieschmoll: done

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
kenzieschmoll: done. also added code to push a notification when the filter attempt fails.

packages/devtools_app/lib/src/profiler/cpu_profiler.dart
```diff
@@ -253,3 +267,73 @@ class CpuProfilerDisabled extends StatelessWidget {
...
+  const UserTagDropdown(this.controller);
+
+  final CpuProfilerController controller;
+
+  @override
+  Widget build(BuildContext context) {
+    const filterByTag = 'Filter by tag:';
+    // This needs to rebuild whenever there is new CPU profile data because the
+    // user tags will change with the data.
+    return ValueListenableBuilder<CpuProfileData>(
```
jacob314, line 284: Add a todo referencing https://github.com/flutter/devtools/issues/2989
to remove the nested value listenables.

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart

packages/devtools_app/lib/src/profiler/cpu_profiler.dart
```diff
@@ -253,3 +267,73 @@ class CpuProfilerDisabled extends StatelessWidget {
...
+                  // case.
+                  if (!(userTags.length == 1 &&
+                      userTags.first == UserTag.defaultTag.label))
+                    for (final tag in userTags)
+                      _buildMenuItem(
+                        display: '$filterByTag $tag',
+                        value: tag,
+                      ),
+                ],
+                onChanged: userTags.isNotEmpty ? _onUserTagChanged : null,
```
jacob314: shouldn't this be items.length > 1 instead of userTags.isNotEmpty?
I expect you don't really want this to be selectable if the only tag was the default tag.

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
kenzieschmoll: This isn't a value listenable because nothing needs to be rebuild when this changes. We are already rebuilding when the _tabController has changes. `selectedProfilerTabIndex` is only accessed by _CpuProfilerState one time in initState, so we don't need to listen to changes to it.

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -418,7 +418,8 @@ class PerformanceController
     }
 
     if (offlinePerformanceData.cpuProfileData != null) {
-      cpuProfilerController.loadData(offlinePerformanceData.cpuProfileData);
+      cpuProfilerController
+          .loadProcessedData(offlinePerformanceData.cpuProfileData);
```
bkonyi, line 422: Nit: trailing comma will format this more nicely... ;-)

packages/devtools_app/test/cpu_profiler_controller_test.dart
```diff
@@ -109,6 +110,68 @@ void main() {
...
+      expect(
+          controller.dataNotifier.value.cpuProfileRoot.toStringDeep(),
+          equals(
+            '''
+  all - children: 1 - excl: 0 - incl: 1
+    Frame1 - children: 1 - excl: 0 - incl: 1
+      Frame2 - children: 1 - excl: 0 - incl: 1
+        Frame4 - children: 0 - excl: 1 - incl: 1
+''',
+          ));
```
jacob314, line 160: nit: consider some trailing commas to force flutter style.

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
```diff
@@ -109,9 +130,40 @@ class CpuProfilerController with SearchControllerMixin<CpuStackFrame> {
...
+  }
+
+  Future<void> loadDataWithTag(String tag) async {
+    _userTagFilter.value = tag;
+
+    _dataNotifier.value = null;
+    _processingNotifier.value = true;
+
+    try {
+      assert(false);
```
bkonyi, line 146: I'm guessing this isn't supposed to be here?

packages/devtools_app/test/cpu_profiler_controller_test.dart
kenzieschmoll: done

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/profiler/cpu_profiler.dart
kenzieschmoll: this is what this code is doing. `items` above will always have one item because we are showing a `none` option to negate any filters and show all samples. Then after that item, we will add the items from `userTags`, so if `userTags` is empty, `items` will have length 1 with just the `none` option. If `userTags` has length 1, then the two options we'd want to show are the one user tag and the none item.

packages/devtools_app/lib/src/profiler/cpu_profiler.dart
kenzieschmoll: done

packages/devtools_app/lib/src/profiler/cpu_profile_transformer.dart
```diff
@@ -46,6 +46,23 @@ class CpuProfileTransformer {
     _stackFrameKeys = cpuProfileData?.stackFramesJson?.keys?.toList() ?? [];
     _stackFrameValues = cpuProfileData?.stackFramesJson?.values?.toList() ?? [];
 
+    // Initialize all stack frames before we start to for the tree.
+    for (int i = 0; i < _stackFramesCount; i++) {
+      final k = _stackFrameKeys[i];
```
bkonyi, line 51: Uber nit: I'm personally not a fan of single character variables for anything other than iterators. Maybe change these to `key` and `values`?

packages/devtools_app/lib/src/profiler/cpu_profiler.dart
```diff
@@ -1,11 +1,14 @@
 // Copyright 2019 The Chromium Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
+import 'dart:developer';
```
bkonyi, line 4: What's this import for?

bkonyi (commented):
Quick first pass

packages/devtools_app/lib/src/profiler/cpu_profiler.dart
kenzieschmoll: for accessing `UserTag.defaultTag`

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
kenzieschmoll: you are correct, removed

packages/devtools_app/lib/src/performance/performance_controller.dart
kenzieschmoll: done 😆

packages/devtools_app/lib/src/profiler/cpu_profile_transformer.dart
kenzieschmoll: done

PR #2994, opened 2021-05-07
Fix lints to unbreak build.

kenzieschmoll:


kenzieschmoll:
In fixing these lints, a "quick fix all" option for sorting directives alphabetically would have been really useful. FYI @bwilkerson 

Screen Shot 2021-05-07 at 9 45 12 AM

kenzieschmoll:
landing TBR to fix build

PR #2995, opened 2021-05-07
Port devtool_shared to null safety

kenzieschmoll:


packages/devtools_shared/lib/src/memory/event_sample.dart
```diff
@@ -80,10 +78,10 @@ class ExtensionEvent {
 
   factory ExtensionEvent.fromJson(Map<String, dynamic> json) =>
       ExtensionEvent.custom(
-        json['timestamp'] as int,
-        json['eventKind'] as String,
-        json['customEventName'] as String,
-        json['data'] as Map<String, Object>,
+        json['timestamp'] as int?,
```
terrylucas, line 81: Should the constructors (default and named) have the timestamp, eventKind and data be required so the the ? is necessary?

packages/devtools_shared/lib/src/memory/event_sample.dart
```diff
@@ -102,13 +100,13 @@ class ExtensionEvent {
       data == null &&
       customEventName == null;
 
-  final int timestamp;
+  final int? timestamp;
```
terrylucas, line 103: than these fields except the customEventName would just be final w/o the ?

packages/devtools_shared/lib/src/memory/heap_sample.dart
```diff
@@ -48,37 +48,37 @@ class HeapSample {
   /// Version of HeapSample JSON payload.
   static const version = 1;
 
-  final int timestamp;
+  final int? timestamp;
```
terrylucas, line 51: The default constructor arg should all be required, the ? isn't necessary?

packages/devtools_shared/lib/src/memory/memory_json.dart
```diff
@@ -25,11 +25,11 @@ abstract class MemoryJson<T> implements DecodeEncode<T> {
   /// List of HeapSample.
   MemoryJson.decode(
     String payloadName, {
-    String argJsonString,
-    Map<String, dynamic> argDecodedMap,
-  }) {
+    String? argJsonString,
+    Map<String, dynamic>? argDecodedMap,
+  }) : assert(argJsonString != null || argDecodedMap != null) {
```
terrylucas, line 30: Can we make named arg required?

packages/devtools_shared/lib/src/memory/event_sample.dart
kenzieschmoll: The `ExtensionEvent.empty` constructor initializes these values to null, so that is currently the only reason they are nullable

packages/devtools_shared/lib/src/memory/memory_json.dart
kenzieschmoll: should both argJsonString and argDecodedMap be required? or only argJsonString?

terrylucas (approved):
Just a couple of areas where I wonder if with need ? for arguments that are always required.

packages/devtools_shared/lib/src/memory/heap_space.dart
```diff
@@ -13,24 +13,24 @@ class HeapSpace {
...
         used = json['used'];
 
-  static HeapSpace parse(Map<String, dynamic> json) =>
+  static HeapSpace? parse(Map<String, dynamic>? json) =>
       json == null ? null : HeapSpace._fromJson(json);
 
   final Map<String, dynamic> json;
 
-  final double avgCollectionPeriodMillis;
+  final double? avgCollectionPeriodMillis;
```
kenzieschmoll, line 21: @terrylucas do any of these need to be nullable?

packages/devtools_shared/lib/src/memory/memory_json.dart
```diff
@@ -231,8 +234,8 @@ class AllocationMemoryJson extends MemoryJson<ClassHeapDetailStats> {
   /// Given a JSON string representing an array of HeapSample, decode to a
   /// List of HeapSample.
   AllocationMemoryJson.decode({
-    String argJsonString,
-    Map<String, dynamic> argDecodedMap,
+    String? argJsonString,
```
jacob314, line 237: this should probably be non-nullable and required as there isn't much point in trying to decode without specifying what you are decoding. I'm assuming that you lways need to pass argJsonString and that passing in just argDecodedMap isn't required.

packages/devtools_shared/lib/src/memory/heap_sample.dart
```diff
@@ -60,25 +60,25 @@ class HeapSample {
 
   final bool isGC;
 
-  EventSample _memoryEventInfo;
+  EventSample? _memoryEventInfo;
```
jacob314, line 63: this is another case where you want to use `late` to avoid spurious nullable types and `!`

jacob314 (approved):
lgtm
with some nits mainly about using `late` everywhere you have an object that is eventually non-null.

packages/devtools_shared/lib/src/memory/heap_sample.dart
kenzieschmoll: How will a late variable be used with the ??= operator below?

packages/devtools_shared/lib/src/memory/heap_sample.dart
kenzieschmoll: removed

packages/devtools_shared/lib/src/memory/event_sample.dart
kenzieschmoll: this has to be nullable because of the `empty()` constructor

packages/devtools_shared/lib/src/memory/memory_json.dart
kenzieschmoll: discussed offline - added a todo for @terrylucas

packages/devtools_shared/lib/src/memory/heap_sample.dart
kenzieschmoll: re-wrote these fields so that we don't need to use nullable types nor late

packages/devtools_shared/lib/src/memory/memory_json.dart
```diff
@@ -57,18 +60,18 @@ abstract class MemoryJson<T> implements DecodeEncode<T> {
     }
   }
 
-  int _payloadVersion;
+  int? _payloadVersion;
```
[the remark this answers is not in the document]
kenzieschmoll: used late final

packages/devtools_shared/lib/src/memory/memory_json.dart
kenzieschmoll: done

PR #3000, opened 2021-05-08
Remove the element count on classes and add item index in lists

### LINKED ISSUE #2987: Aligning provider and debugger object views
The provider view lists the number of "elements" for classes as well as collections. This doesn't seem useful and just adds noise. For example, the `(1 element)` for CatalogModel isn't useful as users don't care exactly how many fields classes have.

Cases where the provider screen is better:
displays hashcodes for objects (feature request for debugger).
Better color coding although risk colors are busy.
Using a variable width font is probably better.

Remove `(1 element) from here
Screen Shot 2021-05-06 at 10 50 24 AM

Consider aligning list display.
Screen Shot 2021-05-06 at 10 53 18 AM

List display in debugger view:
Comparison view in the debugger page:
Screen Shot 2021-05-06 at 10 52 58 AM

rrousselGit:
fixes #2987


jacob314 (commented):
Remove the open+close parens with no content inside for regular Dart objects.
image

PR #3006, opened 2021-05-12
Add toggle to allow showing/hiding internal properties

### LINKED ISSUE #3004: Hide data type internals on the provider page
In this example, we are exposing the internals of `_CompactLinkedHashSet`. Does exposing this data add any value for the user, and should we be allowing them to edit these values? 

Screen Shot 2021-05-11 at 2 10 49 PM

@rrousselGit @jacob314

rrousselGit:
fixes #3004 

This adds a toggle to show/hide internal properties inherited from dependencies:

Screenshot 2021-05-12 at 14 37 47

Screenshot 2021-05-12 at 14 38 20


jacob314:
I would match the behavior of the settings icon and dialog shown on the performance page rather than using the thumbprint icon. That will be consistent with the rest of devtools and there are a few other settings I expect we need to add.
For example: probably disable editing property values unless an experimental setting is enabled.

rrousselGit:
Done, this now uses a dialog with a setting button like in other pages:

Screenshot 2021-05-12 at 21 32 18

packages/devtools_app/lib/src/provider/provider_screen.dart
```diff
@@ -122,3 +137,42 @@ void showProviderErrorBanner(BuildContext context) {
...
   );
 }
+
+class _StateInspectorSettingsDialog extends ConsumerWidget {
+  @override
+  Widget build(BuildContext context, ScopedReader watch) {
+    final theme = Theme.of(context);
+
+    return DevToolsDialog(
+      title: dialogTitleText(theme, 'State inspector configurations'),
```
kenzieschmoll, line 147: put 'State inspector configurations' into a static const on this class and use it here as well as for the tooltip above

packages/devtools_app/lib/src/provider/provider_screen.dart
```diff
@@ -122,3 +137,42 @@ void showProviderErrorBanner(BuildContext context) {
...
+      content: Column(
+        mainAxisSize: MainAxisSize.min,
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          InkWell(
+            onTap: () => _toggleShowInternals(context),
+            child: Row(
+              children: [
+                Checkbox(
```
kenzieschmoll, line 156: can you use a NotifierCheckbox instead of watching and reading from the context? See https://github.com/flutter/devtools/blob/8c1ab6862af372b489fb791dff995279dee303b2/packages/devtools_app/lib/src/performance/performance_screen.dart#L441 for an example

jacob314 (changes requested):
lgtm

packages/devtools_app/test/instance_viewer/instance_viewer_test.dart
```diff
@@ -280,7 +382,9 @@ void main() {
           container: container,
           child: const MaterialApp(
             home: Scaffold(
-              body: InstanceViewer(rootPath: InstancePath.fromInstanceId('0')),
+              body: InstanceViewer(
+                  showInternalProperties: true,
+                  rootPath: InstancePath.fromInstanceId('0')),
```
kenzieschmoll, line 387: nit: trailing comma

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/provider/provider_screen.dart
rrousselGit: There is no ValueListenable involved

kenzieschmoll: I was suggesting that we both use a NotifierCheckbox and move the value to be in a valuenotifier on some controller. This is the pattern we use throughout the rest of DevTools and works nicely for managing state. Not blocking though, this can be done as a cleanup.

Issue #2978, opened 2021-05-04
Port devtools server and devtools shared to use null safety


jacob314:
This is blocking packages that depend on devtools_shared and devtools_server from running in sound null safety mode. For example, flutter_tools.

kenzieschmoll:
devtools_shared port here: https://github.com/flutter/devtools/pull/2995.
devtools_server port is blocked on package:sse and package:browser_launcher being migrated. @grouma @natebosch @kevmoo

kenzieschmoll:
package:browser_launcher is migrated and published 1.0.0. Now just blocked on package:sse

PR #3009, opened 2021-05-12
Migrate devtools_server to null safety

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/2978


packages/devtools_server/lib/src/external_handlers.dart
```diff
@@ -27,28 +27,32 @@ import 'server_api.dart';
 /// DevTools project.
 Future<shelf.Handler> defaultHandler(
   ClientManager clients, {
-  String customDevToolsPath,
+  String? customDevToolsPath,
   bool debugMode = false,
 }) async {
-  String buildDir = customDevToolsPath;
+  String? buildDir = customDevToolsPath;
```
jacob314, line 33: nit: rewrite s
final buildDir = customDevToolsPath ?? await _resolveBuildDir() to eliminate the null checks.

packages/devtools_server/lib/src/memory_profile.dart
```diff
@@ -58,68 +58,68 @@ class MemoryProfile {
...
-    return service.callMethod(
+    return service!.callMethod(
       registered.first,
       isolateId: isolateId,
       args: args,
     );
   }
 
-  Map<String, List<String>> get registeredMethodsForService =>
+  Map<String, List<String>?> get registeredMethodsForService =>
```
jacob314, line 99: seems odd that this is `List<String>?` instead of `List<String>`

packages/devtools_server/lib/src/external_handlers.dart
kenzieschmoll: fixed

packages/devtools_server/lib/src/memory_profile.dart
kenzieschmoll: fixed

packages/devtools_server/lib/src/memory_profile.dart
```diff
@@ -131,54 +131,53 @@ class MemoryProfile {
...
 
   void stopPolling() {
-    _pollingTimer?.cancel();
+    _pollingTimer.cancel();
     service = null;
   }
 
   Future<void> _pollMemory() async {
-    final VM vm = await service.getVM();
+    assert(service != null);
```
jacob314, line 154: cleanup by writing
final service = !this.service;
optionally: after each async block you could verify that service == this.service and abort otherwise as the service has changed so the memory polling request should be abandoned.

packages/devtools_server/lib/src/memory_profile.dart
```diff
@@ -131,54 +131,53 @@ class MemoryProfile {
...
       id: isolate.id,
       name: isolate.name,
       number: isolate.number,
       isSystemIsolate: isolate.isSystemIsolate,
     );
 
-    if (hasConnection &&
-        vm.operatingSystem == 'android' &&
-        _selectedIsolate != null) {
+    if (hasConnection && vm.operatingSystem == 'android') {
```
jacob314, line 180: nice. the migration caught one spurious null check.

packages/devtools_server/lib/src/memory_profile.dart
```diff
@@ -200,28 +199,30 @@ class MemoryProfile {
...
   /// @returns view id of selected isolate's 'FlutterView'.
   /// @throws Exception if no 'FlutterView'.
-  Future<String> getFlutterViewId(IsolateRef selectedIsolate) async {
-    final flutterViewListResponse = await callService(
+  Future<String?> getFlutterViewId(IsolateRef selectedIsolate) async {
+    final flutterViewListResponse = await (callService(
       registrations.flutterListViews,
       isolateId: selectedIsolate.id,
-    );
+    ) as FutureOr<Response>);
```
jacob314, line 223: move the as outside the await so it can just be `as Response`

packages/devtools_server/lib/src/memory_profile.dart
```diff
@@ -245,27 +247,27 @@ class MemoryProfile {
...
   ///   layerBytes - layer raster cache entries in bytes
   ///   pictureBytes - picture raster cache entries in bytes
-  Future<Response> getRasterCacheMetrics(IsolateRef selectedIsolate) async {
+  Future<Response?> getRasterCacheMetrics(IsolateRef selectedIsolate) async {
     final viewId = await getFlutterViewId(selectedIsolate);
 
     return await callService(
       registrations.flutterEngineRasterCache,
-      args: <String, String>{
+      args: <String, String?>{
```
jacob314, line 255: can we just omit the types of this literal and let dart infer it?

packages/devtools_server/lib/src/memory_profile.dart
```diff
@@ -245,27 +247,27 @@ class MemoryProfile {
...
 
     isolateHeaps.clear();
 
-    for (Isolate isolate in isolates) {
+    for (Isolate? isolate in isolates) {
       if (isolate != null) {
-        final List<HeapSpace> heaps = getHeaps(isolate).toList();
-        isolateHeaps[isolate.id] = heaps;
+        final List<HeapSpace?> heaps = getHeaps(isolate).toList();
+        isolateHeaps[isolate.id!] = heaps as List<HeapSpace>;
```
jacob314, line 270: this isn't right.  For this to make sense, getHeaps should just return a List<HeapSpace>
I would be worried this code as written will just throw errors at runtime.

packages/devtools_server/lib/src/memory_profile.dart
kenzieschmoll: changed to `final service = this.service!`

packages/devtools_server/lib/src/memory_profile.dart
kenzieschmoll: fixed getHeaps to exclude null HeapSpaces returned from HeapSpace.parse

packages/devtools_server/lib/src/service_registrations.dart
```diff
@@ -4,8 +4,8 @@
 
 class RegisteredServiceDescription {
   const RegisteredServiceDescription._({
-    this.service,
-    this.title,
+    /*required*/ required this.service,
```
jacob314, line 7: remove /*required*/

jacob314 (approved):
lgtm

packages/devtools_server/lib/src/service_registrations.dart
kenzieschmoll: done

PR #3013, opened 2021-05-13
Eval Console Autocomplete

grouma:
Bug fixes to the autocomplete logic:
- The optional decorator was never actually applied
- Keyboard events would not propagate if they weren't applicable to the autocomplete infrastructure
- Rebuilds could cause issue with focus and text state

New features:
- The evaluation input now uses the autocomplete infrastructure

Follow up work:
- The autocomplete overlay should be pinned to the `.` in tracking mode
- The mixin approach complicates composition. Long term we should model the autocomplete strictly as a stateful widget with the right hooks.

Although not blocking, this feature requires https://github.com/dart-lang/webdev/pull/1319 to be fully functional for Dart web applications.

https://user-images.githubusercontent.com/1840773/118082805-af158980-b372-11eb-8614-4506d85ab1c2.mov


packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -2,14 +2,19 @@
...
 
+import '../auto_dispose.dart';
+import '../auto_dispose_mixin.dart';
 import '../notifications.dart';
+import '../theme.dart';
+import '../ui/search.dart';
 import 'debugger_controller.dart';
 
-// TODO(devoncarew): We'll want some kind of code completion w/ eval.
 // TODO(devoncarew): We should insert eval result objects into the console as
```
jacob314, line 18: remove this TODO on line 13 as well as it is obsolete.

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -24,17 +29,142 @@ class ExpressionEvalField extends StatefulWidget {
...
-    textController = TextEditingController();
-    textFocus = FocusNode();
+    _autoCompleteController = _AutoCompleteController();
+  }
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+
+    addAutoDisposeListener(_autoCompleteController.searchNotifier, () {
```
jacob314, line 53: you need to cancel the existing listeners at the start of this method to avoid ending up with n unbounded number of listeners.
What dependency has changed? I'm not spotting why the autoCompleteController needs a new listener here instead of in initState.

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -24,17 +29,142 @@ class ExpressionEvalField extends StatefulWidget {
...
+        selection,
+        handleFields: isField,
+      );
+
+      // Only show pop-up if there's a real variable name or field.
+      if (parts.activeWord.isEmpty && !parts.isField) return;
+
+      final matches = await _matchesFor(parts);
+
+      final normalizedMatches = matches.toSet().toList()..sort();
```
jacob314, line 96: remove the toSet.toList as matches is already deduped.

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -24,17 +29,142 @@ class ExpressionEvalField extends StatefulWidget {
...
+        // Display the response to the user.
+        if (response is InstanceRef) {
+          final Instance instance = await widget.controller.getObject(response);
+          result.addAll(await _autoCompleteProperties(instance.classRef));
+          // TODO(grouma) - This shouldn't be necessary but package:dwds does
+          // not properly provide superclass information.
+          result.addAll(instance.fields.map((field) => field.decl.name));
+          result.removeWhere((prop) => !prop.startsWith(parts.activeWord));
+        }
+      } catch (_) {}
```
jacob314, line 133: move catch just round the the async calls that are expected to possibly throw.

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -24,17 +29,142 @@ class ExpressionEvalField extends StatefulWidget {
...
+      result.addAll(await _autoCompleteProperties(clazz.superClass));
+      result.removeWhere((member) => !_isAccessible(member, clazz));
+    }
+    return result;
+  }
+
+  bool _validFunction(FuncRef funcRef, Class clazz) {
+    return !funcRef.isStatic &&
+        !_isContructor(funcRef, clazz) &&
+        funcRef.name != '==';
```
jacob314, line 156: filter all operators not just ==
add a helper method isOperator somewhere.

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -24,17 +29,142 @@ class ExpressionEvalField extends StatefulWidget {
...
+  }
+
+  bool _isContructor(FuncRef funcRef, Class clazz) =>
+      funcRef.name == clazz.name || funcRef.name.startsWith('${clazz.name}.');
+
+  bool _isAccessible(String member, Class clazz) {
+    final frame = widget.controller.selectedStackFrame.value?.frame ??
```
jacob314, line 163: why is the `??` needed. Can we just always have the first frame be selected? Seems like that is the UI state in the debugger anyway.

packages/devtools_app/lib/src/debugger/evaluate.dart
grouma: Oversight. Moved to initState.

packages/devtools_app/lib/src/debugger/evaluate.dart
grouma: This is a bit of copied code from the evaluation logic. I'm not exactly sure why it's required but the added defensiveness seemed important.

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -24,17 +26,71 @@ class ExpressionEvalField extends StatefulWidget {
   _ExpressionEvalFieldState createState() => _ExpressionEvalFieldState();
 }
 
-class _ExpressionEvalFieldState extends State<ExpressionEvalField> {
-  TextEditingController textController;
-  FocusNode textFocus;
+class _AutoCompleteController extends DisposableController
+    with SearchControllerMixin, AutoCompleteSearchControllerMixin {}
```
kenzieschmoll, line 30: should this be public and in the search.dart file?

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: I would add it to the controller. Maybe "currentStackFrame".

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -349,3 +283,69 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
+        result.removeWhere((prop) => !prop.startsWith(parts.activeWord));
+      }
+    } catch (_) {}
+  }
+  return result.toList();
+}
+
+Future<List<String>> _autoCompleteMembersFor(
+    ClassRef classRef, DebuggerController controller) async {
+  final result = <String>[];
```
jacob314, line 313: nit: make this return a Set<String> as well as you are currently adding dupes.

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -349,3 +283,69 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
+    result.addAll(await _autoCompleteMembersFor(clazz.superClass, controller));
+    result.removeWhere((member) => !_isAccessible(member, clazz, controller));
+  }
+  return result;
+}
+
+bool _validFunction(FuncRef funcRef, Class clazz) {
+  return !funcRef.isStatic &&
+      !_isContructor(funcRef, clazz) &&
+      funcRef.name != '==';
```
jacob314: need to detect other operators or you will suggest them as invalid autocompletes because the syntax is
`foo+bar` not `foo.+bar`

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -349,3 +283,69 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
+}
+
+bool _isContructor(FuncRef funcRef, Class clazz) =>
+    funcRef.name == clazz.name || funcRef.name.startsWith('${clazz.name}.');
+
+bool _isAccessible(String member, Class clazz, DebuggerController controller) {
+  final frame = controller.selectedStackFrame.value?.frame ??
+      controller.stackFramesWithLocation.value.first.frame;
+  final currentScript = frame.location.script;
+  return !(member.startsWith('_') &&
```
jacob314, line 362: nit:  I think
`!member.startsWith('_') || currentScript.id == clazz.location?.script?.id`
is a little easier to read.

packages/devtools_app/test/evaluate_test.dart
```diff
@@ -0,0 +1,191 @@
...
+                  name: 'field2',
+                  owner: null,
+                  declaredType: null,
+                  isConst: false,
+                  isFinal: false,
+                  isStatic: false,
+                  id: ''),
+            ],
+            functions: [
+              FuncRef(
```
jacob314, line 97: Add a couple constructors to make sure they are filtered out.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -167,10 +268,97 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField> {
...
+        result.addAll(instance.fields.map((field) => field.decl.name));
+      }
+    } catch (_) {}
+  }
+  return result.where((name) => name.startsWith(parts.activeWord)).toList()
+    ..sort();
+}
+
+Future<List<String>> _autoCompleteMembersFor(
+    ClassRef classRef, DebuggerController controller) async {
```
kenzieschmoll, line 312: nit: trailing comma

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -21,15 +21,12 @@ const defaultTopMatchesLimit = 10;
...
-  // Initial values for searching (richness primarily auto-complete).
-  TextEditingValue searchTextFieldValue = const TextEditingValue();
-
   final _searchNotifier = ValueNotifier<String>('');
 
   /// Notify that the search has changed.
   ValueListenable get searchNotifier => _searchNotifier;
 
-  bool isField = false;
+  bool get isField => search.endsWith('.');
```
kenzieschmoll, line 29: this seems specific to the debugger and not something that should be in a general Search mixin. Can we move this to the debugger?

packages/devtools_app/test/evaluate_test.dart
```diff
@@ -0,0 +1,205 @@
...
+    when(debuggerController.variables).thenReturn(
+      ValueNotifier(
+        [
+          Variable.create(
+            BoundVariable(
+                name: 'foo',
+                value: null,
+                declarationTokenPos: 0,
+                scopeStartTokenPos: 0,
+                scopeEndTokenPos: 0),
```
kenzieschmoll, line 41: trailing comma for all these BoundVariable, FieldRef, Class, FuncRef, etc. instances

packages/devtools_app/test/evaluate_test.dart
```diff
@@ -0,0 +1,205 @@
+// Copyright 2021 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: nit: can we rename this file to `debugger_evalution_test.dart` so it is clear that this test is testing functionality in the debugger

grouma:
Merging knowing that we'll need to update the goldens one more time. Currently failing with a `0.02%` diff detected.

Issue #3015, opened 2021-05-13
Settings icons are not centered on desktop or web.


jacob314:
This impacts all screens with settings icons. It may have regressed on current flutter dev channel.
![Screen Shot 2021-05-13 at 10 20 16 AM](https://user-images.githubusercontent.com/1226812/118165014-a22b8100-b3d8-11eb-8188-f0819e6faf8c.png)
![Screen Shot 2021-05-13 at 10 22 27 AM](https://user-images.githubusercontent.com/1226812/118165045-aeafd980-b3d8-11eb-95ed-0eba6622f212.png)

PR #3016, opened 2021-05-13
Fix bug where settings widgets were not centered by using a shared widget

jacob314:
Fixes https://github.com/flutter/devtools/issues/3015

Screen Shot 2021-05-13 at 10 50 18 AM

### DIFF
--- a/packages/devtools_app/lib/src/common_widgets.dart
+++ b/packages/devtools_app/lib/src/common_widgets.dart
@@ -254,30 +254,24 @@ class StopRecordingButton extends StatelessWidget {
 class SettingsOutlinedButton extends StatelessWidget {
   const SettingsOutlinedButton({
     @required this.onPressed,
-    @required this.tooltip,
+    @required this.label,
   });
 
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -254,30 +254,24 @@ class StopRecordingButton extends StatelessWidget {
...
-        tooltip: tooltip,
-        child: OutlinedButton(
-          onPressed: onPressed,
-          child: const Icon(
-            Icons.settings,
-            size: defaultIconSize,
-          ),
-        ),
-      ),
+    return IconLabelButton(
```
kenzieschmoll, line 266: why did we remove the tooltip?

jacob314: tooltip became label

kenzieschmoll (approved):

kenzieschmoll:
This CL should be completely independent of the bot failures - I think this is safe to land event with red bots.

PR #3027, opened 2021-05-14
Migrate ansi_up to null safety

kenzieschmoll:


helin24 (approved):

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -6,18 +6,17 @@
...
 
-  String _text;
+  late String _text;
   bool bold;
-  List<List<AnsiUpColor>> ansiColors;
-  List<AnsiUpColor> palette256;
-  AnsiUpColor fg;
-  AnsiUpColor bg;
-  RegExp _csiRegex;
+  late List<List<AnsiUpColor>> ansiColors;
```
jacob314, line 15: try to avoid `late` when it is not needed as it has some performance overhead and gives you less static checking than other non-nullable types. In this case, _setupPlettes can just be inlined in the constructor to void the late.

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -297,8 +299,8 @@ enum PacketKind {
   OSCURL,
 }
 
-class TextPacket {
-  TextPacket({this.kind, this.text, this.url});
+class _TextPacket {
```
jacob314, line 302: why was this made private?

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -345,8 +347,8 @@ class StyledText {
       return '';
     }
     return <String>[
-      if (bgColor != null) 'background-color: ${_colorToCss(bgColor)}',
-      if (fgColor != null) 'color: ${_colorToCss(fgColor)}',
+      if (bgColor != null) 'background-color: ${_colorToCss(bgColor!)}',
```
jacob314, line 350: for these cases the recommended pattern is the slightly verbose
final bgColor = bgColor;
final fgColor = fgColor;
return [
  if (bgColor != null) 'background-color: ${_colorToCss(bgColor!)}',
...
];
Which avoids the extra `!`

kenzieschmoll: I don't see the difference between the code and what you wrote? bgColor is still being unwrapped?

third_party/packages/ansi_up/lib/ansi_up.dart
kenzieschmoll: From what I could tell, it looked like a class internal to this package. I can change back to public if that is incorrect

third_party/packages/ansi_up/lib/ansi_up.dart
kenzieschmoll: done

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -247,9 +250,9 @@ class AnsiUp {
             }
           }
           if (modeCmd == '2' && index + 2 < sgrCmds.length) {
-            final r = int.tryParse(sgrCmds[index++], radix: 10);
-            final g = int.tryParse(sgrCmds[index++], radix: 10);
-            final b = int.tryParse(sgrCmds[index++], radix: 10);
+            final r = int.tryParse(sgrCmds[index++], radix: 10)!;
+            final g = int.tryParse(sgrCmds[index++], radix: 10)!;
+            final b = int.tryParse(sgrCmds[index++], radix: 10)!;
```
jacob314, line 255: the original Dart code had a bug when porting from JS. The intent is to skip treating as a color if the input is invalid not to blow up.
Remove the `!` and add
`r != null && g != null && b != null`
to the checks.
bonus points, add a test that verifies it does not blowup on invalid input.

third_party/packages/ansi_up/lib/ansi_up.dart
jacob314: leave it as private. I think you are right.

third_party/packages/ansi_up/lib/ansi_up.dart
jacob314: My fault. I left out the `!` that can be removed in the example.
@pq any plants to add a lint to suggest quick fixes that add local variables to avoid an extra `!`?
```dart
final bgColor = bgColor;
final fgColor = fgColor;
return [
  if (bgColor != null) 'background-color: ${_colorToCss(bgColor)}',
...
];
```

PR #3038, opened 2021-05-17
Re-write update_version script in Dart.

kenzieschmoll:
Writing this script in dart will make the script easier to iterate upon and maintain


tool/update_version.dart
```diff
@@ -0,0 +1,156 @@
...
+// This script must be executed from the top level devtools/ directory.
+
+void main(List<String> args) async {
+  final pubspecs = [
+    File('packages/devtools/pubspec.yaml'),
+    File('packages/devtools_app/pubspec.yaml'),
+    File('packages/devtools_server/pubspec.yaml'),
+    File('packages/devtools_shared/pubspec.yaml'),
+    File('packages/devtools_testing/pubspec.yaml'),
+  ];
```
jacob314, line 13: nit: create a List<String> and then map it to a List<File>

kenzieschmoll: done

tool/update_version.dart
```diff
@@ -0,0 +1,156 @@
...
+  final versionString = '## $version';
+  if (lines.first.endsWith(versionString)) {
+    print('Changelog already has an entry for version $version');
+    return;
+  }
+  changelog.writeAsString([
+    '$versionString\n',
+    'TODO: update changelog',
+    ...lines,
+  ].joinWithNewLine());
```
jacob314, line 134: seems like this adds an extra trailing newline?

High level, consider https://github.com/dart-lang/pubspec_parse
rather than using regexp.

kenzieschmoll: it does. We want a new line at the end of the file.

jacob314:
lgtm
Add a TODO to consider rewriting to use https://github.com/dart-lang/pubspec_parse if we make changes to this.

PR #3045, opened 2021-05-20
Improve eval autocomplete

grouma:
- Don't change focus with tab as that can be frustrating, especially in slow autocomplete environments
  - Note that tab to select is already supported
- Prevent showing inaccessible members from a superclass on the web
- Swallow `getObject` issues in the recursive walk of the class hierarchy.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/evaluate.dart
+++ b/packages/devtools_app/lib/src/debugger/evaluate.dart
@@ -296,7 +296,8 @@ Future<List<String>> autoCompleteResultsFor(
         );
         // TODO(grouma) - This shouldn't be necessary but package:dwds does
         // not properly provide superclass information.
-        result.addAll(instance.fields.map((field) => field.decl.name));
+        result.addAll(instance.fields.map((field) => field.decl.name).where(
+            (member) => !_isAccessible(member, instance.classRef, controller)));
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -296,7 +296,8 @@ Future<List<String>> autoCompleteResultsFor(
         );
         // TODO(grouma) - This shouldn't be necessary but package:dwds does
         // not properly provide superclass information.
-        result.addAll(instance.fields.map((field) => field.decl.name));
+        result.addAll(instance.fields.map((field) => field.decl.name).where(
+            (member) => !_isAccessible(member, instance.classRef, controller)));
```
kenzieschmoll, line 300: Should this be negated? Wouldnt we want to add results that are accessible?

kenzieschmoll (approved):
Lgtm with one comment

packages/devtools_app/lib/src/debugger/evaluate.dart
grouma: Yup! A test caught it thankfully. I also expanded this with another test. Also there was surprisingly a runtime type error.

Issue #3008, opened 2021-05-12
Colors in the performance page frame chart are confusing


jacob314:
![image](https://user-images.githubusercontent.com/1226812/118020234-b6a34700-b30e-11eb-824e-944e5f83bec4.png)
In this trace it looks like the raster thread janked even though it didn't because it is colored in red.
We should tweak the chart so that the blue shades are always used for parts of frames that didn't jank.
This matters now tht we are more focused on performance issues that only impact the raster or ui threads.

PR #1035, opened 2019-09-10
Create devtools_testing package for sharing integration tests

jakemac53:
Towards https://github.com/dart-lang/webdev/issues/633. Actually I think only flutter/flutter will use these though.

I was able to validate that given all these changes I could run tests using webdev/dwds by creating my own FlutterTestDriver and passing that in through my own FlutterTestEnvironment.

There is a lot to digest here but basically this PR:

- Creates a new devtools_testing package which can either be published on pub or depended on via a git constraint.
- Moves all the flutter integration tests and testing utilities in devtools into this new package
  - Each test gets its own entrypoint under lib, and a top level function that is identical to the old `main` except that it takes in a `FlutterTestEnvironment` argument.
  - The old test files in devtools now import and call these functions with the same environment as before.
- Moves all the test fixtures and goldens under the `devtools_testing` package, and updates the matchers to look for them in that package (they now have to live in a package so other packages can get them from pub).
  - The FlutterTestEnvironment figures out and sets the root devtools_testing package directory using a PackageResolver. This is a bit funky that it sets this up, but it basically has to be this way. The matchers can't do async code and resolving a package location is async.
 
Feel free to push back strongly - I know this seems a bit drastic but I think it is the easiest way to actually share the tests without copying them so they stay up to date.

PR #3049, opened 2021-05-21
Fix flutter frame coloring to reflect which part of the frame is janky.

kenzieschmoll:
Screen Shot 2021-05-21 at 10 32 57 AM

Fixes https://github.com/flutter/devtools/issues/3008

### DIFF
--- a/packages/devtools_testing/lib/support/performance_test_data.dart
+++ b/packages/devtools_testing/lib/support/performance_test_data.dart
@@ -71,6 +71,30 @@ final jankyFrame = FlutterFrame('jankyFrame')
   ..eventFlows[1] = (goldenRasterTimelineEvent.deepCopy()
     ..time = (TimeRange()
       ..start = const Duration(milliseconds: 68)
+      ..end = const Duration(milliseconds: 88)));
+
+final jankyFrameUiOnly = FlutterFrame('jankyFrameUiOnly')
+// ignore: invalid_use_of_visible_for_testing_member
+  ..eventFlows[0] = (goldenUiTimelineEvent.deepCopy()
+    ..time = (TimeRange()
+      ..start = const Duration(milliseconds: 50)
+      ..end = const Duration(milliseconds: 70)))
+// ignore: invalid_use_of_visible_for_testing_member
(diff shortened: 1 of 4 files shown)


packages/devtools_testing/lib/support/performance_test_data.dart
```diff
@@ -71,6 +71,30 @@ final jankyFrame = FlutterFrame('jankyFrame')
   ..eventFlows[1] = (goldenRasterTimelineEvent.deepCopy()
     ..time = (TimeRange()
       ..start = const Duration(milliseconds: 68)
+      ..end = const Duration(milliseconds: 88)));
+
+final jankyFrameUiOnly = FlutterFrame('jankyFrameUiOnly')
+// ignore: invalid_use_of_visible_for_testing_member
```
grouma, line 77: Curious why this is required.

grouma (approved):

packages/devtools_testing/lib/support/performance_test_data.dart
kenzieschmoll: I think it's because I'm using a visibleForTesting method `deepCopy` from a different package. The test data is in devtools_testing but the model class is in devtools_app. I would be in favor of moving all the devtools_testing code back into devtools_app - I don't think it was ever used as intended. @jakemac53 do you recall why it was moved into a separate package to begin with and if it is still useful to be kept separate?

jakemac53: The [original PR](https://github.com/flutter/devtools/pull/1035) has a pretty thorough explanation. There was a goal to share integration tests between devtools and webdev it sounds like? I have no idea if this is still relevant.

kenzieschmoll: Given the original PR was almost 2 years ago and devtools_app is still the only package that depends on devtools_testing, I think we could simplify things by bringing the devtools_testing code back into devtools_app. Plus, once we are no longer shipping on pub, I don't know that the devtools_testing package will have any use anyway @jacob314 @grouma thoughts?

grouma: Sounds good to me.

Issue #3050, opened 2021-05-21
Debugger clean up


kenzieschmoll:
eval text field needs a different outline color for light theme:
<img width="899" alt="Screen Shot 2021-05-21 at 10 55 32 AM" src="https://user-images.githubusercontent.com/43759233/119178897-260dea80-ba23-11eb-9ea4-776377dec28c.png">

gutter needs a right side outline to match the rest of the view
<img width="309" alt="Screen Shot 2021-05-21 at 10 56 17 AM" src="https://user-images.githubusercontent.com/43759233/119178949-332ad980-ba23-11eb-8c4d-fa485a7bb7ae.png">

PR #3052, opened 2021-05-21
Cleanup some UI in the debugger page

kenzieschmoll:
- Add right side border to gutter to match rest of page
- Remove Eval field border. The double border looked a little busy. This is what it looks like now.
Screen Shot 2021-05-21 at 11 28 36 AM
Screen Shot 2021-05-21 at 11 28 44 AM

Fixes https://github.com/flutter/devtools/issues/3050


jacob314 (approved):
lgtm

Issue #2482, opened 2020-11-02
Refactor InspectorTreeControllerFlutter to not use data that should be part of a theme


helin24:
This is related to trying to thread font size through DevTools components (https://github.com/flutter/devtools/pull/2459). `InspectorTreeControllerFlutter` currently uses a row height constant that we'd like to be able to get from a theme instead.

Potentially also related classes (ones that I encountered while trying to thread a theme to the first file):
- `InspectorController`
- `LoggingController`

helin24:
We decided to use a global variable for font size instead: https://github.com/flutter/devtools/pull/3054

I'm not sure if this could be useful for other cases but closing for now since the original use case no loner applies.

PR #3054, opened 2021-05-22
Use font size from IDE theme

helin24:
https://github.com/flutter/devtools/pull/2459 was the original attempt to implement variable font size (see for more context), but this PR uses a global variable set from main instead of setting a factor based on theme later. We were originally going to refactor some controller classes to enable passing font size through (see https://github.com/flutter/devtools/issues/2482), but this is easier and seems like a reasonable method.

### DIFF
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -292,9 +292,9 @@ class DevToolsAppState extends State<DevToolsApp> {
     return MaterialApp.router(
       debugShowCheckedModeBanner: false,
       theme: themeFor(
-          isDarkTheme: isDarkThemeEnabled,
-          ideTheme: ideTheme,
-          theme: Theme.of(context),
+        isDarkTheme: isDarkThemeEnabled,
+        ideTheme: ideTheme,
+        theme: Theme.of(context),
       ),
       builder: (context, child) => Notifications(child: child),
       routerDelegate: DevToolsRouterDelegate(_getPage),
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -64,12 +64,10 @@ bool isExternalBuild = true;
 class DevToolsApp extends StatefulWidget {
   const DevToolsApp(
     this.screens,
-    this.ideTheme,
     this.analyticsProvider,
   );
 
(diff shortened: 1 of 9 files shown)


packages/devtools_app/lib/src/config_specific/ide_theme/ide_theme.dart
```diff
@@ -15,4 +15,8 @@ class IdeTheme {
   Color backgroundColor;
   Color foregroundColor;
   double fontSize;
+
+  double fontSizeFactor() {
```
jacob314, line 19: nit: make this a getter as it is cheap to compute. While you are at it, make the other fields of this class final as they should have always been final.

packages/devtools_app/lib/src/inspector/inspector_tree.dart
```diff
@@ -42,8 +43,8 @@ extension InspectorColorScheme on ColorScheme {
 const double iconPadding = 5.0;
 const double chartLineStrokeWidth = 1.0;
 const double columnWidth = 16.0;
-const double verticalPadding = 10.0;
-const double rowHeight = 24.0;
+double get verticalPadding => 10.0 * ideTheme.fontSizeFactor();
```
jacob314, line 46: For discussion: it might be a good idea to round the verticalPadding and row height o the nearest px to ensure the ux is always in whole number of pixels unless there is a strong reason not to.

helin24: That sounds reasonable. I added a util function for this. Do you think it's necessary to restrict font sizes to reasonable min/max numbers?

packages/devtools_app/lib/src/app.dart
```diff
@@ -296,7 +291,10 @@ class DevToolsAppState extends State<DevToolsApp> {
   Widget build(BuildContext context) {
     return MaterialApp.router(
       debugShowCheckedModeBanner: false,
-      theme: themeFor(isDarkTheme: isDarkThemeEnabled, ideTheme: ideTheme),
+      theme: themeFor(
+          isDarkTheme: isDarkThemeEnabled,
+          ideTheme: ideTheme,
+          theme: Theme.of(context)),
```
kenzieschmoll, line 297: nit: trailing comma

kenzieschmoll (approved):

PR #3055, opened 2021-05-24
Make a copy of performance page files under performance/legacy

kenzieschmoll:
We will have to support two versions of the performance page: the current implementation that calculates flutter frames based off of timeline events, and a new version that uses the flutter FrameTiming API. Until the engine and framework changes related to flutter frame ids in timeline events and flutter.frame events are present in flutter stable (https://github.com/flutter/flutter/commit/78a96b09d64dc2a520e5b269d5cea1b9dde27d3f is the latest), we need to support both of these impls in the performance page.

No functional changes here, just copying current code into legacy/ to prepare for the migration to using the FrameTiming api.


jacob314 (approved):
lgtm

Issue #2918, opened 2021-04-19
Migrate flutter frames chart (performance page) to use data from FrameTiming API


kenzieschmoll:
Flutter.Frame events are [sent from the Flutter Framework](https://github.com/flutter/flutter/blob/41507a0408cccfd3d068bee163a8bafbaa5ecf28/packages/flutter/lib/src/scheduler/binding.dart#L1105) and they contain information from a [FrameTiming](https://api.flutter.dev/flutter/dart-ui/FrameTiming-class.html) object. We should calculate flutter frame information in the performance page from this set of data. Flutter.Frame events come over the vm extension stream, so we have access to event history and can pull all available frames from that stream.

This will allow the flutter frames information to have "live" data, which is something we know users want based on studies and feedback.

kenzieschmoll:
This issue is blocked on https://github.com/flutter/flutter/issues/80735, as we don't currently have a way to associate a Flutter.Frame event with its corresponding timeline events.

PR #3056, opened 2021-05-25
Show LegacyPerformanceScreen or PerformanceScreen based on the current flutter version

kenzieschmoll:
This is work towards #2918.

A couple additional files (performance_screen.dart and event_details.dart) were copied over into the performance/legacy dir.

### DIFF
--- a/packages/devtools_app/lib/src/version.dart
+++ b/packages/devtools_app/lib/src/version.dart
@@ -101,6 +101,22 @@ class SemanticVersion implements Comparable {
   bool isSupported({@required SemanticVersion supportedVersion}) =>
       compareTo(supportedVersion) >= 0;
 
+  bool operator <(other) {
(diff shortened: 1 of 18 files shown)


packages/devtools_app/lib/src/version.dart
```diff
@@ -101,6 +101,22 @@ class SemanticVersion implements Comparable {
   bool isSupported({@required SemanticVersion supportedVersion}) =>
       compareTo(supportedVersion) >= 0;
 
+  bool operator <(other) {
```
jacob314, line 104: there should be some mixin that derives all of these from compareTo for. If not, you should add one.

kenzieschmoll: added mixin

jacob314 (approved):
lgtm

PR #3057, opened 2021-05-25
Prepare for v2.2.5 release.

terrylucas:


packages/devtools/pubspec.yaml
```diff
@@ -8,16 +8,16 @@ description: A suite of web-based performance tooling for Dart and Flutter.
...
-version: 2.2.4
+version: 2.2.5
 
 homepage: https://github.com/flutter/devtools
 
 environment:
   sdk: '>=2.3.0 <3.0.0'
 
 dependencies:
-  devtools_server: 2.3.0
```
kenzieschmoll, line 19: We are downgrading the deps here. If you need to make a patch release, cherry pick onto the 2.2.4 tag

terrylucas: I'm going to turn on the provider page for google3.  I thought it would be easier to create pull the code in and remove
`// Google3-Only`
comments and the 4 or 5 lines that disabled the provider page.

This is only for Google3 not public.

terrylucas: The downgrade must have happened with the tool/update_version.dart

terrylucas: What is the patching mechanism? Update to 2.2.4+1?

PR #3058, opened 2021-05-25
Use ListValueNotifier for filtered data

kenzieschmoll:
Please review with hide whitespace diffs.

### DIFF
--- a/packages/devtools_app/lib/src/network/network_controller.dart
+++ b/packages/devtools_app/lib/src/network/network_controller.dart
@@ -382,11 +382,10 @@ class NetworkController
   void filterData(QueryFilter filter) {
     serviceManager.errorBadgeManager.clearErrors(NetworkScreen.id);
     if (filter == null) {
-      final requests = List<NetworkRequest>.from(_requests.value.requests);
-      requests.forEach(_checkForError);
+      _requests.value.requests.forEach(_checkForError);
       filteredData
         ..clear()
-        ..addAll(requests);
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -384,49 +384,53 @@ class NetworkController
...
+            return false;
           }
-          return false;
-        }
-        _checkForError(r);
-        return true;
-      }).toList();
+          _checkForError(r);
+          return true;
+        }).toList());
```
jacob314, line 433: the `toList()`
isn't needed.

packages/devtools_app/lib/src/network/network_controller.dart
```diff
@@ -384,49 +384,53 @@ class NetworkController
     if (filter == null) {
       final requests = List<NetworkRequest>.from(_requests.value.requests);
```
jacob314, line 385: you can avoid this List.from

jacob314 (approved):
lgtm
This change is fine and it is nice to be consistent using ListValueNotifier but in this case there isn't a huge amount of benefit as I don't see any cases where elements are being incrementally added to the list.

packages/devtools_app/lib/src/network/network_controller.dart
kenzieschmoll: without toList() an error like `'WhereIterable<LogData>' is not a subtype of type 'List<LogData>'` is thrown

Issue #3062, opened 2021-05-25
Debug hover card is not scrollable


grouma:
Highly nested variables can take up a lot of space and extend off screen. The hover card should have a max size and support scrolling to prevent this issue.

grouma:
cc @terrylucas @kenzieschmoll

PR #3065, opened 2021-05-26
Hover fixes

grouma:
- Increase delay in showing the hover card (as requested by some internal users)
- Ensure only one hover card is displayed at a time
- Make the hover card scrollable for large inspected variables

Closes https://github.com/flutter/devtools/issues/3062

### DIFF
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -690,10 +690,15 @@ class _LineItemState extends State<LineItem> {
           if (_hasMouseExited) return;
           _hoverCard?.remove();
           _hoverCard = HoverCard(
-            contents: Material(
-              child: ExpandableVariable(
-                debuggerController: _debuggerController,
-                variable: ValueNotifier(variable),
+            contents: SingleChildScrollView(
+              child: Container(
+                constraints: const BoxConstraints(maxHeight: 250.0),
+                child: Material(
+                  child: ExpandableVariable(
+                    debuggerController: _debuggerController,
+                    variable: ValueNotifier(variable),
+                  ),
+                ),
               ),
             ),
             event: event,
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -633,6 +633,7 @@ class LineItem extends StatefulWidget {
   }) : super(key: key);
 
   static const _hoverDelay = Duration(milliseconds: 150);
+  static const _removeDelay = Duration(milliseconds: 50);
   static const _hoverWidth = 400.0;
 
   final TextSpan lineContents;
@@ -662,7 +663,7 @@ class _LineItemState extends State<LineItem> {
   void _onHoverExit() {
     _showTimer?.cancel();
     _hasMouseExited = true;
-    _removeTimer = Timer(LineItem._hoverDelay, () {
+    _removeTimer = Timer(LineItem._removeDelay, () {
       _hoverCard?.maybeRemove();
       _previousHoverWord = '';
     });
@@ -692,7 +693,8 @@ class _LineItemState extends State<LineItem> {
           _hoverCard = HoverCard(
             contents: SingleChildScrollView(
               child: Container(
-                constraints: const BoxConstraints(maxHeight: 250.0),
+                constraints:
+                    const BoxConstraints(maxHeight: maxHoverCardHeight),
                 child: Material(
                   child: ExpandableVariable(
                     debuggerController: _debuggerController,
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -632,7 +632,7 @@ class LineItem extends StatefulWidget {
     this.activeSearchMatch,
   }) : super(key: key);
 
-  static const _hoverDelay = Duration(milliseconds: 50);
+  static const _hoverDelay = Duration(milliseconds: 150);
```
jacob314, line 635: why is this better?

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -684,6 +687,8 @@ class _LineItemState extends State<LineItem> {
           if (response is! InstanceRef) return;
           final variable = Variable.fromRef(response);
           await _debuggerController.buildVariablesTree(variable);
+          if (_hasMouseExited) return;
```
jacob314, line 690: we should do more.  _onHoverExit only gets called if you changed the line. If the user moved more than a couple pixels within the same line, this hover is also obsolete.
we need to verify that we haven't gotten a mouseMove to a different location within the line as opposed to an exit from the line.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -657,9 +657,11 @@ class _LineItemState extends State<LineItem> {
   DebuggerController _debuggerController;
 
   String _previousHoverWord = '';
+  bool _hasMouseExited = false;
 
   void _onHoverExit() {
     _showTimer?.cancel();
+    _hasMouseExited = true;
     _removeTimer = Timer(LineItem._hoverDelay, () {
```
jacob314, line 665: this should be a separate delay. I don't think this delay should be made longer.

packages/devtools_app/lib/src/debugger/codeview.dart
grouma: We don't want to show the hover card too quickly. This _feels_ about right.

packages/devtools_app/lib/src/debugger/codeview.dart
grouma: We already handle that scenario. If the hoverWord is the same we don't change the hover card just like in Chrome.

packages/devtools_app/lib/src/debugger/codeview.dart
grouma: I disagree. We are less likely to "flash" a hover card if a user accidentally moves outside the line.

kenzieschmoll: I agree with jacob. Look at the difference between the hover card show and hover card hide delays in IntelliJ. There is a longer delay to show than exit.

hover-intellij

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -684,11 +687,18 @@ class _LineItemState extends State<LineItem> {
...
+          if (_hasMouseExited) return;
+          _hoverCard?.remove();
           _hoverCard = HoverCard(
-            contents: Material(
-              child: ExpandableVariable(
-                debuggerController: _debuggerController,
-                variable: ValueNotifier(variable),
+            contents: SingleChildScrollView(
+              child: Container(
+                constraints: const BoxConstraints(maxHeight: 250.0),
```
kenzieschmoll, line 695: make 250.0 a const var on the theme. `maxHoverCardHeight` or similar

packages/devtools_app/lib/src/debugger/codeview.dart
grouma: I can shorten the delay but we don't want it too small as we want to allow users to hover "into" the card which keeps it around. Otherwise it's frustrating to expand variables displayed in the hover card.

packages/devtools_app/lib/src/debugger/codeview.dart
grouma: Done. PTAL.

packages/devtools_app/lib/src/debugger/codeview.dart
jacob314: i would hope this code path isn't needed to make the case of hovering into a card work. ideally the area treated as in the card and the current location should have a border so you don't need to rely on the delay to move over the hover crd.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/debugger/codeview.dart
grouma: That is not the case. Note that the card is rendered below the source line which is why this is required. This also mirrors the Chrome DevTools behavior. There appears to be a grace period where you can move your mouse outside the hovercard and into before it is removed.

PR #3066, opened 2021-05-27
Stop checking flutter version for connected app in flutter web apps

kenzieschmoll:
This also makes the version checking code more safe by ensuring that the flutter version service is available before attempting to make a call to it.

### DIFF
--- a/packages/devtools_app/lib/src/connected_app.dart
+++ b/packages/devtools_app/lib/src/connected_app.dart
@@ -4,9 +4,12 @@
 
 import 'dart:async';
 
+import 'package:flutter/foundation.dart';
+
 import 'config_specific/logger/logger.dart' as logger;
 import 'eval_on_dart_library.dart';
 import 'globals.dart';
+import 'service_registrations.dart' as registrations;
 import 'title.dart';
 import 'version.dart';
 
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/connected_app.dart
```diff
@@ -109,16 +121,33 @@ class ConnectedApp {
...
-        logger.log(
-          'Failed to fetch flutter version from '
-          '`ConnectedApp.initializeValues`: $e',
-        );
-      }
+
+    // No need to check the flutter version for Flutter web apps, as the
+    // performance tools that consume [flutterVersionNow] are not available for
+    // flutter web apps.
+    if (isFlutterNativeAppNow) {
```
jacob314, line 128: Can you file a bug tracking adding support for flutterVersion for Flutter web apps? At some point it would be useful if that worked for flutter web so devtools extension authors don't have to worry about this..

kenzieschmoll: oddly enough I just checked with the gallery flutter web app and the flutterVersion service does register as expected. I don't see anything here that would prevent this from being registered on the web https://github.com/flutter/flutter/blob/5c09eecd0d5ea428d163de12d12d6bd9a7a8b16c/packages/flutter_tools/lib/src/vmservice.dart#L221.

jacob314: @grouma is this only broken in g3?

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/connected_app.dart
grouma: Likely for now as Flutter Tools and Webdev have tighter restrictions on DevTools.

PR #3067, opened 2021-05-27
Fix a bug with app initialization

kenzieschmoll:


packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -93,10 +93,10 @@ class ServiceConnectionManager {
   VM vm;
   String sdkVersion;
 
-  bool get hasConnection =>
-      service != null &&
-      connectedApp != null &&
-      connectedApp.connectedAppInitialized;
+  bool get hasConnection => service != null && connectedApp != null;
```
kenzieschmoll, line 96: this is called from framework_core to make sure we do not try to initialize the vm service twice. We do not need to wait for all the connected app values to initialize for this check. At the point of the check FrameworkCore.initVmService, the connected app values (isFlutterApp, isDartWebApp, etc.) are in the process of being initialized, but we still have the app connection and want to prevent initVmService from occurring again

jacob314 (approved):
lgtm

PR #3070, opened 2021-06-02
Make update_version script null safe.

jacob314:


tool/update_version.dart
```diff
@@ -55,7 +50,7 @@ String incrementVersion(String oldVersion) {
   final parts = semVer.split('.');
 
   // Versions should have the form x.y.z
-  if (parts.length != 3) return null;
+  assert (parts.length == 3);
```
kenzieschmoll, line 53: remove extra space after assert

kenzieschmoll (approved):

tool/update_version.dart
jacob314: done

PR #3071, opened 2021-06-02
Update version and changelog.

jacob314:


kenzieschmoll (commented):
should we be using 2.3.1 since some packages have already had 2.3.0 published?

PR #3081, opened 2021-06-03
Update to package:vm_service 7.0.0

bkonyi:
Checks will fail until we can publish a new DDS with `vm_service: ^7.0.0` constraints, but we'll need a new `devtools_shared` before we can do that.


jacob314:
Closing as all these changes are part of in https://github.com/flutter/devtools/pull/3082

PR #3082, opened 2021-06-03
Update vm_service dependency.

jacob314:
To make rolling in the future easier, removed the dependency on package:test from package:devtools_app. This avoids problems in cases where there isn't a version of package:test that is compatible with all of the other packages required by devtools_app. All of the tests from devtools_app can be run with flutter_test anyway so there isn't much of a loss of functionality by switching to flutter_test.

### DIFF
--- a/packages/devtools_app/pubspec.yaml
+++ b/packages/devtools_app/pubspec.yaml
@@ -26,7 +26,7 @@ dependencies:
   async: ^2.0.0
   codicon: ^0.0.3
   collection: ^1.15.0-nnbd
-  dds: ^1.8.0
+  dds: ^2.0.1
   devtools_shared: 2.3.0
   file: ^6.0.0
   file_selector: ^0.8.0
@@ -52,9 +52,9 @@ dependencies:
   string_scanner: ^1.1.0
   url_launcher: ^5.0.0
   url_launcher_web: ^0.1.1+6
-  vm_service: ^6.2.0
+  vm_service: ^7.1.0
   vm_snapshot_analysis: ^0.6.0
-  web_socket_channel: ^2.0.0
+  web_socket_channel: ^2.1.0
 
 dev_dependencies:
   build_runner: ^2.0.4
@@ -64,7 +64,6 @@ dev_dependencies:
   flutter_test:
     sdk: flutter
   mockito: ^5.0.9
-  test: any # This version is pinned by Flutter so we don't need to set one explicitly.
   webkit_inspection_protocol: '>=0.5.0 <2.0.0'
 
 flutter:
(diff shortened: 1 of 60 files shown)


packages/devtools_app/pubspec.yaml
```diff
@@ -52,9 +52,9 @@ dependencies:
   string_scanner: ^1.1.0
   url_launcher: ^5.0.0
   url_launcher_web: ^0.1.1+6
-  vm_service: ^6.2.0
+  vm_service: ^7.1.0
```
kenzieschmoll, line 55: don't know if it matters, but the dds vm_service dep is ^7.0.0

jacob314: So ^7.0.0 is the same as '>=7.0.0 <8.0.0'

So the versions are compatible. As our dependency on vm_service is fragile as all changes are breaking due to implementing a class that isn't expected to be an interface, I like making our dependency a bit on the stricter side. `7.1.0` is the latest published version.

jacob314:
This change got much bigger than I'd like due to version match pain. I stripped the dep on `package:test` to avoid version mismatch pain. I don't see any reason why we should be sometimes using package:test instead of package:flutter_test. Given how tightly bound our versions are to packages that are also package:test and package:flutter_test deps I think this is a reasonable tradeoff.

packages/devtools_app/pubspec.yaml
```diff
@@ -130,3 +129,5 @@ dependency_overrides:
     path: ../devtools_shared #OVERRIDE_FOR_DEVELOPMENT
   devtools_testing: #OVERRIDE_FOR_DEVELOPMENT
     path: ../devtools_testing #OVERRIDE_FOR_DEVELOPMENT
+  devtools:  #OVERRIDE_FOR_DEVELOPMENT
+    path: ../devtools  #OVERRIDE_FOR_DEVELOPMENT
```
kenzieschmoll, line 133: why is this necessary if the dev dependency on devtools already depends on devtools from path? is devtools a transitive dep somewhere?

kenzieschmoll (commented):
looks like devtools_app, devtools_testing, and devtools need their pubspec versions to be bumped to 2.3.1 also, as well dependency versions for those packages

packages/devtools_app/pubspec.yaml
jacob314: I've now tweaked the other dependency to not be a path dependency for clarity. Using dependency overrides makes the version solving a little less frustrating.

jacob314:
> looks like devtools_app, devtools_testing, and devtools need their pubspec versions to be bumped to 2.3.1 also, as well dependency versions for those packages. Might as well prep for the release here too since some of the package versions are changing.
Actually need to bump to 2.3.2. I'll send that a separate CL as I already have that update and the changelog tweak ready to go in a separate CL.

kenzieschmoll (approved):

Issue #3043, opened 2021-05-19
Surface shader compilation time on the DevTools performance page


jacob314:
Now that some CLs for https://github.com/flutter/flutter/issues/78500 has landed, we should be able to surface ski shader events by adding a flag when a Flutter app app runs to enble tracing `skia.shaders`.
It would be good to surface the total time spent each frame spent on shader compilation so it is very clear just from the chart of frames which of the slow frames were due to shader compilation.
Currently, teams have a lot of difficulty telling whether they have gotten all the benefit from shader caching because they can't immediately tell whether their slow frames are due to shader caching or some other performance issue.

jacob314:
Fyi @kenzieschmoll

kenzieschmoll:
Do we have specific information on what the shader trace events will look like, and if not, who should we loop in to find out? Maybe @jason-simmons? 

Will shader time also be added to the Flutter.Frame events coming from the framework? @jonahwilliams

jonahwilliams:
I believe this would only show up in engine trace events and not the frame timings

kenzieschmoll:
Engine trace events meaning the events that come on the Raster thread (same as GPU::RasterizerDraw events and similar)? If we can pull the data from there, that should be sufficient once we have the frame ids included in the Flutter.Frame events (https://github.com/flutter/flutter/issues/80735).

kenzieschmoll:
@jason-simmons is going to add a flutter run flag `--trace-skia-allowlist` to support an easy option for turning these trace events on. 

@jonahwilliams @iskakaushik what would it take to get the shader time per frame piped into the Flutter.Frame events? I know this was just done for frame ids - would it also be possible to make shader time part of the frame timing api?

iskakaushik:
@kenzieschmoll this would require that we have the ability to pass additional args to skia trace events, @jason-simmons is that possible? We would need to add `frame_number` as an argument with value `FrameTimingsRecorder->GetFrameNumberTraceArg()` to the relevant timeline events.

@kenzieschmoll , If that is not possible, we would have to either:

1. Have skia return the stats about shader compilation in a consumable way to the Flush/Raster operations in rasterizer so we could capture them in the `FrameTimings` object to send it back to the framework and then devtools.
2. Devtools will have to use some heuristics to infer the frame number from the nearest `GpuRasterizer::Draw` trace event which would have the `frame_number` trace arg.

jonahwilliams:
[...]

PR #3083, opened 2021-06-03
Rank skia shader events first in the performance page timeline

kenzieschmoll:
Screen Shot 2021-06-03 at 10 38 04 AM

Work towards https://github.com/flutter/devtools/issues/3043. These events are traced and visible in the timeline by adding the `--trace-skia-allowlist=skia.shaders` flag and value to `flutter run`.

We need to https://github.com/flutter/flutter/issues/83835 to be fixed in order for us to do this in the most robust way (grouping all skia shader events together in a single group in the performance page timeline). For now, we prioritize any event groups with "shade" in the name.

This PR also fixes some issues with trace event processing that were exposed by the skia shader traces. We were not taking event categories into account when matching for async ids. This fixes that, along with some other small tracing / rendering bugs.

Note: the changes are duplicated between `performance/` and `performance/legacy/` files. We will have to do this for performance page changes until https://github.com/flutter/flutter/commit/78a96b09d64dc2a520e5b269d5cea1b9dde27d3f is in flutter stable.


packages/devtools_app/lib/src/performance/legacy/performance_model.dart
```diff
@@ -127,9 +127,17 @@ class LegacyPerformanceData {
   static int eventGroupComparator(String a, String b) {
     if (a == b) return 0;
 
-    // Order Unknown buckets last.
-    if (a == unknownKey) return 1;
-    if (b == unknownKey) return -1;
+    // TODO(kenz): Once https://github.com/flutter/flutter/issues/83835 is
+    // addressed, match on the group key that all skia.shader events will have.
+    // Order shader buckets first.
+    if (a.toLowerCase().contains('shade')) return -1;
```
jacob314, line 133: This isn't a valid comparator.
```dart
compare('shade A', 'shade B') == -1
compare('shade B', 'shade A') == -1
compare('shade A', 'shade A') == -1
```dart
This will cause sort to do funky things
What you want is
```dart
final shadeCompare = a.contains('shade').compareTo(b.contains('shade'));
if (shadeCompare != 0) return -shadeCompare; // because you want it first.
```
The same would also simplify the logic for the unknownKey logic.

packages/devtools_app/lib/src/performance/performance_model.dart
```diff
@@ -121,9 +121,17 @@ class PerformanceData {
   static int eventGroupComparator(String a, String b) {
     if (a == b) return 0;
 
-    // Order Unknown buckets last.
-    if (a == unknownKey) return 1;
-    if (b == unknownKey) return -1;
+    // TODO(kenz): Once https://github.com/flutter/flutter/issues/83835 is
+    // addressed, match on the group key that all skia.shader events will have.
+    // Order shader buckets first.
```
jacob314, line 126: duplicate code. also has same issue.

packages/devtools_app/lib/src/trace_event.dart
```diff
@@ -140,6 +137,18 @@ class TraceEventWrapper implements Comparable<TraceEventWrapper> {
   }
 }
 
+String generateAsyncUID({
+  @required String id,
+  @required String category,
+  String scope,
+}) {
+  if (scope == null) {
```
jacob314, line 145: nit:
```dart
[category, if (scope != null) scope, id].join(':')
```
is probably a bit slower but easier to read.

jacob314 (commented):
Lgtm once sort bug is fixed. Also add a unittest to verify that the sort is robust.

packages/devtools_app/lib/src/performance/performance_model.dart
kenzieschmoll: moved to a different file to de-dupe.

packages/devtools_app/lib/src/performance/legacy/performance_model.dart
kenzieschmoll: fixed and added more tests.

packages/devtools_app/lib/src/trace_event.dart
kenzieschmoll: done

PR #3085, opened 2021-06-04
Eval autocomplete polish.

jacob314:
Polish items:
Fix flicker while typing in autocomplete.
Fix missing autocompletes for names in scope from the current class.
Fix missing autocompletes for names in scope for the current library.
Fix missing autocompletes for static members of classes.
Add caching to avoid performing duplicated work looking up autocompletes for librries
Limitations: we show more top level more autocompletes than we should as we can't tell what the show clauses are for library imports and exports.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/debugger_controller.dart
+++ b/packages/devtools_app/lib/src/debugger/debugger_controller.dart
@@ -89,6 +89,9 @@ class DebuggerController extends DisposableController
 
   VmServiceWrapper get _service => serviceManager.service;
 
+  Map<ObjRef, Future<Set<String>>> autocompleteCache = {};
+  Map<ObjRef, Future<Set<String>>> autocompleteInclusiveCache = {};
+
   final ScriptCache _scriptCache = ScriptCache();
 
   final ScriptsHistory scriptsHistory = ScriptsHistory();
@@ -202,6 +205,10 @@ class DebuggerController extends DisposableController
 
   ValueListenable<StackFrameAndSourcePosition> get selectedStackFrame =>
       _selectedStackFrame;
+
+  Frame get frameForEval =>
+      _selectedStackFrame.value?.frame ??
+      _stackFramesWithLocation.value.first.frame;
 
   final _variables = ValueNotifier<List<Variable>>([]);
 
@@ -594,6 +601,7 @@ class DebuggerController extends DisposableController
     // ignore: unused_local_variable
     final status = reloadEvent.status;
 
+    _clearAutocompleteCaches();
     // Refresh the list of scripts.
     final scriptRefs = await _retrieveAndSortScripts(isolateRef);
     for (var scriptRef in scriptRefs) {
@@ -769,6 +777,12 @@ class DebuggerController extends DisposableController
     _breakPositionsMap.clear();
     _stdio.value = [];
     _uriToScriptMap.clear();
+    _clearAutocompleteCaches();
+  }
+
+  void _clearAutocompleteCaches() {
+    autocompleteCache.clear();
+    autocompleteInclusiveCache.clear();
   }
 
   /// Get the populated [Obj] object, given an [ObjRef].
--- a/packages/devtools_app/lib/src/debugger/debugger_controller.dart
+++ b/packages/devtools_app/lib/src/debugger/debugger_controller.dart
@@ -89,8 +89,21 @@ class DebuggerController extends DisposableController
 
   VmServiceWrapper get _service => serviceManager.service;
 
-  Map<ObjRef, Future<Set<String>>> autocompleteCache = {};
-  Map<ObjRef, Future<Set<String>>> autocompleteInclusiveCache = {};
+  /// Cache of autocomplete matches to show for a library when that library is
+  /// imported.
+  ///
+  /// This cache includes autocompletes from libraries exported by the library
+  /// but does not include autocompletes for libraries imported by this library.
+
+  Map<LibraryRef, Future<Set<String>>> libraryMemberAutocompleteCache = {};
+
+  /// Cache of autocomplete matches for a library for code written within that
+  /// library.
+  ///
+  /// This cache includes autocompletes from all libraries imported and exported
+  /// by the library as well as all private autocompletes for the library.
+  Map<LibraryRef, Future<Set<String>>>
+      libraryMemberAndImportsAutocompleteCache = {};
 
   final ScriptCache _scriptCache = ScriptCache();
 
@@ -190,6 +203,33 @@ class DebuggerController extends DisposableController
       highlighter: highlighter,
       executableLines: executableLines,
     );
+  }
+
+  /// Find the owner library for a ClassRef, FuncRef, or LibraryRef.
+  ///
+  /// If Dart had union types, ref would be type ClassRef | FuncRef | LibraryRef
+  Future<LibraryRef> findOwnerLibrary(Object ref) async {
+    if (ref is LibraryRef) {
+      return ref;
+    }
+    if (ref is ClassRef) {
+      final clazz = await classFor(ref);
+      return clazz?.library;
+    }
+    if (ref is FuncRef) {
+      return findOwnerLibrary(ref.owner);
+    }
+    return null;
+  }
+
+  /// Returns the class for the provided [ClassRef].
+  ///
+  /// May return null.
+  Future<Class> classFor(ClassRef classRef) async {
+    try {
+      return await getObject(classRef);
+    } catch (_) {}
+    return null;
   }
 
   // A cached map of uris to ScriptRefs.
@@ -781,8 +821,8 @@ class DebuggerController extends DisposableController
   }
 
   void _clearAutocompleteCaches() {
-    autocompleteCache.clear();
-    autocompleteInclusiveCache.clear();
+    libraryMemberAutocompleteCache.clear();
+    libraryMemberAndImportsAutocompleteCache.clear();
   }
 
   /// Get the populated [Obj] object, given an [ObjRef].
--- a/packages/devtools_app/lib/src/debugger/debugger_controller.dart
+++ b/packages/devtools_app/lib/src/debugger/debugger_controller.dart
@@ -213,6 +213,10 @@ class DebuggerController extends DisposableController
       return ref;
     }
     if (ref is ClassRef) {
+      if (ref.library != null) {
+        return ref.library;
+      }
+      // Fallback for older VMService versions.
       final clazz = await classFor(ref);
       return clazz?.library;
     }
--- a/packages/devtools_app/lib/src/debugger/evaluate.dart
+++ b/packages/devtools_app/lib/src/debugger/evaluate.dart
@@ -63,8 +63,9 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
         return;
       }
 
-      // No exact match, return the list of possible matches.
-      _autoCompleteController.clearSearchAutoComplete();
+      // We avoid clearing the list of possible matches here even though the
+      // current matches may probably out of data as that results in flicker
+      // as Flutter will render a frame before the new matches are available.
 
       // Find word in TextField to try and match (word breaks).
       final textFieldEditingValue = searchTextFieldController.value;
@@ -77,13 +78,22 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
       );
 
       // Only show pop-up if there's a real variable name or field.
-      if (parts.activeWord.isEmpty && !parts.isField) return;
-
+      if (parts.activeWord.isEmpty && !parts.isField) {
+        _autoCompleteController.clearSearchAutoComplete();
+        return;
+      }
       final matches = await autoCompleteResultsFor(parts, widget.controller);
-      _autoCompleteController.searchAutoComplete.value = matches.sublist(
-        0,
-        min(defaultTopMatchesLimit, matches.length),
-      );
+      if (matches.length == 1 && matches.first == parts.activeWord) {
+        // It is not useful to show a single autocomplete that is exactly what
+        // the already typed.
+        _autoCompleteController.clearSearchAutoComplete();
+      } else {
+        _autoCompleteController.searchAutoComplete.value = matches.sublist(
+          0,
+          min(defaultTopMatchesLimit, matches.length),
+        );
+        _autoCompleteController.currentDefaultIndex = 0;
+      }
     } else {
       _autoCompleteController.closeAutoCompleteOverlay();
     }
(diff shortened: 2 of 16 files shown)


packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -89,6 +89,9 @@ class DebuggerController extends DisposableController
 
   VmServiceWrapper get _service => serviceManager.service;
 
+  Map<ObjRef, Future<Set<String>>> autocompleteCache = {};
+  Map<ObjRef, Future<Set<String>>> autocompleteInclusiveCache = {};
```
kenzieschmoll, line 93: can you add dart doc to these to explain the difference?

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -63,8 +63,9 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
         return;
       }
 
-      // No exact match, return the list of possible matches.
-      _autoCompleteController.clearSearchAutoComplete();
+      // We avoid clearing the list of possible matches here even though the
+      // current matches may probably out of data as that results in flicker
```
kenzieschmoll, line 67: nit: "may probably be out of date"?

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -274,55 +284,232 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
   }
 }
 
+// If Dart had union types, ref would be type ClassRef | FuncRef | LibraryRef
+Future<LibraryRef> findOwnerLibrary(
+    Object ref, DebuggerController controller) async {
```
kenzieschmoll, line 289: nit: trailing comma

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -274,55 +284,232 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
+
 Future<List<String>> autoCompleteResultsFor(
   EditingParts parts,
   DebuggerController controller,
 ) async {
   final result = <String>{};
   if (!parts.isField) {
-    result.addAll(
-        controller.variables.value.map((variable) => variable.boundVar.name));
+    final variables = controller.variables.value;
```
kenzieschmoll, line 309: do we need to use List.from here if the intent is to make a copy?

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -274,55 +284,232 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
+    result.addAll(variables.map((variable) => variable.boundVar.name));
+
+    final thisVariable = variables.firstWhere(
+        (variable) => variable.boundVar.name == 'this',
+        orElse: () => null);
```
kenzieschmoll, line 314: nit: trailing comma

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -274,55 +284,232 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
+      if (function != null) {
+        final libraryRef = await findOwnerLibrary(function, controller);
+        if (libraryRef != null) {
+          controller.autocompleteInclusiveCache.clear();
+          controller.autocompleteCache.clear();
+          result.addAll(await libraryMemberAndImportsAutocompletes(
+              libraryRef, controller));
+        }
+      }
+      // TODO(jacobr): consider
```
kenzieschmoll, line 337: finish comment?

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -274,55 +284,232 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
+      }
+    }
+    final frame = controller.frameForEval;
+    if (frame != null) {
+      final function = frame.function;
+      if (function != null) {
+        final libraryRef = await findOwnerLibrary(function, controller);
+        if (libraryRef != null) {
+          controller.autocompleteInclusiveCache.clear();
+          controller.autocompleteCache.clear();
```
kenzieschmoll, line 332: can you make `_clearAutocompleteCaches` public instead and use that here?

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -274,55 +284,232 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
 }
 
-Future<Set<String>> _autoCompleteMembersFor(
-  ClassRef classRef,
+// Due to https://github.com/dart-lang/sdk/issues/46221
+// we cannot tell what the show clause for an export was so it is unsafe to
+// surface exports as if they were library members as there tend to be
+// significant false positives for libraries such as Flutter where all of
+// dart:ui shows up as in scope from flutter:foundation when it should not be.
+bool debugIncludeExports = true;
```
kenzieschmoll, line 373: based on the comment, should this be false then?

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -274,55 +284,232 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
+
+Future<Set<String>> libraryMemberAndImportsAutocompletes(
+    LibraryRef libraryRef, DebuggerController controller) {
```
kenzieschmoll, line 376: nit: tc

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -274,55 +284,232 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
+  return controller.autocompleteInclusiveCache.putIfAbsent(libraryRef,
+      () => _libraryMemberAndImportsAutocompletes(libraryRef, controller));
+}
+
+Future<Set<String>> _libraryMemberAndImportsAutocompletes(
+    LibraryRef libraryRef, DebuggerController controller) async {
```
kenzieschmoll, line 382: tc

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -274,55 +284,232 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
+  return result;
+}
+
+Future<Set<String>> libraryMemberAutocompletes(
+  DebuggerController controller,
+  LibraryRef libraryRef, {
+  @required bool includePrivates,
+}) async {
+  var result = await controller.autocompleteCache.putIfAbsent(
+      libraryRef, () => _libraryMemberAutocompletes(controller, libraryRef));
```
kenzieschmoll, line 419: nit: tc

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -274,55 +284,232 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
+      }
+    }
+    (await Future.wait(futures)).forEach(result.addAll);
+  } catch (_) {
+    // Silently skip library completions if there is a failure.
+  }
+  return result;
+}
+
+Future<Set<String>> libraryMemberAutocompletes(
```
grouma, line 413: Why does this need to be public?

packages/devtools_app/lib/src/debugger/debugger_controller.dart
jacob314: done. tried tweaking the names as well to make them a bit clearer.

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -274,55 +284,232 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
+    final variables = controller.variables.value;
+    result.addAll(variables.map((variable) => variable.boundVar.name));
+
+    final thisVariable = variables.firstWhere(
+        (variable) => variable.boundVar.name == 'this',
+        orElse: () => null);
+    if (thisVariable != null) {
+      final thisValue = thisVariable.boundVar.value;
+      if (thisValue is InstanceRef) {
+        await _addAllInstanceMembersToAutocompleteList(
```
grouma, line 318: Why must we now check if we have a `this`?

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: cleaned up comment.

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -274,55 +284,232 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
-Future<Set<String>> _autoCompleteMembersFor(
-  ClassRef classRef,
+// Due to https://github.com/dart-lang/sdk/issues/46221
+// we cannot tell what the show clause for an export was so it is unsafe to
+// surface exports as if they were library members as there tend to be
+// significant false positives for libraries such as Flutter where all of
+// dart:ui shows up as in scope from flutter:foundation when it should not be.
+bool debugIncludeExports = true;
+
+Future<Set<String>> libraryMemberAndImportsAutocompletes(
```
grouma, line 375: Not clear to me why this needs to be public.

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: done

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -366,13 +553,13 @@ bool _isOperator(FuncRef funcRef) => [
...
-  final frame = controller.selectedStackFrame.value?.frame ??
-      controller.stackFramesWithLocation.value.first.frame;
+  final frame = controller.frameForEval;
   final currentScript = frame.location.script;
-  return !member.startsWith('_') ||
-      currentScript.id == clazz?.location?.script?.id;
+  return !isPrivate(member) || currentScript.id == clazz?.location?.script?.id;
 }
+
+bool isPrivate(String member) => member.startsWith('_');
```
grouma, line 565: This doesn't need to be public.

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: We're not making a copy. We are just storing as a local variable two places within this method.

packages/devtools_app/lib/src/debugger/debugger_controller.dart
grouma: +1 - it's not immediately obvious to me

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: done

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -274,55 +284,232 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
   }
 }
 
+// If Dart had union types, ref would be type ClassRef | FuncRef | LibraryRef
+Future<LibraryRef> findOwnerLibrary(
```
grouma, line 288: This can be private

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: Obsolete. Removed.

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -274,55 +284,232 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
+          thisValue,
+          controller,
+        );
+      }
+    }
+    final frame = controller.frameForEval;
+    if (frame != null) {
+      final function = frame.function;
+      if (function != null) {
+        final libraryRef = await findOwnerLibrary(function, controller);
```
grouma, line 329: Why must we go through the function? I thought a frame has a corresponding library reference.

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: removed this code. I clearly should have waited till morning to send out for review. This was just debugging code so I would get new caches every time.

grouma:
This PR really needs a description of what it's trying to achieve. There's a lot to this polish.

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: done.

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: with this level of extreme use of trailing commas, we could probably write a lint to enforce the style and quick fix.

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: done

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: there is a matching helper method with the same name that is private.  Doesn't have to be public but I don't see a lot of harm in making it public.

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: documented and added a case for static members of the this class as well.

packages/devtools_app/lib/src/debugger/evaluate.dart
grouma: It's just good practice to limit the public interface when possible, even for internal classes, to reduce the difficulty of refactors.

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: Made this a public member on the controller. Going from the owner to the library is a generally useful thing to do so it shouldn't be private.

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: I agree this is annoying but I fail to see where the library is on the frame. You can go to the source file directly but that is not the same thing as the library.
@bkonyi  am I missing something?
https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#frame

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: > This PR really needs a description of what it's trying to achieve. There's a lot to this polish.

done.

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: actually it needs to be a public utility uses everywhere... we reinvented the wheel a lot...
lib/src/debugger/evaluate.dart:bool _isPrivate(String member) => member.startsWith('_');
lib/src/inspector/inspector_service.dart:      if (name.startsWith('_')) {
lib/src/memory/memory_graph_model.dart:                  className.startsWith('_')) ||
lib/src/memory/memory_graph_model.dart:                actual.name.startsWith('_')) ||
lib/src/provider/instance_viewer/instance_details.dart:  bool get isPrivate => name.startsWith('_');
lib/src/ui/icons.dart:    final bool isPrivate = name.startsWith('_');
lib/src/utils.dart:  final isAPrivate = a.startsWith('_');
lib/src/utils.dart:  final isBPrivate = b.startsWith('_')

jacob314:
Replied to all comments. This also really needs some test coverage. Let me try to add a test that connects to a real sample app and fetches autocompletes.

packages/devtools_app/lib/src/debugger/evaluate.dart
bkonyi: As of `3.46` you can get access to the `@Library` of a `@Function` by doing something like `function.owner as @Library` (for top level functions) or `(function.owner as @Class).library`. That functionality just landed this week though, so most applications in the wild will require requesting the full objects to eventually get to the `@Library` in either the `Class` or `Function` object.

jacob314: Thanks! I'm now taking advantage of that where possible.

jacob314:
PTAL. Changed the tests to use the real VM instead of mocks and added tests for more cases.

packages/devtools_testing/fixtures/flutter_app/lib/src/autocomplete.dart
```diff
@@ -0,0 +1,66 @@
+import 'dart:developer' as developer;
```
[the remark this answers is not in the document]
jacob314: done

PR #3086, opened 2021-06-04
Support multi-line eval output

grouma:
The `toString()` of an eval result can be multi-line. For example `'\n ${[0,0]} \n'`. Prior to this change this would not render correctly. I also believe, but I'm not certain, that this was the root cause of the flaky overlapping console lines.

### DIFF
--- a/packages/devtools_app/lib/src/tree.dart
+++ b/packages/devtools_app/lib/src/tree.dart
@@ -14,7 +14,7 @@ class TreeView<T extends TreeNode<T>> extends StatefulWidget {
     this.dataDisplayProvider,
     this.onItemPressed,
     this.shrinkWrap = false,
-    this.itemExtent = defaultListItemHeight,
+    this.itemExtent,
     this.onTraverse,
   });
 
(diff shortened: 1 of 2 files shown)


kenzieschmoll (approved):

jacob314 (approved):
We should audit that any other uses of Tree still look good with this change.
lgtm

packages/devtools_app/lib/src/tree.dart
```diff
@@ -68,7 +68,6 @@ class _TreeViewState<T extends TreeNode<T>> extends State<TreeView<T>>
     if (items.isEmpty) return const SizedBox();
     return ListView.builder(
       itemCount: items.length,
-      itemExtent: widget.itemExtent,
```
kenzieschmoll, line 71: instead of removing this, can we remove the default value in the constructor? this would make sure that other consumers of tree view can still pass in item extent if they need

grouma: Ah yeah. That's much better. Done.

Issue #3079, opened 2021-06-03
add crash reporting to our analytics


devoncarew:
We should add crash reporting - exception reporting - to our analytics.

devoncarew:
@kenzieschmoll @terrylucas - do you happen to know a good place to hook into uncaught / unexpected exceptions in DevTools? Are we handling them now in some way (logging them to the browser console / other?

I found some info here wrt error handling and Flutter apps: https://stackoverflow.com/questions/57879455/flutter-catching-all-unhandled-exceptions, but am not sure what we're currently doing.

jacob314:
I don't believe we do anything currently. The one thing we should do differently than that stack overflow example is only override the default onError in release builds or make sure we continue to call the error handler we overrode.
https://stackoverflow.com/questions/57879455/flutter-catching-all-unhandled-exceptions

PR #3100, opened 2021-06-08
catch any unhandled exceptions in devtools and report via our analytics channel

devoncarew:
- catch any unhandled exceptions in devtools and report via our analytics channel
- fix https://github.com/flutter/devtools/issues/3079

### DIFF
--- a/packages/devtools_app/lib/src/app_error_handling.dart
+++ b/packages/devtools_app/lib/src/app_error_handling.dart
@@ -41,7 +41,7 @@ void setupErrorHandling(Future Function() appStartCallback) {
 }
 
 void _reportError(Object error, StackTrace stack) {
-  final String terseStackTrace = stack_trace.Trace.from(stack).terse.toString();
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/app_error_handling.dart
```diff
@@ -0,0 +1,49 @@
...
+    };
+
+    return appStartCallback();
+  }, (Object error, StackTrace stack) {
+    _reportError(error, stack);
+  });
+}
+
+void _reportError(Object error, StackTrace stack) {
+  final String terseStackTrace = stack_trace.Trace.from(stack).terse.toString();
```
kenzieschmoll, line 44: drop type on LHS

kenzieschmoll (approved):

packages/devtools_app/lib/src/app_error_handling.dart
devoncarew: done

jacob314:
Fyi @terrylucas who was also thinking of adding this back to DevTools.

Issue #2012, opened 2020-05-29
Flame chart - jumpTo called before scroll position offsets have new size


kenzieschmoll:
In flame_chart.dart, we call
`linkedHorizontalScrollControllerGroup.jumpTo(.clamp(FlameChart.minScrollOffset, maxScrollOffset));` to update the scroll offset as we zoom. At times when we call this, `scrollController.position.maxScrollOffset` is not updated for the new zoom level and layout size, causing BallisticScrollActivities to bounce the scroll offset to what it thinks is the `maxScrollOffset`.

We need to have a way to call `jumpTo` when we can guarantee that the min / max scroll offsets are updated for the new zoom value. 

I attempted to set these offsets manually using `scrollController.position.applyContentDimensions(minScrollOffset, maxScrollOffset)` but this results in an error regarding gesture detectors: `The replaceGestureRecognizers() method can only be called during the layout phase.`

kenzieschmoll:
@dnfield @HansMuller for any ideas on how to workaround this.

kenzieschmoll:
Some more speculation on this from https://github.com/flutter/devtools/pull/2808#discussion_r595362000

"Perhaps we need a custom scroll controller class that lets us update its extent immediately based on the new size to avoid all these annoying edge cases?"

kenzieschmoll:
Some updated code links here:
The call to update the horizontal scroll position where the `position.maxScrollExtent` may not have been updated for the new size of the scroll controller (size changes as the flame chart zooms in and out via an animation controller): https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/charts/flame_chart.dart#L460

Here we are forced to use a post frame callback to ensure that the maxScrollExtent has been updated for the new zoom level after we zoom: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/charts/flame_chart.dart#L543-L551

cc @Piinks

Piinks:
Thanks for the update! I think the primary issue here is that the Scrollable is being manipulated and accessed at the same time. I am not sure how we can fix that, cc also @goderbauer who is looking into official support for 2D scrolling in Flutter.

kenzieschmoll:
@goderbauer has there been any progress on a lazy 2D view? The scrolling issues we are experiencing in DevTools flame chart are proving very difficult to avoid.

Piinks:
[...]

PR #3107, opened 2021-06-10
Improve scrolling / zooming behavior in flame charts

kenzieschmoll:
Some of these changes will reduce the number of rebuilds, and one (increasing size of right padding for the last node in each row) is a band-aid fix for a larger and more difficult to solve issue (https://github.com/flutter/devtools/issues/2012).

### DIFF
--- a/packages/devtools_app/lib/src/charts/flame_chart.dart
+++ b/packages/devtools_app/lib/src/charts/flame_chart.dart
@@ -436,7 +436,6 @@ abstract class FlameChartState<T extends FlameChart,
     final newZoom = zoomController.value;
     if (previousZoom == newZoom) return;
 
-
     // Store current scroll values for re-calculating scroll location on zoom.
     final lastScrollOffset = horizontalControllerGroup.offset;
 
--- a/packages/devtools_app/lib/src/charts/flame_chart.dart
+++ b/packages/devtools_app/lib/src/charts/flame_chart.dart
@@ -433,8 +433,8 @@ abstract class FlameChartState<T extends FlameChart,
 
   void _handleZoomControllerValueUpdate() {
     final previousZoom = currentZoom;
-    currentZoom = zoomController.value;
-    if (previousZoom == currentZoom) return;
+    final newZoom = zoomController.value;
+    if (previousZoom == newZoom) return;
 
 
     // Store current scroll values for re-calculating scroll location on zoom.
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -434,27 +432,28 @@ abstract class FlameChartState<T extends FlameChart,
...
+    // Calculate the new horizontal scroll position.
+    final newScrollOffset = fixedX >= 0
+        ? fixedX * currentZoom / previousZoom +
+            widget.startInset -
+            safeMouseHoverX
+        // We are in the fixed portion of the window - no need to transform.
+        : lastScrollOffset;
 
+    setState(() {
+      currentZoom = zoomController.value;
```
jacob314, line 456: we already change currentZoom to this value outside of setState. I would avoid updating currentZoom until this line. Maybe add a 
nextZoom local variable to disambiguate.

packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -530,27 +529,24 @@ abstract class FlameChartState<T extends FlameChart,
...
+    // Call these in a post frame callback so that the scroll controllers have
+    // had time to update their scroll extents. Otherwise, we can hit a race
+    // where are trying to scroll to an offset that is beyond what the scroll
+    // controller thinks its max scroll extent is.
+    WidgetsBinding.instance.addPostFrameCallback((_) async {
+      await Future.wait([
+        scrollHorizontallyToData(data),
+        if (scrollVertically) scrollVerticallyToData(data),
+      ]);
+    });
```
jacob314, line 549: why does zooToTimeRange impact the vertical scroll controller? I would think that would only impact horizontal.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/charts/flame_chart.dart
kenzieschmoll: It doesn't impact the vertical scroll controller, but when we call scrollVerticallyToData(data), that triggers a rebuild that may occur before all the horizontal scroll controllers have updated.

packages/devtools_app/lib/src/charts/flame_chart.dart
kenzieschmoll: done

PR #3109, opened 2021-06-10
Add a loading message when loading app size data from file paths

kenzieschmoll:
This message is provided as an extension point so that we can customize the message in google3. Internally, there is high latency due to the file store and we should warn users that it could a little while.


jacob314 (approved):
lgtm

PR #3114, opened 2021-06-11
Fix bug where you could trigger simultaneous connects by accidentally clicking or pressing return twice

jacob314:
The previous approach of delaying by 250ms was insufficient. This matches other UI in DevTools where we block actions while the current action is in progress.

### DIFF
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -202,7 +202,10 @@ class DevToolsAppState extends State<DevToolsApp> {
             if (tabs.isEmpty) {
               return DevToolsScaffold.withChild(
                 child: CenteredMessage(
-                    'The "$page" screen is not available for this application.'),
+                  page != null
+                      ? 'The "$page" screen is not available for this application.'
+                      : 'No tabs available for this application.',
+                ),
                 ideTheme: ideTheme,
                 analyticsProvider: widget.analyticsProvider,
               );
--- a/packages/devtools_app/lib/src/blocking_action_mixin.dart
+++ b/packages/devtools_app/lib/src/blocking_action_mixin.dart
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
+
 import 'package:flutter/widgets.dart';
 
 /// Provides functionality to track actionInProgress state while executing long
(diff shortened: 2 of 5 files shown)


packages/devtools_app/lib/src/app.dart
```diff
@@ -202,7 +202,10 @@ class DevToolsAppState extends State<DevToolsApp> {
             if (tabs.isEmpty) {
               return DevToolsScaffold.withChild(
                 child: CenteredMessage(
-                    'The "$page" screen is not available for this application.'),
+                  page != null
+                      ? 'The "$page" screen is not available for this application.'
+                      : 'No tabs available for this application.',
```
jacob314, line 207: This should now never occur but the previous message was confusing.

packages/devtools_app/lib/src/initializer.dart
```diff
@@ -117,7 +117,7 @@ class _InitializerState extends State<Initializer>
 
   /// Shows a "disconnected" overlay if the [service.serviceManager] is not currently connected.
   void _handleNoConnection() {
-    WidgetsBinding.instance.addPostFrameCallback((_) {
+    WidgetsBinding.instance.scheduleFrameCallback((_) {
```
jacob314, line 120: I don't understand why we were using a postFrameCallback. A frame might not render and then it wouldn't be called till some strange later time.

packages/devtools_app/lib/src/blocking_action_mixin.dart
```diff
@@ -19,17 +21,37 @@ mixin BlockingActionMixin<T extends StatefulWidget> on State<T> {
...
+  Future<void> blockWhileInProgress(Future callback()) async {
     setState(() {
       _actionInProgress = true;
     });
     try {
-      // TODO(jacobr): handle actions that timeout gracefully.
-      await callback();
+      // If we await a callback that does not complete until after this object
+      // is disposed, we will leak State objects.
+      await Future.any([callback(), _disposed.future]);
```
kenzieschmoll, line 38: should we block the `callback()` from being called if `_disposed.future` completes first? `Future.any` doesn't stop the other futures from completing.

jacob314: There isn't a way to cancel the in progress `callback()` if _disposed completes first .

kenzieschmoll: You can't cancel the callback but you can prevent it from running with a boolean flag.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1305,3 +1281,18 @@ extension BoolExtension on bool {
     return -1;
   }
 }
+
+Future<T> whenValueNonNull<T>(ValueListenable<T> listenable) {
```
kenzieschmoll, line 1285: I don't see where this is used in this PR?

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1305,3 +1281,18 @@ extension BoolExtension on bool {
...
+  if (listenable.value != null) return Future.value(listenable.value);
+  final completer = Completer<T>();
+  void listener() {
+    final value = listenable.value;
+    if (value != null) {
+      completer.complete(value);
+    }
+    listenable.removeListener(listener);
```
kenzieschmoll, line 1293: what if `value` is null? Then we would remove the listener and the completer would never complete. If `value` is null, should we keep listening?

packages/devtools_app/lib/src/utils.dart
jacob314: it is used in a future pr.

packages/devtools_app/lib/src/utils.dart
jacob314: agreed. Fixed the logic.

kenzieschmoll (approved):

PR #3117, opened 2021-06-11
Start of refactoring hovercards.

jacob314:
I'd push the code that uses this refactor with the same CL but it is too entangled with another change.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/hover.dart
+++ b/packages/devtools_app/lib/src/debugger/hover.dart
@@ -2,10 +2,14 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
+import 'dart:math' as math;
+
 import 'package:flutter/cupertino.dart';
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/debugger/hover.dart
```diff
@@ -158,3 +174,84 @@ class HoverCard {
     _isRemoved = true;
   }
 }
+
+class HoverCardTip extends StatefulWidget {
```
kenzieschmoll, line 178: I don't understand the naming here. "tip" makes me think multiple things initially. Is it a hint? is it a tooltip? Is it describing the point at the bottom of the hover card?

kenzieschmoll (commented):
LGTM with a comment about the class name

kenzieschmoll (approved):

packages/devtools_app/lib/src/debugger/hover.dart
jacob314: Changed name to HoverCardTooltip and added a comment.

PR #3118, opened 2021-06-11
Add a clear method to history_manager

jacob314:
Moving this off to reduce the # of non-noop things in a big refactor CL.

### DIFF
--- a/packages/devtools_app/lib/src/history_manager.dart
+++ b/packages/devtools_app/lib/src/history_manager.dart
@@ -17,7 +17,6 @@ class HistoryManager<T> {
 
   void clear() {
     _history.clear();
-    _historyIndex = null;
     _historyIndex = -1;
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/history_manager.dart
```diff
@@ -15,6 +15,13 @@ class HistoryManager<T> {
   final _history = <T>[];
   int _historyIndex = -1;
 
+  void clear() {
+    _history.clear();
+    _historyIndex = null;
+    _historyIndex = -1;
```
kenzieschmoll, line 21: Why are we setting this twice?

jacob314: wow that code was dumb. fixed.

kenzieschmoll (approved):

Issue #2631, opened 2021-01-27
Cache cpu profile information when refreshing data in the Performance view


fzyzcjy:
Hi thanks for the lib! I find that, once in a while, the Performance page says "No CPU profile data" while sometimes it behaves normally. Even stranger, once, I successfully see the profile data, but when I click somewhere in the same frame, it re-analyzes and then tells me *no* cpu profile data...

This is quite annoying, because I need the flamechart to see where the performance is bad.

Thanks!

e.g.

![image](https://user-images.githubusercontent.com/5236035/105933007-b5fce280-6088-11eb-8826-71576a250be6.png)

kenzieschmoll:
The dart vm collects CPU profile samples at a specific frequency (specified by profile granularity selector at top of performance page), and stores those samples on a buffer. So it could either be the case that A) the vm did not collect any cpu samples during the time period you have selected in the timeline, or B) activity is continuing to occur in your application while you are looking at this timeline, and in between the time you refreshed the performance page and clicked on an event to view the cpu profile for, the samples have fallen out of the buffer. 

Based on the frame times shown in this screenshot, B) is more likely. Based on my description, does it sound like B) could be what's happening in your app? If this were the case, you'd be more likely to be able to access CPU profile data for later frames in the flutter frames chart (far right) than you would for earlier frames.

We could do something to mitigate this. We could pull all the CPU profiles available at "Refresh" and then generate "sub-profiles" for each time the profiler needs new data. This would allow us to hang on to all the available CPU profile data even if the samples fall out of the VM's buffer.

fzyzcjy:
Hi thanks for the reply! I originally thought that, the `Refresh` will stop once the buffer is full (actually, it does not collect forever, but only collect a bunch of frames and then no more). So how can I manually pause the profiling such that I can look into a frame without its samples lost?

kenzieschmoll:
Unfortunately, I don't think there is a way to do this. We can help on our end by caching the data available upon refreshing. 

Clicking Refresh pulls all available timeline data, so it is not intended to be a live view. If something happens in your app that you want to take a look at, clicking Refresh again would pull all recent data. We are considering changing the performance page to more of a "live" view of performance, that constantly pulls new data in instead of requiring users to refresh.

fzyzcjy:
By the way, I also find the `CPU Profiler` tab quite helpful. That profiler faithfully record the cpu instructions :)

[...]

PR #3121, opened 2021-06-11
Cache CPU profiles for selected frames

kenzieschmoll:
This will prevent users from losing CPU profile data for frames they have already selected. We have seen users in user studies hit problems with CPU samples falling out of the CPU sample buffer. 

We now cache profiles with their respective frames in the performance page. For timeline events within that frame, we will generate a sub profile from the frame's cached profile upon selection.

This PR also improves the performance of the CPU profiler by caching the `callTreeRoots` and `bottomUpRoots` for an instance of `CpuProfileData`.

This will improve https://github.com/flutter/devtools/issues/2631

### DIFF
--- a/packages/devtools_app/lib/src/performance/performance_controller.dart
+++ b/packages/devtools_app/lib/src/performance/performance_controller.dart
@@ -61,8 +61,6 @@ class PerformanceController
   ValueListenable<List<FlutterFrame>> get flutterFrames => _flutterFrames;
   final _flutterFrames = ValueNotifier<List<FlutterFrame>>([]);
 
-  final _flutterFramesById = <String, FlutterFrame>{};
-
   /// Whether an empty timeline recording was just recorded.
   ValueListenable<bool> get emptyTimeline => _emptyTimeline;
   final _emptyTimeline = ValueNotifier<bool>(false);
@@ -165,25 +163,22 @@ class PerformanceController
 
   Future<void> selectTimelineEvent(
     TimelineEvent event, {
-    bool pullCpuProfile = true,
+    bool updateProfiler = true,
   }) async {
     if (event == null || data.selectedEvent == event) return;
 
     data.selectedEvent = event;
     _selectedTimelineEventNotifier.value = event;
 
-    if (pullCpuProfile && event.isUiEvent) {
-      if (event.frameId != null) {
-        final frameProfile = _flutterFramesById[event.frameId]?.cpuProfileData;
-        if (frameProfile != null) {
-          cpuProfilerController.reset();
-          await cpuProfilerController.generateSubProfile(
-            frameProfile,
-            event.time,
-            processId: 'subProfile for ${event.name} from ${event.frameId}',
-          );
-          data.cpuProfileData = cpuProfilerController.dataNotifier.value;
-        }
+    if (event.isUiEvent && updateProfiler) {
+      final storedProfile =
+          cpuProfilerController.cpuProfileStore.storedProfile(event.time);
+      if (storedProfile != null) {
+        await cpuProfilerController.processAndSetData(
+          storedProfile,
+          processId: 'Stored profile for ${event.time}',
+        );
+        data.cpuProfileData = cpuProfilerController.dataNotifier.value;
       } else if ((!offlineMode || offlinePerformanceData == null) &&
           cpuProfilerController.profilerEnabled) {
         // Fetch a profile if not in offline mode and if the profiler is enabled
@@ -192,6 +187,10 @@ class PerformanceController
           startMicros: event.time.start.inMicroseconds,
           extentMicros: event.time.duration.inMicroseconds,
           processId: '${event.traceEvents.first.id}',
+        );
+        cpuProfilerController.cpuProfileStore.storeProfile(
+          event.time,
+          cpuProfilerController.dataNotifier.value,
         );
         data.cpuProfileData = cpuProfilerController.dataNotifier.value;
       }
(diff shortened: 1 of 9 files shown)


packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -61,6 +61,8 @@ class PerformanceController
   ValueListenable<List<FlutterFrame>> get flutterFrames => _flutterFrames;
   final _flutterFrames = ValueNotifier<List<FlutterFrame>>([]);
 
+  final _flutterFramesById = <String, FlutterFrame>{};
```
kenzieschmoll, line 64: The changes in this file are duplicated from performance/legacy/performance_controller.dart. These two files will diverge soon once the files under performance/ are migrated to use the FrameTiming API for flutter frames, though changes like this will continue to need to be applied to both files until the changes we need to use the FrameTiming API hit flutter stable.

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -206,9 +216,28 @@ class PerformanceController
     data.selectedFrame = frame;
     _selectedFrameNotifier.value = frame;
 
-    await selectTimelineEvent(frame.uiEventFlow);
+    // We do not need to pull the CPU profile because we will pull the profile
+    // for the entire frame. The order of selecting the timeline event and
+    // pulling the CPU profile for the frame (directly below) matters here.
+    // If the selected timeline event is null, the event details section will
+    // not show the progress bar while we are processing the CPU profile.
+    await selectTimelineEvent(frame.uiEventFlow, pullCpuProfile: false);
```
jacob314, line 224: What if we added a CPU Profile model/databse that tracked all the time ranges of profiling data we have already cached. That way we wouldn't need to track whether to set pullCpuProfile to true or not. If the time range is already cached. That would also enable supporting offline mode for the cpu profiler data in the future as the offline data could include the database of cpu profile samples that were cached.

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
```diff
@@ -407,3 +423,47 @@ int stackFrameIdCompare(String a, String b) {
     throw error;
   }
 }
+
+class CpuProfileStore {
+  final _store = <TimeRange, CpuProfileData>{};
+
+  CpuProfileData storedProfile(TimeRange time) {
```
jacob314, line 430: nit: I would remove the worst store from most of these names. It just makes it a little confusing.

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
```diff
@@ -407,3 +423,47 @@ int stackFrameIdCompare(String a, String b) {
     throw error;
   }
 }
+
+class CpuProfileStore {
+  final _store = <TimeRange, CpuProfileData>{};
```
jacob314, line 428: nit:
_store => _profiles

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
```diff
@@ -407,3 +423,47 @@ int stackFrameIdCompare(String a, String b) {
...
+
+  CpuProfileData storedProfile(TimeRange time) {
+    // If we have a profile for a time range encompassing [time], then we can
+    // generate and cache the profile for [time] without needing to pull data
+    // from the vm service.
+    _maybeGenerateSubProfile(time);
+    return _store[time];
+  }
+
+  void storeProfile(TimeRange time, CpuProfileData profile) {
```
jacob314, line 438: nit: maybe call this method just `add` for consistency with the Dart collection types. `addProfile` or `saveProfile` would also be fine.

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
```diff
@@ -407,3 +423,47 @@ int stackFrameIdCompare(String a, String b) {
...
+    _store[time] = profile;
+  }
+
+  void _maybeGenerateSubProfile(TimeRange time) {
```
jacob314, line 442: nit: this is generating a bunch of overlapping profiles in the cache. That would make it a bit harder to later serialize this.   Is computing the sub profiles expensive enough that it worth it to add completely overlapping ranges to the cache. It makes sense we will need some overlap as  frames overlap but having completely overlapping entries seems a little odd.

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
jacob314, line 430: nit: maybe call this lookupProfile and note that it will be null if the profile isn't available? Took me a second reading it to realize what storedProfile meant.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
kenzieschmoll: renamed to `profileForTimeRange` and `addProfile`

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
kenzieschmoll: done

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
kenzieschmoll: done

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
kenzieschmoll: done

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
kenzieschmoll: when a sub profile is generated, it has to be created (not expensive) and then processed - this is the expensive part. So if we don't cache each subprofile, there will be processing cost for each time we have to regenerate the sub profile from a super profile.

PR #3125, opened 2021-06-14
Surface shader time per frame in the Performance page

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/3043. This PR:

- Adds logic to calculate the amount of time spent on Shader compilation for a single frame.
- Adds a warning icon on flutter frames that have Shader compilation time
- Adds info about shader compilation time to the tooltip for a flutter frame
- Colors shader events in the timeline with the Raster Jank Color we use in the flutter frames chart

Screen Shot 2021-06-14 at 4 15 26 PM
Screen Shot 2021-06-14 at 4 12 03 PM

**[EDIT with some UI changes from review comments]**
@InMatrix for some UX input on colors here

shader-compilation
Screen Shot 2021-06-15 at 12 10 01 PM
Screen Shot 2021-06-15 at 12 10 20 PM

### DIFF
--- a/packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
+++ b/packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
@@ -257,6 +257,10 @@ class FlutterFramesChartItem extends StatelessWidget {
   static const selectedFrameIndicatorKey =
       Key('flutter frames chart - selected frame indicator');
 
+  // TODO(kenz): instead of hard coding 4, factor in the display refresh rate to
+  // calculate 1/4 of the target frame time.
+  static const shaderTimeThreshold = Duration(milliseconds: 4);
+
   final FlutterFrame frame;
 
   final bool selected;
@@ -273,6 +277,10 @@ class FlutterFramesChartItem extends StatelessWidget {
 
     final bool uiJanky = frame.isUiJanky(displayRefreshRate);
     final bool rasterJanky = frame.isRasterJanky(displayRefreshRate);
+    final bool hasShaderJank = frame.hasShaderTime &&
+        rasterJanky &&
+        frame.shaderDuration > shaderTimeThreshold;
+
     // TODO(kenz): add some indicator when a frame is so janky that it exceeds the
     // available axis space.
     final ui = Container(
@@ -293,7 +301,7 @@ class FlutterFramesChartItem extends StatelessWidget {
       children: [
         // TODO(kenz): make tooltip to persist if the frame is selected.
         DevToolsTooltip(
-          tooltip: _tooltipText(frame),
+          tooltip: _tooltipText(frame, hasShaderJank),
           padding: const EdgeInsets.all(denseSpacing),
           child: Container(
             padding: const EdgeInsets.symmetric(horizontal: densePadding),
@@ -319,7 +327,7 @@ class FlutterFramesChartItem extends StatelessWidget {
             color: defaultSelectionColor,
             height: selectedIndicatorHeight,
           ),
-        if (frame.hasShaderTime)
+        if (hasShaderJank)
           Row(
             mainAxisAlignment: MainAxisAlignment.center,
             children: const [
@@ -330,11 +338,11 @@ class FlutterFramesChartItem extends StatelessWidget {
     );
   }
 
-  String _tooltipText(FlutterFrame frame) {
+  String _tooltipText(FlutterFrame frame, bool hasShaderJank) {
     return [
       'UI: ${msText(frame.uiEventFlow.time.duration)}',
       'Raster: ${msText(frame.rasterEventFlow.time.duration)}',
-      if (frame.hasShaderTime) 'Skia.Shaders: ${msText(frame.shaderDuration)}',
+      if (hasShaderJank) ...['Skia.Shaders: ${msText(frame.shaderDuration)}'],
     ].join('\n');
   }
 }
--- a/packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
+++ b/packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
 import 'dart:math' as math;
 
 import 'package:flutter/material.dart';
@@ -201,6 +202,8 @@ class _FlutterFramesChartState extends State<FlutterFramesChart>
         _legendItem('Frame Time (Raster)', mainRasterColor),
         const SizedBox(height: denseRowSpacing),
         _legendItem('Jank (slow frame)', uiJankColor),
+        const SizedBox(height: denseRowSpacing),
+        _legendItem('Shader Compilation', shaderCompilationColor),
       ],
     );
   }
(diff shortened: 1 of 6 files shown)


kenzieschmoll:
@jacob314 currently we are showing the warning icon for any frame that has shader time, should we only show the warning icon if the GPU portion of the frame exceeds 16.6 ms?

jacob314:
This is really exciting.
We should update https://flutter.dev/docs/perf/rendering/shader to use a screenshot from DevTools. It is MUCH clearer than the one from observatory that is there currently.

High level: I'd love to make it even more obvious you have shader jank when there is significant shader jank that is causing missed frame.
This is because when have shader compilation jank, there are specific steps you can take to fix the problem that are very different from steps you would take for other performance problems.
Ways to make it more obvious: 
1. Make the tooltip link directly to https://flutter.dev/docs/perf/rendering/shader#how-to-use-sksl-warmup to tell people how to fix the problem.
2. As a follow up CL, surface errors in the IDE when there is particularly bad shader compilation jank on app startup.
3. Perhaps show the fraction of Raster time that is shader jank as part of the bar (make it a stacked bar chart)?
4. Make the warning icon on the timeline blink when there is shader compilation jank or at least when there is bad shader compilation jank.

jacob314:
We should only show the warning when the GPU portion of a frame is > 16ms and probably only if the shader compilation jank is a significant portion of the time (e.g. maybe > 8ms?).

packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
```diff
@@ -315,13 +319,23 @@ class FlutterFramesChartItem extends StatelessWidget {
             color: defaultSelectionColor,
             height: selectedIndicatorHeight,
           ),
+        if (frame.hasShaderTime)
```
jacob314, line 322: use a minimum threshold for the amount of shader time (maybe 4 or 8ms) and only warn if the frame missed the raster time target.

packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
```diff
@@ -315,13 +319,23 @@ class FlutterFramesChartItem extends StatelessWidget {
...
     );
   }
 
   String _tooltipText(FlutterFrame frame) {
-    return 'UI: ${msText(frame.uiEventFlow.time.duration)}\n'
-        'Raster: ${msText(frame.rasterEventFlow.time.duration)}';
+    return [
+      'UI: ${msText(frame.uiEventFlow.time.duration)}',
+      'Raster: ${msText(frame.rasterEventFlow.time.duration)}',
+      if (frame.hasShaderTime) 'Skia.Shaders: ${msText(frame.shaderDuration)}',
```
jacob314, line 337: Add a link to the tooltip and a message explaining that shader compilation jank can usually be eliminated with caching.

packages/devtools_app/lib/src/performance/performance_model.dart
```diff
@@ -427,14 +427,30 @@ class FlutterFrame {
...
+
+  Duration get shaderDuration {
+    if (_shaderTime != null) return _shaderTime;
+    int sumShaderMicros = 0;
+    breadthFirstTraversal<TimelineEvent>(
+      rasterEventFlow,
+      action: (TimelineEvent event) {
+        // Shader events with a shader event parent should not be included in the
+        // sum because the parent event will encompass the time of [event].
+        if (event.isShaderEvent &&
```
jacob314, line 442: to make this less fragile: we could union all intervals of time with shader events. Add a TODO to add a helper class for performing efficient time interval union computation. Basically you should just be able to call addInterval repeatedly on the class and then query for the total interval without relying on  a specific structure of shader events. This current code could be broken if there ever happen to be non-shader events interleaved with shader events in which case it would report overly large ranges.

jacob314 (commented):
Can't wait to see this land!

packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
kenzieschmoll: done.

packages/devtools_app/lib/src/performance/performance_model.dart
kenzieschmoll: added todo

packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
kenzieschmoll: AFAIK flutter doesn't support rich tooltips with links, so we'll likely have to use the hovercard impl we use in the debugger for this. I'll look into that.

packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
```diff
@@ -330,11 +338,11 @@ class FlutterFramesChartItem extends StatelessWidget {
     );
   }
 
-  String _tooltipText(FlutterFrame frame) {
+  String _tooltipText(FlutterFrame frame, bool hasShaderJank) {
     return [
       'UI: ${msText(frame.uiEventFlow.time.duration)}',
       'Raster: ${msText(frame.rasterEventFlow.time.duration)}',
-      if (frame.hasShaderTime) 'Skia.Shaders: ${msText(frame.shaderDuration)}',
+      if (hasShaderJank) ...['Skia.Shaders: ${msText(frame.shaderDuration)}'],
```
jacob314, line 345: Lets give this a clearer name than Skia.Shaders. Perhaps `Shader Compilation` or use other terminology consistent with the documentation on early onset jank.

kenzieschmoll: done

kenzieschmoll:
> Make the warning icon on the timeline blink when there is shader compilation jank or at least when there is bad shader compilation jank.

dark-mode-shader-jank-animation light-mode-shader-jank-animation

packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
kenzieschmoll: added a TODO to do this in a follow up PR

jacob314: One other idea is to use the sort of overlay we use to warn about profile vs debug mode to warn about this.
The overlay could summarize the total impact of shader jank for the CPU profile. For example: 
"7 frames janked due to shader compilation with a total of ???ms spent on shader compilation.
To precomple shaders, see the instructions on this doc. That also has the benefit that users do not have to hover to see the message. We could also dispatch this as an error so it is shown to IDE users using the same Flutter.Error used for errors dispatched from Flutter. Fine to do that as a follow up PR.

jacob314:
> > Perhaps show the fraction of Raster time that is shader jank as part of the bar (make it a stacked bar chart)?
> 
> > Make the warning icon on the timeline blink when there is shader compilation jank or at least when there is bad shader compilation jank.
> 
> dark-mode-shader-jank-animation
> shader-compilation
> 
> > surface errors in the IDE when there is particularly bad shader compilation jank on app startup
> 
> filed [flutter/flutter-intellij#5581](https://github.com/flutter/flutter-intellij/issues/5581) to track

The yellow is a bit confusing as it seems like less of an error than red when if anything this is more of an error as there are more concrete steps to eliminate the time.

jacob314:
lgtm

kenzieschmoll:
> The yellow is a bit confusing as it seems like less of an error than red when if anything this is more of an error as there are more concrete steps to eliminate the time.

This matches the dark red we use in treemap diffs. WDYT?
Screen Shot 2021-06-15 at 1 09 42 PM
Screen Shot 2021-06-15 at 1 11 02 PM

InMatrix:
> Make the tooltip link directly to https://flutter.dev/docs/perf/rendering/shader#how-to-use-sksl-warmup to tell people how to fix the problem.

+1

Deep red for shader compilation looks fine.  

image

This tooltip could be misread, since the time for shader compilation is part of the time spent on the raster thread. Perhaps indenting shader compilation?

PR #3129, opened 2021-06-15
some cleanup to the CLI UI

devoncarew:
- some cleanup to the CLI UI

Here's the new output:

```
usage: devtools <options> [service protocol uri]

Open a DevTools instance in a browser and optionally connect to an existing application.

-h, --help                            Prints help output.
-v, --verbose                         Output more informational messages.
    --host=<host>                     Hostname to serve DevTools on (defaults to localhost).
    --port=<port>                     Port to serve DevTools on; specify 0 to automatically use any available port.
                                      (defaults to "9100")
    --[no-]launch-browser             Launches DevTools in a browser immediately at start.
                                      (defaults to on unless in --machine mode)
    --machine                         Sets output format to JSON for consumption in tools.

Memory profiling options:
    --record-memory-profile=<file>    Start devtools headlessly and write memory profiling samples to the indicated file.
                                      (defaults to "memory_samples.json")

App size options;
    --appSizeBase=<appSizeBase>       Path to the base app size file used for app size debugging.
    --appSizeTest=<appSizeTest>       Path to the test app size file used for app size debugging.
                                      This file should only be specified if --appSizeBase is also specified.

Advanced options:
    --try-ports=<count>               The number of ascending ports to try binding to before failing with an error. 
                                      (defaults to "10")
    --enable-notifications            Requests notification permissions immediately when a client connects back to the server.
    --allow-embedding                 Allow embedding DevTools inside an iframe.
    --headless                        Causes the server to spawn Chrome in headless mode for use in automated testing.
```

### DIFF
--- a/packages/devtools_server/lib/src/server.dart
+++ b/packages/devtools_server/lib/src/server.dart
@@ -39,7 +39,6 @@ const argMachine = 'machine';
 const argHost = 'host';
 const argPort = 'port';
 const argProfileMemory = 'record-memory-profile';
-const argProfileMemoryOld = 'profile-memory';
 const argTryPorts = 'try-ports';
 const argVerbose = 'verbose';
 const launchDevToolsService = 'launchDevTools';
@@ -119,8 +118,6 @@ Future<HttpServer?> _serveDevToolsWithArgs(
   String? profileFilename;
   if (args.wasParsed(argProfileMemory)) {
     profileFilename = args[argProfileMemory];
-  } else if (args.wasParsed(argProfileMemoryOld)) {
-    profileFilename = args[argProfileMemoryOld];
   }
   if (profileFilename != null && !path.isAbsolute(profileFilename)) {
     profileFilename = path.absolute(profileFilename);
@@ -498,7 +495,6 @@ ArgParser _createArgsParser(bool verbose) {
     )
     ..addOption(
       argHost,
-      //defaultsTo: 'localhost',
       valueHelp: 'host',
       help: 'Hostname to serve DevTools on (defaults to localhost).',
     )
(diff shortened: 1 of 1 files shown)


packages/devtools_server/lib/src/server.dart
```diff
@@ -586,7 +582,17 @@ ArgParser _createArgsParser(bool verbose) {
       hide: !verbose,
     );
 
-  // Hidden args.
+  // Deprecated and hidden args.
+  // TODO: Remove this - prefer that clients use the rest arg.
+  parser
+    ..addOption(
+      argVmUri,
```
devoncarew, line 589: Do we know whether this can be removed yet? `--vm-uri`

kenzieschmoll: @jacob314 do you know if this is used anywhere?

jacob314: I just verified. It is not used in vscode, flutter-intellij or flutter.

packages/devtools_server/lib/src/server.dart
```diff
@@ -514,47 +510,47 @@ ArgParser _createArgsParser(bool verbose) {
...
-    )
+    );
+
+  if (verbose) {
+    parser.addSeparator('App size options;');
+  }
+
+  parser
     ..addOption(
       argAppSizeBase,
```
devoncarew, line 537: Longer term, we way want to deprecate `--appSizeBase` and `--appSizeTest` in favor of naming the options something like `--app-size-base` and `--app-size-test`.

kenzieschmoll: Would you mind filing an issue for this and linking it in a TODO here?

kenzieschmoll (approved):

packages/devtools_server/lib/src/server.dart
devoncarew: done! https://github.com/flutter/devtools/issues/3146

PR #3130, opened 2021-06-15
lint for depend_on_referenced_packages

devoncarew:
- lint for `depend_on_referenced_packages`

cc @pq

### DIFF
--- a/packages/devtools_app/assets/scripts/inspector_polyfill_script.dart
+++ b/packages/devtools_app/assets/scripts/inspector_polyfill_script.dart
@@ -74,7 +74,6 @@ import 'package:flutter/src/widgets/debug.dart';
 import 'package:flutter/src/widgets/framework.dart';
 import 'package:flutter/src/widgets/gesture_detector.dart';
 import 'package:flutter/src/widgets/widget_inspector.dart';
-// ignore: depend_on_referenced_packages
 import 'package:vector_math/vector_math_64.dart';
 
 // This extension method is simply a hack to make a private member defined on
--- a/packages/devtools_app/assets/scripts/inspector_polyfill_script.dart
+++ b/packages/devtools_app/assets/scripts/inspector_polyfill_script.dart
@@ -74,6 +74,7 @@ import 'package:flutter/src/widgets/debug.dart';
 import 'package:flutter/src/widgets/framework.dart';
 import 'package:flutter/src/widgets/gesture_detector.dart';
 import 'package:flutter/src/widgets/widget_inspector.dart';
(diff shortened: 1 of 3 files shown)


packages/devtools_app/assets/scripts/inspector_polyfill_script.dart
```diff
@@ -74,6 +74,7 @@ import 'package:flutter/src/widgets/debug.dart';
 import 'package:flutter/src/widgets/framework.dart';
 import 'package:flutter/src/widgets/gesture_detector.dart';
 import 'package:flutter/src/widgets/widget_inspector.dart';
+// ignore: depend_on_referenced_packages
```
pq, line 77: Why not add a dependency on `package:vector_math` tp the pubspec?

devoncarew: Good question! This dart file likely shouldn't participate in the package's direct dependencies? It's not used in the context of this package, but is loaded into a running flutter app and eval'd via the service protocol.

I'm guessing that this should actually be re-worked to:
- remove the ignore
- add a dev_dep on package:vector_math

devoncarew:
Updated to add package:vector_math as a dev dependency.

devoncarew:
Interesting - from the CI results, `depend_on_referenced_packages` is in the master channel of Flutter, but not yet in the `dev` or `beta` channels.

pq (approved):
👍

packages/devtools_app/pubspec.yaml
```diff
@@ -64,6 +64,7 @@ dev_dependencies:
   flutter_test:
     sdk: flutter
   mockito: ^5.0.9
+  vector_math: ^2.0.0
```
kenzieschmoll, line 67: why is this in dev_dependenicies and not dependencies? Is the inspector_polyfil_script not used in a release build at runtime? @jacob314

devoncarew: I believe that this script - `packages/devtools_app/assets/scripts/inspector_polyfill_script.dart` - is not used by the devtools source itself, but is eval'd in the context of the running flutter app. So it really needs `vector_math` to be in the deps of the app under debug. We just add an entry into the devtools pubspec so that we can get complete analysis results for the file - to know that it analyzes cleanly.

Completely aside, we might want to work to reduce the amount of code that we're evaling in the running app - to incrementally push more of the functionality into the in-app inspector implementation.

kenzieschmoll (approved):

devoncarew:
closing - I assume this is stale

PR #3134, opened 2021-06-16
Handle pre-release versions in Flutter version parsing and comparison

kenzieschmoll:
This PR also uses this new version parsing code to determine whether to show the legacy or new performance page.

### DIFF
--- a/packages/devtools_app/lib/src/version.dart
+++ b/packages/devtools_app/lib/src/version.dart
@@ -17,18 +17,34 @@ class FlutterVersion extends SemanticVersion {
     @required this.engineRevision,
     @required this.dartSdkVersion,
   }) {
+    final splitOnDash = version.split('-');
+
+    final semVersion = splitOnDash.first;
+
     // Flutter versions can come in as '1.10.7-pre.42', so we strip out any
     // characters that are not digits. We do not currently have a need to know
     // more version parts than major, minor, and patch. If this changes, we can
     // add support for the extra values.
-    final _versionParts = version
-        .split('.')
-        .map((part) => RegExp(r'\d+').stringMatch(part) ?? '0')
-        .toList();
+    final _versionParts = semVersion.split('.');
     major =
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/version.dart
```diff
@@ -17,18 +17,34 @@ class FlutterVersion extends SemanticVersion {
     @required this.engineRevision,
     @required this.dartSdkVersion,
   }) {
+    final splitOnDash = version.split('-');
```
jacob314, line 20: comment what the format of a pre release version is expected to look like here.

packages/devtools_app/lib/src/version.dart
```diff
@@ -17,18 +17,34 @@ class FlutterVersion extends SemanticVersion {
...
-        .split('.')
-        .map((part) => RegExp(r'\d+').stringMatch(part) ?? '0')
-        .toList();
+    final _versionParts = semVersion.split('.');
     major =
         _versionParts.isNotEmpty ? int.tryParse(_versionParts.first) ?? 0 : 0;
     minor = _versionParts.length > 1 ? int.tryParse(_versionParts[1]) ?? 0 : 0;
     patch = _versionParts.length > 2 ? int.tryParse(_versionParts[2]) ?? 0 : 0;
+
+    if (splitOnDash.length == 2) {
```
jacob314, line 34: assert splitOnDash.length <=2

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/version.dart
kenzieschmoll: done

packages/devtools_app/lib/src/version.dart
kenzieschmoll: done

kenzieschmoll:
Build is broken due to unrelated Inspector Golden failures. Will send a follow up PR to fix the bots.

PR #3136, opened 2021-06-16
Make provider code  bit more resilient to null values.

jacob314:
These exceptions don't impact the UI but are annoying when running with pause on exception.

### DIFF
--- a/packages/devtools_app/lib/src/provider/instance_viewer/instance_providers.dart
+++ b/packages/devtools_app/lib/src/provider/instance_viewer/instance_providers.dart
@@ -294,6 +294,12 @@ final AutoDisposeFutureProviderFamily<InstanceDetails, InstancePath>
 
   final instance = await eval.getInstance(instanceRef, isAlive);
 
+  if (instance == null) {
+    // TODO(jacobr): we might want to surface this as unavailable rather than
+    // null.
+    return InstanceDetails.nill(setter: setter);
+  }
+
   switch (instance.kind) {
     case InstanceKind.kNull:
       return InstanceDetails.nill(setter: setter);
@@ -360,6 +366,9 @@ final AutoDisposeFutureProviderFamily<InstanceDetails, InstancePath>
       if (enumDetails != null) return enumDetails;
 
       final classInstance = await eval.getClass(instance.classRef, isAlive);
+      if (classInstance == null) {
+        return InstanceDetails.nill(setter: setter);
+      }
       final evalForInstance =
           ref.watch(libraryEvalProvider(classInstance.library.uri).future);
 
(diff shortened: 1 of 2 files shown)


jacob314:
Fyi @rrousselGit

packages/devtools_app/lib/src/provider/provider_nodes.dart
```diff
@@ -49,9 +49,13 @@ final _rawProviderIdsProvider =
       eval.getInstance(idRef, isAlive)
   ]);
 
-  return [
-    for (final idInstance in providerIdInstances) idInstance.valueAsString,
-  ];
+  final ret = <String>[];
+  for (final idInstance in providerIdInstances) {
+    if (idInstance != null) {
+      ret.add(idInstance.valueAsString);
```
rrousselGit, line 55: Hum this is weird that "null" is received here

In which case does this happen?

packages/devtools_app/lib/src/provider/instance_viewer/instance_providers.dart
```diff
@@ -360,6 +366,9 @@ final AutoDisposeFutureProviderFamily<InstanceDetails, InstancePath>
       if (enumDetails != null) return enumDetails;
 
       final classInstance = await eval.getClass(instance.classRef, isAlive);
+      if (classInstance == null) {
```
rrousselGit, line 369: How is it possible that an object has no associated class?

packages/devtools_app/lib/src/provider/provider_nodes.dart
```diff
@@ -49,9 +49,13 @@ final _rawProviderIdsProvider =
       eval.getInstance(idRef, isAlive)
   ]);
 
-  return [
-    for (final idInstance in providerIdInstances) idInstance.valueAsString,
-  ];
+  final ret = <String>[];
```
kenzieschmoll, line 52: nit: can we have a more descriptive name here?

kenzieschmoll (approved):

packages/devtools_app/lib/src/provider/instance_viewer/instance_providers.dart
jacob314: if isAlive got disposed part way through, eval will short circuit with a null response.

packages/devtools_app/lib/src/provider/provider_nodes.dart
jacob314: Just using the https://github.com/flutter/samples/tree/master/provider_shopper
sample.

packages/devtools_app/lib/src/provider/instance_viewer/instance_providers.dart
rrousselGit: Oh I see. I forgot that only "safeEval" throws if the isAlive was disposed.

Throwing "CancelledException" should be ideal then. That won't cause undesired logs as it is filtered

rrousselGit: Could you add a to-do that asks me to inspect all getClass/co calls to spot issues?

I doubt this is the only case.

Hixie:
What is the status of this PR? It seems to have been abandoned?

PR #3154, opened 2021-06-22
Show the event summary view for UI events in the performance page

kenzieschmoll:
We show this view for non-UI events in the performance page, but we currently only show the CPU profiler for UI events. After this PR, the summary view will be available for all events in the timeline.

Screen Shot 2021-06-22 at 9 11 14 AM

### DIFF
--- a/packages/devtools_app/lib/src/performance/event_details.dart
+++ b/packages/devtools_app/lib/src/performance/event_details.dart
@@ -134,81 +134,172 @@ class EventSummary extends StatelessWidget {
 
   TraceEvent get firstTraceEvent => event.traceEvents.first.event;
 
-  // TODO(kenz): make this view more dense.
   @override
   Widget build(BuildContext context) {
     return ListView(
+      padding: const EdgeInsets.symmetric(vertical: denseSpacing),
       children: [
-        ListTile(
-          title: const Text('Time'),
-          subtitle: Text('${event.time.start.inMicroseconds} μs —  '
-              '${event.time.end.inMicroseconds} μs'),
+        // Wrap with horizontal padding so that these items align with the
+        // expanding data items. Adding horizontal padding to the entire list
+        // affects the hover boundary for the clickable expanding tiles.
+        Padding(
+          padding: const EdgeInsets.symmetric(horizontal: defaultSpacing),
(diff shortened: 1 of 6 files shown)


jacob314:
I would include the total time elapsed for the event in the summary as well as the title as that is more actionable than the stand and end times which are pretty meaningless.

packages/devtools_app/lib/src/profiler/cpu_profiler.dart
```diff
@@ -225,7 +229,12 @@ class _CpuProfilerState extends State<CpuProfiler>
       },
     );
     // TODO(kenz): make this order configurable.
-    return [bottomUp, callTree, cpuFlameChart];
+    return [
+      if (widget.summaryView != null) widget.summaryView,
+      widget.data.isEmpty ? _buildEmptyDataView() : bottomUp,
```
jacob314, line 234: it would be nice to not show these tabs at all rather than showing them with empty data when they are not available.

kenzieschmoll: done

kenzieschmoll:
Cleaned up the event details view. PTAL
Screen Shot 2021-06-22 at 2 53 04 PM

packages/devtools_app/lib/src/performance/event_details.dart
```diff
@@ -134,81 +134,172 @@ class EventSummary extends StatelessWidget {
...
-          subtitle: FormattedJson(json: _eventArgs),
+        FormattedJson(
+          json: _eventArgs,
+          useSubtleStyle: true,
         ),
       ],
     );
   }
+
+  Widget _buildDataItem({
```
jacob314, line 255: nit: add a TODO to make this and other widget like methods in this class widgets.
I think a good rule is that if a function takes a BuildContext, it should probably be a Widget.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/performance/event_details.dart
kenzieschmoll: went ahead and converted to stateless widgets in this PR

PR #3155, opened 2021-06-23
convert the repo tool lib to null safety

devoncarew:
- convert the repo `tool/` lib to null safety

If the `generate_changelog.dart` command is not used (I'm guessing not?), we should just delete the command.

### DIFF
--- a/packages/devtools/analysis_options.yaml
+++ b/packages/devtools/analysis_options.yaml
@@ -1,0 +1,5 @@
+include: ../../analysis_options.yaml
+
+analyzer:
+  exclude:
+    - build/**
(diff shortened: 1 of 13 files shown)


kenzieschmoll:
> If the generate_changelog.dart command is not used (I'm guessing not?), we should just delete the command.

I actually have some work on a local branch to fix up this script. This would save us a lot of time for when we release.

devoncarew:
> I actually have some work on a local branch to fix up this script. This would save us a lot of time for when we release.

I can back my changes out for this file to avoid merge conflicts here -

devoncarew:
Updated to back out changes to `tool/lib/commands/generate_changelog.dart`.

kenzieschmoll (approved):

Issue #3158, opened 2021-06-23
devtools can't open the service URL that flutter tools prints


devoncarew:
When you run an app from the command line (`flutter run`), it prints the following text:

```
The Flutter DevTools debugger and profiler on iPhone 12 Pro Max is available at:
http://127.0.0.1:9101?uri=http%3A%2F%2F127.0.0.1%3A56142%2FHOwgrxalK00%3D%2F
```

When you try and copy and paste the given url into an existing DevTools instance, in the initial connect page, the given url doesn't work - we can't resolve it to a service protocol uri. You have to know to copy the portion after the `...9101?uri=` into the connect field. We should:

- have flutter tools print a more user friendly url - for the one above, the service protocol url is itself uri encoded. We'd want something like `'http://127.0.0.1:9101?host=127.0.0.1&port=56142&token=HOwgrxalK00'`
- have the DevTools app understand the host, port, and token url parameters and know how to then create the service protocol ws: uri
- special case a devtools url (look for `?uri=http`) and paste the bits after the equals

devoncarew:
cc @jonahwilliams wrt the extra encoded service url appended to the devtools uri; we should make that printout a bit cleaner (through, that will take some upstream work in how devtools processes the uri params).

jonahwilliams:
I could un-format things so they come out more normal looking. Let me try that

devoncarew:
I have a DevTools PR which will let us do: `'<devtools uri>?port=xxx&token=yyy'` - that should clean up the URI quite a bit.

PR #3161, opened 2021-06-23
support more uri params for specifying the service uri

devoncarew:
- support more uri params for specifying the service uri

This will allow us to improve the uri that flutter tools prints out (see https://github.com/flutter/devtools/issues/3158).


packages/devtools_app/lib/src/framework/framework_core.dart
```diff
@@ -72,8 +72,8 @@ class FrameworkCore {
   }
 
   /// Gets a VM Service URI from the querystring (in preference from the 'uri'
-  /// value, but otherwise from 'port').
-  static Uri _getUriFromQuerystring(String location) {
+  /// value, but otherwise from 'port' and 'token').
+  static Uri _getUriFromQueryString(String location) {
```
kenzieschmoll, line 76: we should add a unit test for this method.

kenzieschmoll (approved):

packages/devtools_app/lib/src/framework/framework_core.dart
devoncarew: Added!

devoncarew:
Updated the PR w/ a test. I'll look into addressing the other CI failures - in separate PRs - in order to be able to land this on green.

Issue #2958, opened 2021-04-29
Create minimalist `package:devtools_server_api` for SDK integration


bkonyi:
We'd like to avoid depending directly on `package:devtools_server` (and to a lesser extent, `package:devtools_shared`) in DDS to try and keep the number of transitive dependencies down where possible. DDS uses functionality from `server_api.dart`, so `package:devtools_server_api` would only require the following files to be migrated (these are currently forked within DDS):

- `file_system.dart`
- `server_api.dart`
- `usage.dart`

Which would only depend on:
- `package:devtools_shared`
- `package:usage`
- `package:path`
- `package:shelf`

Most of which are currently used within the SDK.

cc @kenzieschmoll @jacob314

PR #3164, opened 2021-06-28
refactor devtools_server to minimize transitive deps

devoncarew:
- refactor devtools_server to minimize transitive deps
- fix https://github.com/flutter/devtools/issues/2958

This separates code out of package:devtools_server (and into package:devtools_shared) to allow code in the dart sdk (DDS) to depend on package:devtools_shared.

There are a lot of use cases that are supported by package:devtools, package:devtools_server, and package:devtools_shared, so it's hard to make changes to this area.

I'm not committed to this specific refactoring of the packages. We could also choose to instead move some code into package:devtools, and have package:devtools_server retain just the code that DDS needs to depend on.


devoncarew:
cc @bkonyi

bkonyi:
This looks perfect! Pulling the common server functionality into `devtools_shared` makes sense to me.

packages/devtools_server/lib/src/memory_profile.dart
```diff
@@ -9,7 +9,7 @@ import 'package:devtools_shared/devtools_shared.dart';
 import 'package:intl/intl.dart' as intl;
 import 'package:vm_service/vm_service.dart';
 
-import 'service_registrations.dart' as registrations;
```
kenzieschmoll, line 12: I don't think this file rename makes sense, as not all of the service code in service_registrations.dart are pertaining to memory features

packages/devtools_shared/pubspec.yaml
```diff
@@ -11,4 +11,7 @@ environment:
   sdk: '>=2.12.0 <3.0.0'
 
 dependencies:
+  path: ^1.8.0
```
kenzieschmoll, line 14: should we be able to remove these deps from the devtools_server pubspec now?

kenzieschmoll:
Note that this change will require some mirrored modifications to the google3 entry point, but I don't see any red flags.

packages/devtools_server/lib/src/memory_profile.dart
devoncarew: OK, I reverted the rename.

packages/devtools_shared/pubspec.yaml
devoncarew: I'll verify which deps are actually necessary now.

devoncarew: I was able to remove the dep on package:usage.

PR #3166, opened 2021-06-28
ignore a reference to a deprecated item

devoncarew:
- ignore a reference to a deprecated item

This should fix a recent CI breakage.


jacob314 (approved):
lgtm

Issue #3144, opened 2021-06-17
Debug toggle buttons redesign proposal


CodemateTechAgency:
# Overview

Hello, here are the Codemate team's initial suggestions for redesigning the debug toggle buttons.

We initially interviewed several of our own developers, including some unfamiliar with Flutter or the DevTools, to understand how well they understood the current design.

There were some toggles that almost nobody fully understood, including developers familiar with Flutter. We've used their feedback to guide our redesign.

For each debug toggle we provide a summary of our findings from interviewing users, and a proposal for a new design.

# Proposal

We propose making the icons monochrome and more legible, and grouping the buttons together:

![Screenshot 2021-06-17 at 13 20 40](https://user-images.githubusercontent.com/82218649/122461485-164edb00-cfbc-11eb-8c7b-5ed8c7275952.png)

## **Slow Animations**

**Findings:** all users we interviewed fully understood what this toggle does without requiring any additional explanation.

**Proposal:** change the clock icon to a stopwatch. Not a huge change, but we feel this reflects a bit more of the “toggling time” intention of the action.

<img src="https://user-images.githubusercontent.com/82218649/122461899-99703100-cfbc-11eb-9e7b-b36e3c4bc0ef.png" height="50">

## **Debug Paint**

**Findings:** users found it extremely hard to guess what this button does; guesses included believing it would be used to find performance issues. Once enabling it, users found the visual guidelines overwhelming, and found it difficult to understand that the various colors, borders and arrows represent. They also found it hard to understand what the guidelines would be useful for.

**Proposal:** change the icon to be simpler and more legible, rename the action to “Show Guidelines” to better describe the functionality.

<img src="https://user-images.githubusercontent.com/82218649/122462023-bf95d100-cfbc-11eb-8c0e-24e15ed9713b.png" height="50">

For further improvements outside the current scope, we propose finer-grained control of which guidelines to show; this would require adding functionality to the Flutter framework:

<img alt="Screenshot 2021-06-18 at 19 01 27" src="https://user-images.githubusercontent.com/82218649/122588391-b6157300-d067-11eb-9854-8a89600b0f14.png">

## **Paint Baselines**

**Findings:** users who were familiar with the concept of font baselines understood this function, but several were unaware of what baselines are, so didn't understand the meaning of this button. Developers struggled with understanding what this function is useful for.

[...]

PR #3167, opened 2021-06-28
Implement redesign of debug toggle buttons (#3144)

alexlindroos:
Hello from Codemate! 

Here is our PR for the implementation of the redesign for debug toggle buttons. 

We have changed the icons, fill color, title colors and grouping of the button. Tooltips will come in a different PR.

Have a nice day!


tomgilder:
@googlebot I consent

jacob314:
In the future, please include screenshots showing before and after for the UI changed by the CL. That makes the code quicker to review without requiring us to manually run the code.

packages/devtools_app/lib/src/ui/icons.dart
```diff
@@ -251,6 +251,27 @@ Image createImageIcon(String url, {double size = defaultIconSize}) {
   );
 }
 
+class ThemedImageIcon extends StatelessWidget {
```
jacob314, line 254: a lot of the calls to this class have the exact same file for both light mode and dark mode. Create a separate ImageIcon class for that case to improve readability.

packages/devtools_app/lib/src/service_extensions.dart
```diff
@@ -258,7 +339,14 @@ final toggleSelectWidgetMode = ToggleableServiceExtensionDescription<bool>._(
 final enableOnDeviceInspector = ToggleableServiceExtensionDescription<bool>._(
   extension: 'ext.flutter.inspector.enable',
   description: 'Enable on-device inspector',
-  icon: createImageIcon('icons/general/locate@2x.png'),
+  disabledIcon: const ThemedImageIcon(
```
jacob314, line 342: example of a case where ThemedImageIcon shouldn't be used as the light and dark mode assets are the same. Create a separate simpler ImageIcon class or just use createImageIcon as before. Specifying the same asset twice for this case just makes the code harder to read.

jacob314 (commented):
Some minor comments then LGTM.

alexlindroos:
Hi @jacob314, thank you for the review. I created a simpler AssetImageIcon to match non themed use cases, like you suggested.

alexlindroos:
@jacob314 I also changed the includeTextWidths for the buttons. My suggestion would be that first we will collapse the "Refresh tree"-button and then right side service extension group, then last the "Select widget mode"-button. 

What you think? 

pr_collapsing_buttons

alexlindroos:
Fixed the name of "Invert Oversized Images" to new "Highlight Oversized Images". I also did some cleaning for the old name references of inspector debug tools.

Latest:

Screenshot 2021-06-30 at 12 42 00

pr_collapsing_buttons_fixed

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -199,18 +201,11 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
-      ),
-      const SizedBox(width: denseSpacing),
-      ServiceExtensionButtonGroup(
-        minIncludeTextWidth: 1050,
-        extensions: [extensions.debugPaint, extensions.debugPaintBaselines],
-      ),
-      const SizedBox(width: denseSpacing),
-      ServiceExtensionButtonGroup(
-        minIncludeTextWidth: 1250,
+        minIncludeTextWidth: 1150,
```
kenzieschmoll, line 204: put this in a const above `serviceExtensionButtonsIncludeTextWith`

packages/devtools_app/lib/src/service_extensions.dart
```diff
@@ -96,44 +101,68 @@ final debugAllowBanner = ToggleableServiceExtensionDescription<bool>._(
...
+  ),
+  enabledIcon: const ThemedImageIcon(
+    lightModeAsset: 'icons/baselines-white.png',
+    darkModeAsset: 'icons/baselines-dgrey.png',
+  ),
   enabledValue: true,
   disabledValue: false,
-  enabledTooltip: 'Hide Paint Baselines',
-  disabledTooltip: 'Show Paint Baselines',
+  enabledTooltip: 'Hide Baselines',
```
kenzieschmoll, line 153: We should add a TODO in this file to add richer tooltips, as described in https://docs.google.com/document/d/1xBJpVykI7-UYP61MX6bKZMl6awQwNvEc3jurDjSZEVE/edit?ts=60db89a6#heading=h.ytka7us6qnox. If we don't already have an issue in the repo to track this, we should also create one and link to it in the TODO

packages/devtools_app/lib/src/service_extensions.dart
```diff
@@ -51,7 +52,8 @@ class ToggleableServiceExtensionDescription<T>
 
 class ServiceExtensionDescription<T> {
   ServiceExtensionDescription({
-    this.icon,
+    this.disabledIcon,
+    this.enabledIcon,
```
kenzieschmoll, line 56: it seems there are a few cases where enabledIcon and disabledIcon are the same. Perhaps we should have `disabledIcon` default to `enabledIcon` when `enabledIcon != null` and disabled icon is not specified.

packages/devtools_app/lib/src/service_extensions.dart
```diff
@@ -51,7 +52,8 @@ class ToggleableServiceExtensionDescription<T>
 
 class ServiceExtensionDescription<T> {
   ServiceExtensionDescription({
-    this.icon,
+    this.disabledIcon,
```
kenzieschmoll, line 55: should these both be required fields?

packages/devtools_app/lib/src/ui/service_extension_widgets.dart
```diff
@@ -143,7 +145,9 @@ class _ServiceExtensionButtonGroupState
               : 0.0,
         ),
         child: ImageIconLabel(
-          description.icon,
+          extensionState.isSelected
+              ? description.enabledIcon
+              : description.disabledIcon,
```
kenzieschmoll, line 150: what if either are null? If we make these fields required (one or both - see earlier comment about making disabledIcon default to the value of enabledIcon when null), we can avoid this consideration

packages/devtools_app/lib/src/service_extensions.dart
alexlindroos: Service extensions `httpEnableTimelineLogging` and `socketProfiling` does not have any kind of icon.  Either we need to add an icon for those or keep icon not required. Do we have any icons that could be used?

packages/devtools_app/lib/src/inspector/inspector_screen.dart
alexlindroos: done

packages/devtools_app/lib/src/service_extensions.dart
alexlindroos: done

packages/devtools_app/lib/src/service_extensions.dart
alexlindroos: It seems that they are not under any button and are called in network recording. So could I just add some icons for them to be used in the future if needed, but for now just to please required for enabled and disabled icon?

For the icons, what about `Icons.timeline` for `httpEnableTimelineLogging` and `Icons.bar_chart` for `socketProfiling`? 
Screenshot 2021-07-01 at 14 25 26
Screenshot 2021-07-01 at 14 35 36

alexlindroos: Or one option is to just add Container-widget or Placeholder-widget for it, since the type for it is a Widget

kenzieschmoll: Let's just make both of these String consts with their extension values `'ext.dart.io.httpEnableTimelineLogging'` and `'ext.dart.io.socketProfilingEnabled'`. This is all we use them for right now anyway

packages/devtools_app/lib/src/service_extensions.dart
alexlindroos: Yep, I will make `enabledIcon` required and default `disabledIcon` to it if not specified

packages/devtools_app/lib/src/ui/service_extension_widgets.dart
alexlindroos: Yep, I will make `enabledIcon` required and default `disabledIcon` to it if not specified

packages/devtools_app/lib/src/service_extensions.dart
alexlindroos: I think we can't do that, because `httpEnableTimelineLogging` and `socketProfilingEnabled` needs to be in `Map<String, ServiceExtensionDescription> serviceExtensionsAllowlist`. Which is used to do extension checks in multiple places even for `httpEnableTimelineLogging` and `socketProfilingEnabled`. For now I will put Placeholder() as `enabledIcon` to pass the required parameter, is it okay or am I missing something?

alexlindroos:
@kenzieschmoll Does the PR look good now or is there something that needs to be changed?

packages/devtools_app/lib/src/service_extensions.dart
jacob314: We should add these extensions to serviceExtensionAllowlist probably adding extra functionality to the related logic for serviceExtensionAllowlist to handle the fact that these extensions should be triggered for all isolates not just the main isolate.

kenzieschmoll (approved):
The golden file change in the PR is already submitted at head - you may want to merge master, then LGTM

kenzieschmoll:
Submitted a PR that fixes the issue seen on the bots here. Landing this.

PR #3168, opened 2021-06-28
Make Flutter Frame timeline live and migrate to the FrameTiming API

kenzieschmoll:
live-timeline

Fixes https://github.com/flutter/devtools/issues/2918

Tests are WIP - will work on those while this change is under review.

Apologies for the messy commit history - avoiding a nasty rebase.


packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -688,15 +690,15 @@ class AreaPaneHeader extends StatelessWidget implements PreferredSizeWidget {
...
-              ),
+            DefaultTextStyle(
+              maxLines: maxLines,
+              overflow: TextOverflow.ellipsis,
+              style: theme.textTheme.subtitle2,
+              child: title,
             ),
-            ...actions,
+            ...leftActions,
+            const Spacer(),
```
jacob314, line 700: only add spacer if leftActions and rightActions are both not empty.

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -130,6 +123,43 @@ class PerformanceController
   /// This list is cleared and repopulated each time "Refresh" is clicked.
   List<TraceEventWrapper> allTraceEvents = [];
 
+  /// The tracking index for the first unprocessed trace event collected.
+  int nextTraceIndexToProcess = 0;
```
jacob314, line 127: nit: can these be made private or perhaps with a public getter if needed?

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -356,14 +517,37 @@ class PerformanceController
 
   void addTimelineEvent(TimelineEvent event) {
     data.addTimelineEvent(event);
+    _timelineEvents.value.add(event);
```
jacob314, line 520: Adding elements to a ValueListenable without notifying isn't safe. Can you use ListValueNotifier or change the _timelineEvents so they are not a ValueListenable?

packages/devtools_app/lib/src/common_widgets.dart
kenzieschmoll: done.

packages/devtools_app/lib/src/performance/performance_controller.dart
kenzieschmoll: done

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -130,6 +123,43 @@ class PerformanceController
...
+  /// recording has been paused.
+  final _pendingFlutterFrames = <FlutterFrame>[];
+
+  /// The collection of [TimelineEvent]s that should be linked to
+  /// [FlutterFrame]s but have not yet been assigned.
+  ///
+  /// These timeline events are keyed by the [FlutterFrame] ID specified in the
+  /// event arguments, which matches the ID for the corresponding
+  /// [FlutterFrame].
+  final unassignedFlutterFrameEvents = <int, List<TimelineEvent>>{};
```
jacob314, line 154: Maybe replace List<TimelineEvent> with a class with a clearer api.
Having the indexes in the List happen to be FlutterFrame ids is a little surprising.
class could be `UnasignedEvents` or similar.

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -138,27 +168,84 @@ class PerformanceController
...
+
+      _processing.value = true;
+      await processTraceEvents(allTraceEvents);
+      _processing.value = false;
+
+      // Listen for new timeline events.
+      autoDispose(serviceManager.service.onTimelineEvent.listen((event) {
+        final eventBatch = <TraceEventWrapper>[];
+        if (event.json['kind'] == 'TimelineEvents') {
+          final List<dynamic> traceEvents = event.json['timelineEvents']
```
jacob314, line 228: why is this a List<dynamic>?

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -138,27 +168,84 @@ class PerformanceController
...
+              .map(
+                (e) => TraceEventWrapper(
+                  TraceEvent(e),
+                  DateTime.now().millisecondsSinceEpoch,
+                ),
+              )
+              .toList();
+          final List<TraceEventWrapper> wrappedTraceEvents =
```
jacob314, line 236: avoid types on LHS.

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -138,27 +168,84 @@ class PerformanceController
...
+              traceEvents.cast<TraceEventWrapper>();
+          eventBatch.addAll(wrappedTraceEvents);
+        }
+        allTraceEvents.addAll(eventBatch);
+      }));
+    }
+  }
+
+  FutureOr<void> processAvailableEvents() async {
+    _processing.value = true;
```
jacob314, line 246: nit: add 
assert(!_processing.value) at the start of the method to make sure we don't get into a state where we have two pending processing requests.

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -208,6 +295,16 @@ class PerformanceController
...
     }
 
+    if (!frame.isWellFormed &&
+        (_firstWellFormedFrameMicros == null ||
+            frame.timeFromFrameTiming.start.inMicroseconds >=
+                _firstWellFormedFrameMicros)) {
+      // Only try to pull timeline events for frames that are after the first
+      // well formed frame. Timeline events that occurred before this frame will
+      // have already fallen out of the buffer.
+      await processAvailableEvents();
```
jacob314, line 305: Are we concerned this will hide timeline events on startup that occur before the first flutter frame (assuming such events can exist)?
We could always add an artificial "before startup" frame to disambiguate this case from the case you are probably trying to avoid of incomplete frames due to full buffers.

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -248,56 +353,112 @@ class PerformanceController
...
+        frame.setEventFlow(event, type: type);
+        if (frame.isWellFormed) {
+          unassignedFlutterFrames.remove(frameNumber);
+          _updateFirstWellFormedFrameMicros(frame);
+        }
+      } else {
+        final unassignedEventsForFrame =
+            unassignedFlutterFrameEvents.putIfAbsent(
+          frameNumber,
+          () => List.generate(2, (_) => null),
```
jacob314, line 432: This is a case where having an UnasignedEvents class would make it clearer.

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -248,56 +353,112 @@ class PerformanceController
...
+      if (frame.isJanky(_displayRefreshRate.value)) {
+        serviceManager.errorBadgeManager
+            .incrementBadgeCount(PerformanceScreen.id);
       }
     }
   }
 
+  void toggleRecordingFrames(bool recording) {
+    _recordingFrames.value = recording;
+    if (_recordingFrames.value) {
```
jacob314, line 457: tiny nit: maybe condition on `recording`  instead of the value of _recordingFrames.value

packages/devtools_app/lib/src/performance/performance_model.dart
```diff
@@ -284,7 +292,15 @@ class OfflinePerformanceData extends PerformanceData {
     final CpuProfileData cpuProfileData =
         cpuProfileJson.isNotEmpty ? CpuProfileData.parse(cpuProfileJson) : null;
 
-    final String selectedFrameId = json[PerformanceData.selectedFrameIdKey];
+    final int selectedFrameId = json[PerformanceData.selectedFrameIdKey];
+
```
jacob314, line 296: nit: avoid dynamic where possible preferring Object so we get better static errors.

packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
```diff
@@ -390,16 +437,36 @@ class TimelineFlameChartState
...
       // Zoom and scroll to the frame's UI event.
+      TimeRange time;
+      TimelineEvent event;
+      if (_selectedFrame.uiEventFlow == null &&
+          _selectedFrame.rasterEventFlow == null) {
+        // TODO(kenz): should we zoom to the latest available frame?
+        Notifications.of(context)
+            .push('No timeline events available for the selected frame');
+        return;
+      } else if (_selectedFrame.uiEventFlow == null) {
```
jacob314, line 456: should these checks be added to a helper on _slectedFrame object? get the current time and event?

packages/devtools_app/lib/src/performance/timeline_flame_chart.dart
```diff
@@ -390,16 +437,36 @@ class TimelineFlameChartState
...
+        time = _selectedFrame.rasterEventFlow.time;
+        event = _selectedFrame.rasterEventFlow;
+      } else if (_selectedFrame.rasterEventFlow == null) {
```
jacob314, line 459: Also document why these checks are the way they are. I would have expected the check to be that
`uiEventFlow != null` rather than checking that `rasterEventFlow == null` as the current way would lead to a NPE if both were null for a frame but perhaps that isn't possible.

packages/devtools_app/lib/src/performance/performance_controller.dart
kenzieschmoll: turns out we weren't even using the _timelineEvents notifier in the UI. removed it completely.

packages/devtools_app/lib/src/performance/performance_controller.dart
kenzieschmoll: The timeline events that occur before the first frame will still be present in the timeline flame chart. This check is to prevent us from trying to process new events for frames that we know occurred before the first frame-related timeline events that we have data for. This code is only triggered when a frame is selected, so it is not responsible for the initial load of timeline events.

kenzieschmoll (commented):
addressed review comments - working on adding tests

jacob314:
lgtm

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -237,10 +224,33 @@ class PerformanceController extends DisposableController
         : PerformanceData();
   }
 
-  FutureOr<void> processAvailableEvents() async {
-    if (data == null) {
-      await _initData();
+  Future<void> _pullTraceEventsFromVmTimeline({
+    bool shouldPrimeThreadIds = false,
+  }) async {
+    final currentVmTime = await serviceManager.service.getVMTimelineMicros();
```
jacob314, line 230: need to make sure we don't have multiple simultaneous timeline event pulls at once for the case a pull took more than 2 seconds.
Use the RateLimiter class and perhaps a delay of only 200 ms to achieve that goal.
As you are now guaranteed to only have one pull at a time it would be safe to pull more frequently.
https://github.com/flutter/devtools/blob/554a0cbbda362785cab18172a3457f5de1fbbc38/packages/devtools_app/lib/src/utils.dart#L452

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -237,10 +224,33 @@ class PerformanceController extends DisposableController
...
+
+    // TODO(kenz): move this priming logic into the loop below.
+    if (shouldPrimeThreadIds) primeThreadIds(timeline);
+    for (final event in timeline.traceEvents) {
+      final eventWrapper = TraceEventWrapper(
+        TraceEvent(event.json),
+        DateTime.now().millisecondsSinceEpoch,
+      );
+      allTraceEvents.add(eventWrapper);
+      debugTraceEventLog(eventWrapper.event.json.toString());
```
jacob314, line 249: Can you remove this debugTraceEventLog or pass the raw json object instead of calling toString? This code will call the toString on the JSON even in release mode which could be somewhat expensive.

jacob314 (commented):
lgtm
once case where fetching polls could take more than 2 secs is handled.

packages/devtools_app/lib/src/performance/performance_controller.dart
kenzieschmoll: changing this method to be a callback so that the toString() won't get called unless the callback is called

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -208,15 +210,17 @@ class PerformanceController extends DisposableController
       await processTraceEvents(allTraceEvents);
       _processing.value = false;
 
+      _timelinePollingRateLimiter = RateLimiter(
+        timelinePollingRateLimit,
+        () async => await _pullTraceEventsFromVmTimeline(),
```
jacob314, line 215: replace:
`() async => await _pullTraceEventsFromVmTimeline()`
with
`_pullTraceEventFromVmTimeline`

@pq is there a lint tracking this case?

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -208,15 +210,17 @@ class PerformanceController extends DisposableController
...
+      );
+
       // Poll for new timeline events.
       // We are polling here instead of listening to the timeline event stream
       // because the event stream is sending out of order and duplicate events.
       // See https://github.com/dart-lang/sdk/issues/46605.
       _pollingTimer = Timer.periodic(const Duration(seconds: 2), (_) async {
```
jacob314, line 222: any reason not to tweak this to 1 second or 500ms?

kenzieschmoll: done

packages/devtools_app/lib/src/performance/performance_controller.dart
kenzieschmoll: done

pq: Hrmmm, closest thing would be `unnecessary_lambdas`.

Issue #3172, opened 2021-06-29
Move code out of devtools_testing and delete package


kenzieschmoll:
Original discussion here: https://github.com/flutter/devtools/pull/3049#discussion_r638056933. 

devtools_testing is an additional layer of abstraction that was never used as intended. If we need files to be shared, we can move them to package:devtools_shared, as that is becoming a place for more shared code anyway.

PR #3173, opened 2021-06-29
Move performance tests out of package:devtools_testing

kenzieschmoll:
No functional changes - just a refactor. Progress towards https://github.com/flutter/devtools/issues/3172


jacob314 (approved):
lgtm
Perhaps we should remove package:devtools_testing.
@grouma, @jakemac53 are we taking advantage of it in dwds? I think the plan was to use it to provide some devtools test coverage upstream but I don't think that happened so devtools_testing is just making it a little harder for us to see what tests we have in devtools.

grouma:
Nope. I don't think we use it.

jakemac53:
I don't have any problem with removing it if it's not being used :)

PR #3185, opened 2021-07-06
ignore use of a deprecated parameter

devoncarew:
- ignore use of a deprecated parameter

This will address a CI failure in https://github.com/flutter/devtools/pull/3161.


devoncarew:
I'm going to TBR this since it looks like there are 1-2 other failures to address.

PR #3188, opened 2021-07-08
Update release script.

jacob314:
TBR. Rev the canvaskit verison to match Flutter.


helin24:
Looks fine, though I don't understand the context.

PR #3193, opened 2021-07-09
Allow embedding unless specified

helin24:
We want DevTools to default to allow embedding so that it can be shown in the VSCode iframe consistently.

CC @DanTup


jacob314 (approved):
lgtm

PR #3197, opened 2021-07-13
experimenting with building against a specific flutter sdk

devoncarew:
experimenting with building against a specific flutter sdk:
- remove some older (`TRAVIS_OS_NAME`) build file cruft
- build and test against a specific flutter version (currently the most recent dev - `2.4.0-4.0.pre`)
- add a CI time check to validate that we're testing that `flutter-version.txt` is the same version that the CI is testing


devoncarew:
After this, we might want to follow up with one mor more of:
- remove testing against `dev`, `beta`, and `stable` on this CI?
- have a script to automatically populate `flutter-sdk/` with the flutter sdk version from `flutter-version.txt`?
- have the `tool/build_release.sh` script ensure that the flutter sdk in `flutter-sdk/` is the same as that mentioned in flutter-version.txt?

cc @kenzieschmoll @jacob314

.github/workflows/build.yaml
```diff
@@ -23,11 +22,11 @@ jobs:
       fail-fast: false
       matrix:
         channel:
-          - master
+          # When updating this version, also update the version in flutter-version.txt.
+          - 2.4.0-4.0.pre
```
kenzieschmoll, line 26: can you add a comment describing how often we need to update this and which version we should be selecting when we do?

devoncarew: done!

kenzieschmoll (approved):

PR #3202, opened 2021-07-15
Only run the formatter on the bots for the master channel and add missing VmService method

kenzieschmoll:
We are hitting an issue currently where the formatter has been changed between the flutter dev and flutter beta branch. We should only run the formatter for one branch so that we do not hit this issue on bot runs. I selected master since most of us develop on tip of trunk flutter.


jacob314 (approved):
lgtm

Issue #3205, opened 2021-07-19
Live timeline polish


kenzieschmoll:
- [ ] Add text to toast describing situation where no timeline events are available for flutter frame because event for that frame have fallen out of the buffer
- [ ] When selecting a very recent frame fails to find timeline events for that frame, try to pull timeline events again after ~1 second

PR #3209, opened 2021-07-20
Polish to the live timeline

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/3205

New toast when we will never have timeline events for a given frame:
Screen Shot 2021-07-21 at 8 23 52 AM

### DIFF
--- a/packages/devtools_app/lib/src/performance/performance_controller.dart
+++ b/packages/devtools_app/lib/src/performance/performance_controller.dart
@@ -317,10 +317,11 @@ class PerformanceController extends DisposableController
       return;
     }
 
-    if (!frame.isWellFormed &&
-        (firstWellFormedFrameMicros == null ||
-            frame.timeFromFrameTiming.start.inMicroseconds >=
-                firstWellFormedFrameMicros)) {
+    final bool frameBeforeFirstWellFormedFrame =
+        firstWellFormedFrameMicros == null ||
+            frame.timeFromFrameTiming.start.inMicroseconds <
+                firstWellFormedFrameMicros;
+    if (!frame.isWellFormed && !frameBeforeFirstWellFormedFrame) {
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -315,16 +317,28 @@ class PerformanceController extends DisposableController
...
       // well formed frame. Timeline events that occurred before this frame will
       // have already fallen out of the buffer.
       await processAvailableEvents();
     }
 
+    // If the frame is still not well formed after processing all available
+    // events, wait a short delay and try to process events again after the
+    // VM has been polled one more time.
+    if (!frame.isWellFormed && !frameBeforeFirstWellFormedFrame) {
+      _processing.value = true;
```
jacob314, line 335: we should ensure we are not already processing otherwise we might kick off multiple calls to processTraceEvents. Maybe that is fine but seems strange.

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -315,16 +317,28 @@ class PerformanceController extends DisposableController
...
+      await Future.delayed(timelinePollingInterval, () async {
+        await processTraceEvents(allTraceEvents);
+        _processing.value = false;
+      });
+    }
+
     data.selectedFrame = frame;
```
jacob314, line 342: what if someone clicks another frame while awaiting this async operation? Add a 
_nextSelectedFrame member and skip if frame != _nextSelectedFrame.

kenzieschmoll: added `_currentFrameBeingSelected` member and a check after each await. PTAL.

jacob314 (approved):
lgtm

Issue #3195, opened 2021-07-12
Inspector widget tree design improvements proposal


CodemateTechAgency:
Hello again! This is our proposal for our second goal: making it easier to find widgets around the Flutter Inspector.

## Background

The widget tree is currently a fairly plain list of widgets within an application, with little visual help for developers to find which part of the widget tree matches which part of their app.

This is how it currently looks:

<img width="269" alt="Screenshot 2021-07-12 at 23 05 22" src="https://user-images.githubusercontent.com/82218649/125349011-a424a880-e365-11eb-8ff8-a5d543590b05.png">

Our aim is to tweak the design to assist users with identifying widgets in the tree more easily, including when scanning through the tree to find a widget, and to better orient themselves when looking at a portion of the tree.

## Design tweaks

We have two design proposals for helping make widgets easier to identify: just changing colors and icon shapes, or adding icons to various common widgets. 

We’d appreciate your feedback as to whether you think the icon version is the way to go; or if it's too visually confusing. 

### Proposed design version 1: just adding icon color coding

This version just categories icons based on color and icon shape, keeping the current initial letter from the current widget design:

<img width="323" alt="Screenshot 2021-07-12 at 23 09 33" src="https://user-images.githubusercontent.com/82218649/125349429-3cbb2880-e366-11eb-8f6c-3daccb18126c.png">

### Proposed design version 2: adding icons

This version adds custom icons to various widgets. Our concern with adding many different icons is that the tree becomes somewhat cluttered and visually confusing. We feel that highlighting container widgets such as containers, columns, rows and alignment is most useful for helping users orient themselves:

<img width="300" alt="Screenshot 2021-07-12 at 23 23 22" src="https://user-images.githubusercontent.com/82218649/125350997-362db080-e368-11eb-957e-cdb6ae969274.png">

### Color coding

These are our current categories of color coding for both options:

* **Grey** for high-level widgets (Root, MaterialApp, etc)
* **Green** to highlight basic widgets (Icons, Text, Images)
* **Orange** for animations, transitions and transforms (to highlight widgets that are moving/rotated etc)
* **Blue** for all other widgets (a brighter color to help improve contrast)

### Icon alignment 

We also suggest tweaking the alignments of the tree so that icons are directly aligned with the expand/collapse arrows, see this before and after:

[...]

PR #3218, opened 2021-07-23
Add text previews to widget tree

tomgilder:
As discussed in #3195.

For now this only brings the text preview for `RenderParagraph` objects, such as `Text` and `RichText`:

Screenshot 2021-07-23 at 22 15 48

We probably want to support other widgets such as `SelectableText`, but those will require special cases.

Am in the process of adding tests hence the draft, but I wanted to check you're happy with this implementation of adding an extra polyfill script @jacob314 @kenzieschmoll?

### DIFF
--- a/packages/devtools_app/lib/src/inspector/diagnostics.dart
+++ b/packages/devtools_app/lib/src/inspector/diagnostics.dart
@@ -89,8 +89,9 @@ class DiagnosticsNodeDescription extends StatelessWidget {
 
     final textPreview = diagnostic.json['textPreview'];
     if (textPreview is String) {
+      final preview = textPreview.replaceAll('\n', ' ');
       yield TextSpan(
-        text: ': "$textPreview"',
+        text: ': "$preview"',
         style: textStyle.merge(inspector_text_styles.unimportant(colorScheme)),
       );
     }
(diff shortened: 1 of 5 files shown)


kenzieschmoll (commented):
This approach LGTM 👍

packages/devtools_app/lib/src/inspector/diagnostics.dart
```diff
@@ -82,9 +82,18 @@ class DiagnosticsNodeDescription extends StatelessWidget {
         return;
       }
     }
+
     if (description?.isNotEmpty == true) {
       yield TextSpan(text: description, style: textStyle);
     }
+
+    final textPreview = diagnostic.json['textPreview'];
+    if (textPreview is String) {
```
kenzieschmoll, line 91: should we be checking for `is String` or `!= null` ?

tomgilder: I now need to check it's a string to strip newlines from it

packages/devtools_app/test/support/inspector_tree.dart
```diff
@@ -0,0 +1,63 @@
+import 'package:devtools_app/src/inspector/diagnostics_node.dart';
```
kenzieschmoll, line 1: missing copyright header

kenzieschmoll (approved):

packages/devtools_app/test/support/inspector_tree.dart
tomgilder: Fixed

Issue #3163, opened 2021-06-24
display information about widget rebuild and repaint counts


devoncarew:
We should display information about widget rebuild and repaint counts. The Flutter IntelliJ plugin has this capability - we should add this to DevTools as well. The likely steps are:

- expose a toggle (on the Performance page?) to enable widget rebuild counts
- collect the resulting rebuild/repaint events - one per frame - and accumulate the counts; clear them out whenever we get a reload or restart event
- display the information somewhere - as per frame info? - in the performance page

Once we have a steel thread through in terms of collecting and displaying the information, we should iterate on where and how the best way to display the info is. 

cc @kenzieschmoll @jacob314 @InMatrix

jacob314:
Exposing on the performance page SGTM.
I could see a version of the performance page for Flutter Web that just showed these counts along with the overall frame timing being useful.

Just showing the same table of aggregated stats like in intellij seems reasonable. We do get a fair amount of positive feedback about those stats helping users catch issues where they did something very inefficient. 


Displaying as per frame info would be interesting. We didn't have a clickable per-frame graph in IntelliJ. One catch is we'll need to make sure we can associate the counts with the correct frame. We might need to add a frame id to the event dispatched by package:flutter.

kenzieschmoll:
This is complete

PR #3219, opened 2021-07-23
process stats for widget rebuild events

devoncarew:
- process stats for widget rebuild events
- working towards https://github.com/flutter/devtools/issues/3163

This doesn't have a UI as yet, but we do listen for events and accumulate the stats in a model object.

### DIFF
--- a/packages/devtools_app/lib/src/performance/performance_controller.dart
+++ b/packages/devtools_app/lib/src/performance/performance_controller.dart
@@ -201,17 +201,13 @@ class PerformanceController extends DisposableController
       data?.displayRefreshRate = _displayRefreshRate.value;
 
       // Listen for Flutter.Frame events with frame timing data.
+      // Listen for Flutter.RebuiltWidgets events.
       autoDispose(
           serviceManager.service.onExtensionEventWithHistory.listen((event) {
         if (event.extensionKind == 'Flutter.Frame') {
           final frame = FlutterFrame.parse(event.extensionData.data);
           addFrame(frame);
-        }
-      }));
-
-      // Listen for Flutter.RebuiltWidgets events.
-      autoDispose(serviceManager.service.onExtensionEvent.listen((event) {
-        if (event.extensionKind == 'Flutter.RebuiltWidgets') {
+        } else if (event.extensionKind == 'Flutter.RebuiltWidgets') {
           rebuildCountModel.processRebuildEvent(event.extensionData.data);
         }
       }));
--- a/packages/devtools_app/lib/src/performance/performance_controller.dart
+++ b/packages/devtools_app/lib/src/performance/performance_controller.dart
@@ -26,6 +26,7 @@ import '../utils.dart';
 import 'performance_model.dart';
 import 'performance_screen.dart';
 import 'performance_utils.dart';
+import 'rebuild_counts.dart';
 import 'timeline_event_processor.dart';
 import 'timeline_streams.dart';
 
@@ -154,6 +155,8 @@ class PerformanceController extends DisposableController
   /// frame id in the corresponding [TimelineEvent]s.
   final _unassignedFlutterFrames = <int, FlutterFrame>{};
 
+  final RebuildCountModel rebuildCountModel = RebuildCountModel();
+
   Timer _pollingTimer;
 
   int _nextPollStartMicros = 0;
@@ -165,6 +168,7 @@ class PerformanceController extends DisposableController
   RateLimiter _timelinePollingRateLimiter;
 
   Future<void> _initialized;
+
   Future<void> get initialized => _initialized;
 
   Future<void> _init() {
@@ -202,6 +206,13 @@ class PerformanceController extends DisposableController
         if (event.extensionKind == 'Flutter.Frame') {
           final frame = FlutterFrame.parse(event.extensionData.data);
           addFrame(frame);
+        }
+      }));
+
+      // Listen for Flutter.RebuiltWidgets events.
+      autoDispose(serviceManager.service.onExtensionEvent.listen((event) {
+        if (event.extensionKind == 'Flutter.RebuiltWidgets') {
+          rebuildCountModel.processRebuildEvent(event.extensionData.data);
         }
       }));
 
--- a/packages/devtools_app/lib/src/performance/rebuild_counts.dart
+++ b/packages/devtools_app/lib/src/performance/rebuild_counts.dart
@@ -1,0 +1,136 @@
+// Copyright 2021 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+import 'package:path/path.dart' as path;
+
+// {
+//   "startTime": 2352949,
+//   "events": [1, 1, 2, 1, ...],
+//   "locations": {
+//     "file": {
+//       "ids": [1, 2, ...],
+//       "lines": [23, 32, ...],
+//       "columns": [10, 12, ...],
+//       "names": ["PlanetsApp", "MaterialApp", ...]
+//     }
+//   },
+//   "newLocations": {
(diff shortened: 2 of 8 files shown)


packages/devtools_app/lib/src/performance/rebuild_counts.dart
```diff
@@ -0,0 +1,136 @@
+// Copyright 2021 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+import 'package:path/path.dart' as path;
+
+// {
```
jacob314, line 8: nit add a commend clarifying this is an example of the data format.

packages/devtools_app/lib/src/performance/rebuild_counts.dart
```diff
@@ -0,0 +1,136 @@
...
+    // parse locations
+    if (json.containsKey('locations')) {
+      final fileLocationsMap =
+          (json['locations'] as Map).cast<String, dynamic>();
+
+      for (final String file in fileLocationsMap.keys) {
+        final Map<String, List<dynamic>> entries =
+            (fileLocationsMap[file] as Map).cast<String, List<dynamic>>();
+
+        final List<int> ids = entries['ids'].cast<int>();
```
jacob314, line 42: nit: remove types on LHS that don't add clarity.

packages/devtools_app/lib/src/performance/rebuild_counts.dart
```diff
@@ -0,0 +1,136 @@
...
+//   "newLocations": {
+//     "file": [id1, line1, column1, id2, line2, column2, ...],
+//     }
+//   }
+// }
+
+class RebuildCountModel {
+  final Map<int, RebuildLocation> _locationMap = {};
+
+  final _locations = PropertyValueNotifier<List<RebuildLocation>>([]);
```
jacob314, line 28: Can you use ListValueNotifier?
If there are extra methods you need to add to make ListValueNotifier work, feel free to add them.

packages/devtools_app/lib/src/performance/rebuild_counts.dart
```diff
@@ -0,0 +1,136 @@
...
+  final int id;
+  final String path;
+  final int line;
+  final int column;
+  final String name;
+
+  int buildCount = 0;
+}
+
+class PropertyValueNotifier<T> extends ValueNotifier<T> {
```
jacob314, line 129: use ListValueNotifier instead if possible. It is a little more robust in terms of ensuring that the value surfaced is not identical if it contains different

packages/devtools_app/lib/src/service_extensions.dart
```diff
@@ -361,6 +373,7 @@ final List<ServiceExtensionDescription> _extensionDescriptions = [
   httpEnableTimelineLogging,
   socketProfiling,
   invertOversizedImages,
+  trackRebuildWidgets
```
jacob314, line 376: nit: add trailing comma for better merges.

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -539,6 +540,7 @@ class IsolateManager extends Disposer {
   Stream<IsolateRef> get onIsolateExited => _isolateExitedController.stream;
 
   final _mainIsolate = ValueNotifier<IsolateRef>(null);
```
jacob314, line 542: rather than adding a line break can you move the getter before the private field? That is actually the convention Flutter uses.

packages/devtools_app/test/performance_rebuild_counts_test.dart
```diff
@@ -0,0 +1,76 @@
...
+      expect(getLocation(9).buildCount, 22);
+
+      rebuildCountModel.clearCurrentCounts();
+      expect(getLocation(9).buildCount, 0);
+    });
+  });
+}
+
+const event1 =
+    '{"startTime":20558388,"events":[1,1,2,1,3,1,4,1,6,1,7,2,9,22,10,22,11,22,12,22],"locations":{"file:///Users/devoncarew/projects/devoncarew/planets/lib/main.dart":{"ids":[1,2,3,4,6,7,9,10,11,12],"lines":[23,32,35,85,106,111,132,247,251,258],"columns":[10,12,13,12,18,20,18,12,14,16],"names":["PlanetsApp","MaterialApp","SolarSystemWidget","Scaffold","CustomPaint","ValueListenableBuilder","PlanetWidget","Positioned","GestureDetector","Container"]}}}';
```
jacob314, line 72: nit: file doesn't seem like it was formatted.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -205,6 +209,13 @@ class PerformanceController extends DisposableController
         }
       }));
 
+      // Listen for Flutter.RebuiltWidgets events.
+      autoDispose(serviceManager.service.onExtensionEvent.listen((event) {
+        if (event.extensionKind == 'Flutter.RebuiltWidgets') {
```
kenzieschmoll, line 214: can we move this code into the listener above instead of initializing another listener?

kenzieschmoll (commented):
Overall, looks good. My one concern with this feature is where the best place to put it is. Considering we tell users to use the performance page in profile mode, it seems like mixed messaging to put a debug only feature on the page.

packages/devtools_app/lib/src/performance/rebuild_counts.dart
```diff
@@ -0,0 +1,136 @@
...
+//     }
+//   },
+//   "newLocations": {
+//     "file": [id1, line1, column1, id2, line2, column2, ...],
+//     }
+//   }
+// }
+
+class RebuildCountModel {
+  final Map<int, RebuildLocation> _locationMap = {};
```
kenzieschmoll, line 26: nit: put type on RHS

packages/devtools_app/lib/src/performance/rebuild_counts.dart
devoncarew: Sure - switched to ListValueNotifier.

devoncarew:
> Overall, looks good. My one concern with this feature is where the best place to put it is. Considering we tell users to use the performance page in profile mode, it seems like mixed messaging to put a debug only feature on the page.

Good point.  I don't know that I have a specific solution, but I think that we should expose it somewhere in the near term even if it moves around later. And generally that - in the long term - we should work towards not recommending against using debug mode.

packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -201,17 +201,13 @@ class PerformanceController extends DisposableController
...
         if (event.extensionKind == 'Flutter.Frame') {
           final frame = FlutterFrame.parse(event.extensionData.data);
           addFrame(frame);
-        }
-      }));
-
-      // Listen for Flutter.RebuiltWidgets events.
-      autoDispose(serviceManager.service.onExtensionEvent.listen((event) {
-        if (event.extensionKind == 'Flutter.RebuiltWidgets') {
+        } else if (event.extensionKind == 'Flutter.RebuiltWidgets') {
```
kenzieschmoll, line 210: not a request but just a comment to think about for the future - it would be great if there is a way to connect rebuild stats with the flutter frame they occurred in.

kenzieschmoll (approved):

packages/devtools_app/lib/src/performance/performance_controller.dart
devoncarew: For sure. We have a 'startTime' info field for the rebuild events, but I don't know if that's reliable enough for us to use in terms of matching the events up with a frame.

We could also change the rebuild event to include a frame number, which would make things much easier for downstream tools. I assume the number would be available to the place where we generate rebuild events, though it might take a bit of framework surgery to expose it.

PR #3220, opened 2021-07-27
Refactor CpuProfileData in preparation for filtering support

kenzieschmoll:


packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
```diff
@@ -16,82 +16,166 @@ import 'cpu_profile_transformer.dart';
...
-        sampleCount: subTraceEvents.length,
+        sampleCount: subSamples.length,
         samplePeriod: superProfile.profileMetaData.samplePeriod,
         stackDepth: superProfile.profileMetaData.stackDepth,
         time: subTimeRange,
       ),
     );
   }
 
+  factory CpuProfileData.fromUserTag(CpuProfileData originalData, String tag) {
```
kenzieschmoll, line 122: the code for this factory constructor was copied from what used to be the `dataForUserTag` method below. There are also some changes for using the new format of CpuProfileData

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
```diff
@@ -16,82 +16,166 @@ import 'cpu_profile_transformer.dart';
...
+      stackFrames[stackFrame.id] = stackFrame;
+    }
+
+    // Initialize all CPU samples.
+    final stackTraceEvents =
+        (json[traceEventsKey] ?? []).cast<Map<String, dynamic>>();
+    final samples = stackTraceEvents
+        .map((trace) => CpuSample.parse(trace))
+        .toList()
+        .cast<CpuSample>();
```
jacob314, line 74: nit: why is this cast() required?

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
```diff
@@ -203,10 +233,19 @@ class CpuProfileData {
...
           timeExtentKey: profileMetaData.time.duration.inMicroseconds,
         stackFramesKey: stackFramesJson,
-        traceEventsKey: stackTraceEvents,
+        traceEventsKey: cpuSamples.map((sample) => sample.json).toList(),
       };
 
   bool get isEmpty => profileMetaData.sampleCount == 0;
+
+  @visibleForTesting
+  Map<String, dynamic> get stackFramesJson {
```
jacob314, line 242: nit: prefer Map<String, Object> so we get more static checking.

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
```diff
@@ -379,6 +446,15 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
     }
   }
 
+  Map<String, dynamic> get json => {
```
jacob314, line 449: nit: rename to toJson() for consistency and because this generates a new object each time.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
kenzieschmoll: without the cast, we get the error `type 'List<dynamic>' is not a subtype of type 'List<CpuSample>'`

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
kenzieschmoll: done

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart

Issue #3075, opened 2021-06-02
Improve debugger file search UX


grouma:
From an internal user:

The current fuzzy search experience has some usability issues.

1. The expand if < 5 results often times hides useful information due to repeated matches from generated files
2. Folders do not stay expanded on new search

We should try to take better advantage of the space and save the user a click when possible,

ex1) always show the top X results from each folder expanded with show more... as the final options.
ex2) dont seperate by folder

kenzieschmoll:
After an offline discussion, here are some improvements we will make to the libraries pane, search experience, and history picker
- [ ] remove filter text field from libraries view
- [ ] (internally) add edited files to the top of the libraries view with a subheader and a separator between those files and the rest
- [ ] convert the libraries view to be a tree file explorer
- [ ] search will become a hotkey (cmd + p or cmd + o) that launches a modal dialog where a query can be entered and search results will be populated for the user to select
- [ ] remove duplicate entries in history picker
- [ ] color edited files a different color (maybe blue like IntelliJ) in the libraries pane and history picker

kenzieschmoll:
The remaining work for this is tracked at https://github.com/flutter/devtools/issues/3383 and https://github.com/flutter/devtools/issues/3418. Closing.

PR #3227, opened 2021-07-29
Replace `ScriptPicker` with `ProgramExplorer` on `DebuggerScreen`

bkonyi:
This change replaces the current `ScriptPicker` implementation with a more fully featured `ProgramExplorer`.

**Functionality:**

- Navigate the program structure, including libraries, scripts, classes, functions, and fields
- Display generic parameter information for generic classes and functions
- View function and field signatures in the program explorer
- Color-coded icons for each object type to match existing syntax highlighting
- Jump to definition of classes, functions, and fields
- Lazy population of class nodes
- Filtering based on object names

**Additional Changes:**

- Updated `CodeView` to allow for scrolling to a location aligned to the top of the view
- Fixed issue where `CodeView` background was transparent
- `TreeView` nodes are now highlighted when selected
- Added support for different behaviors for `TreeView` selection vs expansion

**Future Work:**

- Display code objects in VM developer mode
- Allow for inspecting object information in an "object inspector" window in the debugger page


bkonyi:
@kenzieschmoll @jacob314 I think this is ready for an initial pass. PTAL when you have a chance.

kenzieschmoll:
Can you add a screenshot of the new program explorer to the PR description?

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -78,6 +80,13 @@ class _CodeViewState extends State<CodeView>
     gutterController = verticalController.addAndGet();
     textController = verticalController.addAndGet();
 
+    if (widget.initialPosition != null) {
```
kenzieschmoll, line 83: This likely also needs to be in didUpdateWidget so that this state object can manage the scroll position if the widget was updated with a new value for `initialPosition`

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -78,6 +80,13 @@ class _CodeViewState extends State<CodeView>
...
+      final location = widget.initialPosition.location;
+      final lineIndex = location.line - 1;
```
kenzieschmoll, line 85: can you add a comment as to why we are subtracting 1

packages/devtools_app/lib/src/debugger/controls.dart
```diff
@@ -124,7 +124,7 @@ class _DebuggingControlsState extends State<DebuggingControls>
           child: Container(
             color: visible ? Theme.of(context).highlightColor : null,
             child: DebuggerButton(
-              title: 'Libraries',
+              title: 'Program Explorer',
```
kenzieschmoll, line 127: I'm not completely sold on this name. Maybe "Source Explorer"? Input here @jacob314 @devoncarew?

packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -278,6 +285,9 @@ class DebuggerController extends DisposableController
   /// Return the sorted list of ScriptRefs active in the current isolate.
   ValueListenable<List<ScriptRef>> get sortedScripts => _sortedScripts;
 
+  bool get centerScrollLocation => _centerScrollLocation;
```
kenzieschmoll, line 288: is it possible that this value could get accessed for a script it is not associated with? Storing this at the controller level vs the per call level seems problematic

packages/devtools_app/lib/src/debugger/debugger_model.dart
```diff
@@ -94,10 +94,24 @@ class ScriptLocation {
   String toString() => '${scriptRef.uri} $location';
 }
 
-/// A line, column, and an optional tokenPos.
 class SourcePosition {
-  SourcePosition({@required this.line, @required this.column, this.tokenPos});
+  const SourcePosition(
+      {@required this.line, @required this.column, this.file, this.tokenPos});
```
kenzieschmoll, line 99: nit: trailing comma

packages/devtools_app/lib/src/debugger/debugger_screen.dart
```diff
@@ -79,6 +80,7 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
   static const breakpointsTitle = 'Breakpoints';
 
   DebuggerController controller;
+  //ObjectTreeController objectSelectorController = ObjectTreeController();
```
kenzieschmoll, line 83: delete comment?

packages/devtools_app/lib/src/theme.dart
```diff
@@ -399,6 +399,11 @@ extension ThemeDataExtension on ThemeData {
   TextStyle get subtleFixedFontStyle =>
       fixedFontStyle.copyWith(color: unselectedWidgetColor);
 
+  TextStyle get toolTipFixedFontStyle => //colorScheme.isLight ?
```
kenzieschmoll, line 402: remove comment

packages/devtools_app/lib/src/history_viewport.dart
```diff
@@ -59,17 +63,40 @@ class HistoryViewport<T> extends StatelessWidget {
...
+            ToolbarAction(
+              icon: Icons.chevron_right,
+              onPressed: history.hasNext
+                  ? () {
+                      final current = history.current.value;
+                      history.moveForward();
+                      if (onChange != null) {
+                        onChange(
+                          history.current.value,
+                          current,
```
kenzieschmoll, line 90: why not implement a `peekBack` method to mirror what is done above with `moveBack` and `peekNext`?

kenzieschmoll:
At a high level, is this change something we have discussed in the past? It seems inconsistent with other discussions we've had regarding debugger UI improvements (see https://github.com/flutter/devtools/issues/3075). This is probably something that could use some UX input on (@InMatrix). @devoncarew @jacob314 for input on using a program explorer (classes, functions, fields, etc.) vs a file tree explorer in the debugger (described in linked issue https://github.com/flutter/devtools/issues/3075). 

My concern is making this view more complicated and thus more difficult for users to find what they are looking for. Do users expect to see a file explorer or a program explorer in a debugger? Should we make displaying low level details (classes, functions, etc.) an optional setting?

InMatrix:
@bkonyi Could you post some before-and-after screenshots to help me better understand this change?

bkonyi:
There wasn't an explicit discussion, but I did chat with @jacob314 offline. In general, we need to be able to examine details of fields, functions, and classes to reach feature parity with Observatory (that work isn't included in this PR as the explorer was already a big change) in addition to examining code objects (in VM developer mode). Inspecting code objects will require a code view in addition to an object details view, so the debugger pane seemed like the right place for this rather than creating another top-level tab.

Here's some screenshots:

**Before**:
Screen Shot 2021-08-03 at 7 25 19 PM
Screen Shot 2021-08-03 at 7 25 33 PM

**After:**
Screen Shot 2021-08-03 at 7 22 27 PM
Screen Shot 2021-08-03 at 7 23 14 PM
Screen Shot 2021-08-03 at 7 23 43 PM

bkonyi:
@InMatrix it's not clear from the screenshots but selecting classes, functions, and fields currently jumps to their definition in the script.

kenzieschmoll:
Some things that are confusing to me:
Screen Shot 2021-08-04 at 9 43 18 AM
1. package and file names are repeated (e.g. dart:http & dart:http on the same line item). Perhaps for packages and files, we could densify the line item and just show the package name (as one line, not two), and file names can just have the file name without the whole path (the path could be displayed as a tooltip or something). This would make the UI look more like a tree file explorer at the package / file level.
2. Icons for packages / paths should be more basic (e.g. a folder icon for packages and a file or page icon for a file). For example: 
Screen Shot 2021-08-04 at 9 48 03 AM

Icons for underlying classes, functions, etc should be more descriptive - for example I like the sigma icon for functions.

jacob314:
With regards to #3075, I don't see this in conflict with it. We won't use this view or the script picker for searching but the view will be more useful than the existing ScriptPicker for browsing.

packages/devtools_app/lib/src/debugger/codeview.dart
jacob314: agree we need to do something in didUpdateWidget but we will need to be very careful. You don't want to update the scrollPosition unless widget.initialPosition != oldWidget.initialPosition.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -580,7 +603,7 @@ class _LinesState extends State<Lines> with AutoDisposeMixin {
 
         if (isOutOfViewTop || isOutOfViewBottom) {
           // Scroll this search token to the middle of the view.
-          final targetOffset = math.max(
+          final targetOffset = math.max<double>(
```
jacob314, line 606: why was this type added? seems like it should be inferred.

packages/devtools_app/lib/src/debugger/controls.dart
jacob314: Can we leave it as Libraries for now? We are just showing more information about the libraries.

packages/devtools_app/lib/src/debugger/program_explorer_model.dart
```diff
@@ -0,0 +1,337 @@
...
+  ScriptRef script;
+
+  @override
+  bool isSelected = false;
+
+  /// This exists to allow for O(1) lookup of children when building the tree.
+  final Map<String, VMServiceObjectNode> _childrenAsMap = {};
+
+  @override
+  bool get isExpandable => super.isExpandable || object is ClassRef;
```
jacob314, line 104: what if a class has zero members?

jacob314:
I would suggest aligning the icons with what the structure and project views in IntelliJ or VSCode use.
There are existing icons for files, classes, and methods that users should be somewhat familiar with.

We might also need an option to toggle whether to show the contents of a library by default or not. I am a little worried that showing the contents of a library by default may make the tree harder to browse. One option would be to switch to a master-details style view where we show the list of libraries separate from the "structure" view that shows the contents of the selected library but I don't have a strong opinion on that.
Screen Shot 2021-08-09 at 3 16 33 PM

jacob314:
Should this pull request be closed?

bkonyi:
> Should this pull request be closed?

No, not yet. The reworked change has mostly the same underlying logic, and I'll hopefully have that up today for review. I've just run into a couple more bugs that need to be ironed out.

bkonyi:
I think we're ready to review this change. There's one subtle UI glitch that occurs when you expand/collapse tree nodes quickly (animations sometimes play for carets on nodes that aren't being selected), but it's minor enough (and maybe pre-existing?) so I think we can land this change as-is.

Here's an updated screenshot of the explorer:
image

packages/devtools_app/devtools_app.iml
```diff
@@ -9,6 +9,27 @@
       <excludeFolder url="file://$MODULE_DIR$/.idea" />
       <excludeFolder url="file://$MODULE_DIR$/.pub" />
       <excludeFolder url="file://$MODULE_DIR$/build" />
+      <excludeFolder url="file://$MODULE_DIR$/linux/flutter/ephemeral/.plugin_symlinks/file_selector_linux/.dart_tool" />
```
kenzieschmoll, line 12: why are these being added? Since all of these plugins were added a long time ago, I would think these entries would already be here if they were needed

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -89,6 +91,22 @@ class _CodeViewState extends State<CodeView>
     textController = verticalController.addAndGet();
     horizontalController = ScrollController();
 
+    if (widget.initialPosition != null) {
```
kenzieschmoll, line 94: This block is duplicated directly below. delete one of them.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -147,8 +175,10 @@ class _CodeViewState extends State<CodeView>
     if (parsedScript.lineCount * CodeView.rowHeight > extent) {
       // Scroll to the middle of the screen.
```
kenzieschmoll, line 176: this comment is outdated. Also, can we rename `centerScrollLocation` to `shouldCenterScrollLocation` so that it is more clearly a bool just by reading the code. When I first read this I assumed it would be a double

packages/devtools_app/devtools_app.iml
bkonyi: Accidental commit on my end. Will revert (not sure why these files get updated or aren't part of .gitignore).

packages/devtools_app/lib/src/debugger/debugger_model.dart
```diff
@@ -94,10 +94,28 @@ class ScriptLocation {
...
+  const SourcePosition({
+    @required this.line,
+    @required this.column,
+    this.file,
+    this.tokenPos,
+  });
+
+  factory SourcePosition.calculatePosition(Script script, int tokenPos) {
+    final List<List<int>> table = script.tokenPosTable;
+    if (table == null) {
```
kenzieschmoll, line 107: since we aren't using the `table` var can we just check `if (script.tokenPosTable == null)` directly?

packages/devtools_app/lib/src/framework/framework_core.dart
```diff
@@ -26,6 +27,7 @@ class FrameworkCore {
     setGlobal(MessageBus, MessageBus());
     setGlobal(FrameworkController, FrameworkController());
     setGlobal(SurveyService, SurveyService());
+    setGlobal(ProgramExplorerController, ProgramExplorerController());
```
kenzieschmoll, line 30: why is this a global instead of being part of the debugger controller?

packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -92,6 +93,7 @@ class _FileSearchFieldState extends State<FileSearchField>
   void _onSelection(String scriptUri) {
     final scriptRef = _scriptsCache[scriptUri];
     widget.controller.showScriptLocation(ScriptLocation(scriptRef));
+    programExplorerController.selectScriptNode(scriptRef);
```
kenzieschmoll, line 96: do we always want to call `programExplorerController.selectScriptNode` when `showScriptLocation` is called? if so, maybe we should put `programExplorerController.selectScriptNode` directly in `showScriptLocation`

packages/devtools_app/lib/src/theme.dart
```diff
@@ -419,6 +419,10 @@ extension ThemeDataExtension on ThemeData {
   TextStyle get subtleFixedFontStyle =>
       fixedFontStyle.copyWith(color: unselectedWidgetColor);
 
+  TextStyle get toolTipFixedFontStyle => fixedFontStyle.copyWith(
+        color: colorScheme.isLight ? Colors.white : Colors.black,
```
kenzieschmoll, line 423: use`color: colorScheme.tooltipTextColor` instead

packages/devtools_app/lib/src/trees.dart
```diff
@@ -105,6 +108,28 @@ abstract class TreeNode<T extends TreeNode<T>> {
     child.index = children.length - 1;
   }
 
+  void addChildAtIndex(T child, int index) {
```
kenzieschmoll, line 111: instead of creating a new method for this, let's add an optional parameter `index` to `addChild`. That way we don't have to duplicate the logic of `child.parent = this;` and `child.index = index;`

packages/devtools_app/lib/src/trees.dart
```diff
@@ -85,6 +87,7 @@ abstract class TreeNode<T extends TreeNode<T>> {
   ///
   /// When using this, consider caching the value. It is O([level]) to compute.
   bool shouldShow() {
+    print('${parent == null}');
```
kenzieschmoll, line 90: remove print

packages/devtools_app/lib/src/trees.dart
```diff
@@ -43,6 +43,8 @@ abstract class TreeNode<T extends TreeNode<T>> {
 
   bool get isRoot => parent == null;
 
+  bool get isSelected => false;
```
kenzieschmoll, line 46: Should this getter be on the subclass instead of in the TreeNode base class? seems odd to have a getter return a set bool value when there aren't also methods in this TreeNode class to modify the state of `isSelected`

packages/devtools_app/test/debugger_screen_test.dart
```diff
@@ -26,9 +28,12 @@ void main() {
   DebuggerScreen screen;
   FakeServiceManager fakeServiceManager;
   MockDebuggerController debuggerController;
+  MockProgramExplorerController mockProgramExplorerController;
   fakeServiceManager = FakeServiceManager();
   when(fakeServiceManager.connectedApp.isProfileBuildNow).thenReturn(false);
+  mockProgramExplorerController = MockProgramExplorerController.withDefaults();
   setGlobal(ServiceConnectionManager, fakeServiceManager);
+  setGlobal(ProgramExplorerController, mockProgramExplorerController);
```
kenzieschmoll, line 36: we should initialize variables, call `setGlobal` and set up mocks in a `setUp` method, not directly in main. This will allow each test to start with the same state instead of potentially having left over effects from the previous test

packages/devtools_app/lib/src/debugger/program_explorer.dart
```diff
@@ -0,0 +1,407 @@
...
+                minSizes: const [0.0, 0.0],
+                headers: const <PreferredSizeWidget>[
+                  AreaPaneHeader(
+                    title: Text('File Explorer'),
+                    needsTopBorder: false,
+                  ),
+                  AreaPaneHeader(title: Text('Outline')),
+                ],
+                children: [
+                  ValueListenableBuilder<List<VMServiceObjectNode>>(
```
kenzieschmoll, line 307: This build method is deep. Instead, create 2 stateless widgets `FileExplorer` and `FileOutline` or similar, and use those widgets here.

packages/devtools_app/lib/src/debugger/program_explorer.dart
```diff
@@ -0,0 +1,407 @@
...
+                    valueListenable: controller.rootObjectNodes,
+                    builder: (context, nodes, _) {
+                      return TreeView<VMServiceObjectNode>(
+                        itemExtent: 25,
```
kenzieschmoll, line 311: magic number? and why is this not the same as for the outline view? These also need to be scaled by font factor. You can add a top level getter to this file `double get programExplorerRowHeight => scaleByFontFactor(25.0);` or similar

packages/devtools_app/lib/src/debugger/program_explorer.dart
```diff
@@ -0,0 +1,407 @@
...
+  ProgramExplorer({
+    Key key,
+    @required this.debugController,
+    @required this.onSelected,
+  }) : controller = programExplorerController, super(key: key) {
+    controller.selectScriptNode(debugController.currentScriptRef.value);
+  }
+
+  final ProgramExplorerController controller;
+  final DebuggerController debugController;
```
kenzieschmoll, line 281: make programExplorerController a member of the DebuggerController class, and then we can access it from there without it being a global.

packages/devtools_app/lib/src/debugger/program_explorer.dart
```diff
@@ -0,0 +1,407 @@
...
+
+/// Picker that displays the program's structure, allowing for navigation and
+/// filtering.
+class ProgramExplorer extends StatelessWidget {
+  ProgramExplorer({
+    Key key,
+    @required this.debugController,
+    @required this.onSelected,
+  }) : controller = programExplorerController, super(key: key) {
+    controller.selectScriptNode(debugController.currentScriptRef.value);
```
kenzieschmoll, line 277: this should be handled in the controller, not in the widget constructor. The ProgramExplorer widget should not need to do this on initialization, but rather just read data from a value notifier in the controller.

packages/devtools_app/lib/src/framework/framework_core.dart
bkonyi: Mostly so it can be mocked out in tests. I originally had it as a member of the debugger controller but ran into issues when we were testing the non-mock DebuggerController.

packages/devtools_app/lib/src/debugger/program_explorer.dart
```diff
@@ -0,0 +1,407 @@
...
+                            onTap: () {
+                              controller.selectNode(node);
+                              onTap();
+                            },
+                          );
+                        },
+                      );
+                    },
+                  ),
+                  // TODO(bkonyi): add outline view.
```
kenzieschmoll, line 328: remove todo?

packages/devtools_app/lib/src/tree.dart
```diff
@@ -106,13 +129,15 @@ class _TreeViewState<T extends TreeNode<T>> extends State<TreeView<T>>
 }
 
 class TreeViewItem<T extends TreeNode<T>> extends StatefulWidget {
-  const TreeViewItem(this.data, {this.buildDisplay, this.onItemPressed});
+  const TreeViewItem(this.data,
+      {this.buildDisplay, this.onItemExpanded, this.onItemSelected});
```
kenzieschmoll, line 133: nit: trailing comma

packages/devtools_app/lib/src/tree.dart
```diff
@@ -124,23 +149,27 @@ class _TreeViewItemState<T extends TreeNode<T>> extends State<TreeViewItem<T>>
...
-                      size: defaultIconSize,
+      child: Container(
+        color:
+            widget.data.isSelected ? Theme.of(context).selectedRowColor : null,
+        child: Row(
+          mainAxisSize: MainAxisSize.min,
+          children: [
+            widget.data.isExpandable
+                ? InkWell(
+                    onTap: _onExpanded ?? _onSelected,
```
kenzieschmoll, line 160: _onExpanded doesn't look like it will ever be null since it is a private method in this class. I would think that we would want to default to _onSelected and have _onSelected call widget.onItemExpanded if the node is expandable

packages/devtools_app/lib/src/tree.dart
```diff
@@ -158,8 +187,16 @@ class _TreeViewItemState<T extends TreeNode<T>> extends State<TreeViewItem<T>>
...
+  void _onExpanded() {
+    if (widget.onItemExpanded == null) {
+      return;
+    }
+    widget.onItemExpanded(widget.data);
+    setExpanded(widget.data.isExpanded);
+  }
+
+  void _onSelected() {
+    widget.onItemSelected(widget.data);
```
kenzieschmoll, line 199: should we be checking that onItemSelected is not null here?

packages/devtools_app/lib/src/tree.dart
```diff
@@ -177,6 +214,9 @@ mixin TreeMixin<T extends TreeNode<T>> {
     for (T root in roots) {
       traverse(root, (n) {
         if (onTraverse != null) onTraverse(n);
+        if (!n.shouldShow()) {
+          return false;
+        }
```
kenzieschmoll, line 219: can you add a test for a case that would have been incorrect without this check?

packages/devtools_app/lib/src/debugger/program_explorer_model.dart
```diff
@@ -0,0 +1,512 @@
...
+    this.fields,
+  });
+
+  final Library lib;
+  final List<ClassRef> classes;
+  final List<FuncRef> functions;
+  final List<FieldRef> fields;
+
+  static Future<VMServiceLibraryContents> getLibraryContents(
+      LibraryRef libRef) async {
```
kenzieschmoll, line 26: nit: tc

packages/devtools_app/lib/src/debugger/program_explorer_model.dart
```diff
@@ -0,0 +1,512 @@
...
+
+    // If this is a top-level node and a library is specified, this must be a
+    // library node.
+    if (parts.isEmpty && lib != null) {
+      node.object = lib;
+    }
+    return node;
+  }
+
+  VMServiceObjectNode _getCreateChild(
```
kenzieschmoll, line 308: nit: rename to lookupChild

packages/devtools_app/lib/src/debugger/program_explorer_model.dart
```diff
@@ -0,0 +1,512 @@
...
+    ObjRef object, {
+    bool isSelectable = true,
+  }) {
+    final child = VMServiceObjectNode(
+      controller,
+      name,
+      object,
+      isSelectable: isSelectable,
+    );
+    child.parent = this;
```
kenzieschmoll, line 330: use `addChild` from the superclass instead of duplicating that code here

packages/devtools_app/lib/src/debugger/program_explorer_model.dart
```diff
@@ -0,0 +1,512 @@
...
+    packageAndCoreLibDirectoryNodes.sort((a, b) {
+      return a.name.compareTo(b.name);
+    });
+
+    packageAndCoreLibLibraryNodes.sort((a, b) {
+      return a.name.compareTo(b.name);
+    });
+
+    children.clear();
+    children.addAll([
```
kenzieschmoll, line 477: nit: children..clear()..addAll

kenzieschmoll (commented):
Done with review

packages/devtools_app/lib/src/debugger/codeview.dart
bkonyi: Bad merge resolution I guess. Done.

packages/devtools_app/lib/src/trees.dart
bkonyi: I think this needs to be here since `_TreeViewItemState` uses this to determine whether or not to highlight a given `TreeNode` item.

packages/devtools_app/lib/src/debugger/program_explorer.dart
bkonyi: I guess I was fiddling with these heights at some point during the refactor and they skewed from one another. Anyway, I'll create a top-level private getter for this.

packages/devtools_app/lib/src/debugger/program_explorer.dart
```diff
@@ -0,0 +1,407 @@
...
+      text = toolTip;
+    } else if (node.script != null) {
+      toolTip = node.script.uri;
+    }
+    return Tooltip(
+      waitDuration: tooltipWait,
+      preferBelow: false,
+      message: toolTip ?? node.name,
+      textStyle: theme.toolTipFixedFontStyle,
+      child: Material(
```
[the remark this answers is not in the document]
bkonyi: Very good question... probably from a much earlier iteration. Removed.

packages/devtools_app/lib/src/debugger/program_explorer.dart
```diff
@@ -0,0 +1,407 @@
...
+
+    assert((icon == null && character != null && isShortCharacter != null) ||
+        (icon != null && character == null && isShortCharacter == null));
+
+    return SizedBox(
+      height: defaultIconSize,
+      width: defaultIconSize,
+      child: Container(
+        width: defaultIconSize,
+        height: defaultIconSize,
```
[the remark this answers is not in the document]
bkonyi: We need the `SizedBox` to force the icon to be of a certain size and the `Container` to apply the circle / color decorations. We don't need these `width` and `height` values in the `Container` though.

packages/devtools_app/lib/src/debugger/program_explorer_model.dart
bkonyi: Will rename to `_lookupOrCreateChild` since it returns an existing child when possible, but falls back to creating a new one.

packages/devtools_app/lib/src/tree.dart
bkonyi: Not 100% sure why this was necessary. Doesn't seem to break anything if I remove it, so I'll just do that.

packages/devtools_app/lib/src/debugger/file_search.dart
bkonyi: Good point, that seems to work. Done.

packages/devtools_app/lib/src/trees.dart
kenzieschmoll: if that is the case, then we should move the logic for managing the state of isSelected into this class as well. Subclasses can always override and call super if they need to do additional work on changing selection, in which case we'd want to annotate with @mustCallSuper

packages/devtools_app/lib/src/tree.dart
bkonyi: Ah yeah, this isn't right. However, I do think that we should keep both `_onExpanded` and `_onSelected` as we can have slightly different expansion behaviors depending on whether you tap on the expansion arrow or the row itself. For example, in this PR I've set the tree up so that selecting a node will expand it but won't collapse if it's already expanded, making the expansion arrow the only way to collapse subtrees.

packages/devtools_app/lib/src/tree.dart
bkonyi: Yup. Will add a null-aware operator.

packages/devtools_app/lib/src/debugger/program_explorer.dart
```diff
@@ -266,16 +269,91 @@ class ProgramStructureIcon extends StatelessWidget {
   }
 }
 
+class _FilePicker extends StatelessWidget {
```
kenzieschmoll, line 272: nit: can we rename to _FileExplorer to match the text name for the view?

packages/devtools_app/lib/src/performance/legacy/performance_model.dart
```diff
@@ -681,7 +681,8 @@ abstract class LegacyTimelineEvent extends TreeNode<LegacyTimelineEvent>
   }
 
   @override
-  void addChild(LegacyTimelineEvent child) {
+  void addChild(LegacyTimelineEvent child, [int index]) {
```
kenzieschmoll, line 684: nit: prefer optional named parameters to optional

packages/devtools_app/lib/src/tree.dart
```diff
@@ -188,15 +192,12 @@ class _TreeViewItemState<T extends TreeNode<T>> extends State<TreeViewItem<T>>
...
-    }
-    widget.onItemExpanded(widget.data);
+    widget?.onItemExpanded(widget.data);
     setExpanded(widget.data.isExpanded);
   }
 
   void _onSelected() {
-    widget.onItemSelected(widget.data);
+    widget?.onItemSelected(widget.data);
     setExpanded(widget.data.isExpanded);
```
kenzieschmoll, line 201: should we be calling onItemExpanded here as well?

kenzieschmoll (approved):
a few more comments then LGTM

packages/devtools_app/lib/src/tree.dart
bkonyi: No, since `onItemExpanded` corresponds to clicking the `>`, whereas `onItemSelected` corresponds to clicking on any other part of the list item.

kenzieschmoll (approved):

Issue #2852, opened 2021-04-01
DevTools does not show POST request body


giaur500:
I can show request url, method, headers, but not request body. So I am not able to inspect api calls. There should be also shown requests body, without it network inspector is unusable

kenzieschmoll:
@bkonyi do we have the ability to support request bodies?

bkonyi:
> @bkonyi do we have the ability to support request bodies?

The HTTP profiler does record request bodies, but I don't think we surface them anywhere in DevTools at this point.

giaur500:
Any plans to add it?

Anjula:
bump.

giaur500:
Please, anybody? Without ability to see request, how even network monitor can be considered usable? In my case, it's useless, as I need to use other tools to capture network traffic and I can see all traffic there.

PR #3233, opened 2021-08-02
Show HTTP POST request body

bleroux:
Basic implementaion for https://github.com/flutter/devtools/issues/2852

Had a new Tab in Network page to display request body (only for POST requests) : 

image

I need some guidance to add a test for this. Existing test (test/network_controller_test.dart) set dartIOVersion to 1.2. To test this PR dartIOVersion 1.6 is required. Is it better to create a new test file (what name ?) or to upgrade the existing one to rely on dartIOVersion 1.6 ?


packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -273,6 +277,11 @@ class TimelineHttpRequestData extends HttpRequestData {
   final TraceEvent _startEvent;
   TraceEvent _endEvent;
 
+  /// For timeline event, HTTP request/response body logging is disabled.
```
kenzieschmoll, line 280: will you add this same comment to responseBody below since it also applies to responseBody? Also, you can set equal to null there as well.

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -620,6 +629,25 @@ class DartIOHttpRequestData extends HttpRequestData {
...
 
+  @override
+  String get requestBody {
+    if (_request is! HttpProfileRequest) {
+      return null;
+    }
+    final fullRequest = _request as HttpProfileRequest;
+    try {
+      if (!_request.isResponseComplete) return null;
+      if (_request.method != 'POST') return null;
```
kenzieschmoll, line 640: should we also be supporting other methods ('CONNECT', 'DELETE', GET', HEAD', PATCH', 'PUT', etc.) @bkonyi

kenzieschmoll (commented):
Generally looks good! Let's wait for @bkonyi 's input here as well. As for adding a test, I would bump the version to 1.6 just in the new test you are writing, and then you'll need to add an implementation for `getHttpProfile` to the `FakeVmService` class. You'll need to add some fake data for the fake vm service to return for this test method. You can add that to test/support/network_test_data.dart

packages/devtools_app/lib/src/http/http_request_data.dart
bleroux: I wanted to do so :-) , but responseBody is initialised in TimelineHttpRequest._ constructor. and this constructor is called from 
TimelineHttpRequestData.fromTimeline factory.

test/support/network_test_data.dart calls this factory to initialise fake data for several tests. And some of those tests check that responseBody is not null.

TimelineHttpRequest seems to be obsolete since dartIOVersion 1.6, but does it make sense to update those fake data in tests ? Is it supported to use new DevTools build with old Dart SDK ? If yes we should keep those tests and add new ones.

packages/devtools_app/lib/src/http/http_request_data.dart
bkonyi: I think we should support `PATCH`, `PUT`, and `POST` since those are the request types that have bodies (that I know of, anyway).

packages/devtools_app/lib/src/http/http_request_data.dart
kenzieschmoll: Ah I see. Can you file an issue to cleanup the TimelineHttp code to default to the post dartIOVersion 1.6 world? If you can link to these comments and assign to me, that would be great.

bleroux:
@googlebot I signed it!

packages/devtools_app/test/network_controller_test.dart
```diff
@@ -279,4 +279,156 @@ void main() {
...
+        service: FakeServiceManager.createFakeService(
+          socketProfile: socketProfile,
+          httpProfile: httpProfile,
+        ),
+      );
+      setGlobal(ServiceConnectionManager, fakeServiceManager);
+      // Enables getHttpProfile support.
+      final fakeVmService = fakeServiceManager.service as FakeVmService;
+      fakeVmService.dartIoVersion = SemanticVersion(major: 1, minor: 6);
+      //fakeVmService.httpE
```
kenzieschmoll, line 302: remove this commented out line

packages/devtools_app/test/network_controller_test.dart
```diff
@@ -279,4 +279,156 @@ void main() {
...
+          .cast<HttpRequestData>()
+          .where((request) => request.method == 'PATCH')
+          .toList();
+
+      expect(httpRequests.length, 1);
+      expect(httpRequests.first.responseBody, isNotEmpty);
+      expect(httpRequests.first.requestBody, isNotEmpty);
+      controller.stopRecording();
+    });
+
```
kenzieschmoll, line 431: remove extra new lines

packages/devtools_app/test/support/network_test_data.dart
```diff
@@ -465,3 +465,292 @@ final Map<String, dynamic> testSocket3Json = {
...
+  ...[111, 111, 39, 44, 10, 32, 32, 32, 32, 32, 32, 98, 111, 100, 121, 58, 32, 39, 98, 97, 114],
+  ...[39, 44, 10, 32, 32, 32, 32, 32, 32, 117, 115, 101, 114, 73, 100, 58, 32, 49, 44, 10, 32],
+  ...[32, 32, 32, 125, 10, 32, 32, 32, 32],
+];
+final httpPatchResponseBodyData = [  
+  ...[32, 32, 32, 32, 123, 10, 32, 32, 32, 32, 32, 32, 116, 105, 116, 108, 101, 58, 32, 39, 102],
+  ...[111, 111, 39, 44, 10, 32, 32, 32, 32, 32, 32, 98, 111, 100, 121, 58, 32, 39, 98, 97, 114],
+  ...[39, 44, 10, 32, 32, 32, 32, 32, 32, 117, 115, 101, 114, 73, 100, 58, 32, 49, 44, 10, 32],
+  ...[32, 32, 32, 125, 10, 32, 32, 32, 32],
+];
```
kenzieschmoll, line 756: new line at eof

packages/devtools_app/test/support/network_test_data.dart
```diff
@@ -465,3 +465,292 @@ final Map<String, dynamic> testSocket3Json = {
...
+    'reasonPhrase': 'OK',
+    'redirects': [],
+    'statusCode': 200,
+    'endTime': 6327091628,
+  },
+  'requestBody': [],
+  'responseBody': httpGetResponseBodyData,
+};
+
+
```
kenzieschmoll, line 531: remove extra new lines

packages/devtools_app/test/support/network_test_data.dart
```diff
@@ -465,3 +465,292 @@ final Map<String, dynamic> testSocket3Json = {
...
+  'responseBody': httpPutResponseBodyData,
+};
+final httpPutRequestBodyData = [
+  ...[32, 32, 32, 32, 123, 10, 32, 32, 32, 32, 32, 32, 116, 105, 116, 108, 101, 58, 32, 39, 102],
+  ...[111, 111, 39, 44, 10, 32, 32, 32, 32, 32, 32, 98, 111, 100, 121, 58, 32, 39, 98, 97, 114],
+  ...[39, 44, 10, 32, 32, 32, 32, 32, 32, 117, 115, 101, 114, 73, 100, 58, 32, 49, 44, 10, 32],
+  ...[32, 32, 32, 125, 10, 32, 32, 32, 32],
+];
+final httpPutResponseBodyData = [123, 10, 32, 32, 34, 105, 100, 34, 58, 32, 49, 48, 49, 10, 125];
+
```
kenzieschmoll, line 677: remove all but one new line here as well

bleroux: I'm so used to automatic formatting :-)

kenzieschmoll (approved):
LGTM thanks for the contribution!

PR #3234, opened 2021-08-02
Initial Diff work.

terrylucas:
work towards https://github.com/flutter/devtools/issues/3949

### DIFF
--- a/packages/devtools_app/lib/src/memory/memory_controller.dart
+++ b/packages/devtools_app/lib/src/memory/memory_controller.dart
@@ -21,6 +21,9 @@ import '../shared/table_data.dart';
 import '../shared/utils.dart';
 import '../shared/version.dart';
 import '../ui/search.dart';
+//import '../utils.dart';
+//import '../version.dart';
+import 'memory_diffing.dart';
 import 'memory_filter.dart';
 import 'memory_graph_model.dart';
 import 'memory_protocol.dart';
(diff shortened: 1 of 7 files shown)


terrylucas:
No ready for review.

Hixie:
@terrylucas Should this PR be closed?

packages/devtools_app/lib/src/memory/memory_diffing.dart
```diff
@@ -0,0 +1,504 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
```
jacob314, line 1: nit: 2022

packages/devtools_app/lib/src/memory/memory_diffing.dart
```diff
@@ -0,0 +1,504 @@
...
+import 'package:devtools_app/src/ui/icons.dart';
+import 'package:flutter/material.dart';
+
+import '../shared/globals.dart';
+import 'memory_controller.dart';
+import 'memory_graph_model.dart';
+import 'memory_snapshot_models.dart';
+
+// Diff Image.
+/*
```
jacob314, line 18: delete commented out code.

packages/devtools_app/lib/src/memory/memory_diffing.dart
```diff
@@ -0,0 +1,504 @@
...
+  int id;
+}
+
+class ListComputation<T> {
+  ListComputation(List<T> previousList, List<T> currentList, this.valueFunction)
+      : deletedOnes = previousList.toList(),
+        newOnes = currentList.toList();
+
+  /// Items newly created (did not exist in previous).
+  final List<T> newOnes;
```
jacob314, line 49: Add a TODO to use sets instead of List. Right now I believe this logic is O(N^2) number of objects in the new and old lists but it can be O(N) if we leverage using sets or maps where the identity hashcode is the key. 
Gotcha is there might be a couple identity hash collisions we need to handle gracefully.

packages/devtools_app/lib/src/memory/memory_controller.dart
```diff
@@ -1259,6 +1304,92 @@ class MemoryController extends DisposableController
...
+
+  static String formatDiffClassInfo(int totalNew, int totalDelete) =>
+      '+$totalNew(-$totalDelete)';
+
+  List<LibraryReference> buildDiffLibrary(SnapshotDiffing diffSnapshot) {
+    final result = <String, LibraryReference>{};
+
+    diffSnapshot.diffComputations.forEach((
+      HeapGraphClassLive key,
+      ListComputation<HeapGraphElementLive> value,
```
jacob314, line 1349: @terrylucas. I'm having a bit of trouble understanding the diff computations. Is there a reason we need to have multiple diff computations vs having a single ahrd coded computation based on the identity hash code and perhaps field values to reduce false positives slightly?

jacob314:
@polina-c @terrylucas what is the plan for this PR?

polina-c:
Thank you, Terry. We used this to push forward the feature.

PR #3235, opened 2021-08-02
Prepare for v2.5.0 release.

terrylucas:


tool/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: devtools_repo
 description: A repo management tool for DevTools.
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
```
kenzieschmoll, line 5: is this change necessary?

kenzieschmoll (approved):

tool/pubspec.yaml
terrylucas: Certainly caused bots to not work.  However, update_version.sh would not run without the sdk change e.g.
```

tool/update_version.dart:50:7: Error: Null safety features are disabled for this library.
Try removing the package language version or setting the language version to 2.12 or higher.
String? incrementVersion(String oldVersion) {
```

Issue #3204, opened 2021-07-16
Support filtering CPU profiles by source paths


kenzieschmoll:
This will allow for things like hiding native code, framework code, etc.

PR #3236, opened 2021-08-02
Support filtering CPU profiles

kenzieschmoll:
Screen Shot 2021-07-29 at 2 19 38 PM
Screen Shot 2021-07-29 at 2 07 08 PM

Fixes https://github.com/flutter/devtools/issues/3204


packages/devtools_app/lib/src/network/network_screen.dart
```diff
@@ -244,13 +244,9 @@ class _NetworkScreenBodyState extends State<NetworkScreenBody>
       context: context,
       builder: (context) => FilterDialog(
         controller: _networkController,
-        onApplyFilter: (query) => _networkController.filterData(
-          QueryFilter.parse(
-            query,
-            _networkController.filterArgs,
-          ),
-        ),
+        onApplyFilter: _networkController.filterData,
```
jacob314, line 247: should we remove onApplyFilter and just let the filterDialoag call 
controller.filterData?

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
```diff
@@ -51,6 +55,35 @@ class CpuProfilerController with SearchControllerMixin<CpuStackFrame> {
...
+  // TODO(kenz): make this lazy once we migrate to null safety.
+  List<ToggleFilter> get toggleFilters => _toggleFilters ??= [
+        ToggleFilter<CpuStackFrame>(
+          name: 'Hide Native code',
+          includeCallback: (stackFrame) => stackFrame.url.isNotEmpty,
+        ),
+        ToggleFilter<CpuStackFrame>(
+          name: 'Hide core Dart libraries',
+          includeCallback: (stackFrame) =>
+              !stackFrame.url.contains('org-dartlang-sdk'),
```
jacob314, line 71: can this be more specific? Would be nice to make this 
url.startsWith instead of contains and would be good to have a link to a bug tracking why these urls show as org-dartlang-sdk instead of `dart:`?

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
```diff
@@ -51,6 +55,35 @@ class CpuProfilerController with SearchControllerMixin<CpuStackFrame> {
...
+        ),
+        if (serviceManager.connectedApp?.isFlutterAppNow ?? true)
+          // TODO(kenz): should we also hide flutter engine frames as part of this
+          // filter? We don't have a url to go off of but we could check for a
+          // stack frame name containing 'flutter::'
+          ToggleFilter<CpuStackFrame>(
+            name: 'Hide core Flutter libraries',
+            includeCallback: (stackFrame) =>
+                !(stackFrame.url.contains('packages/flutter/') ||
```
jacob314, line 80: should be startsWith not contains to avoid false positives. Also add the engine as users really really don't care about the engine.

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
```diff
@@ -212,6 +252,36 @@ class CpuProfilerController with SearchControllerMixin<CpuStackFrame> {
     _processingNotifier.dispose();
     transformer.dispose();
   }
+
+  /// Tracks the identifier for the attempt to filter the data.
+  ///
+  /// We use this to prevent multiple filter calls from processing data at the
+  /// same time.
+  int filterIdentifier = 0;
```
jacob314, line 260: make this private? I don't see it used outside of the class.

packages/devtools_app/lib/src/network/network_screen.dart
kenzieschmoll: done

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
kenzieschmoll: done

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
kenzieschmoll: we can't use starts with because the beginning of the url will change depending on which flutter a user is using. For example, file:///path/to/my/flutter/checkout/packages/flutter...

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
kenzieschmoll: Is this something we should be simplifying on our end (like we do currently for package/flutter urls). For example, 
`org-dartlang-sdk:///third_party/dart/sdk/lib/_internal/vm/lib/compact_hash.dart` -> `dart:sdk/lib/_internal/vm/lib/compact_hash.dart`

WDYT

jacob314: what do we show if you open compact_hash.dart in the debugger page? It would be nice to keep it consistent with that.
I imagine it should ideally be
`dart:core/compact_hash.dart`
although I doubt we surface it that way.

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
jacob314: hmm. what is we pre-process all stackFrame urls to use our best guess at package: and dart: before they get to this point?
that way the check can be `.startsWith('package:flutter/')`

jacob314:
Fyi @terrylucas has some similar filtering code in the memory page to filter objects from libraries that aren't usually useful. Could make sense to make sure the terminology is the same although the libraries to filter out shouldn't be the same as a different set of options make sense.

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
```diff
@@ -172,6 +173,92 @@ class CpuProfileData {
...
+    final filteredCpuSamples = <CpuSample>[];
+    void includeSampleOrWalkUp(
+      CpuSample sample,
+      Map<String, Object> sampleJson,
+      CpuStackFrame stackFrame,
+    ) {
+      if (includeFilter(stackFrame)) {
+        // Modify the name value for this sample trace event to reflect the
+        // leaf most node for the filtered sample.
+        sampleJson[stackFrameIdKey] = stackFrame.id;
```
jacob314, line 189: Can we avoid modifying the json? Rather than having to copy the json it might be clearer and faster to optionally pass in a stackFrameIdOverride to CpuSample.

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
```diff
@@ -172,6 +173,92 @@ class CpuProfileData {
...
+      }
+    }
+
+    for (final sample in originalData.cpuSamples) {
+      final sampleJson = Map<String, Object>.from(sample.json);
+      final leafStackFrame = originalData.stackFrames[sample.leafId];
+      includeSampleOrWalkUp(sample, sampleJson, leafStackFrame);
+    }
+
+    // Use a SplayTreeMap so that map iteration will be in sorted key order.
```
jacob314, line 210: can you comment on why that matters?

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
```diff
@@ -495,6 +582,14 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
 
 @visibleForTesting
 int stackFrameIdCompare(String a, String b) {
+  // Order the root first.
+  if (a == CpuProfileData.rootId) {
```
jacob314, line 586: need to short circuit with
if (a == b) return 0;

kenzieschmoll: done

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
kenzieschmoll: you are right, in the debugger we show `dart:collection-patch/compact_hash.dart`. Switched to doing this and pre-processing urls for stack frames this way.

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
kenzieschmoll: done

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
kenzieschmoll: done

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
kenzieschmoll: How would you suggest filtering out flutter engine code? We don't have a uri for flutter engine stack frames since. Should we check for stack frame names that start with `flutter::`? Related, what about stack frame names that start with `dart::` - what category do those stack frames belong to?

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
kenzieschmoll: we are doing this in DevTools for now and I filed an issue for the VM to send package uris instead of file path uris https://github.com/dart-lang/sdk/issues/46872

packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
kenzieschmoll: done. moved this modification into a toJson getter so that exports of the filtered CPU profile data are accurate

jacob314 (approved):
lgtm

PR #3250, opened 2021-08-04
Delete devtools_testing package

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/3172

This PR moves all the code from devtools_testing into devtools_app/test. This PR also adds flutter-sdk/ to the analysis-options exclude section and reverts the test fixtures provider_app/ and riverpod_app/ into non-null-safe mode (since devtools_app is not yet null safe and this PR moves the test fixtures into the devtools_app package).


devoncarew:
This is still a WIP? Happy to review when ready.

tool/bots.sh
```diff
@@ -95,7 +95,8 @@ if [ "$BOT" = "main" ]; then
     repo_tool repo-check
 
     # Analyze the source.
-    dart analyze --fatal-infos
+    dart analyze --fatal-infos lib
```
devoncarew, line 98: duplicated line

devoncarew: Also, why just analyze the source in `lib/`? Don't we want all the source in devtools_app?

If there are some sub-dirs we want to exclude we can do that w/ some analysis options surgery.

devoncarew (approved):
lgtm w/ one comment on the tool/bots.sh script

tool/bots.sh
kenzieschmoll: good catch, this change accidentally leaked in. reverting.

Issue #3251, opened 2021-08-05
External links don't work when embedded in VSCode


tomgilder:
External links don't launch when DevTools are embedded within VSCode.

This affects a several places, including the warning banners on the performance and memory views, and the new debug toggle tooltips.

The links fail silently - this is because the embedded VSCode Chrome view has the "allow-popups" sandbox flag disabled. Links work correctly in Android Studio.

I have a fix for this that also requires support in DartCode.

PR #3252, opened 2021-08-05
Fix links opening when embedded in VSCode

tomgilder:
Fixes https://github.com/flutter/devtools/issues/3251

Buddy PR on DartCode: https://github.com/Dart-Code/Dart-Code/pull/3517

### DIFF
--- a/packages/devtools_app/lib/src/config_specific/launch_url/_launch_url_web.dart
+++ b/packages/devtools_app/lib/src/config_specific/launch_url/_launch_url_web.dart
@@ -1,0 +1,11 @@
+// Copyright 2021 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be found
+// in the LICENSE file.
+
+import 'dart:html';
+
+final isVSCodeEmbedded = window.navigator.userAgent.contains('Electron');
+
+void launchUrlVSCode(String url) {
+  window.parent?.postMessage({'command': 'launchUrl', 'data': url}, '*');
+}
--- a/packages/devtools_app/lib/src/config_specific/launch_url/_launch_url_web.dart
+++ b/packages/devtools_app/lib/src/config_specific/launch_url/_launch_url_web.dart
@@ -4,8 +4,12 @@
 
 import 'dart:html';
 
-final isVSCodeEmbedded = window.navigator.userAgent.contains('Electron');
-
 void launchUrlVSCode(String url) {
-  window.parent?.postMessage({'command': 'launchUrl', 'data': url}, '*');
+  window.parent?.postMessage(
+    {
+      'command': 'launchUrl',
+      'data': {'url': url}
+    },
+    '*',
+  );
 }
--- a/packages/devtools_app/lib/src/config_specific/launch_url/launch_url.dart
+++ b/packages/devtools_app/lib/src/config_specific/launch_url/launch_url.dart
@@ -10,11 +10,14 @@ import '_launch_url_stub.dart'
     if (dart.library.io) '_launch_url_desktop.dart';
 
 Future<void> launchUrl(String url, BuildContext context) async {
-  if (isVSCodeEmbedded) {
-    launchUrlVSCode(url);
-  } else if (await url_launcher.canLaunch(url)) {
+  if (await url_launcher.canLaunch(url)) {
     await url_launcher.launch(url);
   } else {
     Notifications.of(context).push('Unable to open $url.');
   }
+
+  // When embedded in VSCode, url_launcher will silently fail, so we send a
+  // command to DartCode to launch the URL. This will do nothing when not
+  // embedded in VSCode.
+  launchUrlVSCode(url);
 }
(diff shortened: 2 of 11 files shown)


packages/devtools_app/lib/src/config_specific/launch_url/_launch_url_web.dart
```diff
@@ -0,0 +1,11 @@
+// Copyright 2021 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be found
+// in the LICENSE file.
+
+import 'dart:html';
+
+final isVSCodeEmbedded = window.navigator.userAgent.contains('Electron');
```
DanTup, line 7: It might be nice to have this somewhere more shared and use it here too:

https://github.com/flutter/devtools/blob/4e1ccbec3f95f0cd30fb51754845d95c9034c68b/packages/devtools_app/lib/src/config_specific/framework_initialize/_framework_initialize_web.dart#L60

DanTup (commented):
@jacob314 it feels a little awkward leaking more VS Code details into DevTools, but I can't think of a better way to do this. VS Code's iframes are sandboxes and can't open new windows. VS Code attaches click handlers to normal webviews and does this itself, but since DevTools runs inside another iframe inside that, it can't attach a handler to them (same reason we have to send the key events up too).

One option could be to try to eliminate the webview (this would mean injecting the raw index HTML into the webview and setting a base href so the links are all correct), but I suspect that might not play nice with routing etc.

packages/devtools_app/lib/src/config_specific/launch_url/_launch_url_web.dart
tomgilder: Would it better to inject a param from DartCode, to make sure it's running in VSCode? I guess it's possible there would be some other Electron-based browser running DevTools

DanTup: In the other code (where keystrokes are passed up) this check was to prevent another app hosting DevTools and enabling this (see https://github.com/flutter/devtools/issues/2775#issuecomment-794266068). It's probably less of a concern here, though it might be good to be consistent (I'll defer to @jacob314!).

jacob314: There is also the existing query parameter passed in that indicates whether to use the embedded version of DevTools. That could be reasonable to use in combination with checking for electron to ensure that we only show in VSCode.

packages/devtools_app/lib/src/config_specific/launch_url/launch_url.dart
```diff
@@ -0,0 +1,20 @@
...
+import 'package:flutter/widgets.dart';
+import 'package:url_launcher/url_launcher.dart' as url_launcher;
+import '../../notifications.dart';
+import '_launch_url_stub.dart'
+    if (dart.library.html) '_launch_url_web.dart'
+    if (dart.library.io) '_launch_url_desktop.dart';
+
+Future<void> launchUrl(String url, BuildContext context) async {
+  if (isVSCodeEmbedded) {
+    launchUrlVSCode(url);
```
jacob314, line 14: can we call `launchUrlVSCode(url)` in addition to rather than instead of the regular url launch call?  My impression is that wouldn't be a problem for vscode as popups are blocked anyway. That way if we are wrong about an app actually being VSCode, it won't be a problem as the url will still launch if it can.

jacob314:
> @jacob314 it feels a little awkward leaking more VS Code details into DevTools, but I can't think of a better way to do this. VS Code's iframes are sandboxes and can't open new windows. VS Code attaches click handlers to normal webviews and does this itself, but since DevTools runs inside another iframe inside that, it can't attach a handler to them (same reason we have to send the key events up too).
> 
> One option could be to try to eliminate the webview (this would mean injecting the raw index HTML into the webview and setting a base href so the links are all correct), but I suspect that might not play nice with routing etc.

I agree it is a bit awkward but I think it is simpler than the alternatives. As long as we only trigger the custom behavior for the embedded version of DevTools or if we continue to dispatch the regular browser open request I think there is low risk.

packages/devtools_app/lib/src/config_specific/launch_url/_launch_url_web.dart
```diff
@@ -0,0 +1,11 @@
...
+
+void launchUrlVSCode(String url) {
+  window.parent?.postMessage({'command': 'launchUrl', 'data': url}, '*');
```
DanTup, line 10: To mirror my comment in the Dart-Code repo, I think it's better if this is `'data': { 'url': url }` so that if we need to pass additional data through in future it's not a breaking change (from a string to a map/object).

tomgilder: Done

packages/devtools_app/lib/src/config_specific/launch_url/launch_url.dart
tomgilder: Yep, done. Have tested in VSCode and in browser and this solution works.

DanTup:
@jacob314 this looks good to me and I plan to merge the VS Code part. I'll let you check/merge this if you're otherwise happy.

PR #3254, opened 2021-08-06
move provisioning the canvaskit resources from the build step to the update sdk step

devoncarew:
- move provisioning the canvaskit resources from the build step (`build_release.sh`) to the update sdk step (`update_flutter_sdk.sh`). This will mean that less downloading will have to happen during the build step, and the git repository builds will be more reproducible
- rev to a slightly newer version of the flutter sdk
 
Done a several commits to make reviewing easier.


devoncarew:
Note that this is ~16MB of javascript.

tool/update_flutter_sdk.sh
```diff
@@ -24,3 +24,33 @@ else
...
+# solution would be to either upstream this functionality into the
+# flutter_tools, (https://github.com/flutter/flutter/issues/70101), or to read
+# this from a manifest provided (https://github.com/flutter/flutter/issues/74934).
+
+canvaskit_url=https://unpkg.com/canvaskit-wasm@0.28.1/bin/
+
+flutter_bin=$(which flutter)
+canvaskit_dart_file=$(dirname $flutter_bin)/cache/flutter_web_sdk/lib/_engine/engine/canvaskit/initialization.dart
+if ! grep -q "defaultValue: '$canvaskit_url'" "$canvaskit_dart_file"; then
+  echo "CanvasKit $canvaskit_url does not match local web engine copy. Please update before continuing."
```
kenzieschmoll, line 45: can we add what the local web engine copy version is to this cli output? This has been a pain point when doing the release in the past. Or at least output a command to easily grep and print the correct version use?

kenzieschmoll (approved):
One suggestion then lgtm

tool/update_flutter_sdk.sh
devoncarew: Sure, I'll add some cli output here.

devoncarew:
@hterkelsen, @yjbanov - we're considering committing the canvaskit js resources to the devtools repo to help make builds more hermitic. They're pretty large - ~16MB.

I see two related issues to this - https://github.com/flutter/flutter/issues/70101 and 
https://github.com/flutter/flutter/issues/74934. Are these likely to be addressed in the near term? If no, what could also help us would be to have a package that contained the JS resources which devtools could refer to, instead of bundling the canvaskit resources directly into our repo.

hterkelsen:
FYI the grep in this script will no longer work once https://github.com/flutter/engine/pull/28056 lands

devoncarew:
> FYI the grep in this script will no longer work once [flutter/engine#28056](https://github.com/flutter/engine/pull/28056) lands

Gotcha - looks like we might be able to switch to something based on `const String canvaskitVersion = ...`.

The CIPD resource will be useful - we're going to want to be able to build devtools as part of the dart SDK.

devoncarew:
I'm going to close this; we might look into committing the compiled resources from a release, in which case this wouldn't be necessary. Also, the webkit resources will be available as a CIPD package soon, which would give us better options (instead of committing them here) when we build devtools in the dart sdk.

Issue #2254, opened 2020-08-08
Add jump-to-line shortcut to the debugger page


xu-baolin:
Or there are other ways to quickly position to a specified line(the dart file may have thousands of lines).
window10 + chrome

kenzieschmoll:
This is supported. Closing.

PR #3262, opened 2021-08-11
Add a scrollbar for horizontal scrolling of source files in the debugger

elliette:
### Demo:

working_scrollbar


elliette:
Not sure which test I should modify for this change. I was thinking maybe `/integration_tests/debugger.dart`, and checking if the scrollbar is visible on a wide file?

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -298,15 +312,39 @@ class _CodeViewState extends State<CodeView>
...
+                                      longestLength,
+                                      line.toPlainText().length));
+                              final fileWidth =
+                                  CodeView.assumedCodeCharacterWidth *
+                                      longestLineLength;
+                              final boxWidth = math.max(
+                                  constraints.minWidth, constraints.maxWidth);
+
+                              return Scrollbar(
+                                isAlwaysShown: fileWidth > boxWidth,
```
kenzieschmoll, line 327: What does it look like if you set this to true on a narrow file?

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -34,6 +34,9 @@ import 'variables.dart';
 final debuggerCodeViewSearchKey =
     GlobalKey(debugLabel: 'DebuggerCodeViewSearchKey');
 
+// This is a conservative estimate, most fonts have a width/height ratio of ~0.5.
+const approximateFontWidthHeightRatio = 0.6;
```
jacob314, line 38: Rather than guessing, you can draw a character (e.g. ' ') and measure what the ratio is.
as the console is assuming a fixed width font, that is safe.
Existing example:
https://github.com/flutter/devtools/blob/b9950aa0c6af0e095f2c5b254f41a127c9481a70/packages/devtools_app/lib/src/common_widgets.dart#L1149

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: It shows up at the bottom even though the file isn't scrollable:
image

This also happens on hover currently unfortunately (scrollbar shows up when hovering at the bottom of the file even if it's not scrollable). I'm looking into why that is - once I fix that I think it would make sense to set this to `true`.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -264,6 +273,11 @@ class _CodeViewState extends State<CodeView>
             child: Expanded(
               child: Scrollbar(
                 controller: textController,
+                // Only listen for vertical scroll notifications (ignore those
```
jacob314, line 276: why are we getting notifications for the horizontal scoll controller? It looks like we are specifying just the textController which is the vertical scroll controller although it could use a better name.

packages/devtools_app/lib/src/debugger/codeview.dart
jacob314: the scrollbar seems like it is the wrong size. why isn't the scrollbar the entire width of the window given the content is no larger than the size of the window? what happens if you scroll in that case?

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -298,15 +312,39 @@ class _CodeViewState extends State<CodeView>
...
+                                isAlwaysShown: fileWidth > boxWidth,
+                                controller: horizontalController,
+                                child: SingleChildScrollView(
+                                  scrollDirection: Axis.horizontal,
+                                  controller: horizontalController,
+                                  child: Lines(
+                                    constraints: BoxConstraints(
+                                      maxHeight: constraints.maxHeight,
+                                      minHeight: constraints.minHeight,
+                                      minWidth: math.max(fileWidth, boxWidth),
```
jacob314, line 336: why not constrain to exactly the fileWidth?

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -298,15 +312,39 @@ class _CodeViewState extends State<CodeView>
...
+                                  constraints.minWidth, constraints.maxWidth);
+
+                              return Scrollbar(
+                                isAlwaysShown: fileWidth > boxWidth,
+                                controller: horizontalController,
+                                child: SingleChildScrollView(
+                                  scrollDirection: Axis.horizontal,
+                                  controller: horizontalController,
+                                  child: Lines(
+                                    constraints: BoxConstraints(
```
jacob314, line 333: Nit: in Flutter the general rule is to prefer composition over inheritance.
In this case what that means is rather than writing
```dart
Lines(constraints: foo, ...)
```
where the Lines build method adds a ConstrainedBox, write
```dart
ConstrainedBox(constraints: foo, child: Lines(...))
```
If there was a reason the Lines widget needed to know about the constraints then that would be different and passing in the constraints would be fine.

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: From the Scrollbar class [docs](https://api.flutter.dev/flutter/material/Scrollbar-class.html): "If the scrollbar is wrapped around multiple ScrollViews, it only responds to the nearest scrollView and shows the corresponding scrollbar thumb by default. The notificationPredicate allows the ability to customize which ScrollNotifications the Scrollbar should listen to."

The horizontal scrollview is the first nested scrollview, and then the vertical scrollview (see image). So this is to ignore notifications from the horizontal scrollview. Other option would be to try to change the nesting, so the list view is the first scrollable child under the vertical scrollbar.

image

packages/devtools_app/lib/src/debugger/codeview.dart
jacob314: In addition, I suspect you do not need to specify any constraint for the Y axis and that for the X axis you can clarify things by making the constraint more fixed. I may be getting things mixed up but I believe you should be able to just write
SizedBox(width: fileWidth)

elliette: `Lines` was already being passed `constraints` (it looks like to figure out where to scroll to when searching the file: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/debugger/codeview.dart#L572-L593). However all it uses is the `maxHeight`, so I could refactor to use `SizedBox` and only pass the height to `Lines`. Wdyt?

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Ah I see what the issue is. This happens when I open a file with the `Libraries` button, which opens the side panel. With the side panel open, the horizontal scrollbar is necessary. When I then close the side panel, the scrollbar isn't redrawn.

jacob314: hmm. Is the LayoutBuilder called again when the Libraries panel is closed? I would think it is.

packages/devtools_app/lib/src/debugger/codeview.dart
jacob314: I thought that is only the case if you do not specify a scrollController explicitly like you are.
Perhaps the Flutter docs should be updated to be more clear on that or I am misunderstanding something.

packages/devtools_app/lib/src/debugger/codeview.dart
jacob314: you are right. fine to leave as constraints or switch to SizedBox to be clearer.

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Ah interesting, it is being called. But the scrollbar isn't being removed, even with the new constraints. This looks similar to some issues that were marked as being recently fixed with https://github.com/flutter/flutter/pull/82687. I'm currently using the stable/official Flutter SDK, which doesn't have that change yet. I'm going to switch to a local fork of `flutter` and see if this issue goes away.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -264,6 +273,11 @@ class _CodeViewState extends State<CodeView>
             child: Expanded(
               child: Scrollbar(
                 controller: textController,
+                // Only listen for vertical scroll notifications (ignore those
+                // from the nested horizontal SingleChildScrollView):
+                notificationPredicate: (ScrollNotification notification) =>
```
xu-baolin, line 278: We should use `notification.depth` to predicate, otherwise, all the descendants of Scrollables will interfere.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -264,6 +273,11 @@ class _CodeViewState extends State<CodeView>
             child: Expanded(
               child: Scrollbar(
```
xu-baolin, line 274: In addition, if you want this scrollbar to be displayed automatically (without user scrolling operations), you need to set `isAlwaysShown` to true.

packages/devtools_app/lib/src/debugger/codeview.dart
jacob314: Addressing this comment will help the content be the correct size.

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Thanks! This scrollbar already existed, so I don't think we want to change how it's displayed.

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Thanks, changed to use `notificationDepth`.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -298,15 +315,35 @@ class _CodeViewState extends State<CodeView>
...
-                                activeSearchMatchNotifier:
-                                    widget.controller.activeSearchMatch,
+                              final num fileWidth = lines.fold(
+                                  0,
+                                  (widestWidth, line) => math.max(widestWidth,
+                                      CodeView.calculateLineWidth(line)));
+                              final boxWidth = math.max(
+                                  constraints.minWidth, constraints.maxWidth);
+
+                              return Scrollbar(
```
xu-baolin, line 325: "With this change: If the viewport starts out big enough that a horizontal scrollbar is not required, and then is resized smaller so that a scrollbar is now needed, no scrollbar will show up."

Did you mean this scrollbar's behavior is abnormal?

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Good point, now constraining to just `fileWidth`

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Switched to `SizedBox`, and now passing only the height to `Lines`

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Yes, it's abnormal. Let me try to capture some short videos, that might be easiest. Thanks!

xu-baolin: Thanks,could you also test with this change #87818?

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -48,6 +48,16 @@ class CodeView extends StatefulWidget {
   static double get rowHeight => scaleByFontFactor(20.0);
   static double get assumedCharacterWidth => scaleByFontFactor(16.0);
 
+  static num calculateLineWidth(textSpan) {
```
jacob314, line 51: move this to ui/util.dart

jacob314: give textSpan a type/

packages/devtools_app/lib/src/debugger/codeview.dart
jacob314: are we sure this notificationPredicate is needed? If it is needed, there are likely multiple other places in the UI where we need notificationPredicates. I thought it wasn't needed as we explicitly specify the scrollController.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -527,15 +561,15 @@ class GutterItem extends StatelessWidget {
...
+    @required this.height,
     @required this.scrollController,
     @required this.lines,
     @required this.pausedFrame,
     @required this.searchMatchesNotifier,
     @required this.activeSearchMatchNotifier,
   }) : super(key: key);
 
-  final BoxConstraints constraints;
+  final num height;
```
jacob314, line 572: flutter style: prefer to always be explicit about using `double` rather than `num` for num that can be a double. that is better for performance and avoids gotchas where an implicit cast from a num to double for a value that is an int could cause problems.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -298,15 +315,32 @@ class _CodeViewState extends State<CodeView>
...
-                                lines: lines,
-                                pausedFrame: pausedFrame,
-                                searchMatchesNotifier:
-                                    widget.controller.searchMatches,
-                                activeSearchMatchNotifier:
-                                    widget.controller.activeSearchMatch,
+                              final num fileWidth = lines.fold(
+                                  0,
+                                  (widestWidth, line) => math.max(widestWidth,
+                                      CodeView.calculateLineWidth(line)));
```
jacob314, line 321: measuring every single line in a large file will be fairly slow. You will probably notice a slowdown from this code given a 5000+ line file when you resize the window. 
Instead, take advantage of the fact that the font is fixed width and measure a single character and multiply or only measure the length of the longest line.

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Hopefully this is helpful. Here is the behavior with the stable Flutter SDK (as you can see, the issue described in https://github.com/flutter/flutter/issues/75613 is present - a non-functioning horizontal scrollbar is shown when it is not needed):

stable_scrollbar

And now here is the behavior using Flutter from head (including https://github.com/flutter/flutter/pull/87818). I'm shrinking the viewport so that the horizontal scrollbar is necessary. The horizontal scrollbar briefly appears for a second, but as I keep shrinking it disappears. 

scrollbar_fork

xu-baolin: Did you see the overflowing assertion of the second gif on the top right corner?

elliette: Hmm I think it's unrelated. Here is another file that doesn't need to be shrunk as much to need the scrollbar. As you can see, the horizontal scrollbar doesn't appear while shrinking (and there is no overflow assertion). I then switch tabs to reload the page, and on reload the scrollbar shows up:

scrollbar_fork2

xu-baolin: Did we change the bar's configurations such as 'isalwaysshow' during the resizing?

elliette: Nope, `isAlwaysShow` doesn't change (it's set to `true`). I haven't made any changes since my last commit (https://github.com/flutter/devtools/pull/3262/commits/1057cd965c753a334652b67369ca92e6bdcd43bf), so all the code should be visible.

xu-baolin: Thanks very much!
I will try to reproduce this problem locally. If it can be reproduced, I believe it will be resolved soon.

packages/devtools_app/lib/src/debugger/codeview.dart
kenzieschmoll: Also use double as the return type instead of num

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Great, thank you so much! The [contributing docs](https://github.com/flutter/devtools/blob/b9950aa0c6af0e095f2c5b254f41a127c9481a70/CONTRIBUTING.md) should be update with instructions on running devtools locally.

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: The `notificationPredicate` is needed, otherwise the vertical scrollbar responds to horizontal scroll events (and shows up at the bottom along with the horizontal scrollbar, see image). I'm guessing scrollbars in other parts of the UI would be similarly broken if this was needed for them?

image

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Resolving this in lieu of the other comment thread about the `notificationPredicate`

packages/devtools_app/lib/src/debugger/codeview.dart
xu-baolin: I think I found the root cause, and I will fix it on the master branch today.

xu-baolin: https://github.com/flutter/flutter/pull/88152 should fix this, @elliette, Could you help verfy this patch actually fix the issue? Thanks very much!

elliette: Yay, it's working with the patch! Thanks so much for your help getting this resolved, I really appreciate it. I will leave a comment in https://github.com/flutter/flutter/pull/88152 as well confirming that this fixes the issue I was seeing.

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Resolving this thread since https://github.com/flutter/flutter/pull/88152 should fix it

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Done, thanks!

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Makes sense, switched to double

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Good point, now find the longest line and then measuring that

packages/devtools_app/lib/src/ui/utils.dart
```diff
@@ -111,6 +111,17 @@ TextSpan truncateTextSpan(TextSpan span, int length) {
   return truncateHelper(span);
 }
 
+/// Returns a [double] representing the width in pixels of the [span].
```
jacob314, line 114: nit on doc comment writing. Try to avoid  saying things in doc comments that are obvious from the method name and signature.
For example, this comment can be written as
```dart
/// Returns the width in pixels of the [span].
```
without a reduction in clarity.

jacob314:
lgtm

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Hi @xu-baolin! I see you added the [`notificationPredicate`](https://github.com/flutter/flutter/commit/e7772d0e2ecb9529f692556e8b446de65219372c) 😄  Could you clarify what the purpose of the `notificationPredicate` is vs. the `scrollController`? Is the `scrollController` no longer used? Thanks!

packages/devtools_app/lib/src/ui/utils.dart
elliette: Makes sense, updated the doc comment!

jacob314 (approved):
lgtm_dog2

packages/devtools_app/lib/src/debugger/codeview.dart
xu-baolin: Ok I will check that.

xu-baolin:
Nice job!
I have posted a similar issue #2254 a year ago.

packages/devtools_app/lib/src/debugger/codeview.dart
Piinks: Oh this may already be fixed, we noticed this recently in https://github.com/flutter/flutter/issues/87697.
The scrollbar uses the scroll controller but also listens and responds to scroll events. This is useful like in this 2D case where the scrollbars might want to know about both axes and respond when either one scrolls. They just aren't behaving correctly here, should be fixed by https://github.com/flutter/flutter/pull/87698.

PR #3263, opened 2021-08-11
Changes to when analytics dialog appears and how it sets analytics

kenzieschmoll:
Changes:
- Changed some prompt styling. Added a close button in the upper right of the prompt (consistent with other banner-type messages in DevTools.
Screen Shot 2021-08-11 at 10 30 12 AM

- The prompt will show again once for all users because we have changed the 'firstRun' bit in the .devtools storage file to 'isFirstRun' (just to force the prompt to show again). At the time the prompt is shown, analytics will be enabled for a user, and the user will be given the opportunity to opt-out by responding to the dialog. The user can still opt out of analytics at any time through the setting menu.
- The 'enabled' bit in the .devtools storage file has been changed to 'analyticsEnabled' for more clarity.

Here is what the .devtools file will look like after these changes:
```
{
  "analyticsEnabled": true,
  "isFirstRun": false,
  "2020Q4": {
    "surveyActionTaken": false,
    "surveyShownCount": 0
  }
}
```

Verified that these prompts show up as expected in the embedded version of DevTools as well:

IntelliJ
Screen Shot 2021-08-11 at 11 56 19 AM

VS Code
Screen Shot 2021-08-11 at 11 32 25 AM

### DIFF
--- a/packages/devtools_app/lib/src/analytics/prompt.dart
+++ b/packages/devtools_app/lib/src/analytics/prompt.dart
@@ -5,6 +5,7 @@
 import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 
+import '../common_widgets.dart';
 import '../theme.dart';
 import '../utils.dart';
 import 'provider.dart';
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/analytics/prompt.dart
```diff
@@ -21,48 +22,68 @@ class AnalyticsPrompt extends StatefulWidget {
...
-  final AnalyticsProvider _provider;
+  Widget _child;
+  AnalyticsProvider _provider;
 
   bool _isVisible = false;
 
   @override
   void initState() {
     super.initState();
+    _provider = widget.provider;
```
jacob314, line 37: rather than caching, add a
`AnalyticsProvider get provider => _widget.provider` getter instead.

kenzieschmoll: done

packages/devtools_shared/lib/src/server/usage.dart
```diff
@@ -63,26 +63,44 @@ class DevToolsUsage {
...
+
+    properties['isFirstRun'] = properties['isFirstRun'] == null;
+    return properties['isFirstRun'];
   }
 
-  bool get enabled {
-    if (properties['enabled'] == null) {
-      properties['enabled'] = false;
+  bool get analyticsEnabled {
+    // TODO(kenz): remove this in ~6 months. The `enabled` property has been
```
jacob314, line 88: specify a removal date instead of 6 months as someone reading this code doesn't know when 6 months from now is. Also consider filing a tracking bug for the cleanup and reference that here.

jacob314 (approved):
lgtm

packages/devtools_shared/lib/src/server/usage.dart
kenzieschmoll: done. created https://github.com/flutter/devtools/issues/3264

PR #3266, opened 2021-08-11
Some syntax modifications to the analytics dimensions

kenzieschmoll:
From the gtags documentation, it appears the dimensions should be in quotes: https://developers.google.com/analytics/devguides/collection/gtagjs/custom-dims-mets

### DIFF
--- a/packages/devtools_app/lib/src/analytics/analytics.dart
+++ b/packages/devtools_app/lib/src/analytics/analytics.dart
@@ -442,7 +442,7 @@ void computeDevToolsQueryParams() {
   }
 }
 
-void computeFlutterClientId() async {
+Future<void> computeFlutterClientId() async {
   flutterClientId = await server.flutterGAClientID();
 }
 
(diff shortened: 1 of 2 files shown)


packages/devtools_app/web/devtools_analytics.js
```diff
@@ -44,18 +44,18 @@ function initializeGA() {
...
         // Custom dimensions:
-        dimension1: 'user_app',
-        dimension2: 'user_build',
-        dimension3: 'user_platform',
-        dimension4: 'devtools_platform',
-        dimension5: 'devtools_chrome',
-        dimension6: 'devtools_version',
-        dimension7: 'ide_launched',
-        dimension8: 'flutter_client_id',
+        'dimension1': 'user_app',
```
jacob314, line 47: in JS, writing dimension1 in this position is equivalent to writing 'dimension1' so this shouldn't change any actual functionality but I agree writing 'dimension1' is clearer.

packages/devtools_app/lib/src/analytics/analytics.dart
```diff
@@ -475,12 +475,12 @@ void setupAndGaScreen(String screenName) async {
   }
 }
 
-void setupDimensions() {
+Future<void> setupDimensions() async {
```
jacob314, line 478: should we now be awaiting this function everywhere we are calling it?

kenzieschmoll: the only place we call this is in `DevToolsAppState.initState`, so we cannot await it there. Regardless of this though, I think it's better to have the proper return type here so we can await `computeFlutterClientId` before setting the `_analyticsComputed` bit to true at the end of this method.

jacob314:
lgtm

Issue #3270, opened 2021-08-12
Add analytics for all important UX elements in DevTools


kenzieschmoll:
We need to track events for button presses, selections, etc. to better understand how our users are using DevTools and to measure impact of features.

kenzieschmoll:
Closing this as complete. We can add more analytics events as needs come up.

PR #3272, opened 2021-08-12
Add some analytics to the memory screen and clean up code.

kenzieschmoll:
Work towards https://github.com/flutter/devtools/issues/3270

### DIFF
--- a/packages/devtools_app/lib/src/memory/memory_screen.dart
+++ b/packages/devtools_app/lib/src/memory/memory_screen.dart
@@ -8,6 +8,7 @@ import 'package:provider/provider.dart';
 
 import '../analytics/analytics_stub.dart'
     if (dart.library.html) '../analytics/analytics.dart' as ga;
+import '../analytics/constants.dart' as analytics_constants;
 import '../auto_dispose_mixin.dart';
 import '../banner_messages.dart';
 import '../charts/chart_controller.dart';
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -207,7 +172,10 @@ class MemoryBodyState extends State<MemoryBody>
     addAutoDisposeListener(controller.androidChartVisibleNotifier, () {
       setState(() {
         if (controller.androidChartVisibleNotifier.value) {
-          MemoryScreen.gaAction(key: MemoryScreen.androidChartButtonKey);
+          ga.select(
```
jacob314, line 175: maybe add a helper to DevTools screen that 
always passes in the analytics constant for the current screen?

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/memory/memory_screen.dart
kenzieschmoll: This is tricky because the calls to `ga.select` do not happen in the Widget subclasses of Screen. At the time that we call `ga.select`, we often times do not even have access to the Screen object, which is why we would have to get the id from a static method on the screen class, but we don't have access to the screenId from a screen in a static context.

PR #3274, opened 2021-08-12
Performance improvements for the CPU profiler

kenzieschmoll:
The recent introduction of profile filtering added more calls to `getSimplePackageUrl` than we would have liked. This PR improves the performance of the profiler by caching some values:
- the `processedUrl` returned by `getSimplePackageUrl` for a stack frame. We then pass this value as a parameter during stack frame copy operations (we were previously computing it for every copy, which is unnecessary).
- the `isNative`, `isDartCore`, and `isFlutterCore` values for a single stack frame. These no longer have to be re-computed for filter operations.

### DIFF
--- a/packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
+++ b/packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
@@ -441,6 +441,18 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
     @required this.profileMetaData,
   });
 
+  /// Prefix for packages from the core Dart libraries.
+  static const dartPackagePrefix = 'dart:';
+
+  /// Prefix for packages from the core Flutter libraries.
+  static const flutterPackagePrefix = 'package:flutter/';
+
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/profiler/cpu_profile_model.dart
```diff
@@ -434,6 +457,22 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
...
+
+  bool _isNative;
+
+  bool get isDartCore => _isDartCore ??= processedUrl.startsWith('dart:');
+
+  bool _isDartCore;
+
+  bool get isFlutterCore =>
+      _isFlutterCore ??= processedUrl.startsWith('package:flutter/') ||
+          name.startsWith('flutter::') ||
```
jacob314, line 471: nit: maybe make these constants so you have an easy place to document that 
`flutter::` is the Flutter namespace in C++?

jacob314 (approved):
lgtm

Issue #3276, opened 2021-08-13
Logging filters broken at HEAD


goderbauer:
I was running the devtools app with `flutter run -d chrome` and when trying to use Filters on the logging screen it is throwing the following error:

```
Expected a value of type 'Filter<LogData>', but got one of type 'Filter<dynamic>'
══╡ EXCEPTION CAUGHT BY GESTURE ╞═══════════════════════════════════════════════════════════════════
The following TypeErrorImpl was thrown while handling a gesture:
Expected a value of type 'Filter<LogData>', but got one of type 'Filter<dynamic>'

When the exception was thrown, this was the stack:
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 251:49      throw_
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 84:3        castError
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 452:10  cast
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/types.dart 395:9        as
packages/devtools_app/src/logging/logging_controller.dart 624:35                  filterData
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 334:14  _checkAndCall
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 398:10  callMethod
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 402:5   dsend
packages/devtools_app/src/ui/filter.dart 101:46                                   <fn>
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 334:14  _checkAndCall
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 398:10  callMethod
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 402:5   dsend
packages/devtools_app/src/dialogs.dart 105:43                                     <fn>
packages/flutter/src/material/ink_well.dart 989:21                                [_handleTap]
packages/flutter/src/gestures/recognizer.dart 198:24                              invokeCallback
packages/flutter/src/gestures/tap.dart 608:48                                     handleTapUp
packages/flutter/src/gestures/tap.dart 296:5                                      [_checkUp]
packages/flutter/src/gestures/tap.dart 230:7                                      handlePrimaryPointer
[...]

PR #3277, opened 2021-08-13
Fix type errors in Filter code

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/3276

### DIFF
--- a/packages/devtools_app/lib/src/network/network_screen.dart
+++ b/packages/devtools_app/lib/src/network/network_screen.dart
@@ -223,7 +223,7 @@ class _NetworkScreenBodyState extends State<NetworkScreenBody>
   void _showFilterDialog() {
     showDialog(
       context: context,
-      builder: (context) => FilterDialog(
+      builder: (context) => FilterDialog<NetworkController, NetworkRequest>(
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/network/network_screen.dart
```diff
@@ -223,7 +223,7 @@ class _NetworkScreenBodyState extends State<NetworkScreenBody>
   void _showFilterDialog() {
     showDialog(
       context: context,
-      builder: (context) => FilterDialog(
+      builder: (context) => FilterDialog<NetworkController, NetworkRequest>(
```
jacob314, line 226: why aren't these types able to be inferred correctly?

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/network/network_screen.dart
kenzieschmoll: we need to pass these types along to the Filter classes and Filter mixin that expect a generic type to be defined. This lets us use the same filter code across different screens with different data types.

PR #3278, opened 2021-08-13
[WIP] Use new RawTableView widget for Logging table

goderbauer:
This is not ready to be merged. It's an exemplary integration of the new table widget that I am experimenting with over at https://github.com/goderbauer/scratchpad/tree/main/table. The main goal here is to learn if the widget can do all the things it needs to do by using it for a real world use case (here: the logging table of devtools) and to identify any bugs.

Currently, the table widget only offers a very low-level API where developers have to write and manage their own delegate to use it. I plan to provide a higher level abstraction around this API for row-based tables like the Logging table. The experience from integrating the table into this real-world use case is helping in shaping this higher-level API.

There are also plans to add additional features to the table (e.g. mouse-resizable columns, sizing a column based on content, etc).

Leaving this here as a draft to gather some feedback.

To provide feedback on table implementation, comment here: https://github.com/goderbauer/scratchpad/pull/1/files

### DIFF
--- a/packages/devtools_app/devtools_app.iml
+++ b/packages/devtools_app/devtools_app.iml
@@ -9,6 +9,18 @@
       <excludeFolder url="file://$MODULE_DIR$/.idea" />
       <excludeFolder url="file://$MODULE_DIR$/.pub" />
       <excludeFolder url="file://$MODULE_DIR$/build" />
+      <excludeFolder url="file://$MODULE_DIR$/test/fixtures/flutter_error_app/build" />
+      <excludeFolder url="file://$MODULE_DIR$/test/fixtures/flutter_error_app/.pub" />
+      <excludeFolder url="file://$MODULE_DIR$/test/fixtures/flutter_error_app/.dart_tool" />
+      <excludeFolder url="file://$MODULE_DIR$/test/fixtures/riverpod_app/.dart_tool" />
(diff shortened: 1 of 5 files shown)


kenzieschmoll:
Without reviewing the code yet, here are a few things I noticed:
1. column sorting is missing
2. scrollbar is missing
3. When scrolling through the list, the hover row is constantly updated. Is this expected? What is the design standard for scrolling through a list of items with a hover property (e.g. is there normally a delay to update hover until after the scroll is complete)?
flat-table-hover-state

goderbauer:
Thanks for the feedback!

> column sorting is missing

Yeah, that one is not implemented yet... :)

> scrollbar is missing
 
Scrollbars do show up for me. Are you running this with the latest flutter (master)? That one should add scrollbars automatically here.

> When scrolling through the list, the hover row is constantly updated. Is this expected?

I believe the old table here was also doing that. It looks like the hover color may have been a little more subtile to make this less noticeable (plus I think it is fading in slowly).  Having said that, the way hover is done in the new table does not satisfy me yet. I am still experimenting with it...

kenzieschmoll:
Another observation: we need support for arrow key navigation (up / down)

Hixie:
Since this hasn't been touched in many months I'm going to close it to get it off the review queue.

PR #3280, opened 2021-08-16
add a package:args Command implementation

devoncarew:
- add a package:args Command implementation

This will make it easier to implement something like `dart devtools` - we can delegate to the package:devtools and package:devtools_server implementation instead of doing something like copying CLI args and calling `serveDevToolsWithArgs`.

### DIFF
--- a/packages/devtools_server/lib/src/devtools_command.dart
+++ b/packages/devtools_server/lib/src/devtools_command.dart
@@ -1,0 +1,39 @@
+// Copyright 2021 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:args/command_runner.dart';
+
+import '../devtools_server.dart';
+import 'server.dart';
+
+class DevToolsCommand extends Command<int> {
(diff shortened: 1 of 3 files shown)


packages/devtools_server/lib/src/devtools_command.dart
```diff
@@ -0,0 +1,39 @@
...
+// found in the LICENSE file.
+
+import 'package:args/command_runner.dart';
+
+import '../devtools_server.dart';
+import 'server.dart';
+
+class DevToolsCommand extends Command<int> {
+  DevToolsCommand({
+    this.customDevToolsPath,
```
kenzieschmoll, line 12: what is the customDevToolsPath for?

devoncarew: In the default shelf handler, we resolve the location of package:devtools in order to find its build/ directory. For the pub published version of the package that contains the compiled web resources. In the sdk distro those don't exist relative to package:devtools but in the sdk's `bin/resources/devtools` directory; this override lets us point the devtools server to that location.

kenzieschmoll (approved):

PR #3281, opened 2021-08-16
Add analytics to performance and cpu profiler screens

kenzieschmoll:
This PR also

- adds a timing API to the analytics code so that we can track the timing of operations in DevTools.
- adds some extra custom dimensions related to performance and cpu profiling tasks

Work towards https://github.com/flutter/devtools/issues/3270

### DIFF
--- a/packages/devtools_app/lib/src/analytics/analytics.dart
+++ b/packages/devtools_app/lib/src/analytics/analytics.dart
@@ -20,8 +20,11 @@ import '../config_specific/logger/logger.dart';
 import '../config_specific/server/server.dart' as server;
 import '../config_specific/url/url.dart';
 import '../globals.dart';
+import '../performance/performance_screen.dart';
+import '../profiler/profiler_screen.dart';
 import '../ui/gtags.dart';
 import '../version.dart';
+import 'analytics_common.dart';
 import 'constants.dart' as analytics_constants;
 
 // Dimensions1 AppType values:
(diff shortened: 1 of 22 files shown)


packages/devtools_app/lib/src/analytics/analytics.dart
```diff
@@ -214,6 +226,35 @@ void screen(
   );
 }
 
+void timing(
+  String screenName,
+  String timedOperation, {
+  @required Duration duration,
```
jacob314, line 232: Comment to clarify that duration should always refers to time DevTools took not time an app took
I would suggest creating a DevToolsScreen metrics objects that groups 
cpuSampleCount, cpuStackDepth, and traceEventCount. Those are very performance page specific so it is a bit surprising to see them in this method. Keeping them separate will make the important parts of this API clearer and make it easier if you chose to refactor how those metrics are exposed. Internally it could be just a key value pair map or something.
I would also suggest a timing api that is more similar to what dart:developer does. So for example:
```dart
AnalyticsTimeline.timeSync("screenName", "someTimedOperation", () {
  doSomething();
}, ScreenMetrics(cpuSampleCount: 1000, inspectorTreeSize: 100000)
);```
https://api.dart.cn/stable/2.13.4/dart-developer/Timeline-class.html
I'm less sure about the async version of the API as the one in dart:developer seems like it was written before Dart supported async. Something like this might make sense:
```dart
AnalyticsTimeline.timeAsync("screenName", "someTimedOperation", someFutureThatCompleteAtTheEndTime, ScreenMetrics(cpuSampleCount: 1000, inspectorTreeSize: 100000)
);```

Comment (kenzieschmoll) [reply packages/devtools_app/lib/src/analytics/analytics.dart:232]: did some significant refactoring - PTAL

File: packages/devtools_app/lib/src/analytics/analytics.dart [hunk 2]
```diff
@@ -239,18 +280,20 @@ void select(
 
 // Used only for Timeline Frame selection.
 void selectFrame(
-  String screenName,
-  String selectedItem, [
-  int rasterDuration, // Custom metric
-  int uiDuration, // Custom metric
-]) {
+  String screenName, {
+  @required Duration rasterDuration, // Custom metric
```
jacob314, line 284: I would omit these custom metric comments. not clear what they are communicating.

packages/devtools_app/lib/src/analytics/analytics.dart
```diff
@@ -86,14 +89,69 @@ class GtagEventDevTools extends GtagEvent {
...
+      value: value,
+      user_app: userAppType,
+      user_build: userBuildType,
+      user_platform: userPlatformType,
+      devtools_platform: devtoolsPlatformType,
+      devtools_chrome: devtoolsChrome,
+      devtools_version: devtoolsVersion,
+      ide_launched: ideLaunched,
+      flutter_client_id: flutterClientId,
+      ui_duration_micros: screenMetrics is PerformanceScreenMetrics
```
jacob314, line 133: having is checks for subtypes isn't very robust. Can you instead switch to adding a toJson() or similar method in the ScreenAnalyticsMetrics subclasses that surfaces the custom metrics each screen cares about?
Will requiring tweaking GtagEventDevTools to take a JSON map of extra parameters rather than just a fixed list of fields but probably worth it to keep adding extra metrics simple.

jacob314: To land this now add a TODO here  and a comment in the ScreenAnalyticsMetrics base class clarifying that extending this class is fragile.

packages/devtools_app/lib/src/analytics/analytics_common.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright 2021 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// Code in this file should be able to be imported by both dart:html and
+// dart:io dependent libraries.
+
+/// Base class for all screen metrics classes.
```
jacob314, line 8: Describe what is involved in creating a subclass or switch to a 
toJson() based scheme so that it isn't magical.

jacob314 (commented):
Gode to submit once one of the two suggested changes is made.

packages/devtools_app/lib/src/analytics/analytics.dart
kenzieschmoll: I thought about this. Since the `GtagEventDevTools` is an external factory constructor, I'm not sure how passing json in this would affect the custom dimension setup in JS.

I will go ahead and add a TODO to consider trying to make this take a json map in the future, but for now I will add better documentation.

packages/devtools_app/lib/src/analytics/analytics_common.dart
kenzieschmoll: done

PR #3285, opened 2021-08-17
Sort timeline events before processing

kenzieschmoll:
As suggested on https://github.com/flutter/flutter/issues/88056, this will help us to handle out of order input from the vm

### DIFF
--- a/packages/devtools_app/lib/src/performance/timeline_event_processor.dart
+++ b/packages/devtools_app/lib/src/performance/timeline_event_processor.dart
@@ -111,8 +111,6 @@ class TimelineEventProcessor {
 
   int rasterThreadId;
 
-  final _lastProcessedTimestampByThreadId = <int, int>{};
-
   /// Process the given trace events to create [TimelineEvent]s.
   ///
   /// [traceEvents] must be sorted in increasing timestamp order before calling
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/performance/timeline_event_processor.dart
```diff
@@ -130,71 +128,22 @@ class TimelineEventProcessor {
 //        .map((e) => TraceEventWrapper(
 //            TraceEvent(e), DateTime.now().microsecondsSinceEpoch))
 //        .toList();
-    const sixteenHoursMicros = 57600000000;
-    int firstTimestampMicros;
-    final _traceEvents = (traceEvents.whereFromIndex(
+
+    final _traceEvents = (traceEvents.sublist(startIndex)
```
jacob314, line 132: could you still have an edge case where the events before startIndex are out of timestamp order?

kenzieschmoll: I suppose that could be possible, but everything before `startIndex` will have already been processed and displayed in the timeline. So if we have set A that contains events from index 0 to index 10000, and set B that contains events from index 10001 to 20000, there could theoretically always be possibility that an event belonging in set A (according to timestamps) came in set B or even set C for that matter. There really is no way to guarantee this doesn't happen, as we are at the mercy of when events are sent by the framework, engine, etc. - but this is unlikely even though it is possible.

jacob314: thanks the part about incomplete events being processed later was what I wasn't thinking about.

jacob314 (approved):
lgtm

Issue #3286, opened 2021-08-17
Cannot select network request by clicking on Uri


kenzieschmoll:
This is because the Uri column is rendered with selectable text, so currently, the click is captured by the SelectableText widget and is not passed through to the row.

PR #3287, opened 2021-08-17
Fix selection bug in the network profiler

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/3286

### DIFF
--- a/packages/devtools_app/lib/src/table.dart
+++ b/packages/devtools_app/lib/src/table.dart
@@ -1037,7 +1037,12 @@ abstract class ColumnRenderer<T> {
   ///
   /// This method can return `null` to indicate that the default rendering
   /// should be used instead.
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/table.dart
```diff
@@ -1215,7 +1220,8 @@ class _TableRowState<T> extends State<TableRow<T>>
 
   @override
   Widget build(BuildContext context) {
-    final row = tableRowFor(context);
+    final onPressed = () => widget.onPressed(widget.node);
```
jacob314, line 1223: nit: probably cleaner to just inline this in the class to
tableRowFor.

jacob314 (approved):
nice regregression test!

jacob314:
lgtm

PR #3293, opened 2021-08-18
Updated macos/Runner.xcodeproj/project.pbxproj

elliette:
The following was added automatically as a result of running `flutter run -d macos`. 

I also trying deleting the file entirely and adding it to `.gitignore`, but I ran into an error trying to run DevTools with the file deleted.


jacob314 (approved):
lgtm

Issue #3290, opened 2021-08-18
HTTP requests cause memory leak when devtools is attached.


natebosch:
Filing this here since attaching devtools is what surfaces the bug. The issue may be in the VM.

This was originally filed as a bug on `package:http`. https://github.com/dart-lang/http/issues/592

You can see the behavior with the following server and client:

```dart
// server
import 'dart:convert';
import 'dart:io';
import 'dart:math';

void main() async {
  var server = await HttpServer.bind('localhost', 9090);
    server.listen((request) async {
    request.response.write(base64Encode(_randomBytes));
    await request.response.close();
  });

}

final _random = Random();
List<int> get _randomBytes =>
    List.generate((1024 * 1024 * 10), (_) => _random.nextInt(255));


```

```dart
// client - run and attach devtools
import 'dart:convert';

import 'package:http/http.dart' as http;

void main() async {
  while (true) {
    var result = await http.get(Uri.http('localhost:9090', ''));
    var decoded = base64Decode(result.body);
    await Future.delayed(const Duration(milliseconds: 500));
  }
}
```

When I run this without attaching the debugger the memory usage for the `dart` process stays low. When I attach devtools to the observatory URL the memory usage starts growing.

kenzieschmoll:
@bkonyi

bkonyi:
Yes, that's expected since we cache the network request data when we start profiling. Invoking `ext.dart.io.clearHttpProfile` will free the cached data.

natebosch:
> Invoking `ext.dart.io.clearHttpProfile` will free the cached data.

@kenzieschmoll - is there a way to do this from the UI? Should we consider adding one?

kenzieschmoll:
>> Invoking ext.dart.io.clearHttpProfile will free the cached data.

> @kenzieschmoll - is there a way to do this from the UI? Should we consider adding one?

We are not currently calling this when the 'Clear' button is tapped. I'll push up a PR to do this.

kenzieschmoll:
Fix PR: https://github.com/flutter/devtools/pull/3294

caoquanyong:
> Yes, that's expected since we cache the network request data when we start profiling. Invoking `ext.dart.io.clearHttpProfile` will free the cached data.
How to invoking `ext.dart.io.clearHttpProfile`, should i add it to my code ，Where to add it

PR #3294, opened 2021-08-19
Clear HttpProfile on VM when pressing 'Clear' on network page

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/3290


bkonyi (approved):
LGTM. Just a heads up, this won't clear the socket profiling data, but we don't actually store data sent / received over the socket so it shouldn't be much of a memory overhead compared to the HTTP profiling data.

kenzieschmoll:
We actually were already clearing the socket profile on pressing the 'clear' button. We are making a call to `service.clearSocketProfile` for each isolate.

Issue #3295, opened 2021-08-20
UI thread timeline events for a flutter frame are being improperly grouped


kenzieschmoll:
Sometimes a `MessageLoop::FlushTasks` event can parent the event groups for multiple flutter frames. We currently look for an event with the "frame_number" arg and assign the root of that event to a flutter frame object. This scenario breaks that logic.
<img width="1107" alt="Screen Shot 2021-08-20 at 9 21 09 AM" src="https://user-images.githubusercontent.com/43759233/130264122-2303d1a1-8154-4720-9605-fc6d0fb564d3.png">

LeonManolo:
Could someone please clarify if the "MessageLoop::FlushTasks" event has been removed or renamed? I'm looking to measure the total CPU time spent running Dart code by examining the extent of this event in the timeline. However, I can't seem to find it in my timeline anymore. Any guidance or updates on this would be greatly appreciated. Thank you!

kenzieschmoll:
To measure CPU usage, I would recommend using the DevTools CPU profiler to record a CPU profile and look at CPU samples. Right now, CPU samples are not included in the Timeline events. This is blocked on https://github.com/flutter/devtools/issues/4755.

Regarding the "MessageLoop::FlushTasks" event, @dnfield can you comment on what conclusions are safe to draw around this event? Or if it exists anymore?

LeonManolo:
> To measure CPU usage, I would recommend using the DevTools CPU profiler to record a CPU profile and look at CPU samples. Right now, CPU samples are not included in the Timeline events. This is blocked on #4755.
> 
> Regarding the "MessageLoop::FlushTasks" event, @dnfield can you comment on what conclusions are safe to draw around this event? Or if it exists anymore?

Thank you, @kenzieschmoll, for your suggestion regarding the DevTools CPU profiler. I'll definitely explore using CPU samples for a broader performance analysis.

My current approach is somewhat targeted, as I'm specifically looking to compare the performance of two slightly different algorithms within a feature of my app. To do this, I've written an integration test for this feature and used `IntegrationTestWidgetsFlutterBinding.traceAction` to capture the timeline JSON. My understanding is that Dart code executes on the UI thread `name.startsWith('1.ui')`, as I'm not utilizing isolates for this test. I then focused on the top-level events in the timeline (those that appear at the top in the chrome://tracing visualization) for each algorithm's implementation.
[...]

PR #3297, opened 2021-08-20
Tweak how we assign timeline events to flutter frames

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/3295. This assigns the event with the "frame_number" argument to the FlutterFrame object. This does change how we display the selected frame brackets slightly. They are now around the event with the "frame_number" arg instead of the event root (as seen in the linked issue, sometimes the event root is also the parent of many other unrelated events).

Screen Shot 2021-08-20 at 2 56 37 PM
Screen Shot 2021-08-20 at 2 56 48 PM

### DIFF
--- a/packages/devtools_app/lib/src/performance/performance_model.dart
+++ b/packages/devtools_app/lib/src/performance/performance_model.dart
@@ -922,29 +922,17 @@ class SyncTimelineEvent extends TimelineEvent {
     return frameNumber != null ? int.tryParse(frameNumber) : null;
   }
 
-  List<SyncTimelineEvent> get uiFrameEvents {
-    if (_uiFrameEvents != null) return _uiFrameEvents;
-    _uiFrameEvents = nodesInTreeWithCondition((TimelineEvent event) {
-      return event.name.contains(uiEventName) &&
-          event.traceEvents.first.event.args
+  bool get isUiFrameEvent => name.contains(uiEventName) &&
+          traceEvents.first.event.args
               .containsKey(TraceEvent.frameNumberArg);
-    }).cast<SyncTimelineEvent>();
-    return _uiFrameEvents;
-  }
 
-  List<SyncTimelineEvent> _uiFrameEvents;
+  bool get isRasterFrameEvent => name.contains(rasterEventName) &&
+      traceEvents.first.event.args
+          .containsKey(TraceEvent.frameNumberArg);
 
-  List<SyncTimelineEvent> get rasterFrameEvents {
-    if (_rasterFrameEvents != null) return _rasterFrameEvents;
-    _rasterFrameEvents = nodesInTreeWithCondition((TimelineEvent event) {
-      return event.name.contains(rasterEventName) &&
-          event.traceEvents.first.event.args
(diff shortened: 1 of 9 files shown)


packages/devtools_app/lib/src/performance/performance_model.dart
```diff
@@ -917,43 +917,34 @@ abstract class TimelineEvent extends TreeNode<TimelineEvent>
...
-    if (gpuRasterizerDrawEvent != null) {
-      try {
-        return int.parse(gpuRasterizerDrawEvent
-            .traceEvents.first.event.args[TraceEvent.frameNumberArg]);
-      } catch (_) {
-        return null;
-      }
-    }
-    return null;
+  List<SyncTimelineEvent> get uiFrameEvents {
```
jacob314, line 925: shouldn't be a getter given it is traversing the entire tree.

jacob314: never mind. it is cached so I'm less concerned about it being a getter.

kenzieschmoll (commented):
Took a different approach with better performance. PTAL

packages/devtools_app/lib/src/performance/performance_model.dart
```diff
@@ -917,43 +917,26 @@ abstract class TimelineEvent extends TreeNode<TimelineEvent>
...
-      try {
-        return int.parse(animatorBeginFrameEvent
-            .traceEvents.first.event.args[TraceEvent.frameNumberArg]);
-      } catch (_) {
-        return null;
-      }
-    }
-    return null;
+  int get frameNumberFromArgs {
+    final frameNumber = traceEvents.first.event.args[TraceEvent.frameNumberArg];
```
jacob314, line 921: Does the getter need to clarify where it is from?
Can this just be called
frameNumber.
You might want to still cache the value as int.tryParse is somewhat expensive.

packages/devtools_app/lib/src/performance/performance_model.dart
```diff
@@ -917,43 +917,26 @@ abstract class TimelineEvent extends TreeNode<TimelineEvent>
...
-            .traceEvents.first.event.args[TraceEvent.frameNumberArg]);
-      } catch (_) {
-        return null;
-      }
-    }
-    return null;
-  }
+  /// Whether this event is contains the flutter frame identifier for the UI
+  /// thread in its trace event args.
+  bool get isUiFrameIdentifier => name.contains(uiEventName) &&
```
jacob314, line 927: also consider caching these if they are called more than once.
Would it be possible to perform an exact match on the name rather than using contains?
Contains can false positive and is slow. If an exact match isn't posible, using startsWith or endsWith is faster and less false positive prone.

kenzieschmoll: cached and used ==

packages/devtools_app/lib/src/performance/performance_model.dart
kenzieschmoll: cached and renamed

jacob314 (approved):
lgtm

PR #3298, opened 2021-08-20
Cleanup console with eval that scrolls and less busy splitter.

jacob314:
Screen Shot 2021-08-20 at 3 28 15 PM
New implicit splitter that doubles as the title line better matches the rest of the debugger page.
Eval line is now just another line in the console leaving more room to view content when scrolling.
small_gif

### DIFF
--- a/packages/devtools_app/lib/src/debugger/console.dart
+++ b/packages/devtools_app/lib/src/debugger/console.dart
@@ -10,7 +10,6 @@ import '../console.dart';
 import '../console_service.dart';
 import '../globals.dart';
 import '../theme.dart';
-import '../utils.dart';
 import 'debugger_controller.dart';
 import 'evaluate.dart';
 
--- a/packages/devtools_app/lib/src/debugger/console.dart
+++ b/packages/devtools_app/lib/src/debugger/console.dart
@@ -6,10 +6,10 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
 
-import '../common_widgets.dart';
 import '../console.dart';
 import '../console_service.dart';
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/debugger/console.dart
```diff
@@ -30,34 +30,21 @@ class DebuggerConsole extends StatelessWidget {
...
-                    onPressed: () => serviceManager.consoleService.clearStdio(),
-                  ),
-                ],
+    return Column(
+      children: [
+        Expanded(
+          child: Console(
+            lines: stdio,
+            footer: SizedBox(
+              height: scaleByFontFactor(18.0),
```
kenzieschmoll, line 39: put 18.0 in a const in theme.dart

packages/devtools_app/lib/src/scaffold.dart
```diff
@@ -78,7 +79,7 @@ class DevToolsScaffold extends StatefulWidget {
         horizontalPadding.left,
         isEmbedded() ? 2.0 : 16.0,
         horizontalPadding.right,
-        0.0,
+        isEmbedded() ? 0.0 : 8.0,
```
kenzieschmoll, line 82: while you're here, can we use defaultSpacing and denseSpacing for 16.0 and 8.0

kenzieschmoll (approved):

packages/devtools_app/lib/src/debugger/console.dart
jacob314: done

packages/devtools_app/lib/src/scaffold.dart
jacob314: done

PR #3299, opened 2021-08-21
Support custom font sizes better across all pages.

jacob314:
Make most of the devtools responsive to font size changes.
Default to a naive strategy of scaling icons and all fonts but not changing padding.
All widths that were based on fitting a certain amount of text now scale with font size.
All row heights now scale with font size. Before tons of UI was badly broken.
Diff is large because of Dart const meaning that many const vals are no longer const.

Default to all font sizes scaling.
Default to not scaling padding
Default to scaling all icons
Default to scaling all sizes that were picked to fit a fixed amount of text
Fix a ton of places where we were accidentally not passing in a font size and it mattered as we were just getting the default font size not the one from our theme.
Add DevToolsTab base class to support tab scaling so large font sizes don't cause wrap issues
Fix table row heights to respect the font size.
Unpack a bunch of heights that were a combo of implicit padding and font size.

Limitations:
Not attempting to scale fonts on the memory chart or in axis labels in the flame chart as those assume a font that can fit
between ticks.
As always, font scaling changes will not reload so to test this you will need to start the app again. This could also be an issue if users changed the font size with a query parameter.

### Before examples:
All these examples have a silly large font size to make it easier to  spot where we were using inconsistent fonts.

Screen Shot 2021-08-13 at 4 09 21 PM
Screen Shot 2021-08-13 at 3 36 30 PM
Screen Shot 2021-08-13 at 1 37 25 PM
Screen Shot 2021-08-12 at 3 35 17 PM
Screen Shot 2021-08-12 at 2 30 53 PM

### After examples:
All these examples have a silly large font size to verify things are still ok.
Screen Shot 2021-08-20 at 5 26 20 PM
Screen Shot 2021-08-20 at 5 25 42 PM
Screen Shot 2021-08-20 at 5 25 18 PM
Screen Shot 2021-08-20 at 5 25 00 PM
Screen Shot 2021-08-20 at 5 24 26 PM
Screen Shot 2021-08-20 at 5 12 23 PM
Screen Shot 2021-08-20 at 4 56 38 PM


packages/devtools_app/lib/src/app_size/app_size_screen.dart
```diff
@@ -81,10 +82,11 @@ class AppSizeBody extends StatefulWidget {
 
 class _AppSizeBodyState extends State<AppSizeBody>
     with AutoDisposeMixin, SingleTickerProviderStateMixin {
-  static const diffTab = Tab(text: 'Diff', key: AppSizeScreen.diffTabKey);
-  static const analysisTab =
-      Tab(text: 'Analysis', key: AppSizeScreen.analysisTabKey);
-  static const tabs = [analysisTab, diffTab];
+  static final diffTab =
+      DevToolsTab(text: 'Diff', key: AppSizeScreen.diffTabKey);
```
jacob314, line 86: The regular tab class doesn't gracefully handle font size changes.

packages/devtools_app/lib/src/app.dart
```diff
@@ -419,7 +419,7 @@ class OpenAboutAction extends StatelessWidget {
           width: DevToolsScaffold.actionWidgetSize,
           height: DevToolsScaffold.actionWidgetSize,
           alignment: Alignment.center,
-          child: const Icon(
+          child: Icon(
```
jacob314, line 422: lots of values have to stop being const due to fewer size and font defaults being const.

packages/devtools_app/lib/src/banner_messages.dart
```diff
@@ -234,26 +234,33 @@ class DebugModePerformanceMessage {
...
       textSpans: [
-        const TextSpan(
+        TextSpan(
           text: '''
 You are running your app in debug mode. Debug mode performance is not indicative of release performance.
 Relaunch your application with the '--profile' argument, or ''',
-          style: TextStyle(color: _BannerError.foreground),
+          style: TextStyle(
+            color: _BannerError.foreground,
+            fontSize: defaultFontSize,
```
jacob314, line 243: we failed to use our default font size every time we emitted a custom TextStyle without a font size.

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -600,14 +600,13 @@ class ChartPainter extends CustomPainter {
...
-      color: Colors.black,
-      fontSize: 30,
-    );
     final span = TextSpan(
       // TODO(terry): All text in a chart is grey. A chart like a Trace
       //              should have PaintCharacteristics.
-      style: TextStyle(color: Colors.grey[600], fontSize: 10),
+      style: TextStyle(
+        color: Colors.grey[600],
+        fontSize: 10,
```
jacob314, line 608: intentionally left this alone as there are other hard coded consts for sizes in charts I didn't want to muck with and label text is ok to leave consistently small.

packages/devtools_app/lib/src/charts/chart.dart
```diff
@@ -600,14 +600,13 @@ class ChartPainter extends CustomPainter {
   }
 
   TextPainter createText(String textValue, double scale) {
-    const TextStyle(
```
jacob314, line 603: @pq why don't our dead code lints catch const values that can never be referenced like this one? If this was a non-const class maybe it could have side effects but as is it is pointless.

packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -25,9 +25,16 @@ import '../ui/search.dart';
...
+// Flame chart rows contain text so are not readable if they do not scale with
+// the font factor.
+double get rowHeight => scaleByFontFactor(25.0);
+double get rowHeightWithPadding => rowHeight + rowPadding;
+
+// This spacing needs to be scaled by the font factor otherwise section
+// labels will not have enough room. Typically spacing values should not depend
+// on the font size scale factor. TODO(jacobr): clean up the section spacing so
+// it is not used in a case where it is not really spacing.
+double get sectionSpacing => scaleByFontFactor(16.0);
```
jacob314, line 37: @kenzieschmoll any idea about this? the flame chart starts throwing negative padding value exceptions if you leave this as 16.0

packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -25,9 +25,16 @@ import '../ui/search.dart';
...
 const double sideInset = 70.0;
```
jacob314, line 38: @kenzieschmoll should I scale these two by the font size?

packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -741,7 +744,7 @@ class ToggleButton extends StatelessWidget {
         child: MaterialIconLabel(
           label: text,
           iconData: icon,
-          includeTextWidth: includeTextWidth,
```
jacob314, line 744: a perhaps poor judgement call was I changed to an ugly parameter name for this class so that the unscaled width could still be used.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -634,7 +634,7 @@ class LineItem extends StatefulWidget {
 
   static const _hoverDelay = Duration(milliseconds: 150);
   static const _removeDelay = Duration(milliseconds: 50);
-  static const _hoverWidth = 400.0;
+  static double get _hoverWidth => scaleByFontFactor(400.0);
```
jacob314, line 637: widths like this all should be claled because they are based on ensuring the size is large enough to fit a reasonable amount of text.

packages/devtools_app/lib/src/file_import.dart
```diff
@@ -99,13 +99,14 @@ class _FileImportContainerState extends State<FileImportContainer> {
   }
 
   Widget _buildImportFileRow() {
-    const rowHeight = 37.0;
+    // TODO(jacobr): why is this 37.0 instead of a more standard height?
+    final rowHeight = scaleByFontFactor(37.0);
```
jacob314, line 103: @kenzieschmoll any idea why we used this odd height?

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -75,8 +75,8 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
 
   static const summaryTreeKey = Key('Summary Tree');
   static const detailsTreeKey = Key('Details Tree');
-  static const includeTextWidth = 900.0;
-  static const includeRefreshTreeWidth = 1255.0;
+  static const unscaledIncludeTextWidth = 900.0;
```
jacob314, line 78: most of the time the width values we pass into widgets are all real pixels. For this case, there are some classes that are nicer to keep const so we pass in the unscaled version of the width instead.

packages/devtools_app/lib/src/charts/flame_chart.dart
kenzieschmoll: no - sideInset is the distance we leave from the left and right edges of the chart to the first/last flame chart nodes in the timeline. This is unaffected by font size.

packages/devtools_app/lib/src/file_import.dart
kenzieschmoll: I do not know. Feel free to change to a more standard height as long as we verify the layout still looks good

packages/devtools_app/lib/src/charts/flame_chart.dart
kenzieschmoll: negative padding in the X or Y direction?

packages/devtools_app/lib/src/banner_messages.dart
kenzieschmoll: can we add defaultFontSize to the theme so that any `TextStyle` without a `fontSize` specified gets `defaultFontSize` as its `fontSize`?

packages/devtools_app/lib/src/charts/chart.dart
kenzieschmoll: should we put this in a const in the theme?

packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -103,7 +103,7 @@ class IconLabelButton extends StatelessWidget {
 
   final String label;
 
-  final double includeTextWidth;
+  final double unscaledIncludeTextWidth;
```
kenzieschmoll, line 106: do we need to scale this instead of leaving it unscaled? I imagine a large font size would cause overflow errors if we don't also scale this value inversely to the font size.

packages/devtools_app/lib/src/debugger/hover.dart
```diff
@@ -91,9 +92,9 @@ class HoverCardData {
   HoverCardData({
     this.title,
     @required this.contents,
-    this.width = HoverCardTooltip.defaultHoverWidth,
+    double width,
     this.position = HoverCardPosition.cursor,
-  });
+  }) : width = width ?? HoverCardTooltip.defaultHoverWidth;
```
kenzieschmoll, line 97: this change is surprising - why is it neccesary?

packages/devtools_app/lib/src/screen.dart
```diff
@@ -158,8 +160,7 @@ abstract class Screen {
           // size for the [BadgePainter]
           final painter = TextPainter(
             text: TextSpan(
-              text: title,
-            ),
+                text: title, style: Theme.of(context).regularTextStyle),
```
kenzieschmoll, line 163: nit: tc

packages/devtools_app/lib/src/ui/icons.dart
```diff
@@ -68,8 +72,8 @@ class CircleIcon extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return Container(
-      width: 15,
-      height: 15,
+      width: defaultIconSize - 1,
```
kenzieschmoll, line 75: can you add a comment why subtracting 1 is necessary?

packages/devtools_app/lib/src/ui/icons.dart
```diff
@@ -78,7 +82,8 @@ class CircleIcon extends StatelessWidget {
       child: Text(
         text,
         textAlign: TextAlign.center,
-        style: const TextStyle(fontSize: 9, color: Color(0xFF231F20)),
+        style: TextStyle(
+            fontSize: scaleByFontFactor(9.0), color: const Color(0xFF231F20)),
```
kenzieschmoll, line 86: nit: trailing comma

packages/devtools_app/lib/src/ui/label.dart
```diff
@@ -5,11 +5,12 @@ import 'icons.dart';
 
 /// Label including an image icon and optional text.
 class ImageIconLabel extends StatelessWidget {
-  const ImageIconLabel(this.icon, this.text, {this.minIncludeTextWidth});
+  const ImageIconLabel(this.icon, this.text,
+      {this.unscaledMinIncludeTextWidth});
```
kenzieschmoll, line 9: nit: tc

packages/devtools_app/lib/src/ui/service_extension_widgets.dart
```diff
@@ -123,7 +123,8 @@ class _ServiceExtensionButtonGroupState
         selectedColor: theme.colorScheme.serviceExtensionButtonsTitleSelected,
         fillColor: theme.colorScheme.serviceExtensionButtonsFillSelected,
         textStyle: theme.textTheme.bodyText1,
-        constraints: const BoxConstraints(minWidth: 32.0, minHeight: 32.0),
+        constraints: BoxConstraints(
+            minWidth: defaultButtonHeight, minHeight: defaultButtonHeight),
```
kenzieschmoll, line 127: tc

kenzieschmoll (commented):
done reviewing

packages/devtools_app/lib/src/charts/flame_chart.dart
jacob314: -23.0 top padding (Y). topSpacer is computed as -23
I don't quite understand why rowOffsetForTopPadding is multiplied by sectionSpacing.

kenzieschmoll: This is surprising to me that this needs to be scaled at all. This is supposed to be the vertical spacing between sections of the flame chart, and should be independent of the size of the rows (which would be affected by font size).

`rowOffsetForTopPadding` is the number of rows we want to offset for padding at the top of the flame chart. For example, we need more rows of padding in the timeline flame chart that has the timestamp grid at the top than we need in the Cpu profiler flame chart, which doesn't have anything to account for at the top.

packages/devtools_app/lib/src/banner_messages.dart
jacob314: Hmm, I think the only way to do that is MediaQueryData.textScaleFactor
which might be a lot cleaner than what we are currently doing but we would need to be careful not to play badly with accessibility tools.

packages/devtools_app/lib/src/common_widgets.dart
jacob314: got a better idea for a the variable name? What I am trying to say is this is the width you would require with the default text scale, but you will need to scale this width if using a larger font.

packages/devtools_app/lib/src/debugger/hover.dart
jacob314: because the default value is now not a constant and you can only use constants as default values for Dart constants.

packages/devtools_app/lib/src/ui/icons.dart
jacob314: done

packages/devtools_app/lib/src/common_widgets.dart
kenzieschmoll: some questions: What is the consumer of this widget supposed to do? If the font size is arbitrarily set by the user, then the value passed into `unscaledIncludeTextWidth` would need to be a function of the font size, right? Is this function going to be consistent for all uses of this widget, or defined on a per use basis? How can the consumer of this widget guarantee they are passing in a size that will work for all font sizes?

packages/devtools_app/lib/src/charts/chart.dart
pq: This is really interesting. Normal constructors can be side effecting so we can never really treat them as dead but const constructors, I think you're right could be flagged.

packages/devtools_app/lib/src/banner_messages.dart
jacob314: Tested and it didn't fix this case and caused other problems so sticking with this approach.

packages/devtools_app/lib/src/charts/chart.dart
jacob314: done. const chartTextFontSize = 10.0;

packages/devtools_app/lib/src/charts/chart.dart
jacob314: Where should a bug to track this be filed?

packages/devtools_app/lib/src/charts/flame_chart.dart
jacob314: Filed https://github.com/flutter/devtools/issues/3306 to track. I'm not sure how to fix this.

packages/devtools_app/lib/src/theme.dart
```diff
@@ -160,22 +160,23 @@ const defaultElevation = 4.0;
...
 const densePadding = 4.0;
 
-const smallProgressSize = 12.0;
+double get smallProgressSize => scaleByFontFactor(12.0);
 
-const defaultListItemHeight = 28.0;
+double get defaultListItemHeight => scaleByFontFactor(28.0);
 
-const defaultChartHeight = 150.0;
+double get defaultChartHeight => scaleByFontFactor(150.0);
```
kenzieschmoll, line 167: I'm not sure we should be scaling this. For example, the flutter frames chart and the memory chart that use this value for their height do not have any text in them that would be affected by font factor

PR #3304, opened 2021-08-23
Cancel timeline polling timer when the vmService is closed

kenzieschmoll:
Without this call, the timeline polling timer is never cancelled when the vm service disconnects.


jacob314 (approved):
lgtm

PR #3309, opened 2021-08-27
Cache the results of `getVersion` and `getIsolate`

elliette:
Reduces the number of calls we make to `getVersion` and `getIsolate` in order to help improvement initial page load time. 

Previously we were calling both multiple times:

```
...
{jsonrpc: 2.0, method: getVM, id: 6}
{jsonrpc: 2.0, method: getIsolate, id: 7, params: {isolateId: 1}}
{jsonrpc: 2.0, method: getVersion, id: 9, params: {}}
{jsonrpc: 2.0, method: getVersion, id: 10, params: {}}
{jsonrpc: 2.0, method: getFlagList, id: 11, params: {}}
{jsonrpc: 2.0, method: getVM, id: 12, params: {}}
{jsonrpc: 2.0, method: getIsolate, id: 13, params: {isolateId: 1}}
{jsonrpc: 2.0, method: getIsolate, id: 14, params: {isolateId: 1}}
{jsonrpc: 2.0, method: streamListen, id: 15, params: {streamId: GC}}
{jsonrpc: 2.0, method: streamListen, id: 16, params: {streamId: Timeline}}
{jsonrpc: 2.0, method: streamListen, id: 17, params: {streamId: VM}}
{jsonrpc: 2.0, method: streamListen, id: 18, params: {streamId: Service}}
{jsonrpc: 2.0, method: getIsolate, id: 19, params: {isolateId: 1}}
{jsonrpc: 2.0, method: getScripts, id: 20, params: {isolateId: 1}}
{jsonrpc: 2.0, method: getObject, id: 21, params: {isolateId: 1, objectId: 400}}
...
```

After this change, we call `getVersion` only once, and `getIsolate` twice. The second `getIsolate` call comes from the `debugger_controller`.


packages/devtools_app/lib/src/vm_service_wrapper.dart
```diff
@@ -370,8 +371,9 @@ class VmServiceWrapper implements VmService {
   }
 
   @override
-  Future<Isolate> getIsolate(String isolateId) {
-    return trackFuture('getIsolate', _vmService.getIsolate(isolateId));
+  Future<Isolate> getIsolate(String isolateId) async {
```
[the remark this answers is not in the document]
jacob314: yeah feel free to flow the value through vmServiceOpened so none of the calls within it repeat fetching the isolate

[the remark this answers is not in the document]
elliette: Ok, I'm caching the value in `IsolateManager` and using it in `vmServiceOpened` (then clearing that value at the end of `vmServiceOpened`). However, it's possible I still might be overly caching - for example, as part of `vmServiceOpened`, the `_onMainIsolateChanged` method gets called, and I'm using the cached value there: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/service_manager.dart#L914. Wdyt?

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -40,12 +40,14 @@ const debugLogServiceProtocolEvents = false;
 
 const defaultRefreshRate = 60.0;
 
+typedef IsolateF = Future<Isolate> Function(String isolateId);
```
jacob314, line 43: Dart style: avoid abbreviations.https://dart.dev/guides/language/effective-dart/design#avoid-abbreviations

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -315,6 +317,9 @@ class ServiceConnectionManager {
     }
 
     _connectionAvailableController.add(service);
+
+    // Clear the isolates cache:
+    isolateManager.clearIsolateIdCache();
```
jacob314, line 322: why is this needed?

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -764,14 +779,16 @@ class IsolateManager extends Disposer {
...
-  ServiceExtensionManager(this._mainIsolate);
+  ServiceExtensionManager(this._mainIsolate, this._getIsolateCached);
 
   VmServiceWrapper _service;
 
   bool _checkForFirstFrameStarted = false;
 
   final ValueListenable<IsolateRef> _mainIsolate;
 
+  final IsolateF _getIsolateCached;
```
jacob314, line 790: why not just use the serviceManager isolate cache directly instead of passing in this function?

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -542,6 +547,8 @@ class IsolateManager extends Disposer {
   ValueListenable<IsolateRef> get mainIsolate => _mainIsolate;
   final _mainIsolate = ValueNotifier<IsolateRef>(null);
 
+  final _isolatesCache = <String, Isolate>{};
```
jacob314, line 550: there is already an isolateCache. Use it directly for cases where a cache is suitable rather than creating another one. I believe it is in IsolateManager.

elliette: It looks like the existing cache is being used to cache the isolate states, not the isolate itself. It never actually sends a request for the isolate[1], which we need to do. Though I don't understand how the existing cache is getting the isolate without requesting it explicitly, so I might be mistaken!

https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/service_manager.dart#L734-L738

packages/devtools_app/lib/src/service_manager.dart
elliette: Thanks! This typedef is now gone

packages/devtools_app/lib/src/service_manager.dart
elliette: I was mostly worried about clearing the cached isolate value because it was being used in `onMainIsolateChanged` (which was called on start up). I've now split out the registration logic from `onMainIsolateChanged` (because that seems to be why we were calling it on start up, to register the isolate).  So now:

- on start up, we call `registerIsolate` with the cached isolate
- when the isolate changes, `onMainIsolateChanged` is called, which calls `registerIsolate` with the new isolate

packages/devtools_app/lib/src/service_manager.dart
elliette: Now using the pre existing cache :)

packages/devtools_app/lib/src/service_manager.dart
elliette: Realized why the pre-existing cache wasn't working is because I was using it in `_loadIsolateState`. Now we are using the cache everywhere except `loadIsolateState`, which is where we make the request for the isolate.

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -903,21 +902,30 @@ class ServiceExtensionManager extends Disposer {
...
-    final Isolate isolate = await _service.getIsolate(isolateRef.id);
-    if (isolateRef != _mainIsolate.value) {
+    final isolateRef = _isolateManager.mainIsolate.value;
+    final Isolate isolate = await _isolateManager.getIsolateCached(isolateRef);
+
+    await _registerIsolate(isolate, isolateRef);
+  }
+
+  Future<void> _registerIsolate(Isolate isolate,
+      [IsolateRef previousIsolateRef]) async {
```
jacob314, line 918: in general avoid optional positional parameters. they just lead to bugs when they are sometimes used and sometimes not used. In general prefer optional named parameters as they provide more clarity about the intent of the parameters for cases where optional parameters are needed. not sure if this is in the style guide but is implicitly the case for flutter.

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -903,21 +902,30 @@ class ServiceExtensionManager extends Disposer {
...
     }
     _checkForFirstFrameStarted = false;
 
-    final isolateRef = _mainIsolate.value;
-    final Isolate isolate = await _service.getIsolate(isolateRef.id);
-    if (isolateRef != _mainIsolate.value) {
+    final isolateRef = _isolateManager.mainIsolate.value;
+    final Isolate isolate = await _isolateManager.getIsolateCached(isolateRef);
+
+    await _registerIsolate(isolate, isolateRef);
```
jacob314, line 914: call this _registerMainIsolate for clarity.

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -903,21 +902,30 @@ class ServiceExtensionManager extends Disposer {
...
+  }
+
+  Future<void> _registerIsolate(Isolate isolate,
```
jacob314, line 917: parameter name should not be previousIsolateRef. It should be expectedMainIsolateRef or similar.

jacob314 (approved):
lgtm
lgtm once the two minor nit comments are addressed.

PR #3315, opened 2021-08-30
Add a key set command for opening a file

elliette:
Adds `CTRL + P` / `CMD + P` as a key set, which will be used to open the file opener dialog. 

Also pulls the key sets definitions out into their own file within the debugger directory.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/key_sets.dart
+++ b/packages/devtools_app/lib/src/debugger/key_sets.dart
@@ -32,5 +32,5 @@ final LogicalKeySet openFileKeySet = LogicalKeySet(
   HostPlatform.instance.isMacOS
       ? LogicalKeyboardKey.meta
       : LogicalKeyboardKey.control,
-  LogicalKeyboardKey.keyP,
+  LogicalKeyboardKey.keyO,
 );
--- a/packages/devtools_app/lib/src/debugger/key_sets.dart
+++ b/packages/devtools_app/lib/src/debugger/key_sets.dart
@@ -28,6 +28,7 @@ final LogicalKeySet escapeKeySet = LogicalKeySet(
   LogicalKeyboardKey.escape,
 );
 
+// TODO(elliette): Change to cmd/ctrl + P once focus library filter is removed.
 final LogicalKeySet openFileKeySet = LogicalKeySet(
   HostPlatform.instance.isMacOS
       ? LogicalKeyboardKey.meta
--- a/packages/devtools_app/lib/src/debugger/key_sets.dart
+++ b/packages/devtools_app/lib/src/debugger/key_sets.dart
@@ -1,0 +1,36 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/debugger/key_sets.dart
```diff
@@ -0,0 +1,36 @@
...
+
+final LogicalKeySet escapeKeySet = LogicalKeySet(
+  LogicalKeyboardKey.escape,
+);
+
+final LogicalKeySet openFileKeySet = LogicalKeySet(
+  HostPlatform.instance.isMacOS
+      ? LogicalKeyboardKey.meta
+      : LogicalKeyboardKey.control,
+  LogicalKeyboardKey.keyP,
```
kenzieschmoll, line 35: this is the same as `focusLibraryFilterKeySet` above. Maybe use 'o' for this one for now? At least until @bkonyi's changes land and we have ripped out the filter in the libraries window. Then we can decide whether we want to have this hot key be cmd + p or cmd + o long term. I expect both the open file feature and Ben's changes will be done for the Oct release.

elliette: Oh good catch! Switched to cmd + o

bkonyi: CTRL-P is definitely preferable since most editors use that as a shortcut for file searching. Will it help if I upload a small PR to remove the filter box? It shouldn't be very difficult to do.

elliette: Sure, if that's not too much trouble that would be great! Thanks

bkonyi: https://github.com/flutter/devtools/pull/3319

elliette: Thank you!

kenzieschmoll: commented on Ben's PR, but will add a note here too - we shouldn't land #3319 until after the release goes out tomorrow. We don't want to remove the filtering functionality before we replace it with the open file dialog feature.

packages/devtools_app/lib/src/debugger/key_sets.dart
```diff
@@ -0,0 +1,36 @@
...
+      ? LogicalKeyboardKey.meta
+      : LogicalKeyboardKey.control,
+  LogicalKeyboardKey.keyF,
+);
+
+final LogicalKeySet escapeKeySet = LogicalKeySet(
+  LogicalKeyboardKey.escape,
+);
+
+final LogicalKeySet openFileKeySet = LogicalKeySet(
```
kenzieschmoll, line 31: add a TODO to change this to P once the library filter is removed.

kenzieschmoll (approved):

packages/devtools_app/lib/src/debugger/key_sets.dart
elliette: Done, thanks!

PR #3319, opened 2021-08-31
Remove filter text field from ScriptPicker

bkonyi:


elliette:
LGTM!

kenzieschmoll:
I'd like to land this after the release goes out tomorrow. We don't want to remove the library filtering functionality before we add a replacement feature (open file dialog) that allows a user to find a file.

PR #3320, opened 2021-08-31
Don't wait on stream listening in DevTools start up

elliette:
We are currently `awaiting` setting up listeners on all the streams in the `vmServiceOpened` method. This can take ~300 ms to complete. I don't think it's necessary for us to block on getting the listeners registered.

Note: I've also opened https://github.com/dart-lang/sdk/issues/47042 to investigate the weird throttling behavior I'm seeing on these requests.

### DIFF
--- a/packages/devtools_app/lib/src/service_manager.dart
+++ b/packages/devtools_app/lib/src/service_manager.dart
@@ -273,7 +273,7 @@ class ServiceConnectionManager {
       serviceStreamName,
     ];
 
-    await Future.wait(streamIds.map((String id) async {
+    unawaited(Future.wait(streamIds.map((String id) async {
       try {
         await service.streamListen(id);
       } catch (e) {
@@ -287,7 +287,7 @@ class ServiceConnectionManager {
           );
         }
       }
-    }));
+    })));
     if (service != this.service) {
       // A different service has been opened.
       return;


kenzieschmoll (approved):

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -273,7 +273,7 @@ class ServiceConnectionManager {
       serviceStreamName,
     ];
 
-    await Future.wait(streamIds.map((String id) async {
```
jacob314, line 276: remove the Future.wait as well. it isn't doing anything useful now that the result is unawaited.
This change could cause some tricky race conditions.
Main race condition: need to ensure that we are listening for isolate events before we fetch the isolate. Otherwise we could incorrectly believe the isolate is paused because we missed the resume event.
@bkonyi I know a lot of the streams now have history. If all the streams where race conditions matter are ones where we can fetch with history then we are safe.

elliette: Oh interesting! Unless I'm misunderstanding something, it looks like currently we are fetching the isolate before listening for the isolate events. (`await isolateManager.init(isolates)` is the call right before this call to start listening to the streams). Might make sense to send off the stream listen requests earlier.

Issue #2974, opened 2021-05-04
customDimensions for DartDevtools aren't showing up in analytics


jacob314:
They all show up as an empty list. I would expect 

Repro steps:
go/custom-dimensions-devtools-repro

terrylucas:
With the fix to the prompt work (done by Kenzie) analytics is now being sent out.  The original problem was initializeGA and the listener setup was not being called after the RemoteProvider work.

After properly setting up my machine for release using:

Builds against Flutter release used on bots:

```
> ./tool/build_release.sh

May have to update Flutter SDK, will get message to do that e.g.,
> ./tool/update_flutter_sdk.sh

> ./tool/build_release.sh
> dart packages/devtools/bin/devtools.dart
```

Now, connecting up and open up the Chrome dev console (with the gaDebug plugin installed will display:

```
       _                 _
  __ _| |_ __ _  __ _   (_)___
 / _` | __/ _` |/ _` |  | / __|
| (_| | || (_| | (_| |_ | \__ \
 \__, |\__\__,_|\__, (_)/ |___/
 |___/          |___/ |__/


Processing commands (0)
Processing commands (1)
Processing data layer push: {event: "gtm.init_consent"}
Processing data layer push: {event: "gtm.init"}
Processing data layer push: {event: "gtm.dom"}
Processing commands (1)
Processing data layer push: {event: "gtm.load"}
main.dart.js:4153 DevTools version 2.5.0.
DevTools failed to load source map: Could not load content for https://www.googletagmanager.com/debug/badge.css.map: HTTP error: status code 404, net::ERR_HTTP_RESPONSE_CODE_FAILURE
Processing commands (1)
Processing GTAG command: ["event", "landing", {event_category: "select", event_label: "connectToApp", value: 0, user_app: "", user_build: "", user_platform: "", devtools_platform: "MacIntel", devtools_chrome: "Chrome/92.0.4515.131", devtools_version: "2.5.0", ide_launched: "CLI", flutter_client_id: "1c91d630-095d-40af-8bf1-00e439622ead"}]
Processing commands (1)
Processing GTAG command: ["event", "inspector", {event_category: "screen", value: 0, user_app: "flutter", user_build: "debug", user_platform: "ios", devtools_platform: "MacIntel", devtools_chrome: "Chrome/92.0.4515.131", devtools_version: "2.5.0", ide_launched: "CLI", flutter_client_id: "1c91d630-095d-40af-8bf1-00e439622ead"}]
[...]

PR #3323, opened 2021-08-31
Fix analytics initialization

kenzieschmoll:
This PR simplifies the way we check whether analytics is enabled and ensures that we actually initialize gtags analytics each time DevTools is run (only when analytics are enabled).

This PR also adds a couple missing analytics to the landing screen as well as hot reload and hot restart actions (work towards https://github.com/flutter/devtools/issues/3270)

I have verified with GA Debug that I can now see custom dimensions coming over the wire with each GA call. Fixes https://github.com/flutter/devtools/issues/2974


packages/devtools_app/lib/main.dart
```diff
@@ -5,7 +5,7 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 
-import 'src/analytics/stub_provider.dart'
+import 'src/analytics/provider.dart'
```
jacob314, line 8: can this be cleaned up to match our other conditional imports with
src/analytics/provider.dart internally using conditional imports so that call sites like this don't have to remember to add the conditional imports. Would be easy to accidentally only import 'src/analytics/provider.dart' and end up with stubbed out analytics on html.

packages/devtools_app/lib/src/analytics/analytics.dart
```diff
@@ -245,27 +235,17 @@ class GtagExceptionDevTools extends GtagException {
...
-
-// Exposed function to JS via allowInterop.
-bool gaEnabled() => _gaEnabledNotifier.value;
-
 /// Request DevTools property value 'enabled' (GA enabled) stored in the file
 /// '~/.flutter-devtools/.devtools'.
 Future<bool> isAnalyticsEnabled() async {
-  _gaEnabledNotifier.value = await server.isAnalyticsEnabled();
-  return _gaEnabledNotifier.value;
+  print('in analytics isAnalyticsEnabled');
```
jacob314, line 241: remove all print statements before commit or change to logging.

packages/devtools_app/lib/src/analytics/prompt.dart
```diff
@@ -4,44 +4,45 @@
...
+import 'package:provider/provider.dart';
 
 import '../common_widgets.dart';
 import '../config_specific/launch_url/launch_url.dart';
 import '../theme.dart';
 import 'provider.dart';
 
 /// Conditionally displays a prompt to request permission for collection of
-/// usage analytics.
+/// usage analytics and manages initializing GTags analytics.
```
jacob314, line 15: seems fragile to have the prompt responsible for initializing gtags. can we ignore that?

packages/devtools_app/lib/src/analytics/provider.dart
```diff
@@ -2,11 +2,15 @@
...
+
+class AnalyticsProvider {
+  final ValueListenable<bool> analyticsEnabled = ValueNotifier<bool>(false);
+  final ValueListenable<bool> shouldPrompt = ValueNotifier<bool>(false);
+  void enableAnalytics() {}
+  void disableAnalytics() {}
+  void setUpAnalytics() {}
 }
+
+Future<AnalyticsProvider> get analyticsProvider async => _provider;
```
jacob314, line 15: getter seems async for no reason.

packages/devtools_app/lib/src/analytics/remote_provider.dart
```diff
@@ -4,74 +4,65 @@
...
-  bool get isGtagsEnabled => _isGtagsEnabled;
-  final bool _isGtagsEnabled;
+  bool analyticsInitialized = false;
 
   @override
-  void setAllowAnalytics() {
-    setUpAnalytics();
-    analytics.setAllowAnalytics();
+  Future<void> enableAnalytics() async {
+    _analyticsEnabled.value = await analytics.enableAnalytics();
```
jacob314, line 32: do you want to set
_analyticsEnabled.value = true;
before the async call completes?

packages/devtools_app/lib/src/analytics/remote_provider.dart
```diff
@@ -4,74 +4,65 @@
...
   }
 
   @override
-  void setDontAllowAnalytics() {
-    analytics.setDontAllowAnalytics();
+  Future<void> disableAnalytics() async {
+    _analyticsEnabled.value = await analytics.disableAnalytics();
+    _shouldPrompt.value = false;
```
jacob314, line 38: Seems like we could set 
_shouldPrompt.value = false; before awaiting the async operation. Defensively could avoid cases where the prompt was shown multiple times.

packages/devtools_app/test/scaffold_test.dart
```diff
@@ -158,22 +159,25 @@ void main() {
...
-        DevToolsScaffold(
-          tabs: const [
-            _TestScreen(
-              DebuggerScreen.id,
-              debuggerScreenKey,
-              tabKey: debuggerTabKey,
-              showFloatingDebuggerControls: false,
+      await tester.pumpWidget(
+        wrapWithAnalyticsProvider(
+          wrapWithControllers(
```
jacob314, line 164: can you make wrapWithControllers support the analytics providers to reduce the boilerplate in writing these tests?

packages/devtools_app/web/devtools_analytics.js
```diff
@@ -2,79 +2,44 @@
...
-        'metric2': 'ui_duration_micros',
-        'metric3': 'shader_compilation_duration_micros',
-        'metric4': 'cpu_sample_count',
-        'metric5': 'cpu_stack_depth',
-      }
-    });
-
-    _initializedGA = true;
-  }
+  console.log('initializing ga from inside devtools_analytics.js');
```
jacob314, line 13: remove console log before commit

packages/devtools_app/lib/src/analytics/analytics.dart
kenzieschmoll: done.

packages/devtools_app/lib/main.dart
kenzieschmoll: done - here and with the analytics.dart conditional files as well

packages/devtools_app/lib/src/analytics/provider.dart
kenzieschmoll: this is the stub version of this getter, but the real html getter needs to be async. Removed `async` and changed return type to `FutureOr` so that we can still call `await analyticsProvider` in the code.

packages/devtools_app/lib/src/analytics/remote_provider.dart
kenzieschmoll: done

packages/devtools_app/test/scaffold_test.dart
kenzieschmoll: done

packages/devtools_app/web/devtools_analytics.js
kenzieschmoll: done

packages/devtools_app/web/devtools_analytics.js
```diff
@@ -2,79 +2,44 @@
...
-      if (window.gaDevToolsEnabled()) {
-        gtag('event', document.visibilityState, {
-          event_category: 'application',
-        });
-      }
-    });
-  }
+  // Record when DevTools browser tab is selected (visible), not selected (hidden) or browser minimized.
+  document.addEventListener('visibilitychange', function (e) {
+    if (window.gaDevToolsEnabled()) {
```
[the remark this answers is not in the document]
kenzieschmoll: done

packages/devtools_app/lib/src/analytics/remote_provider.dart
kenzieschmoll: done

packages/devtools_app/lib/src/analytics/prompt.dart
kenzieschmoll: moved the initialization code into the provider (now renamed to controller to be consistent with DevTools naming as a whole)

packages/devtools_app/lib/src/analytics/analytics.dart
```diff
@@ -1,640 +1,5 @@
...
-import '../globals.dart';
-import '../performance/performance_screen.dart';
-import '../profiler/profiler_screen.dart';
-import '../ui/gtags.dart';
-import '../version.dart';
-import 'analytics_common.dart';
-import 'constants.dart' as analytics_constants;
-
-// Dimensions1 AppType values:
-const String appTypeFlutter = 'flutter';
```
jacob314, line 31: so none of these strings needed to be in the public api?

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/analytics/analytics.dart
kenzieschmoll: this file was just moved, not deleted - the git diff is a little off. see _analytics_web.dart

PR #3310, opened 2021-08-27
Color stack frames by categories: native, dart core, flutter core

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/578

Screen Shot 2021-08-27 at 10 57 42 AM
Screen Shot 2021-08-27 at 10 52 55 AM
Note: user code and code from outside of native and dart/flutter core
Screen Shot 2021-08-27 at 10 57 05 AM
Light theme:
Screen Shot 2021-08-27 at 10 58 02 AM

@InMatrix any UX input on colors? I grabbed the flutter color and the color for the "other" category from the flutter brand guidelines deck.

PR #3324, opened 2021-08-31
Update CPU profiler colors to make them accessible

kenzieschmoll:
Color changes per UX guidance in discussion on https://github.com/flutter/devtools/pull/3310

This required some refactoring to support theme dependent colors where `context` is not available (`initState`).

### DIFF
--- a/packages/devtools_app/lib/src/ui/utils.dart
+++ b/packages/devtools_app/lib/src/ui/utils.dart
@@ -9,6 +9,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter/scheduler.dart';
 
 import '../auto_dispose_mixin.dart';
+import '../theme.dart';
 
 /// Stateful Checkbox Widget class using a [ValueNotifier].
 ///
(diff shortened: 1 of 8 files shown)


packages/devtools_app/lib/src/ui/utils.dart
```diff
@@ -227,3 +228,39 @@ class ColorPair {
...
+  const ThemedColor({this.light, this.dark});
+
+  factory ThemedColor.fromSingle(Color color) =>
+      ThemedColor(light: color, dark: color);
+
+  final Color light;
+
+  final Color dark;
+
+  Color colorFor(BuildContext context) {
```
jacob314, line 263: this method should take the theme or colorScheme as a parameter not a context.
That way you can lookup the theme once at the top of a build method and use it multiple places.
Theme.of calls are relatively expensive.

packages/devtools_app/lib/src/ui/utils.dart
```diff
@@ -227,3 +228,39 @@ class ColorPair {
...
+/// A theme-dependent color.
+///
+/// When possible, themed colors should be specified in an extension on
+/// [ColorScheme] using the [ColorScheme.isLight] getter. However, this class
+/// may be used when access to the [BuildContext] is not available at the time
+/// the color needs to be specified.
+class ThemedColor {
+  const ThemedColor({this.light, this.dark});
+
+  factory ThemedColor.fromSingle(Color color) =>
```
jacob314, line 256: this should be a const constructor as well which will enable a lot more of the ThemedColor to be const as well.

kenzieschmoll: factory constructors can't be const since Dart requires that const constructors have no body

kenzieschmoll: Ah nevermind - I can make this const and remove the factory

packages/devtools_app/lib/src/ui/utils.dart
kenzieschmoll: done

PR #3325, opened 2021-08-31
Add service worker logic from sample Flutter web app

elliette:
Copies over the service worker logic from the sample Flutter web app to cache `main.js`

### DIFF
--- a/packages/devtools_app/web/index.html
+++ b/packages/devtools_app/web/index.html
@@ -10,7 +10,6 @@
 <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
 
     <title></title>
     <link href="favicon.png" rel="icon" sizes="64x64">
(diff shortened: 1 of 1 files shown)


kenzieschmoll:
@jacob314 more of a general question of if we want to do this or not? I know at some point we removed the service worker code intentionally but I don't remember why.

jacob314 (approved):
We should test how this works after you make a local version of DevTools. We should also test whether it works if you change your port. There might or might not be a way for a service worker to still help in that case.

elliette:
Converting this back to draft until I figure out what is going on when running this from g3. Unfortunately this actually decreases IPL in that case, since there is a timeout to load the backup script tag if the service worker registration fails.

jacob314:
It would be nice to land this and apply a workaround to g3.

elliette:
> It would be nice to land this and apply a workaround to g3.

Sounds good! I'll try to get this landed this week. I always need to create the actual service worker itself first.

packages/devtools_app/web/index.html
```diff
@@ -55,17 +54,69 @@
...
+              waitForActivation(reg.installing);
+            } else {
+              // Existing service worker is still good.
+              loadMainDartJs();
+            }
+          });
+
+        // If service worker doesn't succeed in a reasonable amount of time,
+        // fallback to plain <script> tag.
+        setTimeout(() => {
```
jacob314, line 107: When might this occur?

tool/update_version.dart
```diff
@@ -41,6 +51,10 @@ void main(List<String> args) async {
   print('Updating CHANGELOG to version $version...');
   writeVersionToChangelog(File('packages/devtools/CHANGELOG.md'), version);
 
+  print('Updating index.html to version $version...');
+  writeVersionToIndexHtml(
```
jacob314, line 55: Glad to see you are updating the version automatically!

tool/update_version.dart
```diff
@@ -140,6 +154,28 @@ void writeVersionToChangelog(File changelog, String version) {
...
 
+void writeVersionToIndexHtml(
+  File indexHtml,
+  String oldVersion,
+  String newVersion,
+) {
+  final lines = indexHtml.readAsLinesSync();
+  final revisedLines = <String>[];
+  for (final line in lines) {
+    if (line.contains(oldVersion)) {
```
jacob314, line 165: make this more robust. If the file doesn't contain the oldVersion we need to throw an exception as we failed to update.

Issue #3279, opened 2021-08-14
Fixed - [Network] HTTP Requests with error are shown as 'Pending'


bleroux:
When an HTTP request failed (HandshakeException, SocketException, ...) :
- it is shown as 'pending' in the requests list
- when selected, details pane is grey

![image](https://user-images.githubusercontent.com/840911/129445817-95b5483e-7f8b-4a48-9c5a-c781b715cd7f.png)

Issue #3260, opened 2021-08-10
Fixed - Grey panels when selecting some HTTP Requests


bleroux:
When selecting an HTTP Requests, 'Overview', 'Header' and 'Response' tabs sometimes display a grey panel (or a red one when launching DevTools in debug mode).

![ReleaseMode](https://user-images.githubusercontent.com/840911/128894364-0e227f05-de03-4d81-96df-3218302e50af.png)

This is at least true for the following requests :
- request to a non-existing domain
- request to initiate a WebSocket to ```echo.websocket.org```
 
Underlying problem is that there are several calls to nullable fields. Existing tests (```network_profiler_test.dart```) don't cover those cases because they rely on HTTP Requests collected from the timeline and not on ones collected by the new ```HttpProfile```

@kenzieschmoll Kenzie, if you want you can assign this issue to me. I will try to make a PR with :
- a new group in network_profiler_test.dart that will turn on HttpProfile
- fixes in UI and non UI code (DartIOHttpRequestData)

kenzieschmoll:
Sure @bleroux, thanks! Go ahead and add me as a reviewer when you are ready.

PR #3328, opened 2021-09-01
Fix grey panels when selecting some HTTP Requests

bleroux:
- Update network tests to cover HTTP requests collected by the new HttpProfiler 
- Add two HTTP requests to test data : a failed HTTP request and a GET request to initialize a WebSocket connection (handshake)
- Fix issues described in #3260 #3279


packages/devtools_app/test/network_controller_test.dart
```diff
@@ -335,107 +335,48 @@ void main() {
...
+        expect(request.general.length, greaterThan(0));
+        expect(request.hasCookies, isNotNull);
+        expect(request.inProgress, false);
+        expect(request.instantEvents, isNotNull);
+        expect(httpMethods.contains(request.method), true);
+        expect(request.requestCookies, isNotNull);
+        expect(request.responseCookies, isNotNull);
+        expect(request.startTimestamp, isNotNull);
+        expect(request.status, isNotNull);
+        expect(request.uri, isNotNull);
```
kenzieschmoll, line 372: looks like the checks for request and response bodies have been removed - please add those back

packages/devtools_app/test/network_profiler_test.dart
```diff
@@ -268,32 +261,6 @@ void main() {
       await clearTimeouts(tester);
     });
 
-    // Regression test for https://github.com/flutter/devtools/issues/3286.
-    testWidgetsWithWindowSize('can select by clicking on url', windowSize,
```
kenzieschmoll, line 272: why was this regression test removed?

packages/devtools_app/test/network_controller_test.dart
bleroux: I copied those lines from Timeline tests (same file, lines 116-128), there were no check for request and response bodies. 

Request and response bodies are tested in network_model_test.dart  (line 792-808).

packages/devtools_app/test/network_profiler_test.dart
bleroux: Good catch !
My bad : I started working on this PR before your PR #3287 was merged (the one adding this regression test) and I didn't notice this when merging...

I need some guidance to properly merge our changes on this file.
In my PR, I kept the old tests (timeline) and created a new group for HttpProfile tests (I didn't want to remove the old ones before asking, but didn't like to have to duplicate test code).
In your PR, you "upgraded" the old tests from Timeline profiler to Http profiler.

I really prefer your way, as Timeline profiling for HTTP requests seems to be obsolete now. 
Can I assume this ? (if so I could throw away my changes on this file)

kenzieschmoll: yes it is safe to "upgrade" from timeline to the http profile impl. Feel free to do that here.

kenzieschmoll (approved):

PR #3333, opened 2021-09-02
Don't wait on stream listening in DevTools start up 

elliette:
Follow-up to https://github.com/flutter/devtools/pull/3320 (reverted in https://github.com/flutter/devtools/pull/3321).


elliette:
Made some changes from the original in response to https://github.com/flutter/devtools/pull/3320#discussion_r699602505:

* Removed the `Future.wait`
* Moved the stream listening up so that it happens before initializing the isolates

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -286,7 +273,21 @@ class ServiceConnectionManager {
           );
         }
       }
-    }));
+    }
```
kenzieschmoll, line 276: if we are worried about missing isolate events, what if we only await listening to streams that we are concerned about? perhaps kIsolate and kVM?

kenzieschmoll: cc @jacob314 for input on this

jacob314: Ok thought about this a bit more. If the VM is handling our requests in order we don't need to await any of these as long as we ensure that we send the request over the VM Service for the streams before we preform operations getting the current states of the streams. I would double check that we aren't loading isolate objects before the stream listen code and we should be good.

bkonyi: I'm not sure if it matters in this case, but there's no guarantee that the VM service will finish processing a request before a more recent request is completed. Obviously this won't be an issue if you're the only client and you're awaiting on each service request, but just something to keep in mind.

elliette: Confirmed that we are triggering the requests for streams before loading any isolate objects

jacob314:
lgtm

PR #3336, opened 2021-09-02
Restructure analytics code so that the AnalyticsController can be tested

kenzieschmoll:
This allows the core functionality of AnalyticsController to be used for desktop and web platforms, and breaks out the platform specific code into callbacks. This also allows for better testing in that we can actually use a real controller in tests instead of a fake controller that was a copy of the web implementation. This was fragile and easy to fall out of date with the real controller code.


jacob314 (approved):
lgtm
This is much more testable!

PR #3337, opened 2021-09-02
Add is_external_build dimension to analytics

kenzieschmoll:
This will allow us to remove the fake 'google3' screen internally. The isExternalBuild const is set to false internally, so this will just work when this change is rolled in.


jacob314 (approved):
lgtm

PR #3339, opened 2021-09-03
Fix some state management issues in the inspector

kenzieschmoll:
We were trying to mark an ancestor widget as needing to build in the middle of a build phase. This PR avoid making these setState calls during the build phase of the inspector.


packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -83,8 +83,11 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
     super.initState();
     ga.screen(InspectorScreen.id);
 
-    autoDispose(
-        serviceManager.onConnectionAvailable.listen(_handleConnectionStart));
+    autoDispose(serviceManager.onConnectionAvailable.listen((service) {
+      setState(() {
```
jacob314, line 87: this refactor seems to be in the wrong direction. setState calls should be as close as to where state is set as possible. what is the motivation for doing this?

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -281,10 +284,8 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
   }
 
   void _handleConnectionStart(VmService service) async {
```
jacob314, line 286: this method seems like it doesn't need to be async anymore. probably the cleanest solution is to remove the async and leave the setState where it was.

packages/devtools_app/lib/src/inspector/inspector_screen.dart
kenzieschmoll: _handleConnectionStart is called immediately below (line 92) which is in initState. We should not be calling `setState` from `initState` and that is what was causing an exception to be thrown. 

We were also calling `setState` twice in `_handleConnectionStart`, which contained only synchronous code, so it seemed unnecessary to call it twice instead of once for the entire method.

jacob314: calling _handleConnectionStart immediately in initState is probably the real bug. I would suggest we defer it to run after the frame or figure out how to refactor further so initState isn't triggering controller business logic like this.

kenzieschmoll: Did some refactoring PTAL.

packages/devtools_app/lib/src/inspector/inspector_controller.dart
```diff
@@ -133,6 +133,34 @@ class InspectorController extends DisposableController
...
+      serviceManager.onConnectionClosed.listen(_handleConnectionStop),
+    );
+  }
+
+  void _handleConnectionStart(VmService service) {
+    // Clear any existing badge/errors for older errors that were collected.
+    // Do this in a post frame callback so that we are not trying to clear the
+    // error notifiers for this screen while the framework is already in the
+    // process of building widgets.
+    WidgetsBinding.instance.addPostFrameCallback((_) {
```
jacob314, line 153: can you remove this addPostFrameCallback now that this is in the controller? I would have thought its lifecycle was different from State objects so this wouldn't be needed.

jacob314: nm. maybe add a TODO that when the inspectorController isn't called from 
createState() this can be removed.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/inspector/inspector_controller.dart
kenzieschmoll: We still need this in a post frame callback because it triggers a value listenable builder elsewhere in the widget tree that is already building. Added TODO in https://github.com/flutter/devtools/pull/3347

PR #3340, opened 2021-09-03
Fix network_model_test.dart breaking the bots

kenzieschmoll:


elliette (approved):

bleroux:
@kenzieschmoll By curiosity, why previous code broke the bots ?

kenzieschmoll:
The hard coded data for the http patch response body was throwing an exception when being json decoded. I'm not sure why the bots were passing before this change, but they are back to green now.

PR #3350, opened 2021-09-09
Enable the file opener

elliette:
Enables the file opener dialog. Since it's now enabled, also changes the text when no file is opened to specify how to open a new file.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -32,8 +32,6 @@ import 'file_search.dart';
 import 'hover.dart';
 import 'variables.dart';
 
-const openFileDialogEnabled = false;
-
 final debuggerCodeViewSearchKey =
     GlobalKey(debugLabel: 'DebuggerCodeViewSearchKey');
 
@@ -172,7 +170,7 @@ class _CodeViewState extends State<CodeView>
     if (scriptRef == null) {
       return Center(
         child: Text(
-          'No script selected',
+          'Open a file: ⌘ P',
           style: theme.textTheme.subtitle1,
         ),
       );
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -172,7 +170,7 @@ class _CodeViewState extends State<CodeView>
     if (scriptRef == null) {
       return Center(
         child: Text(
-          'No script selected',
+          'Open a file: ⌘ P',
           style: theme.textTheme.subtitle1,
```
jacob314, line 174: nice polish. One thing... we need to show the correct shortcut on both mac, linux, and windows so add a helper so we generate `control` instead of `⌘` on linux and windows. @kenzieschmoll probably knows if we already have a helper like that. There are a few other places where we show keyboard shortcuts in the UI

kenzieschmoll: Yes we have a helper for this: https://github.com/flutter/devtools/blob/a05828770ebfcf9af19384c5f4b3f4135b80675d/packages/devtools_app/lib/src/utils.dart#L951

kenzieschmoll (approved):

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Oh nice! Switched to using that. One thing I noticed though is that the special character ⌘ isn't rendered properly when DevTools is running on the web. 

image

This seems to be a known Flutter issue, and happens when Canvaskit is the renderer:

* https://github.com/flutter/flutter/issues/79172
* https://github.com/flutter/flutter/issues/77465
* https://github.com/flutter/flutter/issues/76473

Would it make sense to use "cmd" instead of ⌘ on web until the rendering issue is fixed?

kenzieschmoll: Yes, let's do that. Add a todo and link to the issue that will unblock this. Can you also file an issue on our repo to make sure we switch this back once the blockers are fixed? thanks

PR #3351, opened 2021-09-10
Revert "Don't wait on stream listening in DevTools start up "

elliette:
Reverts flutter/devtools#3333

Need to investigate failure, but this assertion that the isolate is not null is failing: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/service_manager.dart#L466


jacob314:
lgtm

Issue #3352, opened 2021-09-10
Special character ⌘ renders incorrectly on web


elliette:
![image](https://user-images.githubusercontent.com/21270878/132905009-141ce7fd-3baa-4542-8ebd-4a1448454539.png)

This seems to be a known Flutter issue, and happens when Canvaskit is the renderer:

* https://github.com/flutter/flutter/issues/79172
* https://github.com/flutter/flutter/issues/77465
* https://github.com/flutter/flutter/issues/76473

One solution is to use "cmd" until this is resolved in Flutter.

sperochon:
Same issue here.

The fix is to use a font that includes this special character as indicated in the logs:
`Could not find a set of Noto fonts to display all missing characters. Please add a font asset for the missing characters. See: https://flutter.dev/docs/cookbook/design/fonts`

I've used 'Inter' font and it is now displaying fine.
https://fonts.google.com/specimen/Inter?query=inter&preview.text=⌘&preview.text_type=custom

elliette:
Fixed in https://github.com/flutter/devtools/pull/3353

PR #3353, opened 2021-09-10
Use "Command" instead of special character ⌘ on web

elliette:
See https://github.com/flutter/devtools/issues/3352 for details

Temporary workaround until rendering issue is fixed in Flutter.

### DIFF
--- a/packages/devtools_app/lib/src/utils.dart
+++ b/packages/devtools_app/lib/src/utils.dart
@@ -966,7 +966,8 @@ extension LogicalKeySetExtension on LogicalKeySet {
     return sortedKeys.map((key) {
       if (_modifiers.contains(key)) {
         if (isMacOS && key == LogicalKeyboardKey.meta) {
-          return '⌘';
+          // TODO(issue #3352) Switch back to using ⌘ once supported on web.
+          return kIsWeb ? 'Command-' : '⌘';
         }
         return '${_modifierNames[key]}-';
       } else {
--- a/packages/devtools_app/lib/src/utils.dart
+++ b/packages/devtools_app/lib/src/utils.dart
@@ -966,7 +966,7 @@ extension LogicalKeySetExtension on LogicalKeySet {
     return sortedKeys.map((key) {
       if (_modifiers.contains(key)) {
         if (isMacOS && key == LogicalKeyboardKey.meta) {
-          // TODO(issue #3352) Switch back to using ⌘ once supported on web.
+          // TODO(elliette) Switch back to using ⌘ once supported on web.
           return kIsWeb ? 'Command-' : '⌘';
         }
         return '${_modifierNames[key]}-';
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/utils.dart
```diff
@@ -966,7 +966,8 @@ extension LogicalKeySetExtension on LogicalKeySet {
     return sortedKeys.map((key) {
       if (_modifiers.contains(key)) {
         if (isMacOS && key == LogicalKeyboardKey.meta) {
-          return '⌘';
+          // TODO(issue #3352) Switch back to using ⌘ once supported on web.
```
jacob314, line 969: See other comments in DevTools for examples of how we link to issues in todos.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -966,7 +966,8 @@ extension LogicalKeySetExtension on LogicalKeySet {
...
+          return kIsWeb ? 'Command-' : '⌘';
```
kenzieschmoll, line 970: will there be a space on the other side of this before the hotkey? e.g. `Command- P`, or will it be `Command-P`? Either way, we should have a consistent number of spaces on either side of this dash. Also, slight preference for `cmd` to be more similar to how we write an abbreviated `ctrl`. If we use `cmd`, I don't think the dash is necessary.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -966,7 +966,8 @@ extension LogicalKeySetExtension on LogicalKeySet {
     return sortedKeys.map((key) {
       if (_modifiers.contains(key)) {
         if (isMacOS && key == LogicalKeyboardKey.meta) {
-          return '⌘';
+          // TODO(elliette) Switch back to using ⌘ once supported on web.
```
kenzieschmoll, line 969: When there is a specific bug, you can either write `TODO(elliette): some description. See <github link to issue>` Or you can write the todo specifically attached to the issue link `TODO(<github link to issue): use ⌘ instead of command`

packages/devtools_app/lib/src/utils.dart
elliette: No space before the "P", it looks like this:

image

I chose `Command` to be consistent with how we are showing the modifier names here: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/utils.dart#L943-L948, but happy to change to `cmd` instead. Wdyt?

packages/devtools_app/lib/src/utils.dart
elliette: Done, thanks!

packages/devtools_app/lib/src/utils.dart
elliette: Great, thanks!

packages/devtools_app/lib/src/utils.dart
kenzieschmoll: Oh, I thought we were using `ctrl`. In that case, lgtm on what you have!

kenzieschmoll (approved):

packages/devtools_app/lib/src/utils.dart
kenzieschmoll: You may need to skip the 'meta mac' test for web. See other examples of this `skip: kIsWeb`. If you skip that test add a TODO link to stop skipping that links to the issue you filed

PR #3354, opened 2021-09-10
Moves the file opener out of a dialog

elliette:
Moves the file opener out of a dialog so that it shows up at the top of the codeview. It now works similarly to the "search in file" field. 

new file opener

Material dialogs can't be positioned anywhere other than center, so this gives us more flexibility with the positioning.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -32,7 +32,7 @@ import 'file_search.dart';
 import 'hover.dart';
 import 'variables.dart';
 
-const openFileDialogEnabled = false;
+const openFileDialogEnabled = true;
 
 final debuggerCodeViewSearchKey =
     GlobalKey(debugLabel: 'DebuggerCodeViewSearchKey');
@@ -167,6 +167,28 @@ class _CodeViewState extends State<CodeView>
 
   @override
   Widget build(BuildContext context) {
+    return ValueListenableBuilder(
+        valueListenable: widget.controller.showFileOpener,
+        builder: (context, showFileOpener, _) {
+          return Stack(
+            children: [
+              buildCodeviewContent(context),
+              if (showFileOpener)
+                Positioned(
+                  // top: denseSpacing,
+                  left: 110.0,
+                  child: wrapInElevatedCard(
+                    FileSearchField(
+                      controller: widget.controller,
+                    ),
+                  ),
+                ),
+            ],
+          );
+        });
+  }
+
+  Widget buildCodeviewContent(BuildContext context) {
     final theme = Theme.of(context);
 
     if (scriptRef == null) {
@@ -377,7 +399,7 @@ class _CodeViewState extends State<CodeView>
     );
   }
 
-  Widget buildSearchInFileField() {
+  Widget wrapInElevatedCard(Widget widget) {
     return Card(
       elevation: defaultElevation,
       color: Theme.of(context).scaffoldBackgroundColor,
@@ -388,14 +410,20 @@ class _CodeViewState extends State<CodeView>
         width: wideSearchTextWidth,
         height: defaultTextFieldHeight + 2 * denseSpacing,
         padding: const EdgeInsets.all(denseSpacing),
-        child: buildSearchField(
-          controller: widget.controller,
-          searchFieldKey: debuggerCodeViewSearchKey,
-          searchFieldEnabled: parsedScript != null,
-          shouldRequestFocus: true,
-          supportsNavigation: true,
-          onClose: () => widget.controller.toggleSearchInFileVisibility(false),
-        ),
+        child: widget,
+      ),
+    );
+  }
+
+  Widget buildSearchInFileField() {
+    return wrapInElevatedCard(
+      buildSearchField(
+        controller: widget.controller,
+        searchFieldKey: debuggerCodeViewSearchKey,
+        searchFieldEnabled: parsedScript != null,
+        shouldRequestFocus: true,
+        supportsNavigation: true,
+        onClose: () => widget.controller.toggleSearchInFileVisibility(false),
       ),
     );
   }
@@ -1083,19 +1111,14 @@ const goToLineOption = ScriptPopupMenuOption(
   onSelected: showGoToLineDialog,
 );
 
-void showOpenFileDialog(BuildContext context, DebuggerController controller) {
-  if (openFileDialogEnabled) {
-    showDialog(
-      context: context,
-      builder: (context) => OpenFileDialog(controller),
-    );
-  }
+void showFileOpener(BuildContext context, DebuggerController controller) {
+  controller.toggleFileOpener(true);
 }
 
 const openFileOption = ScriptPopupMenuOption(
   label: 'Open file (⌘ P)',
   icon: Icons.folder_open,
-  onSelected: showOpenFileDialog,
+  onSelected: showFileOpener,
 );
 
 class GoToLineDialog extends StatelessWidget {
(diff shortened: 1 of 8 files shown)


kenzieschmoll:
We may want to change the text field label to "Open" instead of "Search" for this use case

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -166,6 +166,27 @@ class _CodeViewState extends State<CodeView>
...
   Widget build(BuildContext context) {
+    return ValueListenableBuilder(
+        valueListenable: widget.controller.showFileOpener,
+        builder: (context, showFileOpener, _) {
+          return Stack(
+            children: [
+              buildCodeviewContent(context),
+              if (showFileOpener)
+                Positioned(
+                  left: 110.0,
```
kenzieschmoll, line 177: prefer consts with a descriptive name over "magic" numbers. For example, see another const in this class `searchFieldRightPadding`

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -166,6 +166,27 @@ class _CodeViewState extends State<CodeView>
 
   @override
   Widget build(BuildContext context) {
+    return ValueListenableBuilder(
+        valueListenable: widget.controller.showFileOpener,
+        builder: (context, showFileOpener, _) {
+          return Stack(
```
kenzieschmoll, line 172: this will result in two Stack widgets. One here and one as part of buildCodeviewContent. Instead of adding the buildCodeviewContent helper, leave this build method as it was, and you can add your ValueListenableBuilder and extra Positioned widget directly to the old build method.

We should also make sure that both the file opener and the search in file fields cannot be showing at the same time. Opening one of the two while the other is open should automatically close the other field. Make sure to try this out with an active search query in both.

kenzieschmoll:
we should also make sure that the file name is centered vertically in each row of of the autocomplete dialog

packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -1018,10 +1018,22 @@ class DebuggerController extends DisposableController
...
   }
 
+  void hideSearchInFileIfVisible() {
+    if (showSearchInFileField.value) {
+      toggleSearchInFileVisibility(false);
+    }
+  }
+
+  void hideFileOpenerIfVisible() {
+    if (showFileOpener.value) {
```
kenzieschmoll, line 1032: ValueNotifier will not notify listeners if the new value is the same as the current value, so there is no need to add extra methods `hideSearchInFileIfVisible` and `hideFileOpenerIfVisible` to check the current state of the respective value notifiers. You can just call the `toggleSearchInFileVisibility` and `toggleFileOpenerVisibility` directly where you are calling these hide methods

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -170,6 +170,8 @@ class AutoCompleteState extends State<AutoComplete> with AutoDisposeMixin {
 
     // Approximation but it's pretty accurate. Could consider using a layout builder
     // or maybe build in an overlay (that's isn't visible) to compute.
+    // TODO(https://github.com/flutter/devtools/issues/3075) This approximation
+    // doesn't work for the new file opener, should compute the actual height.
     final tileEntryHeight = box.size.height;
```
kenzieschmoll, line 175: Instead of doing the work to compute the height of the tiles, can you just make the tileEntryHeight configurable by a parameter on the `AutoComplete` widget instead?

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -607,10 +613,20 @@ mixin SearchFieldMixin<T extends StatefulWidget> on State<T> {
         autoCompleteLayerLink: controller.autoCompleteLayerLink,
         decoration: decoration,
         tracking: tracking,
+        onClose: onClose,
       ),
     );
   }
 
+  KeyEventResult _determineKeyEventResult(
+      int keyEventId, Set<LogicalKeyboardKey> keyEventsToPropogate) {
```
kenzieschmoll, line 622: nit: trailing comma

packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -66,6 +66,7 @@ class _FileSearchFieldState extends State<FileSearchField>
       context: context,
       searchFieldKey: fileSearchFieldKey,
       onTap: _onSelection,
+      autocompleteMatchTileHeight: 50.0,
```
kenzieschmoll, line 69: make this a const in this class.

kenzieschmoll (approved):
One more comment, then LGTM!

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -166,6 +166,27 @@ class _CodeViewState extends State<CodeView>
...
+                  left: 110.0,
+                  child: wrapInElevatedCard(
+                    FileSearchField(
+                      controller: widget.controller,
+                    ),
+                  ),
+                ),
+            ],
+          );
+        });
```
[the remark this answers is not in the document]
elliette: Done, thanks!

packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -156,6 +156,10 @@ class DebuggerController extends DisposableController
 
   ValueListenable<ScriptLocation> get scriptLocation => _scriptLocation;
 
+  final _showFileOpener = ValueNotifier<bool>(false);
+
+  ValueListenable<bool> get showFileOpener => _showFileOpener;
```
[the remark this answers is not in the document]
elliette: Done, thanks! Also reordered the members above it

packages/devtools_app/lib/src/debugger/debugger_screen.dart
```diff
@@ -299,22 +299,26 @@ class EscapeIntent extends Intent {
...
   @override
   void invoke(EscapeIntent intent) {
-    Navigator.of(intent._context).pop(dialogDefaultContext);
-    intent._controller.toggleSearchInFileVisibility(false);
+    if (intent._controller.showSearchInFileField.value) {
+      Navigator.of(intent._context).pop(dialogDefaultContext);
+      intent._controller.toggleSearchInFileVisibility(false);
+    }
+    if (intent._controller.showFileOpener.value) {
+      intent._controller.toggleFileOpener(false);
```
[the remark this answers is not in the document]
elliette: I don't think so - I actually don't think we need to be popping off the dialog for the search in file field (I don't see a dialog being used). I've removed it and it still seems to work fine

packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -1014,6 +1018,10 @@ class DebuggerController extends DisposableController
     }
   }
 
+  void toggleFileOpener(bool visible) {
```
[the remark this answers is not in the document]
elliette: Done, thanks!

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Done, thanks!

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Switched to everything being in one stack, and making sure the file opener closes when the search in file field is open and visa versa.

packages/devtools_app/lib/src/debugger/debugger_controller.dart
elliette: Ah good to know, thanks!

packages/devtools_app/lib/src/ui/search.dart
elliette: Good idea, done!

packages/devtools_app/lib/src/ui/search.dart
elliette: Done, thanks!

packages/devtools_app/lib/src/debugger/file_search.dart

Issue #3317, opened 2021-08-30
Cache filtered CPU profiles 


kenzieschmoll:
Currently if you filter a CPU profile to hide Flutter framework stack frames, for example, then disable the filter, then re-enable it, we will reprocess the CPU samples to generate the filtered profile with hidden Flutter framework stack frames. We should cache these so filter changes are quick the second time.

Issue #2982, opened 2021-05-05
Add ability to profile app start in the CPU Profiler page


kenzieschmoll:
https://github.com/flutter/devtools/pull/2943 should be updated to fix this once we have the appropriate APIs to work with in DDS.

cc @bkonyi

PR #3357, opened 2021-09-10
Add ability to profile app start up and improve CPU profile caching

kenzieschmoll:
Work towards https://github.com/flutter/devtools/issues/2982. We will still use the DDS apis once they are available.
Screen Shot 2021-09-09 at 3 44 17 PM
When an app start up profile is available, clicking this button will load the profile and automatically select the AppStartUp usertag:
Screen Shot 2021-09-10 at 3 45 31 PM
When an app start up profile is not available, a warning message will be displayed. The default profiling rate may cause this case to occur more often than we like. See https://github.com/flutter/flutter/issues/84884#issuecomment-917050444 for some more details.
Screen Shot 2021-09-09 at 3 39 27 PM

The PR also fixes https://github.com/flutter/devtools/issues/3317 and improves the caching for the CPU profiler in general. This will prevent duplicate processing actions when a user is changing filters back and forth. This also fixes bugs where the filters were not applied correctly to profiles cached by user tag.

### DIFF
--- a/packages/devtools_app/lib/src/common_widgets.dart
+++ b/packages/devtools_app/lib/src/common_widgets.dart
@@ -94,6 +94,7 @@ class IconLabelButton extends StatelessWidget {
     this.minScreenWidthForTextBeforeScaling,
     this.elevatedButton = false,
     this.tooltip,
+    this.tooltipPadding,
   })  : assert((icon == null) != (imageIcon == null)),
         super(key: key);
 
@@ -114,6 +115,8 @@ class IconLabelButton extends StatelessWidget {
 
   final String tooltip;
 
+  final EdgeInsetsGeometry tooltipPadding;
+
   @override
   Widget build(BuildContext context) {
     final iconLabel = MaterialIconLabel(
@@ -125,8 +128,9 @@ class IconLabelButton extends StatelessWidget {
     );
     if (elevatedButton) {
       return maybeWrapWithTooltip(
-        tooltip,
-        ElevatedButton(
+        tooltip: tooltip,
+        tooltipPadding: tooltipPadding,
+        child: ElevatedButton(
           onPressed: onPressed,
           child: iconLabel,
         ),
@@ -135,8 +139,9 @@ class IconLabelButton extends StatelessWidget {
     // TODO(kenz): this SizedBox wrapper should be unnecessary once
     // https://github.com/flutter/flutter/issues/79894 is fixed.
     return maybeWrapWithTooltip(
-      tooltip,
-      SizedBox(
+      tooltip: tooltip,
+      tooltipPadding: tooltipPadding,
+      child: SizedBox(
         height: defaultButtonHeight,
         width: !includeText(context, minScreenWidthForTextBeforeScaling)
             ? buttonMinWidth
@@ -1294,10 +1299,15 @@ class Link {
   final String url;
 }
 
-Widget maybeWrapWithTooltip(String tooltip, Widget child) {
+Widget maybeWrapWithTooltip({
+  @required String tooltip,
+  EdgeInsetsGeometry tooltipPadding,
+  @required Widget child,
+}) {
   if (tooltip != null) {
     return DevToolsTooltip(
       tooltip: tooltip,
+      padding: tooltipPadding,
       child: child,
     );
   }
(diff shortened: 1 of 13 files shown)


packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -114,6 +115,8 @@ class IconLabelButton extends StatelessWidget {
 
   final String tooltip;
 
+  final EdgeInsetsGeometry tooltipPadding;
```
jacob314, line 118: why is this needed? Is there a reason we can't just give all tooltips a reasonable default?

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
```diff
@@ -32,12 +32,28 @@ class CpuProfilerController
   /// the message that no filters are applied.
   static const userTagNone = 'none';
 
+  /// User tag that marks app start up for Flutter apps.
+  ///
+  /// This needs to match the tag set and unset at these locations,
+  /// respectively:
+  ///
+  /// Flutter Engine - https://github.com/flutter/engine/blob/master/runtime/dart_isolate.cc#L369
```
jacob314, line 40: nit: those lines and columns will probably become wrong over time as unrelated lines are added to the files.
Maybe specify a git hash so the link shows the relevant line although perhaps that will be more confusing?
https://github.com/flutter/engine/blob/f52653ba4568b4a40e8624d47773d156740ea9f7/runtime/dart_isolate.cc#L369

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
```diff
@@ -61,10 +77,13 @@ class CpuProfilerController
   ValueListenable<String> get userTagFilter => _userTagFilter;
   final _userTagFilter = ValueNotifier<String>(userTagNone);
 
-  @visibleForTesting
-  final dataByTag = <String, CpuProfileData>{};
+  Iterable<String> get userTags =>
+      cpuProfileStore.lookupProfile(label: userTagNone)?.userTags ?? [];
```
jacob314, line 81: `[]` ==> `const []`
to avoid an allocation for the common case.

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
```diff
@@ -61,10 +77,13 @@ class CpuProfilerController
...
 
-  Iterable<String> get userTags => dataByTag[userTagNone]?.userTags ?? [];
+  bool get isToggleFilterActive =>
+      toggleFilters.firstWhere((filter) => filter.enabled.value,
```
jacob314, line 84: https://api.dart.dev/stable/2.14.1/dart-core/Iterable/any.html
does this more cleanly.

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
```diff
@@ -229,10 +253,109 @@ class CpuProfilerController
     return matches;
   }
 
-  void loadProcessedData(CpuProfileData data) {
+  Future<void> loadAllSamples() async {
+    reset();
+    await pullAndProcessProfile(
+      startMicros: 0,
+      // Using [maxJsInt] as [extentMicros] for the getCpuProfile requests will
+      // give us all cpu samples we have available
```
jacob314, line 261: missing period at end of sentence.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/common_widgets.dart
kenzieschmoll: Multiline tooltips will look bad without specifying the padding. Single line tooltips have adequate spacing on the top and bottom, but if there is a new line char in the tooltip message, that top and bottom spacing disappears.

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
kenzieschmoll: done

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
kenzieschmoll: nice. thanks.

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
kenzieschmoll: done

packages/devtools_app/lib/src/profiler/cpu_profile_controller.dart
kenzieschmoll: done

PR #3358, opened 2021-09-10
Do not wait on stream listening during start up

elliette:


jacob314:
Do we want to land this or close this pull request?

packages/devtools_app/test/service_manager_test.dart
```diff
@@ -72,16 +74,17 @@ void main() async {
...
-            'streamListen',
           ]),
         );
+
+        expect(
+          serviceManager.service.vmServiceCalls
+              .where((call) => call == 'streamListen')
+              .toList()
+              .length,
+          equals(2 + 8), // 2 from package:devtools_testing:flutter_test_driver.
```
kenzieschmoll, line 85: instead of working around the calls from test driver in our test, let's clear out vmServiceCalls once the test driver has finished setup. That will let us remove the test driver calls in the above expect statement too

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -248,19 +248,6 @@ class ServiceConnectionManager {
 
     service.onEvent(serviceStreamName).listen(handleServiceEvent);
 
-    _connectedState.value = const ConnectedState(true);
```
kenzieschmoll, line 251: why was all this code moved to be after the streams?

kenzieschmoll: never mind, this seems fine. If all this was being called before the stream listeners were initialized, then it shouldn't matter to be called after the stream listeners are initialized unawaited.

kenzieschmoll (approved):
LGTM after test is fixed

PR #3360, opened 2021-09-14
Add selection analytics for NetworkScreen

kenzieschmoll:
Work towards https://github.com/flutter/devtools/issues/3270.

This PR tracks selections for the network profiler controls (pause, resume, clear) as well as for tab changes on the request inspector to understand usage of the different request inspector views.


packages/devtools_app/lib/src/network/network_model.dart
```diff
@@ -1,3 +1,4 @@
+import 'package:flutter/material.dart';
```
jacob314, line 1: nit: this is an import of foundation.dart not material.
Doesn't matter but otherwise made me double take to why material was being used in a data model class.

packages/devtools_app/lib/src/network/network_request_inspector.dart
```diff
@@ -36,73 +38,140 @@ class NetworkRequestInspector extends StatelessWidget {
...
+    with TickerProviderStateMixin {
+  TabController _tabController;
+
+  List<DevToolsTab> _tabs;
+
+  int _currentTabControllerIndex = 0;
+
+  void _initTabController() {
+    _tabController?.removeListener(_onTabChanged);
+    _tabController?.dispose();
```
jacob314, line 56: I imagine we want to start tracking analytics on other cases we have tabs in DevTools. Would it make sense to define a AnalyticsTabController that handles logging the tab name each time a different tab is selected?
Seem annoying to have to change from a stateless to stateful widget just to make tabs work with our analytics.

jacob314 (approved):
lgtm
lgtm with one question about generalizing to handle the generic case of tabs with associated analytics ids.

packages/devtools_app/lib/src/network/network_model.dart
kenzieschmoll: done

packages/devtools_app/lib/src/network/network_request_inspector.dart
kenzieschmoll: Done - Added AnalyticsTabbedView that handles sending tab switches to analytics.

elliette (approved):

Issue #3226, opened 2021-07-29
CPU profiler only works on the Flutter isolate


Piero512:
Hey!

Since @bkonyi suggested I open an issue here about an [issue](https://github.com/dart-lang/sdk/issues/46756#issuecomment-889352243) I have with devtools. 

My issue with Devtools is that the CPU profiler seems to only work for the main isolate and I haven't find a way to choose a different isolate to get profiler samples.

I did find my problem with the observatory and it was some library that was doing numerical calculation on Lists, where 27% of the time was spent allocating new spaces (hey, maybe I can make an issue to the Dart SDK about it). 

As a relevant fact, the isolate was spawned through Flutter's compute function, so if the reason that I wasn't able to choose it was because the isolate name starts with an underscore (I think the isolate name at the observatory was _spawn), please feel free to move it to the appropiate repository.

PR #3362, opened 2021-09-14
Add multi-isolate support to the CPU profiler

kenzieschmoll:
Screen Shot 2021-09-14 at 11 13 05 AM
Screen Shot 2021-09-14 at 11 13 10 AM

Add the isolate selector to the CPU profiler page. Each isolate is given an instance of `CpuProfileStore` to track cached profiles for the isolate.

Fixes https://github.com/flutter/devtools/issues/3226

### DIFF
--- a/packages/devtools_app/lib/src/connected_app.dart
+++ b/packages/devtools_app/lib/src/connected_app.dart
@@ -39,8 +39,7 @@ class ConnectedApp {
   bool _isFlutterApp;
 
   FlutterVersion get flutterVersionNow {
-    assert(isFlutterNativeAppNow);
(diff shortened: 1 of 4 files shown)


jacob314 (approved):
lgtm

jacob314:
Might be worth filing a bug for future work to capture how users can profile short lived isolates. Right now I imagine that isn't possible.

kenzieschmoll:
This should be possible if a user switches to the isolate and then clicks "Load all CPU samples" if a couple conditions are met. Filed https://github.com/flutter/devtools/issues/3363 with details.

Issue #3249, opened 2021-08-04
File:line:col numbers are wrong color in hover cards and console output


kenzieschmoll:
![Screen Shot 2021-08-04 at 11 57 23 AM](https://user-images.githubusercontent.com/43759233/128240327-2427e433-31f8-447e-913c-7ac6077cfa55.png)
![Screen Shot 2021-08-04 at 11 57 29 AM](https://user-images.githubusercontent.com/43759233/128240331-dc8df636-5a59-4ee7-8f39-fc2e447f5b69.png)

alexlindroos:
@kenzieschmoll For me dark mode and light mode are looking like this. Dark mode seems fine, but light mode has white text for file:line:col which makes it hard to spot.

![darkmode_filecol_text](https://user-images.githubusercontent.com/17140689/133577806-bf9a0638-f804-4d20-9f5d-630fef4a229a.png)

![lightmode_filecol_text](https://user-images.githubusercontent.com/17140689/133577810-4e47a505-dd21-4012-8932-6bb4ac8e8c84.png)

flutter doctor -v

[✓] Flutter (Channel dev, 2.6.0-5.1.pre, on macOS 11.5.2 20G95 darwin-x64,
    locale en-FI)
    • Flutter version 2.6.0-5.1.pre at /Users/codemate/Desktop/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision f2903d3da5 (8 days ago), 2021-09-08 12:44:29 -0700
    • Engine revision 0bd9be379a
    • Dart version 2.15.0 (build 2.15.0-82.0.dev)

[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
    • Android SDK at /Users/codemate/Library/Android/sdk
    • Platform android-30, build-tools 30.0.3
    • Java binary at: /Applications/Android
      Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7281165)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 12.5.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • CocoaPods version 1.10.1

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2020.3)
    • Android Studio at /Applications/Android Studio.app/Contents
[...]

PR #3365, opened 2021-09-16
Fix file:line:col color fixed (#3249)

alexlindroos:
@kenzieschmoll 

Before:

darkmode_filecol_text
lightmode_filecol_text

After:

new_darkmode_fielcol

new_lightmode_filecol


packages/devtools_app/lib/src/theme.dart
```diff
@@ -242,6 +242,8 @@ extension DevToolsColorScheme on ColorScheme {
 
   Color get serviceExtensionButtonsFillSelected => devtoolsBlue[400];
 
+  Color get fileLineColColor => isLight ? Colors.black : Colors.white;
```
jacob314, line 245: is there a reason not to use defaultForeground instead? 
Using full white in the dark color scheme is brighter than other text so should be avoided unless there is a strong reason.

alexlindroos: My bad, was focused on creating own color for clarity. I will change it to use defaultForeground.

alexlindroos:
@jacob314 new colors using defaultForeground below:

Screenshot 2021-09-30 at 10 47 54
Screenshot 2021-09-30 at 10 48 41

kenzieschmoll (approved):

PR #3367, opened 2021-09-16
[Cleanup] Merge InspectorTreeControllerFlutter and superclass into single InspectorTreeController class

matt-ragonese:
Previously, `InspectorTreeControllerFlutter` implemented the `InspectorTreeController` abstract class and `InspectorTreeFixedRowHeightController` mixin.

This change merges these three classes/mixins into a single `InspectorTreeController` class.


jacob314 (approved):
lgtm
Good to land once the bots go green.
Thanks for cleaning this up.

PR #3368, opened 2021-09-16
minor fix in my_application.h file

kapilkumar2001:


kapilkumar2001:
@googlebot I signed it!

jacob314:
What problem was that line causing? We didn't manually create that file. Has the code in Flutter to create this file been changed?

kapilkumar2001:
No, Actually this line was giving an error because of that semicolon. I have added that semicolon so it will work fine now. I think this is coming from template code, I have opened a PR on flutter/flutter for this :)

Hixie:
This doesn't seem right. Looking at the docs for G_DECLARE_FINAL_TYPE it doesn't suggest using a semicolon there either. (I couldn't find the definition of that macro with a quick Google search.)

Hixie:
Since there's been no progress here since November I'm going to close this PR. Please don't hesitate to open a new one if this is still something you want to work on!

PR #3371, opened 2021-09-17
Improves searching in the file picker 

elliette:
Makes a few changes to the file search to speed it up:

* while the current query is a continuation of the previous query (eg, "cat" is a continuation of "ca"), search through the previous matches only (before this, we were searching through all the files each time the query changed)
* only fuzzy-match on the file name, not the entire file path
* combine the exact match and fuzzy match for files into one loop, instead of doing two separate loops (this also fixes an edge case where you could have a file listed multiple times in the results because it would be included in the exact matches and the fuzzy matches)

Also adds a test to the file search to validate that it returns the expected results.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -398,7 +398,6 @@ class _CodeViewState extends State<CodeView>
     return wrapInElevatedCard(
       FileSearchField(
         debuggerController: widget.controller,
-        autoCompleteController: AutoCompleteController(),
       ),
     );
   }
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -397,7 +397,8 @@ class _CodeViewState extends State<CodeView>
   Widget buildFileSearchField() {
     return wrapInElevatedCard(
       FileSearchField(
-        controller: widget.controller,
+        debuggerController: widget.controller,
+        autoCompleteController: AutoCompleteController(),
       ),
     );
   }
--- a/packages/devtools_app/lib/src/debugger/file_search.dart
+++ b/packages/devtools_app/lib/src/debugger/file_search.dart
@@ -61,8 +61,8 @@ class _FileSearchFieldState extends State<FileSearchField>
     final previousQuery = _query;
     final currentQuery = widget.autoCompleteController.search;
 
-    // If the current query is a continuation of the previous query, then wittle
-    // down the matches. Otherwise search through all scripts:
+    // If the current query is a continuation of the previous query, then
+    // whittle down the matches. Otherwise search through all scripts:
     final scripts = currentQuery.startsWith(previousQuery)
         ? _matches
         : widget.debuggerController.sortedScripts.value;
--- a/packages/devtools_app/lib/src/debugger/file_search.dart
+++ b/packages/devtools_app/lib/src/debugger/file_search.dart
@@ -18,10 +18,12 @@ const double autocompleteMatchTileHeight = 50.0;
 
 class FileSearchField extends StatefulWidget {
   const FileSearchField({
-    @required this.controller,
+    @required this.debuggerController,
+    @required this.autoCompleteController,
   });
 
-  final DebuggerController controller;
+  final DebuggerController debuggerController;
+  final AutoCompleteController autoCompleteController;
 
   @override
   _FileSearchFieldState createState() => _FileSearchFieldState();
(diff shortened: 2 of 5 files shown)


packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -397,7 +397,8 @@ class _CodeViewState extends State<CodeView>
   Widget buildFileSearchField() {
     return wrapInElevatedCard(
       FileSearchField(
-        controller: widget.controller,
+        debuggerController: widget.controller,
+        autoCompleteController: AutoCompleteController(),
```
jacob314, line 401: should you create the autoCompleteController once in initState instead of creating it every single build?

packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -18,52 +18,73 @@ const double autocompleteMatchTileHeight = 50.0;
...
-    final query = _autoCompleteController.search;
-    final matches = findMatches(query, widget.controller.sortedScripts.value);
+    final previousQuery = _query;
+    final currentQuery = widget.autoCompleteController.search;
+
+    // If the current query is a continuation of the previous query, then
+    // whittle down the matches. Otherwise search through all scripts:
+    final scripts = currentQuery.startsWith(previousQuery)
+        ? _matches
+        : widget.debuggerController.sortedScripts.value;
```
jacob314, line 68: the line widget.autocompleteController occurs a lot now. Feel free to add an
autocompleteController getter to the state class that just returns widget.autocompleteController to make this code a little less verbose.

packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -18,52 +18,73 @@ const double autocompleteMatchTileHeight = 50.0;
...
   }
 
   void _handleSearch() {
-    final query = _autoCompleteController.search;
-    final matches = findMatches(query, widget.controller.sortedScripts.value);
+    final previousQuery = _query;
+    final currentQuery = widget.autoCompleteController.search;
+
+    // If the current query is a continuation of the previous query, then
+    // whittle down the matches. Otherwise search through all scripts:
```
jacob314, line 65: tiny nit: whittle down the matches => filter the previous matches.

packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -91,20 +112,20 @@ class _FileSearchFieldState extends State<FileSearchField>
...
+    widget.autoCompleteController.closeAutoCompleteOverlay();
+    widget.debuggerController.toggleFileOpenerVisibility(false);
     _scriptsCache.clear();
   }
 
   @override
   void dispose() {
     _onClose();
-    _autoCompleteController.dispose();
+    widget.autoCompleteController.dispose();
```
jacob314, line 128: this shouldn't be disposed here. we don't own the _autoCompleteController. The state object that created it and passed it in should own it.

packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -114,22 +135,23 @@ List<ScriptRef> findMatches(
...
-      .where((scriptRef) => scriptRef.uri.caseInsensitiveContains(query))
-      .toList();
+  final exactMatches = [];
+  final fuzzyMatches = [];
 
-  if (exactMatches.length >= numOfMatchesToShow) {
-    return takeTopMatches(exactMatches);
+  for (final scriptRef in scriptRefs) {
+    final fullPath = scriptRef.uri;
+    final fileName = scriptRef.uri.split('/').last;
```
jacob314, line 146: to optimize the performance of this process, catch a list of split script URIs or fileNames rather than splitting every time.

packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -114,22 +135,23 @@ List<ScriptRef> findMatches(
...
+    if (fullPath.caseInsensitiveContains(query)) {
+      exactMatches.add(scriptRef);
+    } else if (fileName.caseInsensitiveFuzzyMatch(query)) {
```
jacob314, line 149: might want to also check to see if caseInsensitiveFuzzyMatch can be optimized

packages/devtools_app/test/file_search_test.dart
```diff
@@ -0,0 +1,164 @@
...
+            'animals/cats/purr.dart',
+            'animals/insects/caterpillar.dart',
+            'animals/insects/cicada.dart',
+            'food/catering/party.dart',
+            'food/carton/milk.dart',
+            'food/milk/carton.dart',
+            'travel/adventure/cave_tours_europe.dart',
+            'travel/canada/banff.dart',
+          ]),
+          reason: 'Correct search results for "c" query.');
```
jacob314, line 75: don't worry about supplying a reason unless it is really surprising. For example, here it is clear the query is 'c' so no need to document that.

packages/devtools_app/test/file_search_test.dart
```diff
@@ -0,0 +1,164 @@
...
+
+      autoCompleteController.search = 'ca';
+      expect(
+          autoCompleteController.searchAutoComplete.value,
+          equals([
+            // Exact matches:
```
jacob314, line 81: helpful to call out the exact and fuzzy matches. you might even go a step further and provide a version of the api where you can query for exactly which matches were exact and fuzzy

jacob314 (commented):
Done with around of comments.

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Switched to no longer passing it in - I think file search widget should own it, not codeview. I've switched back to creating it within the `initState` method for the file search widget. I've also made FileSearchFieldState/autocompleteController public, so they can be accessed from the test.

packages/devtools_app/lib/src/debugger/file_search.dart
elliette: Back to creating it in the file search state, so it's no longer on the widget

packages/devtools_app/lib/src/debugger/file_search.dart

packages/devtools_app/lib/src/debugger/file_search.dart
elliette: I took a look at it before - it's linear so I don't think it can be optimized further

packages/devtools_app/test/file_search_test.dart
jacob314: failures will have a line and column clarifying what the failure is.

packages/devtools_app/test/file_search_test.dart
jacob314: I mean change the controller to also specify exactly what the exact and fuzzy matches are

packages/devtools_app/test/file_search_test.dart
elliette: Ah good to know, done!

packages/devtools_app/test/file_search_test.dart
elliette: Ah got it. That will be coming in a forthcoming PR (when I add the autocomplete matches)

packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -42,40 +43,58 @@ class _FileSearchFieldState extends State<FileSearchField>
...
+    _query = autoCompleteController.search;
     _matches = widget.debuggerController.sortedScripts.value;
 
     // Open the autocomplete results immediately before a query is entered:
     SchedulerBinding.instance.addPostFrameCallback((_) => _handleSearch());
   }
 
+  @override
+  Widget build(BuildContext context) {
+    return buildAutoCompleteSearchField(
```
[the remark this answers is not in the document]
elliette: Hmm I'm not sure I understand. This is a pre-existing method in `search.dart`: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/ui/search.dart#L533, I don't think it can be easily converted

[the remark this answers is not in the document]
kenzieschmoll: What Jacob is suggesting is to basically take the buildAutoCompleteSearchField method and convert it to 

`class AutoCompleteSearchField extends StatefulWidget {` where the parameters in the constructor match the parameters for this method. Then the build method does the work of constructing this widget tree..

However, in this case, I think we should actually leave this as is, because we only want classes that mixin `SearchFieldMixin` to be using this auto complete search field widget, and we can't enforce that if this is a public widget class. I suppose we could make a private widget _AutoCompleteSearchField and have this `buildAutoCompleteSearchField` method return that private widget. That would be a good solution.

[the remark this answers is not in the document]
elliette: Ah got it! I've left a TODO for having it return a private ` _AutoCompleteSearchField`. Will address that in a separate PR.

Issue #3375, opened 2021-09-21
Evaluate autocomplete is formatted incorrectly 


elliette:
![evaluate 2](https://user-images.githubusercontent.com/21270878/134093453-7651f23b-be34-4e8c-ad5d-46d66ef9e8f9.gif)

I can reproduce this on the current version 2.5.0, but not on the previous version (2.5.0). 

`flutter pub global activate devtools 2.6.0`
`flutter pub global run devtools`
--> strange formatting on autocomplete

`flutter pub global activate devtools 2.5.0`
`flutter pub global run devtools`
--> no strange formatting on autocomplete

elliette:
FYI @kenzieschmoll @jacob314 @annagrin - I'm currently looking into why this is happening

PR #3377, opened 2021-09-21
Calculate tile size for the autocomplete widget

elliette:
Fixes https://github.com/flutter/devtools/issues/3375

Calculates the size of the autocomplete tiles to be dependent on the text size. Instead of using the `ListTile` widget (which has a minimum height of 48px: https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/material/list_tile.dart#L1379-L1380), we now use our own implementation so that they can be narrower.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/evaluate.dart
+++ b/packages/devtools_app/lib/src/debugger/evaluate.dart
@@ -47,7 +47,6 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
         onTap: _onSelection,
         bottom: false,
         maxWidth: false,
-        autoCompleteMatchTileHeight: 50.0,
       );
     });
     addAutoDisposeListener(
--- a/packages/devtools_app/lib/src/debugger/evaluate.dart
+++ b/packages/devtools_app/lib/src/debugger/evaluate.dart
@@ -47,6 +47,7 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
         onTap: _onSelection,
         bottom: false,
         maxWidth: false,
+        autoCompleteMatchTileHeight: 50.0,
       );
     });
     addAutoDisposeListener(
--- a/packages/devtools_app/lib/src/debugger/file_search.dart
+++ b/packages/devtools_app/lib/src/debugger/file_search.dart
@@ -14,7 +14,6 @@ import 'debugger_controller.dart';
 import 'debugger_model.dart';
 
 const int numOfMatchesToShow = 10;
-const double autoCompleteMatchTileHeight = 50.0;
 
 class FileSearchField extends StatefulWidget {
   const FileSearchField({
@@ -67,7 +66,6 @@ class _FileSearchFieldState extends State<FileSearchField>
       context: context,
       searchFieldKey: fileSearchFieldKey,
       onTap: _onSelection,
-      autoCompleteMatchTileHeight: autoCompleteMatchTileHeight,
     );
   }
 
--- a/packages/devtools_app/lib/src/debugger/file_search.dart
+++ b/packages/devtools_app/lib/src/debugger/file_search.dart
@@ -14,7 +14,7 @@ import 'debugger_controller.dart';
 import 'debugger_model.dart';
 
 const int numOfMatchesToShow = 10;
-const double autocompleteMatchTileHeight = 50.0;
+const double autoCompleteMatchTileHeight = 50.0;
 
 class FileSearchField extends StatefulWidget {
   const FileSearchField({
@@ -67,7 +67,7 @@ class _FileSearchFieldState extends State<FileSearchField>
       context: context,
       searchFieldKey: fileSearchFieldKey,
       onTap: _onSelection,
-      autocompleteMatchTileHeight: autocompleteMatchTileHeight,
+      autoCompleteMatchTileHeight: autoCompleteMatchTileHeight,
     );
   }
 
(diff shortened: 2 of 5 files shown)


packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -67,7 +67,7 @@ class _FileSearchFieldState extends State<FileSearchField>
       context: context,
       searchFieldKey: fileSearchFieldKey,
       onTap: _onSelection,
-      autocompleteMatchTileHeight: autocompleteMatchTileHeight,
+      autoCompleteMatchTileHeight: autoCompleteMatchTileHeight,
```
jacob314, line 70: this size needs to be scaled with the font size.

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -47,6 +47,7 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
         onTap: _onSelection,
         bottom: false,
         maxWidth: false,
+        autoCompleteMatchTileHeight: 50.0,
```
jacob314, line 50: This seems way too high for eval. I can imagine a higher value for file search as you are showing 2 lines of text for each entry.
Please include before and after screenshots to clarify.
You'll also need to scale this value by the font size so it still works well with larger or smaller fonts.

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -176,7 +176,7 @@ class AutoCompleteState extends State<AutoComplete> with AutoDisposeMixin {
     // Approximation but it's pretty accurate. Could consider using a layout builder
     // or maybe build in an overlay (that's isn't visible) to compute.
     final tileEntryHeight =
-        autoComplete.autocompleteMatchTileHeight ?? box.size.height;
```
jacob314, line 179: prefer only providing one solution if possible. 
If the box.size.height is not working, we should remove it and make autoCompleteTileHeight `required` rather than having a safe solution and a broken solution.

jacob314 (commented):
Done with a round of comments.

packages/devtools_app/lib/src/debugger/evaluate.dart
elliette: Dug into this a bit more. The formatting gets very strange for anything less than 48px in height (see photos below),  which is what Flutter is defining as the default list tile height for one line of text & dense spacing: https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/material/list_tile.dart#L1379-L1380

I'm guessing this is because the Material specs says that clickable elements should be at least 48px for accessibility. So if we want it to be any narrower, we would probably have to use a different widget (maybe row?) 

I've changed the height to be whatever is larger of 48px or the text height. That way we don't need the `autoCompleteMatchTileHeight` parameter at all. 

How do I choose a larger font size to verify this scales nicely?


**48 px (current minimum height):**
Screen Shot 2021-09-21 at 2 17 26 PM

**40px:**
Screen Shot 2021-09-21 at 2 05 22 PM

**30px:**
Screen Shot 2021-09-21 at 2 05 02 PM

**20px:**
Screen Shot 2021-09-21 at 2 04 42 PM

packages/devtools_app/lib/src/debugger/file_search.dart
elliette: Should now be scaling, but as mentioned in other comment - how do I choose a larger font size to verify this scales nicely?

packages/devtools_app/lib/src/ui/search.dart
elliette: As mentioned above, I got rid of the `autocompleteMatchTileHeight` and am instead calculating the height to use

packages/devtools_app/lib/src/debugger/evaluate.dart
jacob314: we should write our own trivial widget for this. Even 30px looks like it is adding too much space.

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -168,16 +164,29 @@ class AutoCompleteState extends State<AutoComplete> with AutoDisposeMixin {
...
+    final tileContents = searchAutoComplete.value == null
+        ? <TextSpan>[]
+        : searchAutoComplete.value
+            .map((text) => TextSpan(
+                  text: text,
+                  style: autoCompleteTextStyle,
+                ))
+            .toList();
+
+    const minTileHeight = 48.0;
```
jacob314, line 180: remove this and fix the ui to stop using tile.

jacob314 (commented):
lgtm once the issue having to set a min tile height is fixed.

packages/devtools_app/lib/src/debugger/evaluate.dart
elliette: Sg, removed ListTile! This is what it looks like now:
Screen Shot 2021-09-22 at 10 24 49 AM

jacob314:
lgtm

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -177,10 +177,10 @@ class AutoCompleteState extends State<AutoComplete> with AutoDisposeMixin {
                 ))
             .toList();
 
-    const minTileHeight = 48.0;
+    const padding = 8.0;
```
jacob314, line 180: use our constands for padding rather than hardcoding 8.0. If we ever tweak our padding that makes it easier.

packages/devtools_app/lib/src/ui/search.dart
elliette: Done!

PR #3384, opened 2021-09-23
Highlight matches in the file picker dropdown

elliette:
Dark mode:
autocomplete dark mode

Light mode:
autocomplete light mode


jacob314:
Nit: yellow is too bright. I think other similar file match autocompletes tend to bold matches rather than using a different color.

kenzieschmoll:
Here's what intellij does: 
Screen Shot 2021-09-23 at 12 28 38 PM

elliette:
Here's how it looks with just bolding - wdyt?

Screen Shot 2021-09-23 at 1 30 55 PM
Screen Shot 2021-09-23 at 1 32 59 PM

kenzieschmoll:
I think I actually like VS code's approach the best. Bold + blue? We could match the blue color that is already highlighting the search field

packages/devtools_app/lib/src/ui/utils.dart
```diff
@@ -134,6 +134,19 @@ double calculateTextSpanHeight(TextSpan span) {
   return textPainter.height;
 }
 
+TextSpan findLongestTextSpan(List<TextSpan> spans) {
```
kenzieschmoll, line 137: now that this is easier to test, can you add tests for this in utils_test.dart?

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -899,3 +960,17 @@ mixin TreeDataSearchStateMixin<T extends TreeNode<T>>
 
 class AutoCompleteController extends DisposableController
     with SearchControllerMixin, AutoCompleteSearchControllerMixin {}
+
+class AutoCompleteMatchSegment {
```
kenzieschmoll, line 964: instead of creating an AutoComplete specific class for this, can you use Range from utils.dart?

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -316,19 +316,8 @@ class _CodeViewState extends State<CodeView>
...
-                                    line.toPlainText().length;
-                                if (currentLength > longestLength) {
-                                  longestLength = currentLength;
-                                  longestLine = line;
-                                }
-                              }
-                              final double fileWidth =
-                                  calculateTextSpanWidth(longestLine);
+                              final double fileWidth = calculateTextSpanWidth(
+                                  findLongestTextSpan(lines));
```
kenzieschmoll, line 320: nit: trailing comma

packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -125,6 +127,36 @@ class FileSearchFieldState extends State<FileSearchField>
...
+      return AutoCompleteMatch(match);
+    }
+
+    final autoCompleteResultSegments = <AutoCompleteMatchSegment>[];
+    if (match.contains(query)) {
+      final start = match.indexOf(query);
+      final end = start + query.length;
+      autoCompleteResultSegments.add(AutoCompleteMatchSegment(start, end));
+    } else {
+      final fileName = match.split('/').last;
```
kenzieschmoll, line 141: can you add a comment as the first line of this else statement that describes this is for a fuzzy match?

packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -125,6 +127,36 @@ class FileSearchFieldState extends State<FileSearchField>
...
+          final start = matchIndex;
+          final end = matchIndex + 1;
+          autoCompleteResultSegments.add(AutoCompleteMatchSegment(start, end));
+          queryIndex++;
+        }
+      }
+    }
+
+    return AutoCompleteMatch(match,
+        matchedSegments: autoCompleteResultSegments);
```
kenzieschmoll, line 157: nit: tc

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -168,19 +169,29 @@ class AutoCompleteState extends State<AutoComplete> with AutoDisposeMixin {
...
-            .map((text) => TextSpan(
-                  text: text,
-                  style: autoCompleteTextStyle,
-                ))
-            .toList();
+    final autoCompleteHighlightedTextStyle = Theme.of(context)
+        .regularTextStyle
+        .copyWith(
+            color: colorScheme.autoCompleteHighlightedTextColor,
+            fontWeight: FontWeight.bold);
```
kenzieschmoll, line 176: nit: tc

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -264,6 +275,55 @@ class AutoCompleteState extends State<AutoComplete> with AutoDisposeMixin {
...
+    int previousEndIndex = 0;
+
+    for (final segment in matchedSegments) {
+      if (previousEndIndex < segment.start) {
+        // Add the unhighlighted segment before the current highlighted segment:
+        final segmentBefore = text.substring(previousEndIndex, segment.start);
+        spans.add(TextSpan(
+          text: segmentBefore,
+          style: regularTextStyle,
+        ));
```
kenzieschmoll, line 304: nit: tc here and for spans.add calls below

kenzieschmoll (commented):
done with review

packages/devtools_app/lib/src/ui/utils.dart
elliette: Good idea, added tests

packages/devtools_app/lib/src/ui/search.dart
elliette: Oh nice, done

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Done! OOC, what cases do we need trailing commas? All function parameters?

packages/devtools_app/lib/src/debugger/file_search.dart
elliette: Good idea, done

elliette:
Now using the search input blue color:

Screen Shot 2021-09-23 at 4 50 25 PM
Screen Shot 2021-09-23 at 4 50 38 PM

packages/devtools_app/lib/src/debugger/codeview.dart
kenzieschmoll: general rule is always* for flutter build methods (*except some cases like a one line SizedBox or similar or a widget with a single unnamed param), and then whenever a function call or method signature has to break to a new line.

kenzieschmoll (approved):
lgtm_army

kenzieschmoll:
@echo27713 for some input on whether these colors should be made more accessible. Not blocking though - this can be done as a follow up if necessary.

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Ah got it, thanks!

echo27713:
@kenzieschmoll @elliette bold+blue looks good to me as well!

PR #3390, opened 2021-09-24
Remove unused imports and package:meta dependency

kenzieschmoll:
The latest flutter triggered a new lint about unused imports - most cases where we were importing two files for the same thing. For example, looks like package:flutter/foundation.dart now has an @required annotation, so we can remove package:meta.


kenzieschmoll:
This PR may need to sit around until the version of flutter with this lint hits dev (that's the version we run on the bots and want to release from).

kenzieschmoll:
closing in favor of https://github.com/flutter/devtools/pull/3440

Issue #3372, opened 2021-09-17
Scrollbar issue on debugger page


xu-baolin:
![image](https://user-images.githubusercontent.com/61075224/133750272-584ad6ae-4f2e-40c1-8d49-cec2840ea9d1.png)
### Video

https://user-images.githubusercontent.com/61075224/134300303-d52e2a9f-74ce-46b0-a2e4-d1fc2eba276d.mp4

windows + Chrome

jacob314:
Fyi @kenzieschmoll @elliette. Looks like we still have some scrollbar issues.
@xu-baolin, thanks for the bug report! Can you provide some more details about the bad srollbar behavior you are seeing? If you interact with the debugger page, does the phantom scrollbar go away? What happens if you resize the window? Sending an animated gif could also help clarify these sorts of questions.

xu-baolin:
> Fyi @kenzieschmoll @elliette. Looks like we still have some scrollbar issues.
> @xu-baolin, thanks for the bug report! Can you provide some more details about the bad srollbar behavior you are seeing? If you interact with the debugger page, does the phantom scrollbar go away? What happens if you resize the window? Sending an animated gif could also help clarify these sorts of questions.

Updated.

PR #3393, opened 2021-09-24
Fix duplicate scrollbars shown on web

elliette:
Fixes https://github.com/flutter/devtools/issues/3372

With https://flutter.dev/docs/release/breaking-changes/default-desktop-scrollbars, ScrollBehaviors now automatically apply Scrollbars to scrolling widgets on desktop platforms. For some reason, an extra scrollbar is automatically being added in the debugger on web, but not on desktop. I've opened up https://github.com/flutter/flutter/issues/90697 in Flutter since this doesn't seem to be intended. 

However, for now disabling the default scrollbars for web fixes the issue we are seeing.

### DIFF
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -311,6 +311,10 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
       ),
       routerDelegate: DevToolsRouterDelegate(_getPage),
       routeInformationParser: DevToolsRouteInformationParser(),
+      // Disable default scrollbar behavior on web to fix duplicate scrollbars
+      // bug, see https://github.com/flutter/flutter/issues/90697:
+      scrollBehavior:
+          const MaterialScrollBehavior().copyWith(scrollbars: !kIsWeb),
     );
   }
 }


packages/devtools_app/lib/src/app.dart
```diff
@@ -311,6 +311,10 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
       ),
       routerDelegate: DevToolsRouterDelegate(_getPage),
       routeInformationParser: DevToolsRouteInformationParser(),
+      // Disable default scrollbar behavior on web to fix duplicate scrollbars
+      // bug, see https://github.com/flutter/flutter/issues/90697:
+      scrollBehavior:
+          const MaterialScrollBehavior().copyWith(scrollbars: !kIsWeb),
```
kenzieschmoll, line 317: what does this look like on web? Are scrollbars now gone? We really only care about how web looks since that's the only platform we ship on.

elliette: We still have the scrollbars, but the "phantom" scrollbar that was appearing in the debugger is now gone.

scrollbars

kenzieschmoll: I see another problem here that the vertical scrollbar on the debugger does not show up until after the file has been scrolled, but I think this may be an independent issue of the phantom scrollbar problem.

kenzieschmoll (approved):

packages/devtools_app/lib/src/app.dart
elliette: Oh interesting! I think this might be intended, we have `isAlwaysShown` set to `false`. From the [definition](https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/widgets/scrollbar.dart#L717-L721):

> By default, the thumb will fade in and out as the child scroll view scrolls. When [isAlwaysShown] is true, the scrollbar thumb will remain visible without the fade animation. 

I'll open a separate PR to set `isAlwaysShown` to true.

Issue #3386, opened 2021-09-24
Opening the FileExplorer in the debugger loses codeview scroll position


kenzieschmoll:
![file-explorer-scroll-bug](https://user-images.githubusercontent.com/43759233/134709843-a469e334-8070-4c72-b982-b11ee64e8b37.gif)
@bkonyi

PR #3395, opened 2021-09-27
Don't scroll code view to top when opening/closing file explorer

bkonyi:
Fixes https://github.com/flutter/devtools/issues/3386

### DIFF
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -82,6 +82,10 @@ class _CodeViewState extends State<CodeView>
 
   ParsedScript get parsedScript => widget.parsedScript;
 
+  // Used to ensure we don't update the scroll position when expanding or
+  // collapsing the file explorer.
+  ScriptRef _lastScriptRef;
+
   @override
   void initState() {
     super.initState();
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -145,18 +150,22 @@ class _CodeViewState extends State<CodeView>
...
-        verticalController.animateTo(
-          0,
-          duration: longDuration,
-          curve: defaultCurve,
-        );
-      } else {
-        verticalController.jumpTo(0);
+      // Don't scroll to top if we're just rebuilding the code view for the
+      // same script.
+      if (_lastScriptRef?.uri != scriptRef?.uri) {
```
kenzieschmoll, line 157: will this still scroll properly when selecting different members for the same file from the outline view?

bkonyi: Yup. This just makes sure that we don't try and reset the scroll position if we're just rebuilding the code view for the same script. When we select a member from the outline view, we don't actually trigger a rebuild of the code view (AFAIK).

kenzieschmoll (approved):

Issue #3361, opened 2021-09-14
Store connected app information in offline exports


kenzieschmoll:
Storing this information in the exported json will allow us to customize app dependent display features when the data is imported into devtools (for example, showing Flutter-specific filters in the CPU profiler).

PR #3397, opened 2021-09-27
Add connected app information to offline snapshots

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/3361.

This will allow for things like hiding flutter-specific features when the exported data was for a non-flutter app.

### DIFF
--- a/packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
+++ b/packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
@@ -5,6 +5,7 @@
 import 'dart:convert';
 
 import '../../../devtools.dart';
+import '../../connected_app.dart';
 import '../../globals.dart';
 import '../../notifications.dart';
 import '../../performance/legacy/performance_screen.dart';
@@ -18,6 +19,11 @@ import '_export_stub.dart'
 const devToolsSnapshotKey = 'devToolsSnapshot';
 const activeScreenIdKey = 'activeScreenId';
 const devToolsVersionKey = 'devtoolsVersion';
+const connectedAppKey = 'connectedApp';
+const isFlutterAppKey = 'isFlutterApp';
+const isProfileBuildKey = 'isProfileBuild';
+const isDartWebAppKey = 'isDartWebApp';
+const isRunningOnDartVMKey = 'isRunningOnDartVM';
 const flutterVersionKey = 'flutterVersion';
 const nonDevToolsFileMessage = 'The imported file is not a Dart DevTools file.'
     ' At this time, DevTools only supports importing files that were originally'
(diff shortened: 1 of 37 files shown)


packages/devtools_app/lib/src/example/conditional_screen.dart
```diff
@@ -93,9 +93,9 @@ class _ExampleConditionalScreenBodyState
 
   @override
   bool shouldLoadOfflineData() {
-    return offlineMode &&
-        offlineDataJson.isNotEmpty &&
-        offlineDataJson[ExampleConditionalScreen.id] != null;
+    return offlineController.offlineMode &&
```
jacob314, line 96: nit: seems like this could be a helper on offlineController.

packages/devtools_app/lib/src/performance/legacy/performance_screen.dart
```diff
@@ -321,10 +321,10 @@ class LegacyPerformanceScreenBodyState
 
   @override
   bool shouldLoadOfflineData() {
-    return offlineMode &&
-        offlineDataJson.isNotEmpty &&
-        offlineDataJson[LegacyPerformanceScreen.id] != null &&
-        offlineDataJson[LegacyPerformanceData.traceEventsKey] != null;
+    return offlineController.offlineMode &&
```
jacob314, line 324: nit: seems like room for a helper checking that you are in offline mode and that the screen id namtches.

packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
```diff
@@ -117,3 +133,25 @@ abstract class ExportController {
     return jsonEncode(_contents..addAll({activeScreenId: contents}));
   }
 }
+
+class OfflineModeController {
+  bool get offlineMode => _offlineMode;
```
jacob314, line 138: should this be a ValueListenable?
Existing logic seemed like it was calling setState on changes to this.

packages/devtools_app/lib/src/example/conditional_screen.dart
kenzieschmoll: done

packages/devtools_app/lib/src/performance/legacy/performance_screen.dart
kenzieschmoll: done

packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
kenzieschmoll: done

jacob314 (approved):
lgtm

PR #3398, opened 2021-09-27
Hide synthetic classes created by mixin applications in outline view

### LINKED ISSUE #3388: Hide Mixins from Outline view in the debugger
Mixins show up in the outline view, and selecting one of their properties navigates to the file where the mixin is defined. The feels like a jarring experience - perhaps we should hide Mixins all together.
mixin-file-explorer
@bkonyi

bkonyi:
Fixes #3388

### DIFF
--- a/packages/devtools_app/lib/src/debugger/program_explorer_model.dart
+++ b/packages/devtools_app/lib/src/debugger/program_explorer_model.dart
@@ -161,6 +161,10 @@ class VMServiceObjectNode extends TreeNode<VMServiceObjectNode> {
     }
 
     for (final clazz in lib.classes) {
+      // Don't surface synthetic classes created by mixin applications.
+      if (clazz.name.contains('&')) {
+        continue;
+      }
       if (clazz.location.script.uri == uri) {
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/debugger/program_explorer_model.dart
```diff
@@ -161,6 +161,10 @@ class VMServiceObjectNode extends TreeNode<VMServiceObjectNode> {
     }
 
     for (final clazz in lib.classes) {
+      // Don't surface synthetic classes created by mixin applications.
+      if (clazz.name.contains('&')) {
```
kenzieschmoll, line 165: if `class A with B, C` exists, will the name be `A&B&C`, and in that case, wouldn't we still want class `A` to be surfaced, even though we don't want `B` or `C`?

bkonyi: They actually show up as different classes. In your example, we would currently show the following class entries:

- `A`
- `A&B`
- `A&B&C`

kenzieschmoll (approved):

PR #3400, opened 2021-09-27
Fix issue where initial script was not always selected in File Explorer

bkonyi:


packages/devtools_app/lib/src/debugger/debugger_controller.dart
```diff
@@ -41,7 +41,12 @@ class DebuggerController extends DisposableController
       _showScriptLocation(ScriptLocation(scriptsHistory.current.value));
     };
     scriptsHistory.current.addListener(_scriptHistoryListener);
-    programExplorerController.initialize(currentScriptRef.value);
+    addAutoDisposeListener(currentScriptRef, () async {
+      if (!programExplorerController.initialized.value) {
+        await programExplorerController.initialize();
+      }
+      programExplorerController.selectScriptNode(currentScriptRef.value);
```
kenzieschmoll, line 48: should this be `currentScriptRef?.value`? Looks like `currentScriptRef` can be null according to `showScriptLocation`:

`_currentScriptRef.value = scriptLocation?.scriptRef;`

bkonyi: `currentScriptRef` is never null, but we should check for `currentScriptRef.value == null`

kenzieschmoll (approved):

Issue #3394, opened 2021-09-27
Widget icons package


tomgilder:
Hello, it was suggested that other developers would find it useful to have access to the icons for DevTools widgets.

We've turned the widgets into an icon font, and created a package for it: [https://github.com/CodemateLtd/widget_icons](https://github.com/CodemateLtd/widget_icons).

I've got quite a few questions about how to publish this though...

- Who should host and publish the package, Codemate or Google? To me it feels like something that should an official Google project, given it would be a key DevTools dependency.
- However, we would be happy to publish and maintain it, if you think that's appopriate.
- If it's published by Google, what are the next steps to get this done?
- We probably need a guide to updating the package. Again this might be better done internally within Google, following whatever your procedure is for Material/Cupertino icon fonts.

Happy to submit a PR switching DevTools to use this package once everything is decided.

Thanks!
Tom

cc @InMatrix

InMatrix:
Thanks for publishing this icons package. Cc: @csells @jacob314 for their opinions on ownership.

jacob314:
I'd prefer it if this was published under github.com/flutter/devtools/packages/widget_icons

There are already multiple packages under that directory and the mono-repo structure makes it convenient to iterate on any package under the directory when developing DevTools.

alexlindroos:
#3408 https://github.com/flutter/devtools/pull/3408

kenzieschmoll:
Did we decide who is going to be responsible for publishing this package, Google or Codemate? If we publish it, I would think we'd want the publisher to be flutter.dev as it is for the other devtools packages.

PR #3409, opened 2021-09-29
Widget Icons package added

alexlindroos:
Added widget_icons package under devtools/packages. I will update the icons and themes after this PR is merged.

https://github.com/flutter/devtools/issues/3394


kenzieschmoll:
We probably want this under `devtools/third_party/packages` instead of `devtools/packages`. See other third party packages we have there like codicon, ansi_up, etc.

alexlindroos:
@kenzieschmoll I moved the package to `devtools/third_party/packages`

packages/devtools_app/pubspec.yaml
```diff
@@ -129,3 +129,5 @@ dependency_overrides:
     path: ../devtools_shared #OVERRIDE_FOR_DEVELOPMENT
   devtools:  #OVERRIDE_FOR_DEVELOPMENT
     path: ../devtools  #OVERRIDE_FOR_DEVELOPMENT
+  widget_icons: #OVERRIDE_FOR_DEVELOPMENT
+    path: ../../third_party/packages/widget_icons  #OVERRIDE_FOR_DEVELOPMENT
```
kenzieschmoll, line 133: we should remove this. Once we publish 0.0.1 we can depend on it normally in the main dependency sections above. For now add a widget_icons dep in the mail dependencies section and depend on it from path there.

alexlindroos: Fixed

packages/devtools_app/pubspec.yaml
```diff
@@ -56,6 +56,8 @@ dependencies:
   vm_service: ^7.1.1
   vm_snapshot_analysis: ^0.6.0
   web_socket_channel: ^2.1.0
+  widget_icons:
+    path: ../../third_party/packages/widget_icons
```
kenzieschmoll, line 60: Actually, just write `widget_icons: ^0.0.1` and comment it out, since we aren't actually depending on the widget_icons package at all in this PR. Then we can publish and uncomment once this PR lands.

alexlindroos: Makes sense, I commented it out.

kenzieschmoll (approved):

PR #3421, opened 2021-10-01
Tweak positioning of the file picker

elliette:
Makes the file picker wider, and moves it a little to the left.

Screen Shot 2021-10-01 at 3 19 11 PM


packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -402,15 +402,18 @@ class _CodeViewState extends State<CodeView>
     );
   }
 
-  Widget wrapInElevatedCard(Widget widget) {
+  Widget wrapInElevatedCard(
+    Widget widget, {
+    double width = wideSearchTextWidth,
```
jacob314, line 407: this extra parameter is never passed in anywhere so should be removed.

packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -423,6 +426,7 @@ class _CodeViewState extends State<CodeView>
       FileSearchField(
         debuggerController: widget.controller,
       ),
+      width: extraWideSearchTextWidth,
```
jacob314, line 429: what effect does specifying this width and the width for the container achieve?

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Which extra parameter? The `width` parameter is being used on line 429 below

packages/devtools_app/lib/src/debugger/codeview.dart
elliette: Hmm I'm confused - this width is the width for the container (in`wrapInElevatedCard`)

kenzieschmoll (approved):

packages/devtools_app/lib/src/debugger/codeview.dart
jacob314: What confuses me is its child also has a width (line 416). I'd hope you could just specify one width for the outer parent and have all widths be computed based on that.

kenzieschmoll: you may need to expand out the few lines that are omitted in the git diff. This line (429) is part of the `buildFileSearchField` method, which calls `wrapInElevatedCard` with the parameter `width: extraWideSearchTextWidth,`.  The above use of `width` on line 416, is part of the `wrapInElevatedCard` method.

Issue #3412, opened 2021-09-29
inspector_service_test is very flaky


kenzieschmoll:
```
03:44 +244 ~4 -1: /home/runner/work/devtools/devtools/packages/devtools_app/test/inspector_service_test.dart: inspector service tests track widget creation on [E]                                     
  ext.flutter.inspector.getSelectedWidget: (-32000) Service connection disposed
  dart:async                                                         _Completer.completeError
  package:devtools_app/src/eval_on_dart_library.dart 558:20          EvalOnDartLibrary.addRequest.wrappedRequest
  ===== asynchronous gap ===========================
  dart:async                                                         _asyncErrorWrapperHelper
  package:devtools_app/src/eval_on_dart_library.dart                 EvalOnDartLibrary.addRequest.wrappedRequest
  package:devtools_app/src/eval_on_dart_library.dart 582:7           EvalOnDartLibrary.addRequest
  ===== asynchronous gap ===========================
  dart:async                                                         _asyncThenWrapperHelper
  package:devtools_app/src/inspector/inspector_service.dart 686:29   ObjectGroup._callServiceExtension
  package:devtools_app/src/inspector/inspector_service.dart 714:18   ObjectGroup.invokeServiceMethodDaemonParams
  package:devtools_app/src/inspector/inspector_service.dart 677:12   ObjectGroup.invokeServiceMethodDaemonArg
  package:devtools_app/src/inspector/inspector_service.dart 719:12   ObjectGroup.invokeServiceMethodDaemonInspectorRef
  package:devtools_app/src/inspector/inspector_service.dart 952:11   ObjectGroup.invokeServiceMethodReturningNodeInspectorRef
  package:devtools_app/src/inspector/inspector_service.dart 1050:30  ObjectGroup.getSelection
  package:devtools_app/src/inspector/inspector_service.dart 386:42   InspectorService.notifySelectionChanged
  package:devtools_app/src/inspector/inspector_service.dart 410:7    InspectorService.onDebugVmServiceReceived
  ===== asynchronous gap ===========================
  dart:async                                                         _StreamImpl.listen
  package:devtools_app/src/inspector/inspector_service.dart 74:45    new InspectorService
  package:devtools_app/src/service_manager.dart 303:27               ServiceConnectionManager.vmServiceOpened
  ===== asynchronous gap ===========================
  dart:async                                                         _asyncThenWrapperHelper
[...]

PR #3422, opened 2021-10-02
Dispose inspectorService on disconnect.

jacob314:
Fix case where inspectorService wasn't disposed even though the app had disconnected.

### DIFF
--- a/packages/devtools_app/lib/src/service_manager.dart
+++ b/packages/devtools_app/lib/src/service_manager.dart
@@ -345,7 +345,6 @@ class ServiceConnectionManager {
     _connectedState.value = connectionState;
     _connectionClosedController.add(null);
 
-    print('ZZZ vmServiceClosed');
     _inspectorService?.onIsolateStopped();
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -344,6 +344,11 @@ class ServiceConnectionManager {
 
     _connectedState.value = connectionState;
     _connectionClosedController.add(null);
+
+    print('ZZZ vmServiceClosed');
```
kenzieschmoll, line 348: remove print

kenzieschmoll (approved):

kenzieschmoll:
This fixes https://github.com/flutter/devtools/issues/3412, correct?

packages/devtools_app/lib/src/service_manager.dart
jacob314: > #3412
Correct.
https://github.com/flutter/devtools/issues/3412

PR #3426, opened 2021-10-04
Update release scripts

kenzieschmoll:
We need to use the version of flutter from `/flutter-sdk` for the `publish.sh` script (we already do this for the `build_release.sh` script), and the canvaskit version is now stored differently on flutter (https://github.com/flutter/engine/commit/057c49c4271c94f3b039b4f6fe3ac51a8f39f03c#diff-0569728ab50c0b8a004cd2d3eba60fb7b6990a090457fd23e5a930a0547e359f), so that changes how we need to look it up.

### DIFF
--- a/tool/build_release.sh
+++ b/tool/build_release.sh
@@ -27,23 +27,34 @@ fi
 set -ex
 
 # This avoids requiring an internet connection for CanvasKit at runtime.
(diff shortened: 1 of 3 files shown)


jacob314 (approved):
lgtm

PR #3432, opened 2021-10-07
Listen to timeline stream changes from the VM service and cleanup perf settings

kenzieschmoll:
- Completes a long standing TODO to listen to `kTimelineStreamSubscriptionsUpdate` events from the VM service and update timeline streams accordingly. 
- cleans up the UI to use more widget classes instead of `_buildXYZ` methods
- Simplifies `NotifierCheckbox` to use `ValueListenableBuilder` instead of maintaining its own state - also moves this class from `ui/utils.dart` to `common_widgets.dart`


packages/devtools_app/lib/src/performance/legacy/performance_screen.dart
```diff
@@ -403,15 +402,19 @@ class LegacyPerformanceSettingsDialog extends StatelessWidget {
     @required bool advanced,
   }) {
     final settings = <Widget>[];
-    final streams = controller.recordedStreams
-        .where((s) => s.advanced == advanced)
-        .toList();
+    final streams = advanced
```
kenzieschmoll, line 405: no need to do all the Widget code cleanup on the legacy perf screen

packages/devtools_app/lib/src/service_manager.dart
```diff
@@ -1322,61 +1329,6 @@ class ServiceExtensionState {
       );
 }
 
-class VmFlagManager extends Disposer {
```
kenzieschmoll, line 1325: moved this manager to the existing `vm_flags.dart` file so that `service_manager.dart` has more logically grouped classes.

packages/devtools_app/test/common_widgets_test.dart
```diff
@@ -116,4 +116,74 @@ void main() {
       expect(progressIndicator.value, equals(0.5));
     });
   });
+
+  group('NotifierCheckbox', () {
```
kenzieschmoll, line 120: these tests were moved from ui_utils_test

packages/devtools_app/test/vm_flags_test.dart
```diff
@@ -0,0 +1,52 @@
...
+// found in the LICENSE file.
+
+@TestOn('vm')
+import 'package:devtools_app/src/globals.dart';
+import 'package:devtools_test/flutter_test_driver.dart'
+    show FlutterRunConfiguration;
+import 'package:devtools_test/flutter_test_environment.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() async {
```
kenzieschmoll, line 12: moved these tests from service_manager_test.dart

packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -1388,3 +1388,116 @@ class CopyToClipboardControl extends StatelessWidget {
...
+    this.onChanged,
+  }) : super(key: key);
+
+  final ValueNotifier<bool> notifier;
+
+  final void Function(bool newValue) onChanged;
+
+  void _updateValue(bool value) {
+    notifier.value = value;
+    if (onChanged != null) {
```
jacob314, line 1416: probably only call onChanged if notifier.value != value.

kenzieschmoll: done

packages/devtools_app/lib/src/common_widgets.dart
```diff
@@ -1388,3 +1388,116 @@ class CopyToClipboardControl extends StatelessWidget {
...
+    );
+
+    if (description != null) {
+      textContent = Row(
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          textContent,
+          Expanded(
+            child: RichText(
+              overflow: TextOverflow.visible,
```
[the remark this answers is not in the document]
jacob314: https://stackoverflow.com/questions/60467972/how-to-use-an-ellipsis-overflow-with-multiline-text-in-flutter
I would suggest setting max lines to 2 for the best of both options.

jacob314 (approved):
lgtm

Issue #3439, opened 2021-10-08
Several 'unnecessary imports' in package:devtools_app


bleroux:
Last version of Dart analysis server reports several 'unnecessary imports' in `devtools_app` package files.
(I don't remember if this is a new lint ?)

PR #3440, opened 2021-10-08
Fix 'unnecessary import' warnings

bleroux:
Fixes https://github.com/flutter/devtools/issues/3439


kenzieschmoll (commented):
We should now be able to remove package:meta entry from the pubspec

Issue #2778, opened 2020-12-14
Expose performance toggles related to rendering in devtools


dnfield:
- [ ] [`debugDisableOpacityLayers`](https://github.com/flutter/flutter/blob/78910062997c3a836feee883712c241a5fd22983/packages/flutter/lib/src/rendering/debug.dart#L233)
- [ ] [`debugDisableClipLayers`](https://github.com/flutter/flutter/blob/78910062997c3a836feee883712c241a5fd22983/packages/flutter/lib/src/rendering/debug.dart#L208)
- [ ] [`debugDisablePhysicalShapeLayers`](https://github.com/flutter/flutter/blob/78910062997c3a836feee883712c241a5fd22983/packages/flutter/lib/src/rendering/debug.dart#L219)

There may be others in that file worth enabling if they're not already, but these three in particular (especially opacity and clip) can be helpful when debugging performance related issues.

/cc @jacob314 @kenzieschmoll

PR #3441, opened 2021-10-08
Add debug disable layer toggles to the Performance page

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/2778, in combination with https://github.com/flutter/flutter/pull/91499

In this PR, I also consolidated all the debugging options into a dialog that is launched by clicking a button "Debugging Options". Looking for some feedback on this layout. Optionally, we can use switches instead of checkboxes.

Screen Shot 2021-10-08 at 12 06 07 PM

Screen Shot 2021-10-08 at 12 06 46 PM

### DIFF
--- a/packages/devtools_app/test/performance_screen_test.dart
+++ b/packages/devtools_app/test/performance_screen_test.dart
@@ -65,7 +65,7 @@ void main() {
     expect(find.byType(PerformanceScreenBody), findsOneWidget);
   }
 
-  const windowSize = Size(2050.0, 1000.0);
+  const windowSize = Size(3000.0, 1000.0);
 
   group('PerformanceScreen', () {
     setUp(() async {


kenzieschmoll:
CC @InMatrix @cobblest for UX input on how to best expose these perf debugging features

jacob314:
I worry "debugging options" isn't an obvious. @cobblest & @InMatrix, any ideas for an alternate name for the options list to make it more discoverable? We already have the problem that too people notice "track widget rebuilds."

packages/devtools_app/lib/src/performance/performance_screen.dart
```diff
@@ -360,6 +363,45 @@ class _SecondaryControls extends StatelessWidget {
...
+        ),
+      ),
+      actions: [
+        DialogCloseButton(),
+      ],
+    );
+  }
+
+  Widget _maybeIncludeOption(ToggleableServiceExtensionDescription service) {
+    if (serviceManager.serviceExtensionManager
```
jacob314, line 397: I would instead make ServiceExtensionCheckbox itself show up as disabled or render as a SizedBox rather than adding the check here. 
Doing this here makes the screen fragile as if you happened to open the dialog right when the app is starting up or right after a hot restart, the checkboxes won't show up unless you close the dialog and reopen. If the logic is tied to the checkbox itself, it can listen for when the service extension becomes available to change its state as appropriate.

kenzieschmoll: done

kenzieschmoll:
@cobblest 
> I also noticed the last 3 checkbox are "Disable" options to "Ignore" some effects. To make it more consistent with other options and clearer to users, I am wondering if we could make them as "Enable" options. Unchecked box naturally means "disable" and we keep it unchecked by default if that's a typical setting for most users.

I am not completely opposed to changing these to "enable", though I will push back a bit with some thoughts from the developer perspective. The user action here is truly to disable a layer for debugging purposes only, which I think may be better communicated by having "disable" in the action title. If "enable" is in the title, we need to make sure we are communicating that each of these rendering layers are enabled by default with flutter, and that they are not settings that DevTools is manually enabling, or that a user can expect to turn off to improve performance in a release build. We do have some other areas in DevTools where the setting action is phrased in the negative, like "hiding" certain classes or code in memory snapshots and CPU profiles, so this is not completely inconsistent with other DevTools settings.

@ayanogawa 
You can read what each of these toggles do in [rendering/debug.dart](https://github.com/flutter/flutter/blob/0d3b44e835b04249610b32f2a3f71bf5e819cecd/packages/flutter/lib/src/rendering/debug.dart#L153-L188) from the Flutter framework, originally added in https://github.com/flutter/flutter/pull/15548.

Looping in someone from the framework to fact check me on these answers: @goderbauer 

> Does widget tree = layer tree?

From my understanding, no. The [layer tree](https://api.flutter.dev/flutter/rendering/Layer-class.html) is generated during paint. 

> Does clip layers / clipping effects mean Clip Behavior?

Yes. Anything that uses a [ClipRectLayer](https://api.flutter.dev/flutter/rendering/ClipRectLayer-class.html), [ClipRRectLayer](https://api.flutter.dev/flutter/rendering/ClipRRectLayer-class.html), or a [ClipPathLayer](https://api.flutter.dev/flutter/rendering/ClipPathLayer-class.html).

> Does opacity layers mean Opacity class?

This applies to anything with an [OpacityLayer](https://api.flutter.dev/flutter/rendering/OpacityLayer-class.html). The Opacity widget is one widget that uses an OpacityLayer for rendering, but there are also others. A list could be put together by tracing the code references in the framework, but there are several places where OpacityLayer is used.

> Does physical shape layers mean PhysicalModel or PhysicalShape? 

Yes. This applies to anything that uses a [PhysicalModelLayer](https://api.flutter.dev/flutter/rendering/PhysicalModelLayer-class.html), which include RenderPhysicalModel and RenderPhysicalShape.

I do have a concern with using the terminology "filter" as these options will not filter data that is currently visible in the performance page, but rather will affect how Flutter is rendering your app. To see a difference after turning on one of these debug toggles, the user will have to interact with their app again to reproduce the activity they want to profile. Communicating this required "reproducing" behavior has been a sticking point with Track Widget Builds as well, since users do not know that they have to reproduce activity in their app to get widget build trace events to show up.

cobblest:
In general UX principles, checkbox should be used with positive phrasing. For example, use do instead of do not, and notify instead of do not notify: 
https://docs.microsoft.com/en-us/windows/win32/uxguide/text-ui#control-labels
https://ux.stackexchange.com/questions/5748/checkbox-label-negating

The reason is that a checked checkbox means "Yes", and users would need to wrap their head around the negative phrase "Disabled" to decide if the actual state is "on" or "off". I guess that's why I got confused the first time I saw the design.

Issue #3447, opened 2021-10-13
Enable Outline view in `ProgramExplorer` for web-based applications


bkonyi:
Requires the following issues to be resolved before this functionality can be enabled for web targets:

- [ ] https://github.com/dart-lang/webdev/issues/1427
- [ ] https://github.com/dart-lang/webdev/issues/1428

PR #3448, opened 2021-10-13
Reland `ProgramExplorer`

bkonyi:
Take two at landing the `ProgramExplorer`. Libraries, scripts, and classes are now lazily loaded and the outline view is disabled for web targets (see https://github.com/flutter/devtools/issues/3447).


packages/devtools_app/lib/src/debugger/program_explorer_model.dart
```diff
@@ -206,11 +133,7 @@ class VMServiceObjectNode extends TreeNode<VMServiceObjectNode> {
         );
         await controller.populateNode(node);
         root.addChild(
-          VMServiceObjectNode(
-            controller,
-            field.name,
-            field,
-          ),
+          node,
```
kenzieschmoll, line 136: nit: no need for trailing comma if there is only one param

packages/devtools_app/lib/src/debugger/program_explorer.dart
```diff
@@ -377,32 +378,53 @@ class ProgramExplorer extends StatelessWidget {
         if (!initialized) {
           body = const CenteredCircularProgressIndicator();
         } else {
+          const fileExplorerHeader = AreaPaneHeader(
```
kenzieschmoll, line 381: either remove this const since it is only used once, or keep it and change `needsTopBorder` to `serviceManager.connectedApp.isDartWebAppNow` so it can be used as part of the `Column` and the `FlexSplitColumn`

Issue #3095, opened 2021-06-07
Autcomplete polish burndown list


jacob314:
Autocomplete is close to being done but there a few more edge cases that should be polished.
![autocomplete_demo](https://user-images.githubusercontent.com/1226812/121062950-94201480-c77a-11eb-82e9-b1774cdb8aa5.gif)
- [x] The autocomplete card should be aligned with the period, not the cursor. This is how Chrome and other IDEs render their autocomplete options.
- [x] The current autocomplete match should be show in grey directly in the text box as well as in the autocomplete list.
- [x] Autocomplete is slow for very large classes https://github.com/flutter/devtools/issues/3089
- [x] Reduce padding between list items.

TimWhiting:
I would suggest having less padding as well for the list items. (Similar spacing to VSCode or Intellij).

PR #3449, opened 2021-10-13
Expression evaluation autocomplete overlay is positioned over the last `.` in the expression

elliette:
- Previously we had a boolean parameter `tracking` which if `true` would display the autocomplete overlay over the cursor
- Instead, clients now provide a callback function that determines the X-position of the overlay. For the case of the expression evaluation autocomplete, this callback returns the position of the last `.` in the expression (if there is a `.`). 

**Demo:**

expression evaluation 2

This PR is part of the work towards https://github.com/flutter/devtools/issues/3095

### DIFF
--- a/packages/devtools_app/lib/src/analytics/constants.dart
+++ b/packages/devtools_app/lib/src/analytics/constants.dart
@@ -55,6 +55,9 @@ const refreshTimelineEvents = 'refreshTimelineEvents';
 const timelineFlameChartHelp = 'timelineFlameChartHelp';
 const selectFlutterFrame = 'selectFlutterFrame';
 const traceEventProcessingTime = 'traceEventProcessingTime';
+const disableClipLayersOption = 'disableClipLayers';
+const disableOpacityLayersOption = 'disableOpacityLayers';
+const disablePhysicalShapeLayersOption = 'disablePhysicalShapeLayers';
 
 // CPU profiler UX actions:
 const profileGranularityPrefix = 'profileGranularity';
--- a/packages/devtools_app/lib/src/common_widgets.dart
+++ b/packages/devtools_app/lib/src/common_widgets.dart
@@ -596,7 +596,7 @@ class DevToolsTooltip extends StatelessWidget {
     @required this.child,
     this.waitDuration = tooltipWait,
     this.preferBelow = false,
-    this.padding,
+    this.padding = const EdgeInsets.all(defaultSpacing),
     this.decoration,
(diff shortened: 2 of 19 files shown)


packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -556,6 +556,11 @@ typedef ClearSearchField = Function(
   bool force,
 });
 
+/// Provided by clients to specify where the autocomplete overlay should be
+/// positioned relative to the input text.
+typedef DetermineOverlayXPosition = double Function(
+    String inputValue, TextStyle inputStyle);
```
kenzieschmoll, line 562: nit: trailing comma

packages/devtools_app/lib/src/debugger/evaluate.dart
```diff
@@ -135,14 +136,27 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
                 border: OutlineInputBorder(),
                 focusedBorder: OutlineInputBorder(borderSide: BorderSide.none),
                 enabledBorder: OutlineInputBorder(borderSide: BorderSide.none),
                 labelText: 'Eval',
               ),
+              determineOverlayXPosition:
+                  (String inputValue, TextStyle inputStyle) {
+                // X-coordinate is equivalent to the width of the input text
+                // up to the last "." or the insertion point (cursor):
+                final textSegment = inputValue.contains('.')
```
jacob314, line 150: nit: slightly cleaner is:
```dart
final indexOfDot = inputValue.lastIndexOf('.')
final textSegment = indexOfDot != -1 ? inputValue.substring(0, indexOfDot + 1) : inputValue;
```
A little better as it doesn't have two different expressions with the same goal (contains and lastIndexOf) and is a little faster due to less dupe work not that it matters in this particular case.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -556,6 +556,11 @@ typedef ClearSearchField = Function(
   bool force,
 });
 
+/// Provided by clients to specify where the autocomplete overlay should be
+/// positioned relative to the input text.
+typedef DetermineOverlayXPosition = double Function(
```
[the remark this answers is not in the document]
elliette: Done, renamed. Still using the typedef since it's being used twice

jacob314 (commented):
lgtm

PR #3451, opened 2021-10-14
Add Track Paints and Track Layouts toggles to the performance page.

kenzieschmoll:
Screen Shot 2021-10-14 at 10 55 26 AM

IMO Adding these toggles crowds the button. Here a couple ideas:

1. Never show text in Pause, Resume, Clear, and Export buttons. Text could be visible in the tooltip still.
2. Put all the "Track XYZ" buttons into a dialog. The button to launch the dialog could say something like "Enhance Tracing" or similar, and in the dialog, the options to track builds, layouts, and paints would be accessible.

Thoughts? @cobblest @InMatrix @ayanogawa

When these new options are enabled, timeline events are added to for RenderObject paints and layouts respectively.
Screen Shot 2021-10-14 at 10 53 56 AM
CC @goderbauer 

Depends on https://github.com/flutter/flutter/pull/91822.


packages/devtools_app/lib/src/performance/performance_screen.dart
```diff
@@ -375,6 +380,206 @@ class _SecondaryControls extends StatelessWidget {
...
+            ),
+          ],
+          selectedStates: [tracingEnhanced],
+          onPressed: (_) => _insertOverlay(context),
+        );
+      },
+    );
+  }
+
+  void _insertOverlay(BuildContext context) {
```
jacob314, line 461: maybe comment on the effect you are trying to achieve by adding this overlay like this.

packages/devtools_app/lib/src/performance/performance_screen.dart
```diff
@@ -375,6 +380,206 @@ class _SecondaryControls extends StatelessWidget {
...
+}
+
+class _EnhanceTracingButtonState extends State<EnhanceTracingButton>
+    with AutoDisposeMixin {
+  static const _hoverYOffset = 10.0;
+
+  final _tracingServiceExtensions = [
+    profileWidgetBuilds,
+    profileRenderObjectLayouts,
+    profileRenderObjectPaints,
```
jacob314, line 397: this list occurs twice. would be good to avoid that to prevent the UI from getting out of sync. I would drive the list of actual buttons from this list of extensions maybe adding some extra fields to the extension object data classes as needed.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/performance/performance_screen.dart
```diff
@@ -375,6 +380,206 @@ class _SecondaryControls extends StatelessWidget {
...
+
+  OverlayEntry _enhanceTracingOverlay;
+
+  bool _enhanceTracingOverlayHovered = false;
+
+  @override
+  void initState() {
+    super.initState();
+    _extensionStates =
+        List.generate(_tracingServiceExtensions.length, (index) => false);
```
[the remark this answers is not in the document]
kenzieschmoll: done

Issue #3457, opened 2021-10-18
Problems with the debugger empty state page


elliette:
The empty state page (see image below) shows up when selecting a file fails. When that happens, no keyboard shortcuts work, so opening with command+P (as the empty state suggests) doesn't work. 

We should:
1) Add a button to the empty state that will trigger the file opener (this can be a stop gap until we fix 2 below) 
2) Figure out why the empty state shows up in the first place and why keyboard shortcuts don't work. 

Unfortunately the empty state page only shows up sporadically so it's hard to reproduce.

![Screen Shot 2021-10-18 at 10 24 56 AM](https://user-images.githubusercontent.com/21270878/137780223-740f8828-eb3e-4f2c-97f4-52e650cd74bf.png)

PR #3458, opened 2021-10-18
Keyboard shortcuts are set on the top-level scaffold

elliette:
Work towards https://github.com/flutter/devtools/issues/3457

Moves keyboard shortcuts to the top-level scaffold so that they can be triggered for a particular screen no matter where the focus is.


packages/devtools_app/lib/src/debugger/codeview.dart
```diff
@@ -170,34 +170,42 @@ class _CodeViewState extends State<CodeView>
...
-                    child: buildSearchInFileField(),
-                  ),
-              ],
-            );
-          },
-        );
+    return MouseRegion(
+      onEnter: (_) {
+        if (!focusNode.hasFocus) {
+          focusNode.requestFocus();
```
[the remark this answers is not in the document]
jacob314: We should find a different way to fix this.  Mouse hover should not change focus.

packages/devtools_app/lib/src/screen.dart
```diff
@@ -304,3 +308,20 @@ class BadgePainter extends CustomPainter {
...
+class ShortcutsConfiguration {
+  ShortcutsConfiguration({
+    @required this.shortcuts,
+    @required this.actions,
+  });
+
+  final Map<ShortcutActivator, Intent> shortcuts;
+  final Map<Type, Action<Intent>> actions;
+
+  static ShortcutsConfiguration empty =
```
kenzieschmoll, line 321: make this a factory constructor

packages/devtools_app/lib/src/screen.dart
```diff
@@ -67,6 +67,10 @@ abstract class Screen {
   /// Whether to show the console for this screen.
   bool showConsole(bool embed) => false;
 
+  // Which keyboard shortcuts shoud be enabled for this screen.
```
kenzieschmoll, line 70: nit: use /// instead of // for dartdoc

packages/devtools_app/lib/src/screen.dart
```diff
@@ -304,3 +308,20 @@ class BadgePainter extends CustomPainter {
     return true;
   }
 }
+
+class ShortcutsConfiguration {
+  ShortcutsConfiguration({
+    @required this.shortcuts,
+    @required this.actions,
+  });
```
kenzieschmoll, line 316: assert that shortcuts.length == actions.length

packages/devtools_app/lib/src/scaffold.dart
```diff
@@ -340,28 +363,8 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
...
-                        splitters: [
-                          DebuggerConsole.buildHeader(),
-                        ],
-                        initialFractions: const [0.8, 0.2],
-                      )
-                    : content,
-                bottomNavigationBar: widget.embed ? null : _buildStatusLine(),
-              ),
+              child: _maybeWrapInShortcuts(
+                  scaffold, _currentScreen.keyboardShortcuts(context)),
```
kenzieschmoll, line 367: nit: trailing comma

packages/devtools_app/lib/src/scaffold.dart
```diff
@@ -470,6 +473,22 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
     );
   }
 
+  Widget _maybeWrapInShortcuts(
```
kenzieschmoll, line 476: instead of making this a method, create a widget class KeyboardShortcuts that takes required named params. And you can change `widget` to `child` to match flutter naming conventions. The KeyboardShortcuts widget can handle the logic here in its build method

packages/devtools_app/lib/src/scaffold.dart
```diff
@@ -326,6 +326,29 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
       ],
     );
     final theme = Theme.of(context);
+    final scaffold = Scaffold(
```
kenzieschmoll, line 329: instead of breaking this out into its own var, put back inline in the return statement and see my comment below regarding KeyboardShortcuts. Nesting widgets directly is preferred for build methods when possible. There are some cases where we need to break widgets out into local variables, but in this case, the var `scaffold` is only being used once, so we can leave directly in the widget tree.

packages/devtools_app/lib/src/debugger/debugger_screen.dart
```diff
@@ -44,6 +44,24 @@ class DebuggerScreen extends Screen {
   @override
   bool showConsole(bool embed) => true;
 
+  @override
+  ShortcutsConfiguration keyboardShortcuts(BuildContext context) {
```
jacob314, line 48: nit: maybe call this buildKeyboardShortcuts?

packages/devtools_app/lib/src/screen.dart
```diff
@@ -304,3 +308,20 @@ class BadgePainter extends CustomPainter {
...
+
+  final Map<ShortcutActivator, Intent> shortcuts;
+  final Map<Type, Action<Intent>> actions;
+
+  static ShortcutsConfiguration empty =
+      ShortcutsConfiguration(shortcuts: {}, actions: {});
+
+  bool isEmpty() {
```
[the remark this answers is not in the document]
elliette: Done, switched to getter

packages/devtools_app/lib/src/debugger/debugger_screen.dart
elliette: Done

packages/devtools_app/lib/src/scaffold.dart
elliette: Sg, done!

packages/devtools_app/lib/src/scaffold.dart
elliette: Makes sense, switched it to a widget

packages/devtools_app/lib/src/screen.dart
elliette: Good to know! Switched to ///

packages/devtools_app/lib/src/screen.dart
elliette: Let me know what you think of it now - it's a factory constructor, and it's passed to the widget

PR #3461, opened 2021-10-20
Cleanup "More Debugging Options" button on Performance page

kenzieschmoll:
This PR removes the dialog that was used for these settings before, and uses the same overlay that is used for the "Enhance Tracing" button. The code for this overlay is now generalized for use with any set of service extensions and lives in service_extension_widgets.dart.
Screen Shot 2021-10-20 at 3 36 22 PM

### DIFF
--- a/packages/devtools_app/lib/src/performance/performance_screen.dart
+++ b/packages/devtools_app/lib/src/performance/performance_screen.dart
@@ -4,7 +4,6 @@
 
 import 'dart:async';
 
-import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
 
@@ -24,7 +23,6 @@ import '../split.dart';
 import '../theme.dart';
 import '../ui/colors.dart';
 import '../ui/icons.dart';
-import '../ui/label.dart';
 import '../ui/service_extension_widgets.dart';
 import '../ui/vm_flag_widgets.dart';
 import '../version.dart';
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/performance/performance_screen.dart
```diff
@@ -380,244 +364,70 @@ class _SecondaryControls extends StatelessWidget {
   }
 }
 
-class EnhanceTracingButton extends StatefulWidget {
+class EnhanceTracingButton extends StatelessWidget {
```
jacob314, line 367: always nice to see fewer stateless widgets!

packages/devtools_app/lib/src/performance/performance_screen.dart
```diff
@@ -380,244 +364,70 @@ class _SecondaryControls extends StatelessWidget {
...
+        disablePhysicalShapeLayers,
       ],
+      overlayDescription: Text(
+        'When toggling on/off a rendering layer, you will need '
+        'to reproduce activity in your app to see the effects of the '
+        'debugging option. All layers are rendered by default - disabling a '
+        'layer may help you identify expensive operations in your app.',
+        style: Theme.of(context).subtleTextStyle,
+      ),
+      overlayWidth: _width,
```
jacob314, line 430: why is this width not scaled by the font scale factor?

packages/devtools_app/lib/src/ui/service_extension_widgets.dart
```diff
@@ -508,6 +509,241 @@ class _ServiceExtensionCheckboxState extends State<ServiceExtensionCheckbox>
   }
 }
 
+/// A button that, when pressed, will display an overlay directly below that has
+/// a list of service extension checkbox settings.
+class ServiceExtensionCheckboxOverlayButton extends StatefulWidget {
```
jacob314, line 514: The name doesn't capture that it is for multiple service extensions rather than one.

How about ServiceExtensionCheckboxGroupButton ?

packages/devtools_app/lib/src/performance/performance_screen.dart
kenzieschmoll: done

Issue #3434, opened 2021-10-07
service manager related test flakes


kenzieschmoll:
This same assertion error for `'isolate != null'` has been seen on other tests as well.
```
02:58 +241 ~3: /home/runner/work/devtools/devtools/packages/devtools_app/test/inspector_service_test.dart: inspector service tests track widget creation on                                            
PreferencesController: storage not initialized

02:59 +241 ~3 -1: /home/runner/work/devtools/devtools/packages/devtools_app/test/inspector_service_test.dart: inspector service tests track widget creation on [E]                                     
  'package:devtools_app/src/service_manager.dart': Failed assertion: line 489 pos 12: 'isolate != null': is not true.
  dart:core                                             _AssertionError._throwNew
  package:devtools_app/src/service_manager.dart 489:12  ServiceConnectionManager.libraryUriAvailableNow
  package:devtools_app/src/service_manager.dart 499:12  ServiceConnectionManager.libraryUriAvailable
  
Retry: inspector service tests track widget creation on
  'package:devtools_app/src/inspector/inspector_service.dart': Failed assertion: line 53 pos 16: 'serviceManager.connectedAppInitialized': is not true.
  dart:core                                                        _AssertionError._throwNew
  package:devtools_app/src/inspector/inspector_service.dart 53:16  new InspectorService
  test/inspector_service_test.dart 26:24                           main.<fn>
  test/inspector_service_test.dart 25:25                           main.<fn>
  package:devtools_test/flutter_test_environment.dart 107:57       FlutterTestEnvironment.setupEnvironment
  test/inspector_service_test.dart 44:19                           main.<fn>.<fn>
  test/inspector_service_test.dart 43:40                           main.<fn>.<fn>
  ```

PR #3474, opened 2021-10-28
Fix test flakes with the service manager.

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/3434.


jacob314 (approved):
Fyi @elliette: this might impact debugger page startup negatively.
lgtm

PR #3476, opened 2021-10-29
Don't register service worker when running DevTools locally

elliette:
Adds a check to see if DevTools is running locally in `index.html`, which is determined by:
* the host name is `localhost` OR
* the `index.html` file is called `index.ddc.html` 

If so, we skip the service worker registration step (before, this would fail after 4 seconds). 

I've also caught any error coming from registering the service worker, and logging it to the console as a warning.

### DIFF
--- a/packages/devtools_app/web/index.html
+++ b/packages/devtools_app/web/index.html
@@ -72,8 +72,7 @@
     }
 
     if ('serviceWorker' in navigator) {
-      if (window.location.hostname === 'localhost' ||
-          window.location.href.includes('index.ddc.html')) {
+      if (window.location.href.includes('index.ddc.html')) {
         // If we are running DevTools locally, immediately load `main.dart.js` and
         // unregister any service workers:
         loadMainDartJs();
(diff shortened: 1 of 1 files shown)


packages/devtools_app/web/index.html
```diff
@@ -72,51 +72,69 @@
...
-        // Wait for registration to finish before dropping the <script> tag.
-        // Otherwise, the browser will load the script multiple times,
-        // potentially different versions.
-        var serviceWorkerUrl = 'service_worker.js?v=' + version;
-        navigator.serviceWorker.register(serviceWorkerUrl)
-          .then((reg) => {
-            function waitForActivation(serviceWorker) {
-              serviceWorker.addEventListener('statechange', () => {
-                if (serviceWorker.state == 'activated') {
+      if (window.location.hostname === 'localhost' ||
```
jacob314, line 75: checking for localhost doesn't seem safe. Could easily be running devtools from localhost in a release build.

elliette: SG, modified to just check for `index.ddc.html`

jacob314 (approved):
lgtm

PR #3477, opened 2021-10-30
Preferences keep split fractions base on screen and children

quyenvsp:
Save split panel size after changed to preferences, so users can resize panel and do not need do it again each time run new Devtools instance.

Storage key is string base on current page path and hashCode of list children widget. Make it a bit unique and do not conflict with other Split / Page.

By this way I just need start and it will auto show my favorites layout, perfect! (Second monitor vertical, android emulator always on top)

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/devtools_app/lib/src/preferences.dart
+++ b/packages/devtools_app/lib/src/preferences.dart
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:convert';
+
 import 'package:flutter/foundation.dart';
 
 import 'config_specific/logger/logger.dart';
@@ -13,10 +15,12 @@ class PreferencesController {
   final ValueNotifier<bool> _darkModeTheme = ValueNotifier(true);
   final ValueNotifier<bool> _vmDeveloperMode = ValueNotifier(false);
   final ValueNotifier<bool> _denseMode = ValueNotifier(false);
+  final ValueNotifier<String> _splitFractions = ValueNotifier('{}');
 
   ValueListenable<bool> get darkModeTheme => _darkModeTheme;
   ValueListenable<bool> get vmDeveloperModeEnabled => _vmDeveloperMode;
   ValueListenable<bool> get denseModeEnabled => _denseMode;
+  ValueListenable<String> get splitFractions => _splitFractions;
 
   Future<void> init() async {
     if (storage != null) {
@@ -37,6 +41,13 @@ class PreferencesController {
       toggleDenseMode(value == 'true');
       _denseMode.addListener(() {
         storage.setValue('ui.denseMode', '${_denseMode.value}');
+      });
+
+      final String _splitFractionsValue =
+          await storage.getValue('ui.splitFractions');
+      setSplitFractions(_splitFractionsValue);
+      _splitFractions.addListener(() {
+        storage.setValue('ui.splitFractions', '${_splitFractions.value}');
       });
     } else {
       // This can happen when running tests.
@@ -59,5 +70,19 @@ class PreferencesController {
   /// Change the value for the dense mode setting.
   void toggleDenseMode(bool enableDenseMode) {
     _denseMode.value = enableDenseMode;
+  }
+
+  /// Change the value for the split fractions setting.
+  void setSplitFractions(String splitFractions) {
+    dynamic decoded;
+    try {
+      decoded = jsonDecode(splitFractions);
+    } catch (e) {
+      decoded = false;
+    }
+    _splitFractions.value = ((splitFractions?.isNotEmpty ?? false) &&
+            decoded is Map<String, dynamic>)
+        ? splitFractions
+        : '{}';
   }
 }
--- a/packages/devtools_app/lib/src/preferences.dart
+++ b/packages/devtools_app/lib/src/preferences.dart
@@ -12,15 +12,16 @@ import 'vm_service_wrapper.dart';
 
 /// A controller for global application preferences.
 class PreferencesController {
+  Map<String, List<double>> _initialFractions = {};
+
   final ValueNotifier<bool> _darkModeTheme = ValueNotifier(true);
   final ValueNotifier<bool> _vmDeveloperMode = ValueNotifier(false);
   final ValueNotifier<bool> _denseMode = ValueNotifier(false);
-  final ValueNotifier<String> _splitFractions = ValueNotifier('{}');
 
   ValueListenable<bool> get darkModeTheme => _darkModeTheme;
   ValueListenable<bool> get vmDeveloperModeEnabled => _vmDeveloperMode;
   ValueListenable<bool> get denseModeEnabled => _denseMode;
-  ValueListenable<String> get splitFractions => _splitFractions;
+  final splitFractions = <String, ValueListenable<List<double>>>{};
 
   Future<void> init() async {
     if (storage != null) {
@@ -43,12 +44,8 @@ class PreferencesController {
         storage.setValue('ui.denseMode', '${_denseMode.value}');
       });
 
-      final String _splitFractionsValue =
-          await storage.getValue('ui.splitFractions');
-      setSplitFractions(_splitFractionsValue);
-      _splitFractions.addListener(() {
-        storage.setValue('ui.splitFractions', '${_splitFractions.value}');
-      });
+      _initialFractions = fractionsFromJson(
+          jsonDecode(await storage.getValue('ui.splitFractions') ?? '{}'));
     } else {
       // This can happen when running tests.
       log('PreferencesController: storage not initialized');
@@ -72,17 +69,33 @@ class PreferencesController {
     _denseMode.value = enableDenseMode;
   }
 
-  /// Change the value for the split fractions setting.
-  void setSplitFractions(String splitFractions) {
-    dynamic decoded;
-    try {
-      decoded = jsonDecode(splitFractions);
-    } catch (e) {
-      decoded = false;
-    }
-    _splitFractions.value = ((splitFractions?.isNotEmpty ?? false) &&
-            decoded is Map<String, dynamic>)
-        ? splitFractions
-        : '{}';
+  /// Get the value for the split fractions.
+  ValueNotifier<List<double>> lookupSplitFractions(String key) {
+    return splitFractions.putIfAbsent(
+        key, () => ValueNotifier<List<double>>(_initialFractions[key] ?? []));
+  }
+
+  /// Update the value for the split fractions.
+  void updateSplitFractions(String key, List<double> value) {
+    if ((key?.isEmpty ?? true) ||
+        (value?.length ?? 0) < 2 ||
+        isListEqual(lookupSplitFractions(key).value, value)) return;
+    lookupSplitFractions(key).value = value;
+    storage?.setValue(
+        'ui.splitFractions', jsonEncode(fractionsToJson(splitFractions)));
+  }
+
+  Map<String, List<double>> fractionsFromJson(Map<String, dynamic> json) {
+    return json
+        .map((key, fractions) => MapEntry(key, fractions?.cast<double>()));
+  }
+
+  Map<String, List<double>> fractionsToJson(
+      Map<String, ValueListenable<List<double>>> fractions) {
+    return fractions.map((key, fractions) => MapEntry(key, fractions.value));
+  }
+
+  bool isListEqual(List value, List other) {
+    return jsonEncode(value) == jsonEncode(other);
   }
 }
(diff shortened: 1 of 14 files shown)


packages/devtools_app/lib/src/split.dart
```diff
@@ -102,13 +108,42 @@ class _SplitState extends State<Split> {
...
+
   Widget _buildLayout(BuildContext context, BoxConstraints constraints) {
     final width = constraints.maxWidth;
     final height = constraints.maxHeight;
     final axisSize = isHorizontal ? width : height;
 
     final availableSize = axisSize - _totalSplitterSize();
 
+    final routerDelegate = DevToolsRouterDelegate.of(context);
+    String childrenRuntimeType = '';
```
jacob314, line 131: this is very clever. I like using the current page as part of the key. I'm not sure about using the types of the children widgets. Instead, I'd prefer to make specifying a name for each splitter required and using that name as part of the key.
In either case, remove the hashCode and just include the entire string in the key.

quyenvsp: I use list children as storage key because already have many widget use Split and don't want pass 'storageKey' argument to them all.
This way helpful in case Split has dynamic children since Split required children length == fractions length.
So if list children widget of Split change then fractions need change too.

In case save with passed 'storageKey' string, when list children change then saved fractions length can not equal children length (storageKey still same if you not dynamic change), it lead to can not use saved fractions for this Split.
Also resize Split override saved fractions (new and old different length) => when old list children show it will can not use again and will override again more time.
=> If save with 'storageKey' argument you will need handle this argument each time list children widget change for avoid conflict.

Are you think this worthy for use list children name instead of pass storageKey argument? If not I will try make specifying a name for each splitter.

About hashCode, yes your suggest good. Before I use it for 'shorten' string because can have long storage key 'widget1_widget2_widget3_widget4...' but infact Split just have some children, so It good to not use hashCode.

jacob314: I think this change is worthy of landing as long as we switch to using a split names. This is definitely solving a problem users care about fixing.
I'm fine with passing in the # of split children after the split name to ensure there aren't conflicts if the # of children changes.

packages/devtools_app/lib/src/split.dart
```diff
@@ -102,13 +108,42 @@ class _SplitState extends State<Split> {
...
+    childrenRuntimeType +=
+        widget.children.map((e) => '_${e.runtimeType.toString()}').join();
+    final String fractionsStorageKey =
+        '${routerDelegate.currentConfiguration?.page}_hash${childrenRuntimeType.hashCode}';
+    final List<dynamic> fractionsStorageData = (fractionsStorage != null &&
```
jacob314, line 136: remove the List<dynamic> or change it to List<double>

packages/devtools_app/lib/src/split.dart
```diff
@@ -262,6 +298,10 @@ class _SplitState extends State<Split> {
                   isHorizontal ? updateSpacing(details, i) : null,
               onVerticalDragUpdate: (details) =>
                   isHorizontal ? null : updateSpacing(details, i),
+              onHorizontalDragEnd: (detail) =>
+                  preferences.setSplitFractions(jsonEncode(fractionsStorage)),
```
jacob314, line 302: add a helper method called _saveSplitFractions()

packages/devtools_app/lib/src/split.dart
```diff
@@ -102,13 +108,42 @@ class _SplitState extends State<Split> {
...
   }
 
+  @override
+  void dispose() {
+    preferences.splitFractions.removeListener(onChangeSplitFractions);
+    super.dispose();
+  }
+
+  void onChangeSplitFractions() {
+    setState(() {
```
jacob314, line 118: add a helper to perform all the json decoding and updating fractions.  That logic is currently duplicated and we probably want to use a slightly different decoding scheme + check to make sure the decoded fractions aren't identical to the existing fractions. That would avoid possible infinite loops repeatedly decoding and encoding and performing extra rebuilds.

packages/devtools_app/lib/src/preferences.dart
```diff
@@ -13,10 +15,12 @@ class PreferencesController {
   final ValueNotifier<bool> _darkModeTheme = ValueNotifier(true);
   final ValueNotifier<bool> _vmDeveloperMode = ValueNotifier(false);
   final ValueNotifier<bool> _denseMode = ValueNotifier(false);
+  final ValueNotifier<String> _splitFractions = ValueNotifier('{}');
```
jacob314, line 18: It would be cleaner to have a separate
ValueNotifier<String> or better yet ValueNotifier<List<double>>. for each named split rather than one that returns a value that is JSON for all splits.
With the current approach, the size of the json blob will continue increasing as more splits are added to the app. There are only a couple splits on each page so not that many separate blobs will be loaded.

jacob314: A reasonable api would be 
```dart
ValueListenable<List<double> lookupSplitFractions(String key) {
  return _splitFractions.putIfAbsent(key, () { 
    final listenable = ValueListenable<List<double>>;
    // do work to listen on a preference with name 'splitters_$key' or similar
    // ....
    return listenable;
  })
}

Map<String, ValueListenable<List<double>>> _splitFractions;
```

jacob314: Actually never mind. Your current approach is probably better as it would be hard to make this approach sync.

jacob314: Ok here is an approach that would be cleaner but not require big changes.
```dart
ValueListenable<List<double> lookupSplitFractions(String key) {
  return _splitFractions.putIfAbsent(key, () { 
    final listenable = ValueListenable<List<double>>(_initialFractions[key]);
    return listenable;
  })
}

Map<String, ValueListenable<List<double>>> _splitFractions;
// Fill out _initialFractions in the init() method.
Map<String, List<double>> _initialFractions;
```

packages/devtools_app/lib/src/split.dart
```diff
@@ -26,10 +28,12 @@ class Split extends StatefulWidget {
     @required this.initialFractions,
     this.minSizes,
     this.splitters,
+    this.fractionsKey,
```
kenzieschmoll, line 31: make this required

packages/devtools_app/lib/src/preferences.dart
```diff
@@ -10,13 +12,16 @@ import 'vm_service_wrapper.dart';
 
 /// A controller for global application preferences.
 class PreferencesController {
+  Map<String, List<double>> _initialFractions = {};
```
kenzieschmoll, line 15: nit: put type on RHS `final _initialFractions = <String, List<double>>{};`

packages/devtools_app/lib/src/preferences.dart
```diff
@@ -60,4 +68,34 @@ class PreferencesController {
   void toggleDenseMode(bool enableDenseMode) {
     _denseMode.value = enableDenseMode;
   }
+
+  /// Get the value for the split fractions.
+  ValueNotifier<List<double>> lookupSplitFractions(String key) {
+    return splitFractions.putIfAbsent(
+        key, () => ValueNotifier<List<double>>(_initialFractions[key] ?? []));
```
kenzieschmoll, line 75: nit: put a trailing comma after the second param to `putIfAbsent`

packages/devtools_app/lib/src/preferences.dart
```diff
@@ -60,4 +68,34 @@ class PreferencesController {
...
+  }
+
+  /// Update the value for the split fractions.
+  void updateSplitFractions(String key, List<double> value) {
+    if ((key?.isEmpty ?? true) ||
+        (value?.length ?? 0) < 2 ||
+        isListEqual(lookupSplitFractions(key).value, value)) return;
+    lookupSplitFractions(key).value = value;
+    storage?.setValue(
+        'ui.splitFractions', jsonEncode(fractionsToJson(splitFractions)));
```
kenzieschmoll, line 85: put 'ui.splitFractions' into a const and use here and above. Also, add a trailing comma in this method call after `jsonEncode(fractionsToJson(splitFractions))`

packages/devtools_app/lib/src/preferences.dart
```diff
@@ -60,4 +68,34 @@ class PreferencesController {
...
+  }
+
+  Map<String, List<double>> fractionsFromJson(Map<String, dynamic> json) {
+    return json
+        .map((key, fractions) => MapEntry(key, fractions?.cast<double>()));
+  }
+
+  Map<String, List<double>> fractionsToJson(
```
kenzieschmoll, line 93: this method as well as `fractionsFromJson` can be private

packages/devtools_app/lib/src/preferences.dart
```diff
@@ -60,4 +68,34 @@ class PreferencesController {
...
+      Map<String, ValueListenable<List<double>>> fractions) {
```
kenzieschmoll, line 94: nit: trailing comma after `fractions`

packages/devtools_app/lib/src/preferences.dart
```diff
@@ -60,4 +68,34 @@ class PreferencesController {
...
+    return fractions.map((key, fractions) => MapEntry(key, fractions.value));
+  }
+
+  bool isListEqual(List value, List other) {
```
kenzieschmoll, line 98: add the `<double>` type to these lists

packages/devtools_app/lib/src/preferences.dart
```diff
@@ -60,4 +68,34 @@ class PreferencesController {
...
+        (value?.length ?? 0) < 2 ||
+        isListEqual(lookupSplitFractions(key).value, value)) return;
+    lookupSplitFractions(key).value = value;
+    storage?.setValue(
+        'ui.splitFractions', jsonEncode(fractionsToJson(splitFractions)));
+  }
+
+  Map<String, List<double>> fractionsFromJson(Map<String, dynamic> json) {
+    return json
+        .map((key, fractions) => MapEntry(key, fractions?.cast<double>()));
```
kenzieschmoll, line 90: `fractions?.cast<double>()` will yield null when `fractions` is null. Instead of null, should we be using an empty list?

packages/devtools_app/lib/src/split.dart
```diff
@@ -88,15 +95,36 @@ class Split extends StatefulWidget {
...
 
   bool get isHorizontal => widget.axis == Axis.horizontal;
 
   @override
   void initState() {
     super.initState();
+    if (widget.fractionsKey != null) {
+      _fractionsStorage = preferences.lookupSplitFractions(widget.fractionsKey);
+      fractionsStorage = _fractionsStorage.value;
+      _fractionsStorage.addListener(_onChangeSplitFractions);
```
kenzieschmoll, line 109: instead of manually adding and removing the listener, use the `addAutoDisposeListener` helper that is used throughout devtools

packages/devtools_app/lib/src/split.dart
```diff
@@ -280,6 +316,11 @@ class _SplitState extends State<Split> {
     );
   }
 
+  // Save split fractions helpers
```
kenzieschmoll, line 319: nit: use three slashes for dart doc ///

packages/devtools_app/test/preferences_controller_test.dart
```diff
@@ -57,5 +58,33 @@ void main() {
       expect(valueChanged, isTrue);
       expect(controller.denseModeEnabled.value, isNot(originalValue));
     });
+
+    test('updateSplitFractions', () {
+      bool valueChanged = false;
+      const String fractionsKey = 'test';
+      const List<double> testValue = [0.33342342423, 0.98989899343];
+      const List<double> testValue2 = [0.33342342424, 0.98989899344];
```
kenzieschmoll, line 66: you can remove types on LHS

packages/devtools_app/lib/src/split.dart
```diff
@@ -109,6 +137,11 @@ class _SplitState extends State<Split> {
 
     final availableSize = axisSize - _totalSplitterSize();
 
+    if (fractions.length == fractionsStorage?.length) {
+      fractions = List.from(fractionsStorage);
```
kenzieschmoll, line 141: add type `List<double>.from(fractionsStorage)`

packages/devtools_app/lib/src/preferences.dart
kenzieschmoll: additionally, return early `if ((value == null) != (other == null))` to avoid unnecessary encoding. Since we expect these lists to be very small, it may be more performant to use [ListEquality](https://api.flutter.dev/flutter/package-collection_collection/ListEquality-class.html) to compare the elements directly instead of encoding the lists.

Hixie:
@quyenvsp Are you still interested in working on this PR?

Hixie:
@quyenvsp since we haven't heard from you I'm going to close this PR for now, but please don't hesitate to submit a new one if you are still working on this!

PR #3484, opened 2021-11-01
remove package meta dependency and imports

kenzieschmoll:
the imports in devtools_app were blocking a pub publish since we no longer depend on package:meta. Not sure how our bots weren't failing before this change.


jacob314 (approved):
lgtm

kenzieschmoll:
hitting https://github.com/flutter/devtools/issues/3434

PR #3487, opened 2021-11-02
Do not focus a line in the code view when no `outlineNode` is selected

elliette:
If we don't have a location and line on the `outlineNode`, use `-1` as the index.

This fixes an error when launching DevTools from g3, where the debugger page would try to load without a location.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/debugger/codeview.dart
@@ -708,7 +708,7 @@ class _LinesState extends State<Lines> with AutoDisposeMixin {
               widget.debugController.programExplorerController.outlineSelection,
           builder: (context, outlineNode, _) {
             final isFocusedLine =
-                outlineNode.location?.location?.line == lineNum;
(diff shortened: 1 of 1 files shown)


kenzieschmoll:
LGTM. Let's change the PR title for accuracy though for when we generate the changelog. Change to "Do not focus a line in the code view when no outlineNode is selected", as that is really what this is doing. thanks!

kenzieschmoll (approved):

bkonyi:
LGTM! I don't think we necessarily need to default to -1 if the result is null since the comparison will always return false in that case, but it's probably the right thing to do assuming we intend to migrate to null safety eventually... :-)

kenzieschmoll:
Agreed. This was my original suggestion to Elliott and it could probably be made more clear by doing something like this: 
```dart
final selectedOutlineNodeLine = outlineNode?.location?.location?.line;
final isFocusedLine = selectedOutlineNodeLine != null ? selectedOutlineNodeLine == lineNum : false;
```

Issue #3139, opened 2021-06-17
Support a rich tooltip for frames in the flutter frames chart


kenzieschmoll:
The tooltip should link to shader compilation jank docs.

This would also enable us the make it more clear that the shader compilation time shown is a subset of the raster time.

See https://github.com/flutter/devtools/pull/3125#issuecomment-862782005

PR #3493, opened 2021-11-10
Create rich tooltip for Flutter frames in the performance view

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/3139
Screen Shot 2021-11-10 at 10 58 22 AM
Screen Shot 2021-11-10 at 10 57 46 AM

This PR also has some cleanup like creating a `LinkTextSpan` helper widget and using where appropriate, moving the hover.dart file to a common location, etc.


packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
```diff
@@ -399,6 +384,107 @@ class FlutterFramesChartItem extends StatelessWidget {
...
+          ),
+        ),
+      ),
+    );
+  }
+
+  double _calculateTooltipWidth(List<String> lines) {
+    var maxWidth = 0.0;
+    for (final line in lines) {
+      final textPainter = TextPainter(
```
elliette, line 470: Could use / modify `calcluateTextSpanWidth` helper function instead: https://github.com/flutter/devtools/blob/fa9124840df46b49a36635a3d86e0f4f1d2b5ac1/packages/devtools_app/lib/src/ui/utils.dart#L47-L56

elliette (approved):
One suggestion but LGTM!

packages/devtools_app/lib/src/performance/flutter_frames_chart.dart
kenzieschmoll: good suggestion - done

PR #3494, opened 2021-11-10
Track whether an app is a Flutter Web app in analytics

elliette:
Adds `flutter_web` type along with `web` and `flutter`.

### DIFF
--- a/packages/devtools_app/lib/src/analytics/_analytics_web.dart
+++ b/packages/devtools_app/lib/src/analytics/_analytics_web.dart
@@ -29,6 +29,7 @@ import 'constants.dart' as analytics_constants;
 // Dimensions1 AppType values:
 const String appTypeFlutter = 'flutter';
 const String appTypeWeb = 'web';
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/analytics/_analytics_web.dart
```diff
@@ -557,6 +558,10 @@ Future<void> computeUserApplicationCustomGTagData() async {
   if (serviceManager.connectedApp.isDartWebAppNow) {
     userAppType = appTypeWeb;
   }
+  if (serviceManager.connectedApp.isFlutterWebAppNow) {
+    userAppType = appTypeFlutterWeb;
+  }
```
kenzieschmoll, line 563: can we put all of these checks into an if else block to make sure nothing is getting mistakenly overwritten in the future? also, in the default case, we should return 'dart_cli_app'.

kenzieschmoll (approved):
LGTM after the comment is addressed

PR #3506, opened 2021-11-15
Implement `noSuchMethod` in VmServiceWrapper

kenzieschmoll:
Implementing `noSuchMethod` will ensure that DevTools is not broken by updates to the VmService class that we implement. When new methods are added to VmService, we can implement them as we need.


kenzieschmoll:
closing in favor of https://github.com/flutter/devtools/pull/3505

PR #3513, opened 2021-11-19
Refactor `Variable` class and rename it to `DartObjectNode`

elliette:
Addresses two TODOs on the `Variable` class:

1. Consider a new class name. This class is more just the data model for a tree of Dart objects with properties rather than a "Variable".
2. Refactor to make `name`, `ref,` `text` named (and potentially required?) parameters. Give `ref` a type.

### DIFF
--- a/packages/devtools_app/lib/src/console_service.dart
+++ b/packages/devtools_app/lib/src/console_service.dart
@@ -74,7 +74,7 @@ class ConsoleService extends Disposer {
     bool expandAll = false,
   }) async {
     _stdioTrailingNewline = false;
-    final variable = Variable.fromValue(
+    final variable = DartObjectNode.fromValue(
       name: name,
       value: value,
       diagnostic: diagnostic,
--- a/packages/devtools_app/lib/src/console_service.dart
+++ b/packages/devtools_app/lib/src/console_service.dart
@@ -23,7 +23,7 @@ class ConsoleLine {
         forceScrollIntoView: forceScrollIntoView,
       );
 
-  factory ConsoleLine.variable(
+  factory ConsoleLine.dartObjectNode(
     DartObjectNode variable, {
     bool forceScrollIntoView = false,
   }) =>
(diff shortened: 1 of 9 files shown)


packages/devtools_app/lib/src/console_service.dart
```diff
@@ -24,7 +24,7 @@ class ConsoleLine {
       );
 
   factory ConsoleLine.variable(
```
kenzieschmoll, line 26: rename this factory constructor too

packages/devtools_app/lib/src/debugger/debugger_model.dart
```diff
@@ -328,12 +328,13 @@ Future<void> addExpandableChildren(
 }
 
 /// Builds the tree representation for a [Variable] object by querying data,
-/// creating child Variable objects, and assigning parent-child relationships.
+/// creating child DartObjectNode objects, and assigning parent-child
```
kenzieschmoll, line 331: wrap DartObjectNode in [ ]

packages/devtools_app/lib/src/debugger/debugger_model.dart
```diff
@@ -695,13 +696,14 @@ List<Variable> _createVariablesForFields(
...
-// TODO(elliette): Refactor to make name, ref, text named (and potentially
-// required?) parameters. Give ref a type.
-class Variable extends TreeNode<Variable> {
-  Variable._(this.name, ref, this.text, {int offset, int childCount})
-      : _ref = ref,
+class DartObjectNode extends TreeNode<DartObjectNode> {
+  DartObjectNode._({
+    this.name,
+    this.text,
+    GenericInstanceRef ref,
```
kenzieschmoll, line 703: make these required

kenzieschmoll (approved):
LGTM with a couple comments

kenzieschmoll (approved):
LGTM with a couple comments

packages/devtools_app/lib/src/console_service.dart
elliette: Done

packages/devtools_app/lib/src/debugger/debugger_model.dart
elliette: Done

packages/devtools_app/lib/src/debugger/debugger_model.dart
elliette: Looks like none of them can be required (for example, `DartObjectNode.text` only provides the `text` parameter).

PR #3522, opened 2021-12-01
Add is_embedded dimension to DevTools analytics

kenzieschmoll:
This will allow us to see whether a user is using the embedded version of DevTools in an IDE. We already have information as to which IDE a user is using from the `?ide=` query parameter


DanTup (approved):
lgtm!

(I don't know if `dimension10` needs adding to some analytics account somewhere, but I presume if it does that will be done too)

kenzieschmoll:
> (I don't know if dimension10 needs adding to some analytics account somewhere, but I presume if it does that will be done too)

Yep - it needed to be added to our analytics 360 account, and that has been completed too.

Issue #3489, opened 2021-11-08
Proposal: Improve widget selection in the inspector


CodemateTechAgency:
# KR1

## Background

> “The inspector allows the user to search the summary tree and the details tree by keyword. The inspector searches the part of the details tree that has been loaded by default with an option to search the full details tree enabled by KR 2.”

## Proposal

- Add bordered headers for both widget tree and explorer/details areas to isolate browsing/filtering related actions and separate them from the action toolbar above. Tab bar look should reflect the rest of the DevTools (without the grey background).
- Add search functionality to the Widget Tree header. Move the refresh widget tree to the header area.
  - Search can target different contexts based on the user input. (Widget/Details/All etc.)
- Rename “Details tree” to “Widget Details Tree” to pair it better with the Widget Tree.

**Question**: Should we now move the select widget mode to the button group on the right?

### Before

![Screenshot 2021-11-05 at 10 11 30](https://user-images.githubusercontent.com/82218649/140737406-09100e5e-eefb-47f1-a10f-2349af0be678.png)

### After

![Screenshot 2021-11-05 at 10 12 17](https://user-images.githubusercontent.com/82218649/140737603-b9b9eca9-c154-44f7-b117-00e4c00b1338.png)

![Screenshot 2021-11-05 at 10 12 09](https://user-images.githubusercontent.com/82218649/140737648-d8056a42-d59b-4891-aef7-e2df6540182f.png)

### Search field variations

![Screenshot 2021-11-08 at 10 50 32](https://user-images.githubusercontent.com/82218649/140737678-c2fb72da-f6a5-4156-8985-4363e55ff5e1.png)

Normal state **a)**, active search fields with **b)** and without **c)** search options (search from widget names, details, all, etc). Floating version **d)** for narrow spaces 

---

# KR2

## Background

> “The inspector provides a UI affordance to show the complete detail tree. This is essentially a shortcut for performing the following steps: 1) select the root widget in the summary tree and 2) click the details tree tab. Progressive loading might be needed for very large trees.“

Additional input and background information is welcome. Assumption here is that KR2 will support KR1 with more detailed search and navigation. As KR1 can also include “Widget Details Tree” as a search target, this proposal is to make the navigation tree more discoverable.

## Proposal

- Add breadcrumb navigation atop of the details tree for easier navigation and to make root level always accessible and easily discoverable.

### Before

[...]

PR #3525, opened 2021-12-07
Inspector widget selection improvements (#3489)

aednlaxer:
Original ticket: https://github.com/flutter/devtools/issues/3489

Changes in this pull request:

* add widget search with text selection
* rename "Details Tree" to "Widget Details Tree"
* add breadcrumbs navigator to Widget Details Tree

demo


packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -542,6 +585,151 @@ class InspectorTreeController extends Object {
       }
     }
   }
+
+  /* Search support */
+  @override
+  void previousMatch() {
+    super.previousMatch();
+    final currentSearchMatches = searchMatches.value;
```
kenzieschmoll, line 593: why can't we use `matchIndex` from `SearchControllerMixin` for this to grab the current match? 

More generally, perhaps instead of overriding `previousMatch` and `nextMatch`, we should have another method, `SearchControllerMixin.onMatchChanged` that `SearchControllerMixin.previousMatch` and `SearchControllerMixin.nextMatch` can call, and subclasses can override and implement `onMatchChanged` instead.

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -542,6 +585,151 @@ class InspectorTreeController extends Object {
...
+    if (_searchDebounce?.isActive ?? false) {
+      _searchDebounce.cancel();
+    }
+  }
+
+  Future _searchOperation;
+  Timer _searchDebounce;
+
+  @override
+  List<InspectorTreeRow> matchesForSearch(String search) {
```
kenzieschmoll, line 636: Can you add support for searching through previous matches? See the super class for method signature (you may need to sync to head - this framework was added a couple of weeks ago)

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -542,6 +585,151 @@ class InspectorTreeController extends Object {
...
+      _searchDebounce.cancel();
+    }
+    _searchDebounce =
+        Timer(Duration(milliseconds: search.isEmpty ? 0 : 300), () async {
+      // do something with query
+
+      // Abort any ongoing search operations and start a new one
+      if (_searchOperation != null) {
+        try {
+          await _searchOperation;
```
kenzieschmoll, line 650: I'm not seeing how this aborts the ongoing search operation? this looks like it actually waits for the ongoing operation to complete

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -542,6 +585,151 @@ class InspectorTreeController extends Object {
...
+      _searchOperation = matchesForSearchAsyncOld(search, _searchTarget);
+
+      await _searchOperation;
+      searchInProgress = false;
+    });
+    return [];
+  }
+
+  Future<void> matchesForSearchAsyncOld(
+      String search, SearchTargetType searchTarget) async {
```
kenzieschmoll, line 667: nit: trailing comma

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -542,6 +585,151 @@ class InspectorTreeController extends Object {
...
+  List<InspectorTreeRow> matchesForSearch(String search) {
+    updateMatches([]);
+    searchInProgress = true;
+
+    if (_searchDebounce?.isActive ?? false) {
+      _searchDebounce.cancel();
+    }
+    _searchDebounce =
+        Timer(Duration(milliseconds: search.isEmpty ? 0 : 300), () async {
+      // do something with query
```
kenzieschmoll, line 645: remove comment?

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -1007,48 +1217,66 @@ class InspectorRowContent extends StatelessWidget {
...
+                        child: RotationTransition(
+                          turns: expandArrowAnimation,
+                          child: Icon(
+                            Icons.expand_more,
+                            size: defaultIconSize,
+                          ),
+                        ),
+                      )
+                    : const SizedBox(
+                        width: defaultSpacing, height: defaultSpacing),
```
kenzieschmoll, line 1240: nit: trailing comma

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -557,11 +745,15 @@ class InspectorTree extends StatefulWidget {
     Key key,
     @required this.controller,
     @required this.debuggerController,
+    this.inspectorTreeController,
     this.isSummaryTree = false,
     this.widgetErrors,
-  }) : super(key: key);
+  })  : assert(isSummaryTree && inspectorTreeController == null ||
+            !isSummaryTree && inspectorTreeController != null),
```
kenzieschmoll, line 752: nit this can be `assert(isSummaryTree == (inspectorTreeController == null))`

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -542,6 +585,151 @@ class InspectorTreeController extends Object {
...
+
+      if (matches.isNotEmpty && _firstSearchMatch) {
+        _firstSearchMatch = false;
+        // Always select the first search match
+        onSelectRow(matches[0]);
+      }
+    }
+
+    int _statsSearchOps = 0;
+    int _statsWidgets = 0;
```
kenzieschmoll, line 682: if these are just for debugging, let's prefix the names with `debug`

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -542,6 +585,151 @@ class InspectorTreeController extends Object {
...
+    return [];
+  }
+
+  Future<void> matchesForSearchAsyncOld(
+      String search, SearchTargetType searchTarget) async {
+    final matches = <InspectorTreeRow>[];
+
+    bool _firstSearchMatch = true;
+    void _updateSearchMatches() {
+      updateMatches(matches);
```
kenzieschmoll, line 672: we should be using `refreshSearchMatches` instead of `updateMatches`. In fact, can you rename `updateMatches` to `_updateMatches` to enforce this?

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -542,6 +585,151 @@ class InspectorTreeController extends Object {
...
+
+      if (matches.isNotEmpty && _firstSearchMatch) {
+        _firstSearchMatch = false;
+        // Always select the first search match
+        onSelectRow(matches[0]);
```
kenzieschmoll, line 677: can we use `matchIndex.value` instead of trying to manually manage the index? Generally we should be trying to use the logic from `SearchControllerMixin` instead of re-inventing the wheel

packages/devtools_app/lib/src/inspector/inspector_tree.dart
```diff
@@ -284,6 +286,7 @@ class InspectorTreeRow {
   final bool lineToParent;
 
   bool get isSelected => node.selected;
+  bool get hasSearchMatch => node.hasSearchMatch;
```
kenzieschmoll, line 289: why do we need `hasSearchMatch` when DataSearchStateMixin has `isSearchMatch`?

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -542,6 +585,151 @@ class InspectorTreeController extends Object {
...
+      return;
+    }
+
+    debugPrint('Search started: ' + _searchTarget.toString());
+    final caseInsensitiveSearch = search.toLowerCase();
+
+    // Reset search matches
+    for (final row in cachedRows) {
+      _statsWidgets++;
+      setSearchMatch(row.node, false);
```
kenzieschmoll, line 699: we shouldn't need to do this. SearchControllerMixin already resets the isSearchMatch field in _updateMatches

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -542,6 +585,151 @@ class InspectorTreeController extends Object {
...
+      if (searchTarget == SearchTargetType.widget) {
+        final description = diagnostic.toStringShort();
+        final textPreview = diagnostic.json['textPreview'];
+
+        final searchValue = textPreview is String
+            ? description + ' ' + textPreview.replaceAll('\n', ' ')
+            : description;
+
+        _statsSearchOps++;
+        if (searchValue.toLowerCase().contains(caseInsensitiveSearch)) {
```
kenzieschmoll, line 717: use caseInsensitiveContains helper

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -542,6 +585,151 @@ class InspectorTreeController extends Object {
...
+      // Start new search operation
+      _searchOperation = matchesForSearchAsyncOld(search, _searchTarget);
+
+      await _searchOperation;
+      searchInProgress = false;
+    });
+    return [];
+  }
+
+  Future<void> matchesForSearchAsyncOld(
```
kenzieschmoll, line 666: I'm not sure I understand the naming here. What does async old mean?

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -542,6 +585,151 @@ class InspectorTreeController extends Object {
...
+        final textPreview = diagnostic.json['textPreview'];
+
+        final searchValue = textPreview is String
+            ? description + ' ' + textPreview.replaceAll('\n', ' ')
+            : description;
+
+        _statsSearchOps++;
+        if (searchValue.toLowerCase().contains(caseInsensitiveSearch)) {
+          matches.add(row);
+          setSearchMatch(row.node, true);
```
kenzieschmoll, line 719: same comment as above. SearchControllerMixin._updateMatches should already handle this

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -542,6 +585,151 @@ class InspectorTreeController extends Object {
...
+          _updateSearchMatches();
```
kenzieschmoll, line 720: matchesForSearch also shouldn't need to worry about updating the search matches. A class that mixes in SearchControllerMixin is responsible for implementing matchesForSearch. matchesForSearch is called automatically in SearchControllerMixin when the `search` setter is called.

kenzieschmoll (commented):
Done with first round of comments. Generally, we should be trying to use the search functionality in SearchControllerMixin where possible instead of trying to manage matches and match indexes manually. Happy to discuss further or chat offline for help with hooking the inspector up to the existing search functionality in DevTools.

packages/devtools_app/lib/src/inspector/inspector_breadcrumbs.dart
```diff
@@ -0,0 +1,169 @@
...
+        ),
+      ),
+    );
+  }
+
+  List<_InspectorBreadcrumbData> _getBreadcrumbs(List<InspectorTreeRow> rows) {
+    final List<_InspectorBreadcrumbData> items = rows
+        .map((e) => _InspectorBreadcrumbData.wrap(e, e == rows.safeLast))
+        .toList();
+    if (items.length > 5) {
```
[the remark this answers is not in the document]
aednlaxer: Max number of breadcrumbs is 4 (excluding "[root]" and "…"), this limit is enforced on that line. If there are more breadcrumbs between root and the first visible breadcrumb, ellipsis item is displayed. It should actually be 4 instead of 5 in code. I've changed it to 4 and pulled this value out into a const.

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -198,6 +212,27 @@ class InspectorTreeController extends Object {
...
+    final selectedItem = cachedRows.firstWhere(
+      (element) => element.isSelected,
+      orElse: () => null,
+    );
+    if (selectedItem == null) return [];
+
+    final pathToRoot = <InspectorTreeRow>[selectedItem];
+    InspectorTreeNode nextParentNode = selectedItem.node.parent;
+    while (nextParentNode != null) {
+      final rowItem = cachedRows.firstWhere(
```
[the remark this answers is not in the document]
aednlaxer: In that case `rowItem` would be null (`orElse: null,`) which should break out the loop (`if (rowItem == null) break`)

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
aednlaxer: `onMatchChanged()` has been added

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
aednlaxer: Sure, this is something I'll look into early next week

aednlaxer:
I've made some changes according to the review comments above. There are still couple more things to be checked\fixed.

The reason behind using 'SearchControllerMixin.updateMatches()' is that there is a timer to debounce user's input. I removed the `updateMatches()` calls but had to use `set search(String search)` and `resetSearch()` to manage temporary list of results. I'd like to know your opinion.

packages/devtools_app/lib/src/inspector/inspector_breadcrumbs.dart
```diff
@@ -46,29 +49,27 @@ class InspectorBreadcrumbNavigator extends StatelessWidget {
...
+        row: row,
+        isSelected: row == rows.safeLast,
+      );
+    }).toList();
+    List<_InspectorBreadcrumbData> breadcrumbs;
+    if (items.length > _maxNumberOfBreadcrumbs) {
+      breadcrumbs = [
+        items[0],
+        _InspectorBreadcrumbData.more(),
+        ...items.sublist(items.length - _maxNumberOfBreadcrumbs, items.length),
```
kenzieschmoll, line 66: shouldn't this be `items.length - _maxNumberOfBreadcrumbs - 1`? Otherwise this would happen
`[b1, b2, b3, b4, b5] --> b1 > more > b2, b3, b4, b5`

packages/devtools_app/lib/src/inspector/inspector_breadcrumbs.dart
```diff
@@ -129,41 +135,62 @@ class _InspectorBreadcrumb extends StatelessWidget {
...
   final bool isSelected;
 
-  String get text => _alternativeText ?? row?.node?.diagnostic?.description;
+  String get text => alternativeText ?? row?.node?.diagnostic?.description;
 
-  Widget get icon => _alternativeIcon ?? row?.node?.diagnostic?.icon;
+  Widget get icon => alternativeIcon ?? row?.node?.diagnostic?.icon;
 
-  bool get isChevron => row == null && _alternativeText == null;
+  bool get isChevron =>
```
kenzieschmoll, line 189: why not just check if `row == null && alternativeIcon == Icons.chevron_right`. Same for isEllipsis: `row == null && alternativeText == _ellipsisText` where _ellipsisText is a const value pulled out and used both here and in the .more constructor

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -125,19 +130,19 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
           summaryTreeController.search.isEmpty &&
-          !searchPreventCloseOnBlur) {
+          !searchPreventClose) {
         setState(() {
           searchVisible = false;
         });
       }
 
-      // Reset searchPreventCloseOnBlur state to false after the search field gains focus.
+      // Reset searchPreventClose state to false after the search field gains focus.
```
kenzieschmoll, line 142: dart-doc nit: wrap var names that are in comments in brackets: [searchPreventClose]

packages/devtools_app/lib/src/inspector/inspector_screen.dart
```diff
@@ -414,6 +373,89 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+  Widget build(BuildContext context) {
+    return Column(
+      children: [
+        _controlsContainer(
+          context,
+          Row(
+            children: <Widget>[
+              const SizedBox(width: denseSpacing),
+              const Text('Widget Tree'),
+              const SizedBox(width: denseSpacing),
```
kenzieschmoll, line 404: nit: just wrap `Text('WidgetTree')` in a padding widget with horizontal padding `denseSpacing`

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -588,98 +596,82 @@ class InspectorTreeController extends Object
...
+    _searchResults = [];
+    super.resetSearch();
+  }
+
+  @override
+  set search(String search) {
+    _searchResults = [];
+    super.search = search;
+
+    _startSearchDebounceTimer();
```
kenzieschmoll, line 626: This debounceTimer seems like it would be useful for all searching. Instead of overriding the search mixin methods to use it, feel free to add it to the search code directly. to be safe, you could add a parameter that allows each consumer of the search mixin to opt in or out.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1066,6 +1066,17 @@ extension StringExtension on String {
     }
     return true;
   }
+
```
kenzieschmoll, line 1069: add tests for these to utils.dart

packages/devtools_app/lib/src/inspector/inspector_breadcrumbs.dart
aednlaxer: I didn't like that `_maxNumberOfBreadcrumbs` meant max number of breadcrumbs excluding root so I've changed it to 5 (root + 4 widget breadcrumbs) with a comment. Now the logic looks like this:
```dart
[
    items[0],
    _InspectorBreadcrumbData.more(),
    ...items.sublist(items.length - _maxNumberOfBreadcrumbs + 1, items.length),
]
```

For 5 items it should take all items (root + allowed 4 at the end) without adding 'more' item. For 6+ items it should take first one and last 4 items, for example: `[b1, b2, b3, b4, b5, b6] -> b1 > more > b3, b4, b5, b6`

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
aednlaxer: `setSearchMatch` does two things: it expands current node and it marks a `node` (not row) as search match. 
`InspectorTree` uses `getRow()` to get row data which is different from what's returned from `getCachedRow()`. `_updateMatches()` marks a row as search match but this row is a cached row so `isSearchMatch` is not reflecting the actual status for rows returned by `getRow()`.

I have the following questions regarding this:
* can cached rows be used in `InspectorTree`?
* what would be a good way to expand a node if `setSearchMatch()` is removed?

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -109,6 +170,10 @@ mixin SearchControllerMixin<T extends DataSearchStateMixin> {
     onMatchChanged(activeMatchIndex);
   }
 
+  /// Delay to reduce the amount of search queries
+  /// Duration.zero (default) disables debounce
+  Duration get debounceDelay => Duration.zero;
```
kenzieschmoll, line 175: Instead of returning Duration.zero, why not just return null? Since we are never actually waiting for a Duration of zero, null seems like it would better indicate the state of "no value" or disabled

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -721,6 +665,18 @@ class InspectorTreeController extends Object
...
+    return matches;
+  }
+}
+
+extension RemoteDiagnosticsNodeExtension on RemoteDiagnosticsNode {
+  String get searchValue {
+    final description = toStringShort();
+    final textPreview = json['textPreview'];
+    return textPreview is String
+        ? description + ' ' + textPreview.replaceAll('\n', ' ')
```
kenzieschmoll, line 678: use string interpolation here instead
`'$description ${textPreview.replaceAll('\n', ' ')}'`

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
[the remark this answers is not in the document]
kenzieschmoll: right, but then wouldn't the path to root be incomplete, meaning the breadcrumb navigator would be incomplete and not show the root node?

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
kenzieschmoll: A couple questsions:
1. Why do we need to mark both the row and the node as having a search match? as far as I can tell, everywhere `hasSearchMatch` is used, we have access to the row, so we could just write `row.isSearchMatch` instead.
2. Why are we expanding nodes at all for search?

If I understand correctly, the issue with cached rows vs non-cached rows is that we are setting `isSearchMatch` on cached rows and not on rows returned by `getRow()` - is that correct? If so, can we do a comparison where ever we need to call `getRow()` and need to know the value of `isSearchMatch` to check if there is a matching cached row with `isSearchMatch` set to true?

aednlaxer:
>Why do we need to mark both the row and the node as having a search match?

It was made as a work around because `isSearchMatch` is set on cached rows but different instances of rows are used when displaying the tree.

>Why are we expanding nodes at all for search?

Search is looking for matches in `cachedRows`. In order to have all rows in `cachedRows` list all nodes need to be expanded.

>If I understand correctly, the issue with cached rows vs non-cached rows is that we are setting isSearchMatch on cached rows and not on rows returned by getRow() - is that correct?

Yes, that's correct.

>can we do a comparison where ever we need to call getRow() and need to know the value of isSearchMatch

Yes. We need to know the value of `isSearchMatch` in `_InspectorTreeState` but `getRows()` is used there. 

I've made some changes to the code:
* `setSearchMatch()` and `hasSearchMatch` have been removed
* `InspectorTree` now uses `getCachedRow()`. It should be okay to use it since it may still call `root?.getRow()` under the hood.

There's still need to expand the tree to get all rows cached. I did this by expanding all nodes of the tree in the overridden `set search` setter (so nodes are expanded before submitting search query). I don't think it's ideal, maybe it's better to find a solution that keeps current state of collapsed and expanded nodes.

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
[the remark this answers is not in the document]
aednlaxer: I think you're right and I might have found a way to see this issue in action. I wonder if there's a better way to get a path to the root

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
aednlaxer: >Why do we need to mark both the row and the node as having a search match?

It was made as a work around because `isSearchMatch` is set on cached rows but different instances of rows are used when displaying the tree.

>Why are we expanding nodes at all for search?

Search is looking for matches in `cachedRows`. In order to have all rows in `cachedRows` list all nodes need to be expanded.

>If I understand correctly, the issue with cached rows vs non-cached rows is that we are setting isSearchMatch on cached rows and not on rows returned by getRow() - is that correct?

Yes, that's correct.

>can we do a comparison where ever we need to call getRow() and need to know the value of isSearchMatch

Yes. We need to know the value of `isSearchMatch` in `_InspectorTreeState` but `getRows()` is used there. 

I've made some changes to the code:
* `setSearchMatch()` and `hasSearchMatch` have been removed
* `InspectorTree` now uses `getCachedRow()`. It should be okay to use it since it may still call `root?.getRow()` under the hood.

There's still need to expand the tree to get all rows cached. I did this by expanding all nodes of the tree in the overridden `set search` setter (so nodes are expanded before submitting search query). I don't think it's ideal, maybe it's better to find a solution that keeps current state of collapsed and expanded nodes.

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -594,6 +593,13 @@ class InspectorTreeController extends Object
   }
 
   /* Search support */
+  @override
+  set search(String value) {
+    // Expand tree so all rows are available for search
+    _expandAll(root);
```
kenzieschmoll, line 599: instead of overriding search to expand the tree. Why don't we create a copy of the tree at the time of tree creation. This copy can be fully expanded from the start, which has the advantages that A) we won't have to modify the expand/collapse state of the original tree, and B) we won't have to pay the cost of expanding the tree each time we perform a search.

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -616,8 +622,9 @@ class InspectorTreeController extends Object
     if (searchPreviousMatches) {
       final previousMatches = searchMatches.value;
       for (final previousMatch in previousMatches) {
-        if (previousMatch.node.diagnostic.searchValue
-            .caseInsensitiveContains(search)) {
+        if (cachedRows.contains(previousMatch) &&
```
kenzieschmoll, line 625: will this ever be false? If a row was a previous match, wouldn't it always be in cachedRows?

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
aednlaxer: Having a copy leads to a previous issue when actual\cached rows and rows from the copy are not the same instances so `isSearchMatch` doesn't reflect the actual value. Anyway I tried using a copy of the tree for search and finding a path to the root for breadcrumbs. `isSearchMatch` was made to show the actual value because `getCachedRow()` sets the value to the cached row before returning it.

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -211,27 +223,26 @@ class InspectorTreeController extends Object
...
+    final selectedItem = _cachedSelectedRow;
     if (selectedItem == null) return [];
 
     final pathToRoot = <InspectorTreeRow>[selectedItem];
     InspectorTreeNode nextParentNode = selectedItem.node.parent;
     while (nextParentNode != null) {
-      final rowItem = cachedRows.firstWhere(
-        (element) => element.node == nextParentNode,
-        orElse: null,
+      final rowItem = _searchableCachedRows.firstWhere(
```
kenzieschmoll, line 243: have you tried this with a very large app? The worst case scenario of this is that you are trying to get the path from the deepest leaf node to the root. Where N is the depth of the leaf and M is the total number of nodes in the tree, this could be NxM, which could be very slow for a large app.

seems like it would be more performant to either A) make rows aware of their parent rows so that `row.parent` would yield the next parent, or B) to give Node's access to their associated row, so that `row.node.parent.row` would yield the next parent.

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -184,6 +185,10 @@ class InspectorTreeController extends Object
   final List<InspectorTreeRow> cachedRows = [];
   InspectorTreeRow _cachedSelectedRow;
 
+  /// Cached rows of the tree. Similar to [cachedRows] but items are populated
```
kenzieschmoll, line 188: dart-doc nit: there should be a blank new line between the first sentence and the rest. 

Also just to clarify, `_searchableCachedRows` contains every row in the tree, is that correct? not just those that are visible (nodes with expanded parents)?

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -184,6 +185,10 @@ class InspectorTreeController extends Object
...
+  /// when root is changed and don't change when nodes are expanded or collapsed
+  final List<InspectorTreeRow> _searchableCachedRows = [];
```
kenzieschmoll, line 190: nit: put type on RHS

`final _searchableCachedRows = <InspectorTreeRow>[];`

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -211,27 +223,26 @@ class InspectorTreeController extends Object
     if (cachedRows[index]?.isSelected == true) {
       _cachedSelectedRow = cachedRows[index];
     }
-    return cachedRows[index];
+    final cachedRow = cachedRows[index];
```
kenzieschmoll, line 226: nit: move this up before the previous if block so the local var can be used in that if statement as well.

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
aednlaxer: Yes, it contains every row in the tree (including collapsed nodes, children of collapsed nodes), it's now reflected in the doc

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
aednlaxer: Turns out it's enough to have an instance of `InspectorTreeNode` for the selected row to find a way to the root. Such instance is available from an instance of `InspectorTreeRow` and `_cachedSelectedRow` in particular. This method now uses parent of the selected `row.node` to find a path to the root. Should be more performant now as it doesn't iterate through the whole tree to find the right item on every iteration of the `while` loop.

packages/devtools_app/lib/src/inspector/inspector_breadcrumbs.dart
```diff
@@ -53,12 +53,12 @@ class InspectorBreadcrumbNavigator extends StatelessWidget {
...
-    List<InspectorTreeRow> rows,
+    List<InspectorTreeNode> nodes,
   ) {
-    final List<_InspectorBreadcrumbData> items = rows.map((row) {
+    final List<_InspectorBreadcrumbData> items = nodes.map((node) {
       return _InspectorBreadcrumbData.wrap(
-        row: row,
-        isSelected: row == rows.safeLast,
+        node: node,
+        isSelected: node  == nodes.safeLast,
```
kenzieschmoll, line 61: pull `nodes.safeLast` out into a var before calling List.map so that it doesn't need to be called for every item in the list.

packages/devtools_app/lib/src/inspector/inspector_tree_controller.dart
```diff
@@ -220,33 +224,30 @@ class InspectorTreeController extends Object
...
     cachedRows[index] ??= root?.getRow(index);
-    if (cachedRows[index]?.isSelected == true) {
-      _cachedSelectedRow = cachedRows[index];
-    }
+
     final cachedRow = cachedRows[index];
     cachedRow?.isSearchMatch =
         _searchableCachedRows.safeGet(index)?.isSearchMatch ?? false;
+
+    if (cachedRows[index]?.isSelected == true) {
```
kenzieschmoll, line 232: use `cachedRow` here instead of `cachedRows[index]`

aednlaxer:
I'm having difficulties figuring out why goldens tests fail. I'm unable to reproduce it or update golden files locally when using Flutter master or stable. As far as I understand, devtools uses specific version of SDK in CI/CD (`2.8.0-3.1.pre` currently) but using same version locally didn't help. @kenzieschmoll do you have any idea what could be the reason here?

aednlaxer:
The goldens and other tests seem to be okay now

kenzieschmoll (approved):

kenzieschmoll:
I just checked this out locally and played with it against a large application to verify the performance is acceptable. It looks fantastic - great work!

Here is one polish item that should be done as a follow up: https://github.com/flutter/devtools/issues/3561

aednlaxer:
Thanks for thoroughly reviewing this PR! I'll have a look at the issue

Issue #3485, opened 2021-11-02
Support per-frame analysis for janky frames


kenzieschmoll:
When a janky frame occurs, the user wants to know what took a long time to build/layout/paint. With any of the “Enhance Tracing” toggles enabled, DevTools should be able to provide per-frame analysis on the widgets built and render objects laid out and painted. 

One idea is to show a bottom up view for each of the build, layout, paint events to identify widgets with expensive self time. Self time in this case would be `event.time - sum(children.time)`.

kenzieschmoll:
Event when "Enhance Tracing" is not enabled, we should nudge the users to use the feature when we encounter a long Build, Layout, Paint phase for a janky frame - see https://github.com/flutter/devtools/issues/3483.

kenzieschmoll:
![frame-analysis-gif](https://user-images.githubusercontent.com/43759233/142070654-e1738ec8-9629-41c7-b204-e9c47e6f98b6.gif)

@InMatrix @cobblest  @ayanogawa here is an implementation of the "tab-controller-like" view for navigating between the timeline events chart and frame analysis views (as proposed in the design doc). The general workflow is as such:
- in the flutter frames chart, frames with jank are marked with an icon suggesting more analysis is available (right now, marked only for UI jank until we have more actionable data from the raster thread)
- When this frame is selected, a button with the same icon and tooltip "Analyze selected frame" will appear in the Timeline Events section header.
- Clicking this button will open a tab for this frame's analysis

Some other items to note:
- Tabs will remain open until manually closed.
- Selecting a new frame will open the timeline frame chart to that frame's time range (current behavior)

Please let me know what you think, thanks!

InMatrix:
Thanks for the update, @kenzieschmoll.

I like the general direction this is taking. There are a few questions I'd like to discuss:

1. Having both the "shader jank" icon and the "analyze" icon for some frames look busy. I wonder if we could combine the two icons. Related, how useful is it to identify frames with shader compilation jank, since the workaround doesn't seem to be very useful? 
2. How long does it take to analyze an individual frame? 
3. If individual frame can be analyzed and visualized quickly, I wonder if it would make sense to show the individual frame view instead of the full Timeline Events chart by default. The full timeline events chart is difficult to navigate and sometimes slow. Is there a use case where navigating in that view is beneficial?

kenzieschmoll:
> Having both the "shader jank" icon and the "analyze" icon for some frames look busy. I wonder if we could combine the two icons. 

[...]

PR #3526, opened 2021-12-09
Add frame numbers to the flutter frames chart in the performance page

kenzieschmoll:
Work towards https://github.com/flutter/devtools/issues/3485, specifically addressing the feedback [here](https://github.com/flutter/devtools/issues/3485#issuecomment-982022144) - @ayanogawa

Screen Shot 2021-12-08 at 4 43 47 PM

When the frame analysis feature is turned on, these numbers will match the frame tab titles:
Screen Shot 2021-12-08 at 4 39 49 PM


elliette (approved):

ayanogawa:
Awesome work!
For the magnifying glasses, can we style it like the attachment?
Circle - .8 opacity of #303030, 20 X 20 px
Magnifying glass - #ffffff, 12 X 12 px

Thank you in advance.

timeline

Issue #3521, opened 2021-11-30
File explorer does not show file contents after a hot restart


annagrin:
**Repro**

- run a flutter example using `flutter run -d chrome`
- run devtools and connect it to the running app above. 
- open file explorer, find a main.dart file and observe the code opening.
- put a breakpoint anywhere in that file
- click hot restart
- try to open main.dart or changing to another file. 
- observe that the same contents (in my case. web_entrypoint.dart) is shown instead, and the devtools reports "Invalid object id" exceptions.

Note that after hot restart, refresh, or reload file explorer shows two or more sets of the same files, where only the last set is actually working. Looks like the fix needs to make sure that file explorer clears and re-initializes its state on hot restart and refresh.

![image](https://user-images.githubusercontent.com/12737689/144146520-1a800114-e700-43c0-8323-dc7bb33998cf.png)

elliette:
This issue is not restricted to web, it is reproducible on all platforms.

PR #3480, opened 2021-11-01
Add support for selecting objects in the program explorer outline view

bkonyi:
Selecting an object from the outline view currently only highlights the corresponding definition in the codeview and tries to scroll to the definition. Will eventually be used to display additional information about the selected program object.

Fixes https://github.com/flutter/devtools/issues/3389

**Demo:**
ezgif-3-3f50744802db

PR #3527, opened 2021-12-09
Fixes broken program explorer on hot restart

elliette:
Fixes https://github.com/flutter/devtools/issues/3521

### DIFF
--- a/packages/devtools_app/lib/src/debugger/program_explorer.dart
+++ b/packages/devtools_app/lib/src/debugger/program_explorer.dart
@@ -327,25 +327,29 @@ class _ProgramOutlineView extends StatelessWidget {
         if (isLoadingOutline) {
           return const CenteredCircularProgressIndicator();
         }
-        if (controller.outlineNodes.value == null ||
-            controller.outlineNodes.value.isEmpty) {
-          return const Center(
-            child: Text('Nothing to inspect'),
-          );
-        }
-        return TreeView<VMServiceObjectNode>(
-          itemExtent: _programExplorerRowHeight,
-          dataRootsListenable: controller.outlineNodes,
-          onItemSelected: onItemSelected,
-          onItemExpanded: onItemExpanded,
-          dataDisplayProvider: (node, onTap) {
-            return _ProgramExplorerRow(
-              controller: controller,
-              node: node,
-              onTap: () async {
-                await node.populateLocation();
-                controller.selectOutlineNode(node);
-                onTap();
+        return ValueListenableBuilder<List<VMServiceObjectNode>>(
+          valueListenable: controller.outlineNodes,
+          builder: (context, nodes, _) {
+            if (nodes == null || nodes.isEmpty) {
+              return const Center(
+                child: Text('Nothing to inspect'),
+              );
+            }
+            return TreeView<VMServiceObjectNode>(
+              itemExtent: _programExplorerRowHeight,
+              dataRootsListenable: controller.outlineNodes,
+              onItemSelected: onItemSelected,
+              onItemExpanded: onItemExpanded,
+              dataDisplayProvider: (node, onTap) {
+                return _ProgramExplorerRow(
+                  controller: controller,
+                  node: node,
+                  onTap: () async {
+                    await node.populateLocation();
+                    controller.selectOutlineNode(node);
+                    onTap();
+                  },
+                );
               },
             );
           },
--- a/packages/devtools_app/lib/src/debugger/program_explorer.dart
+++ b/packages/devtools_app/lib/src/debugger/program_explorer.dart
@@ -327,32 +327,25 @@ class _ProgramOutlineView extends StatelessWidget {
         if (isLoadingOutline) {
           return const CenteredCircularProgressIndicator();
         }
-        return ValueListenableBuilder<List<VMServiceObjectNode>>(
-          valueListenable: controller.outlineNodes,
-          builder: (context, nodes, _) {
-            if (nodes == null || nodes.isEmpty) {
-              return const Center(
-                child: Text('Nothing to inspect'),
-              );
-            }
-            return TreeView<VMServiceObjectNode>(
-              itemExtent: _programExplorerRowHeight,
-              dataRootsListenable: controller.outlineNodes,
-              onItemSelected: onItemSelected,
-              onItemExpanded: onItemExpanded,
-              dataDisplayProvider: (node, onTap) {
-                return _ProgramExplorerRow(
-                  controller: controller,
-                  node: node,
-                  onTap: () async {
-                    await node.populateLocation();
-                    controller.selectOutlineNode(node);
-                    onTap();
-                  },
-                );
+        return TreeView<VMServiceObjectNode>(
+          itemExtent: _programExplorerRowHeight,
+          dataRootsListenable: controller.outlineNodes,
+          onItemSelected: onItemSelected,
+          onItemExpanded: onItemExpanded,
+          dataDisplayProvider: (node, onTap) {
+            return _ProgramExplorerRow(
+              controller: controller,
+              node: node,
+              onTap: () async {
+                await node.populateLocation();
+                controller.selectOutlineNode(node);
+                onTap();
               },
             );
           },
+          emptyTreeViewBuilder: () => const Center(
+            child: Text('Nothing to inspect'),
+          ),
         );
       },
     );
--- a/packages/devtools_app/lib/src/debugger/program_explorer.dart
+++ b/packages/devtools_app/lib/src/debugger/program_explorer.dart
@@ -289,23 +289,18 @@ class _FileExplorer extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return ValueListenableBuilder<List<VMServiceObjectNode>>(
-      valueListenable: controller.rootObjectNodes,
-      builder: (context, nodes, _) {
-        return TreeView<VMServiceObjectNode>(
-          itemExtent: _programExplorerRowHeight,
-          dataRoots: nodes,
-          onItemSelected: onItemSelected,
-          onItemExpanded: onItemExpanded,
-          dataDisplayProvider: (node, onTap) {
-            return _ProgramExplorerRow(
-              controller: controller,
-              node: node,
-              onTap: () {
-                controller.selectNode(node);
-                onTap();
-              },
-            );
+    return TreeView<VMServiceObjectNode>(
+      itemExtent: _programExplorerRowHeight,
+      dataRootsListenable: controller.rootObjectNodes,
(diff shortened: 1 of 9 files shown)


packages/devtools_app/lib/src/debugger/program_explorer_controller.dart
```diff
@@ -77,15 +77,11 @@ class ProgramExplorerController extends DisposableController
       this,
       libraries,
     );
-    _rootObjectNodes.addAll(nodes);
+    _rootObjectNodes.replaceAll(nodes);
     _initialized.value = true;
   }
 
   void initListeners() {
-    addAutoDisposeListener(
```
elliette, line 85: This was removed because it was redundant - the `debugger_controller` was also listening for `serviceManager.isolateManager.selectedIsolate` [code link](https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/debugger/debugger_controller.dart#L123-L125), and was re-populating the scripts when the isolate changed. Therefore, what was happening was:

1. isolate changes -> refresh
2. sorted scripts changes -> refresh

Since we only want to refresh once we have the scripts, this first listener could be removed.

packages/devtools_app/lib/src/tree.dart
```diff
@@ -63,10 +63,8 @@ class _TreeViewState<T extends TreeNode<T>> extends State<TreeView<T>>
   @override
   void didUpdateWidget(TreeView<T> oldWidget) {
     super.didUpdateWidget(oldWidget);
-    if (widget.dataRoots != oldWidget.dataRoots) {
```
elliette, line 66: I don't think this check was ever `true`. It seems like the `oldWidget.dataRoots` were always identical to the `widget.dataRoots`, regardless of whether the `dataRoots` had changed or not. This meant the tree view wasn't being re-drawn on refresh.

packages/devtools_app/lib/src/debugger/program_explorer_controller.dart
```diff
@@ -130,6 +126,7 @@ class ProgramExplorerController extends DisposableController
 
   /// Marks [node] as the currently selected node, clearing the selection state
   /// of any currently selected node.
+  /// TODO(elliette): Scroll to node when selected.
```
kenzieschmoll, line 129: is this a regression? this was originally implemented by @bkonyi here https://github.com/flutter/devtools/pull/3480

packages/devtools_app/lib/src/utils.dart
```diff
@@ -1223,6 +1223,15 @@ class ListValueNotifier<T> extends ChangeNotifier
     _listChanged();
   }
 
+  /// Replaces all elements in the list and notifies listeners. It's preferred
+  /// to calling .clear() then .addAll(), because it only notifies listeners
+  /// once.
+  void replaceAll(Iterable<T> elements) {
+    _rawList.clear();
```
kenzieschmoll, line 1230: let's use `_rawList = <T>[];` here to be consistent with how the `clear` method clears out the list

packages/devtools_app/lib/src/debugger/program_explorer_controller.dart
elliette: That looks like it implements scrolling within the file, this comment is for scrolling within the program explorer tree view - eg: user opens `some_file.dart` at the bottom of the tree view, and then does a hot-restart. On restart, `main.dart` is selected and opened, but the tree is still scrolled to `some_file.dart`. I've added more details to the comment to make it clearer.

packages/devtools_app/lib/src/utils.dart
elliette: Good point, done!

packages/devtools_app/lib/src/tree.dart
kenzieschmoll: I looked into this for the uses of `TreeView` throughout devtools. I think we should change TreeView to accept `dataRootsListenable` instead of `dataRoots`. Then all the `ValueListenableBuilders` around `TreeView` could be removed, and `TreeView` would be responsible for listening to changes in `dataRootsListenable` and calling `setState`. The one place where we aren't using a `List<T>` listenable for the tree view is `ExpandableVariable` (where the listenable has type `T` not a `List<T>`). For this case you can just pass in the raw variable to `ExpandableVariable` and use a `FixedValueListenable` with this value for the `TreeView`

packages/devtools_app/lib/src/debugger/program_explorer_controller.dart
kenzieschmoll: got it. On another note, after a hot restart, should the selected and opened file be some_file.dart if it is still available instead of main.dart?

packages/devtools_app/lib/src/tree.dart
kenzieschmoll: Also as a cleanup as part of this change, looks like the VariablesList class can be deleted

packages/devtools_app/lib/src/debugger/variables.dart
```diff
@@ -85,26 +52,20 @@ class ExpandableVariable extends StatelessWidget {
...
-              displayProvider(context, variable, onPressed, debuggerController),
-          onItemSelected: (variable) =>
-              onItemPressed(variable, debuggerController),
-        );
-      },
+    if (variable == null) return const SizedBox();
+    // TODO(kenz): preserve expanded state of tree on switching frames and
+    // on stepping.
+    return TreeView<DartObjectNode>(
+      dataRootsListenable: ValueNotifier([variable]),
```
kenzieschmoll, line 64: Use FixedValueListenable here instead

packages/devtools_app/lib/src/tree.dart
```diff
@@ -56,21 +57,30 @@ class _TreeViewState<T extends TreeNode<T>> extends State<TreeView<T>>
   @override
   void initState() {
     super.initState();
-    _initData();
-    _updateItems();
+    widget.dataRootsListenable.addListener(_updateTreeView);
```
kenzieschmoll, line 60: can you use addAutoDisposeListener instead?

packages/devtools_app/lib/src/tree.dart
```diff
@@ -56,21 +57,30 @@ class _TreeViewState<T extends TreeNode<T>> extends State<TreeView<T>>
...
-    if (widget.dataRoots != oldWidget.dataRoots) {
-      _initData();
-      _updateItems();
-    }
+  void dispose() {
+    widget.dataRootsListenable.removeListener(_updateTreeView);
+    super.dispose();
   }
 
+  // @override
```
kenzieschmoll, line 70: remove commented out code

packages/devtools_app/lib/src/tree.dart
```diff
@@ -56,21 +57,30 @@ class _TreeViewState<T extends TreeNode<T>> extends State<TreeView<T>>
...
+  // void didUpdateWidget(TreeView<T> oldWidget) {
+  //   super.didUpdateWidget(oldWidget);
+  //   _initData();
+  //   _updateItems();
+  // }
+
   void _initData() {
-    dataRoots = List.from(widget.dataRoots);
+    dataRoots = List.from(widget.dataRootsListenable.value);
```
kenzieschmoll, line 78: just put this directly in _updateTreeView since _initData is not used anywhere else

kenzieschmoll (approved):
LGTM after comments

packages/devtools_app/lib/src/debugger/variables.dart
```diff
@@ -25,14 +25,19 @@ class Variables extends StatelessWidget {
...
-    // on stepping.
-    return TreeView<DartObjectNode>(
-      dataRootsListenable: controller.variables,
-      dataDisplayProvider: (variable, onPressed) =>
-          displayProvider(context, variable, onPressed, controller),
-      onItemSelected: (variable) => onItemPressed(variable, controller),
+    return ValueListenableBuilder<List<DartObjectNode>>(
+      valueListenable: controller.variables,
+      builder: (context, variables, _) {
+        if (variables.isEmpty) return const SizedBox();
```
kenzieschmoll, line 31: we can actually remove the ValueListenableBuilder and the special case for variables.isEmpty. TreeView returns a SizedBox in its build method when there is an empty data set

packages/devtools_app/lib/src/debugger/program_explorer.dart
```diff
@@ -327,25 +327,29 @@ class _ProgramOutlineView extends StatelessWidget {
...
-              onTap: () async {
-                await node.populateLocation();
-                controller.selectOutlineNode(node);
-                onTap();
+        return ValueListenableBuilder<List<VMServiceObjectNode>>(
+          valueListenable: controller.outlineNodes,
+          builder: (context, nodes, _) {
+            if (nodes == null || nodes.isEmpty) {
+              return const Center(
+                child: Text('Nothing to inspect'),
```
kenzieschmoll: Let's add an optional parameter to TreeView, `emptyTreeDisplayBuilder` or some better name if you can think of one that defaults to building `const SizedBox()`. Then use this value in the TreeView build method when the data is empty. I just took a look at the code, and I don't think controller.outlineNodes.value will ever be null, so we shouldn't have to worry about the null case.

packages/devtools_app/lib/src/tree.dart
elliette: Done, thanks! Also deleted VariablesList

packages/devtools_app/lib/src/debugger/program_explorer_controller.dart
elliette: Hmm good point - added a TODO to see if that's possible

packages/devtools_app/lib/src/debugger/variables.dart
elliette: Done

packages/devtools_app/lib/src/tree.dart
elliette: Done

packages/devtools_app/lib/src/tree.dart
elliette: Oops thanks!

packages/devtools_app/lib/src/tree.dart
elliette: Done

packages/devtools_app/lib/src/debugger/variables.dart
elliette: Oh nice! Good point

packages/devtools_app/lib/src/debugger/program_explorer.dart
elliette: Done, added optional parameter

PR #3531, opened 2021-12-10
Fix state issue with CpuProfiler user tags

kenzieschmoll:
This allows us to remove the additional ValueListenableBuilder for `controller.cpuProfileData`. We weren't properly clearing out the user tag state in the controller, and we weren't updating `_CpuProfilerState` in `didUpdateWidget` when the cpu profile data changed.

### DIFF
--- a/packages/devtools_app/lib/src/profiler/cpu_profiler.dart
+++ b/packages/devtools_app/lib/src/profiler/cpu_profiler.dart
@@ -82,9 +82,12 @@ class _CpuProfilerState extends State<CpuProfiler>
         SearchFieldMixin<CpuProfiler> {
   TabController _tabController;
 
+  CpuProfileData data;
+
   @override
   void initState() {
     super.initState();
+    data = widget.data;
     _initTabController();
   }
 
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/profiler/cpu_profiler.dart
```diff
@@ -94,6 +94,9 @@ class _CpuProfilerState extends State<CpuProfiler>
     if (widget.tabs.length != oldWidget.tabs.length) {
       _initTabController();
     }
+    if (widget.data != oldWidget.data) {
+      setState(() {});
```
elliette, line 98: OOC, what does calling `setState` with no data do? Trigger a refresh of the widget?

elliette (approved):

packages/devtools_app/lib/src/profiler/cpu_profiler.dart
kenzieschmoll: setState will force the widget to be rebuilt. Calling an empty setState should be avoided though where possible, for this reason - it makes it hard to read and understand what state has changed and why we need to rebuild the widget. I violated best practices here, so I updated it to actually set a `data` var with the updated value we need to rebuild with.

PR #3532, opened 2021-12-10
Follow best practices for creating FocusNode objects 

elliette:
Continuing work towards [3457](https://github.com/flutter/devtools/issues/3457). 

This PR works towards fixing the behavior described in https://github.com/flutter/devtools/issues/3457#issuecomment-946243954. It updates how we are using `FocusNode` objects to follow Flutter's ["Best practices for creating FocusNode objects"](https://docs.flutter.dev/development/ui/advanced/focus#best-practices-for-creating-focusnode-objects), specifically:

* Don’t allocate a new FocusNode for each build. This can cause memory leaks, and occasionally causes a loss of focus when the widget rebuilds while the node has focus.
* Do create FocusNode and FocusScopeNode objects in a stateful widget. FocusNode and FocusScopeNode need to be disposed of when you’re done using them, so they should only be created inside of a stateful widget’s state object, where you can override dispose to dispose of them.

### DIFF
--- a/packages/devtools_app/lib/src/charts/flame_chart.dart
+++ b/packages/devtools_app/lib/src/charts/flame_chart.dart
@@ -250,6 +250,8 @@ abstract class FlameChartState<T extends FlameChart,
         }
       });
     }
+
+    autoDisposeFocusNode(focusNode);
   }
 
   @override
@@ -268,7 +270,6 @@ abstract class FlameChartState<T extends FlameChart,
   @override
   void dispose() {
     zoomController.dispose();
-    focusNode.dispose();
     super.dispose();
   }
 
--- a/packages/devtools_app/lib/src/charts/flame_chart.dart
+++ b/packages/devtools_app/lib/src/charts/flame_chart.dart
@@ -268,6 +268,7 @@ abstract class FlameChartState<T extends FlameChart,
   @override
   void dispose() {
     zoomController.dispose();
+    focusNode.dispose();
     super.dispose();
   }
 
--- a/packages/devtools_app/lib/src/memory/memory_screen.dart
+++ b/packages/devtools_app/lib/src/memory/memory_screen.dart
@@ -98,7 +98,7 @@ class MemoryBodyState extends State<MemoryBody>
   /// Updated when the MemoryController's _androidCollectionEnabled ValueNotifier changes.
   bool isAndroidCollection = MemoryController.androidADBDefault;
 
-  final FocusNode _focusNode = FocusNode(debugLabel: 'memory');
+  final _focusNode = FocusNode(debugLabel: 'memory');
 
   @override
   void initState() {
--- a/packages/devtools_app/lib/src/memory/memory_screen.dart
+++ b/packages/devtools_app/lib/src/memory/memory_screen.dart
@@ -98,10 +98,13 @@ class MemoryBodyState extends State<MemoryBody>
   /// Updated when the MemoryController's _androidCollectionEnabled ValueNotifier changes.
   bool isAndroidCollection = MemoryController.androidADBDefault;
 
+  final FocusNode _focusNode = FocusNode(debugLabel: 'memory');
+
   @override
   void initState() {
     super.initState();
     ga.screen(MemoryScreen.id);
+    autoDisposeFocusNode(_focusNode);
   }
 
   @override
@@ -303,7 +306,7 @@ class MemoryBodyState extends State<MemoryBody>
 
     // TODO(terry): Can Flutter's focus system be used instead of listening to keyboard?
     return RawKeyboardListener(
-      focusNode: FocusNode(debugLabel: 'memory'),
+      focusNode: _focusNode,
       onKey: (RawKeyEvent event) {
         if (event.isKeyPressed(LogicalKeyboardKey.escape)) {
           hideHover();
--- a/packages/devtools_app/lib/src/ui/search.dart
+++ b/packages/devtools_app/lib/src/ui/search.dart
@@ -587,6 +587,7 @@ typedef OverlayXPositionBuilder = double Function(
 mixin SearchFieldMixin<T extends StatefulWidget> on State<T> {
   TextEditingController searchTextFieldController;
   FocusNode _searchFieldFocusNode;
+  FocusNode _rawKeyboardFocusNode;
   SelectAutoComplete _onSelection;
   void Function() _closeHandler;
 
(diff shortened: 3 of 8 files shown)


packages/devtools_app/lib/src/memory/memory_screen.dart
```diff
@@ -98,10 +98,13 @@ class MemoryBodyState extends State<MemoryBody>
   /// Updated when the MemoryController's _androidCollectionEnabled ValueNotifier changes.
   bool isAndroidCollection = MemoryController.androidADBDefault;
 
+  final FocusNode _focusNode = FocusNode(debugLabel: 'memory');
```
kenzieschmoll, line 101: nit: you can remove type on LHS

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -606,6 +609,7 @@ mixin SearchFieldMixin<T extends StatefulWidget> on State<T> {
     super.dispose();
     searchTextFieldController?.dispose();
     _searchFieldFocusNode?.dispose();
+    _rawKeyboardFocusNode?.dispose();
```
kenzieschmoll, line 612: why not use autoDisposeFocusNode instead of manually disposing?

packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -268,6 +268,7 @@ abstract class FlameChartState<T extends FlameChart,
   @override
   void dispose() {
     zoomController.dispose();
+    focusNode.dispose();
```
kenzieschmoll, line 271: same question - why not use autoDisposeFocusNode?

packages/devtools_app/lib/src/memory/memory_screen.dart
elliette: Done!

packages/devtools_app/lib/src/ui/search.dart
elliette: Hmm can I use a `mixin` in a `mixin` (ie `AutoDisposeMixin` within the `SearchFieldMixin`)? I'm not finding documentation about that.

packages/devtools_app/lib/src/charts/flame_chart.dart
elliette: Good point, done!

packages/devtools_app/lib/src/ui/search.dart
kenzieschmoll: You should be able to add `AutoDisposeMixin` to the `on` list:
`mixin SearchFieldMixin<T extends StatefulWidget> on State<T>, AutoDisposeMixin<T>`

elliette: Hmm I thought that restricted what classes could use the `mixin` (eg, `SearchFieldMixin<T>` can only be used on classes that implement both `State<T>` and `AutoDisposeMixin<T>`?

I'm seeing the following error on `FileSearchFieldState` (and other places): 

'SearchFieldMixin<StatefulWidget>' can't be mixed onto 'State<FileSearchField>' because 'State<FileSearchField>' doesn't implement 'AutoDisposeMixin<StatefulWidget>'.
Try extending the class 'SearchFieldMixin<StatefulWidget>'.

kenzieschmoll: If the `SearchFieldMixin` depends on functionality from `AutoDisposeMixin`, I think it is fine to force consumers of `SearchFieldMixin` to mixin both `SearchFieldMixin` and `AutoDisposeMixin`.

kenzieschmoll: CC @leafpetersen @munificent Is there a best practice around mixins depending on other mixins?

leafpetersen: cc @lrhn @natebosch for thoughts on best practices here.

lrhn: I'm not aware of any official or semi-official best practices.
I'd personally have no problem making one mixin depend on another, in order to force you to always include the other.

You can do that using the `on`, which requires the `AutoDisposeMixin` to be mixed in first. That's probably the safest approach, and it's very explicit about not being allowed to mix in `SearchFieldMixin` without first mixing in `AutoDisposeMixin`. (Not amazing, though, "Try extending the class 'SearchFieldMixin'" isn't really helping.)

You can also let `SearchFieldMixin` do `implements AutoDisposeMixin`, without actually implementing the members. That means that the user must mix in `AutoDisposeMixin` too (or implement its interface manually) in order to satisfy the interface, but the error message are not going to be as directly actionable.


Anyway, why is `SearchFieldMixin` a mixin to begin with, and not just a `Widget`. This feels like something where you'd want composition rather than inheritance, so a widget shouldn't be a search field *as well* as something else, it should be (or have) the something else and also *have* a search field. (Can something have two search fields?)

The two uses of `SearchFieldMixin` I see in internal code are using it as `extends State<Something> with SearchFieldMixin, AutoDisposeMixin`, so they would be equally well helped by having `class SearchFieldMixin<T extends Widget> extends State<T> with AutoDisposeMixin { ... }` and extending that. 
I'm not seeing the advantage of being a `mixin` here.

kenzieschmoll: > why is SearchFieldMixin a mixin to begin with, and not just a Widget

It's been a while since this code was written, but I believe this was made a mixin so that the State classes that mixin SearchFieldMixin can call these three methods from the mixin: https://github.com/flutter/devtools/blob/136617a5e3f44e6837f2ebb4ee137f46d0f236c3/packages/devtools_app/lib/src/ui/search.dart#L704-L735. These may be able to be pushed into the controller, but this would require some extra plumbing to have the text field controller respond to events from the business logic controller, which may be a little messy.

Either way, it sounds like it is fine for now to force classes mixing in SearchFieldMixin to also mixin AutoDisposeMixin.

kenzieschmoll (approved):

packages/devtools_app/lib/src/ui/search.dart
natebosch: +1 on moving towards composition for this design. FWIW I find most designs with an imposed `with` or `extends` hard to follow especially when they mix methods which are supposed to be called by others (`initState` and `dispose` here I think) alongside methods that are supposed to be called by the extending class.

elliette: Thanks everyone for clarifying! I've added a TODO for refactoring `SearchFieldMixin`. It looks like we currently can't use the `AutoDisposeMixin` with it due to complications with how the `AutoDisposeMixin` is structured (opened https://github.com/flutter/devtools/issues/3538 for that). Will address both in separate PRs to keep the scope of this change limited to `FocusNodes`.

PR #3544, opened 2021-12-16
Stop using package:intl in devtools_server.

scheglov:
This is a long tail of changes.
https://dart-review.googlesource.com/c/sdk/+/224520
https://github.com/dart-lang/intl/pull/436
https://dart-review.googlesource.com/c/sdk/+/224660

### DIFF
--- a/packages/devtools_app/lib/src/utils.dart
+++ b/packages/devtools_app/lib/src/utils.dart
@@ -617,7 +617,7 @@ class TimeRange {
 }
 
 String formatDateTime(DateTime time) {
-  return DateFormat('H:mm:ss.S a').format(time);
+  return DateFormat('H:mm:ss.S').format(time);
 }
 
 bool isDebugBuild() {
--- a/packages/devtools_app/lib/src/utils.dart
+++ b/packages/devtools_app/lib/src/utils.dart
@@ -617,7 +617,7 @@ class TimeRange {
 }
 
 String formatDateTime(DateTime time) {
-  return DateFormat('h:mm:ss.S a').format(time);
+  return DateFormat('H:mm:ss.S a').format(time);
 }
 
 bool isDebugBuild() {
--- a/packages/devtools_server/lib/src/memory_profile.dart
+++ b/packages/devtools_server/lib/src/memory_profile.dart
@@ -337,7 +337,7 @@ class MemoryProfile {
       return '0' * (length - result.length) + result;
     }
 
-    return toStringLength(value.hour, 2) +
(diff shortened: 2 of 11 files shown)


packages/devtools_server/lib/src/memory_profile.dart
```diff
@@ -331,6 +330,21 @@ class MemoryProfile {
     }
     return heapList;
   }
+
+  static String _formatTime(DateTime value) {
+    String toStringLength(int value, int length) {
+      final result = '$value';
+      return '0' * (length - result.length) + result;
```
kenzieschmoll, line 337: I'm not familiar with this syntax. When length - result.length is negative, what will this look like?

Take the example value = 123456, length = 3. What does this method yield?

scheglov: We call this function for values that have less digits than `length`.
I will add `assert`.

kenzieschmoll: instead of asserting, I think we need to handle all cases:
```dart
if (length == result.length) {
  return result;
}
if (length > result.length) {
 return '0' * (length - result.length) + result;
}
return result.substring(0, length);
```

scheglov: OK, although we know that we never have more than 24 hours, 60 minutes, and 60 seconds. So, never longer than 2 digits. But this is your code, so this is fine with me.

kenzieschmoll (approved):

packages/devtools_server/lib/src/memory_profile.dart
kenzieschmoll: Ah I see what you are saying. In that case you can revert to what you had. Apologies for the churn.

packages/devtools_app/lib/src/utils.dart
```diff
@@ -617,7 +617,7 @@ class TimeRange {
 }
 
 String formatDateTime(DateTime time) {
-  return DateFormat('h:mm:ss.S a').format(time);
+  return DateFormat('H:mm:ss.S a').format(time);
```
kenzieschmoll, line 620: if we are using military time, we can get rid of AM / PM

scheglov: Done.

scheglov:
I made a few more changes, specifically to test expectations to adapt them to the fact that we are using military time. Please check if these changes are what we want.

kenzieschmoll (approved):

PR #3545, opened 2021-12-17
Update package:vm_service to ^8.1.0

bkonyi:


kenzieschmoll:
Looks like we may need to update dds as well?
```
Because dds >=2.1.6 depends on vm_service ^7.5.0 and dds >=2.1.0 <2.1.6 depends on vm_service ^7.2.0, dds >=2.1.0 requires vm_service ^7.2.0.
So, because devtools_app depends on both dds ^2.1.3 and vm_service ^8.1.0, version solving failed.
```

elliette (approved):

PR #3546, opened 2021-12-17
Add --version flag to DevTools server command

bkonyi:
Also supports machine readable output

### DIFF
--- a/packages/devtools_server/lib/src/server.dart
+++ b/packages/devtools_server/lib/src/server.dart
@@ -41,6 +41,7 @@ const argPort = 'port';
 const argProfileMemory = 'record-memory-profile';
 const argTryPorts = 'try-ports';
 const argVerbose = 'verbose';
+const argVersion = 'version';
 const launchDevToolsService = 'launchDevTools';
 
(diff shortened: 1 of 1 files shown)


packages/devtools_server/lib/src/server.dart
```diff
@@ -113,6 +115,18 @@ Future<HttpServer?> _serveDevToolsWithArgs(
     return null;
   }
 
+  if (version) {
+    final versionStr = await _getVersion();
+    printOutput(
+      'Dart DevTools version $versionStr',
+      {
+        'version': versionStr,
```
DanTup, line 123: It looks like `_getVersion` can return the string `'unknown'` if it can't get the version. Would it be better to do something else (like `null` or `0.0.0`) for the JSON mode?

kenzieschmoll (approved):

packages/devtools_server/lib/src/server.dart
bkonyi: Just returning `'unknown'` is probably more clear than returning `0.0.0` or `null`, so I think this makes the most sense for now.

PR #3552, opened 2021-12-22
Experimental: widget detail search (#3489)

aednlaxer:
This pull request is based on #3525 and adds search of widget properties and property values. I'd like to ask you to evaluate the performance of details search and to give your opinion on that (maybe as a comment to ticket #3489). This PR is not ready for review although all comments are welcome.

Details search can be enabled by choosing 'All' or 'Details' in the search box dropdown:

dropdown

dropdown-selected

There's room for improvement. Some ideas how to make user experience better:
* Search should respect value of `searchPreviousMatches` (commented out for now)
* Colors are stored as `int` components so they're not found when search query is a hex code
* Search matches could be displayed as soon as there's a match, not waiting for the whole search to complete
* Property search matches are not instances of `DataSearchStateMixin` so the total number of search matches doesn't include the number of matching properties. Previous\next buttons don't navigate among these matches for the same reason.


kenzieschmoll:
Now that #3525 is merged, can you merge master or rebase so that this change is easier to review? thank you

aednlaxer:
Done. Please review it as a prototype for evaluating performance rather than a production ready changes. I'd like to know if it makes any sense to continue working on this. I'm personally not very happy with the performance at the moment. I'd want to explore a possible improvement - showing matches as soon as they're found without waiting for search to complete

kenzieschmoll:
Overall, I think the performance is okay. If we can speed up the search at all, even better. I tested this patch out on a large internal app, and with the search scope "All" and a one-letter search query, the search was completing in ~5-6 seconds. The search spinner notifies the user that something is happening, so the UI doesn't feel stalled. If we want, we could warn the user that when they select the "details" or "all" scope, searches may take a bit longer than if the "widget" scope is selected. @InMatrix thoughts?

@aednlaxer - Is the worst case search time I was seeing comparable to what you experienced when working on this feature? Is there another way I should be testing the worst case other than with a large app and a one letter search query?

A couple other comments:
- Perhaps it makes sense to rename the search scopes to "Widgets" and "Properties" instead of "Widget" and "Details".
- I found it a bit confusing that the search navigator arrows don't navigate through matches in the details tree. For example, if item A is selected in the widget tree, and the details tree for item A has two property matches that also match the search query, navigating to the next search match doesn't go to item-A-details-tree-property-match-1, it goes to the next widget tree search match. @InMatrix what is your opinion on this behavior?

kenzieschmoll:
One more comment. It appears in the current state that only the properties are coming up as matches when the search scope is Details or All. Here's an example where the search query is "in" but only properties in the details tree are highlighted, even though there are widgets in the details tree that also contain the "in" string in their name. This seems like a bug and that we would want to include both widgets and properties in search matches within the details / all scopes.
Screen Shot 2022-01-06 at 3 23 45 PM

aednlaxer:
@kenzieschmoll thank you for your comments. Sorry, I didn't have time to focus on this recently but I'd like to work on improving this soon.

I was connecting to devtools itself to experiment with properties search and was getting similar performance results.

Hixie:
@aednlaxer is this still something you are interested in working on?

PR #3566, opened 2022-01-06
Add FrameTimeVisualizer to janky frame analysis view

kenzieschmoll:
Screen Shot 2022-01-06 at 1 44 15 PM
I am still not sure whether the selected color has enough contrast with the background in light mode. @ayanogawa what do you think?
Screen Shot 2022-01-06 at 1 44 59 PM

This feature is still disabled by default. Will continue to iterate in follow up PRs. Up next is adding bottom up / top down visualizations for events in the selected frame phase.


packages/devtools_app/lib/src/performance/performance_model.dart
```diff
@@ -1181,3 +1183,110 @@ class AsyncTimelineEvent extends TimelineEvent {
...
+      buildEvents.firstWhere((buildEvent) => buildEvent.children.isNotEmpty) !=
+      null;
+
+  bool get canAnalyzeLayout => layoutEvent.children.isNotEmpty;
+
+  bool get canAnalyzePaint => paintEvent.children.isNotEmpty;
+
+  // TODO(kenz): calculate ratios to use as flex values. This will be a bit
+  // tricky because sometimes the Build event(s) are children of Layout.
+  // int buildTimeRatio() {
```
elliette, line 1267: Is this intended to be commented out as part of the TODO?

packages/devtools_app/lib/src/performance/timeline_analysis.dart
```diff
@@ -32,7 +33,8 @@ class TimelineAnalysisHeader extends StatelessWidget {
     final showFrameAnalysisButton = frameAnalysisSupported &&
         (controller.selectedFrame.value
                 ?.isUiJanky(controller.displayRefreshRate.value) ??
-            false);
+            false) &&
+        (!(controller.selectedFrame.value?.timelineEventData?.isEmpty ?? true));
```
elliette, line 37: Nit: I think this would be clearer as `controller.selectedFrame.value?.timelineEventData?.isNotEmpty ?? false`

elliette (approved):

packages/devtools_app/lib/src/performance/performance_model.dart
kenzieschmoll: yes it is. This logic is incorrect in it's current state, so commenting out until I fix it.

packages/devtools_app/lib/src/performance/timeline_analysis.dart
kenzieschmoll: good idea. done.

Issue #3554, opened 2021-12-23
File search improvements 


elliette:
A couple file search improvements requested by a customer:

- [ ] Multi-token support (eg, allow searching for both "controller" and "banner" with space in the middle)
- [ ] Regexp support

PR #3569, opened 2022-01-15
Adds a utility function to transform AutoCompleteMatch

elliette:
The new utility function is re-used in the `file_search_test` to capitalize matching segments, making it a lot easier to verify that the output is expected. 

Work towards: https://github.com/flutter/devtools/issues/3554

### DIFF
--- a/packages/devtools_app/lib/src/ui/search.dart
+++ b/packages/devtools_app/lib/src/ui/search.dart
@@ -380,55 +380,21 @@ class AutoCompleteState extends State<AutoComplete> with AutoDisposeMixin {
     TextStyle regularTextStyle,
     TextStyle highlightedTextStyle,
   ) {
-    final text = match.text;
-    final matchedSegments = match.matchedSegments;
-
-    if (matchedSegments == null || matchedSegments.isEmpty) {
-      return TextSpan(
-        text: text,
+    return transformAutoCompleteMatch<TextSpan>(
+      match: match,
+      transformMatchedSegment: (segment) => TextSpan(
+        text: segment,
+        style: highlightedTextStyle,
+      ),
+      transformUnmatchedSegment: (segment) => TextSpan(
+        text: segment,
         style: regularTextStyle,
-      );
-    }
-
-    final spans = <TextSpan>[];
-    int previousEndIndex = 0;
-
-    for (final segment in matchedSegments) {
-      if (previousEndIndex < segment.begin) {
-        // Add the unhighlighted segment before the current highlighted segment:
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -1256,4 +1221,36 @@ class AutoCompleteMatch {
 
   final String text;
   final List<Range> matchedSegments;
+
+  // Transform the autocomplete match somehow (e.g. create a TextSpan where the
+  // matched segments are highlighted).
+  T transformAutoCompleteMatch<T>(
+      {T transformMatchedSegment(String segment),
+      T transformUnmatchedSegment(String segment),
+      T combineSegments(List<T> segments)}) {
```
kenzieschmoll, line 1230: nit: trailing comma

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -380,55 +380,20 @@ class AutoCompleteState extends State<AutoComplete> with AutoDisposeMixin {
...
-      );
-    }
-
-    return TextSpan(
-      text: spans.first.text,
-      style: spans.first.style,
-      children: spans.sublist(1),
+      ),
+      combineSegments: (segments) => TextSpan(
+        text: segments.first.text,
```
kenzieschmoll, line 393: We should have some length checks on `segments`. `.first` and `.sublist` will throw exceptions if the list is empty

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -1256,4 +1221,36 @@ class AutoCompleteMatch {
 
   final String text;
   final List<Range> matchedSegments;
+
+  // Transform the autocomplete match somehow (e.g. create a TextSpan where the
```
kenzieschmoll, line 1225: use /// for dart doc

kenzieschmoll (approved):

packages/devtools_app/lib/src/ui/search.dart
elliette: Good catch, done!

packages/devtools_app/lib/src/ui/search.dart
elliette: Good point, added an `assert` in the `transformAutocompleteMatch` method to verify segments is not empty.

packages/devtools_app/lib/src/ui/search.dart
elliette: Done, thanks!

PR #3573, opened 2022-01-18
Refactors file search to use `FileSearchResults` and `FileQuery` classes 

elliette:
Continued work towards https://github.com/flutter/devtools/issues/3554

This change should be a no-op, it simply refactors the FileSearch widget to make it easier to add support for multi-token and regexp search.

This refactor moves logic from the FileSearch widget into two utility classes:
1. `FileQuery`: has methods to determine whether a file path matches the query, and to build an `AutoCompleteMatch` for the file
2. `FileSearchResults`: finds all the matches for a query and provides a `getter` for the top X matches.

### DIFF
--- a/packages/devtools_app/lib/src/debugger/file_search.dart
+++ b/packages/devtools_app/lib/src/debugger/file_search.dart
@@ -218,6 +218,7 @@ class FileSearchResults {
     @required this.query,
     @required this.allScripts,
   }) {
+    // ignore: prefer_asserts_in_initializer_lists
     assert(!query.isEmpty);
 
     for (final scriptRef in allScripts) {
--- a/packages/devtools_app/lib/src/debugger/file_search.dart
+++ b/packages/devtools_app/lib/src/debugger/file_search.dart
@@ -135,22 +135,9 @@ class FileSearchFieldState extends State<FileSearchField>
       return FileSearchResults.emptyQuery(scriptRefs);
     }
 
-    final exactFullPathMatches = <ScriptRef>[];
-    final fuzzyMatches = <ScriptRef>[];
-    final fileQuery = FileQuery(query);
-
-    for (final scriptRef in scriptRefs) {
-      if (fileQuery.isExactFullPathMatch(scriptRef)) {
-        exactFullPathMatches.add(scriptRef);
-      } else if (fileQuery.isFuzzyMatch(scriptRef)) {
-        fuzzyMatches.add(scriptRef);
-      }
-    }
-
     return FileSearchResults.withQuery(
-      exactFullPathMatches: exactFullPathMatches,
-      fuzzyMatches: fuzzyMatches,
-      query: fileQuery,
+      allScripts: scriptRefs,
+      query: FileQuery(query),
     );
   }
 }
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -124,72 +123,187 @@ class FileSearchFieldState extends State<FileSearchField>
...
+    );
+  }
+}
+
+class FileQuery {
+  FileQuery(this.query) : assert(query.isNotEmpty);
+
+  FileQuery.empty() : query = '';
+
+  String query;
```
kenzieschmoll, line 150: this should be able to be final

packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -124,72 +123,187 @@ class FileSearchFieldState extends State<FileSearchField>
...
+}
 
-  List<ScriptRef> _findMatches(
-    String query,
-    List<ScriptRef> scriptRefs,
-  ) {
+class FileSearchResults {
+  FileSearchResults.emptyQuery(this.allScripts) : query = FileQuery.empty();
+
+  FileSearchResults.withQuery({
```
kenzieschmoll, line 217: nit use factory constructors instead. You should be able to remove the "// ignore: ..." line by using a factory constructor since asserts are allowed in the body of a factory constructor

packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -124,72 +123,187 @@ class FileSearchFieldState extends State<FileSearchField>
...
-          _fileNamesCache[scriptRef.uri] ??= scriptRef.uri.split('/').last;
-      if (fullPath.caseInsensitiveContains(query)) {
-        exactMatches.add(scriptRef);
-      } else if (fileName.caseInsensitiveFuzzyMatch(query)) {
-        fuzzyMatches.add(scriptRef);
-      }
+    if (scriptRefs.length <= numOfMatchesToShow) {
+      // Make a copy of this:
+      return FileSearchResults._(
+        allScripts: allScripts,
```
kenzieschmoll, line 277: you can make a `copyWith` method that defaults to all of the values passed in here and allows you to optionally pass in something different. You could use this below where you pass 
`    exactFullPathMatches: topMatches[0],
      fuzzyMatches: topMatches[1],`

packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -31,12 +33,11 @@ class FileSearchFieldState extends State<FileSearchField>
   AutoCompleteController autoCompleteController;
 
   final _scriptsCache = <String, ScriptRef>{};
-  final _fileNamesCache = <String, String>{};
 
   final fileSearchFieldKey = GlobalKey(debugLabel: 'fileSearchFieldKey');
 
   String _query;
-  List<ScriptRef> _matches;
+  FileSearchResults _searchResults;
```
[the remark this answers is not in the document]
elliette: Thank you, done!

packages/devtools_app/lib/src/debugger/file_search.dart
```diff
@@ -50,7 +51,8 @@ class FileSearchFieldState extends State<FileSearchField>
         _handleAutoCompleteOverlay);
 
     _query = autoCompleteController.search;
-    _matches = widget.debuggerController.sortedScripts.value;
+    _searchResults = FileSearchResults.emptyQuery(
+        widget.debuggerController.sortedScripts.value);
```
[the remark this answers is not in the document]
elliette: Done

packages/devtools_app/lib/src/debugger/file_search.dart

packages/devtools_app/lib/src/debugger/file_search.dart
elliette: Oh perfect, thanks!

packages/devtools_app/lib/src/debugger/file_search.dart
elliette: Good idea - switched to using `copyWith`

kenzieschmoll (approved):

PR #3574, opened 2022-01-19
Quick fix to survey url parsing bug

kenzieschmoll:
This is a quick fix that we will try to cherry pick into flutter. The more robust fix is to solve https://github.com/flutter/devtools/issues/2475 and figure out if we can make our url able to be parsed by `Uri.parse`, but this fix will ensure that we do not prevent the DevTools survey from being able to go out (if we can cherry-pick in to flutter that is).

### DIFF
--- a/packages/devtools_app/lib/src/analytics/_analytics_web.dart
+++ b/packages/devtools_app/lib/src/analytics/_analytics_web.dart
@@ -689,11 +689,19 @@ Map<String, dynamic> generateSurveyQueryParameters() {
   final url = window.location.toString();
   const fromValuePrefix = '#/';
   final startIndex = url.indexOf(fromValuePrefix);
-  final endIndex = url.indexOf('?');
-  final fromPage = url.substring(
-    startIndex + fromValuePrefix.length,
-    endIndex,
-  );
+  // Use the last index because the url can be of the form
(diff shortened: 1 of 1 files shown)


kenzieschmoll:
Website PR here to push back the survey dates and make sure our users don't hit this issue: https://github.com/flutter/website/pull/6696. Today was supposed to be the first day of the survey, so hopefully not many users will hit this issue.

packages/devtools_app/lib/src/analytics/_analytics_web.dart
```diff
@@ -689,11 +689,19 @@ Map<String, dynamic> generateSurveyQueryParameters() {
...
-  final fromPage = url.substring(
-    startIndex + fromValuePrefix.length,
-    endIndex,
-  );
+  // Use the last index because the url can be of the form
+  // 'http://127.0.0.1:9103/?#/?' and we want to be referencing the last '?'
+  // character.
+  final endIndex = url.lastIndexOf('?');
+  var fromPage = '';
+  try {
```
jacob314, line 697: probably check that endIndex !=-1 before this to avoid having to throw an exceptions.

jacob314 (approved):
lgtm

PR #3585, opened 2022-01-24
Switch to using Path URL Strategy

### LINKED ISSUE #2475: DevTools url has fragment and query parameters in the wrong order

DanTup:
This changes the app to use the PageUrlStrategy in Flutter, which means URLs will just be standard like `/foo` instead of being added to the fragment like `/#/foo`.

@kenzieschmoll this seems to work well from my testing, including from VS Code ([with a small tweak](https://github.com/Dart-Code/Dart-Code/commit/cb0bb3f46789048acba7632c5e098fe5151559b8) I'll ship in the next week or so), although I'm not certain there aren't pages with unusual navigation I may have missed.

Some things I'm not certain but should be checked before merging though"

- [x] It doesn't affect IntelliJ/Android Studio negatively (VS Code requires a change because it's over-encoding or URLs was causing issues here)
- [ ] It doesn't affect anyone not using `defaultHandler` in the server... I've had to extend `defaultHandler` to be able to serve up `index.html` when a request comes in for something like `/memory`, because although navigations are usually done client-side, it's possible that the user hits Refresh or we try to launch directly on to a page which will now send `/page_id` to the server instead of just `/` which won't otherwise get the contents of `index.html` to launch the Flutter app.
- [x] Survey parsing may need updating

I'm not sure who best knows the answer to these questions though.

Fixes #2475.


DanTup:
Oh also, survey parsing seems like it'll need updating - I'm not certain how to test that though. If you let me know I can ensure that's working correctly.

DanTup:
@helin24 I don't know if this will affect IntelliJ/Android Studio. The reason VS Code needed a change was that the URLs were over-encoded which only seemed ok when we were using the fragment.

kenzieschmoll:
Generating the survey url does need updating. The value of `fromPage` [here](https://github.com/flutter/devtools/blob/44613a5e009c079457f1a71bf9c3ba703efb742a/packages/devtools_app/lib/src/analytics/_analytics_web.dart#L712) should reflect which page a user clicked the survey link from (e.g. '', 'performance', 'inspector', etc.).

You can force the survey prompt to come up by deleting the .devtools file (`rm ~/.flutter-devtools/.devtools`) and manually changing the date on your computer to a date between 2/19 and 3/19. You'll still need to build the release version of the app to see this since the survey logic requires the server.

DanTup:
Done! I extracted the code that extracts the page to a non-web-specific file so it was easier to add tests for it - I don't know if analytics_common was the right place for it, let me know if it's better elsewhere.

Btw when I was testing this it worked fine running from the CLI on a new port, but when launching on the normal port from VS Code, the app seemed heavily cached and I had to use Chrome devtools to "clear application data" before the change took affect. I don't know if this is just because it's not a real release or if other users may see this.

packages/devtools_app/lib/src/analytics/analytics_common.dart
```diff
@@ -23,3 +23,15 @@
 /// Then, add your fields to the `GtagEventDevTools.withScreenMetrics` factory
 /// constructor.
 abstract class ScreenAnalyticsMetrics {}
+
+/// Extracts the current DevTools page from the given [url].
+String extractCurrentPageFromUrl(String url) {
```
kenzieschmoll, line 28: let's move this to url_utils.dart, and then you can add a test to url_utils_test.dart

kenzieschmoll:
I found an issue when connecting to an app from the connect screen. When doing this, the 'inspector' page does not show up in the url. This is not specific to the inspector, but rather the first page shown after connecting DevTools to the running app.

Repro steps:
1. run devtools
2. copy and paste a vm service uri for a running flutter app into the connect screen
3. observe that 'inspector' is not present in the url. If you navigate to another page, and then click the inspector tab to go to the inspector page, you'll notice 'inspector' is now present.

DanTup:
hmm, yeah - I can repro this. You can also just go direct to a URL like this:

http://127.0.0.1:9104/?uri=ws%3A%2F%2F127.0.0.1%3A59526%2FIRmSRy-94Gk%3D%2Fws

I think this is because we're just showing the first item in the tabbed control when there's no page, so it's valid to not have one and we'll show whatever is the first item.

If we never want this, I think probably we could check if there's a page in the URL after we connect, and if not, explicitly push the first visible screen?

kenzieschmoll:
> I think probably we could check if there's a page in the URL after we connect, and if not, explicitly push the first visible screen

Yeah, once we show a tab, we should verify that the page name is in the url, and if not, we should add it.

packages/devtools_server/lib/src/external_handlers.dart
```diff
@@ -34,11 +34,14 @@ Future<shelf.Handler> defaultHandler(
 
   // Default static handler for all non-package requests.
   Handler? buildDirHandler;
+  File? indexFile;
```
kenzieschmoll, line 37: not your fault, but can you remove line 13 'import 'package:shelf/shelf.dart';' as that is a duplicate import

packages/devtools_server/lib/src/external_handlers.dart
```diff
@@ -115,3 +129,11 @@ Future<String> _resolveBuildDir() async {
   final packageDir = path.dirname(path.dirname(resourceUri!.toFilePath()));
   return path.join(packageDir, 'build');
 }
+
+/// Servers [file] for all requests.
```
kenzieschmoll, line 133: nit: serves file?

packages/devtools_server/lib/src/external_handlers.dart
```diff
@@ -97,7 +100,18 @@ Future<shelf.Handler> defaultHandler(
     if (debugMode) {
       return debugProxyHandler!(request);
     } else {
-      return buildDirHandler!(request);
+      // If the user hits Refresh when on a page being handled by the Flutter
```
kenzieschmoll, line 103: The changes to this file will need to be applied to the g3 handler as well, but I have verified that things work as expected when that is done. I'll update the g3 handler when this change rolls into g3.

packages/devtools_app/lib/main.dart
```diff
@@ -16,8 +16,11 @@ import 'src/extension_points/extensions_external.dart';
 import 'src/globals.dart';
 import 'src/preferences.dart';
 import 'src/provider/riverpod_error_logger_observer.dart';
+import 'src/url_strategy/url_strategy.dart';
 
 void main() async {
+  usePathUrlStrategy();
```
kenzieschmoll, line 22: this will need to be applied to the google3 main file. I'll take an action item to do this when this change is rolled into g3.

DanTup:
I've pushed changes that should fix the URL thing. One updates the URL when the tab controller is initialised (so we know which first tab is being rendered if there isn't one in the QS), and one removes some code that had a bug if you went to `/?page=memory` then tried to navigate to another page, it was a bit confused about the current page. (Ultimately support for `page=` should probably be dropped, but IDEs are currently still using it).

packages/devtools_server/lib/src/external_handlers.dart
```diff
@@ -130,10 +129,10 @@ Future<String> _resolveBuildDir() async {
   return path.join(packageDir, 'build');
 }
 
-/// Servers [file] for all requests.
-Future<Response> _serveStaticFile(
-    Request request, File file, String contentType) async {
+/// Serves [file] for all requests.
+Future<shelf.Response> _serveStaticFile(
+    shelf.Request request, File file, String contentType) async {
```
kenzieschmoll, line 134: nit: trailing comma

kenzieschmoll:
> Ultimately support for page= should probably be dropped, but IDEs are currently still using it

@DanTup @helin24 can we transition IDEs off of this pattern?

kenzieschmoll (approved):
lgtm once tests pass

helin24:
> > Ultimately support for page= should probably be dropped, but IDEs are currently still using it
> 
> @DanTup @helin24 can we transition IDEs off of this pattern?

This would mean instead of `?page=inspector` `/inspector?query=etc`? Should be pretty easy for IntelliJ

DanTup:
> > can we transition IDEs off of this pattern?
>
> This would mean instead of ?page=inspector /inspector?query=etc? Should be pretty easy for IntelliJ

Yep, though we'll need to gate it on a version of DevTools that includes this PR (without the `serveStaticFile` change in this PR, the users would see 404s).

DanTup:
@kenzieschmoll bots are passing, but only because I added a `Future.microtask(() {` around the navigation. It seems not valid to navigate during `initState` (or `didChangeDependencies`) - which kinda makes sense, it's odd that we're "navigating" (even if we are doing `Router.neglect`) as we're building (we can't do it sooner because we don't know which tab will be the "default" one).

Do you know if there's a better way to do this?

packages/devtools_app/lib/src/scaffold.dart
```diff
@@ -274,14 +274,17 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
...
-          _currentScreen.screenId,
-          routerDelegate.currentConfiguration.args,
-        );
+    // without triggering a navigation. Since we can't nagivate during a build
+    // we have to wrap this in `Future.microtask`.
+    if (widget.page == null && _currentScreen is! SimpleScreen) {
+      Future.microtask(() {
+        final routerDelegate = DevToolsRouterDelegate.of(context);
+        Router.neglect(context, () {
+          routerDelegate.navigateIfNotCurrent(
```
kenzieschmoll, line 283: Instead of using Future.microtask(), what about WidgetsBinding.instance.addPostFrameCallback ?

another idea - is there a better way to update the url that doesn't trigger a build in the first place?

kenzieschmoll: I think it is forcing build to be called because `routerDelegate.navigateIfNotCurrent` eventually calls `notifyListeners`. One idea is to hide the `notifyListeners` call behind a bool `if (shouldNotify)`. `shouldNotify` could be an optional named parameter that defaults to true. I haven't tried this locally, so we'll need to make sure that this doesn't break any routing behavior if `notifyListeners` was false. This may cause more trouble than it is worth, but could be worth looking into.

DanTup: My understanding is that `notifyListeners()` is how we notify Flutter that the URL may have changed (and in turn, it may update the visible URL to the user). If we remove that, I think the visible URL in the browser won't change at all (since we've only updated our own local state variables).

(If this doesn't seem right, I can do some testing of it though)

kenzieschmoll: That makes sense. The postFrameCallback seems fine then.

packages/devtools_app/web/index.html
```diff
@@ -10,6 +10,7 @@
 <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <base href="/">
```
kenzieschmoll, line 13: @elliette wanted to get your input on changes to this file to verify that this shouldn't have any effects on serving DevTools in g3 or with DDR / the Dart DevTools extension.

kenzieschmoll (commented):
Thanks for these changes, Danny. Can you please mirror the devtools_server/ changes in this PR to the new home of devtools_server inside dds?

external_handlers.dart --> https://github.com/dart-lang/sdk/blob/main/pkg/dds/lib/src/devtools/handler.dart#L26

server.dart -->  https://github.com/dart-lang/sdk/blob/d1baf327a18dfc92476049a09985b645c08072f5/pkg/dds/lib/devtools_server.dart#L278

Please also leave the changes here for now, as we haven't completely migrated all code off of package:devtools_server and over to the dds implementation.

DanTup:
Thanks! I want to re-test with these final tweaks in VS Code, and @helin24 may want to do the same for IntelliJ. If both are good I'll open the same in DDS.

packages/devtools_app/web/index.html
elliette: This looks safe to me! None of the serving logic should care about the base URL.

DanTup:
This still works in VS Code. I'll make a start on the DDS PR tomorrow. @helin24 I don't anticipate the changes since you last tested (which I believe is just wrapping the router call in `addPostFrameCallback`) but if you want to test it before we merge, let me know.

helin24:
@DanTup we don't need to test again; thanks for checking

DanTup:
@kenzieschmoll 

> Thanks for these changes, Danny. Can you please mirror the devtools_server/ changes in this PR to the new home of devtools_server inside dds?

I see some code in here that seems to assume all paths will start with `/devtools` if `notFoundHandler` is not `null`. How is this currently used? I think in order to support this, the `index.html` would need to have `<base href='/devtools' />` so that Flutters routing would update the URLs to the correct thing (`/devtools/inspector`).

If we don't expect the `/devtools` in the frontend paths, we'd need to know which urls (eg. out of `/inspector` and `/foo`) should serve up the static `index.html` and which should fall through to `NotFoundHandler` (which I think means having knowing a list of valid page URLs here).

kenzieschmoll:
CC @bkonyi ☝️

kenzieschmoll:
> I see some code in here that seems to assume all paths will start with /devtools if notFoundHandler is not null.

Are you referring to this line: https://github.com/dart-lang/sdk/blob/main/pkg/dds/lib/src/devtools/handler.dart#L84? It looks like the notFoundHandler _will_ get hit if the path does not start with '/devtools'. So I think we would want the default not found handler to serve the static file. @bkonyi the optional `notFoundHandler` was a new addition to the server code when you moved it over from devtools_server to dds - what was the intended use for this? Also, what is the best way to test devtools end to end when serving from dds?

bkonyi:
`notFoundHandler` has been there since the standalone VM has been able to serve DevTools (`defaultHandler(...)` isn't the original handler from DevTools). When launched as part of starting DDS (not just serving via `package:dds/devtools_server.dart`), we need to be able to forward HTTP requests to the VM service to A) serve Observatory B) handle HTTP service requests in addition to serving DevTools. To avoid conflicts, DevTools assets are served under `$DDS_URI/devtools/` in this scenario.

I'm not sure what you mean by testing DevTools end to end when serving from DDS. We have some simple smoke tests that check that assets are being served correctly (see `devtools/test/devtools_observatory_connection_test.dart`), but that's about it.

kenzieschmoll:
> I'm not sure what you mean by testing DevTools end to end when serving from DDS.

I mean how can you build and run devtools locally with changes to devtools_app/ on a local branch and changes to the devtools server  code in dds/ on a local branch?

bkonyi:
> > I'm not sure what you mean by testing DevTools end to end when serving from DDS.
> 
> I mean how can you build and run devtools locally with changes to devtools_app/ on a local branch and changes to the devtools server code in dds/ on a local branch?

You could always pass a custom value for `customDevToolsPath` in `DevToolsServer.serveDevTools(...)` to point to the output of the DevTools build. The standalone VM only supports running the SDK packaged DevTools at this point, but we can always add a new option to allow for pointing to local versions (e.g., something similar to Flutter's `flutter run --local-engine-src=/a/b/c`).

DanTup:
> When launched as part of starting DDS (not just serving via package:dds/devtools_server.dart)
> ...
> To avoid conflicts, DevTools assets are served under $DDS_URI/devtools/ in this scenario.

To go via this path locally, would I just compile my local SDK, then run a Dart script with `--observe` (and not `--no-serve-devtools`) and then I can use the base URL (without `uri=`) as a general DevTools instance for testing?


I presume in that case we want the inspector page to look like `/devtools/inspector` in the URL and not just `/inspector`? If yes, then I think we just need a way to use an `index.html` with a different `base href`. If no and we want `/inspector` then I think the `defaultHandler` needs to know which urls belong to DevTools pages so it can serve up the static `index.html` for requests to them (likely from the user pressing Refresh in the browser or an IDE linking directly to one of those pages) before falling back to `notFoundHandler`?

bkonyi:
This is an example of a DevTools URI currently output by the standalone VM:

`http://127.0.0.1:8181/8JA6Uk5_ly4=/devtools/#/?uri=ws%3A%2F%2F127.0.0.1%3A8181%2F8JA6Uk5_ly4%3D%2Fws`

To just get to the DevTools connection screen, you'd just go to:

`http://127.0.0.1:8181/8JA6Uk5_ly4=/devtools/`

To communicate with the VM service or get Observatory assets, requests would just be made to:

`http://127.0.0.1:8181/8JA6Uk5_ly4=/`

DanTup:
Oh, that auto token might complicate things a bit more. Although I wonder if a relative `base href` would work (`<base href="." />`). I'll do some testing (otherwise, perhaps we could dynamically replace it as we serve the static index.html file).

DanTup:
Nope, relative base hrefs are not supported:

```
Error: The base href in "web/index.html" must be absolute (i.e. start with a "/"), but found: `<base href="./">`.
For example, to serve from the root use:
```

I think the only real option is probably to put a placeholder in the `index.html` file like:

```
<base href="$DEVTOOLS_BASE_HREF">
```

And in `serveStaticFile` do a replacement with either `/` or `/devtools/` depending on whether `notFoundHandler` is provided (always using `/` for the local non-DDS server script).

@kenzieschmoll @bkonyi WDYT?

kenzieschmoll:
We may want to take a step back and look at this again with fresh eyes now that https://github.com/flutter/devtools/pull/3617 has landed. Now that the devtools/ and devtools_server/ packages are deleted from DevTools, there should only be one place externally that we serve DevTools from: DDS. We still have a custom handler in g3 that won't expect the `/devtools` prefix. We could potentially look into changing that handler for simplicity if we need to.

Also, @DanTup I added a script in that PR build_e2e.dart and another script in pkg/dds that the build_e2e script calls. This will let you build a local devtools app with a local devtools server from pkg/dds source.

DanTup:
> We may want to take a step back and look at this again with fresh eyes now that https://github.com/flutter/devtools/pull/3617 has landed. Now that the devtools/ and devtools_server/ packages are deleted from DevTools, there should only be one place externally that we serve DevTools from: DDS. We still have a custom handler in g3 that won't expect the /devtools prefix. We could potentially look into changing that handler for simplicity if we need to.

Don't we still have the same difference between whether it's started with `dart devtools` (served at `/`) versus being started as part of ddds (served at `/devtools`)?

If so, I think we'd still need to change the `base href` to match as we serve the static file?

kenzieschmoll:
> Don't we still have the same difference between whether it's started with `dart devtools` (served at /) versus being started as part of ddds (served at /devtools)?

`dart devtools` does go through DDS, so whether it is served through the `dart devtools` command or calling into `serveDevToolsWithArgs` directly from the DDS package, I'd expect the same behavior.

DanTup:
> `dart devtools` does go through DDS, so whether it is served through the `dart devtools` command or calling into `serveDevToolsWithArgs` directly from the DDS package, I'd expect the same behavior.

I'm a little confused. When I run `dart devtools` using a build from today, it's served in the root at `/`, but if I run `dart --observe danny.dart` it's served from `/HWqt3IS3_Fc=/devtools/`.

I assume that this means that for the former, we need serve up `<base href="/" >` and for the latter we need to serve up `<base href="/HWqt3IS3_Fc=/devtools/" >` so that Flutter will build the correct URLs (and when the user presses Refresh on one of those URLs, we can serve up the correct index.html content).

Do you think there's a way we can do this without having to use a dynamic/injected base href?

kenzieschmoll:
Ah I see. I didn't realize the two (`dart devtools` vs `dart run --observe`) were different. @bkonyi is it possible to make `dart run --observe my_dart_script.dart` output a similar DevTools url to what `flutter run` outputs (e.g. `http://127.0.0.1:9100?uri=http://127.0.0.1:60555/1-u_9Qsc__0=/`)? [Here](https://github.com/flutter/flutter/blob/52ae102f182afaa0524d0d01d21b2d86d15a11dc/packages/flutter_tools/lib/src/resident_runner.dart#L1393) is where the DevTools url is built from flutter tools. I'm not sure I understand why the different implementation is necessary, since in both cases (standalone dart app vs a flutter app), the Observatory is started and the url for that is used to connect to DevTools.

So instead of 
```
$ dart run --observe test/fixtures/empty_dart_app.dart 
Observatory listening on http://127.0.0.1:8181/HwOin_ooG18=/
The Dart DevTools debugger and profiler is available at: http://127.0.0.1:8181/HwOin_ooG18=/devtools/#/?uri=ws%3A%2F%2F127.0.0.1%3A8181%2FHwOin_ooG18%3D%2Fws
```
we would output the following (note that `/HwOin_ooG18=/devtools/#/` has been removed from the devtools url)
```
$ dart run --observe test/fixtures/empty_dart_app.dart 
Observatory listening on http://127.0.0.1:8181/HwOin_ooG18=/
The Dart DevTools debugger and profiler is available at: http://127.0.0.1:8181?uri=ws%3A%2F%2F127.0.0.1%3A8181%2FHwOin_ooG18%3D%2Fws
```

bkonyi:
Instead of changing the DevTools URI in the standalone VM, I think we should standardize on `http://127.0.0.1:8181/devtools/...` across all of our tooling. I think serving Observatory on the same path as the VM service wasn't a great idea in the first place, so I'd prefer we keep the root path reserved for VM service communications.

bkonyi:
Also, since Observatory is still being served on the root path, determining which tool is served based solely on URI parameters seems confusing to me.

DanTup:
> http://127.0.0.1:8181/devtools/

Does there need to be an auth token in there?

I suspect so - and if so, I think we'd still need to inject the URL into the `base href` dynamically so it matches up.

bkonyi:
> > http://127.0.0.1:8181/devtools/
> 
> Does there need to be an auth token in there?
> 
> I suspect so - and if so, I think we'd still need to inject the URL into the `base href` dynamically so it matches up.

Yes, there should be. Sorry, I have a habit of passing `--disable-service-auth-codes` when I'm debugging service related stuff... :-)

DanTup:
@kenzieschmoll did you have any thoughts about this? It'll need rebasing and some changes moving into the DDS server, but I don't think we concluded how to handle `base href` so wen ca host the app in a sub-folder. My understanding is that if the URL is going to be `/someAuthToken==/devtools/` that we need to write that into the `base href` so Flutter knows that navigations (eg. to `/inspector`) should leave that first part of the URL alone. Since the auto token is not fixed, we can't write that directly into the `index.html` file, and would need to inject it while serving up the `index.html` page.

kenzieschmoll:
Just to clarify, is the following result what is being suggested?

Output from `dart run --observe test/fixtures/empty_dart_app.dart `:
```
The Dart VM service is listening on http://127.0.0.1:8181/Bpxmk_BkX-Q=/
The Dart DevTools debugger and profiler is available at: http://127.0.0.1:8181/Bpxmk_BkX-Q=/devtools/#/?uri=ws%3A%2F%2F127.0.0.1%3A8181%2FBpxmk_BkX-Q%3D%2Fws
```
Output from `dart devtools`:
```
Serving DevTools at http://127.0.0.1:9100/devtools
```
Output from `flutter run` (which in turn, calls `dart devtools`)
```
An Observatory debugger and profiler on iPhone 13 is available at: http://127.0.0.1:57031/w5becV99ILI=/
The Flutter DevTools debugger and profiler on iPhone 13 is available at:
http://127.0.0.1:9100/devtools?uri=http://127.0.0.1:57031/w5becV99ILI=/
```
and when you change a page in DevTools:
`http://127.0.0.1:9100/devtools/#/performance?uri=http://127.0.0.1:57031/w5becV99ILI=/`

Note that for the `dart devtools` and `flutter run` outputs, DevTools and the vm service are still served on different ports (as is the current implementation), so the `/devtools` addition is just for consistency.

DanTup:
Sorry for the confusion - there are two related - but different - issues. However I think one of them (whether the path starts with `/devtools` always) is superseded by the second one, so I'll just explain the second one.

In a Flutter web app, the `<base href="" />` HTML tag needs to be set to match the folder where the app is being hosted in order to use the "Path" URL strategy ([docs](https://docs.flutter.dev/development/ui/navigation/url-strategies#hosting-a-flutter-app-at-a-non-root-location)). If it's not set correctly - for example it is set to `/` but the app is really at `/devtools` then when the app tries to push a route like `/inspector`, the address bar will be updated to `/inspector` instead of `/devtools/inspector`. Then if the user presses Refresh in their browser, a request will go to the server outside of the DevTools app (here, it would result in a 404 via the original Observatory handler).

Part of the discussion above about using `/devtools/` in all cases was the hope that we could just hard-code that into `index.html` so it didn't need to be dynamic.

However, this won't work because of auth tokens. DevTools is sometimes served up at `/{authToken}/devtools` and that means we need to have `/{authToken}/devtools` in the `<base href="" />` HTML tag. Since the auth token changes, I don't see any way to handle this other than rewriting it _as_ we serve the `index.html` file up. If we're writing it dynamically, then it might not matter whether we always have `/devtools`, as long as the server knows what the base URL is.

This shouldn't be difficult, but I think we need to be happy with dynamically injecting/rewriting the `base href` in the handler that serves up the `index.html` page..


Slightly related, we need to ensure that the server responses to requests for `/inspector` with that same `index.html` content, so the Flutter app is loaded. That means either having a list of the URLs in the server handler, or being able to assume that any request into whatever the root for DevTools is that is not matched by an existing handler can serve up the `index.html` (and the DevTools Flutter app will handle the 404s client-side with its routing).

kenzieschmoll:
> This shouldn't be difficult, but I think we need to be happy with dynamically injecting/rewriting the base href in the handler that serves up the index.html page..

Okay, and it sounds like we would only have to dynamically rewrite the base href in the case of `dart run --observe`, since that is the only time DevTools is served with an auth token in the url path? CC @elliette.

Another idea for @bkonyi - why can't we serve DevTools on a different port when running `dart run --observe`? Just like we do for running `dart devtools`? If we serve DevTools on its own port, we wouldn't have to worry about dynamically rewriting the base href in DevTools because it would always be served at "/".

bkonyi:
> Another idea for @bkonyi - why can't we serve DevTools on a different port when running `dart run --observe`? Just like we do for running `dart devtools`? If we serve DevTools on its own port, we wouldn't have to worry about dynamically rewriting the base href in DevTools because it would always be served at "/".

If we're planning on serving DevTools as a replacement for Observatory, I'd strongly prefer that we use the same port as DDS. Adding more ports creates more opportunities for port conflicts and requires remote users to forward yet another port.

Another option is to just not serve DevTools on a path behind the authentication code. We'll still want to keep it behind `/devtools` though.

kenzieschmoll:
> Another option is to just not serve DevTools on a path behind the authentication code. We'll still want to keep it behind `/devtools` though.

@DanTup would this be a better solution since we wouldn't have to worry about dynamically rewriting `/` to `/{authToken}/devtools`, but rather `/` to `/devtools`?

DanTup:
> > Another option is to just not serve DevTools on a path behind the authentication code. We'll still want to keep it behind /devtools though.
> 
> @DanTup would this be a better solution since we wouldn't have to worry about dynamically rewriting `/` to `/{authToken}/devtools`, but rather `/` to `/devtools`?

If we can skip the auth token and _always_ host at `/devtools/`, we could just hardcode that into the `base href` and not do any rewriting at all.

If we still host at `/` for `dart devtools` then we'd still need to rewrite it (I don't think the difference between rewriting it with/without the token makes much difference, since we'll have the auth token in the incoming request URL anyway).

DanTup:
@kenzieschmoll @bkonyi what's the simplest way for me to run devtools locally in a way that goes through the path that has `/devtools` as the root? Do I need to build the SDK/VM and run through it, or is there an easier way?

I'm using the `build_e2e.dart` script for the other way, but I need to check my changes aren't affecting anything outside of `/devtools` when hosted the other way.

bkonyi:
> @kenzieschmoll @bkonyi what's the simplest way for me to run devtools locally in a way that goes through the path that has `/devtools` as the root? Do I need to build the SDK/VM and run through it, or is there an easier way?
> 
> I'm using the `build_e2e.dart` script for the other way, but I need to check my changes aren't affecting anything outside of `/devtools` when hosted the other way.

Other than running a Dart process that spawns it, I don't think there's a way to do that. Maybe we should be adding the `/devtools` path for `dart devtools` as well...

DanTup:
So, I thought it would reduce complexity - and it does - however it also adds some different complexity :-)

The IDEs only get a host/port and not a full URL, so they will (currently) try to go to `/`. So we would additionally need to have a redirect from `/` -> `/devtools`. That sounds simple, but because we're currently using fragments the IDEs might embed `http://foo:123/#/inspector` and I think the fragment will be lost in a server-side redirect.

So, we may end up having to serve a `/index.html` that uses javascript to redirect to `/devtools/` and retain the fragment to stay compatible with the IDEs as they are today.

I suspect that's less complex than messing with the `base href`, but perhaps only slightly - so if that influences your preference, let me know!

If we do go this route, I'll still need a way to run locally to test that my `/` -> `/devtools` redirect is only active in the local case and not when being hosted through the VM.

@bkonyi 

> Other than running a Dart process that spawns it, I don't think there's a way to do that

I presume this means "build the VM and run a script through the newly compiled VM`? If so, how do I ensure it's using my local copy of the DevTools front-end? (I presume I can just paste it somewhere in the SDK folder? but I can't find anything that looks like a copy of it to overwrite).

bkonyi:
> @bkonyi
> 
> > Other than running a Dart process that spawns it, I don't think there's a way to do that
> 
> I presume this means "build the VM and run a script through the newly compiled VM`? If so, how do I ensure it's using my local copy of the DevTools front-end? (I presume I can just paste it somewhere in the SDK folder? but I can't find anything that looks like a copy of it to overwrite).

If you locally compile DevTools, you can copy the binary output to `$DART_SDK/bin/resources/devtools/`. We just serve the DevTools binaries stored there.

DanTup:
@kenzieschmoll @bkonyi I've not been able to track down why this change fails with old-style (fragment) URLs. It seems that at some point during initialisation (after switching to UrlPathStrategy, although exactly when seemed inconsistent even if I added lots of delays) the fragment was being dropped.

As a fix, I decided that the best thing would be to detect legacy URLs right at the start, and redirect (with `location.replace`) to the new-style equivalent. This also has the benefit that no part of the DevTools code will need to handle legacy URLs anymore, it's all handled just be one small mapping before initialisation.

I've pushed that change to this PR so they'll need reviewing.

Since it's been a while since this was started, a summary:

## DDS Change

CL: https://dart-review.googlesource.com/c/sdk/+/239082/

- Changes the server to rewrite the `base href` _if present in the page_ so support the UrlPathStrategy version of DevTools
- Serves up the contents of `index.html` for requests to URLs like `/inspector` so linking directly to those pages (or hitting refresh) will not result in a 404 from the server

I believe these changes are safe to land before the DevTools changes (the opposite is not true).

### DevTools Change

This PR

- Changes the app to use PathUrlStrategy to remove the `#` fragments in the URLs
- Adds a base href to support the above
- Redirects old legacy URLs to new URLs before initialising the app

These changes _require_ the serve has the changes above to function correctly.

### Testing

To test, I did:

- build release version of DevTools
- build SDK (`./tools/build.py --no-goma --mode release --arch x64 create_platform_sdk`)
- Overwrite DevTools in `third_party/devtools/web` and `sdk/xcodebuild/ReleaseX64/dart-sdk/bin/resources/devtools` with the newly built DevTools

Things I've tested:

- running `xcodebuild/ReleaseX64/dart-sdk/bin/dart devtools` serves a version using Url paths
- running `xcodebuild/ReleaseX64/dart-sdk/bin/dart --observe=0 /tmp/foo.dart` prints a valid URL that loads DevTools (it redirects to remove the fragment, and then functions as expected)
- running VS Code using `xcodebuild/ReleaseX64/dart-sdk` as the active SDK spawns a working embedded devtools
- running VS Code using `xcodebuild/ReleaseX64/dart-sdk` as the active SDK spawns working external devtools pages, and can navigate between pages using the server API (by running commands in VS Code to launch those other pages)

Things I'm not able to test:

- IntelliJ/Android Studio
- Google3

### Future Changes

After both changes are landed, further changes to be made:

- Unskip the two DDS tests that are currently skipped because they test the served page has a `base href`
- Remove the `#` from the URL printed by `dart --observe` as it's no longer necessary (although the redirect handles it so it doesn't cause any problems)

DanTup:
```
   info - test/url_utils_test.dart:36:69 - Missing a required trailing comma. - require_trailing_commas
   info - test/url_utils_test.dart:44:75 - Missing a required trailing comma. - require_trailing_commas
```

Pushed a fix for these, though for some reason I don't see these lints locally (nor can I see where `require_trailing_commas` is being enabled from), so if anyone has any ideas about that, I'm interested to know :-)

kenzieschmoll:
> for some reason I don't see these lints locally

This lint was enabled in a recent PR. If you merge master you should see the lint warnings (you may have to restart the analysis server or your IDE if the change is not picked up automatically).

DanTup:
> This lint was enabled in a recent PR. If you merge master you should see the lint warnings (you may have to restart the analysis server or your IDE if the change is not picked up automatically).

Ah, that's what it was. For some reason I thought GH Actions always ran the PR as it was on the source branch, but I can see from the logs that it merges master in first (which is better, I just didn't realise it did). Thanks!

packages/devtools_app/lib/src/url_strategy/web.dart
```diff
@@ -0,0 +1,5 @@
+import 'package:flutter_web_plugins/flutter_web_plugins.dart';
```
kenzieschmoll, line 1: move the url_strategy folder under src/config_specific

DanTup: Done!

kenzieschmoll:
Tested changes in google3. Google3 does not have a context where devtools is served under `/devtools`, so everything works as expected with a base href of `/`

kenzieschmoll (approved):

DanTup:
@kenzieschmoll I merged master into this (there were some conflicts because url_utils was removed, but I'd added to it here, so this brings it back).

Do you need to do anything in g3 before landing this (the DDS change has landed)? I see https://github.com/flutter/devtools/pull/3585#discussion_r792196617 but suspect that needs doing as it's rolled into g3?

kenzieschmoll:
> Do you need to do anything in g3 before landing this (the DDS change has landed)?

Nope, I have a g3 CL ready to go that applies your changes to the g3 handler. You are free to land this.

packages/devtools_app/lib/src/primitives/url_utils.dart
```diff
@@ -0,0 +1,84 @@
...
+
+// TODO(kenz): try to eliminate this method once
+// https://github.com/dart-lang/sdk/issues/46872 is fixed.
+/// Returns a simplified version of a package url, replacing "path/to/flutter"
+/// with "package:".
+///
+/// This is a bit of a hack, as we have file paths instead of the explicit
+/// "package:uris" we'd like to have. This will be problematic for a use case
+/// such as "packages/my_package/src/utils/packages/flutter/".
+String getSimplePackageUrl(String url) {
```
kenzieschmoll, line 13: can you remove this method and associated tests? (essentially removing everything that was removed as part of https://github.com/flutter/devtools/pull/3932/files#diff-0fb55973e4befafb1d8f5ee26a619a6500f6524ef0dfcf2c252f62aa0644076f)

DanTup: Oops, good catch. I had to abort/re-do the merge and apparently didn't re-remove it the second time 😞

DanTup:
@kenzieschmoll I did another final test and this all looked good, so I've landed it.

Anyone running the `build_e2e.dart` script will need to ensure they're on latest SDK to get the new server version for everything to work correctly.

Please ping me (and/or revert) if you think this has caused any serious issue. If you let me know once this version of the frontend has landed in the SDK, I'll un-skip those tests that rely on the `base href` (and tests again using the landed version of everything). Thanks!

PR #3586, opened 2022-01-24
Manually set R/W permissions on canvaskit binaries

bkonyi:


tool/build_release.sh
```diff
@@ -41,7 +41,6 @@ flutter build web \
   --pwa-strategy=none \
   --profile \
   --dart-define=FLUTTER_WEB_USE_SKIA=true \
-  --dart-define=FLUTTER_WEB_CANVASKIT_URL=canvaskit/ \
```
kenzieschmoll, line 44: if this is removed, how will the web app know where to pull canvas kit from?

bkonyi: AFAICT, if set, this overrides where canvaskit is downloaded/copied from (see [here](https://github.com/flutter/engine/blob/2962099077b36704071802ea5595f4a016a1c214/lib/web_ui/lib/src/engine/configuration.dart#L96)). I don't think we were using this value properly in the first place.

kenzieschmoll: hmm - It looks like the default is from unpkg.com: https://github.com/flutter/engine/blob/2962099077b36704071802ea5595f4a016a1c214/lib/web_ui/lib/src/engine/configuration.dart#L113

We do not want to be downloading this at runtime, which is why we've specified an already downloaded version.

bkonyi: It seems that this actually downloads at compile time (`canvaskit/` is present in the `build/web/` when running `build_release.sh`).

kenzieschmoll: Okay - have you verified that building the release web build with the FLUTTER_WEB_CANVASKIT_URL removed and running it does not make any canvaskit related network requests over the wire? If so, lgtm.

bkonyi: That's weird... we pull it down during the build and then proceed not to use it... I've added back the define and things seem to work properly again.

kenzieschmoll (approved):

Issue #3593, opened 2022-01-27
Warn users that Flutter Web debugging support is not available for internal users


kenzieschmoll:
We can land this externally and hide it behind the `isExternalBuild` bool as well as connected app type = flutter web. 

We might want to only show this warning for the debugger (unless other pages like the inspector are also affected).

elliette:
I know both the inspector and debugger are affected, I'm not sure about other pages.

PR #3597, opened 2022-01-27
Show warning for internal Flutter Web apps

elliette:
Shows a warning that Flutter DevTools is not supported for internal Flutter Web apps, and includes a link to the bug. 

The warning is shown when a user first opens Flutter DevTools:

Screen Shot 2022-01-27 at 11 47 21 AM

Resolves https://github.com/flutter/devtools/issues/3593

### DIFF
--- a/packages/devtools_app/lib/src/shared/common_widgets.dart
+++ b/packages/devtools_app/lib/src/shared/common_widgets.dart
@@ -9,7 +9,6 @@ import 'dart:math';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
-import 'package:url_launcher/url_launcher.dart';
 
 import '../analytics/analytics.dart' as ga;
 import '../config_specific/launch_url/launch_url.dart';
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1775,6 +1776,38 @@ class PubWarningText extends StatelessWidget {
...
+          TextSpan(
+              text:
+                  'Warning: Flutter DevTools is not currently supported for Flutter Web apps.\n\n',
+              style: theme.subtleTextStyle
+                  .copyWith(color: theme.colorScheme.errorTextColor)),
+          TextSpan(
+              text: 'Some debugging features might not work as expected.\n',
+              style: theme.subtleTextStyle),
+          TextSpan(text: 'See ', style: theme.subtleTextStyle),
+          TextSpan(
```
kenzieschmoll, line 1797: Use LinkTextSpan from common_widgets.dart. This widget uses logic from launch_url.dart, which has conditional import logic so that we don't try to use url_launcher on Desktop, which will result in a crash (or at least it used to)

kenzieschmoll (approved):
LGTM with one comment

packages/devtools_app/lib/src/shared/common_widgets.dart
elliette: Done, thanks!

PR #3600, opened 2022-01-28
Update ansicolor to nullsafe version

polina-c:
Created issue to fix @TestOn('browser') :https://github.com/flutter/devtools/issues/3616


polina-c:
Created https://dart-review.googlesource.com/c/sdk/+/230882 to fix one of the bugs

packages/devtools_app/test/debugger_screen_test.dart
```diff
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+@TestOn('browser')
```
kenzieschmoll, line 5: instead of adding this to all the tests that test something related to ansicolor outputs, can we try pulling only the affected tests into ansi_up_test.dart? you can make a group for each set of tests (e.g. 'Debugger ansi tests', 'Logging ansi tests'). This would also eliminate the need below to make logging_screen_browser_test and logging_screen_vm_test

polina-c: done

packages/devtools_app/test/ansi_up_test.dart
```diff
@@ -88,4 +100,197 @@ void main() {
...
+    final fakeLogData = List<LogData>.generate(totalLogs, _generate);
+
+    Future<void> pumpLoggingScreen(WidgetTester tester) async {
+      await tester.pumpWidget(wrapWithControllers(
+        const LoggingScreenBody(),
+        logging: mockLoggingController,
+      ));
+    }
+
+    setUp(() async {
```
kenzieschmoll, line 160: I expect a lot of these set up steps are not necessary for the single test that is in this group. Let's remove any that aren't needed. Same comment for the Debugger screen group below.

polina-c: For now I do not have a way to run the tests and verify they pass. Both locally or with bots.
Added TODO for this.

kenzieschmoll (approved):

Issue #3555, opened 2021-12-23
File opener input should be overlaid above the program explorer


elliette:
For small window sizes, when both the file opener and the program explorer are opened, the program explorer covers the file opener search field:
<img width="718" alt="Screen Shot 2021-12-23 at 9 51 29 AM" src="https://user-images.githubusercontent.com/21270878/147277249-b53717aa-064b-4b4b-b504-56140402de55.png">

The search field modal should be stacked above the program explorer.

elliette:
Closing in favor of https://github.com/flutter/devtools/issues/3604

PR #3602, opened 2022-01-28
Focus workaround so that keyboard shortcuts always work

elliette:
Continued work on https://github.com/flutter/devtools/issues/3457

This change is a workaround to avoid the behavior described in https://github.com/flutter/flutter/issues/97389. It adds a `GestureDetector` that detects click events anywhere on the screen (that isn't already a focusable widget), and moves the focus to the keyboard shortcuts instead.

With this change, a user can do things like resize the console and still use keyboard shortcuts afterwards (this wasn't working before):
Jan-27-2022 17-50-26

### DIFF
--- a/packages/devtools_app/lib/src/shared/scaffold.dart
+++ b/packages/devtools_app/lib/src/shared/scaffold.dart
@@ -546,7 +546,7 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
   }
 }
 
-class KeyboardShortcuts extends StatelessWidget {
+class KeyboardShortcuts extends StatefulWidget {
   const KeyboardShortcuts({
     @required this.keyboardShortcuts,
     @required this.child,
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/shared/scaffold.dart
```diff
@@ -556,16 +556,36 @@ class KeyboardShortcuts extends StatelessWidget {
...
     }
-    return Shortcuts(
-      shortcuts: keyboardShortcuts.shortcuts,
-      child: Actions(
-        actions: keyboardShortcuts.actions,
-        child: child,
+
+    return GestureDetector(
+      behavior: HitTestBehavior.opaque,
+      onTap: () => FocusScope.of(context).requestFocus(_focusNode),
```
kenzieschmoll, line 582: This is a little concerning since KeyboardShortcuts is a top level wrapper widget around the scaffold. If a FocusNode further down the tree requests focus, will those requests take precedence over this one? I would think so because this GestureDetector would be "behind" or further up the tree than other gesture detectors down the tree, but just want to confirm.

kenzieschmoll:
I checked this code out and played with it a bit. I don't think these issues are new or related to this PR, but linking it here just in case: https://github.com/flutter/devtools/issues/3604

packages/devtools_app/lib/src/shared/scaffold.dart
elliette: Yeah, I've added some logging in `FocusManager` to check, it seems like everything that already had a `FocusNode` is still getting focus as we would want, and this new focus node only gets focus when something that doesn't have a focus node in scope is tapped.

elliette:
Thanks for opening https://github.com/flutter/devtools/issues/3604! That was already in issue, I've closed https://github.com/flutter/devtools/issues/3555 in favor of your new issue.

kenzieschmoll (approved):

busslina:
Here the same issue. I try to always capture focus with a hierarchical structure and when I click outside a textfield focus is gained by a focus node with "_ModalScopeState Focus Scope" as debug label.

My first doubt is which element introduces this modal scope. Is it Material App? Is it Scaffold?
My second doubt is how to disable this Focus Scope?

busslina:
It is Material App that adds this Focus Scope. Adding a Focus widget above Material App allows you to not lost focus (on that root node) when clicked outside TextField and _ModalScopeState FocusScope gains the focus. Still not the desired behavior.
Make no sense to have controlled the focus with a hierarchical tree structure which leaf is the textfield (or any other widget) and transfer the focus to Material App (_ModalScopeState) when clicked outside textfield (breaking the focus node tree below Material App).

I found some tricky ways to counter that behavior, but I want to resolve this in a acceptable way. Seems that I will have to dig into MaterialApp code.

Any suggestion will be rewarded.

PR #3603, opened 2022-01-28
Keep mapping of thread ids to thread names up to date

### LINKED ISSUE #1881: Unknown thread shows up in timeline
Run a Flutter app in an ios simulator and you will see an empty "Unknown"
isolate at the bottom of the list of isolates.
We should consider hiding isolates without any profiling data or understand what this isolate is and why we don't have data for it.
Screen Shot 2020-04-27 at 10 13 46 AM

blaugold:
This change updates the map of known threads each time new
timeline data is loaded.

The timeline lanes of threads that did not exist when the first batch of
timeline data was loaded were previously labeled `Unknown`. Now
they have the same labeling as other threads.

Fixes #1881

### DIFF
--- a/packages/devtools_app/lib/src/performance/performance_controller.dart
+++ b/packages/devtools_app/lib/src/performance/performance_controller.dart
@@ -576,6 +576,11 @@ class PerformanceController extends DisposableController
     List<TraceEventWrapper> traceEvents, {
     bool isInitialUpdate = false,
   }) {
+    final isFlutterApp = offlineController.offlineMode.value
+        ? offlinePerformanceData != null &&
+            offlinePerformanceData.frames.isNotEmpty
+        : serviceManager.connectedApp.isFlutterAppNow;
+
     final threadNameEvents = traceEvents
         .map((TraceEventWrapper wrapper) => wrapper.event)
         .where((TraceEvent event) {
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/performance/performance_controller.dart
```diff
@@ -583,45 +586,49 @@ class PerformanceController extends DisposableController
     for (TraceEvent event in threadNameEvents) {
       final name = event.args['name'];
 
-      // Android: "1.ui (12652)"
-      // iOS: "io.flutter.1.ui (12652)"
-      // MacOS, Linux, Windows, Dream (g3): "io.flutter.ui (225695)"
-      if (name.contains('.ui')) {
-        uiThreadId = event.threadId;
-      }
+      if (isInitialUpdate) {
```
kenzieschmoll, line 589: we should also check if we are connected to a flutter app (either offline or online). See an example of the below check here: https://github.com/flutter/devtools/blob/a19d96fc321d047f5b96d3b1d7d5a8a32cdf3eb6/packages/devtools_app/lib/src/performance/performance_screen.dart#L151-L153

```
final isOfflineFlutterApp = offlineController.offlineMode.value &&
       offlinePerformanceData != null &&
        offlinePerformanceData.frames.isNotEmpty;

if (isInitialUpdate && (isOfflineFlutterApp ||
            (!offlineController.offlineMode.value &&
                serviceManager.connectedApp.isFlutterAppNow))
```

blaugold: Done

kenzieschmoll (approved):

blaugold:
Thanks for reviewing this PR.

PR #3611, opened 2022-01-31
Assorted cleanup for `_AutoCompleteSearchField`

elliette:
Various cleanup for the `_AutoCompleteSearchField` widget. This should not change any behavior, it is simply a refactor:

* Switches it from a `StatelessWidget` to a `StatefulWidget`. This is because there was a focus listener in the `build` method. From what I understand, widgets with listeners should be `StatefulWidget`s. We now attach the listener in `initState`
* deletes the unused `closeHandler` callback 
* deletes a bunch of unused key definitions in `SearchFieldMixin`
* Instead of inlining callback functions, moves their body into two methods ( `_closeAutoCompleteOnFocusLost` and `_handleKeyStrokes`) to make them easier to read 
* Deletes the `enterMac` and `tabMac` key definitions. The Flutter desktop bug that they were added to work around no longer seems to be occurring.

### DIFF
--- a/packages/devtools_app/lib/src/ui/search.dart
+++ b/packages/devtools_app/lib/src/ui/search.dart
@@ -951,7 +951,8 @@ class _AutoCompleteSearchField extends StatefulWidget {
       _AutoCompleteSearchFieldState();
 }
 
-class _AutoCompleteSearchFieldState extends State<_AutoCompleteSearchField> {
+class _AutoCompleteSearchFieldState extends State<_AutoCompleteSearchField>
+    with AutoDisposeMixin {
   /// Platform independent (Mac or Linux).
   int get arrowDown =>
       LogicalKeyboardKey.arrowDown.keyId & LogicalKeyboardKey.valueMask;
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -962,111 +945,121 @@ class _AutoCompleteSearchField extends StatelessWidget {
...
+  HighlightAutoComplete get _highlightDropdown =>
+      widget.onHighlightDropdown != null
+          ? widget.onHighlightDropdown
+          : _highlightDropdownDefault;
+
+  @override
+  void initState() {
+    super.initState();
+
+    widget.searchFieldFocusNode.addListener(_closeAutoCompleteOnFocusLost);
```
kenzieschmoll, line 979: use addAutoDisposeListener for this

kenzieschmoll (approved):
lgtm with one request

packages/devtools_app/lib/src/ui/search.dart
elliette: Good catch, done!

PR #3621, opened 2022-02-03
Migrate auto_dispose to null safety.

polina-c:


packages/devtools_app/lib/src/primitives/auto_dispose.dart
```diff
@@ -25,19 +23,19 @@ class Disposer {
...
+  void autoDisposeFocusNode(FocusNode? node) {
     if (node == null) return;
     _focusNodes.add(node);
   }
 
   /// Add a listener to a Listenable object that is automatically removed when
   /// cancel is called.
-  void addAutoDisposeListener(Listenable listenable, [VoidCallback listener]) {
+  void addAutoDisposeListener(Listenable? listenable,
+      [VoidCallback? listener]) {
```
kenzieschmoll, line 38: nit: add a trailing comma after the last parameter. General rule of thumb: if a method signature or method call has to break to a second line, always use the trailing comma. For Flutter code, we always use a trailing comma (with the exception of a few widgets like SizedBox or a Text widget to name a couple).

here and elsewhere in this PR

packages/devtools_app/lib/src/primitives/auto_dispose_mixin.dart
```diff
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-// @dart=2.9
+
```
kenzieschmoll, line 5: you can remove the extra new lines here. One blank line between copy right and imports is standard.

kenzieschmoll (approved):

PR #3624, opened 2022-02-03
Null safety for some primitives (#3).

polina-c:


packages/devtools_app/lib/src/primitives/syntax_highlighting.dart
```diff
@@ -33,34 +31,34 @@ class TextmateGrammar {
   final List<Rule> _fileRules = [];
   final Map<String, Rule> _ruleMap = {};
 
-  Map _definition;
+  Map? _definition;
```
kenzieschmoll, line 34: it looks like the can be late final

kenzieschmoll (approved):

PR #3626, opened 2022-02-04
Make trees.dart null safe.

polina-c:


packages/devtools_app/lib/src/primitives/trees.dart
```diff
@@ -48,26 +46,26 @@ abstract class TreeNode<T extends TreeNode<T>> {
   bool get isSelected => _selected;
   bool _selected = false;
 
-  T get root {
+  T? get root {
```
kenzieschmoll, line 49: this should not be nullable. If a node does not have a parent, we return `this` as the root (e.g. `this.root => this`), and `this.root` would not be able to be accessed if `this` was null.

packages/devtools_app/lib/src/primitives/trees.dart
```diff
@@ -130,13 +128,13 @@ abstract class TreeNode<T extends TreeNode<T>> {
     return children.removeAt(index);
   }
 
-  void addAllChildren(List<T> children) {
-    children.forEach(addChild);
+  void addAllChildren(List<T?> children) {
+    children.forEach(addChild as void Function(T?));
```
kenzieschmoll, line 132: can we drop the `?` marks here? I don't know why they'd be necessary since `addChild` takes a non nullable `T child`

kenzieschmoll (approved):

polina-c:
tested the functionality

PR #3632, opened 2022-02-04
Migrate some libraries to be null safe.

polina-c:


polina-c:
tested

packages/devtools_app/lib/src/shared/version.dart
```diff
@@ -2,21 +2,21 @@
...
-    @required this.version,
-    @required this.channel,
-    @required this.repositoryUrl,
-    @required this.frameworkRevision,
-    @required this.frameworkCommitDate,
-    @required this.engineRevision,
-    @required this.dartSdkVersion,
+    required this.version,
+    required this.channel,
+    this.repositoryUrl,
```
kenzieschmoll, line 15: why is this no longer required?

packages/devtools_app/lib/src/shared/version.dart
```diff
@@ -42,15 +42,15 @@ class FlutterVersion extends SemanticVersion {
 
   final String channel;
 
-  final String repositoryUrl;
+  final String? repositoryUrl;
```
kenzieschmoll, line 45: why is repositoryUrl the only member that is nullable here? I would expect that they are either all nullable or none since we are just parsing json?

packages/devtools_app/lib/src/shared/version.dart
polina-c: We check it for null later in the code, so I assume it sometimes can be null

packages/devtools_app/lib/src/shared/framework_controller.dart
```diff
@@ -2,12 +2,10 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-// @dart=2.9
+// ignore_for_file: import_of_legacy_library_into_null_safe
```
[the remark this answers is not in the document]
polina-c: discussed offline

packages/devtools_app/lib/src/shared/version.dart
polina-c: fixed

kenzieschmoll (approved):

packages/devtools_app/lib/src/shared/version.dart
polina-c: fixed

PR #3633, opened 2022-02-05
Migrate theme.dart to be null safe.

polina-c:


packages/devtools_app/lib/src/debugger/program_explorer_controller.dart
```diff
@@ -98,6 +98,7 @@ class ProgramExplorerController extends DisposableController
       return;
     }
     _selectScriptNode(script, _rootObjectNodes.value);
+    // ignore: invalid_use_of_visible_for_testing_member, invalid_use_of_protected_member
```
kenzieschmoll, line 101: why are these ignores necessary? These should be calling ListValueNotifier.notifyListeners(), which isn't marked visible for testing nor protected.

packages/devtools_app/lib/src/shared/theme.dart
```diff
@@ -554,7 +554,7 @@ double get chartFontSizeSmall => scaleByFontFactor(12.0);
 const lightSelection = Color(0xFFD4D7DA);
 
 bool includeText(
-    BuildContext context, double minScreenWidthForTextBeforeScaling) {
+    BuildContext context, double? minScreenWidthForTextBeforeScaling) {
```
kenzieschmoll, line 557: nit: not your fault but add a trailing comma here

packages/devtools_app/test/analytics_prompt_test.dart
```diff
@@ -32,6 +33,7 @@ void main() {
     setUp(() {
       didCallEnableAnalytics = false;
       setGlobal(ServiceConnectionManager, FakeServiceManager());
+      setGlobal(IdeTheme, IdeTheme());
```
kenzieschmoll, line 36: do all these test fail if the IdeTheme global is null?

polina-c: yes, they fail

packages/devtools_app/lib/src/shared/theme.dart
polina-c: done

packages/devtools_app/lib/src/performance/rebuild_counts.dart
```diff
@@ -100,6 +100,7 @@ class RebuildCountModel {
     }
 
     // We've updated the build counts and possibly the locations.
+    // ignore: invalid_use_of_visible_for_testing_member, invalid_use_of_protected_member
```
kenzieschmoll, line 103: same comment applies to the ignores in this file as well

packages/devtools_app/lib/src/debugger/program_explorer_controller.dart
polina-c: interesting, now they are not needed
may be something changed in master?
removed

packages/devtools_app/lib/src/performance/rebuild_counts.dart
```diff
@@ -116,6 +116,7 @@ class RebuildCountModel {
     }
 
     // We've updated the build counts.
+    // ignore: invalid_use_of_visible_for_testing_member, invalid_use_of_protected_member
```
kenzieschmoll, line 119: this ignore can be removed as well

packages/devtools_app/test/analytics_prompt_test.dart
kenzieschmoll: Would it be better to make sure that we fail gracefully at runtime when IdeTheme is null? If the field is nullable in general, we are saying it is okay to have a null ide theme at runtime, so we want to make sure that our UI gracefully handles the null value and does not throw exceptions.

polina-c: Instead of investing in graceful error handling, I would invest in making sure getIdeTheme(), invoked in main.dart, never returns null.

polina-c: updated ide_theme_stub.dart

packages/devtools_app/lib/src/shared/theme.dart
```diff
@@ -18,9 +18,9 @@ const contrastForegroundWhite = _contrastForegroundWhite;
 /// Constructs the light or dark theme for the app taking into account
 /// IDE-supplied theming.
 ThemeData themeFor({
-  @required bool isDarkTheme,
-  @required IdeTheme ideTheme,
-  ThemeData theme,
+  required bool isDarkTheme,
+  required IdeTheme? ideTheme,
```
kenzieschmoll, line 22: now IdeTheme should be able to be non-nullable everywhere

polina-c: done

kenzieschmoll (approved):

packages/devtools_app/lib/src/shared/theme.dart
polina-c: some tests are fixed by `setGlobal(IdeTheme, IdeTheme());`, but some are not
if test pumps widgets like this `await tester.pumpWidget(wrap(aboutDialog));`

themeFor gets null for ideTheme. it seems globals are not visible in this case.

reverted back to nullable

polina-c: fixed by updating 'wrap'

kenzieschmoll (approved):

packages/devtools_app/test/theme_test.dart
```diff
@@ -12,17 +12,6 @@ import 'package:flutter_test/flutter_test.dart';
 void main() {
   group('Theme', () {
     ThemeData theme;
-    test('can be used without override', () {
-      theme = themeFor(isDarkTheme: true, ideTheme: null);
```
kenzieschmoll, line 16: instead of removing this test, why don't we just replace `null` with `IdeTheme()`?

polina-c: The goal of the test is to verify everything works with null. We do not need this validation any more.
Other tests here cover non-null cases. Makes sense?

kenzieschmoll: In this case an empty IdeTheme would be the equivalent of what used to be null, in terms of testing the "without override" case. The "without override" case is where a user runs DevTools via `dart devtools` instead of using the DevTools embedded in their IDE. So we still want to test the case where IdeTheme does not contain overrides for backgroundColor, foregroundColor, etc.

polina-c: fixed

PR #3639, opened 2022-02-08
Split utils to simplify migration to null safety.

polina-c:
An attempt to migrate utils as is to null safety resulted in 340 failing tests.
Splitting the file to make the migration easier.
Moved the items, that do not depend on other libraries, to primitives/utils.dart.


packages/devtools_app/lib/src/primitives/utils.dart
```diff
@@ -0,0 +1,1422 @@
+// Copyright 2018 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: nit: 2022

Can you also add a todo somewhere at the top of this file (but below the copyright) to move this code back to shared/utils.dart once the null safety migration is complete? Maintaining two utils files doesn't seem necessary moving forward.

polina-c: Discussed offline and agreed to keep two utils.dart: higher level and lower level.

kenzieschmoll (approved):

PR #3640, opened 2022-02-08
Migrate app_error_handling.dart to null safety.

polina-c:


packages/devtools_app/lib/src/shared/app_error_handling.dart
```diff
@@ -26,10 +26,10 @@ void setupErrorHandling(Future Function() appStartCallback) {
   return runZonedGuarded(() async {
     WidgetsFlutterBinding.ensureInitialized();
 
-    final FlutterExceptionHandler oldHandler = FlutterError.onError;
+    final FlutterExceptionHandler? oldHandler = FlutterError.onError;
 
     FlutterError.onError = (FlutterErrorDetails details) {
-      _reportError(details.exception, details.stack);
+      _reportError(details.exception, details.stack!);
```
kenzieschmoll, line 32: instead of adding a bang to `details.stack`, let's make this a little more null safe by doing `details.stack ?? StackTrace.empty`

kenzieschmoll (approved):

PR #3642, opened 2022-02-08
Migrate connected_app.dart to null safety.

polina-c:


packages/devtools_app/lib/src/shared/connected_app.dart
```diff
@@ -36,62 +36,62 @@ class ConnectedApp {
   Future<bool> get isFlutterApp async => _isFlutterApp ??=
       await serviceManager.libraryUriAvailable(flutterLibraryUri);
 
-  bool get isFlutterAppNow {
+  bool? get isFlutterAppNow {
```
kenzieschmoll, line 39: this can be non-nullable

packages/devtools_app/lib/src/shared/connected_app.dart
```diff
@@ -36,62 +36,62 @@ class ConnectedApp {
...
     assert(_isFlutterApp != null);
-    return _isFlutterApp;
+    return _isFlutterApp == true;
   }
 
-  bool _isFlutterApp;
+  bool? _isFlutterApp;
```
kenzieschmoll, line 44: can we make the private fields late instead of nullable?

packages/devtools_app/lib/src/shared/connected_app.dart
```diff
@@ -36,62 +36,62 @@ class ConnectedApp {
...
-  FlutterVersion _flutterVersion;
+  FlutterVersion? _flutterVersion;
 
-  final _flutterVersionCompleter = Completer<FlutterVersion>();
+  final _flutterVersionCompleter = Completer<FlutterVersion?>();
 
   static const _flutterVersionTimeout = Duration(seconds: 3);
 
-  Future<bool> get isProfileBuild async {
+  Future<bool?> get isProfileBuild async {
```
kenzieschmoll, line 56: this, isProfileBuildNow, and isDartWebAppNow can all be non-nullable. Same comment about their respective private fields with the question as to whether we can make them late instead of nullable

packages/devtools_app/lib/src/shared/connected_app.dart
polina-c: All of them can be validated before initialization in connectedAppInitialized. We do not want this getter to fail. So, leaving nullable.

packages/devtools_app/lib/src/shared/connected_app.dart
```diff
@@ -36,62 +36,62 @@ class ConnectedApp {
...
+  bool get isFlutterWebAppNow => isFlutterAppNow! && isDartWebAppNow!;
 
-  bool get isFlutterNativeAppNow => isFlutterAppNow && !isDartWebAppNow;
+  bool get isFlutterNativeAppNow => isFlutterAppNow! && !isDartWebAppNow!;
 
-  bool get isDebugFlutterAppNow => isFlutterAppNow && !isProfileBuildNow;
+  bool get isDebugFlutterAppNow => isFlutterAppNow! && !isProfileBuildNow!;
 
-  bool get isRunningOnDartVM => serviceManager.vm.name != 'ChromeDebugProxy';
+  bool? get isRunningOnDartVM => serviceManager.vm!.name != 'ChromeDebugProxy';
```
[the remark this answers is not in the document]
polina-c: we read it from json, so it can be null

packages/devtools_app/lib/src/shared/connected_app.dart
polina-c: it cannot, because we override it in descendant OfflineConnectedApp, where it is nullable

packages/devtools_app/lib/src/shared/connected_app.dart
polina-c: fixed

packages/devtools_app/lib/src/shared/connected_app.dart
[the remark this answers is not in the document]
kenzieschmoll: `serviceManager.vm.name` may be null, but `serviceManager.vm.name != 'ChromeDebugProxy'` will not be.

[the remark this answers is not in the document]
kenzieschmoll: Ah I see now this overridden in `OfflineConnectedApp` as well. Nevermind then - nullable is fine.

packages/devtools_app/lib/src/shared/connected_app.dart
polina-c: reverted, because they are overriden too

kenzieschmoll (approved):

PR #3646, opened 2022-02-09
Migrate banner_messages.dart to null safety.

polina-c:


packages/devtools_app/lib/src/shared/banner_messages.dart
```diff
@@ -25,13 +26,13 @@ const preCompileShadersDocsUrl =
 
 class BannerMessagesController {
   final _messages = <String, ValueNotifier<List<BannerMessage>>>{};
-  final _dismissedMessageKeys = <Key>{};
+  final Set<Key?> _dismissedMessageKeys = <Key?>{};
```
kenzieschmoll, line 29: why do we need to add the type on the LHS? `<Key?>` should be enough

packages/devtools_app/lib/src/shared/banner_messages.dart
```diff
@@ -423,8 +422,8 @@ void maybePushDebugModePerformanceMessage(
 ) {
   if (offlineController.offlineMode.value) return;
   if (serviceManager.connectedApp?.isDebugFlutterAppNow ?? false) {
-    Provider.of<BannerMessagesController>(context)
-        .addMessage(DebugModePerformanceMessage(screenId).build(context));
+    Provider.of<BannerMessagesController>(context).addMessage(
+        DebugModePerformanceMessage(screenId).build(context) as BannerMessage);
```
kenzieschmoll, line 426: nit: add trailing comma

packages/devtools_app/lib/src/shared/banner_messages.dart
polina-c: done

packages/devtools_app/lib/src/shared/banner_messages.dart
polina-c: done

kenzieschmoll (approved):

PR #3647, opened 2022-02-09
Migrate common_widgets.dart to null safety.

polina-c:


packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -181,20 +181,24 @@ class IconLabelButton extends StatelessWidget {
...
-                style: denseAwareOutlinedButtonStyle(
-                  context,
-                  minScreenWidthForTextBeforeScaling,
-                ),
+                style: minScreenWidthForTextBeforeScaling != null
+                    ? denseAwareOutlinedButtonStyle(
+                        context,
+                        minScreenWidthForTextBeforeScaling!,
+                      )
+                    : null,
```
kenzieschmoll, line 189: instead of explicitly passing null here, can we just continue to have `denseAwareOutlinedButtonStyle` properly handle a null value for `minScreenWidthForTextBeforeScaling`? same comment for below

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -575,7 +579,7 @@ class BulletSpacer extends StatelessWidget {
   Widget build(BuildContext context) {
     final ThemeData theme = Theme.of(context);
 
-    TextStyle textStyle;
+    TextStyle? textStyle;
```
kenzieschmoll, line 582: this can be late instead of nullable

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -624,7 +628,8 @@ class Badge extends StatelessWidget {
       child: Text(
         text,
         // Use a slightly smaller font for the badge.
-        style: theme.primaryTextTheme.bodyText2.apply(fontSizeDelta: -1),
+        style: (theme.primaryTextTheme.bodyText2 ?? const TextStyle())
```
kenzieschmoll, line 631: can we do `theme.primaryTextTheme.bodyText2!` here instead of adding `?? const TextStyle()`

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -646,18 +651,18 @@ class DevToolsTooltip extends StatelessWidget {
...
-  final InlineSpan richMessage;
+  final String? message;
+  final InlineSpan? richMessage;
   final Widget child;
   final Duration waitDuration;
   final bool preferBelow;
-  final EdgeInsetsGeometry padding;
-  final Decoration decoration;
-  final TextStyle textStyle;
+  final EdgeInsetsGeometry? padding;
```
kenzieschmoll, line 659: this can be non-nullable since we pass a default value

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1443,23 +1448,23 @@ class FormattedJson extends StatelessWidget {
...
+    required this.url,
+    required this.gaScreenName,
+    required this.gaSelectedItemDescription,
   }) : super(key: key);
 
-  final String url;
+  final String? url;
 
-  final String gaScreenName;
+  final String? gaScreenName;
```
kenzieschmoll, line 1467: make these non-nullable

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1492,40 +1497,40 @@ class MoreInfoLink extends StatelessWidget {
...
 }
 
 class Link {
   const Link({this.display, this.url});
 
-  final String display;
+  final String? display;
 
-  final String url;
+  final String? url;
```
kenzieschmoll, line 1527: make these non-nullable

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1625,28 +1629,28 @@ class CopyToClipboardControl extends StatelessWidget {
...
   /// The callback to be called on change in addition to the notifier changes
   /// handled by this class.
-  final void Function(bool newValue) onChanged;
+  final void Function(bool? newValue)? onChanged;
 
   /// Whether this checkbox should be enabled for interaction.
   final bool enabled;
 
-  void _updateValue(bool value) {
+  void _updateValue(bool? value) {
```
kenzieschmoll, line 1649: All these bools should be able to be non-nullable

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1655,7 +1659,7 @@ class NotifierCheckbox extends StatelessWidget {
   Widget build(BuildContext context) {
     return ValueListenableBuilder(
       valueListenable: notifier,
-      builder: (context, value, _) {
+      builder: (context, dynamic value, _) {
```
kenzieschmoll, line 1662: this should be bool. You may have to add the type to ValueListenableBuilder (`ValueListenableBuilder<bool>`)

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1667,24 +1671,24 @@ class NotifierCheckbox extends StatelessWidget {
...
   final String title;
 
-  final String description;
+  final String? description;
 
-  final String tooltip;
+  final String? tooltip;
 
-  final void Function(bool newValue) onChanged;
+  final void Function(bool? newValue)? onChanged;
```
kenzieschmoll, line 1691: these bools can be non-nullable

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1667,24 +1671,24 @@ class NotifierCheckbox extends StatelessWidget {
...
+    required this.notifier,
+    required this.title,
     this.description,
     this.tooltip,
     this.onChanged,
     this.enabled = true,
   }) : super(key: key);
 
-  final ValueListenable<bool> notifier;
+  final ValueListenable<bool?> notifier;
```
kenzieschmoll, line 1683: This should be `ValueNotifier<bool>` so we don't have to type cast later in the build method

packages/devtools_app/lib/src/shared/common_widgets.dart
polina-c: denseAwareOutlinedButtonStyle requires non null minScreenWidthForTextBeforeScaling

packages/devtools_app/lib/src/shared/common_widgets.dart
polina-c: nope, because theme.textTheme.bodyText2 can be null

packages/devtools_app/lib/src/shared/common_widgets.dart
polina-c: what if theme.textTheme.bodyText2 is null?

packages/devtools_app/lib/src/shared/common_widgets.dart
polina-c: we pass nullable value here in maybeWrapWithTooltip

packages/devtools_app/lib/src/shared/common_widgets.dart
polina-c: done

packages/devtools_app/lib/src/shared/common_widgets.dart
polina-c: done

packages/devtools_app/lib/src/shared/common_widgets.dart
polina-c: done

packages/devtools_app/lib/src/shared/common_widgets.dart
polina-c: nope: CheckBox.onChanged requires handler for nullable value

packages/devtools_app/lib/src/shared/common_widgets.dart
polina-c: done

packages/devtools_app/lib/src/shared/common_widgets.dart
polina-c: nope: NotifierCheckbox.onChanged requires nullable parameter

packages/devtools_app/lib/src/shared/common_widgets.dart
polina-c: nope: NotifierCheckbox.onChanged requires nullable parameter

packages/devtools_app/lib/src/shared/common_widgets.dart
kenzieschmoll: We can change that. `denseAwareOutlinedButtonStyle ` calls `includeText`, which handles a null value for `minScreenWidthForTextBeforeScaling`

packages/devtools_app/lib/src/shared/common_widgets.dart
kenzieschmoll: ack

packages/devtools_app/lib/src/shared/common_widgets.dart
kenzieschmoll: ack

packages/devtools_app/lib/src/shared/common_widgets.dart
kenzieschmoll: We can still change it to be a ValueNotifier instead of a ValueListenable to eliminate the need to type cast

packages/devtools_app/lib/src/shared/common_widgets.dart
kenzieschmoll: ack

packages/devtools_app/lib/src/shared/common_widgets.dart
kenzieschmoll: Hmm. I think this value is set by default by the flutter framework, so we've always treated it as non-nullable in our code. But I see that all the values of TextTheme are nullable, so it's fine for us to be consistent here.

packages/devtools_app/lib/src/shared/common_widgets.dart
polina-c: done

kenzieschmoll (approved):

PR #3651, opened 2022-02-09
Migrate device_dialog to null safety.

polina-c:


packages/devtools_app/lib/src/shared/device_dialog.dart
```diff
@@ -38,9 +38,7 @@ class DeviceDialog extends StatelessWidget {
 
     final vm = serviceManager.vm;
 
-    if (vm == null) return const SizedBox();
```
kenzieschmoll, line 41: why are we no longer checking for a null vm?

packages/devtools_app/lib/src/shared/device_dialog.dart
```diff
@@ -157,7 +156,7 @@ class _VMFlagsDialogState extends State<VMFlagsDialog> with AutoDisposeMixin {
   }
 
   void _updateFromController() {
-    flags = infoController.flagListNotifier.value.flags
+    flags = infoController.flagListNotifier.value!.flags!
```
kenzieschmoll, line 159: we should either make sure flagListNotifier.value cannot return null (maybe empty list instead), or handle a possible null value here e.g. `infoController.flagListNotifier.value?.flags ?? []`

packages/devtools_app/lib/src/shared/device_dialog.dart
```diff
@@ -256,7 +255,7 @@ class _NameColumn extends ColumnData<_DialogFlag> {
         );
 
   @override
-  String getValue(_DialogFlag dataObject) => dataObject.name;
+  String? getValue(_DialogFlag dataObject) => dataObject.name;
```
kenzieschmoll, line 258: having these ColumnData overrides return null will likely cause issues with the table rendering. I'd set a default value (e.g. '')  instead and have these return types return a non-nullable string

packages/devtools_app/lib/src/shared/device_dialog.dart
polina-c: done

packages/devtools_app/lib/src/shared/device_dialog.dart
polina-c: done

kenzieschmoll: This comment applies to all of the other getValue / getTooltip overrides too

packages/devtools_app/lib/src/shared/device_dialog.dart
kenzieschmoll: It should be. It is set to null in vmServiceClosed: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/shared/service_manager.dart#L332

polina-c: right. ServiceManager is not null safe yet. Thanks for the catch.

packages/devtools_app/lib/src/shared/device_dialog.dart
polina-c: done

kenzieschmoll (approved):

PR #3652, opened 2022-02-09
Migrate dialogs.dart to null safety.

polina-c:


packages/devtools_app/lib/src/shared/dialogs.dart
```diff
@@ -115,17 +115,17 @@ class DialogApplyButton extends StatelessWidget {
 class DialogTextButton extends StatelessWidget {
   const DialogTextButton({this.onPressed, this.child});
 
-  final VoidCallback onPressed;
+  final VoidCallback? onPressed;
 
-  final Widget child;
+  final Widget? child;
```
kenzieschmoll, line 120: make this non-nullable and required if we are going to force unwrap it in the build method anyway. This is also consistent with TextButton

polina-c: done

kenzieschmoll (approved):

polina-c:
tested

PR #3659, opened 2022-02-10
Migrate landing_screen.dart to null safety.

polina-c:


packages/devtools_app/lib/src/shared/landing_screen.dart
```diff
@@ -259,15 +259,14 @@ class ImportFileInstructions extends StatelessWidget {
     final importedFile = await importFileFromPicker(
       acceptedTypes: ['json'],
     );
-    if (importedFile != null) {
-      Provider.of<ImportController>(context, listen: false)
-          .importData(importedFile);
-    }
+
+    Provider.of<ImportController>(context, listen: false)
```
kenzieschmoll, line 263: importFileFromPicker can return null, so we'll want to keep this null check here.

polina-c: done

polina-c:
tested

PR #3657, opened 2022-02-10
Migrate history_viewport.dart to null safety.

polina-c:

PR #3660, opened 2022-02-10
Migrate navigation.dart to null safety.

polina-c:


packages/devtools_app/lib/src/shared/navigation.dart
```diff
@@ -17,17 +15,17 @@ import 'theme.dart';
 ///
 /// This will fail because we can't determine what the theme value for the app
 /// is.
-String routeNameWithQueryParams(BuildContext context, String routeName,
-    [Map<String, String> queryParameters]) {
+String routeNameWithQueryParams(BuildContext? context, String routeName,
+    [Map<String, String>? queryParameters]) {
```
kenzieschmoll, line 19: nit: trailing comma

polina-c: I accidentally addressed this comment in https://github.com/flutter/devtools/pull/3657

PR #3667, opened 2022-02-10
Migrate screen.dart to null safety.

polina-c:


packages/devtools_app/lib/src/shared/screen.dart
```diff
@@ -155,7 +155,7 @@ abstract class Screen {
               Icon(icon, size: defaultIconSize),
               Padding(
                 padding: const EdgeInsets.only(left: denseSpacing),
-                child: Text(title),
+                child: Text(title ?? ''),
```
kenzieschmoll, line 158: instead of doing this here, let's make `title` non-nullable and give it a default value of '' in the constructor

kenzieschmoll (approved):

polina-c:
tested

PR #3676, opened 2022-02-11
Delete legacy performance code and show a warning for old flutter versions

kenzieschmoll:
This warning will rarely get hit since we are shipping in the SDK, but this will catch an edge case where a user ran DevTools and an app from two different versions of flutter.
Screen Shot 2022-02-11 at 2 49 28 PM


packages/devtools_app/lib/src/shared/banner_messages.dart
```diff
@@ -416,6 +417,60 @@ For the most accurate absolute memory stats, relaunch your application with the
...
+  }
+}
+
+void maybePushUnsupportedFlutterVersionWarning(
+  BuildContext context,
+  String screenId, {
+  required SemanticVersion supportedFlutterVersion,
+}) {
+  if (offlineController.offlineMode.value) return;
+  if (serviceManager.connectedApp?.isFlutterAppNow ?? false) {
```
elliette, line 460: Optional suggestion: return early if not a Flutter app to avoid the nested if-clauses

kenzieschmoll: done.

PR #3678, opened 2022-02-12
Migrate service_registrations.dart to null safety

polina-c:


packages/devtools_app/lib/src/shared/service_registrations.dart
```diff
@@ -19,11 +19,11 @@ class RegisteredServiceDescription {
     this.gaItem,
   });
 
-  final String service;
-  final String title;
-  final Widget icon;
-  final String gaScreenName;
-  final String gaItem;
+  final String? service;
+  final String? title;
```
kenzieschmoll, line 23: let's make `service` and `title` required and non-nullable

polina-c: done

PR #3679, opened 2022-02-12
Migrate service.dart to null safety.

polina-c:


packages/devtools_app/lib/src/shared/service.dart
```diff
@@ -24,20 +24,21 @@ Future<VmServiceWrapper> _connectWithSse(
...
+      client.stream?.asBroadcastStream() as Stream<String>;
   final service = VmServiceWrapper.fromNewVmService(
     stream,
-    client.sink.add,
+    client.sink!.add,
     uri,
   );
 
-  unawaited(client.sink?.done?.whenComplete(() {
+  unawaited(client.sink?.done.whenComplete(() {
```
kenzieschmoll, line 35: if sink is nullable, then don't we still need the `?` after `done`? what if `client.sink` is null?

packages/devtools_app/lib/src/shared/service.dart
```diff
@@ -24,20 +24,21 @@ Future<VmServiceWrapper> _connectWithSse(
       ? uri.replace(scheme: 'http')
       : uri.replace(scheme: 'https');
   final client = SseClient('$uri');
-  final Stream<String> stream = client.stream?.asBroadcastStream();
+  final Stream<String> stream =
+      client.stream?.asBroadcastStream() as Stream<String>;
```
kenzieschmoll, line 28: what if `client.stream?.asBroadcastStream()` is null? since `client.stream` can be null, `client.stream?.asBroadcastStream()` may be null. So should this be `Stream<String>?` instead?

packages/devtools_app/lib/src/shared/service.dart
polina-c: we don't: https://dartpad.dev/?id=b0607a3f57081b35099f702c55bc82bd

packages/devtools_app/lib/src/shared/service.dart
polina-c: Right. Force unwrapped `stream`. We invoke `stream.drain()` later in this method. So, we know it is not null.

kenzieschmoll: Maybe I'm missing something here, but since we used to call `client.stream?.asBroadcastStream()`, that would suggest that we expect `client.stream` to sometimes be null. So why are we we force unwrapping `client.stream` here if can sometimes be null? In particular, this looks like it would break our app when running on desktop: https://github.com/flutter/devtools/blob/a616832995d13410a4078b506c91927865ff6904/packages/devtools_app/lib/src/config_specific/sse/_fake_sse.dart#L17.

kenzieschmoll (approved):

polina-c:
tested

PR #3680, opened 2022-02-12
Migrate snapshot_screen.dart to null safety.

polina-c:


polina-c:
Not sure how to test it. I do not see print in 'build' printed.

polina-c:
tested

PR #3682, opened 2022-02-14
Migrate split.dart to null safety.

polina-c:


packages/devtools_app/lib/src/shared/split.dart
```diff
@@ -89,7 +86,7 @@ class Split extends StatefulWidget {
 }
 
 class _SplitState extends State<Split> {
-  List<double> fractions;
+  late List<double> fractions;
```
kenzieschmoll, line 89: can this be late final?

kenzieschmoll (approved):
LGTM with one comment

packages/devtools_app/lib/src/shared/split.dart
polina-c: done

polina-c:
tested

PR #3683, opened 2022-02-14
Migrate status_line.dart to null safety.

polina-c:


packages/devtools_app/lib/src/shared/status_line.dart
```diff
@@ -147,31 +143,31 @@ class StatusLine extends StatelessWidget {
...
-          final color = Theme.of(context).textTheme.bodyText2.color;
+          final color = Theme.of(context).textTheme.bodyText2!.color;
 
           return Row(
             mainAxisAlignment: MainAxisAlignment.end,
             children: [
               ValueListenableBuilder(
                 valueListenable: serviceManager.deviceBusy,
-                builder: (context, isBusy, _) {
+                builder: (context, dynamic isBusy, _) {
```
kenzieschmoll, line 164: always prefer to use the actual type if we know it (for this case, bool)

packages/devtools_app/lib/src/shared/status_line.dart
```diff
@@ -242,12 +238,12 @@ class IsolateSelector extends StatelessWidget {
...
               isDense: true,
               items: isolates.map(
-                (IsolateRef ref) {
+                (IsolateRef? ref) {
                   return DropdownMenuItem<IsolateRef>(
                     value: ref,
                     child: Row(
                       children: [
-                        ref.isSystemIsolate
+                        ref!.isSystemIsolate!
```
kenzieschmoll, line 246: what if `ref` is null? we should either handle that case or make isolateManager.isolates.value have type `<IsolateRef>[]` instead of `<IsolateRef?>[]`. Prefer the latter unless we are explicitly adding a null value to this list somewhere.

polina-c: done

packages/devtools_app/lib/src/shared/status_line.dart
```diff
@@ -238,12 +238,12 @@ class IsolateSelector extends StatelessWidget {
...
               isDense: true,
               items: isolates.map(
-                (IsolateRef? ref) {
+                (IsolateRef ref) {
                   return DropdownMenuItem<IsolateRef>(
                     value: ref,
                     child: Row(
                       children: [
-                        ref!.isSystemIsolate!
+                        ref.isSystemIsolate!
```
kenzieschmoll, line 246: let's make this a bit more null safe: `ref.isSystemIsolate ?? false`

kenzieschmoll (approved):
one comment, then lgtm

polina-c:
tested

PR #3684, opened 2022-02-14
Migrate survey.dart to null safety.

polina-c:


packages/devtools_app/lib/src/shared/survey.dart
```diff
@@ -58,23 +58,23 @@ class SurveyService {
         NotificationAction(
           _noThanksLabel,
           () => _noThanksPressed(
-            message: message,
+            message: message!,
```
kenzieschmoll, line 61: unwrap the `message` var where it is defined, and then you won't have to force unwrap it for every use

kenzieschmoll (approved):
LGTM with one comment

packages/devtools_app/lib/src/shared/survey.dart
polina-c: done

PR #3685, opened 2022-02-14
Migrate table_data.dart to null safety.

polina-c:


kenzieschmoll (commented):
I just took a closer look at this code. Looks like the classes `HoverCellData`, `TableDataClient`, `TableData` and `TreeTableData` aren't used anywhere in DevTools. Let's delete those classes.

Polina: done

packages/devtools_app/lib/src/shared/table_data.dart
```diff
@@ -363,22 +362,22 @@ abstract class ColumnData<T> {
 
   int compare(T a, T b) {
     final Comparable valueA = getValue(a);
-    final Comparable valueB = getValue(b);
+    final Comparable? valueB = getValue(b);
```
kenzieschmoll, line 365: non-nullable

packages/devtools_app/lib/src/shared/table_data.dart
```diff
@@ -363,22 +362,22 @@ abstract class ColumnData<T> {
...
   dynamic getValue(T dataObject);
 
   /// Get the cell's display value from the given [dataObject].
-  String getDisplayValue(T dataObject) => getValue(dataObject).toString();
+  String? getDisplayValue(T dataObject) => getValue(dataObject).toString();
 
   // TODO(kenz): this isn't hooked up to the table elements. Do this.
   /// Get the cell's tooltip value from the given [dataObject].
-  String getTooltip(T dataObject) => getDisplayValue(dataObject);
+  String? getTooltip(T dataObject) => getDisplayValue(dataObject);
```
kenzieschmoll, line 377: make these return non-nullable strings

packages/devtools_app/lib/src/shared/table_data.dart
```diff
@@ -329,10 +329,9 @@ abstract class ColumnData<T> {
   ColumnData(
     this.title, {
     this.titleTooltip,
-    @required this.fixedWidthPx,
+    required double this.fixedWidthPx,
```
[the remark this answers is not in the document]
polina-c: Because type of parameter is different with type of the field. There are two constructors here. One requires fixedWidthPx, and in another it is null.

packages/devtools_app/lib/src/shared/table_data.dart
polina-c: done

packages/devtools_app/lib/src/shared/table_data.dart
polina-c: done

kenzieschmoll (approved):

polina-c:
tested

PR #3691, opened 2022-02-14
draft - Migrate vm_service_wrapper.dart to null safety.

polina-c:


polina-c:
service.dart, line 29:
```
final service = VmServiceWrapper.fromNewVmService(
    stream,
    client.sink!.add,
    uri,
  );
```

How do we want to handle if stream is null?
Throw exception before calling VmServiceWrapper.fromNewVmService?

kenzieschmoll:
> Throw exception before calling VmServiceWrapper.fromNewVmService?

Yes, let's do this because the VmService class expects a non-null `inStream` parameter.

PR #3695, opened 2022-02-15
Fix Google internal error in file_import.dart

polina-c:
https://paste.googleplex.com/5013282847981568?raw

### DIFF
--- a/packages/devtools_app/lib/src/shared/file_import.dart
+++ b/packages/devtools_app/lib/src/shared/file_import.dart
@@ -227,6 +227,10 @@ Future<DevToolsJsonFile?> importFileFromPicker({
   final lastModifiedTime = await file.lastModified();
   // TODO(kenz): this will need to be modified if we need to support other file
   // extensions than .json. We will need to return a more generic file type.
+
+  // We need this assert
+  // ignore: unnecessary_null_comparison
+  assert(file.name != null);
   return DevToolsJsonFile(
     data: data,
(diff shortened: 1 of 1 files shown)


elliette (approved):

packages/devtools_app/lib/src/shared/file_import.dart
```diff
@@ -227,6 +227,10 @@ Future<DevToolsJsonFile?> importFileFromPicker({
   final lastModifiedTime = await file.lastModified();
   // TODO(kenz): this will need to be modified if we need to support other file
   // extensions than .json. We will need to return a more generic file type.
+
+  // We need this assert
+  // ignore: unnecessary_null_comparison
+  assert(file.name != null);
```
kenzieschmoll, line 233: is the proper fix here to unwrap `file.name!` where it is used?

polina-c: We will still get warning, because, file.name is not nullable in latest version. It seems version of file in Google3 is outdated.

kenzieschmoll: @elliette can you just re-run the roll script in google3 to pick up the version of file_import.dart that is null-safe?

elliette: I think I do have the null-safe version, I ran the script ~15 mins ago

PR #3696, opened 2022-02-16
Changes to enable devtools server tests to run on DDS

kenzieschmoll:
Some of these changes are refactors (moving files to devtools_shared to be exposed in dds) and some of these changes are bug fixes.


packages/devtools_app/lib/src/shared/vm_service_wrapper.dart
```diff
@@ -1166,15 +1166,6 @@ class VmServiceWrapper implements VmService {
     }
   }
 
-  /// Gets the name of the service stream for the connected VM service. Pre-v3.22
```
bkonyi, line 1169: I wonder if we can start dropping these version checks now that we're shipping via the SDK... Might be worth getting rid of special cases like there everywhere at some point.

bkonyi (approved):

packages/devtools_app/lib/src/shared/vm_service_wrapper.dart
kenzieschmoll: Agreed. Filed https://github.com/flutter/devtools/issues/3699 to track.

PR #3701, opened 2022-02-16
Upgrade flutter version.

polina-c:


kenzieschmoll (approved):

third_party/packages/ansi_up/lib/ansi_up.dart
```diff
@@ -304,11 +304,11 @@ enum PacketKind {
...
 
 class _TextPacket {
-  _TextPacket({required this.kind, this.text = '', this.url = ''});
+  _TextPacket({required this.kind});
 
   PacketKind kind;
-  String text;
-  final String url;
+  String text = '';
+  final String url = '';
```
kenzieschmoll, line 311: why is this change necessary?

polina-c: I got warning from analyzer, that the parameter is never used, that is true.

polina-c:
On 2.11.0-0.1.pre Gallery crashes when DevTools connects to it with the exception described here: https://github.com/flutter/flutter/issues/97500

We need to wait for fix to be in a tagged version.

PR #3714, opened 2022-02-21
Migrate vm_service_wrapper.dart to null safety.

polina-c:


packages/devtools_app/lib/src/service/service.dart
```diff
@@ -23,8 +23,8 @@ Future<VmServiceWrapper> _connectWithSse(
       ? uri.replace(scheme: 'http')
       : uri.replace(scheme: 'https');
   final client = SseClient('$uri');
-  final Stream<String>? stream =
-      client.stream?.asBroadcastStream() as Stream<String>?;
+  final Stream<String> stream =
+      client.stream!.asBroadcastStream() as Stream<String>;
```
kenzieschmoll, line 27: I don't think we should be unconditionally unwrapping this here. On desktop targets, SseClient will come from _fake_sse.dart, which returns a null value for `stream`

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -37,10 +37,10 @@ class VmServiceWrapper implements VmService {
     );
   }
 
-  VmService _vmService;
-  Version _protocolVersion;
-  Version _dartIoVersion;
-  Version _ddsVersion;
+  VmService? _vmService;
```
kenzieschmoll, line 40: this should be late final and non-nullable

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -58,13 +58,13 @@ class VmServiceWrapper implements VmService {
     }
   }
 
-  Uri get connectedUri => _connectedUri;
-  final Uri _connectedUri;
+  Uri? get connectedUri => _connectedUri;
+  final Uri? _connectedUri;
```
kenzieschmoll, line 62: can this be non-nullable?

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -58,13 +58,13 @@ class VmServiceWrapper implements VmService {
...
 
   final bool trackFutures;
   final Map<String, Future<Success>> _activeStreams = {};
 
-  final Set<TrackedFuture<Object>> activeFutures = {};
+  final Set<TrackedFuture<Object?>> activeFutures = {};
```
kenzieschmoll, line 67: why does object need to be nullable

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -77,9 +77,9 @@ class VmServiceWrapper implements VmService {
   /// finish before completing.
   Future<void> forEachIsolate(
       Future<void> Function(IsolateRef) callback) async {
-    final vm = await _vmService.getVM();
+    final vm = await _vmService!.getVM();
     final futures = <Future>[];
-    for (final isolate in vm.isolates) {
+    for (final isolate in vm.isolates!) {
```
kenzieschmoll, line 82: write `vm.isolates ?? <IsolateRef>[]` instead

packages/devtools_app/lib/src/service/service.dart
polina-c: hm

How do we want to handle this? 
Should the constructor VmServiceWrapper.fromNewVmService  accept null for stream? If yes, what should it pass to the constructor VmService, that requires non-nullable stream.

kenzieschmoll: I see. I just verified - looks like we hit `_connectWithWebSocket` on Desktop, not `_connectWithSse`, so what you have should be fine.

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
jacob314: this is final to be nullable. The type returned by the future doesn't matter.

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
jacob314: nit: `?? []` should be sufficient.

polina-c: done

packages/devtools_app/lib/src/service/vm_flags.dart
```diff
@@ -34,7 +34,7 @@ class VmFlagManager extends Disposer {
   }
 
   Future<void> _initFlags() async {
-    final flagList = await service!.getFlagList() ?? FlagList(flags: []);
+    final flagList = await service?.getFlagList() ?? FlagList(flags: []);
```
kenzieschmoll, line 37: can _service be late and non-nullable instead? that way service can be non-nullable

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -121,16 +121,16 @@ class VmServiceWrapper implements VmService {
 
   @override
   Future<Response> callMethod(String method,
-      {String isolateId, Map<String, dynamic> args}) {
+      {String? isolateId, Map<String, dynamic>? args}) {
```
kenzieschmoll, line 124: nit: trailing comma

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -313,8 +313,8 @@ class VmServiceWrapper implements VmService {
       processStackFrame(current: root, parent: null);
 
       // Build the trace events.
-      for (final sample in cpuSamples.samples) {
-        final tree = _CpuProfileTimelineTree.getTreeFromSample(sample);
+      for (final sample in cpuSamples.samples!) {
```
kenzieschmoll, line 316: make this more null safe: `cpuSamples.samples ?? <CpuSample>[]`

packages/devtools_app/lib/src/service/vm_flags.dart
polina-c: done

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -427,15 +427,15 @@ class VmServiceWrapper implements VmService {
...
+    String? scriptId,
+    int? tokenPos,
+    int? endTokenPos,
+    bool? forceCompile,
+    bool? reportLines,
   }) async {
     // Workaround for https://github.com/flutter/devtools/issues/2981.
     // TODO(bkonyi): remove after Flutter stable is on Dart SDK > 2.12.
-    if (await serviceManager.connectedApp.isProfileBuild) {
+    if (await serviceManager.connectedApp!.isProfileBuild) {
```
kenzieschmoll, line 438: based on the above comment, I think you can remove this whole if block

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -502,17 +502,17 @@ class VmServiceWrapper implements VmService {
       supportedVersion: SemanticVersion(major: 1, minor: 3),
       isolateId: isolateId,
     )) {
-      return isolate.extensionRPCs
+      return isolate.extensionRPCs!
```
kenzieschmoll, line 505: to make this a bit more null safe, add a variable after `isolate` is defined and use that for both these checks

`final extensionRPCs = isolate.extensionRPCs ?? <String>[]`

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -541,7 +541,7 @@ class VmServiceWrapper implements VmService {
   // https://github.com/dart-lang/sdk/blob/master/pkg/vm_service/lib/src/dart_io_extensions.dart
   Future<bool> isHttpProfilingAvailable(String isolateId) async {
     final Isolate isolate = await getIsolate(isolateId);
-    return isolate.extensionRPCs.contains('ext.dart.io.getHttpProfile');
+    return isolate.extensionRPCs!.contains('ext.dart.io.getHttpProfile');
```
kenzieschmoll, line 544: same comment here: `(isolate.extensionRPCs ?? <String>[]).contains...`

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -590,12 +590,12 @@ class VmServiceWrapper implements VmService {
   // https://github.com/dart-lang/sdk/blob/master/pkg/vm_service/lib/src/dart_io_extensions.dart
   Future<bool> isSocketProfilingAvailable(String isolateId) async {
     final Isolate isolate = await getIsolate(isolateId);
-    return isolate.extensionRPCs.contains('ext.dart.io.getSocketProfile');
+    return isolate.extensionRPCs!.contains('ext.dart.io.getSocketProfile');
```
kenzieschmoll, line 593: ditto

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -962,7 +962,7 @@ class VmServiceWrapper implements VmService {
 
   @override
   Future<CpuSamples> getAllocationTraces(String isolateId,
-      {int timeOriginMicros, int timeExtentMicros, String classId}) {
+      {int? timeOriginMicros, int? timeExtentMicros, String? classId}) {
```
kenzieschmoll, line 965: nit: trailing comma

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -974,21 +974,19 @@ class VmServiceWrapper implements VmService {
   @override
   Future<Success> setTraceClassAllocation(
     String isolateId,
-    String classId,
+    String? classId,
```
kenzieschmoll, line 977: the parameter types in our wrapper class should match the parameter types from VmService. Make this non-nullable and then fix errors upstream at the call sites

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -974,21 +974,19 @@ class VmServiceWrapper implements VmService {
...
     bool enable,
   ) async {
     if (await isProtocolVersionSupported(
         supportedVersion: SemanticVersion(major: 3, minor: 43))) {
       return trackFuture('setTraceClassAllocation',
-          _vmService.setTraceClassAllocation(isolateId, classId, enable));
+          _vmService.setTraceClassAllocation(isolateId, classId!, enable));
```
kenzieschmoll, line 983: nit: trailing comma

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -1226,10 +1223,10 @@ class _CpuProfileTimelineTree {
     final root = _CpuProfileTimelineTree._fromIndex(cpuSamples, kRootIndex);
     _CpuProfileTimelineTree current;
     // TODO(bkonyi): handle truncated?
-    for (final sample in cpuSamples.samples) {
+    for (final sample in cpuSamples.samples!) {
```
kenzieschmoll, line 1226: `cpuSamples.samples ?? <CpuSample>[]`

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -1317,18 +1314,18 @@ extension VmServicePrivate on VmServiceWrapper {
   }
 
   /// Forces the VM to perform a full garbage collection.
-  Future<Success> collectAllGarbage() => _privateRpcInvoke(
+  Future<Success?> collectAllGarbage() => _privateRpcInvoke(
```
kenzieschmoll, line 1317: why does this have to be nullable if `_privateRpcInvoke` has a non-nullable return type? Same question for RPCs below

packages/devtools_app/lib/src/shared/device_dialog.dart
```diff
@@ -62,9 +62,7 @@ class DeviceDialog extends StatelessWidget {
         'Framework / Engine': '${flutterVersion!.frameworkRevision} / '
             '${flutterVersion!.engineRevision}',
       },
-      if (serviceManager.service!.connectedUri != null)
-        'VM Service Connection':
-            serviceManager.service!.connectedUri.toString(),
+      'VM Service Connection': serviceManager.service!.connectedUri.toString(),
```
kenzieschmoll, line 65: instead of removing the if check, check that `serviceManager.service?.connectedUri != null`

packages/devtools_app/lib/src/shared/eval_on_dart_library.dart
```diff
@@ -298,7 +298,7 @@ class EvalOnDartLibrary extends DisposableController
     final hash = await (evalInstance(
       'instance.hashCode',
       isAlive: isAlive,
-      scope: {'instance': instance.id},
+      scope: {'instance': instance.id!},
```
kenzieschmoll, line 301: are all the unwraps of `.id!` safe to do? Why are we making these non-nullable if they may be null?

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
polina-c: done

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
polina-c: done

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
polina-c: done

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
polina-c: done

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
polina-c: done

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
polina-c: done

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
polina-c: done

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
polina-c: PortList.parse returns PortList?
So, T=PortList?, i.e. return type is nullable

packages/devtools_app/lib/src/shared/device_dialog.dart
polina-c: done

packages/devtools_app/lib/src/shared/eval_on_dart_library.dart
polina-c: This scope will end up to be passed to SDK's VmService.evaluate, that takes Map<String, String>?.
So, I made all methods, that take scope, taking Map<String, String>?, not Map<String, String?>?

It should be safe, because VmService.evaluate does not fail accepting parameters in the current version, i.e. all existing code paths have id not null.

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
polina-c: done

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
polina-c: done

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
polina-c: done

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
kenzieschmoll: If that is the case, then the `parser` parameter in `_privateRpcInvoke` should be: `required T? Function(Map<String, dynamic>?) parser,` (note the T? instead of T). And `_privateRpcInvoke` should also return a nullable type: `Future<T?>`

polina-c: done

kenzieschmoll (approved):

polina-c:
tested

PR #3720, opened 2022-02-24
Migrate vm_flag_widgets.dart to null safety.

polina-c:


packages/devtools_app/lib/src/ui/vm_flag_widgets.dart
```diff
@@ -54,7 +55,8 @@ class ProfileGranularityDropdown extends StatelessWidget {
             Provider.of<BannerMessagesController>(context);
         if (safeValue == highProfilePeriod) {
           bannerMessageController.addMessage(
-            HighProfileGranularityMessage(screenId).build(context),
+            HighProfileGranularityMessage(screenId).build(context)
+                as BannerMessage,
```
kenzieschmoll, line 59: let's try to change the return type of HighProfileGranularityMessage to BannerMessage instead of Widget. Then we can get rid of this cast

packages/devtools_app/lib/src/ui/vm_flag_widgets.dart
```diff
@@ -68,9 +70,10 @@ class ProfileGranularityDropdown extends StatelessWidget {
           style: Theme.of(context).textTheme.bodyText2,
           value: safeValue,
           items: [
-            _buildMenuItem(ProfileGranularity.low),
-            _buildMenuItem(ProfileGranularity.medium),
-            _buildMenuItem(ProfileGranularity.high),
+            _buildMenuItem(ProfileGranularity.low) as DropdownMenuItem<String>,
```
kenzieschmoll, line 73: similar comment. Make the return type of _buildMenuItem be `DropdownMenuItem<String>`. Then we should be able to remove the casts

packages/devtools_app/lib/src/ui/vm_flag_widgets.dart
```diff
@@ -85,11 +88,11 @@ class ProfileGranularityDropdown extends StatelessWidget {
     );
   }
 
-  Future<void> _onProfileGranularityChanged(String newValue) async {
+  Future<void> _onProfileGranularityChanged(String? newValue) async {
```
kenzieschmoll, line 91: can this be non-nullable?

packages/devtools_app/lib/src/ui/vm_flag_widgets.dart
polina-c: done

packages/devtools_app/lib/src/ui/vm_flag_widgets.dart
polina-c: done

packages/devtools_app/lib/src/ui/vm_flag_widgets.dart
polina-c: nope
This method is assigned to RoundedDropDownButton<String>.onChanged that has nullable parameter

kenzieschmoll (approved):

polina-c:
tested

PR #3721, opened 2022-02-24
Migrate tab.dart to null safety.

polina-c:


packages/devtools_app/lib/src/ui/tab.dart
```diff
@@ -19,13 +19,13 @@ class DevToolsTab extends Tab {
   /// The only difference is this tab makes more of an effort to reflect
   /// changes in font and icon sizes.
   DevToolsTab._({
```
kenzieschmoll, line 21: make key and gaid required

packages/devtools_app/lib/src/ui/tab.dart
```diff
@@ -63,9 +63,9 @@ class DevToolsTab extends Tab {
   }
 
   /// Tab id for google analytics.
-  final String gaId;
+  final String? gaId;
```
kenzieschmoll, line 66: non-nullable

packages/devtools_app/lib/src/ui/tab.dart
```diff
@@ -165,8 +165,8 @@ class _AnalyticsTabbedViewState extends State<AnalyticsTabbedView>
 
   @override
   void dispose() {
-    _tabController?.removeListener(_onTabChanged);
-    _tabController?.dispose();
+    _tabController!.removeListener(_onTabChanged);
+    _tabController!.dispose();
```
kenzieschmoll, line 169: leave the ? null checks here instead of force unwrapping

polina-c: done

packages/devtools_app/lib/src/ui/tab.dart
polina-c: done

packages/devtools_app/lib/src/ui/tab.dart
polina-c: done

packages/devtools_app/lib/src/ui/tab.dart
```diff
@@ -19,13 +19,13 @@ class DevToolsTab extends Tab {
   /// The only difference is this tab makes more of an effort to reflect
   /// changes in font and icon sizes.
   DevToolsTab._({
-    Key key,
-    String text,
-    Icon icon,
+    required Key? key,
```
kenzieschmoll, line 22: required and non-nullable. The only place this should be constructed is from the factory constructor, and we know it is non-null.

kenzieschmoll (approved):
one comment then lgtm

polina-c:
tested

packages/devtools_app/lib/src/ui/tab.dart
polina-c: done

polina-c:
tested

PR #3724, opened 2022-02-24
Migrate icons.dart to null safety

polina-c:


packages/devtools_app/lib/src/ui/icons.dart
```diff
@@ -95,31 +95,31 @@ class CircleIcon extends StatelessWidget {
 }
 
 class CustomIconMaker {
-  final Map<String, Widget> iconCache = {};
+  final Map<String?, Widget> iconCache = {};
```
kenzieschmoll, line 98: String should be non-nullable

kenzieschmoll (approved):
one comment - then lgtm

packages/devtools_app/lib/src/ui/icons.dart
polina-c: done

polina-c:
tested

PR #3734, opened 2022-02-25
Migrate filter.dart to null safety.

polina-c:


packages/devtools_app/lib/src/ui/filter.dart
```diff
@@ -99,7 +97,7 @@ class _FilterDialogState<T> extends State<FilterDialog> with AutoDisposeMixin {
               const SizedBox(height: defaultSpacing),
             ],
             if (widget.toggleFilters != null)
-              for (final toggleFilter in widget.toggleFilters) ...[
+              for (final toggleFilter in widget.toggleFilters ?? []) ...[
```
kenzieschmoll, line 100: can you just use ! here instead since we already checked that widget.toggleFilters != null?

packages/devtools_app/lib/src/ui/filter.dart
```diff
@@ -111,9 +109,9 @@ class _FilterDialogState<T> extends State<FilterDialog> with AutoDisposeMixin {
             Filter<T>(
               queryFilter: widget.includeQueryFilter
                   ? QueryFilter.parse(queryTextFieldController.value.text,
-                      widget.queryFilterArguments)
+                      widget.queryFilterArguments!)
                   : null,
-              toggleFilters: widget.toggleFilters,
+              toggleFilters: widget.toggleFilters as List<ToggleFilter<T>>?,
```
kenzieschmoll, line 114: this type cast shouldn't be necessary. `List<ToggleFilter<T>>?` is already the type of `widget.toggleFilters`

packages/devtools_app/lib/src/ui/filter.dart
```diff
@@ -300,7 +298,7 @@ class QueryFilterArgument {
     return false;
   }
 
-  bool matchesValue(String dataValue, {bool substringMatch = false}) {
+  bool matchesValue(String? dataValue, {bool substringMatch = false}) {
```
kenzieschmoll, line 301: make this non-nullable since we aren't doing any null checking on this value. Deal with the error upstream to ensure we aren't passing a null value for dataValue here.

packages/devtools_app/lib/src/ui/filter.dart
polina-c: done

packages/devtools_app/lib/src/ui/filter.dart
polina-c: done

packages/devtools_app/lib/src/ui/filter.dart
polina-c: done

kenzieschmoll (approved):

polina-c:
tested

PR #3737, opened 2022-02-26
Migrate chart_controller to null safety.

polina-c:


packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -74,21 +74,21 @@ class ChartController extends DisposableController
   double _tickWidth = 10.0;
 
   /// Number of ticks visible (on the X-axis);
-  int visibleTicks;
+  int? visibleTicks;
```
kenzieschmoll, line 77: can this be late?

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -104,13 +104,13 @@ class ChartController extends DisposableController
...
-  double xCanvasChart;
+  double? xCanvasChart;
 
   /// Width of the canvas for plotting data (#).
-  double canvasChartWidth;
+  double? canvasChartWidth;
 
   /// Right-side padding after computing minPadding and max number of integral ticks for canvasChartWidth.
-  double xPaddingRight;
+  double? xPaddingRight;
```
kenzieschmoll, line 113: can these all be late?

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -239,9 +239,12 @@ class ChartController extends DisposableController
       // Greater or equal to range we're zooming in on?
       if (lastDT.difference(firstDT).inMinutes >= duration.inMinutes) {
         // Grab the duration in minutes passed in.
-        final startOfLastNMinutes = timestamps.reversed.firstWhere(
+        final startOfLastNMinutes =
+            // We need this cast to be able to return null if nothing is found.
+            // ignore: unnecessary_cast
+            (timestamps as List<int?>).reversed.firstWhere(
```
kenzieschmoll, line 245: use firstWhereOrNull instead

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -478,20 +481,20 @@ class ChartController extends DisposableController
     return traceIndex;
   }
 
-  int get numberOfVisibleXAxisTicks => visibleTicks;
+  int? get numberOfVisibleXAxisTicks => visibleTicks;
```
kenzieschmoll, line 484: let's remove this getter and just use visibleTicks instead wherever numberOfVisibleXAxisTicks was used. If we need to rename visibleTicks to visibleXAxisTicks to make this more clear, let's do that.

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -478,20 +481,20 @@ class ChartController extends DisposableController
...
     final leftIndex = totalTimestampTicks;
     if (leftIndex > 0) return leftIndex;
 
     // Less ticks than total size of ticks to show.
-    return numberOfVisibleXAxisTicks - timestampsLength;
+    return numberOfVisibleXAxisTicks! - timestampsLength;
   }
 
-  bool isTimestampVisible(int timestamp, int timestampIndex) {
+  bool isTimestampVisible(int? timestamp, int timestampIndex) {
```
kenzieschmoll, line 497: remove the timestamp parameter completely as it is not used

packages/devtools_app/lib/src/charts/chart_controller.dart
```diff
@@ -586,10 +590,10 @@ class TapLocation {
         timestamp = original.timestamp,
         index = original.index;
 
-  final TapDownDetails tapDownDetails;
+  final TapDownDetails? tapDownDetails;
```
kenzieschmoll, line 593: this should be able to be non-nullable

packages/devtools_app/lib/src/charts/chart_controller.dart
polina-c: I do not think so, because there is getter numberOfVisibleXAxisTicks that can be invoked before initialization.

packages/devtools_app/lib/src/charts/chart_controller.dart
polina-c: Nope as it is public filed that can be accessed before initialization

packages/devtools_app/lib/src/charts/chart_controller.dart
polina-c: done

packages/devtools_app/lib/src/charts/chart_controller.dart
polina-c: done

packages/devtools_app/lib/src/charts/chart_controller.dart
polina-c: done

packages/devtools_app/lib/src/charts/chart_controller.dart
polina-c: Nope. It is set to null in 'copy'.

kenzieschmoll (approved):

polina-c:
tested

PR #3739, opened 2022-02-26
Migrate treemap.dart to null safety.

polina-c:


packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -18,28 +16,28 @@ enum PivotType { pivotByMiddle, pivotBySize }
...
-    @required this.levelsVisible,
+    required this.levelsVisible,
     this.isOutermostLevel = false,
-    @required this.width,
-    @required this.height,
-    @required this.onRootChangedCallback,
+    required this.width,
+    required this.height,
+    required this.onRootChangedCallback,
   });
```
kenzieschmoll, line 26: we should add an assert for both of these constructors that `rootNode == null != nodes == null`

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -421,7 +419,7 @@ class _TreemapState extends State<Treemap> {
   /// ----------------------------
   /// ```
   Widget buildTreemap(BuildContext context) {
-    if (widget.rootNode.children.isNotEmpty) {
+    if (widget.rootNode!.children.isNotEmpty) {
```
kenzieschmoll, line 422: add an assert at the beginning of this method to verify widget.rootNode != null

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -370,7 +368,7 @@ class _TreemapState extends State<Treemap> {
           return const SizedBox();
         }
         final positionedChildren = buildTreemaps(
-          children: widget.nodes,
+          children: widget.nodes!,
```
kenzieschmoll, line 371: add an assert at the beginning of this method: `assert(widget.nodes?.isNotEmpty ?? false)`

packages/devtools_app/lib/src/charts/treemap.dart
polina-c: done

packages/devtools_app/lib/src/charts/treemap.dart
polina-c: done

packages/devtools_app/lib/src/charts/treemap.dart
polina-c: `widget.nodes != null && widget.nodes!.isNotEmpty` seems to be more readable

kenzieschmoll (approved):

polina-c:
tested

PR #3744, opened 2022-02-27
Migrate drag_and_drop to null safety.

polina-c:


packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
```diff
@@ -20,13 +18,13 @@ abstract class DragAndDropManager {
     init();
   }
 
-  static DragAndDropManager get instance => _instance ?? DragAndDropManager();
+  static DragAndDropManager get instance => _instance;
 
-  static DragAndDropManager _instance;
+  static late DragAndDropManager _instance = DragAndDropManager();
```
kenzieschmoll, line 23: instead of making this late and assigning _instance to `DragAndDropManager()`, what if we do this:
```
static DragAndDropManager get instance => _instance != null ? instance! : DragAndDropManager();
static late final DragAndDropManager? _instance;
```

This is more of an equivalent to the code we had before, where _instance is only being assigned once.

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
```diff
@@ -112,9 +109,9 @@ class DragAndDrop extends StatefulWidget {
 class DragAndDropState extends State<DragAndDrop> {
   final _dragging = ValueNotifier<bool>(false);
 
-  NotificationsState notifications;
+  NotificationsState? notifications;
```
kenzieschmoll, line 112: can this be late instead?

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
polina-c: late is a right way to implement lazy pattern: https://dev.to/pktintali/late-variables-in-dart-dart-learning-series-1-2opf
For some reasons `init` assigns it to `this`. I do not understand it so I left it as is and that's why the variable is not final.

But the code before initialized the field more than once.

What do you think?

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
polina-c: I would not. It is public field. So it can be accessed before didChangeDependencies, that set it.

kenzieschmoll: didChangeDependencies will be called directly after initState the first time the state is created, so it should be safe to assume `notifications` will be initialized anywhere it is accessed. If it is null, we are going to throw an error anyway when `activeState!.notifications!` is called, so making this late will just throw a late initialization error instead of a null operator error

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
kenzieschmoll: In the previous code, the only place _instance is assigned is in init: `_instance = this;`. This code is a bit confusing though because `_instance = DragAndDropManager()` is essentially the same thing as creating the DragAndDropManager constructor which calls init and sets `_instance` to `this` anyway.

How about this:
```
static DragAndDropManager get instance => _instance ??= DragAndDropManager(); 
static late final DragAndDropManager _instance;
```

And by doing this, you can get rid of the init method body all together. Make init an abstract method `void init();` so that subclasses still have to override it (you can add an empty override for the desktop impl). You can get rid of the mustCallSuper annotation then too.

polina-c: done

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
polina-c: done

packages/devtools_app/lib/src/config_specific/drag_and_drop/drag_and_drop.dart
```diff
@@ -18,17 +18,15 @@ abstract class DragAndDropManager {
...
-
-  static DragAndDropManager? _instance;
+  static late final DragAndDropManager _instance = DragAndDropManager();
 
   final _dragAndDropStates = <DragAndDropState>{};
 
   DragAndDropState? activeState;
 
-  void init();
+  void init() {}
```
kenzieschmoll, line 29: make this abstract instead of empty. You can make the desktop override empty.

polina-c: What's the point to obligate descendants to implement it?

polina-c: done

kenzieschmoll (approved):

polina-c:
tested

PR #3749, opened 2022-02-27
Migrate import_export to null safety.

polina-c:


packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
```diff
@@ -51,9 +51,9 @@ class ImportController {
 
   final void Function(String screenId) _pushSnapshotScreenForImport;
 
-  final NotificationService _notifications;
+  final NotificationService? _notifications;
```
kenzieschmoll, line 54: make this non-nullable since we just unwrap it everywhere it is used here

kenzieschmoll (approved):
One comment, then LGTM

packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
polina-c: done

polina-c: I did this and scaffold_test.dart started failing because the constructor Notifications.of(context) returns null. 
Is there a way to update the test so that `context.dependOnInheritedWidgetOfExactType<_InheritedNotifications>();` returns non null?

kenzieschmoll: Yes, try using the wrapWithNotifications helper. It is used in a few other tests so there are some examples.

polina-c:
tested

PR #3751, opened 2022-02-27
Migrate notifications to null saftety.

polina-c:


packages/devtools_app/lib/src/config_specific/notifications/notifications_web.dart
```diff
@@ -2,16 +2,14 @@
...
 import 'dart:html' as html;
 
 class Notification {
-  Notification(String title, {String body}) {
+  Notification(String title, {String? body}) {
     _impl = html.Notification(title, body: body);
   }
 
-  html.Notification _impl;
+  late html.Notification _impl;
```
kenzieschmoll, line 12: late final?

kenzieschmoll (approved):
one comment - then lgtm

packages/devtools_app/lib/src/config_specific/notifications/notifications_web.dart
polina-c: done

PR #3779, opened 2022-03-04
Migrate http_request_data.dart to null safety.

polina-c:


packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -47,21 +47,21 @@ class DartIOHttpInstantEvent extends HttpInstantEvent {
...
 
   // This is set from within HttpRequestData.
-  TimeRange _timeRange;
+  TimeRange? _timeRange;
 }
 
 /// An abstraction of an HTTP request made through dart:io.
 abstract class HttpRequestData extends NetworkRequest {
-  HttpRequestData(int timelineMicrosBase) : super(timelineMicrosBase);
+  HttpRequestData(int? timelineMicrosBase) : super(timelineMicrosBase);
```
kenzieschmoll, line 64: can we make this non-nullable?

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -173,30 +173,30 @@ abstract class HttpRequestData extends NetworkRequest {
   }
 
   /// All instant events logged to the timeline for this HTTP request.
-  List<HttpInstantEvent> get instantEvents;
+  List<HttpInstantEvent>? get instantEvents;
 
   /// A list of all cookies contained within the request headers.
   List<Cookie> get requestCookies;
```
kenzieschmoll, line 179: should this be nullable too?

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -173,30 +173,30 @@ abstract class HttpRequestData extends NetworkRequest {
...
 
   /// The request headers for the HTTP request.
-  Map<String, dynamic> get requestHeaders;
+  Map<String, dynamic>? get requestHeaders;
 
   /// A list of all cookies contained within the response headers.
   List<Cookie> get responseCookies;
```
kenzieschmoll, line 185: same comment here

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -173,30 +173,30 @@ abstract class HttpRequestData extends NetworkRequest {
...
 
   /// UTF-8 Decoded response body
-  String get responseBody;
+  String? get responseBody;
 
   /// Merges the information from another [HttpRequestData] into this instance.
   void merge(HttpRequestData data);
 
-  static List<Cookie> _parseCookies(List cookies) {
+  static List<Cookie> _parseCookies(List? cookies) {
```
kenzieschmoll, line 199: add type to List

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -314,24 +314,25 @@ class TimelineHttpRequestData extends HttpRequestData {
...
   @override
   bool get isValid =>
-      _startEvent != null && !_startEvent.name.contains('HTTP CLIENT response');
+      _startEvent != null &&
+      !_startEvent!.name!.contains('HTTP CLIENT response');
 
   /// A map of general information associated with an HTTP request.
   @override
-  Map<String, dynamic> get general {
+  Map<String, dynamic>? get general {
```
kenzieschmoll, line 322: looks like this may be able to be non-null?

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -562,7 +564,7 @@ class DartIOHttpRequestData extends HttpRequestData {
 
   /// All instant events logged to the timeline for this HTTP request.
   @override
-  List<HttpInstantEvent> get instantEvents {
+  List<HttpInstantEvent>? get instantEvents {
```
kenzieschmoll, line 567: I think this can be non-null. `_instantEvents` is assigned in the getter body

packages/devtools_app/lib/src/http/http_request_data.dart
polina-c: yes

packages/devtools_app/lib/src/http/http_request_data.dart
polina-c: Nope. Actually both can be non nullable because they can be empty lists.

packages/devtools_app/lib/src/http/http_request_data.dart
polina-c: It is empty list.

packages/devtools_app/lib/src/http/http_request_data.dart
polina-c: done

packages/devtools_app/lib/src/http/http_request_data.dart
polina-c: nope. We return null if !isValid

packages/devtools_app/lib/src/http/http_request_data.dart
polina-c: done

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -196,7 +196,7 @@ abstract class HttpRequestData extends NetworkRequest {
   /// Merges the information from another [HttpRequestData] into this instance.
   void merge(HttpRequestData data);
 
-  static List<Cookie> _parseCookies(List? cookies) {
+  static List<Cookie> _parseCookies(List<dynamic>? cookies) {
```
kenzieschmoll, line 199: should this be type Cookie instead of dynamic?

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -485,9 +485,10 @@ int _dartIoHttpRequestWrapperId = 0;
 
 class DartIOHttpRequestData extends HttpRequestData {
   DartIOHttpRequestData(
-    int? timelineMicrosBase,
+    int timelineMicrosBase,
     this._request,
   )   : wrapperId = _dartIoHttpRequestWrapperId++,
+        _instantEvents = [],
```
kenzieschmoll, line 491: setting this to the empty list here will prevent _instantEvents from being initialized in the `instantEvents` getter, so let's remove.

packages/devtools_app/lib/src/http/http_request_data.dart
polina-c: Nope. Downstream code code passes dynamic here.

kenzieschmoll (approved):

packages/devtools_app/lib/src/http/http_request_data.dart
polina-c: done

PR #3783, opened 2022-03-04
Migrate memory_service.dart to null safety.

polina-c:


packages/devtools_app/lib/src/screens/memory/memory_service.dart
```diff
@@ -65,80 +64,79 @@ class InboundReference extends Response {
...
   /* Parent class that allocated memory. */
   bool owningAllocatorIsAbstract,
   /* is owning class abstract */
 );
 
-ClassHeapDetailStats _searchClass(
+ClassHeapDetailStats? _searchClass(
   List<ClassHeapDetailStats> allClasses,
-  String className,
+  String? className,
```
kenzieschmoll, line 129: should this be non-nullable?

kenzieschmoll (approved):
one comment then lgtm

packages/devtools_app/lib/src/screens/memory/memory_service.dart
polina-c: nope. It gets classRef.name, that is nullable.

Issue #3556, opened 2021-12-23
Jump to current file option in program explorer


elliette:
A customer requested a way to jump to the current open file in the program explorer. I think ideally we should hook up the file opener with the program explorer, so when a user searches and opens a file with the file opener, it also selects that file in the program explorer.

elliette:
Looked into this a bit. We do select the file, we just don't scroll to it. I think all that needs to be done here is to scroll to a fiel when it's selected.

PR #3786, opened 2022-03-04
Automatic scrolling in the Program Explorer 

elliette:
Fixes https://github.com/flutter/devtools/issues/3556

Adds automatic scrolling to the selected file in the Program Explorer.

**Demo:**

program_explorer_scroll

### DIFF
--- a/packages/devtools_app/lib/src/primitives/trees.dart
+++ b/packages/devtools_app/lib/src/primitives/trees.dart
@@ -289,7 +289,7 @@ T? breadthFirstTraversal<T extends TreeNode<T>>(
   );
 }
 
-/// Traverses a tree in depth-first order.
+/// Traverses a tree in depth-first preorder order.
 ///
 /// [returnCondition] specifies the condition for which we should stop
 /// traversing the tree. For example, if we are calling this method to perform
@@ -331,6 +331,7 @@ T? _treeTraversal<T extends TreeNode<T>>(
       action(node);
     }
     if (exploreChildrenCondition == null || exploreChildrenCondition(node)) {
+      // For DFS, reverse the children to gaurantee preorder traversal.
       final children = bfs ? node.children : node.children.reversed;
       children.forEach(toVisit.add);
     }
--- a/packages/devtools_app/lib/src/primitives/trees.dart
+++ b/packages/devtools_app/lib/src/primitives/trees.dart
@@ -281,18 +281,12 @@ T? breadthFirstTraversal<T extends TreeNode<T>>(
   bool returnCondition(T node)?,
   void action(T node)?,
 }) {
-  final queue = Queue.of([root]);
-  while (queue.isNotEmpty) {
-    final node = queue.removeFirst();
-    if (returnCondition != null && returnCondition(node)) {
-      return node;
-    }
-    if (action != null) {
-      action(node);
-    }
-    node.children.forEach(queue.add);
-  }
-  return null;
+  return _treeTraversal(
+    root,
+    bfs: true,
+    returnCondition: returnCondition,
+    action: action,
+  );
 }
 
 /// Traverses a tree in depth-first order.
@@ -311,9 +305,25 @@ T? depthFirstTraversal<T extends TreeNode<T>>(
   void action(T node)?,
   bool exploreChildrenCondition(T node)?,
 }) {
-  final stack = Queue.of([root]);
-  while (stack.isNotEmpty) {
-    final node = stack.removeLast();
+  return _treeTraversal(
+    root,
+    bfs: false,
+    returnCondition: returnCondition,
+    action: action,
+    exploreChildrenCondition: exploreChildrenCondition,
+  );
+}
+
+T? _treeTraversal<T extends TreeNode<T>>(
+  T root, {
+  bool bfs = true,
+  bool returnCondition(T node)?,
+  void action(T node)?,
+  bool exploreChildrenCondition(T node)?,
+}) {
+  final toVisit = Queue.of([root]);
+  while (toVisit.isNotEmpty) {
+    final node = bfs ? toVisit.removeFirst() : toVisit.removeLast();
     if (returnCondition != null && returnCondition(node)) {
       return node;
     }
@@ -321,7 +331,8 @@ T? depthFirstTraversal<T extends TreeNode<T>>(
       action(node);
     }
     if (exploreChildrenCondition == null || exploreChildrenCondition(node)) {
-      node.children.reversed.forEach(stack.add);
+      final children = bfs ? node.children : node.children.reversed;
+      children.forEach(toVisit.add);
     }
   }
   return null;
--- a/packages/devtools_app/lib/src/primitives/trees.dart
+++ b/packages/devtools_app/lib/src/primitives/trees.dart
@@ -294,3 +294,35 @@ T? breadthFirstTraversal<T extends TreeNode<T>>(
   }
   return null;
 }
+
+/// Traverses a tree in depth-first order.
+///
+/// [returnCondition] specifies the condition for which we should stop
+/// traversing the tree. For example, if we are calling this method to perform
+/// DFS, [returnCondition] would specify when we have found the node we are
+/// searching for. [action] specifies an action that we will execute on each
+/// node. For example, if we need to traverse a tree and change a property on
+/// every single node, we would do this through the [action] param.
+/// [exploreChildrenCondition] specifies the condition for which we should
+/// explore the children of a node. By default, all children are explored.
+T? depthFirstTraversal<T extends TreeNode<T>>(
+  T root, {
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/primitives/trees.dart
```diff
@@ -281,16 +281,59 @@ T? breadthFirstTraversal<T extends TreeNode<T>>(
...
+    final node = bfs ? toVisit.removeFirst() : toVisit.removeLast();
     if (returnCondition != null && returnCondition(node)) {
       return node;
     }
     if (action != null) {
       action(node);
     }
-    node.children.forEach(queue.add);
+    if (exploreChildrenCondition == null || exploreChildrenCondition(node)) {
+      final children = bfs ? node.children : node.children.reversed;
```
kenzieschmoll, line 334: can you add a note to explain why we are reversing the children? Looks like it is to guarantee a preorder traversal? We may also want to add a comment that depthFirstTraversal will traverse in preorder

packages/devtools_app/lib/src/screens/debugger/program_explorer.dart
```diff
@@ -289,25 +291,57 @@ class _FileExplorer extends StatelessWidget {
   final Function(VMServiceObjectNode) onItemSelected;
   final Function(VMServiceObjectNode) onItemExpanded;
 
+  double get selectedNodeOffset =>
```
kenzieschmoll, line 294: let's move this getter to the widget state instead

packages/devtools_app/lib/src/screens/debugger/program_explorer.dart
```diff
@@ -289,25 +291,57 @@ class _FileExplorer extends StatelessWidget {
...
+}
+
+class _FileExplorerState extends State<_FileExplorer> with AutoDisposeMixin {
+  final ScrollController _scrollController = ScrollController();
+
+  @override
+  void initState() {
+    super.initState();
+    addAutoDisposeListener(
+        widget.controller.selectedNodeIndex, _maybeScrollToSelectedNode);
```
kenzieschmoll, line 309: nit: trailing comma

packages/devtools_app/lib/src/screens/debugger/program_explorer.dart
```diff
@@ -289,25 +291,57 @@ class _FileExplorer extends StatelessWidget {
...
+              onTap();
+            },
+          );
+        },
+      ),
     );
   }
+
+  void _maybeScrollToSelectedNode() {
+    if (_scrollController.offset != widget.selectedNodeOffset) {
```
kenzieschmoll, line 340: instead of checking for equality of offsets, can we check whether the selected node is already in view based on the offset and scroll extent? This will prevent spurious scrolls when the item is already visible.

packages/devtools_app/lib/src/screens/debugger/program_explorer.dart
```diff
@@ -289,25 +291,57 @@ class _FileExplorer extends StatelessWidget {
...
+      _scrollController.animateTo(widget.selectedNodeOffset,
+          duration: longDuration, curve: defaultCurve);
```
kenzieschmoll, line 342: nit: trailing comma

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
```diff
@@ -105,18 +109,34 @@ class ProgramExplorerController extends DisposableController
...
+      final matchingNode = depthFirstTraversal(
+        node,
+        returnCondition: matchingNodeCondition,
+        exploreChildrenCondition: (node) => node.isExpanded,
+        action: (_) => index++,
+      );
+      if (matchingNode != null) return index;
+    }
+    // If the node wasn't found, return 0.
+    return 0;
```
kenzieschmoll, line 137: maybe we should return -1, and then prevent scrolling the program explorer if the selected node index is -1? Otherwise, we'll jump the scroll view around because we couldn't find the index.

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
```diff
@@ -105,18 +109,34 @@ class ProgramExplorerController extends DisposableController
...
         selectNode(result);
         result.expandAscending();
+        _selectedNodeIndex.value = _calculateNodeIndex(searchCondition);
         return;
       }
     }
   }
 
+  int _calculateNodeIndex(
+      bool matchingNodeCondition(VMServiceObjectNode node)) {
```
kenzieschmoll, line 125: nit: tc

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
```diff
@@ -105,18 +109,34 @@ class ProgramExplorerController extends DisposableController
...
+    var index = 0;
+    for (final node in _rootObjectNodes.value) {
+      final matchingNode = depthFirstTraversal(
+        node,
+        returnCondition: matchingNodeCondition,
+        exploreChildrenCondition: (node) => node.isExpanded,
```
kenzieschmoll, line 131: what if the file uri we are looking for is in a collapsed part of the tree?

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
```diff
@@ -105,18 +109,34 @@ class ProgramExplorerController extends DisposableController
...
         result.expandAscending();
+        _selectedNodeIndex.value = _calculateNodeIndex(searchCondition);
         return;
       }
     }
   }
 
+  int _calculateNodeIndex(
+      bool matchingNodeCondition(VMServiceObjectNode node)) {
+    var index = 0;
```
kenzieschmoll, line 126: just to clarify, index is tracking the position of the node in the flat list representation of the tree, is that correct? if so can we add a comment

packages/devtools_app/lib/src/primitives/trees.dart
elliette: Yes, exactly! Added a comment and note

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
elliette: Yep, added a comment

packages/devtools_app/lib/src/screens/debugger/program_explorer.dart
elliette: Done

packages/devtools_app/lib/src/screens/debugger/program_explorer.dart
elliette: Done

packages/devtools_app/lib/src/screens/debugger/program_explorer.dart
elliette: Good idea! Switched to that

packages/devtools_app/lib/src/screens/debugger/program_explorer.dart
elliette: Done

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
elliette: Good point! Done

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
elliette: Done

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
elliette: We are currently only calling this method directly after expand-ascending up from the selected file, but this is a good point. I've added a parameter to this method to specify whether or not to search in collapsed parts of the tree, that way if it gets re-used we aren't accidentally skipping nodes in the search.

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
```diff
@@ -115,26 +115,34 @@ class ProgramExplorerController extends DisposableController
...
         return;
       }
     }
   }
 
-  int _calculateNodeIndex(
-      bool matchingNodeCondition(VMServiceObjectNode node)) {
+  int _calculateNodeIndex({
+    bool matchingNodeCondition(VMServiceObjectNode node),
+    bool includeCollapsedNodes,
```
kenzieschmoll, line 129: give this a default value. otherwise, the ternary check `includeCollapsedNodes ? ... : ...` will fail on a null value for `includeCollapsedNodes`

PR #3812, opened 2022-03-08
Example of improving performance by reducing # of widgets

jonahwilliams:
A draft example of our discussion from a few weeks back on improving performance. This change replaces the node widgets with a single widget representing a row, placing it in a single child scroll view. This doesn't completely work, as the item extents and sizing are not correctly calculated, but it does demonstrate the principles here:

@kenzieschmoll @goderbauer @dnfield 

What the unfinished chart currently looks like:

image

### DIFF
--- a/packages/devtools_app/lib/src/charts/flame_chart.dart
+++ b/packages/devtools_app/lib/src/charts/flame_chart.dart
@@ -753,30 +753,10 @@ class ScrollingFlameChartRowState<V extends FlameChartDataMixin<V>>
               chartWidth: widget.width,
               selected: selected,
               hovered: hovered,
+              extentDelegate: extentDelegate,
+              controller: scrollController,
             ),
           ),
-          // child: ExtentDelegateListView(
-          //   controller: scrollController,
-          //   scrollDirection: Axis.horizontal,
-          //   extentDelegate: extentDelegate,
-          //   childrenDelegate: SliverChildBuilderDelegate(
-          //     (context, index) {
-          //       final node = nodes[index];
-          //       return FlameChartNodeWidget(
-          //         index: index,
-          //         nodes: nodes,
-          //         zoom: widget.zoom,
-          //         startInset: widget.startInset,
-          //         chartWidth: widget.width,
-          //         selected: node.data == selected,
-          //         hovered: node.data == hovered,
-          //       );
-          //     },
-          //     childCount: nodes.length,
-          //     addRepaintBoundaries: false,
-          //     addAutomaticKeepAlives: false,
-          //   ),
-          // ),
         ),
       ),
     );
@@ -840,6 +820,8 @@ class FlameChartNodeRow extends LeafRenderObjectWidget {
     required this.chartWidth,
     required this.selected,
     required this.hovered,
+    required this.extentDelegate,
+    required this.controller,
   }) : super(key: key);
 
   final List<FlameChartNode> nodes;
@@ -854,6 +836,10 @@ class FlameChartNodeRow extends LeafRenderObjectWidget {
 
   final Object? hovered;
 
+  final _ScrollingFlameChartRowExtentDelegate extentDelegate;
+
+  final ScrollController controller;
+
   @override
   RenderObject createRenderObject(BuildContext context) {
     return _RenderFlameRow()
@@ -863,24 +849,39 @@ class FlameChartNodeRow extends LeafRenderObjectWidget {
       ..chartWidth = chartWidth
       ..selected = selected
       ..hovered = hovered
+      ..extentDelegate = extentDelegate
+      ..controller = controller
       ..colorScheme = Theme.of(context).colorScheme;
   }
 
   @override
   void updateRenderObject(BuildContext context, covariant _RenderFlameRow renderObject) {
-    _RenderFlameRow()
+    renderObject
       ..nodes = nodes
       ..zoom = zoom
       ..startInset = startInset
       ..chartWidth = chartWidth
       ..selected = selected
       ..hovered = hovered
+      ..extentDelegate = extentDelegate
+      ..controller = controller
       ..colorScheme = Theme.of(context).colorScheme;
   }
 }
 
 class _RenderFlameRow extends RenderBox {
    _RenderFlameRow();
+
+  _ScrollingFlameChartRowExtentDelegate? extentDelegate;
+  ScrollController? _controller;
+  set controller(ScrollController value) {
+    if (_controller == value) {
+      return;
+    }
+    _controller?.removeListener(markNeedsPaint);
+    _controller = value;
+    _controller?.addListener(markNeedsPaint);
+  }
 
   List<FlameChartNode> get nodes => _nodes;
   List<FlameChartNode> _nodes = [];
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -839,40 +850,141 @@ class FlameChartNodeWidget extends StatelessWidget {
...
+        canvas: context.canvas,
+        topLeft: offset + Offset(left + accumulated, 0),
+        selected: node == selected,
+        hovered: node == hovered,
+        searchMatch:  node.data.isSearchMatch,
+        activeSearchMatch: node.data.isActiveSearchMatch,
+        zoom: zoom,
+        colorScheme: colorScheme,
+      );
+      accumulated += 10; // TBD
```
jonahwilliams, line 984: Need to correctly compute width and spacing here

jonahwilliams:
Looks a little better:

image


Not quite sure the correct way to implement horizontal scroll with single child render object widget

packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -1073,48 +1204,23 @@ class FlameChartNode<T extends FlameChartDataMixin<T>> {
     // TODO(kenz): this is intended to improve performance but can probably be
     // improved. Perhaps we should still show a solid line and fade it out?
     if (zoomedWidth < 0.5) {
-      return SizedBox(width: zoomedWidth);
+      return;
```
kenzieschmoll, line 1207: is this one of the bigger performance grabs in this PR?

packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -1073,48 +1204,23 @@ class FlameChartNode<T extends FlameChartDataMixin<T>> {
...
     hovered = selectable ? hovered : false;
-
-    final node = Container(
-      key: hovered ? null : key,
-      width: zoomedWidth,
-      height: rect.height,
-      padding: const EdgeInsets.symmetric(horizontal: 6.0),
-      alignment: Alignment.centerLeft,
-      color: _backgroundColor(
+    final drawnRect = Rect.fromLTWH(topLeft.dx, topLeft.dy, rect.width * zoom, rect.height * zoom);
```
kenzieschmoll, line 1212: no need to scale height by zoom. zoom only affects horizontal size of the flame chart and nodes

packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -1035,7 +1149,21 @@ class FlameChartNode<T extends FlameChartDataMixin<T>> {
...
+        textAlign: TextAlign.left,
+        maxLines: 1,
+        fontStyle: FontStyle.normal,
+        textDirection: TextDirection.ltr,
+        fontSize: 10,
+      ),
+    );
+    paragraphBuilder.pushStyle(ui.TextStyle(color: _darkTextColor));
+    paragraphBuilder.addText(text);
+    paragraph = paragraphBuilder.build();
```
kenzieschmoll, line 1165: is it more performant to do this as part of the constructor than in draw? I'm assuming because draw is called more than once and the node should only be created once?

packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -839,40 +832,161 @@ class FlameChartNodeWidget extends StatelessWidget {
...
+      width,
+      constraints.maxHeight,
     );
   }
+
+  @override
+  void paint(PaintingContext context, Offset offset) {
+    var accumulated = 0.0;
+    var index = 0;
+    for (var node in nodes) {
```
kenzieschmoll, line 969: should we be painting only nodes in the viewport instead of all nodes in the row?

kenzieschmoll:
Added a few comments, mostly for my understanding. Did you happen to measure how big of a perf improvement this results in?

packages/devtools_app/lib/src/charts/flame_chart.dart
jonahwilliams: It might be even more performant to do it lazily, since we start out sort of zoomed out with most text invisible. @dnfield is the expert, but I believe paragraph builders are fairly expensive to construct so re-using them is a big win

packages/devtools_app/lib/src/charts/flame_chart.dart
jonahwilliams: Definitely :)

jonahwilliams:
I do need to make scrolling _actually_ work before we get realistic measurements. That said it feels better for zoom currently

packages/devtools_app/lib/src/charts/flame_chart.dart
jonahwilliams: Hard to say its just one thing, really we're cutting out a ton of simple configuration and replacing it with the canvas calls it boils down to. So the real work we are doing should be about the same but there is less stuff in-between (multiple layout/paint calls et cetera) - if that makes sense

jonahwilliams:
In terms of performance, currently its a difference between frames per second and seconds per frame

jonahwilliams:
Remaining work:

Figure out correct way to draw padding (left, right, bottom)
Update text to match old behavior
Ensure scrolling is consistent
update unit tests

jonahwilliams:
once I zoom in performance starts to tank. Looks like drawing text is really expensive in CK. ToWireType, followed by malloc dominates

image

jonahwilliams:
Reference canvaskit issue: https://github.com/flutter/flutter/issues/81224

dnfield:
In this case can you cache/reuse paragraph objects? A lot of those texts and styles are the same.

jonahwilliams:
I am

packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -1056,7 +1179,29 @@ class FlameChartNode<T extends FlameChartDataMixin<T>> {
 
   int sectionIndex;
 
-  Widget buildWidget({
+  Paragraph? _paragraph;
+  Paragraph _getParagraph() {
```
jonahwilliams, line 1183: @dnfield this is basically computed once per node, then laid out based on the size. I have a patch that caches the last layout width and only runs layout if it changes, but no big difference there

dnfield:
Alternatively, a UX change might look like making them color-coded so that the same labels get the same color, and just put the labels in a tooltip or some other visual area that tells you what they represent on hover/click or something....  

Alternatively make text fast on CanvasKit

jonahwilliams:
Yeah would be nice if text was just fast. Seems like some frames will hit a 10-12 ms layout due to needing to allocate too.

Alternatively, devtools could use a bitmap font/ text atlas and we could draw everything with drawAtlas

jonahwilliams:
its also possible I'm culling incorrectly 😓

jonahwilliams:
I'm updating this to use a double scrollable as discussed with @goderbauer . This should allow us to more easily tell what is in the viewport and simplify the implementation at the same time

jonahwilliams:
I think the update was, the single scrollable with a custom render object was getting a viewport that was far too big. I wonder if that is the case with the current nested scrollable approach.

The approach I have now works better, but there is a question of how much work it will be to port all the existing features over.

PR #3818, opened 2022-03-09
Migrate VM Tools screens to be null safe

bkonyi:
Fixes https://github.com/flutter/devtools/issues/3780

### DIFF
--- a/packages/devtools_app/lib/src/screens/vm_developer/isolate_statistics_view.dart
+++ b/packages/devtools_app/lib/src/screens/vm_developer/isolate_statistics_view.dart
@@ -2,8 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-// @dart=2.9
-
 import 'package:flutter/material.dart';
 import 'package:intl/intl.dart';
 import 'package:vm_service/vm_service.dart';
@@ -13,6 +11,7 @@ import '../../shared/common_widgets.dart';
 import '../../shared/split.dart';
 import '../../shared/table.dart';
 import '../../shared/table_data.dart';
+// ignore: import_of_legacy_library_into_null_safe
 import '../profiler/cpu_profiler.dart';
 import 'isolate_statistics_view_controller.dart';
 import 'vm_developer_common_widgets.dart';
@@ -114,23 +113,23 @@ class IsolateStatisticsViewBody extends StatelessWidget {
 ///   - Root library
 ///   - Isolate ID
 class GeneralIsolateStatisticsWidget extends StatelessWidget {
-  const GeneralIsolateStatisticsWidget({@required this.controller});
+  const GeneralIsolateStatisticsWidget({required this.controller});
 
   final IsolateStatisticsViewController controller;
 
-  String _startTime(Isolate isolate) {
+  String? _startTime(Isolate? isolate) {
     if (isolate == null) {
       return null;
     }
     final startedAtFormatter = DateFormat.yMMMMd().add_jms();
     return startedAtFormatter.format(
       DateTime.fromMillisecondsSinceEpoch(
-        isolate.startTime,
+        isolate.startTime!,
       ).toLocal(),
     );
   }
 
-  String _uptime(Isolate isolate) {
+  String? _uptime(Isolate? isolate) {
     if (isolate == null) {
       return null;
     }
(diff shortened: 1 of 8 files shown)


packages/devtools_app/lib/src/screens/vm_developer/isolate_statistics_view.dart
```diff
@@ -352,9 +353,9 @@ class _IsolatePortsWidgetState extends State<IsolatePortsWidget> {
   final id = _PortIDColumn();
   final name = _PortNameColumn();
 
-  final selectedPort = ValueNotifier<InstanceRef>(null);
+  final selectedPort = ValueNotifier<InstanceRef?>(null);
 
-  List<ColumnData<InstanceRef>> get columns => [
+  List<ColumnData<InstanceRef?>> get columns => [
```
kenzieschmoll, line 358: can InstanceRef be non-nullable

packages/devtools_app/lib/src/screens/vm_developer/isolate_statistics_view.dart
```diff
@@ -371,15 +372,15 @@ class _IsolatePortsWidgetState extends State<IsolatePortsWidget> {
               child: Split(
                 axis: Axis.horizontal,
                 children: [
-                  FlatTable<InstanceRef>(
+                  FlatTable<InstanceRef?>(
```
kenzieschmoll, line 375: same question here and for types in callbacks below

packages/devtools_app/lib/src/screens/vm_developer/isolate_statistics_view_controller.dart
```diff
@@ -51,42 +50,42 @@ class IsolateStatisticsViewController extends DisposableController
...
     _refreshing.value = true;
 
+    final isolateId = isolateRef.id!;
     // Retrieve updated isolate information and refresh the page.
-    _isolate = await _service.getIsolate(isolateRef.id);
+    _isolate = await _service.getIsolate(isolateId);
+    final isolate = _isolate!;
     _updateTagCounters(isolate);
-    _ports = (await _service.getPorts(_isolate.id)).ports;
+    _ports = (await _service.getPorts(isolateId)).ports!;
```
kenzieschmoll, line 63: just want to verify isolateRef.id == isolate.id? before this change we were getting the id from the Isolate, and this changes behavior to grab the id from the isolateRef

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_tools_screen.dart
```diff
@@ -75,14 +73,17 @@ class VMDeveloperToolsScreenBody extends StatefulWidget {
...
 
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
     final newController = Provider.of<VMDeveloperToolsController>(context);
-    if (newController == controller) return;
+    // Don't access `controller` if we haven't already initialized it.
+    if (initialized && newController == controller) return;
     controller = newController;
+    initialized = true;
```
kenzieschmoll, line 86: FYI @polina-c we should use this pattern elsewhere when we are initializing `controller` in widget state classes. (https://github.com/flutter/devtools/pull/3822#discussion_r823183816)

polina-c: So, the idea is to set it just once, right?
Good to know thank you. Thank you!

packages/devtools_app/lib/src/screens/vm_developer/isolate_statistics_view_controller.dart
bkonyi: Yes, isolate IDs are stable.

packages/devtools_app/lib/src/screens/vm_developer/isolate_statistics_view.dart
bkonyi: Yup, done.

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_tools_screen.dart
kenzieschmoll: We will still update `controller` if the dependencies changed (i.e. if the provided controller changed), but with this pattern we will not access `controller` to check equality before it is initialized, thus allowing it to be late and non-nullable.

PR #3822, opened 2022-03-09
Migrate memory_allocation_table_view.dart to null safety.

polina-c:


packages/devtools_app/lib/src/screens/memory/memory_allocation_table_view.dart
```diff
@@ -54,7 +54,7 @@ class AllocationTableViewState extends State<AllocationTableView>
     with AutoDisposeMixin {
   AllocationTableViewState() : super();
 
-  MemoryController controller;
+  MemoryController? controller;
```
kenzieschmoll, line 57: since controller is used so much, it might be worth creating a getter `MemoryController get controller => _controller!` and then renaming this to _controller. That way we don't have to null assert everywhere

kenzieschmoll (approved):
lgtm with one comment

packages/devtools_app/lib/src/screens/memory/memory_allocation_table_view.dart
kenzieschmoll: actually, let's use this pattern that Ben used here: https://github.com/flutter/devtools/pull/3818#discussion_r823225182

polina-c: applied https://github.com/flutter/devtools/pull/3818/files/f2c291440c7ea524bf2fa5a015e445595e57a1c3#r823225182

kenzieschmoll (approved):

polina-c:
tested

PR #3825, opened 2022-03-10
Changes required to get null-safety to work in G3

elliette:
Fixes for null-safety related warnings I see when rolling DevTools into g3.

### DIFF
--- a/packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
+++ b/packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
@@ -71,23 +71,20 @@ class ImportController {
     }
     previousImportTime = now;
 
-    final isDevToolsSnapshot =
-        json is Map<String, dynamic> && json[devToolsSnapshotKey] == true;
-    if (!isDevToolsSnapshot) {
+    final devToolsSnapshot = json;
+    if (devToolsSnapshot is Map<String, dynamic> &&
+        devToolsSnapshot[devToolsSnapshotKey] == true) {
+      final activeScreenId = devToolsSnapshot[activeScreenIdKey];
+      offlineController
+        ..enterOfflineMode()
+        ..offlineDataJson = devToolsSnapshot;
+      serviceManager.connectedApp =
+          OfflineConnectedApp.parse(devToolsSnapshot[connectedAppKey]);
+      _notifications.push(attemptingToImportMessage(activeScreenId));
+      _pushSnapshotScreenForImport(activeScreenId);
+    } else {
       _notifications.push(nonDevToolsFileMessage);
-      return;
     }
-
-    final devToolsSnapshot = json as Map<String, dynamic>;
-    // TODO(kenz): support imports for more than one screen at a time.
-    final activeScreenId = devToolsSnapshot[activeScreenIdKey];
-    offlineController
-      ..enterOfflineMode()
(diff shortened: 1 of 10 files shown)


packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
```diff
@@ -71,23 +71,20 @@ class ImportController {
     }
     previousImportTime = now;
 
-    final isDevToolsSnapshot =
-        json is Map<String, dynamic> && json[devToolsSnapshotKey] == true;
-    if (!isDevToolsSnapshot) {
+    final devToolsSnapshot = json;
```
kenzieschmoll, line 74: this change is fine to land externally, but the other changes to remove the null assert operators won't pass on the bots externally. DevTools is built from flutter 2.10.0-0.3.pre externally, which is much older than what is in google3. We can't move to the latest beta because there is a bug in that version of flutter that prevents apps from being able to connect to DevTools. So until then, we'll have to manually remove the `!` operators in google3 only.

packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
```diff
@@ -71,23 +71,20 @@ class ImportController {
...
+      final activeScreenId = devToolsSnapshot[activeScreenIdKey];
+      offlineController
+        ..enterOfflineMode()
+        ..offlineDataJson = devToolsSnapshot;
+      serviceManager.connectedApp =
+          OfflineConnectedApp.parse(devToolsSnapshot[connectedAppKey]);
+      _notifications.push(attemptingToImportMessage(activeScreenId));
+      _pushSnapshotScreenForImport(activeScreenId);
+    } else {
       _notifications.push(nonDevToolsFileMessage);
```
polina-c, line 86: I like two things in original code, that are lost in new code:

1. It follows 'early out' (or 'return early') pattern by returning immediately when it is clear the path is non-happy.
More materials on this:
https://szymonkrajewski.pl/why-should-you-return-early/
https://g3doc.corp.google.com/company/teams/gpay/flutter/development/code_style.md#prefer-early-out-style

2. It defines the variable isDevToolsSnapshot, that described the meaning of the logical expression with its name.

From my perspective the original code is more readable, than the new one.
What do you think?

elliette: Oh yes, I agree. This change wasn't about readability, but about getting rid of the "unnecessary cast" warning in g3. This way both the internal and external linters are happy without suppressing any warnings. The analyzer doesn't seem smart enough to know the type of `devToolsSnapshot` with the early out pattern.

packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
elliette: Oh good to know, I didn't realize that! That explains why this is failing.

packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
elliette: Just to clarify, explicitly setting the cast is what was there before, right?

elliette: Ok suppressing works. I'll just add that change in g3!

PR #3830, opened 2022-03-10
Migrate memory_tracker_model to null safety.

polina-c:


packages/devtools_app/lib/src/screens/memory/memory_tracker_model.dart
```diff
@@ -61,7 +62,7 @@ class TrackerCall extends Tracker {
   TrackerCall(String entry) : super._internal(name: entry);
 
   @override
-  String toString() => name;
+  String toString() => name ?? '';
```
kenzieschmoll, line 65: if entry is non nullable we should be able to null-assert here instead of providing a default value

packages/devtools_app/lib/src/screens/memory/memory_tracker_model.dart
```diff
@@ -85,14 +86,14 @@ class _TrackerClassColumn extends TreeColumnData<Tracker> {
 
   @override
   int compare(Tracker a, Tracker b) {
-    Comparable valueA = getValue(a);
-    Comparable valueB = getValue(b);
+    Comparable? valueA = getValue(a);
+    Comparable? valueB = getValue(b);
```
kenzieschmoll, line 90: these can be non-null

kenzieschmoll (approved):
lgtm with a couple comments

packages/devtools_app/lib/src/screens/memory/memory_tracker_model.dart
polina-c: It is nullable and it is assigned to null in TrackerAllocation constructor.

packages/devtools_app/lib/src/screens/memory/memory_tracker_model.dart
polina-c: done

packages/devtools_app/lib/src/screens/memory/memory_tracker_model.dart
kenzieschmoll: This can still be null-asserted in the toString method because TrackerCall takes a non-nullable parameter `entry`, which it assigns to `Tracker.name`

polina-c:
tested

PR #3832, opened 2022-03-10
IDE theme can be `null`

elliette:
If the IDE theme global hasn't been set, then `ideTheme` is `null`. 

This was causing errors runtime errors when rolling DevTools into g3. I think it might make sense for the other global values to be nullable as well.


kenzieschmoll (commented):
this can be non-nullable. the setGlobal statement happens in main before we run the DevTools app, so it should always be set

PR #3833, opened 2022-03-10
Split logging_screen to make null safety migration easier.

polina-c:


kenzieschmoll (approved):

kenzieschmoll:
Maybe file a bug or add a task to the meta null safety issue to merge some of these classes back into a single file after you migrate

Issue #3711, opened 2022-02-18
Sent event to dwds from Inspector window


annagrin:
Currently DevTools sends an event to dwds from Debugger window. Tools that start DevTools use this information to measure startup times of DevTools.

https://github.com/flutter/devtools/blob/a616832995d13410a4078b506c91927865ff6904/packages/devtools_app/lib/src/debugger/debugger_screen.dart#L158
For flutter apps, however, the first tab that opens is the Inspector, which means the time to start DevTools is measured incorrectly.

Related: https://github.com/dart-lang/webdev/issues/1517

jacob314:
Fyi @elliette

PR #3834, opened 2022-03-11
Send pageReady event to dwds from Inspector page

annagrin:
Send `pageReady` event from the Inspector page, similar to what is already
done from the Debugger page.

Closes: https://github.com/flutter/devtools/issues/3711

### DIFF
--- a/packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
+++ b/packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
@@ -23,6 +23,7 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:vm_service/vm_service.dart';
 
+import '../../analytics/constants.dart' as analytics_constants;
 import '../../config_specific/logger/logger.dart';
 import '../../config_specific/url/url.dart';
 import '../../primitives/auto_dispose.dart';
@@ -412,6 +413,12 @@ class InspectorController extends DisposableController
     }
 
     if (flutterAppFrameReady) {
+      // TODO: measure and send DevTools pageReady analytics:
+      // https://github.com/flutter/devtools/issues/3879
+      await serviceManager.sendDwdsEvent(
+        screen: InspectorScreen.id,
+        action: analytics_constants.pageReady,
+      );
       _rootDirectories = await inspectorService.inferPubRootDirectoryIfNeeded();
       if (_disposed) return;
       // We need to start by querying the inspector service to find out the
--- a/packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
+++ b/packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
@@ -181,6 +181,10 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
                   .hasServiceExtension(
                       extensions.toggleSelectWidgetMode.extension),
               builder: (_, selectModeSupported, __) {
+                serviceManager.sendDwdsEvent(
+                  screen: InspectorScreen.id,
+                  action: analytics_constants.pageReady,
+                );
                 return ServiceExtensionButtonGroup(
                   extensions: [
                     selectModeSupported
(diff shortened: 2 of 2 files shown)


kenzieschmoll:
A couple comments here:
1) the linked issue looks like we want to measure the time it takes to start the inspector, but we aren't measuring anything in the DevTools analytics for that metric in this PR - only sending a dwds event. We should probably do both, just like we do for the debugger.
2) Do we want to measure the time until the widget tree is loaded or time until the inspector screen is displayed? I think we should measure time it takes until the widget tree is displayed, because the inspector screen without the widget tree is not useful for users. Right now the event is being sent from a value listenable builder related to the toggle select widget mode button, so we'll want to send the event in a different spot. I'm looking currently for the best place to do this.
3) Does it matter if the DWDS event is sent more than once? What if the event is sent from the Inspector, and then the user goes to the Debugger page, which sends another dwds pageReady event. Should we guard against this?

kenzieschmoll:
After digging a little, I think this may be the best place to mark end of startup time or "pageReady": at the end of this if block https://github.com/flutter/devtools/blob/ecba4cd9dbdbe2e15ec0d299d24b8f23d9a151ab/packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart#L414. 

@jacob314 what do you think?

annagrin:
> A couple comments here:
> 
> 1. the linked issue looks like we want to measure the time it takes to start the inspector, but we aren't measuring anything in the DevTools analytics for that metric in this PR - only sending a dwds event. We should probably do both, just like we do for the debugger.

The time measurement you are suggesting is done for DevTools analytics, which is separate from dwds analytics. For the purpose of the bug fix it is not needed, but please let me know if you would like me to add this as well. Dwds tracks time starting from the user clicking on the dart extension button, and ending with any page ready event from the DevTools, and it already has the logic to record the IPL event just once.

> 2. Do we want to measure the time until the widget tree is loaded or time until the inspector screen is displayed? I think we should measure time it takes until the widget tree is displayed, because the inspector screen without the widget tree is not useful for users. Right now the event is being sent from a value listenable builder related to the toggle select widget mode button, so we'll want to send the event in a different spot. I'm looking currently for the best place to do this.

We are measuring the time  it takes for the DevTools to be available to the user after the user initiated the debugging session. Not sure what is the best way to define it - I think it is ok not to include some heavy asynchronous operations in that, if the user can already do something. I think there is a lot of expression evaluation happening for the inspector tabs that we probably don't want to consider a part of "IPL time". If we just want to measure the time until time until the inspector screen is displayed, is the event sent in the correct spot?

> 3. Does it matter if the DWDS event is sent more than once? What if the event is sent from the Inspector, and then the user goes to the Debugger page, which sends another dwds pageReady event. Should we guard against this?

It is ok to send the event many times, dwds will ignore all such events after the first `pageReady` coming from either debugger  or inspector.

kenzieschmoll:
> For the purpose of the bug fix it is not needed, but please let me know if you would like me to add this as well.

We can leave it out of this PR then, but would you mind filing an issue and adding a TODO in the code that links to it? Thanks.

> I think it is ok not to include some heavy asynchronous operations in that, if the user can already do something. I think there is a lot of expression evaluation happening for the inspector tabs that we probably don't want to consider a part of "IPL time". 

[1] I think we have gotten complaints in the past that the widget tree is slow to load in g3, and I think slow to load for flutter web specifically. I'll defer to your judgement on what you are interested in measuring for DWDS, but for the devtools analytics we will add, we'll probably end the timer when the widget tree loads. 

> If we just want to measure the time until the inspector screen is displayed, is the event sent in the correct spot?

I think there may be a better place to put this event. As for where to put the event, it will depend on your response to [1] above.

annagrin:
@kenzieschmoll 

> We can leave it out of this PR then, but would you mind filing an issue and adding a TODO in the code that links to it? Thanks.

Sure, will do. Opened a separate bug for DevTools analytics:  https://github.com/flutter/devtools/issues/3879

> [1] I think we have gotten complaints in the past that the widget tree is slow to load in g3, and I think slow to load for flutter web specifically. I'll defer to your judgement on what you are interested in measuring for DWDS, but for the devtools analytics we will add, we'll probably end the timer when the widget tree loads.

I'll measure the whole time until the widget tree is loaded as well, as we discussed offline.

kenzieschmoll (approved):

kenzieschmoll:
Do we also want to add an event somewhere in the isolate connection code that would be triggered at the same time for any devtools launch (independent of whether the first page is the inspector or the debugger)?

PR #3835, opened 2022-03-11
Migrate _message_column.dart to null safety.

polina-c:


packages/devtools_app/lib/src/screens/logging/_message_column.dart
```diff
@@ -24,20 +22,20 @@ class MessageColumn extends ColumnData<LogData>
   bool get supportsSorting => false;
 
   @override
-  String getValue(LogData dataObject) =>
+  String? getValue(LogData dataObject) =>
```
kenzieschmoll, line 25: make this return a non-nullable string. We may need to change the return type of `ColumnData.getValue` to `Object` so that it is non-nullable by default. The table classes and other column classes use `getDisplayValue` without null checking, and `getDisplayValue` uses `getValue` without null checking, assuming the value returned by `getValue` is non-null. So if we return a nullable value from `getValue`, this will likely cause some exceptions to be thrown from tables.

polina-c:
blocked by memory_analyzer migration

polina-c:
blocked by https://github.com/flutter/devtools/pull/3847

PR #3838, opened 2022-03-11
Migrate _log_details.dart to null safety.

polina-c:


packages/devtools_app/lib/src/screens/logging/_log_details.dart
```diff
@@ -25,8 +23,8 @@ class LogDetails extends StatefulWidget {
 
 class _LogDetailsState extends State<LogDetails>
     with SingleTickerProviderStateMixin {
-  String _lastDetails;
-  ScrollController scrollController;
+  String? _lastDetails;
+  ScrollController? scrollController;
```
kenzieschmoll, line 27: late final

packages/devtools_app/lib/src/screens/logging/_log_details.dart
```diff
@@ -45,12 +43,12 @@ class _LogDetailsState extends State<LogDetails>
...
   Future<void> _computeLogDetails() async {
     if (widget.log?.needsComputing ?? false) {
-      await widget.log.compute();
+      await widget.log!.compute();
       setState(() {});
     }
   }
 
-  bool showSimple(LogData log) => log != null && !log.needsComputing;
+  bool showSimple(LogData log) => !log.needsComputing;
```
kenzieschmoll, line 51: looks like this method can be deleted. It is not used anywhere

packages/devtools_app/lib/src/screens/logging/_log_details.dart
```diff
@@ -89,7 +87,11 @@ class _LogDetailsState extends State<LogDetails>
           needsTopBorder: false,
           rightActions: [
             CopyToClipboardControl(
-              dataProvider: disabled ? null : () => log?.prettyPrinted,
+              dataProvider: disabled
+                  ? null
+                  // The parenthesis are actually needed.
+                  // ignore: unnecessary_parenthesis
+                  : (() => log?.prettyPrinted!) as String Function()?,
```
kenzieschmoll, line 94: you should be able to get rid of the cast if you write:
`dataProvider: disabled ? null : log?.prettyPrinted`

and then change prettyPrinted from a getter to a function.

packages/devtools_app/lib/src/screens/logging/_log_details.dart
polina-c: done

packages/devtools_app/lib/src/screens/logging/_log_details.dart
polina-c: done

packages/devtools_app/lib/src/screens/logging/_log_details.dart
polina-c: prettyPrinted can return null, but we want function that  returns non-nullable string for dataProvider. 
Do we want to change dataProvider type to return nullable string?

Still converted prettyPrinted to function as it is too complicated to be getter.

kenzieschmoll: Yeah let's change ClipboardDataProvider to return a nullable string. It looks like down the line, ClipboardData.text is nullable anyway, so this should be fine. We might want to ensure we give a default string value for `data.text` in `Clipboard.setData`

polina-c: done

packages/devtools_app/lib/src/screens/logging/_log_details.dart
```diff
@@ -85,11 +85,7 @@ class _LogDetailsState extends State<LogDetails>
           needsTopBorder: false,
           rightActions: [
             CopyToClipboardControl(
-              dataProvider: disabled
-                  ? null
-                  // The parenthesis are actually needed.
-                  // ignore: unnecessary_parenthesis
-                  : (() => log?.prettyPrinted()!) as String Function()?,
+              dataProvider: disabled ? null : () => log?.prettyPrinted(),
```
kenzieschmoll, line 88: can you write `log?.prettyPrinted` instead of `() => log?.prettyPrinted()`

polina-c: done

PR #3842, opened 2022-03-14
Fix null-safety issue.

polina-c:
Fix compilation error in master.

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/memory_filter.dart
+++ b/packages/devtools_app/lib/src/screens/memory/memory_filter.dart
@@ -186,7 +186,8 @@ class SnapshotFilterState extends State<SnapshotFilterDialog>
       // Don't include external and filtered these are a composite and can't be filtered out.
       if (library.name != externalLibraryName &&
           library.name != filteredLibrariesName) {
-        addLibrary(library.name);
+        assert(library.name != null);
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/screens/memory/memory_filter.dart
```diff
@@ -186,7 +186,8 @@ class SnapshotFilterState extends State<SnapshotFilterDialog>
       // Don't include external and filtered these are a composite and can't be filtered out.
       if (library.name != externalLibraryName &&
           library.name != filteredLibrariesName) {
-        addLibrary(library.name);
+        assert(library.name != null);
```
bkonyi, line 189: Is this assertion necessary?

bkonyi (approved):

packages/devtools_app/lib/src/screens/memory/memory_filter.dart
polina-c: if name is null, next line will fail. We prefer assertion because we assume name is not null here.

bkonyi: `library.name!` is effectively `assert(library.name != null)` though. It's not incorrect to assert as well, but it's a redundant check.

PR #3852, opened 2022-03-15
Migrate memory_instance_tree_view.dart to null safety.

polina-c:


packages/devtools_app/lib/src/screens/memory/memory_instance_tree_view.dart
```diff
@@ -45,42 +44,43 @@ class InstanceTreeViewState extends State<InstanceTreeView>
...
+    return _controller.instanceFieldsTreeTable!;
   }
 }
 
 class _FieldTypeColumn extends TreeColumnData<FieldReference> {
   _FieldTypeColumn() : super('Type');
 
   @override
-  dynamic getValue(FieldReference dataObject) =>
+  Object? getValue(FieldReference dataObject) =>
```
kenzieschmoll, line 83: non-nullable. Also should this be returning a String? it seems most other columns return a string value for their `getValue` method

packages/devtools_app/lib/src/screens/memory/memory_instance_tree_view.dart
```diff
@@ -94,8 +94,8 @@ class _FieldTypeColumn extends TreeColumnData<FieldReference> {
 
   @override
   int compare(FieldReference a, FieldReference b) {
-    final Comparable valueA = getValue(a);
-    final Comparable valueB = getValue(b);
+    final Comparable valueA = getValue(a) as Comparable<Object>;
+    final Comparable valueB = getValue(b) as Comparable<Object>;
```
kenzieschmoll, line 98: these casts should be able to be removed once getValue return type is changed to string. here and below.

packages/devtools_app/lib/src/screens/memory/memory_instance_tree_view.dart
```diff
@@ -111,7 +111,7 @@ class _FieldNameColumn extends ColumnData<FieldReference> {
         );
 
   @override
-  dynamic getValue(FieldReference dataObject) =>
+  Object? getValue(FieldReference dataObject) =>
```
kenzieschmoll, line 114: same comment: `String`, here and below.

packages/devtools_app/lib/src/screens/memory/memory_instance_tree_view.dart
polina-c: done

packages/devtools_app/lib/src/screens/memory/memory_instance_tree_view.dart
polina-c: done

packages/devtools_app/lib/src/screens/memory/memory_instance_tree_view.dart
polina-c: done

kenzieschmoll (approved):

PR #3855, opened 2022-03-15
Migrate inspector ui primitives to null safety.

polina-c:


packages/devtools_app/lib/src/ui/icons.dart
```diff
@@ -130,7 +130,7 @@ class CustomIconMaker {
     final widgetTheme = WidgetTheme.fromName(name);
     if (widgetTheme.iconAsset != null) {
       return iconCache.putIfAbsent(name, () {
-        return AssetImageIcon(asset: widgetTheme.iconAsset);
+        return AssetImageIcon(asset: widgetTheme.iconAsset!);
```
kenzieschmoll, line 133: hoist `widgetTheme.iconAsset` into a local var before the `if (widgetTheme.iconAsset != null)`, then use it both places to avoid a second null check

kenzieschmoll (approved):
lgtm with one comment

packages/devtools_app/lib/src/ui/icons.dart
polina-c: done

polina-c:
tested

PR #3871, opened 2022-03-16
Migrate instance_viewer and integration tests to null safety.

polina-c:


packages/devtools_app/test/instance_viewer/instance_viewer_test.dart
```diff
@@ -129,7 +129,7 @@ void main() {
         (tester) async {
       const objPath = InstancePath.fromInstanceId('obj');
 
-      InstancePath pathForProperty(String name) {
+      InstancePath? pathForProperty(String name) {
```
kenzieschmoll, line 132: why does this need to be nullable

packages/devtools_app/test/integration_tests/app.dart
```diff
@@ -64,12 +62,12 @@ class ConnectDialogManager {
 
   Future<bool> isVisible() async {
     final AppResponse response =
-        await tools.tabInstance.send('connectDialog.isVisible');
+        await tools.tabInstance!.send('connectDialog.isVisible');
```
kenzieschmoll, line 65: this should be non-nullable. same comment below

packages/devtools_app/test/integration_tests/debugger.dart
```diff
@@ -324,77 +323,77 @@ class DebuggingManager {
   final DevtoolsManager tools;
 
   Future<void> resume() async {
-    await tools.tabInstance.send('debugger.resume');
+    await tools.tabInstance!.send('debugger.resume');
```
kenzieschmoll, line 326: non nullable, here and elsewhere

packages/devtools_app/test/integration_tests/integration.dart
```diff
@@ -16,40 +14,40 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart'
...
-WebBuildFixture webBuildFixture;
-BrowserManager browserManager;
+late WebBuildFixture webBuildFixture;
+late BrowserManager browserManager;
 
 class DevtoolsManager {
   DevtoolsManager(this.tabInstance, this.baseUri);
 
-  final BrowserTabInstance tabInstance;
+  final BrowserTabInstance? tabInstance;
```
kenzieschmoll, line 23: non nullable

packages/devtools_app/test/integration_tests/integration.dart
```diff
@@ -16,40 +14,40 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart'
...
   }
 
   Future<void> switchPage(String page) async {
-    await tabInstance.send('switchPage', page);
+    await tabInstance!.send('switchPage', page);
   }
 
-  Future<String> currentPageId() async {
-    final AppResponse response = await tabInstance.send('currentPageId');
+  Future<String?> currentPageId() async {
```
kenzieschmoll, line 49: The other methods that return `response.result` are cast as `String?` - does this return need that also? if not, do the other casts actually need to be there?

packages/devtools_app/test/integration_tests/integration.dart
```diff
@@ -58,13 +56,14 @@ class BrowserManager {
...
-    final Chrome chrome = Chrome.locate();
+    final Chrome? chrome = Chrome.locate();
     if (chrome == null) {
       throw 'unable to locate Chrome';
     }
 
     final ChromeProcess chromeProcess = await chrome.start();
-    final ChromeTab tab = await chromeProcess.getFirstTab();
+    final ChromeTab tab =
+        await (chromeProcess.getFirstTab() as FutureOr<ChromeTab>);
```
kenzieschmoll, line 66: couple things - 1) why do we need the cast, and 2) `getFirstTab` returns `Future<ChromeTab?>`, so doesn't this need to be nullable or null-asserted? same question below.

packages/devtools_app/test/integration_tests/integration.dart
```diff
@@ -198,11 +197,11 @@ class BrowserTabInstance {
     if (message.containsKey('id')) {
       // handle a response: {id: 1}
       final AppResponse response = AppResponse(message);
-      final Completer<AppResponse> completer = _completers.remove(response.id);
+      final Completer<AppResponse>? completer = _completers.remove(response.id);
```
kenzieschmoll, line 200: null assert this var here instead of doing so twice below

packages/devtools_app/test/integration_tests/integration.dart
```diff
@@ -246,11 +245,11 @@ class AppResponse {
 class AppError {
   AppError(this.json);
 
-  final Map<dynamic, dynamic> json;
+  final Map<dynamic, dynamic>? json;
```
kenzieschmoll, line 248: make non nullable if we are null asserting unconditionally elsewhere

packages/devtools_app/test/instance_viewer/instance_viewer_test.dart
polina-c: fixed

packages/devtools_app/test/integration_tests/integration.dart
polina-c: done

packages/devtools_app/test/integration_tests/integration.dart
polina-c: Not sure why we do it differently. Added casting here and in other places for consistency.

packages/devtools_app/test/integration_tests/integration.dart
polina-c: done

packages/devtools_app/test/integration_tests/integration.dart
polina-c: done

kenzieschmoll (approved):

PR #3873, opened 2022-03-16
Migrate test data and infra to null safety.

polina-c:


packages/devtools_app/test/test_infra/flutter_test_driver.dart
```diff
@@ -174,23 +172,25 @@ abstract class FlutterTestDriver {
         message: 'Isolate did not pause');
   }
 
-  Future<Isolate> resume({String step, bool wait = true}) async {
+  Future<Isolate?> resume({String? step, bool wait = true}) async {
     debugPrint('Sending resume ($step)');
     await _timeoutWithMessages<dynamic>(
-        () async => vmService.resume(await getFlutterIsolateId(), step: step),
+        () async => vmService!.resume(
+            await (getFlutterIsolateId() as FutureOr<String>),
```
kenzieschmoll, line 179: remove cast

packages/devtools_app/test/test_infra/flutter_test_driver.dart
```diff
@@ -224,7 +224,7 @@ abstract class FlutterTestDriver {
   }
 
   Future<T> _timeoutWithMessages<T>(Future<T> Function() f,
-      {Duration timeout, String message}) {
+      {Duration? timeout, String? message}) {
```
kenzieschmoll, line 227: nit: trailing comma

packages/devtools_app/test/test_infra/flutter_test_driver.dart
```diff
@@ -261,17 +261,17 @@ abstract class FlutterTestDriver {
...
 
   Future<void> run({
-    @required String flutterExecutable,
+    required String flutterExecutable,
     FlutterRunConfiguration runConfig = const FlutterRunConfiguration(),
-    File pidFile,
+    File? pidFile,
   }) async {
-    final args = <String>[
+    final args = <String?>[
```
kenzieschmoll, line 274: non nullable

packages/devtools_app/test/test_infra/flutter_test_driver.dart
```diff
@@ -373,8 +373,8 @@ class FlutterRunTestDriver extends FlutterTestDriver {
       // to hit breakpoints, etc.
       await waitForPause();
       if (runConfig.pauseOnExceptions) {
-        await vmService.setIsolatePauseMode(
-          await getFlutterIsolateId(),
+        await vmService!.setIsolatePauseMode(
+          await (getFlutterIsolateId() as FutureOr<String>),
```
kenzieschmoll, line 377: remove cast

packages/devtools_app/test/test_infra/flutter_test_environment.dart
```diff
@@ -39,12 +37,12 @@ class FlutterTestEnvironment {
...
   }
 
-  FlutterRunConfiguration _runConfig;
-  FlutterRunConfiguration get runConfig => _runConfig;
-  FlutterRunTestDriver _flutter;
-  FlutterRunTestDriver get flutter => _flutter;
-  VmServiceWrapper _service;
-  VmServiceWrapper get service => _service;
+  FlutterRunConfiguration? _runConfig;
+  FlutterRunConfiguration? get runConfig => _runConfig;
```
kenzieschmoll, line 41: non nullable

packages/devtools_app/test/test_infra/flutter_test_environment.dart
```diff
@@ -39,12 +37,12 @@ class FlutterTestEnvironment {
...
+  FlutterRunTestDriver? _flutter;
+  FlutterRunTestDriver? get flutter => _flutter;
+  VmServiceWrapper? _service;
```
kenzieschmoll, line 44: _service can be late and service getter can be non-nullable

packages/devtools_app/test/test_infra/flutter_test_environment.dart
```diff
@@ -108,13 +106,14 @@ class FlutterTestEnvironment {
 
       _needsSetup = false;
 
-      _flutter = _flutterDriverFactory(Directory(testAppDirectory));
-      await _flutter.run(
+      _flutter = _flutterDriverFactory(Directory(testAppDirectory))
+          as FlutterRunTestDriver?;
```
kenzieschmoll, line 110: why does this need a nullable cast? _flutterDriverFactory should return a non-nullable value

packages/devtools_app/test/test_infra/flutter_test_driver.dart
```diff
@@ -114,53 +112,53 @@ abstract class FlutterTestDriver {
...
-    return proc.exitCode;
+    Process.killPid(procPid!, ProcessSignal.sigkill);
+    return proc!.exitCode;
   }
 
-  String flutterIsolateId;
+  String? flutterIsolateId;
 
-  Future<String> getFlutterIsolateId() async {
+  Future<String?> getFlutterIsolateId() async {
```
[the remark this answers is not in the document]
bkonyi: `vm.isolates` will never be null in practice, but it could be empty so you'll need a check to make sure there's an element before calling `.first`. Also, you can always assume an ID is provided (it's only nullable since tests with mocks like to omit them).

packages/devtools_app/test/test_infra/flutter_test_driver.dart
```diff
@@ -345,25 +345,25 @@ class FlutterRunTestDriver extends FlutterTestDriver {
 
       // Map to WS URI.
       vmServiceWsUri =
-          convertToWebSocketUrl(serviceProtocolUrl: vmServiceWsUri);
+          convertToWebSocketUrl(serviceProtocolUrl: vmServiceWsUri!);
```
bkonyi, line 348: Copy `vmService` and `vmServiceWsUri` into locals to avoid the extra null assertions.

packages/devtools_app/test/test_infra/flutter_test_driver.dart
```diff
@@ -414,12 +414,12 @@ class FlutterRunTestDriver extends FlutterTestDriver {
   Future<int> detach() async {
     if (vmService != null) {
       debugPrint('Closing VM service');
-      await vmService.dispose();
+      await vmService!.dispose();
```
bkonyi, line 417: Assign `vmService` to a local so we only do one null check.

packages/devtools_app/test/test_infra/flutter_test_driver.dart
```diff
@@ -433,18 +433,18 @@ class FlutterRunTestDriver extends FlutterTestDriver {
...
     debugPrint('Waiting for process to end');
-    return proc.exitCode.timeout(quitTimeout, onTimeout: killGracefully);
+    return proc!.exitCode.timeout(quitTimeout, onTimeout: killGracefully);
   }
 
   Future<int> stop() async {
     if (vmService != null) {
       debugPrint('Closing VM service');
-      await vmService.dispose();
+      await vmService!.dispose();
```
bkonyi, line 442: Same as above.

packages/devtools_app/test/test_infra/flutter_test_environment.dart
bkonyi: `_flutter` can probably be marked as a late field instead of being marked nullable (assuming `setupEnvironment` is always invoked at some point).

bkonyi (commented):
Kenzie covered most of the comments I have, but I found a few more places we can remove null checks.

packages/devtools_app/test/test_infra/flutter_test_driver.dart
```diff
@@ -29,32 +27,32 @@ const Duration appStartTimeout = Duration(seconds: 120);
...
 abstract class FlutterTestDriver {
-  FlutterTestDriver(this.projectFolder, {String logPrefix})
+  FlutterTestDriver(this.projectFolder, {String? logPrefix})
       : _logPrefix = logPrefix != null ? '$logPrefix: ' : '';
 
   final Directory projectFolder;
   final String _logPrefix;
-  Process proc;
-  int procPid;
+  Process? proc;
```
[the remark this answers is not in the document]
polina-c: I am confused by this below:
```
    if (proc != null) {
      debugPrint('Waiting for process to end');
      return proc!.exitCode.timeout(quitTimeout, onTimeout: killGracefully);
    }
```

proc is expected to become null at some point, but I do not see the code that assignes it to null. Am I missing something?

packages/devtools_app/test/test_infra/flutter_test_driver.dart
```diff
@@ -29,32 +27,32 @@ const Duration appStartTimeout = Duration(seconds: 120);
...
       StreamController<String>.broadcast();
   final StreamController<String> stderrController =
       StreamController<String>.broadcast();
   final StreamController<String> _allMessages =
       StreamController<String>.broadcast();
   final StringBuffer errorBuffer = StringBuffer();
-  String lastResponse;
-  Uri vmServiceWsUri;
+  String? lastResponse;
+  Uri? vmServiceWsUri;
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/test/test_infra/flutter_test_driver.dart
```diff
@@ -68,10 +66,10 @@ abstract class FlutterTestDriver {
   }
 
   Future<void> setupProcess(
-    List<String> args, {
-    @required String flutterExecutable,
+    List<String?> args, {
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/test/test_infra/flutter_test_driver.dart
```diff
@@ -83,7 +81,7 @@ abstract class FlutterTestDriver {
 
     proc = await Process.start(
       flutterExecutable,
-      _args,
+      _args as List<String>,
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/test/test_infra/flutter_test_driver.dart
```diff
@@ -114,53 +112,53 @@ abstract class FlutterTestDriver {
       return -1;
     }
     debugPrint('Sending SIGTERM to $procPid..');
-    Process.killPid(procPid);
-    return proc.exitCode.timeout(quitTimeout, onTimeout: _killForcefully);
+    Process.killPid(procPid!);
```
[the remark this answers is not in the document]
polina-c: there is just one null assert in this method. Local variable seems overkilling here.

packages/devtools_app/test/test_infra/flutter_test_driver.dart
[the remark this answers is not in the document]
polina-c: as 'first' worked before and it is not production code, skipped validation

packages/devtools_app/test/test_infra/flutter_test_driver.dart
```diff
@@ -114,53 +112,53 @@ abstract class FlutterTestDriver {
...
+      flutterIsolateId = vm.isolates!.first.id;
     }
     return flutterIsolateId;
   }
 
   Future<Isolate> _getFlutterIsolate() async {
-    final Isolate isolate =
-        await vmService.getIsolate(await getFlutterIsolateId());
+    final Isolate isolate = await vmService!
+        .getIsolate(await (getFlutterIsolateId() as FutureOr<String>));
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/test/test_infra/flutter_test_driver.dart
polina-c: done

packages/devtools_app/test/test_infra/flutter_test_driver.dart
polina-c: done

packages/devtools_app/test/test_infra/flutter_test_driver.dart
polina-c: done

packages/devtools_app/test/test_infra/flutter_test_driver.dart
polina-c: done

packages/devtools_app/test/test_infra/flutter_test_environment.dart
polina-c: done

packages/devtools_app/test/test_infra/flutter_test_environment.dart
polina-c: done

packages/devtools_app/test/test_infra/flutter_test_environment.dart
polina-c: it is assigned to null in tearDownEnvironment

packages/devtools_app/test/test_infra/flutter_test_driver.dart
polina-c: removed assertions

packages/devtools_app/test/test_infra/flutter_test_driver.dart
polina-c: If there were at least two null assersions, it would make sense. There is just one.

packages/devtools_app/test/test_infra/flutter_test_driver.dart
polina-c: same answer

packages/devtools_app/test/test_infra/flutter_test_driver.dart
bkonyi: If you do this:

```dart
final service = vmService;
if (service != null) {
  await service.dispose();
}
```

We only perform one null check instead of two. Anywhere with the pattern:

```dart
if (foo != null) {
  foo!.doSomething();
}
```
should either be updated to:

```
final localFoo = foo;
if (localFoo != null) {
  localFoo.doSomething();
}
```

or

```
foo?.doSomething();
```

polina-c: Hm. What's the point to update 

```dart
if (foo != null) {
  foo!.doSomething();
}
```
with
```dart
final localFoo = foo;
if (localFoo != null) {
  localFoo.doSomething();
}
```
?

The second piece seems less readable to me. Is there performance improvement?

packages/devtools_app/test/test_infra/flutter_test_driver.dart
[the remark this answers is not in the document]
kenzieschmoll: searching for "proc = " I only see where `proc` is assigned to the result of `await Process.start(...)`. How does the code snippet you pasted relate to marking `proc` as late? Does it give you an error? Can you remove the `if (proc != null)`?

packages/devtools_app/test/test_infra/flutter_test_driver.dart
polina-c: Yes, we save on null-check. On other side, the additional variable declaration adds to both performance and cognitive load.
There are two options how we can resolve it: (1) agree it is subjective and should be up to author or (2) declare a team standard for such cases. What do you think?

kenzieschmoll: > the additional variable declaration adds to both performance and cognitive load.

The additional variable should be GC'd once it is out of scope.

Let's declare a team standard to avoid multiple null assertions when possible. Consistency across the codebase will help with cognitive load over time.

polina-c: Sounds good.

Is it written standard or de-facto standard throughout the dart core libraries? I do not remember reading it. 
As it is not obvious, and there are pros and cons in both approaches, should we write it to make contribution easier for new members and external contributors? 

In one of these places, for example:
https://dart.dev/guides/language/effective-dart/usage#null
https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo

polina-c:
@bkonyi , replying your comments that do not have button 'Reply':

https://github.com/flutter/devtools/pull/3873#discussion_r828438080
_flutter is null checked, so it cannot be marked as late.

https://github.com/flutter/devtools/pull/3873#discussion_r828434533
I decided tot to check for length before calling first, because this code does not fail for all existing tests. If/when it start failing, the test author will need to figure out what to do in case of empty collection. Makes sense?

packages/devtools_app/test/test_infra/flutter_test_driver.dart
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/test/test_infra/flutter_test_driver.dart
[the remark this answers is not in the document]
polina-c: fixed

packages/devtools_app/test/test_infra/flutter_test_driver.dart
kenzieschmoll: @leafpetersen @munificent do you have an opinion on whether we should recommend this style to users?

leafpetersen: @munificent may have more to say here.  Obviously, when applicable the `foo?.bar()` approach is fine (I like it).  The style guide already softly [recommends](https://dart.dev/guides/language/effective-dart/usage#consider-copying-a-nullable-field-to-a-local-variable-to-enable-type-promotion) the local variable approach ("Consider").  I personally don't mind using a "!" when there's only one occurrence of the variable (even though there is the redundant check), but I'm pretty sure @lrhn yelled at me the last time I did that in a CL, so.... :)

polina-c: Thank you for the link. I see there are three usages of '!' in the example. And, yes, the question is what we think about single usage.

There are actually two questions:
1. Do we want to recommend local variable for this case to Dart users?
2. Do we want to document the style as a standard across Dart code libraries?

leafpetersen: TBH, my preferred answer to this is that you should use the new pattern feature that @munificent is working on to do the null check and the local variable binding in one step inside of the condition.  Unfortunately he hasn't shipped it yet, dunno what the hold up is... :). 

So I'll defer to others as to what we want to do in the interim, but I would caution against putting large effort into standardizing/cleaning up around one of the existing solutions given that I expect (hope) that we will be shipping a better solution within a year.

polina-c: That's cool! Let's wait then. I am curious how the code would look like.

Issue #3773, opened 2022-03-04
Migrate screens/network to null safety.


polina-c:
Child issue of https://github.com/flutter/devtools/issues/3025.
Tips for faster migration: https://github.com/flutter/devtools/issues/3025#issuecomment-1058800219

PR #3880, opened 2022-03-16
Migrate networking screen to null safety

bkonyi:
Fixes https://github.com/flutter/devtools/issues/3773

### DIFF
--- a/packages/devtools_app/lib/src/http/http_request_data.dart
+++ b/packages/devtools_app/lib/src/http/http_request_data.dart
@@ -513,7 +513,7 @@ class DartIOHttpRequestData extends HttpRequestData {
   bool get _hasError => _request.request?.hasError ?? false;
 
   int? get _endTime =>
-      _hasError ? _request.endTime : _request.response!.endTime;
+      _hasError ? _request.endTime : _request.response?.endTime;
 
   @override
   Duration? get duration {
@@ -539,8 +539,8 @@ class DartIOHttpRequestData extends HttpRequestData {
       'method': _request.method,
       'uri': _request.uri.toString(),
       if (!didFail) ...{
-        'connectionInfo': _request.request!.connectionInfo,
-        'contentLength': _request.request!.contentLength,
+        'connectionInfo': _request.request?.connectionInfo,
+        'contentLength': _request.request?.contentLength,
       },
       if (_request.response != null) ...{
         'compressionState': _request.response!.compressionState,
@@ -605,9 +605,11 @@ class DartIOHttpRequestData extends HttpRequestData {
   }
 
   @override
-  DateTime get endTimestamp => DateTime.fromMicrosecondsSinceEpoch(
-        timelineMicrosecondsSinceEpoch(_endTime!),
-      );
+  DateTime? get endTimestamp => _endTime == null
+      ? null
+      : DateTime.fromMicrosecondsSinceEpoch(
+          timelineMicrosecondsSinceEpoch(_endTime!),
+        );
 
   @override
   DateTime get startTimestamp => DateTime.fromMicrosecondsSinceEpoch(
--- a/packages/devtools_app/lib/src/screens/network/network_controller.dart
+++ b/packages/devtools_app/lib/src/screens/network/network_controller.dart
@@ -2,8 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-// @dart=2.9
-
 import 'dart:async';
 
 import 'package:flutter/foundation.dart';
@@ -45,9 +43,9 @@ class NetworkController
 
   final _requests = ValueNotifier<NetworkRequests>(NetworkRequests());
 
-  ValueListenable<NetworkRequest> get selectedRequest => _selectedRequest;
+  ValueListenable<NetworkRequest?> get selectedRequest => _selectedRequest;
 
-  final _selectedRequest = ValueNotifier<NetworkRequest>(null);
+  final _selectedRequest = ValueNotifier<NetworkRequest?>(null);
 
   /// Notifies that the timeline is currently being recorded.
   ValueListenable<bool> get recordingNotifier => _recordingNotifier;
@@ -55,7 +53,7 @@ class NetworkController
 
   @visibleForTesting
   NetworkService get networkService => _networkService;
-  NetworkService _networkService;
+  late NetworkService _networkService;
 
   /// The timeline timestamps are relative to when the VM started.
   ///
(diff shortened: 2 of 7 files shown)


packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -379,41 +378,43 @@ class NetworkController
...
-  void filterData(Filter<NetworkRequest> filter) {
+  void filterData(Filter<NetworkRequest>? filter) {
     serviceManager.errorBadgeManager.clearErrors(NetworkScreen.id);
     if (filter?.queryFilter == null) {
       _requests.value.requests.forEach(_checkForError);
       filteredData
         ..clear()
         ..addAll(_requests.value.requests);
     } else {
+      final queryFilter = filter!.queryFilter!;
```
kenzieschmoll, line 389: if you put this before the if statement above, can we get rid of the second null assertion here?

packages/devtools_app/lib/src/screens/network/network_model.dart
```diff
@@ -16,31 +14,31 @@ abstract class NetworkRequest with DataSearchStateMixin {
...
+  int? get port;
 
   bool get didFail;
 
   /// True if the request hasn't completed yet.
   bool get inProgress;
 
   String get durationDisplay =>
-      'Duration: ${duration != null ? msText(duration) : 'Pending'}';
+      'Duration: ${duration != null ? msText(duration!) : 'Pending'}';
```
kenzieschmoll, line 41: remove the `=>` and use brackets instead so that we can hoist duration into a local var and remove the double null assert

packages/devtools_app/lib/src/screens/network/network_model.dart
```diff
@@ -86,31 +84,31 @@ class WebSocket extends NetworkRequest {
...
 
   @override
-  Duration get duration {
-    if (_socket.startTime == null || _socket.endTime == null) {
+  Duration? get duration {
+    if (_socket.endTime == null) {
       return null;
     }
-    return Duration(microseconds: _socket.endTime - _socket.startTime);
+    return Duration(microseconds: _socket.endTime! - _socket.startTime);
```
kenzieschmoll, line 91: hoist into a local var

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -605,9 +605,11 @@ class DartIOHttpRequestData extends HttpRequestData {
   }
 
   @override
-  DateTime get endTimestamp => DateTime.fromMicrosecondsSinceEpoch(
-        timelineMicrosecondsSinceEpoch(_endTime!),
-      );
+  DateTime? get endTimestamp => _endTime == null
+      ? null
+      : DateTime.fromMicrosecondsSinceEpoch(
+          timelineMicrosecondsSinceEpoch(_endTime!),
```
kenzieschmoll, line 611: hoist endTime into a local

packages/devtools_app/lib/src/screens/network/network_service.dart
```diff
@@ -26,37 +24,38 @@ class NetworkService {
...
+    Timeline? timeline;
+    if (await serviceManager.service!.isDartIoVersionSupported(
       supportedVersion: SemanticVersion(major: 1, minor: 6),
-      isolateId: serviceManager.isolateManager.selectedIsolate.value.id,
+      isolateId: serviceManager.isolateManager.selectedIsolate.value!.id!,
     )) {
       httpRequests = await _refreshHttpProfile();
     } else {
-      timeline = await serviceManager.service.getVMTimeline(
+      timeline = await serviceManager.service!.getVMTimeline(
```
kenzieschmoll, line 53: hoist serviceManager.service into local

packages/devtools_app/lib/src/screens/network/network_service.dart
```diff
@@ -82,9 +81,9 @@ class NetworkService {
   Future<void> _clearHttpProfile() async {
     assert(serviceManager.service != null);
     if (serviceManager.service == null) return;
-
-    await serviceManager.service.forEachIsolate((isolate) async {
-      final future = serviceManager.service.clearHttpProfile(isolate.id);
+    final service = serviceManager.service!;
```
kenzieschmoll, line 84: move this to the top of the method, remove the assert, and then check if the local var is null in the if statement

packages/devtools_app/lib/src/screens/network/network_screen.dart
```diff
@@ -47,7 +45,7 @@ class NetworkScreen extends Screen {
   @override
   Widget buildStatus(BuildContext context, TextTheme textTheme) {
     final networkController = Provider.of<NetworkController>(context);
-    final color = Theme.of(context).textTheme.bodyText2.color;
+    final color = Theme.of(context).textTheme.bodyText2!.color!;
```
polina-c, line 48: How about assigning default color instead of null checking?

packages/devtools_app/lib/src/screens/network/network_screen.dart
```diff
@@ -112,7 +110,8 @@ Example queries:
 
 class _NetworkScreenBodyState extends State<NetworkScreenBody>
     with AutoDisposeMixin {
-  NetworkController _networkController;
+  bool initialized = false;
```
polina-c, line 113: Is there reason for initialized to be public?

bkonyi: Nope, I'll make it private.

packages/devtools_app/lib/src/screens/network/network_screen.dart
```diff
@@ -172,11 +172,11 @@ class _NetworkProfilerControls extends StatefulWidget {
...
 class _NetworkProfilerControlsState extends State<_NetworkProfilerControls>
     with AutoDisposeMixin, SearchFieldMixin<_NetworkProfilerControls> {
-  NetworkRequests requests;
+  late NetworkRequests requests;
 
-  List<NetworkRequest> filteredRequests;
+  late List<NetworkRequest> filteredRequests;
 
-  bool recording;
+  bool recording = false;
```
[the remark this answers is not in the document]
bkonyi: Yup!

packages/devtools_app/lib/src/screens/network/network_screen.dart
bkonyi: @kenzieschmoll may have a different opinion on this, but if these are null that means we've setup our theming incorrectly. These should never actually be null AFAIK.

kenzieschmoll: It is fine to null-assert. Especially when we know this was fine pre-migration.

packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -380,13 +380,13 @@ class NetworkController
   @override
   void filterData(Filter<NetworkRequest>? filter) {
     serviceManager.errorBadgeManager.clearErrors(NetworkScreen.id);
-    if (filter?.queryFilter == null) {
+    if (filter == null || filter.queryFilter == null) {
```
kenzieschmoll, line 383: maybe my comment was a little misleading. I meant this:
```
final queryFilter = filter?.queryFilter;
if (queryFilter == null) {
  ...
} else {
  // no need to create the var and null assert here
  filteredData...
}
```

kenzieschmoll (approved):
lgtm with one comment

packages/devtools_app/lib/src/screens/network/network_controller.dart
bkonyi: Ah, good call. That works better.

PR #3885, opened 2022-03-17
Migrate ansi_up_test to null safety.

polina-c:


packages/devtools_app/test/ansi_up_test.dart
```diff
@@ -145,8 +145,8 @@ void main() {
 
       final detailsComputer = computedDetails == null
           ? null
-          : () =>
-              Future.delayed(const Duration(seconds: 1), () => computedDetails);
+          : () => Future.delayed(
+              const Duration(seconds: 1), () => computedDetails!);
```
kenzieschmoll, line 149: hoist computedDetails into local

packages/devtools_app/test/ansi_up_test.dart
```diff
@@ -178,8 +178,10 @@ void main() {
       when(fakeServiceManager.connectedApp.isFlutterWebAppNow)
           .thenReturn(false);
       when(fakeServiceManager.connectedApp.isProfileBuildNow).thenReturn(false);
-      when(fakeServiceManager.errorBadgeManager.errorCountNotifier(any))
-          .thenReturn(ValueNotifier<int>(0));
+      // TODO(polinach): when we start supporting browser tests, uncomment
+      // and fix this. See https://github.com/flutter/devtools/issues/3616.
+      // when(fakeServiceManager.errorBadgeManager.errorCountNotifier(any))
+      //     .thenReturn(ValueNotifier<int>(0));
```
kenzieschmoll, line 184: why does browser test support affect the errorBadgeManager mocking specifically?

packages/devtools_app/test/ansi_up_test.dart
```diff
@@ -231,13 +233,13 @@ void main() {
 
   group('Debugger Screen', () {
     FakeServiceManager fakeServiceManager;
```
kenzieschmoll, line 235: make this late

packages/devtools_app/test/ansi_up_test.dart
polina-c: computedDetails is local

packages/devtools_app/test/ansi_up_test.dart
polina-c: Because I cannot run tis test to verify the fix. So, it is better to postpone.

packages/devtools_app/test/ansi_up_test.dart
```diff
@@ -231,13 +233,13 @@ void main() {
...
-    MockDebuggerController debuggerController;
+    MockDebuggerController? debuggerController;
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/test/ansi_up_test.dart
polina-c: done

packages/devtools_app/test/ansi_up_test.dart
kenzieschmoll: @bkonyi interesting, so does the flow analysis not work with ternary operators? after the `:` we know `computedDetails` is not null

packages/devtools_app/test/ansi_up_test.dart
kenzieschmoll: If we don't know the impact of changing these lines, it seems odd that we would comment this out then. Can we uncomment and default to leaving as it was if we can't verify?

packages/devtools_app/test/ansi_up_test.dart
```diff
@@ -267,8 +269,10 @@ void main() {
       setGlobal(ServiceConnectionManager, fakeServiceManager);
       fakeServiceManager.consoleService.ensureServiceInitialized();
 
-      when(fakeServiceManager.errorBadgeManager.errorCountNotifier(any))
-          .thenReturn(ValueNotifier<int>(0));
+      // TODO(polinach): when we start supporting browser tests, uncomment
+      // and fix this. See https://github.com/flutter/devtools/issues/3616.
+      // when(fakeServiceManager.errorBadgeManager.errorCountNotifier(any))
```
kenzieschmoll, line 274: same comment here

packages/devtools_app/test/ansi_up_test.dart
bkonyi: I'm not sure this is an issue with the analysis not working for ternary operators, but is probably due to the asynchrony involved or simply the variable capture by the closure creation.

For example, we can invalidate the null check in this simple example:

```dart
void main() {
  String? foo = 'bar';
  print(foo + 'baz'); // No null check needed since we instantiated foo.
  Future.delayed(const Duration(seconds: 1, () => print(foo)));
  foo = null;
}
```

The output from this is:

```
barbaz
null
```

bkonyi: Confirmed this isn't a ternary operator issue since the following example works fine:

```dart
String? foo() => 'foo';

void main() {
  final x = foo();
  // Try to append to 'x' to double check it's not null.
  print(x == null ? null : x + 'bar');
}
```

Which outputs: `foobar`

packages/devtools_app/test/ansi_up_test.dart
polina-c: I commented them because analyzer complains about the parameter 'any' that is of type Null and thus cannot be passed as non-nullable parameter.

kenzieschmoll: Here's an idea that should work. Go to the MockErrorBadgeManager class and override this method:

```
@override
ValueListenable<int> errorCountNotifier(String screenId) {
    return ValueNotifier<int>(0);
  }
```

Then you can remove these lines entirely instead of commenting them out

polina-c: discussed offline and decided to comment it as is for now

packages/devtools_app/test/ansi_up_test.dart
polina-c: updated comment

kenzieschmoll (approved):

PR #3886, opened 2022-03-17
Update bots to run tests with --no-sound-null-safety

polina-c:
We will need to remove this flag after full migration.

Example of failure: https://github.com/flutter/devtools/runs/5579506929?check_suite_focus=true

See https://stackoverflow.com/questions/66333667/cannot-run-with-sound-null-safety-because-the-following-dependencies-dont-supp


kenzieschmoll (approved):
Please add an item to the meta null safety tracking issue to revert this PR, if you haven't already

Issue #3699, opened 2022-02-16
Remove service protocol version checks in DevTools


kenzieschmoll:
There are several places in DevTools where we check the vm service protocol or dds version before doing some action. Most of these came from breaking changes to the api or adding new functionality. Now that we are shipping with the SDK, we do not need to support legacy implementations and can just default to supporting the SDK we ship with.

Example: https://github.com/flutter/devtools/pull/3696/files#r808223584

PR #3907, opened 2022-03-21
Remove service protocol version checks in DevTools

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/3699


packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -507,6 +68,11 @@ class DartIOHttpRequestData extends HttpRequestData {
         .then((updated) => _request = updated);
   }
 
+  static List<Cookie> _parseCookies(List<dynamic>? cookies) {
```
bkonyi, line 71: Should this be `List<dynamic>`? Do we not have a static type to use here?

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -553,18 +117,65 @@ class DartIOHttpRequestData extends HttpRequestData {
     };
   }
 
+  @override
+  String? get contentType {
+    if (responseHeaders == null || responseHeaders![_contentTypeKey] == null) {
```
bkonyi, line 122: Create a local for `responseHeaders` to avoid the extra checks?

packages/devtools_app/lib/src/http/http_request_data.dart
```diff
@@ -553,18 +117,65 @@ class DartIOHttpRequestData extends HttpRequestData {
...
+    return connectionInfo != null ? connectionInfo[_localPortKey] : null;
+  }
+
   /// True if the HTTP request hasn't completed yet, determined by the lack of
   /// an end event.
   @override
   bool get inProgress =>
       _hasError ? !_request.isRequestComplete : !_request.isResponseComplete;
 
   /// All instant events logged to the timeline for this HTTP request.
```
bkonyi, line 177: This comment still references the timeline, but you've removed that in this PR, no?

packages/devtools_app/linux/flutter/generated_plugins.cmake
```diff
@@ -7,6 +7,9 @@ list(APPEND FLUTTER_PLUGIN_LIST
   url_launcher_linux
 )
```
bkonyi, line 9: Were these changes intentional?

bkonyi (approved):
Generally LGTM

packages/devtools_app/lib/src/http/http_request_data.dart
kenzieschmoll: changed to `List<String>`

packages/devtools_app/lib/src/http/http_request_data.dart
kenzieschmoll: done

packages/devtools_app/linux/flutter/generated_plugins.cmake
kenzieschmoll: removed

Issue #3879, opened 2022-03-16
Track inspector page load time for DevTools analytics


annagrin:
Inspector page currently does not track pageReady analytics. 

Example: Debugger page tracks its load time:
Start: 
https://github.com/flutter/devtools/blob/a616832995d13410a4078b506c91927865ff6904/packages/devtools_app/lib/src/debugger/debugger_screen.dart#L110
End:
https://github.com/flutter/devtools/blob/a616832995d13410a4078b506c91927865ff6904/packages/devtools_app/lib/src/debugger/debugger_screen.dart#L157

Issue #393, opened 2019-03-05
Timeline window doesn't respect my scroll direction setting


DaveShuckerow:
I use scroll forward as scroll up on my trackpad, but devtools is zooming out in the timeline view instead of zooming in.

kenzieschmoll:
Interesting - I wonder if this has to do with your scroll direction settings on your computer. On my mac, scrolling up (from the edge of the trackpad closest to me towards the keyboard) zooms the flame chart in.

This same scrolling action scrolls web pages down for me. What does it do for you?

This is a screenshot from my mac trackpad settings.
<img width="317" alt="screen shot 2019-03-05 at 2 30 17 pm" src="https://user-images.githubusercontent.com/43759233/53842080-56a4ff00-3f53-11e9-8cde-54d2a5fb4d61.png">

What device are you using?

PR #3924, opened 2022-03-25
Add analytics to track inspector tree first load time

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/3879
Fixes https://github.com/flutter/devtools/issues/3930


packages/devtools_app/lib/src/analytics/_analytics_web.dart
```diff
@@ -394,6 +394,14 @@ void timeEnd(
   );
 }
 
+void cancelTimingOperation(String screenName, String timedOperation) {
+  final operationKey = _operationKey(
+    screenName,
+    timedOperation,
+  );
+  _timedOperationsInProgress.remove(operationKey);
```
jacob314, line 402: add assert that the timing operation is actually in progress.

kenzieschmoll: done

kenzieschmoll:
FYI @annagrin this may be a better place to send the DWDS event (https://github.com/flutter/devtools/pull/3834) if you actually want to measure up to the point that the inspector tree is loaded. This current bug (https://github.com/flutter/devtools/issues/393) needs to be fixed though, otherwise the DWDS event may not be sent.

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -418,12 +427,14 @@ class InspectorController extends DisposableController
     }
 
     if (flutterAppFrameReady) {
-      // TODO: measure and send DevTools pageReady analytics:
-      // https://github.com/flutter/devtools/issues/3879
-      await serviceManager.sendDwdsEvent(
-        screen: InspectorScreen.id,
-        action: analytics_constants.pageReady,
-      );
+      if (serviceManager.connectedApp!.isDartWebAppNow!) {
```
kenzieschmoll, line 430: @annagrin @elliette is this the proper check that will ensure dwds is available and that the `ext.dwds.sendEvent` service has been registered?

This was failing for a flutter mobile app and was causing https://github.com/flutter/devtools/issues/3930. I removed the await so that sending this event does not block the inspector loading.

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -121,6 +121,10 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
       treeType: FlutterTreeType.widget,
     );
 
+    if (!inspectorController.firstLoadCompleted) {
+      ga.timeStart(InspectorScreen.id, analytics_constants.pageReady);
+    }
+
```
kenzieschmoll, line 127: @jacob314 because the InspectorController is not provided using package:provider like our other controllers, it is recreated in the inspector screen's `initState` method each time the inspector tab is selected. So for example, opening devtools to the inspector, switching to a different tab, then switching back to the inspector tab would trigger this code twice.

We could do one of two things here:
1) time the inspector tree load time for every time the inspector screen is built. This would include the initial load as well as subsequent loads from tab switches. 
2) time the inspector tree load time for _only_ the first load. To do this, we'd store `firstLoadCompleted` as a top level variable instead of as a field on the controller class. This would make more sense if we expect subsequent loads to be faster than the initial load, as that would skew the results of the true "first load" time.

jacob314: Timing only the first load seems reasonable. It would probably also be a good idea to clean up the tech debt that the inspectorController is not provided using package:provider.

kenzieschmoll: sg. Filed https://github.com/flutter/devtools/issues/3950 to track.

PR #3927, opened 2022-03-25
Migrate the debugger panel to null-safety

elliette:


packages/devtools_app/lib/src/screens/debugger/breakpoints.dart
```diff
@@ -114,11 +113,11 @@ class _BreakpointsState extends State<Breakpoints> {
   }
 
   void _onBreakpointSelected(BreakpointAndSourcePosition bp) {
-    controller.selectBreakpoint(bp);
+    controller!.selectBreakpoint(bp);
```
jacob314, line 116: why does this have a bang?

packages/devtools_app/lib/src/screens/debugger/call_stack.dart
```diff
@@ -30,13 +29,13 @@ class _CallStackState extends State<CallStack> {
     final newController = Provider.of<DebuggerController>(context);
     if (newController == controller) return;
```
jacob314, line 30: this line is now wrong. Needs to be newController == _controller. Without that, this code should be crashing at runtime.

packages/devtools_app/lib/src/screens/debugger/breakpoints.dart
```diff
@@ -19,28 +17,29 @@ double get executableLineRadius => scaleByFontFactor(1.5);
...
-  DebuggerController controller;
+  DebuggerController get controller => _controller!;
+  DebuggerController? _controller;
 
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
 
     final newController = Provider.of<DebuggerController>(context);
     if (newController == controller) return;
```
jacob314, line 35: this line needs to be
newController == _controller

to avoid crashing

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -239,16 +241,16 @@ class _CodeViewState extends State<CodeView>
...
     // Ensure the syntax highlighter has been initialized.
     // TODO(bkonyi): process source for highlighting on a separate thread.
-    if (parsedScript.script.source != null) {
-      if (parsedScript.script.source.length < 500000 &&
-          parsedScript.highlighter != null) {
-        final highlighted = parsedScript.highlighter.highlight(context);
+    if (parsedScript?.script.source != null) {
+      final scriptSource = parsedScript!.script.source!;
+      if (scriptSource.length < 500000) {
+        final highlighted = parsedScript!.highlighter.highlight(context);
```
jacob314, line 247: avoid bang

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -290,7 +292,7 @@ class _CodeViewState extends State<CodeView>
 
     return HistoryViewport(
       history: widget.controller.scriptsHistory,
-      generateTitle: (script) => script.uri,
+      generateTitle: (dynamic script) => script.uri,
```
jacob314, line 295: avoid ever adding dynamic to work around null safety issues.

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -302,7 +304,7 @@ class _CodeViewState extends State<CodeView>
           enabled: widget.controller.scriptsHistory.hasScripts,
         ),
       ],
-      contentBuilder: (context, script) {
+      contentBuilder: (context, dynamic script) {
```
jacob314, line 307: solve problem without adding dynamic. you are stopping the static type analysis from catching issues.

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -557,7 +559,7 @@ class GutterItem extends StatelessWidget {
     final theme = Theme.of(context);
 
     final foregroundColor = theme.isDarkTheme
-        ? theme.textTheme.bodyText2.color
+        ? theme.textTheme.bodyText2!.color
```
jacob314, line 562: handle this being null gracefully.  This looks like a sign we are using themes a bit wrong.

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -659,17 +661,17 @@ class _LinesState extends State<Lines> with AutoDisposeMixin {
         activeSearch = widget.activeSearchMatchNotifier.value;
       });
 
-      if (activeSearch != null) {
-        final isOutOfViewTop = activeSearch.position.line * CodeView.rowHeight <
+      if (activeSearch?.position.line != null) {
```
jacob314, line 664: add local variables to avoid all the `!` for activeSearch

packages/devtools_app/lib/src/screens/debugger/breakpoints.dart
```diff
@@ -19,14 +17,14 @@ double get executableLineRadius => scaleByFontFactor(1.5);
...
-  const Breakpoints({Key key}) : super(key: key);
+  const Breakpoints({Key? key}) : super(key: key);
 
   @override
   _BreakpointsState createState() => _BreakpointsState();
 }
 
 class _BreakpointsState extends State<Breakpoints> {
-  DebuggerController controller;
+  DebuggerController? controller;
```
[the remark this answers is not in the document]
elliette: This is done now

packages/devtools_app/lib/src/screens/debugger/controls.dart
```diff
@@ -16,31 +14,31 @@ import '../../ui/label.dart';
...
+    addAutoDisposeListener(controller!.stackFramesWithLocation);
   }
 
   @override
   Widget build(BuildContext context) {
-    final isPaused = controller.isPaused.value;
-    final resuming = controller.resuming.value;
-    final hasStackFrames = controller.stackFramesWithLocation.value.isNotEmpty;
-    final isSystemIsolate = controller.isSystemIsolate;
+    final isPaused = controller!.isPaused.value;
```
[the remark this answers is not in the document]
elliette: Done

packages/devtools_app/lib/src/screens/debugger/breakpoints.dart
elliette: I missed that! It has been removed

packages/devtools_app/lib/src/screens/debugger/call_stack.dart
elliette: Hah yes I discovered that when I tried to run it - fixed now, thanks!

packages/devtools_app/lib/src/screens/debugger/breakpoints.dart
elliette: Done

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -99,11 +99,10 @@ class _CodeViewState extends State<CodeView>
     horizontalController = ScrollController();
     _lastScriptRef = widget.scriptRef;
 
-    if (widget.initialPosition != null) {
-      final location = widget.initialPosition.location;
+    if (initialPosition?.location?.line != null) {
       // Lines are 1-indexed. Scrolling to line 1 required a scroll position of
       // 0.
-      final lineIndex = location.line - 1;
+      final lineIndex = initialPosition!.location!.line! - 1;
```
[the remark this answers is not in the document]
elliette: Doen

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -239,16 +241,16 @@ class _CodeViewState extends State<CodeView>
 
     // Ensure the syntax highlighter has been initialized.
     // TODO(bkonyi): process source for highlighting on a separate thread.
-    if (parsedScript.script.source != null) {
-      if (parsedScript.script.source.length < 500000 &&
-          parsedScript.highlighter != null) {
-        final highlighted = parsedScript.highlighter.highlight(context);
+    if (parsedScript?.script.source != null) {
```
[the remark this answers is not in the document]
elliette: Done

packages/devtools_app/lib/src/screens/debugger/codeview.dart
elliette: Done

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -239,16 +241,16 @@ class _CodeViewState extends State<CodeView>
...
+      final scriptSource = parsedScript!.script.source!;
+      if (scriptSource.length < 500000) {
+        final highlighted = parsedScript!.highlighter.highlight(context);
 
         // Look for [TextSpan]s which only contain '\n' to manually break the
         // output from the syntax highlighter into individual lines.
         var currentLine = <TextSpan>[];
         highlighted.visitChildren((span) {
-          currentLine.add(span);
+          currentLine.add(span as TextSpan);
```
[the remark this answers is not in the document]
elliette: The type of the parameter to `visitChildren`' is `InlineSpanVisitor`

packages/devtools_app/lib/src/screens/debugger/codeview.dart
elliette: Oops! I missed removing these from the null safety tool auto-migration

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -337,7 +339,7 @@ class _CodeViewState extends State<CodeView>
                               breakpoints: breakpoints
                                   .where((bp) => bp.scriptRef == scriptRef)
                                   .toList(),
-                              executableLines: parsedScript.executableLines,
+                              executableLines: parsedScript!.executableLines,
```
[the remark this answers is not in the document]
elliette: Done

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -463,7 +465,7 @@ class _CodeViewState extends State<CodeView>
               overflow: TextOverflow.ellipsis,
             ),
             Text(
-              scriptRef.uri,
+              scriptRef.uri!,
```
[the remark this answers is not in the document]
elliette: Done

packages/devtools_app/lib/src/screens/debugger/codeview.dart
elliette: Done

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -894,8 +896,10 @@ class _LineItemState extends State<LineItem> {
   }
 
   TextSpan searchAwareLineContents() {
-    final activeSearchAwareContents =
-        _activeSearchAwareLineContents(widget.lineContents.children);
+    final children = widget.lineContents.children as List<TextSpan>?;
```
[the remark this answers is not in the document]
elliette: `widget.lineContents.children` has the type `List<InlineSpan>?`

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -178,8 +177,9 @@ class _CodeViewState extends State<CodeView>
 
     // TODO(devoncarew): Adjust this so we don't scroll if we're already in the
     // middle third of the screen.
-    if (parsedScript.lineCount * CodeView.rowHeight > extent) {
-      final lineIndex = location.line - 1;
+    if (parsedScript?.lineCount != null &&
```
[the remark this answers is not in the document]
elliette: Done

packages/devtools_app/lib/src/screens/debugger/breakpoints.dart
```diff
@@ -19,28 +17,29 @@ double get executableLineRadius => scaleByFontFactor(1.5);
...
+  const Breakpoints({Key? key}) : super(key: key);
 
   @override
   _BreakpointsState createState() => _BreakpointsState();
 }
 
 class _BreakpointsState extends State<Breakpoints> {
-  DebuggerController controller;
+  DebuggerController get controller => _controller!;
+  DebuggerController? _controller;
```
kenzieschmoll, line 28: Another pattern we've used elsewhere in DevTools is making the controller var late and using a bool to track whether or not we have initialized the controller. This avoids having to perform a null-assert every time the controller getter is accessed. See https://github.com/flutter/devtools/blob/3f83c3597740677a5d86b7808e2ac5c8addba409/packages/devtools_app/lib/src/screens/performance/performance_screen.dart#L101-L103

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -132,33 +129,36 @@ class DebuggerController extends DisposableController
...
   /// This cache includes autocompletes from all libraries imported and exported
   /// by the library as well as all private autocompletes for the library.
-  Map<LibraryRef, Future<Set<String>>>
+  Map<LibraryRef, Future<Set<String?>>>
       libraryMemberAndImportsAutocompleteCache = {};
 
   ProgramExplorerController get programExplorerController =>
       _programExplorerController;
-  ProgramExplorerController _programExplorerController;
+  late final ProgramExplorerController _programExplorerController;
```
kenzieschmoll, line 158: This getter doesn't appear to be adding any value. Can we just make the late final var public?
`late final ProgramExplorerController programExplorerController;`

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -230,9 +230,9 @@ class DebuggerController extends DisposableController
     _scriptLocation.value = scriptLocation;
   }
 
-  Future<Script> getScriptForRef(ScriptRef ref) async {
-    final cachedScript = scriptManager.getScriptCached(ref);
-    if (cachedScript == null && ref != null) {
+  Future<Script?> getScriptForRef(ScriptRef? ref) async {
+    final cachedScript = scriptManager.getScriptCached(ref!);
```
kenzieschmoll, line 234: if we are unconditionally null asserting `ref` here, then let's make the parameter be non-nullable

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -256,7 +256,8 @@ class DebuggerController extends DisposableController
 
     if (script != null) {
       try {
-        final positions = await getBreakablePositions(script);
+        final positions =
+            await getBreakablePositions(script) as List<SourcePosition>;
```
kenzieschmoll, line 260: have we verified that this code does not throw? @polina-c ran into some issues where using `as NewType` instead of `.cast<NewType>()` was causing issues. Not sure if this is the same situation or not.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -307,38 +308,38 @@ class DebuggerController extends DisposableController
...
       _selectedStackFrame.value?.frame ??
-      _stackFramesWithLocation.value?.safeFirst?.frame;
+      _stackFramesWithLocation.value.safeFirst?.frame;
 
   final _variables = ValueNotifier<List<DartObjectNode>>([]);
 
   ValueListenable<List<DartObjectNode>> get variables => _variables;
 
-  final _breakpoints = ValueNotifier<List<Breakpoint>>([]);
+  final _breakpoints = ValueNotifier<List<Breakpoint>?>([]);
```
kenzieschmoll, line 324: this list can be non-nullable. We are using the empty list as the base state and not null.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -307,38 +308,38 @@ class DebuggerController extends DisposableController
...
 
-  ValueListenable<List<Breakpoint>> get breakpoints => _breakpoints;
+  ValueListenable<List<Breakpoint?>?> get breakpoints => _breakpoints;
```
kenzieschmoll, line 326: both breakpoint and the list should be able to be non-nullable

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -307,38 +308,38 @@ class DebuggerController extends DisposableController
...
-  ValueListenable<BreakpointAndSourcePosition> get selectedBreakpoint =>
+  ValueListenable<BreakpointAndSourcePosition?> get selectedBreakpoint =>
       _selectedBreakpoint;
 
   final _exceptionPauseMode =
-      ValueNotifier<String>(ExceptionPauseMode.kUnhandled);
+      ValueNotifier<String?>(ExceptionPauseMode.kUnhandled);
 
-  ValueListenable<String> get exceptionPauseMode => _exceptionPauseMode;
+  ValueListenable<String?> get exceptionPauseMode => _exceptionPauseMode;
```
kenzieschmoll, line 342: we should be able to make these String types non-nullable by defaulting to `ExceptionPauseMode.kUnhandled` everywhere where a null value may have been assigned.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -351,13 +352,18 @@ class DebuggerController extends DisposableController
     _librariesVisible.value = !_librariesVisible.value;
   }
 
-  IsolateRef isolateRef;
-  bool get isSystemIsolate => isolateRef?.isSystemIsolate ?? false;
+  IsolateRef get isolateRef {
+    return _isolateRef!;
+  }
+
+  IsolateRef? _isolateRef;
```
kenzieschmoll, line 359: if _isolateRef can actually be null, we should not be unconditionally null-asserting here. If it shouldn't be null, then we should try to make _isolateRef late so that it can be non-nullable

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -480,38 +486,38 @@ class DebuggerController extends DisposableController
...
       onUnavailable: onUnavailable,
     );
   }
 
   Future<void> clearBreakpoints() async {
-    final breakpoints = _breakpoints.value.toList();
-    await Future.forEach(breakpoints, (Breakpoint breakpoint) {
-      return removeBreakpoint(breakpoint);
+    final breakpoints = _breakpoints.value!.toList();
+    await Future.forEach(breakpoints, (Breakpoint? breakpoint) {
```
kenzieschmoll, line 511: can breakpoint be non-nulalble? when would we have a null value in the breakpoints list

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -521,27 +527,28 @@ class DebuggerController extends DisposableController
...
       } catch (_) {
         // ignore errors setting breakpoints
       }
     }
   }
 
   Future<void> setIsolatePauseMode(String mode) async {
-    await _service.setIsolatePauseMode(isolateRef.id, exceptionPauseMode: mode);
+    await _service.setIsolatePauseMode(isolateRef.id!,
+        exceptionPauseMode: mode);
```
kenzieschmoll, line 551: nit: trailing comma

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -563,12 +570,12 @@ class DebuggerController extends DisposableController
     // We're resuming and another isolate has started in a paused state,
     // resume any pauseState isolates.
     if (_resuming.value &&
-        event.isolate.id != isolateRef?.id &&
+        event.isolate!.id != isolateRef.id &&
```
kenzieschmoll, line 573: put `event.isolate!.id` into a local variable at the top of this method. This prevents one null-assert

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -847,28 +854,27 @@ class DebuggerController extends DisposableController
...
-    assert(isolate != null);
     final scriptRefs = await scriptManager.retrieveAndSortScripts(isolateRef);
 
     // Update the selected script.
-    final mainScriptRef = scriptRefs.firstWhere((ref) {
-      return ref.uri == isolate.rootLib.uri;
-    }, orElse: () => null);
+    final mainScriptRef = scriptRefs.firstWhereOrNull((ref) {
+      return ref.uri == isolate.rootLib!.uri;
+    })!;
```
kenzieschmoll, line 866: null asserting a `firstWhereOrNull` return is some null-safety code smell. Remove the null-assert here and either handle the null case here or make sure _populateScriptAndShowLocation handles a null value

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -847,28 +854,27 @@ class DebuggerController extends DisposableController
...
 
     // Display the script location.
     _populateScriptAndShowLocation(mainScriptRef);
   }
 
   Future<BreakpointAndSourcePosition> _createBreakpointWithLocation(
-      Breakpoint breakpoint) async {
-    if (breakpoint.resolved) {
+      Breakpoint? breakpoint) async {
```
kenzieschmoll, line 873: nit: tc

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -884,24 +890,24 @@ class DebuggerController extends DisposableController
...
   void selectBreakpoint(BreakpointAndSourcePosition bp) {
     _selectedBreakpoint.value = bp;
 
     if (bp.sourcePosition == null) {
-      showScriptLocation(ScriptLocation(bp.scriptRef));
+      showScriptLocation(ScriptLocation(bp.scriptRef!));
     } else {
       showScriptLocation(
-          ScriptLocation(bp.scriptRef, location: bp.sourcePosition));
+          ScriptLocation(bp.scriptRef!, location: bp.sourcePosition));
```
kenzieschmoll, line 906: hoist bp.scriptRef into a local

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -912,7 +918,7 @@ class DebuggerController extends DisposableController
 
     if (frame?.scriptRef != null) {
       showScriptLocation(
-          ScriptLocation(frame.scriptRef, location: frame.position));
+          ScriptLocation(frame!.scriptRef!, location: frame.position));
```
kenzieschmoll, line 921: hoist frame?.scriptRef into a local. Then when you check that the local var is not null, you can use the local here where it will have been promoted to a non-nullable type

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -923,20 +929,21 @@ class DebuggerController extends DisposableController
...
-      ..sort((a, b) => sortFieldsByName(a.name, b.name));
+      ..sort((a, b) => sortFieldsByName(a.name!, b.name!));
     return variables;
   }
 
   List<Frame> _framesForCallStack(
-    List<Frame> stackFrames, {
-    List<Frame> asyncCausalFrames,
-    InstanceRef reportedException,
+    List<Frame>? stackFrames, {
```
kenzieschmoll, line 940: instead of allowing stackFrames to be nullable, default to an empty list where this method is called (e.g. `stack.frames ?? []`)

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -967,10 +974,10 @@ class DebuggerController extends DisposableController
     return frames;
   }
 
-  final Map<String, List<SourcePosition>> _breakPositionsMap = {};
+  final Map<String?, List<SourcePosition>> _breakPositionsMap = {};
```
kenzieschmoll, line 977: do we really want null keys?

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -1013,7 +1020,7 @@ class DebuggerController extends DisposableController
   // TODO(kenz): search through previous matches when possible.
   @override
   List<SourceToken> matchesForSearch(
-    String search, {
+    String? search, {
```
kenzieschmoll, line 1023: make this non-nullable. Default callers to the empty string if they were previously passing null

kenzieschmoll (commented):
Partial review. Reviewed up through debugger_controller.dart.

packages/devtools_app/lib/src/screens/debugger/breakpoints.dart
jacob314: > Another pattern we've used elsewhere in DevTools is making the controller var late and using a bool to track whether or not we have initialized the controller. This avoids having to perform a null-assert every time the controller getter is accessed. See
> 
> https://github.com/flutter/devtools/blob/3f83c3597740677a5d86b7808e2ac5c8addba409/packages/devtools_app/lib/src/screens/performance/performance_screen.dart#L101-L103

What is the motivation for using this pattern rather than defining the getter + setter pair?
There isn't much magic to `late`. It basically just desugars to the getter setter pair. Keeping `_controllerInitialized` separate from the field leaves room for bugs if one is set but the other isn't and uses more memory as an extra boolean field is tracked.
With late, the Dart runtime still has to do the `!` check, you just didn't have to explicitly write it. There are a few cases where the AOT compiler is clever about optimizing the bang checks for late in ways that it may not for the desugared version but in general that shouldn't matter for DevTools.

kenzieschmoll: > With late, the Dart runtime still has to do the ! check, you just didn't have to explicitly write it. 

I was unaware of this. If calling `controller!` is the same thing that would be happening under the hood for `late controller`, then there is no benefit to using late. CC @bkonyi who originally introduced this pattern in case you had any motivations behind using late over a getter setter pair that I may be missing.

jacob314: Yeah there is no free lunch. `late` still requires the runtime to verify the member is non-null before use.
Another valid case to take `late` variables and switch them to a getter + setter pair is if you need to null the variable in the `dispose()` method.

packages/devtools_app/lib/src/screens/debugger/debugger_screen.dart
```diff
@@ -383,10 +385,10 @@ class _DebuggerStatusState extends State<DebuggerStatus> with AutoDisposeMixin {
       return 'paused$reason';
     }
 
-    final fileName = ' at ' + frame.location.script.uri.split('/').last;
-    final script = await scriptManager.getScript(frame.location.script);
+    final fileName = ' at ' + frame.location!.script!.uri!.split('/').last;
+    final script = await scriptManager.getScript(frame.location!.script!);
     final pos =
-        SourcePosition.calculatePosition(script, frame.location.tokenPos);
+        SourcePosition.calculatePosition(script, frame.location!.tokenPos!);
```
kenzieschmoll, line 391: pull `frame.location!` out into a local var and then use that local to create another local var `location.script!`. Then use in these three lines to avoid the redundant null checks

packages/devtools_app/lib/src/screens/debugger/debugger_screen.dart
```diff
@@ -400,11 +402,11 @@ class FloatingDebuggerControls extends StatefulWidget {
...
 class _FloatingDebuggerControlsState extends State<FloatingDebuggerControls>
     with AutoDisposeMixin {
-  DebuggerController controller;
+  late DebuggerController controller;
 
-  bool paused;
+  late bool paused;
 
-  double controlHeight;
+  double? controlHeight;
```
kenzieschmoll, line 409: can this be late also

packages/devtools_app/lib/src/screens/debugger/evaluate.dart
```diff
@@ -32,11 +32,11 @@ class ExpressionEvalField extends StatefulWidget {
 
 class _ExpressionEvalFieldState extends State<ExpressionEvalField>
     with AutoDisposeMixin, SearchFieldMixin {
-  AutoCompleteController _autoCompleteController;
+  late AutoCompleteController _autoCompleteController;
   int historyPosition = -1;
 
   String _activeWord = '';
-  List<String> _matches = [];
+  List<String?> _matches = [];
```
kenzieschmoll, line 39: do we want to allow null matches? probably not if we are null asserting below on line 117

packages/devtools_app/lib/src/screens/debugger/evaluate.dart
```diff
@@ -218,9 +218,10 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
     );
   }
 
-  List<String> _filterMatches(List<String> previousMatches, String activeWord) {
+  List<String?> _filterMatches(
+      List<String?> previousMatches, String activeWord) {
```
kenzieschmoll, line 222: nit: tc
also, we should make String non nullable

packages/devtools_app/lib/src/screens/debugger/evaluate.dart
```diff
@@ -331,18 +332,17 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
   }
 }
 
-Future<List<String>> autoCompleteResultsFor(
+Future<List<String?>> autoCompleteResultsFor(
   EditingParts parts,
   DebuggerController controller,
 ) async {
-  final result = <String>{};
+  final result = <String?>{};
```
kenzieschmoll, line 339: non nullable strings

packages/devtools_app/lib/src/screens/debugger/evaluate.dart
```diff
@@ -456,38 +456,38 @@ Future<Set<String>> _libraryMemberAndImportsAutocompletes(
   return result;
 }
 
-Future<Set<String>> libraryMemberAutocompletes(
+Future<Set<String?>> libraryMemberAutocompletes(
   DebuggerController controller,
-  LibraryRef libraryRef, {
-  @required bool includePrivates,
+  LibraryRef? libraryRef, {
```
kenzieschmoll, line 461: non nullable if we are unconditionally null asserting below (line 466)

packages/devtools_app/lib/src/screens/debugger/evaluate.dart
```diff
@@ -456,38 +456,38 @@ Future<Set<String>> _libraryMemberAndImportsAutocompletes(
...
-Future<Set<String>> _libraryMemberAutocompletes(
+Future<Set<String?>> _libraryMemberAutocompletes(
   DebuggerController controller,
   LibraryRef libraryRef,
 ) async {
-  final result = <String>{};
-  final Library library = await controller.getObject(libraryRef);
-  result.addAll(library.variables.map((field) => field.name));
-  result.addAll(library.functions
+  final result = <String?>{};
```
kenzieschmoll, line 478: I think all these nullable String types should be able to be made non-nullable

packages/devtools_app/lib/src/screens/debugger/evaluate.dart
```diff
@@ -521,37 +521,37 @@ Future<void> _addAllInstanceMembersToAutocompleteList(
...
       if (_validFunction(funcRef, clazz, staticContext)) {
         final isConstructor = _isConstructor(funcRef, clazz);
         // The VM shows setters as `<member>=`.
-        var name = funcRef.name.replaceAll('=', '');
+        var name = funcRef.name!.replaceAll('=', '');
         if (isConstructor) {
-          assert(name.startsWith(clazz.name));
-          if (name.length <= clazz.name.length + 1) continue;
-          name = name.substring(clazz.name.length + 1);
+          assert(name.startsWith(clazz.name!));
```
kenzieschmoll, line 552: pull clazz.name! out into a local

packages/devtools_app/lib/src/screens/debugger/file_search.dart
```diff
@@ -33,14 +31,14 @@ class FileSearchField extends StatefulWidget {
 
 class FileSearchFieldState extends State<FileSearchField>
     with AutoDisposeMixin, SearchFieldMixin {
-  AutoCompleteController autoCompleteController;
+  late AutoCompleteController autoCompleteController;
 
-  final _scriptsCache = <String, ScriptRef>{};
+  final _scriptsCache = <String?, ScriptRef>{};
```
kenzieschmoll, line 36: do we want to allow null keys?

packages/devtools_app/lib/src/screens/debugger/file_search.dart
```diff
@@ -175,61 +173,62 @@ class FileQuery {
   }
 
   AutoCompleteMatch createExactFileNameAutoCompleteMatch(ScriptRef script) {
-    if (isEmpty) return AutoCompleteMatch(script.uri);
+    if (isEmpty) return AutoCompleteMatch(script.uri!);
```
kenzieschmoll, line 176: pull script.uri! out into a local and use throughout this method

packages/devtools_app/lib/src/screens/debugger/file_search.dart
```diff
@@ -175,61 +173,62 @@ class FileQuery {
...
+          .every((token) => script.uri!.caseInsensitiveContains(token));
     }
 
-    return script.uri.caseInsensitiveContains(query);
+    return script.uri!.caseInsensitiveContains(query);
   }
 
   AutoCompleteMatch createExactFullPathAutoCompleteMatch(ScriptRef script) {
-    if (isEmpty) return AutoCompleteMatch(script.uri);
+    if (isEmpty) return AutoCompleteMatch(script.uri!);
```
kenzieschmoll, line 199: same comment here

packages/devtools_app/lib/src/screens/debugger/file_search.dart
```diff
@@ -175,61 +173,62 @@ class FileQuery {
...
+      return script.uri!.caseInsensitiveFuzzyMatch(query.replaceAll(' ', ''));
     }
 
-    return _fileName(script.uri).caseInsensitiveFuzzyMatch(query);
+    return _fileName(script.uri!).caseInsensitiveFuzzyMatch(query);
   }
 
   AutoCompleteMatch createFuzzyMatchAutoCompleteMatch(ScriptRef script) {
-    if (isEmpty) return AutoCompleteMatch(script.uri);
+    if (isEmpty) return AutoCompleteMatch(script.uri!);
```
kenzieschmoll, line 216: ditto

packages/devtools_app/lib/src/screens/debugger/program_explorer.dart
```diff
@@ -28,20 +26,20 @@ double get _selectedNodeTopSpacing => _programExplorerRowHeight * 3;
...
   const _ProgramExplorerRow({
-    @required this.controller,
-    @required this.node,
+    required this.controller,
+    required this.node,
     this.onTap,
   });
 
-  final ProgramExplorerController controller;
+  final ProgramExplorerController? controller;
```
kenzieschmoll, line 34: non-nullable?

packages/devtools_app/lib/src/screens/debugger/program_explorer.dart
```diff
@@ -155,12 +153,12 @@ class _ProgramExplorerRow extends StatelessWidget {
       buffer.write('>');
     }
     buffer.write('(');
-    String closingTag;
+    String? closingTag;
     for (int i = isInstanceMethod ? 1 : 0;
-        i < signature.parameters.length;
+        i < signature.parameters!.length;
```
kenzieschmoll, line 158: pull signature.parameters! out into a local

packages/devtools_app/lib/src/screens/debugger/program_explorer.dart
```diff
@@ -169,28 +167,28 @@ class _ProgramExplorerRow extends StatelessWidget {
           buffer.write('{');
         }
       }
-      if (param.required != null && param.required) {
+      if (param.required != null && param.required!) {
         buffer.write('required ');
       }
-      if (param.parameterType.name == null) {
-        buffer.write(_buildFunctionTypeText('Function', param.parameterType));
+      if (param.parameterType!.name == null) {
```
kenzieschmoll, line 173: pull param.parameterType! into local. Same for other cases where we find ourselves null asserting the same things multiple times

packages/devtools_app/lib/src/screens/debugger/program_explorer_model.dart
```diff
@@ -32,12 +30,12 @@ class VMServiceObjectNode extends TreeNode<VMServiceObjectNode> {
   static const packagePrefix = 'package:';
 
   final ProgramExplorerController controller;
-  final String name;
+  final String? name;
```
kenzieschmoll, line 33: can we make this non-nullable? perhaps defaulting to the empty string at the call site that may be passing null?

packages/devtools_app/lib/src/screens/debugger/program_explorer_model.dart
```diff
@@ -305,22 +303,23 @@ class VMServiceObjectNode extends TreeNode<VMServiceObjectNode> {
   }
 
   Future<void> populateLocation() async {
-    ScriptRef script = this.script;
-    int tokenPos = 0;
+    // ignore: unused_local_variable
+    ScriptRef? scriptRef = script;
```
kenzieschmoll, line 307: this smells a bit. We aren't using scriptRef everywhere in this method where we previously were using script. If we use scriptRef everywhere we were previously using script, we should be able to get rid of this ignore

packages/devtools_app/lib/src/screens/debugger/span_parser.dart
```diff
@@ -13,12 +11,12 @@ import 'package:string_scanner/string_scanner.dart';
 abstract class SpanParser {
   /// Takes a TextMate [Grammar] and a [String] and outputs a list of
   /// [ScopeSpan]s corresponding to the parsed input.
-  static List<ScopeSpan> parse(Grammar grammar, String src) {
+  static List<ScopeSpan> parse(Grammar? grammar, String src) {
```
kenzieschmoll, line 14: make non nullable

packages/devtools_app/lib/src/screens/debugger/span_parser.dart
```diff
@@ -120,27 +118,27 @@ class ScopeSpan {
...
 
-  bool contains(int token) => (_start <= token) && (token < _end);
+  bool contains(int token) => (_start! <= token) && (token < _end!);
 
   /// Splits the current [ScopeSpan] into multiple spans separated by [cond].
   /// This is useful for post-processing the results from a rule with a while
   /// condition as formatting should not be applied to the characters that
   /// match the while condition.
-  List<ScopeSpan> split(LineScanner scanner, RegExp cond) {
+  List<ScopeSpan> split(LineScanner scanner, RegExp? cond) {
```
kenzieschmoll, line 141: non nullable `cond`. We are unconditionally null asserting later

packages/devtools_app/lib/src/screens/debugger/span_parser.dart
```diff
@@ -213,9 +211,9 @@ class Repository {
     }
   }
 
-  final patterns = <String, List<_Matcher>>{};
+  final patterns = <String?, List<_Matcher>>{};
 
-  Map<String, dynamic> toJson() {
+  Map<String?, dynamic> toJson() {
```
kenzieschmoll, line 216: I'm skeptical anytime I see a nullable key type for a map

packages/devtools_app/lib/src/screens/debugger/codeview.dart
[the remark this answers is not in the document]
jacob314: ok so type inference before null safety just wasn't warning us. Might add a TODO to change all these lists to <InlineSpan> instead of <TextSpan> if that makes sense.

packages/devtools_app/lib/src/screens/debugger/span_parser.dart
```diff
@@ -612,8 +606,8 @@ class _IncludeMatcher extends _Matcher {
   }
 
   @override
-  List<ScopeSpan> scan(Grammar grammar, LineScanner scanner) {
-    final patterns = grammar.repository.patterns[include];
+  List<ScopeSpan>? scan(Grammar? grammar, LineScanner scanner) {
```
kenzieschmoll, line 609: non nullable grammar, probably here and elsewhere

packages/devtools_app/lib/src/screens/debugger/syntax_highlighter.dart
```diff
@@ -54,8 +52,8 @@ class SyntaxHighlighter {
     return TextSpan(
       children: _highlightLoopHelper(
         currentScope: null,
-        loopCondition: () => _currentPosition < source.length,
-        scopes: SpanParser.parse(_grammar, source),
+        loopCondition: () => _currentPosition < source!.length,
+        scopes: SpanParser.parse(_grammar, source!),
```
kenzieschmoll, line 56: hoist source into local

packages/devtools_app/lib/src/screens/debugger/codeview.dart
jacob314: @kenzieschmoll any thoughts on what color we should be using for cases like this?

jacob314: nm. related comment seems resolved.

packages/devtools_app/lib/src/screens/debugger/syntax_highlighter.dart
```diff
@@ -17,21 +15,21 @@ class SyntaxHighlighter {
...
   SyntaxHighlighter.withGrammar({
-    Grammar grammar,
+    Grammar? grammar,
     this.source,
   }) {
     _grammar = grammar;
   }
 
-  final String source;
+  final String? source;
```
kenzieschmoll, line 24: we seem to be null asserting this everywhere - should it be non-nullable?

packages/devtools_app/lib/src/screens/debugger/variables.dart
```diff
@@ -176,16 +175,16 @@ class VariableSelectionControls extends MaterialTextSelectionControls {
...
+    Offset? lastSecondaryTapDownPosition,
   ) {
     return _TextSelectionControlsToolbar(
       globalEditableRegion: globalEditableRegion,
       textLineHeight: textLineHeight,
       selectionMidpoint: selectionMidpoint,
       endpoints: endpoints,
       delegate: delegate,
-      clipboardStatus: clipboardStatus,
+      clipboardStatus: clipboardStatus!,
```
kenzieschmoll, line 187: either drop the null assert here or make the above parameter non nullable

packages/devtools_app/lib/src/screens/debugger/codeview.dart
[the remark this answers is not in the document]
jacob314: This cast seems wrong. We should unify on using List<InlineSpan> everywhere unless there is a strong reason we have to force just List<TextSpan>.

packages/devtools_test/lib/src/mocks.dart
```diff
@@ -1460,10 +1464,12 @@ final mockScriptRef = ScriptRef(
         'libraries/@783137924/scripts/package%3Agallery%2Fmain.dart/17b557e5bc3"',
     id: 'test-script-long-lines');
 
+final mockSyntaxHighlighter = SyntaxHighlighter.withGrammar(
+    grammar: mockGrammar, source: mockScript.source);
```
kenzieschmoll, line 1468: nit: tc

packages/devtools_app/lib/src/screens/debugger/call_stack.dart
```diff
@@ -14,29 +12,30 @@ import 'debugger_controller.dart';
...
+  const CallStack({Key? key}) : super(key: key);
 
   @override
   _CallStackState createState() => _CallStackState();
 }
 
 class _CallStackState extends State<CallStack> {
-  DebuggerController controller;
+  DebuggerController get controller => _controller!;
+  DebuggerController? _controller;
```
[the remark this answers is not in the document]
jacob314: discussed offline with @kenzieschmoll and lets stick with this pattern and move other cases where we used `late + bool` to this pattern.

packages/devtools_app/lib/src/screens/debugger/breakpoints.dart
bkonyi: Yeah, using `late` introduces an unavoidable sentinel check since we need to be able to throw a late initialization error if the variable isn't initialized. I still prefer this pattern over making `_controller` nullable if we never need to explicitly set the controller to `null` since the null assertions give the impression that the check _could_ fail, even if it never will.

packages/devtools_app/lib/src/screens/debugger/codeview.dart
kenzieschmoll: I think there are cases where it is okay to null-assert. For example, if we have this value defined in our theme, it will never be null in our app, even though it is nullable in the framework. So anywhere where we were previously assuming a theme value was non-null, it is safe to continue to assume that.

packages/devtools_app/lib/src/screens/debugger/breakpoints.dart
jacob314: `late` should give you the same impression that any accesses to the late variable could fail with a null assertion. There is no compile time checking that a late variable is initialized properly before use. `late` is just as much of a runtime feature as the getter+field pair. It is 95% just syntactic sugar so you don't have to manually write the getter + field pair.

elliette:
Doing a sanity check / clean up. Will resolve comments and notify when ready for review again!

packages/devtools_app/lib/src/screens/debugger/breakpoints.dart
elliette: Switched to the non-null getter / nullable private field pattern.

packages/devtools_app/lib/src/screens/debugger/codeview.dart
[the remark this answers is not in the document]
elliette: Added TODO

packages/devtools_app/lib/src/screens/debugger/codeview.dart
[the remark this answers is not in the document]
elliette: Might be misunderstanding - I added a TODO to switch to `List<InslineSpan` so that type casting isn't necessary

packages/devtools_app/lib/src/screens/debugger/debugger_model.dart
```diff
@@ -101,28 +100,24 @@ class ScriptLocation {
...
     );
   }
 
-  final String file;
-  final int line;
-  final int column;
-  final int tokenPos;
+  final String? file;
+  final int? line;
+  final int? column;
```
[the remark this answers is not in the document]
elliette: They can be null when SourcePosition is constructed with  the `calculatePosition` factory constructor, therefore keeping as nullable

packages/devtools_app/lib/src/screens/debugger/variables.dart
elliette: Keeping as is, we are overriding Flutter's `buildToolbar` method so we can't make the parameter non-nullable.

packages/devtools_app/lib/src/screens/debugger/program_explorer_model.dart
elliette: Oops! Ignore is now gone. We were setting a variable with type `ScriptRef` to type `Script`. Refactored so they are now separate variables.

packages/devtools_app/lib/src/screens/debugger/file_search.dart
elliette: Nope, changed to no null keys :)

packages/devtools_app/lib/src/screens/debugger/evaluate.dart
elliette: Nope, done!

packages/devtools_app/lib/src/screens/debugger/debugger_model.dart
```diff
@@ -720,10 +715,10 @@ class DartObjectNode extends TreeNode<DartObjectNode> {
   /// [value] should typically be an [InstanceRef] but can also be a [Sentinel]
   /// [ObjRef] or primitive type such as num or String.
   factory DartObjectNode.fromValue({
-    String name = '',
-    @required Object value,
-    RemoteDiagnosticsNode diagnostic,
-    @required IsolateRef isolateRef,
+    String? name = '',
```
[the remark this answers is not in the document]
elliette: Seems like default value is only if no value is passed, not if null value is passed. Modified to set name to an empty string if the parameter is null

packages/devtools_app/lib/src/screens/debugger/debugger_model.dart
```diff
@@ -431,11 +426,11 @@ Future<void> buildVariablesTree(
   }
   if (diagnostic != null && includeDiagnosticChildren) {
     // Always add children last after properties to avoid confusion.
-    final ObjectGroupBase service = diagnostic.inspectorService;
+    final ObjectGroupBase? service = diagnostic.inspectorService;
     final diagnosticChildren = await diagnostic.children;
     if (diagnosticChildren?.isNotEmpty ?? false) {
       final childrenNode = DartObjectNode.text(
-        pluralize('child', diagnosticChildren.length, plural: 'children'),
+        pluralize('child', diagnosticChildren!.length, plural: 'children'),
```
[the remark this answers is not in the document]
elliette: Hmm seems like it's not smart enough to understand that the value is not null in that case, still need the null assert :(

packages/devtools_app/lib/src/screens/debugger/debugger_model.dart
```diff
@@ -353,7 +348,7 @@ Future<void> buildVariablesTree(
   if (diagnostic != null && includeDiagnosticPropertiesInDebugger) {
     final service = diagnostic.inspectorService;
     Future<void> _addPropertiesHelper(
-        List<RemoteDiagnosticsNode> properties) async {
+        List<RemoteDiagnosticsNode?>? properties) async {
```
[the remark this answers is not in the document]
elliette: Nope :)

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
elliette: Nope!

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
elliette: Removed the cast!

kenzieschmoll (approved):

PR #3928, opened 2022-03-26
Migrate inspector to null safety.

polina-c:


packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
```diff
@@ -36,23 +36,23 @@ const bool _showRenderObjectPropertiesAsLinks = false;
...
     this.errorText,
     this.multiline = false,
     this.style,
-    @required this.debuggerController,
+    required this.debuggerController,
     this.nodeDescriptionHighlightStyle,
   });
 
-  final RemoteDiagnosticsNode diagnostic;
+  final RemoteDiagnosticsNode? diagnostic;
```
polina-c, line 48: InspectorTreeController passes InspectorTreeNode.diagnostics here, that is nullable.

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
```diff
@@ -36,23 +36,23 @@ const bool _showRenderObjectPropertiesAsLinks = false;
...
   final bool isSelected;
-  final String errorText;
-  final String searchValue;
+  final String? errorText;
+  final String? searchValue;
```
polina-c, line 51: not specified in variables.dart in constructor

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
```diff
@@ -36,23 +36,23 @@ const bool _showRenderObjectPropertiesAsLinks = false;
...
   final bool multiline;
-  final TextStyle style;
+  final TextStyle? style;
   final DebuggerController debuggerController;
-  final TextStyle nodeDescriptionHighlightStyle;
+  final TextStyle? nodeDescriptionHighlightStyle;
```
polina-c, line 55: not specified in variables.dart in constructor

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
```diff
@@ -36,23 +36,23 @@ const bool _showRenderObjectPropertiesAsLinks = false;
...
-  final RemoteDiagnosticsNode diagnostic;
+  final RemoteDiagnosticsNode? diagnostic;
   final bool isSelected;
-  final String errorText;
-  final String searchValue;
+  final String? errorText;
+  final String? searchValue;
   final bool multiline;
-  final TextStyle style;
+  final TextStyle? style;
```
polina-c, line 53: not specified in inspector_tree_controller.dart in constructor call

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
```diff
@@ -369,7 +369,7 @@ class DiagnosticsNodeDescription extends StatelessWidget {
 
   TextSpan _buildHighlightedSearchPreview(
     String textPreview,
-    String searchValue,
+    String? searchValue,
```
polina-c, line 372: null passed in _buildDescriptionTextSpans

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -70,22 +68,21 @@ class InspectorController extends DisposableController
     with AutoDisposeControllerMixin
     implements InspectorServiceClient {
   InspectorController({
-    @required this.inspectorTree,
-    InspectorTreeController detailsTree,
-    @required this.treeType,
+    required this.inspectorTree,
+    InspectorTreeController? detailsTree,
```
polina-c, line 72: detailsTree is nullable as it is not specified in invocation below

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -97,7 +94,7 @@ class InspectorController extends DisposableController
     );
     if (isSummaryTree) {
       details = InspectorController(
-        inspectorTree: detailsTree,
+        inspectorTree: detailsTree!,
```
polina-c, line 97: InspectorController treats inspectorTree as not nullable, so null-checking here

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -143,7 +140,7 @@ class InspectorController extends DisposableController
     serviceManager.consoleService.ensureServiceInitialized();
   }
 
-  void _handleConnectionStart(VmService service) {
+  void _handleConnectionStart(VmService? service) {
```
polina-c, line 143: Parameter is nullable because of handler signature

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -163,7 +160,7 @@ class InspectorController extends DisposableController
     }
   }
 
-  IsolateRef _mainIsolate;
+  IsolateRef? _mainIsolate;
```
polina-c, line 163: Naullable, because it is cached value

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -194,46 +191,46 @@ class InspectorController extends DisposableController
...
 
   bool _disposed = false;
 
-  RateLimiter _refreshRateLimiter;
+  late RateLimiter _refreshRateLimiter;
 
   /// Groups used to manage and cancel requests to load data to display directly
   /// in the tree.
-  InspectorObjectGroupManager _treeGroups;
+  InspectorObjectGroupManager? _treeGroups;
```
polina-c, line 207: assigned to null in dispose

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -194,46 +191,46 @@ class InspectorController extends DisposableController
...
 
   /// Groups used to manage and cancel requests to determine what the current
   /// selection is.
   ///
   /// This group needs to be kept separate from treeGroups as the selection is
   /// shared more with the details subtree.
   /// TODO(jacobr): is there a way we can unify the selection and tree groups?
-  InspectorObjectGroupManager _selectionGroups;
+  InspectorObjectGroupManager? _selectionGroups;
```
polina-c, line 215: assigned to null in dispose

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -194,46 +191,46 @@ class InspectorController extends DisposableController
   final bool isSummaryTree;
 
   /// Parent InspectorController if this is a details subtree.
-  InspectorController parent;
+  InspectorController? parent;
```
polina-c, line 194: Nullable, because assigned externally.

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -194,46 +191,46 @@ class InspectorController extends DisposableController
...
 
-  InspectorController details;
+  InspectorController? details;
```
polina-c, line 196: Nullable, because assigned externally.

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -194,46 +191,46 @@ class InspectorController extends DisposableController
...
 
-  set currentShowNode(InspectorTreeNode node) => inspectorTree.hover = node;
+  set currentShowNode(InspectorTreeNode? node) => inspectorTree.hover = node;
 
   bool flutterAppFrameReady = false;
 
   bool treeLoadStarted = false;
 
-  RemoteDiagnosticsNode subtreeRoot;
+  RemoteDiagnosticsNode? subtreeRoot;
```
polina-c, line 226: nullable because it is assigned to null in shutdownTree

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -143,58 +143,58 @@ class InspectorTreeController extends Object
...
+  InspectorTreeRow? _cachedSelectedRow;
 
   /// All cached rows of the tree.
   ///
   /// Similar to [cachedRows] but:
   /// * contains every row in the tree (including collapsed rows)
   /// * items don't change when nodes are expanded or collapsed
   /// * items are populated only when root is changed
-  final _searchableCachedRows = <InspectorTreeRow>[];
+  final _searchableCachedRows = <InspectorTreeRow?>[];
```
polina-c, line 197: Populated with nulls in getCashedRow

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -143,58 +143,58 @@ class InspectorTreeController extends Object
...
-  InspectorTreeNode _hover;
+  InspectorTreeNode? get hover => _hover;
+  InspectorTreeNode? _hover;
 
-  double lastContentWidth;
+  double? lastContentWidth;
 
-  final List<InspectorTreeRow> cachedRows = [];
-  InspectorTreeRow _cachedSelectedRow;
+  final List<InspectorTreeRow?> cachedRows = [];
```
polina-c, line 188: Populated with nulls in getCashedRow

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -700,27 +701,27 @@ extension RemoteDiagnosticsNodeExtension on RemoteDiagnosticsNode {
...
+    required this.debuggerController,
     this.inspectorTreeController,
     this.isSummaryTree = false,
     this.widgetErrors,
   })  : assert(isSummaryTree == (inspectorTreeController == null)),
         super(key: key);
 
-  final InspectorTreeController controller;
-  final InspectorTreeController inspectorTreeController;
+  final InspectorTreeController? controller;
```
polina-c, line 720: Assigned to nullable value in InspectorScreen.build.

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -700,27 +701,27 @@ extension RemoteDiagnosticsNodeExtension on RemoteDiagnosticsNode {
...
+  final InspectorTreeController? inspectorTreeController;
```
polina-c, line 721: Nullable because Is not provided by inspector screen.

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -853,43 +842,48 @@ class InspectorController extends DisposableController
...
               .selectAndShowNode(firstAncestorInParentTree(selectedNode.value));
         }
       }
     }
   }
 
-  RemoteDiagnosticsNode firstAncestorInParentTree(InspectorTreeNode node) {
-    if (parent == null) {
-      return node.diagnostic;
+  RemoteDiagnosticsNode? firstAncestorInParentTree(InspectorTreeNode? node) {
```
polina-c, line 866: Nulls passed here in selectionChanged.

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
```diff
@@ -19,11 +19,11 @@ import 'inspector_controller.dart';
 import 'inspector_text_styles.dart' as inspector_text_styles;
 import 'inspector_tree.dart';
 
-final ColorIconMaker _colorIconMaker = ColorIconMaker();
-final CustomIconMaker _customIconMaker = CustomIconMaker();
-final CustomIcon defaultIcon = _customIconMaker.fromInfo('Default');
+final _colorIconMaker = ColorIconMaker();
+final _customIconMaker = CustomIconMaker();
+final defaultIcon = _customIconMaker.fromInfo('Default') as CustomIcon;
```
jacob314, line 24: why is this as check required?

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
```diff
@@ -112,13 +113,8 @@ class DiagnosticsNodeDescription extends StatelessWidget {
...
 
-  Widget buildDescription(
-    String description,
-    TextStyle textStyle,
-    BuildContext context,
-    ColorScheme colorScheme, {
-    bool isProperty,
-  }) {
+  Widget buildDescription(String description, TextStyle textStyle,
+      BuildContext context, ColorScheme colorScheme) {
```
jacob314, line 117: add back trailing comma.

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
jacob314: nit: is a null service actually means the service is stopped, add a TODO to rename the API to be something more accurate.

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
jacob314: also not all controllers have a parent controller. only the details one does.

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
jacob314: for things that are only null due to being nulled in dispose, add a getter that is non-nullable to avoid polluting the code due to things nulled in dispose.

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
jacob314: add  non-nullable getter. otherwise we are polluting the code with null checks that are only needed if there is use after dispose.

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
polina-c: done

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
jacob314: This is the one usage of this method. Make it non-nullable as the null path doesn't make sense anymore

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -118,14 +118,14 @@ class InspectorTreeController extends Object
     final firstClient = _clients.isEmpty;
     _clients.add(value);
     if (firstClient) {
-      config.onClientActiveChange?.call(true);
+      config!.onClientActiveChange?.call(true);
```
jacob314, line 121: make config late so it can be null safe

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -94,9 +94,9 @@ class _InspectorTreeRowState extends State<_InspectorTreeRowWidget>
     setState(() {
       final row = widget.row;
       if (expanded) {
-        widget.inspectorTreeState.controller.onExpandRow(row);
+        widget.inspectorTreeState.controller!.onExpandRow(row);
```
jacob314, line 97: make this controller also  non-nullable perhaps using late.

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -377,25 +383,27 @@ class InspectorTreeController extends Object
     root.children.forEach(_collapseAllNodes);
   }
 
-  int get numRows => root != null ? root.subtreeSize : 0;
+  int get numRows => root?.subtreeSize ?? 0;
```
jacob314, line 386: good cleanup

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -377,25 +383,27 @@ class InspectorTreeController extends Object
...
 
   int getRowIndex(double y) => max(0, (y - verticalPadding) ~/ rowHeight);
 
-  InspectorTreeRow getRowForNode(InspectorTreeNode node) {
-    return getCachedRow(root.getRowIndex(node));
+  InspectorTreeRow? getRowForNode(InspectorTreeNode node) {
+    return getCachedRow(root!.getRowIndex(node));
```
jacob314, line 391: rather than using `root!`
return null early if root is null That is more coherent with the overall method that can return null.

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -377,25 +383,27 @@ class InspectorTreeController extends Object
...
+    final rootLocal = root;
+    if (rootLocal == null) return null;
     final int row = getRowIndex(offset.dy);
-    return row < root.subtreeSize ? getCachedRow(row) : null;
+    return row < rootLocal.subtreeSize ? getCachedRow(row) : null;
   }
 
   void onExpandRow(InspectorTreeRow row) {
     setState(() {
+      final configLocal = config!;
```
jacob314, line 403: adds configLocal but dosn't benefit from it.
I assume what you meant to do was add
onExpandLocal.

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
jacob314: Add a TODO here. something is wrong. One of controller and inspectorTreeController isn't really used.

jacob314: Also add a todo to follow up making this non-nullable. I think null checks may need to be added to InspectorScreen. Would be cleaner if null values aren't allowed to flow in here. Seems like all the usages are assuming this is non-nullable.

packages/devtools_app/lib/src/screens/inspector/layout_explorer/flex/flex.dart
```diff
@@ -42,41 +42,41 @@ class _FlexLayoutExplorerWidgetState extends LayoutExplorerWidgetState<
     FlexLayoutExplorerWidget, FlexLayoutProperties> {
   final scrollController = ScrollController();
 
-  Axis get direction => properties.direction;
+  Axis get direction => properties!.direction;
```
jacob314, line 45: properties should be non-nullable.

packages/devtools_app/lib/src/screens/inspector/layout_explorer/flex/flex.dart
```diff
@@ -42,41 +42,41 @@ class _FlexLayoutExplorerWidgetState extends LayoutExplorerWidgetState<
...
           ? colorScheme.mainAxisTextColor
           : colorScheme.crossAxisTextColor;
 
-  String get flexType => properties.type;
+  String get flexType => properties!.type;
 
   @override
-  RemoteDiagnosticsNode getRoot(RemoteDiagnosticsNode node) {
+  RemoteDiagnosticsNode? getRoot(RemoteDiagnosticsNode? node) {
     if (!shouldDisplay(node)) return null;
```
jacob314, line 73: add an explicit 
if node == null return null
consider making shouldDisplay take a non-nullable node.
That style is a bit more consistent with null safety than the previous style of having a few too many methods that implicitly check for null to be safe.

packages/devtools_app/lib/src/screens/inspector/layout_explorer/flex/flex.dart
```diff
@@ -190,8 +193,10 @@ class _FlexLayoutExplorerWidgetState extends LayoutExplorerWidgetState<
                       Flexible(
                         child: Image.asset(
                           (axis == direction)
-                              ? mainAxisAssetImageUrl(direction, alignment)
-                              : crossAxisAssetImageUrl(direction, alignment),
+                              ? mainAxisAssetImageUrl(
+                                  direction, alignment as MainAxisAlignment)
```
jacob314, line 197: why are these as checks now needed?

packages/devtools_app/lib/src/screens/inspector/layout_explorer/flex/flex.dart
```diff
@@ -201,22 +206,22 @@ class _FlexLayoutExplorerWidgetState extends LayoutExplorerWidgetState<
...
             // newSelection is an object instead of type here because
             // the type is dependent on the `axis` parameter
             // if the axis is the main axis the type should be [MainAxisAlignment]
             // if the axis is the cross axis the type should be [CrossAxisAlignment]
             FlexLayoutProperties changedProperties;
             if (axis == direction) {
-              changedProperties =
-                  properties.copyWith(mainAxisAlignment: newSelection);
+              changedProperties = propertiesLocal.copyWith(
+                  mainAxisAlignment: newSelection as MainAxisAlignment?);
```
jacob314, line 217: more cases with as checks that should be removed or replaced with `!` operators.

packages/devtools_app/lib/src/screens/inspector/layout_explorer/flex/flex.dart
```diff
@@ -540,34 +547,36 @@ class FlexChildVisualizer extends StatelessWidget {
 
   final RenderProperties renderProperties;
 
-  FlexLayoutProperties get root => state.properties;
+  FlexLayoutProperties? get root => state.properties;
```
jacob314, line 550: make these both non-nullable even if it means adding a bang. 
Consider adding a TODO to refactor code to eliminate the bang. The core logic of this visualizer doesn't make sense if these are null.

packages/devtools_app/lib/src/screens/inspector/layout_explorer/flex/utils.dart
```diff
@@ -70,8 +68,8 @@ class AnimatedFlexLayoutProperties
           size: Size.lerp(beginProps.size, endProps.size, t),
           realSize: Size.lerp(beginProps.realSize, endProps.realSize, t),
           layoutProperties: AnimatedLayoutProperties(
-            beginProps.layoutProperties,
-            endProps.layoutProperties,
+            beginProps.layoutProperties!,
```
jacob314, line 71: this also seems like a sign that the beginProps should have been non-nullable.

packages/devtools_app/lib/src/screens/inspector/layout_explorer/flex/utils.dart
```diff
@@ -123,19 +121,19 @@ class AnimatedFlexLayoutProperties
...
   @override
-  TextBaseline get textBaseline => end.textBaseline;
+  TextBaseline? get textBaseline => end.textBaseline;
 
   @override
   TextDirection get textDirection => end.textDirection;
 
   @override
-  double get totalFlex =>
+  double? get totalFlex =>
```
jacob314, line 136: does this have to be null? lerpDouble is a bit strangely nullable but not an api we control. We should probably create our own version that is null safe as the current behavior seems only reasonable given dart:ui wanted to avoid breaking changes when migrating to null safety.

packages/devtools_app/lib/src/screens/inspector/layout_explorer/ui/layout_explorer_widget.dart
```diff
@@ -126,13 +126,13 @@ abstract class LayoutExplorerWidgetState<W extends LayoutExplorerWidget,
   String? id(RemoteDiagnosticsNode? node) => node?.dartDiagnosticRef.id;
 
   void _registerInspectorControllerService() {
-    inspectorController.selectedNode?.addListener(_onSelectionChangedCallback);
+    inspectorController.selectedNode.addListener(_onSelectionChangedCallback);
```
jacob314, line 129: glad null safety migration made it easy to catch this silly code.

jacob314:
Done with full round of review comments. Non comments are serious, just a few cases to push on to make sure we are making the right members non-nullable.

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
polina-c: hm. Today I could change it. May be because of recent upgrade to next version of SDK.

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
polina-c: They can turn out to be already nulls when a method is invoked. In this case usage of getter will be error prone.

So I prefer:
In methods, that assume the field is not null yet, to assign the field to local variable.
```
final treeGroupsLocal = _treeGroups!;
```

For methods that are ok with them to be null, to use null aware:
```
await _treeGroups?.pendingUpdateDone;
```

What do you think?

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
polina-c: see https://github.com/flutter/devtools/pull/3928#discussion_r838787666

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_tree.dart
```diff
@@ -95,22 +93,25 @@ class InspectorTreeNode {
   }
 
   bool get shouldShow {
-    _shouldShow ??= parent == null || parent.isExpanded && parent.shouldShow;
-    return _shouldShow;
+    _shouldShow ??= parent == null || parent!.isExpanded && parent!.shouldShow;
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
polina-c: There is assert in setter that seems to be important:
```
assert(_config == null);
```

Late fields cannot be asserted for null.
There is an alternative to introduce field `_isConfigInitialized`, but is seems to create comparable pollution.

What do you think?

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
polina-c: This controller is sometimes null, so cannot be late. It is always non-null if the node is expanded.

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -700,27 +701,27 @@ extension RemoteDiagnosticsNodeExtension on RemoteDiagnosticsNode {
 abstract class InspectorControllerClient {
   void onChanged();
 
-  void scrollToRect(Rect rect);
+  void scrollToRect(Rect? rect);
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/layout_explorer/flex/flex.dart
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
polina-c: we want to make it late final

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
polina-c: The controller can be null: https://github.com/flutter/devtools/blob/2a707ca56c1a9d5eeef212c28c573548a051fdd2/packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart#L964

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
polina-c: done

packages/devtools_app/lib/src/screens/inspector/layout_explorer/flex/flex.dart
polina-c: done

packages/devtools_app/lib/src/screens/inspector/layout_explorer/flex/flex.dart
polina-c: because alignment is declared as Object

packages/devtools_app/lib/src/screens/inspector/layout_explorer/flex/flex.dart
polina-c: done

packages/devtools_app/lib/src/screens/inspector/layout_explorer/flex/utils.dart
polina-c: there are places in code where RenderProperties.layoutProperties are validated for nullness and, in case of null, other properties are used.
This exact method does not do this.
Added TODO.

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
```diff
@@ -172,40 +172,44 @@ class DiagnosticsNodeDescription extends StatelessWidget {
...
-      final String propertyType = diagnostic.propertyType;
-      final Map<String, Object> properties = diagnostic.valuePropertiesJson;
+      final propertyType = diagnosticLocal.propertyType;
+      final properties = diagnosticLocal.valuePropertiesJson;
 
-      if (name?.isNotEmpty == true && diagnostic.showName) {
-        children.add(Text('$name${diagnostic.separator} ', style: textStyle));
+      if (name?.isNotEmpty == true && diagnosticLocal.showName) {
+        children
+            .add(Text('$name${diagnosticLocal.separator} ', style: textStyle));
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -70,22 +68,21 @@ class InspectorController extends DisposableController
...
   InspectorController({
-    @required this.inspectorTree,
-    InspectorTreeController detailsTree,
-    @required this.treeType,
+    required this.inspectorTree,
+    InspectorTreeController? detailsTree,
+    required this.treeType,
     this.parent,
     this.isSummaryTree = true,
   })  : _treeGroups = InspectorObjectGroupManager(
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -290,26 +287,27 @@ class InspectorController extends DisposableController
...
+      node = findMatchingInspectorTreeNode(parent?.selectedDiagnostic);
     }
 
     currentShowNode = node;
     return true;
   }
 
-  InspectorTreeNode findMatchingInspectorTreeNode(RemoteDiagnosticsNode node) {
+  InspectorTreeNode? findMatchingInspectorTreeNode(
+      RemoteDiagnosticsNode? node) {
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -290,26 +287,27 @@ class InspectorController extends DisposableController
...
     if (node?.valueRef == null) {
       return null;
     }
-    return valueToInspectorTreeNode[node.valueRef];
+    return valueToInspectorTreeNode[node?.valueRef];
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -290,26 +287,27 @@ class InspectorController extends DisposableController
...
   }
 
-  Future<void> getPendingUpdateDone() async {
+  Future<void> _makePendingUpdateDone() async {
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/diagnostics_node.dart
```diff
@@ -61,27 +61,23 @@ class RemoteDiagnosticsNode extends DiagnosticableTree {
...
       double.parse(json['height'] as String),
     );
   }
 
-  static FlexFit? deserializeFlexFit(String? flexFit) {
-    if (flexFit == null) {
-      return null;
-    } else if (flexFit == 'tight') return FlexFit.tight;
+  static FlexFit deserializeFlexFit(String? flexFit) {
+    if (flexFit == 'tight') return FlexFit.tight;
```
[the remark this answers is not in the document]
kenzieschmoll: my question was, is null a valid return for this method? If consumers weren't doing things like returning early or assigning some default value for a null return, then changing this method to _not_ return null may not be a no-op, but rather a functional change.

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -248,12 +245,12 @@ class InspectorController extends DisposableController
...
 
-  RemoteDiagnosticsNode get selectedDiagnostic =>
+  RemoteDiagnosticsNode? get selectedDiagnostic =>
       selectedNode.value?.diagnostic;
 
-  final ValueNotifier<int> _selectedErrorIndex = ValueNotifier<int>(null);
+  final ValueNotifier<int?> _selectedErrorIndex = ValueNotifier<int?>(null);
 
-  ValueListenable<int> get selectedErrorIndex => _selectedErrorIndex;
+  ValueListenable<int?> get selectedErrorIndex => _selectedErrorIndex;
```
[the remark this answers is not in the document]
kenzieschmoll: making this non-nullable allows us to get rid of several null checks and asserts elsewhere. Also, I would argue that the fact that `-1` is a standard for the non-existing-index would be a reason _for_ using it here over `null`. -1 is an intuitive pattern for the index that does not exist.

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -397,11 +394,12 @@ class InspectorController extends DisposableController
...
 
-  InspectorService get inspectorService => serviceManager.inspectorService;
+  InspectorService get inspectorService =>
+      serviceManager.inspectorService as InspectorService;
 
   List<String> get rootDirectories =>
-      _rootDirectories ?? parent.rootDirectories;
-  List<String> _rootDirectories;
+      _rootDirectories ?? parent!.rootDirectories;
+  List<String>? _rootDirectories;
```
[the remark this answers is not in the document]
polina-c: The property was defined 1.5 years ago, so I do not think it is needed. Deleted.

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -520,19 +517,19 @@ class InspectorController extends DisposableController
...
       return null;
     }
-    return valueToInspectorTreeNode[subtreeRoot.valueRef];
+    return valueToInspectorTreeNode[subtreeRoot!.valueRef];
   }
 
-  void refreshSelection(RemoteDiagnosticsNode newSelection,
-      RemoteDiagnosticsNode detailsSelection, bool setSubtreeRoot) {
+  void refreshSelection(RemoteDiagnosticsNode? newSelection,
+      RemoteDiagnosticsNode? detailsSelection, bool setSubtreeRoot) {
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -544,7 +541,7 @@ class InspectorController extends DisposableController
     if (details != null) {
       if (subtreeRoot != null && getSubtreeRootNode() == null) {
         subtreeRoot = newSelection;
-        details.setSubtreeRoot(newSelection, detailsSelection);
+        details!.setSubtreeRoot(newSelection, detailsSelection);
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -632,20 +623,22 @@ class InspectorController extends DisposableController
   }
 
   Future<void> updateSelectionFromService(
-      {@required bool firstFrame, String inspectorRef}) async {
+      {required bool firstFrame, String? inspectorRef}) async {
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -754,9 +747,9 @@ class InspectorController extends DisposableController
     lastExpanded = null; // New selected node takes precedence.
     endShowNode();
     if (details != null) {
-      details.endShowNode();
+      details!.endShowNode();
     } else if (parent != null) {
-      parent.endShowNode();
+      parent!.endShowNode();
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/diagnostics_node.dart
```diff
@@ -133,13 +129,13 @@ class RemoteDiagnosticsNode extends DiagnosticableTree {
...
 
-  BoxConstraints? get constraints =>
-      deserializeConstraints(json['constraints'] as Map<String, Object>?);
+  BoxConstraints get constraints => deserializeConstraints(
+      json['constraints'] as Map<String, Object>? ?? {})!;
 
-  BoxParentData? get parentData =>
-      deserializeParentData(json['parentData'] as Map<String, Object>?);
+  BoxParentData get parentData =>
+      deserializeParentData(json['parentData'] as Map<String, Object>? ?? {});
```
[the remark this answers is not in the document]
jacob314: Never mind. For this case, making it non-nullable is fine. The empty BoxParentData with an offset 0,0 is a fine default. The degree it differs from the flutter framework (not everything is a box) does not matter.

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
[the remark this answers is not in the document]
jacob314: I agree with @polina-c on this case. Using null for a missing index is better because you get language support to make sure you aren't accessing a missing index. With -1, you have to make sure you are checking for -1 in all the right places.

packages/devtools_app/lib/src/screens/inspector/inspector_tree.dart
```diff
@@ -246,14 +249,11 @@ class InspectorTreeNode {
         depth++;
       }
     }
-    assert(false); // internal error.
-    return null;
```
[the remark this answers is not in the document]
jacob314: this is now dead code.

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -68,15 +68,16 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
 
-  InspectorTreeController get summaryTreeController =>
-      inspectorController?.inspectorTree;
+  InspectorTreeController? get summaryTreeController =>
+      inspectorController.inspectorTree;
 
-  InspectorTreeController get detailsTreeController =>
-      inspectorController?.details?.inspectorTree;
+  InspectorTreeController? get detailsTreeController =>
+      inspectorController.details?.inspectorTree;
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -242,18 +248,21 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+                              as LinkedHashMap<String, InspectableWidgetError>?,
                           debuggerController: debuggerController,
                         ),
-                        if (errors.isNotEmpty && inspectorController != null)
+                        if (errors.isNotEmpty)
                           ValueListenableBuilder(
                             valueListenable:
                                 inspectorController.selectedErrorIndex,
-                            builder: (_, selectedErrorIndex, __) => Positioned(
+                            builder: (_, dynamic selectedErrorIndex, __) =>
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -242,18 +248,21 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
                           key: summaryTreeKey,
                           controller: summaryTreeController,
                           isSummaryTree: true,
-                          widgetErrors: inspectableErrors,
+                          widgetErrors: inspectableErrors
+                              as LinkedHashMap<String, InspectableWidgetError>?,
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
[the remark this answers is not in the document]
polina-c: null checks help us to catch unexpected non-existing-index early, and analyzer will give warnings at compile time. -1 will cause out of range exception or unexpected behavior at run time, if not caught early.
null checks seem to be beneficial in this case

packages/devtools_app/lib/src/screens/inspector/inspector_tree.dart
```diff
@@ -95,22 +93,25 @@ class InspectorTreeNode {
...
+    return _shouldShow!;
   }
 
-  bool _shouldShow;
+  bool? _shouldShow;
 
   bool selected = false;
 
-  RemoteDiagnosticsNode _diagnostic;
+  RemoteDiagnosticsNode? _diagnostic;
```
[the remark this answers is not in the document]
polina-c: no, as there is null-allow code for it in couple places and there is no guarantee it will be set

packages/devtools_app/lib/src/screens/inspector/inspector_tree.dart
```diff
@@ -134,23 +135,24 @@ class InspectorTreeNode {
...
-    _isExpanded = v.childrenReady;
+  set diagnostic(RemoteDiagnosticsNode? v) {
+    final value = v!;
+    _diagnostic = value;
+    _isExpanded = value.childrenReady;
     isDirty = true;
   }
 
-  int get childrenCount {
+  int? get childrenCount {
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_tree.dart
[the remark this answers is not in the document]
polina-c: yes, we never break from while (true) above

packages/devtools_app/lib/src/screens/inspector/inspector_tree.dart
```diff
@@ -246,14 +249,11 @@ class InspectorTreeNode {
...
       }
     }
-    assert(false); // internal error.
-    return null;
   }
 
   void removeChild(InspectorTreeNode child) {
     child.parent = null;
-    final removed = _children.remove(child);
-    assert(removed != null);
```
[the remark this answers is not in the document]
polina-c: remove is now not nullable

packages/devtools_app/lib/src/screens/inspector/diagnostics_node.dart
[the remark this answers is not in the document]
polina-c: I think we resolved the question here: https://github.com/flutter/devtools/pull/3928#discussion_r838688505

packages/devtools_app/lib/src/screens/inspector/inspector_tree.dart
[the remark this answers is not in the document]
polina-c: returned assert(remove)

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
kenzieschmoll: with more context of the other null checks in this widget, I agree with Jacob. Add a non-nullable parameter to the _InspectorTreeRowWidget for this controller. Then that can be used both in this method and in the build method above.

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -143,58 +143,52 @@ class InspectorTreeController extends Object
...
-  InspectorTreeNode _hover;
+  InspectorTreeNode? get hover => _hover;
+  InspectorTreeNode? _hover;
 
-  double lastContentWidth;
+  double? lastContentWidth;
 
-  final List<InspectorTreeRow> cachedRows = [];
-  InspectorTreeRow _cachedSelectedRow;
+  final List<InspectorTreeRow?> cachedRows = [];
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -643,10 +641,9 @@ class InspectorTreeController extends Object
     int _debugStatsSearchOps = 0;
     final _debugStatsWidgets = _searchableCachedRows.length;
 
-    if (search == null ||
-        search.isEmpty ||
+    if (search.isEmpty ||
         serviceManager.inspectorService == null ||
-        serviceManager.inspectorService.isDisposed) {
+        serviceManager.inspectorService!.isDisposed) {
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -933,16 +931,16 @@ class _InspectorTreeState extends State<InspectorTree>
     if (event is! RawKeyDownEvent) return KeyEventResult.ignored;
 
     if (event.logicalKey == LogicalKeyboardKey.arrowDown) {
-      controller.navigateDown();
+      controller!.navigateDown();
```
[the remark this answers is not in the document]
polina-c: every check is for it's own condition, so I will not save anything

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -1075,7 +1075,7 @@ Paint _defaultPaint(ColorScheme colorScheme) => Paint()
 class _RowPainter extends CustomPainter {
   _RowPainter(this.row, this._controller, this.colorScheme);
 
-  final InspectorTreeController _controller;
+  final InspectorTreeController? _controller;
```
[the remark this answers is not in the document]
polina-c: yes

kenzieschmoll (approved):

Issue #3802, opened 2022-03-08
Use `package:` uris instead of `file://` uris in CPU profiler


kenzieschmoll:
additionally, we should convert `org-dartlang-sdk` to `dart`.

Requested by @scheglov

scheglov:
Thank you for tracking this!

jacob314:
@bkonyi's API surfacing the mapping between file uris and pretty dart uris solves this exact issue and similar.
Fyi @helin24 who was just using it in IntelliJ.

kenzieschmoll:
related https://github.com/flutter/devtools/issues/3865

kenzieschmoll:
When using the `lookupPackageUris` and `lookupResolvedPackageUris` APIs in DevTools, show a warning if a user is using a vm service protocol version < 7.4.0

kenzieschmoll:
Here is an example of the Flutter IntelliJ plugin using these APIs: https://github.com/flutter/flutter-intellij/pull/6044

PR #3932, opened 2022-03-30
Using lookupPackageUris for stack frame processedUrls

CoderDake:
Resolves: https://github.com/flutter/devtools/issues/3802

# Description

This change makes it so that the devtool source column will show shortened package uris instead of absolute paths.

The uris in the source column will now look something like this:
Screen Shot 2022-04-14 at 2 16 31 PM

# Code Walk Through
- [url_utils](https://github.com/flutter/devtools/pull/3932/files#diff-c5ac6b87801a471fffd57bd29dbf44d520a19840c9d49bfa8a55028fb95a384eL13) is no longer needed so it is being deleted
- `processedUrl` has been renamed to `packageUri` in a bunch of places
- [ResolvedUriManager has been created](https://github.com/flutter/devtools/pull/3932/files#diff-05cf548d9abef51d628ba9e197bcf38608d7f299c8a8cbce89f49be36ae37739R5) to accomodate caching resolvedUrl to packageUri mappings and to facilitate the mapping look up
- `_addPackageUrisToTraceObject` now has been added [after the traceObject has been created](https://github.com/flutter/devtools/pull/3932/files#diff-8ea419af3b96202def061b93440e5b119c49aba8132769f83c9289086559704dR365)
- [_addPackageUrisToTraceObject](https://github.com/flutter/devtools/pull/3932/files#diff-8ea419af3b96202def061b93440e5b119c49aba8132769f83c9289086559704dR377) passes over the stack frames twice
    - [1st pass](https://github.com/flutter/devtools/pull/3932/files#diff-8ea419af3b96202def061b93440e5b119c49aba8132769f83c9289086559704dR386) it will try to update any packageUri that the ResolvedUrlManager already knows about
    - any resolvedUrls could not be fulfilled [are then fetched by the ResolvedUrlManager](https://github.com/flutter/devtools/pull/3932/files#diff-8ea419af3b96202def061b93440e5b119c49aba8132769f83c9289086559704dR401)
    - [The 2nd pass](https://github.com/flutter/devtools/pull/3932/files#diff-8ea419af3b96202def061b93440e5b119c49aba8132769f83c9289086559704dR406) is then run on any frames that were not updated in the 1st pass.
- [A getter has been added](https://github.com/flutter/devtools/pull/3932/files#diff-8ea419af3b96202def061b93440e5b119c49aba8132769f83c9289086559704dR632) for the package uri with the sourceLine attached
- Some [cpu profile model tests](https://github.com/flutter/devtools/pull/3932/files#diff-06dd95875ced799a3e0e4e96bc84b2980ded2720868a36531a32edfb1d3fd17aR115) have been added to test some edge cases that we want to guarantee work.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/devtools_app/test/cpu_profile_model_test.dart
```diff
@@ -107,6 +123,7 @@ void main() {
           verboseName: 'all',
           category: 'Dart',
           rawUrl: '',
+          processedUrl: '',
```
kenzieschmoll, line 126: not your fault, but while you are modifying this fake data, could you also add the missing 'sourceLine' field?

packages/devtools_app/test/test_data/cpu_profile_test_data.dart
```diff
@@ -729,6 +792,8 @@ final CpuStackFrame stackFrameC2 = CpuStackFrame(
   category: 'Dart',
   rawUrl:
       'file:///path/to/flutter/packages/flutter/lib/src/widgets/binding.dart',
+  processedUrl:
+      'file:///path/to/flutter/packages/flutter/lib/src/widgets/binding.dart',
```
kenzieschmoll, line 796: should this be using `package:flutter/...` format? Same comment applies to other processedUrl fields that begin with `file:///` in this fake test data

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -613,7 +617,7 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
       verboseName: verboseName ?? this.verboseName,
       category: category ?? this.category,
       rawUrl: url ?? rawUrl,
-      processedUrl: url != null ? getSimplePackageUrl(url) : processedUrl,
+      processedUrl: processedUrl ?? this.processedUrl,
       sourceLine: sourceLine,
```
kenzieschmoll, line 621: again, not your fault, but while you're here could you add the missing param for sourceLine and use it here `sourceLine ?? this.sourceLine`

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -665,6 +669,7 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
           CpuProfileData.nameKey: verboseName,
           CpuProfileData.categoryKey: category,
           CpuProfileData.resolvedUrlKey: rawUrl,
+          CpuProfileData.processedUrlKey: processedUrl,
```
kenzieschmoll, line 672: could you add sourceLine here too

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -255,6 +256,10 @@ class VmServiceWrapper implements VmService {
           if (parent != null && parent.frameId != 0)
             CpuProfileData.parentIdKey: '$isolateId-${parent.frameId}',
         };
+
+        if (current.resolvedUrl != null && current.resolvedUrl!.isNotEmpty) {
+          resolvedUrlMap[current.resolvedUrl!] = null;
```
kenzieschmoll, line 261: hoist `current.resolvedUrl` up into a local variable before this if statement. Then use the local var wherever you are currently using `current.resolvedUrl`. This will allow you to get rid of all the null checks following the first `!= null`.

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -285,9 +290,57 @@ class VmServiceWrapper implements VmService {
         },
       });
     }
+
+    await _addProcessedUrlsToTraceObject(
+        isolateId, resolvedUrlMap, traceObject);
```
kenzieschmoll, line 295: nit: add a trailing comma after `traceObject` for better formatting. General rule of thumb is if the method call or signature breaks to a second line, we should use the trailing comma, and we should Always use the trailing comma in flutter build methods (with a couple exceptions like a single line SizedBox widget).

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -285,9 +290,57 @@ class VmServiceWrapper implements VmService {
...
+  // [isolateId] The id which is passed to the getIsolate RPC to load this isolate.
+  // [resolvedUrlMap] A map of where each key is a resolved url that has been extracted from the [traceObject]'s
+  // stack frames. The values of the map will be overriden in this helper.
+  // This is passed as a parameter since [resolvedUrls] can be extracted
+  // while [traceObject] is being created.
+  // [traceObject] A map where the CpuProfileData for each frame is stored.
+  Future<void> _addProcessedUrlsToTraceObject(
+      String isolateId,
+      Map<String, String?> resolvedUrlMap,
+      Map<String, dynamic> traceObject) async {
```
kenzieschmoll, line 312: nit: trailing comma

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -285,9 +290,57 @@ class VmServiceWrapper implements VmService {
...
       });
     }
+
+    await _addProcessedUrlsToTraceObject(
+        isolateId, resolvedUrlMap, traceObject);
+
     return CpuProfileData.parse(traceObject);
   }
 
+  // Helper function for fetching simplified package urls for each of
```
kenzieschmoll, line 300: nit: use /// for dartdoc. https://dart.dev/guides/language/effective-dart/documentation#do-use--doc-comments-to-document-members-and-types

And the first sentence should be followed by a single blank line.
e.g.
```
/// This is my method that does something.
///
/// more details. blah blah blah
```

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -285,9 +290,57 @@ class VmServiceWrapper implements VmService {
...
+  // [resolvedUrlMap] A map of where each key is a resolved url that has been extracted from the [traceObject]'s
+  // stack frames. The values of the map will be overriden in this helper.
+  // This is passed as a parameter since [resolvedUrls] can be extracted
+  // while [traceObject] is being created.
+  // [traceObject] A map where the CpuProfileData for each frame is stored.
+  Future<void> _addProcessedUrlsToTraceObject(
+      String isolateId,
+      Map<String, String?> resolvedUrlMap,
+      Map<String, dynamic> traceObject) async {
+    final processedUrlMapping = <String, String>{};
```
kenzieschmoll, line 313: we should store this mapping as a field on the VmServiceWrapper object. That way we can reuse the processed urls for future recorded profiles. As long as we have the same vm service connection, these urls should persist from request to request.

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -285,9 +290,57 @@ class VmServiceWrapper implements VmService {
...
+    final resolvedUrlList = resolvedUrlMap.keys.toList();
+    final uris =
+        (await _vmService.lookupPackageUris(isolateId, resolvedUrlList)).uris;
```
kenzieschmoll, line 316: piggy backing off of my last comment, we should only request the package uri for resolved urls we do not already have a mapped package uri for. We'll probably want to check for this where `resolvedUrlMap` is built

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -285,9 +290,57 @@ class VmServiceWrapper implements VmService {
...
+  // This is passed as a parameter since [resolvedUrls] can be extracted
+  // while [traceObject] is being created.
+  // [traceObject] A map where the CpuProfileData for each frame is stored.
+  Future<void> _addProcessedUrlsToTraceObject(
+      String isolateId,
+      Map<String, String?> resolvedUrlMap,
+      Map<String, dynamic> traceObject) async {
+    final processedUrlMapping = <String, String>{};
+    final resolvedUrlList = resolvedUrlMap.keys.toList();
+    final uris =
```
kenzieschmoll, line 315: nit: rename to `packageUris` for clarity

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -285,9 +290,57 @@ class VmServiceWrapper implements VmService {
...
+              .values
+              .cast<Map<String, dynamic>>();
+      for (final stackFrameJson in stackFrames) {
+        final resolvedUrl =
+            stackFrameJson[CpuProfileData.resolvedUrlKey] as String?;
+        if (resolvedUrl != null && resolvedUrl.isNotEmpty) {
+          final processedUrl = processedUrlMapping[resolvedUrl];
+          if (processedUrl != null && processedUrl.isNotEmpty) {
+            stackFrameJson[CpuProfileData.processedUrlKey] = processedUrl;
+            continue;
```
kenzieschmoll, line 337: no need for the continue here since the loop will continue by default

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
CoderDake: so I left these as just double because I wanted to make sure that my description seemed sufficient before promoting it to ///. I can definitely do that right off the bat though :)

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
CoderDake: ah right saving this in the wrapper service context allows us to persist those over a much longer time so we can squeek some efficiency out!

packages/devtools_app/test/cpu_profile_model_test.dart
CoderDake: @kenzieschmoll where should I be adding the sourceLine field, I'm not sure I understand

packages/devtools_app/test/test_data/cpu_profile_test_data.dart
CoderDake: yes probably. this was me being lazy learning about the specs initially 😓

packages/devtools_app/test/cpu_profile_model_test.dart
kenzieschmoll: right below processedUrl. This should have been added when I added the sourceLine data but I missed it 🙃

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -357,9 +360,53 @@ class CpuProfileData {
...
+    for (final stackFrameJson in stackFrames) {
+      final resolvedUrl =
+          stackFrameJson[CpuProfileData.resolvedUrlKey] as String?;
+      if (resolvedUrl != null && resolvedUrl.isNotEmpty) {
+        final packageUri =
+            serviceManager.resolvedUriManager.getPackageUri(resolvedUrl);
+        if (packageUri != null) {
+          stackFrameJson[CpuProfileData.processedUrlKey] = packageUri;
+        } else {
+          stackFramesWaitingOnPackageUri.add(stackFrameJson);
```
kenzieschmoll, line 393: instead of storing the stack frame json objects, just store the key instead. And then iterate through the keys below instead of all the stackFrameJson objects. You can iterate through `traceObject[CpuProfileData.stackFramesKey].entries` above to have access to both the key and the value in this for loop

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -56,16 +56,19 @@ class CpuProfileData {
         jsonDecode(jsonEncode(json[stackFramesKey] ?? <String, dynamic>{}));
     for (final MapEntry<String, dynamic> entry in stackFramesJson.entries) {
       final stackFrameJson = entry.value;
-      final stackFrame = CpuStackFrame(
+      final resolvedUrl = stackFrameJson[resolvedUrlKey] ?? '';
+      final processedUrl = stackFrameJson[processedUrlKey] ?? resolvedUrl;
```
kenzieschmoll, line 60: rename to `resolvedPackageUri`. Same comment for the key const: `processedUrlKey` --> `resolvedPackageUriKey`

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -56,16 +56,19 @@ class CpuProfileData {
...
         name: getSimpleStackFrameName(stackFrameJson[nameKey]),
         verboseName: stackFrameJson[nameKey],
         category: stackFrameJson[categoryKey],
         // If the user is on a version of Flutter where resolvedUrl is not
         // included in the response, this will be null. If the frame is a native
         // frame, the this will be the empty string.
-        rawUrl: stackFrameJson[resolvedUrlKey] ?? '',
-        sourceLine: stackFrameJson[sourceLine],
+        rawUrl: resolvedUrl,
+        processedUrl: processedUrl,
```
kenzieschmoll, line 70: same comment for this parameter name

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -357,9 +360,53 @@ class CpuProfileData {
         },
       });
     }
+
+    await _addProcessedUrlsToTraceObject(isolateId, traceObject);
```
kenzieschmoll, line 364: nit: rename to _addPackageUrisToTraceObject

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -44,6 +44,8 @@ class ServiceConnectionManager {
     _serviceExtensionManager = ServiceExtensionManager(isolateManager);
   }
 
+  final ResolvedUriManager resolvedUriManager = ResolvedUriManager();
```
kenzieschmoll, line 47: nit: remove type on LHS and group below with other "managers" (`vmFlagManager`, `timelineStreamManager`, etc.)

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -496,6 +498,36 @@ class ServiceConnectionManager {
   }
 }
 
+class ResolvedUriManager {
+  final _resolvedUrlMap = <String, String?>{};
+  final _unknownUris = <String>[];
```
kenzieschmoll, line 503: instead of tracking unknown uris here, track these elsewhere at the call site of `fetchUnknownUris`. This manager knows that it does not know any uri that is not in `_resolvedUrlMap`, so no need to track the list here.

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -496,6 +498,36 @@ class ServiceConnectionManager {
...
+  VmServiceWrapper? service;
+
+  Future<void> fetchUnknownUris(
+    String isolateId,
```
kenzieschmoll, line 507: take a List of Strings as a param instead of using _unknownUris (which as I said in my previous comment should be tracked where `fetchUnknownUris` is being called.

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -496,6 +498,36 @@ class ServiceConnectionManager {
...
+      for (var i = 0; i < _unknownUris.length; i++) {
+        final unknownUri = _unknownUris[i];
+        final resolvedUri = packageUris[i];
+        _resolvedUrlMap[unknownUri] = resolvedUri!;
+      }
+      _unknownUris.clear();
+    }
+  }
+
+  String? getPackageUri(String uri) {
```
kenzieschmoll, line 522: avoid starting method names with get: https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

rename to `lookupPackageUri`

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -40,6 +40,7 @@ class VmServiceWrapper implements VmService {
 
   final bool trackFutures;
   final Map<String, Future<Success>> _activeStreams = {};
+  final _resolvedUrlMap = <String, String?>{};
```
kenzieschmoll, line 43: remove

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -496,6 +498,36 @@ class ServiceConnectionManager {
   }
 }
 
+class ResolvedUriManager {
```
kenzieschmoll, line 501: let's move this class to its own file in the service directory: resolved_uri_manager.dart just to be consistent with how the other manager type classes are separated

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
CoderDake: Going to handle this in https://github.com/flutter/devtools/issues/3974

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
CoderDake: Going to handle this in https://github.com/flutter/devtools/issues/3974

packages/devtools_app/test/cpu_profile_model_test.dart
CoderDake: Going to handle this in https://github.com/flutter/devtools/issues/3974

packages/devtools_app/lib/src/screens/profiler/cpu_profile_columns.dart
```diff
@@ -105,15 +105,15 @@ class SourceColumn extends ColumnData<CpuStackFrame> {
   SourceColumn() : super.wide('Source', alignment: ColumnAlignment.right);
 
   @override
-  dynamic getValue(CpuStackFrame dataObject) => dataObject.processedUrl;
+  dynamic getValue(CpuStackFrame dataObject) => dataObject.packageUri;
 
   @override
   String getDisplayValue(CpuStackFrame dataObject) {
-    return dataObject.processedUrl;
+    return dataObject.packageUri;
```
kenzieschmoll, line 112: we may need to add the sourceLine number here. I think processedUrl contained this before

CoderDake: @kenzieschmoll 
Would maintaining the sourceLine here: https://github.com/flutter/devtools/pull/3932/files/9c7af2f5ca37c22be0486f582c08f2a06741f0a6#diff-8ea419af3b96202def061b93440e5b119c49aba8132769f83c9289086559704dR586
be ok?

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
CoderDake: Since dart passes primitives by value then I think storing the keys as Strings would use a bunch more memory. I believe that the current approach is just an array of Map References, which would each be the size of the pointer to that map.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_columns.dart
kenzieschmoll: instead of adding it there, why don't we just make `packageUriWithLineNumber` or similar a getter on CpuStackFrame and then the getter can add the source line value if present. It may be a little confusing if another place in the code tries to grab the 'packageUri' and wouldn't be expecting a line number to be on the end of it.

CoderDake: gosh I was working on this earlier and thinking the exact same thing! :O

I will make it happen :D

CoderDake: I've added it as a getter for now to avoid some redundancies

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -704,6 +761,7 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
           CpuProfileData.nameKey: verboseName,
           CpuProfileData.categoryKey: category,
           CpuProfileData.resolvedUrlKey: rawUrl,
+          CpuProfileData.resolvedPackageUriKey: packageUriWithSourceLine,
```
kenzieschmoll, line 764: add the packageUri and the  sourceLine to the json separately.

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
```diff
@@ -0,0 +1,51 @@
+import '../shared/globals.dart';
```
kenzieschmoll, line 1: add copyright for 2022 at the top of this file. (you can copy and paste from any other file and then just change the year). All files need to have the copyright header at the top.

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
```diff
@@ -0,0 +1,51 @@
...
+import 'vm_service_wrapper.dart';
+
+/// Manager for handling package Uri lookup and caching.
+class ResolvedUriManager {
+  Map<String, String?>? _resolvedUrlMap;
+  VmServiceWrapper? service;
```
kenzieschmoll, line 7: initialize and null out the service in vmServiceOpened and vmServiceClosed. then use that service instead of serviceManager.service below

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
```diff
@@ -0,0 +1,51 @@
...
+
+  /// Calls out to the [VmService] to lookup unknown uri to package uri mappings.
+  ///
+  /// Known uri mappings are cached to avoid asking [VmService] for the same
+  /// mapping.
+  ///
+  /// [uris] List of uris to fetch package uris for.
+  Future<void> fetchPackageUris(
+    String isolateId,
+    List<String> uris,
```
kenzieschmoll, line 27: nit: I think this can fit on one line. If I'm wrong, add back the trailing comma

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
```diff
@@ -0,0 +1,51 @@
+import '../shared/globals.dart';
+import 'vm_service_wrapper.dart';
+
+/// Manager for handling package Uri lookup and caching.
+class ResolvedUriManager {
+  Map<String, String?>? _resolvedUrlMap;
```
kenzieschmoll, line 6: add dartdoc explaining that this map maps resolved uris to their resolved package uris.

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
```diff
@@ -0,0 +1,51 @@
...
+  void vmServiceClosed() {
+    _resolvedUrlMap = null;
+  }
+
+  /// Calls out to the [VmService] to lookup unknown uri to package uri mappings.
+  ///
+  /// Known uri mappings are cached to avoid asking [VmService] for the same
+  /// mapping.
+  ///
+  /// [uris] List of uris to fetch package uris for.
```
kenzieschmoll, line 24: great dartdoc!

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
```diff
@@ -0,0 +1,51 @@
...
+  Future<void> fetchPackageUris(
+    String isolateId,
+    List<String> uris,
+  ) async {
+    if (_resolvedUrlMap != null) {
```
kenzieschmoll, line 29: once you initialize the service here, check that the service is != null instead of the map

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
kenzieschmoll, line 6: question for @bkonyi - should we be storing a cache per isolate? right now this would store package uris for all isolates in the same cache

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
```diff
@@ -0,0 +1,51 @@
...
+      final packageUris =
+          (await serviceManager.service!.lookupPackageUris(isolateId, uris))
+              .uris;
+      if (packageUris != null) {
+        for (var i = 0; i < uris.length; i++) {
+          final unknownUri = uris[i];
```
kenzieschmoll, line 35: nit: rename to `resolvedUri`

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -308,6 +311,8 @@ class ServiceConnectionManager {
     }
 
     _connectionAvailableController.add(service);
+
+    resolvedUriManager.vmServiceOpened();
```
kenzieschmoll, line 315: call this where all the other managers' vmServiceOpened methods are called 183-190. Same comment for the vmServiceClosed call below

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
CoderDake: oops ty!

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
bkonyi: It's _probably_ fine (I doubt URI mappings for files will be different between isolates in practice), but it's more correct to have a cache per isolate since that's how things are setup in the VM.

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
```diff
@@ -0,0 +1,51 @@
...
+  VmServiceWrapper? service;
+
+  /// Initializes the [ResolvedUriManager]
+  void vmServiceOpened() {
+    _resolvedUrlMap = <String, String?>{};
+  }
+
+  /// Cleans up the resources of the [ResolvedUriManager]
+  void vmServiceClosed() {
+    _resolvedUrlMap = null;
```
bkonyi, line 16: Style nit: this can be a one-liner (not sure if we have preference one way or another in DevTools, @kenzieschmoll may have an opinion).

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
```diff
@@ -0,0 +1,51 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  /// Returns a package uri for the given uri, if one exists in the cache.
+  ///
+  /// [uri] Absolute path uri to look up in the package uri mapping cache.
+  String? lookupPackageUri(String uri) {
```
bkonyi, line 48: Same style nit as above. Can be a one liner.

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
CoderDake: oops I don't know how this snuck in here. it isn't used so I will delete it all together :D

packages/devtools_app/test/cpu_profile_model_test.dart
```diff
@@ -3,9 +3,13 @@
 // found in the LICENSE file.
 
 // ignore_for_file: avoid_redundant_argument_values
+// ignore_for_file: import_of_legacy_library_into_null_safe
```
bkonyi, line 6: What's not null safe here? I thought everything except the Provider page has been migrated already?

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
CoderDake: ty, I have made it so!

packages/devtools_app/test/cpu_profile_model_test.dart
CoderDake: looks like devtools_test is the culprit

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
CoderDake: I've fixed this up and I'll leave the sourceline fix for the follow up issue:
https://github.com/flutter/devtools/issues/3802

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
```diff
@@ -1,41 +1,47 @@
...
-    _resolvedUrlMap = <String, String?>{};
-  }
+  void vmServiceOpened() =>
+      _isolateResolvedUrlMap = <String, Map<String, String?>>{};
 
   /// Cleans up the resources of the [ResolvedUriManager]
-  void vmServiceClosed() {
-    _resolvedUrlMap = null;
-  }
+  void vmServiceClosed() => _isolateResolvedUrlMap = null;
```
kenzieschmoll, line 20: _usually_ we'll use the `=>` style for one liners when we actually need to return something, but it doesn't really matter. More of a style preference than a hard guideline.

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
```diff
@@ -1,41 +1,47 @@
...
+  Future<void> fetchPackageUris(String isolateId, List<String> uris) async {
+    if (_isolateResolvedUrlMap != null) {
       final packageUris =
           (await serviceManager.service!.lookupPackageUris(isolateId, uris))
               .uris;
+
+      if (!_isolateResolvedUrlMap!.containsKey(isolateId)) {
+        _isolateResolvedUrlMap![isolateId] = <String, String?>{};
+      }
+      final resolvedUrlMap = _isolateResolvedUrlMap![isolateId]!;
```
kenzieschmoll, line 38: use putIfAbsent here instead

`final resolvedUrlMap = _isolateResolvedUrlMap.putIfAbsent(isolateId, () => <String, String?>{});`

packages/devtools_test/lib/src/mocks.dart
```diff
@@ -241,6 +247,15 @@ class FakeServiceManager extends Fake implements ServiceConnectionManager {
 
   @override
   ValueListenable<bool> get deviceBusy => ValueNotifier(false);
+
+  @override
+  Future<void> vmServiceOpened(
+    VmServiceWrapper service, {
+    Future<void> onClosed,
+  }) {
+    resolvedUriManager.vmServiceOpened();
```
kenzieschmoll, line 256: move the call to initFlagManager into this method as well

packages/devtools_test/lib/src/mocks.dart
```diff
@@ -117,6 +120,9 @@ class FakeServiceManager extends Fake implements ServiceConnectionManager {
 
   final MockVM _mockVM = MockVM();
 
+  @override
+  ResolvedUriManager resolvedUriManager = ResolvedUriManager();
```
kenzieschmoll, line 124: make this final and remove type on LHS

packages/devtools_test/lib/src/mocks.dart
```diff
@@ -241,6 +247,15 @@ class FakeServiceManager extends Fake implements ServiceConnectionManager {
...
   @override
   ValueListenable<bool> get deviceBusy => ValueNotifier(false);
+
+  @override
+  Future<void> vmServiceOpened(
+    VmServiceWrapper service, {
+    Future<void> onClosed,
+  }) {
+    resolvedUriManager.vmServiceOpened();
+    return Future.value();
```
kenzieschmoll, line 257: you shouldn't have to return here

packages/devtools_test/lib/src/mocks.dart
```diff
@@ -277,6 +293,7 @@ class FakeVmService extends Fake implements VmServiceWrapper {
             });
 
   CpuSamples cpuSamples;
+  final Map<String, String> _resolvedUriMap;
```
kenzieschmoll, line 296: nit: group with the other final variables

packages/devtools_app/test/resolved_uri_manager_test.dart
```diff
@@ -0,0 +1,155 @@
...
+import 'package:vm_service/vm_service.dart';
+
+void main() async {
+  group('ResolvedUriManager', () {
+    const String isolateId = 'anIsolateId';
+    const uri1 = 'this/is/a/uri1';
+    const uri2 = 'this/is/a/uri2';
+    const packageUri1 = 'uri/am/i1';
+    const packageUri2 = 'uri/am/i2';
+    ResolvedUriManager? resolvedUriManager;
```
kenzieschmoll, line 22: make this late and then you can remove all the null asserts where the variable is used later

packages/devtools_test/lib/src/mocks.dart
CoderDake: linter doesn't seem to like when I remove this.

PR #3946, opened 2022-04-01
Move getCpuProfileTimeline from VmServiceWrapper to CpuProfileModel

CoderDake:
This is a refactor PR to prepare for changes to `getCpuProfileTimeline`(see https://github.com/flutter/devtools/issues/3802). 

This PR moves `getCpuProfileTimeline` from `VmServiceWrapper` to `CpuProfileModel`. This is being done to help VmServiceWrapper be truer to its name, so it can mostly just be a wrapper. This also has the added benefit of making the `getCpuProfileTimeline` easier to unit test, so a basic unit test has been added to get the ball rolling in this area.

I have also renamed CpuProfileModel's `CpuSample` to `ProfileDataCpuSample` as it was causing confusing naming conflicts.

https://github.com/flutter/devtools/issues/3802

## Pre-launch Checklist

- [x] I signed the [CLA].

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

### DIFF
--- a/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
+++ b/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
@@ -275,6 +275,14 @@ class CpuProfileData {
 
   factory CpuProfileData.empty() => CpuProfileData.parse({});
 
+  /// Generates [CpuProfileData] by fetching [CpuSamples] from the [VmService]
+  /// that is stored in the global [ServiceConnectionManager].
+  ///
+  /// Only samples collected in the time range [origin, origin + extent] will be reported.
+  ///
+  /// [isolateId] The id which is passed to the getIsolate RPC to load this isolate.
+  /// [origin] origin time
+  /// [extent] extent time
   static Future<CpuProfileData> generateFromCpuSamples(
     String isolateId,
     int origin,
--- a/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
+++ b/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
@@ -6,12 +6,14 @@ import 'dart:collection';
 import 'dart:convert';
 
 import 'package:flutter/foundation.dart';
+import 'package:vm_service/vm_service.dart';
 
 import '../../charts/flame_chart.dart';
 import '../../primitives/trace_event.dart';
 import '../../primitives/trees.dart';
 import '../../primitives/url_utils.dart';
 import '../../primitives/utils.dart';
+import '../../shared/globals.dart';
 import '../../ui/search.dart';
 import 'cpu_profile_transformer.dart';
 
@@ -74,9 +76,9 @@ class CpuProfileData {
     final stackTraceEvents =
         (json[traceEventsKey] ?? []).cast<Map<String, dynamic>>();
     final samples = stackTraceEvents
-        .map((trace) => CpuSample.parse(trace))
+        .map((trace) => ProfileDataCpuSample.parse(trace))
         .toList()
-        .cast<CpuSample>();
+        .cast<ProfileDataCpuSample>();
 
     return CpuProfileData._(
       stackFrames: stackFrames,
@@ -193,15 +195,15 @@ class CpuProfileData {
     CpuProfileData originalData,
     bool Function(CpuStackFrame) includeFilter,
   ) {
-    final filteredCpuSamples = <CpuSample>[];
+    final filteredCpuSamples = <ProfileDataCpuSample>[];
     void includeSampleOrWalkUp(
-      CpuSample sample,
+      ProfileDataCpuSample sample,
       Map<String, Object> sampleJson,
       CpuStackFrame stackFrame,
     ) {
       if (includeFilter(stackFrame)) {
         filteredCpuSamples.add(
-          CpuSample(
+          ProfileDataCpuSample(
             leafId: stackFrame.id,
             userTag: sample.userTag,
             traceJson: sampleJson,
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -75,9 +77,9 @@ class CpuProfileData {
     final stackTraceEvents =
         (json[traceEventsKey] ?? []).cast<Map<String, dynamic>>();
     final samples = stackTraceEvents
-        .map((trace) => CpuSample.parse(trace))
+        .map((trace) => ProfileDataCpuSample.parse(trace))
```
kenzieschmoll, line 80: a couple other things you can do if you don't want to rename just to avoid a class name conflict
1) add a prefix for the import: `import 'package:vm_service/vm_service.dart' as vm_service;`. Then everywhere you need to use the CpuSample class from package:vm_service, you would write `vm_service.CpuSample`
2) hide the conflicting class when you import the library (this won't work in this case because you actually need to use the class with the conflicting name, but just adding it for your future information). `import 'package:vm_service/vm_service.dart' hide CpuSample;`. Then the classes would not conflict. You can also use the keyword `show` if you want to import only a single or a few select classes from a library.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -276,6 +278,99 @@ class CpuProfileData {
...
+  static Future<CpuProfileData> generateFromCpuSamples(
+    String isolateId,
+    int origin,
+    int extent,
+  ) async {
+    // As of service protocol version 3.27 _getCpuProfileTimeline does not exist
+    // and has been replaced by getCpuSamples. We need to do some processing to
+    // get back to the format we expect.
+    final cpuSamples =
+        await serviceManager.service!.getCpuSamples(isolateId, origin, extent);
```
kenzieschmoll, line 298: push this call into `getCpuProfile` in `cpu_profile_service.dart`. Then make this method accept just two params `String isolateId` and `CpuSamples cpuSamples`.

packages/devtools_app/test/cpu_profile_model_test.dart
```diff
@@ -2,11 +2,13 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-// ignore_for_file: avoid_redundant_argument_values
+// ignore_for_file: avoid_redundant_argument_values, import_of_legacy_library_into_null_safe
 
-import 'package:devtools_app/src/primitives/utils.dart';
-import 'package:devtools_app/src/screens/profiler/cpu_profile_model.dart';
+import 'package:devtools_app/devtools_app.dart';
```
kenzieschmoll, line 7: do not import this file. instead import the exact file you need. Your IDE may auto suggest you import this file (this file exports a bunch of other files in devtools), but you'll always want to make sure you are using the exact library you need instead of this catch-all lib.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
CoderDake: I can definitely put it back the way it was. This was more for readability as it was really confusing to be jumping around the code and then realize that I was looking at 2 different classes.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -276,6 +277,91 @@ class CpuProfileData {
...
 
+  /// Generates [CpuProfileData] from the provided [CpuSamples].
+  ///
+  /// Only samples collected in the time range [origin, origin + extent] will be reported.
+  ///
+  /// [isolateId] The isolate id which was used to get the [cpuSamples].
+  /// This will be used to tag the stack frames and trace events.
+  /// [cpuSamples] The CPU samples that will be used to generate the [CpuProfileData]
+  static Future<CpuProfileData> generateFromCpuSamples(
+    String isolateId,
```
bkonyi, line 288: Nit: make these `required` named parameters.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -276,6 +277,91 @@ class CpuProfileData {
...
   factory CpuProfileData.empty() => CpuProfileData.parse({});
 
+  /// Generates [CpuProfileData] from the provided [CpuSamples].
+  ///
+  /// Only samples collected in the time range [origin, origin + extent] will be reported.
+  ///
+  /// [isolateId] The isolate id which was used to get the [cpuSamples].
+  /// This will be used to tag the stack frames and trace events.
+  /// [cpuSamples] The CPU samples that will be used to generate the [CpuProfileData]
+  static Future<CpuProfileData> generateFromCpuSamples(
```
bkonyi, line 287: AFAICT, there's no asynchronous work being done in this method so this can be refactored into a factory. Maybe something like:

```
CpuProfileData.fromCpuSamples({required String isolateId, required vm_service.CpuSamples cpuSamples}) {...}
```

packages/devtools_app/test/performance_model_test.dart
```diff
@@ -90,7 +91,12 @@ void main() {
               'vsyncOverhead': 1000
             },
           ],
-          PerformanceData.cpuProfileKey: goldenCpuProfileDataJson,
+          PerformanceData.cpuProfileKey:
+              (await CpuProfileData.generateFromCpuSamples(
```
bkonyi, line 95: Did `dart format` do this?

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
CoderDake: oh true! this couldn't be a factory before because it had the async call to the vm_service but it has since been moved out. Great call!

packages/devtools_app/test/performance_model_test.dart
CoderDake: yeah,  I think so. I have vscode auto format every time i save my files 😅

packages/devtools_app/lib/src/screens/profiler/cpu_profile_service.dart
```diff
@@ -13,11 +13,10 @@ extension CpuProfilerExtension on VmServiceWrapper {
...
     required int extentMicros,
   }) async {
-    return await serviceManager.service!.getCpuProfileTimeline(
-      serviceManager.isolateManager.selectedIsolate.value!.id!,
-      startMicros,
-      extentMicros,
-    );
+    final isolateId = serviceManager.isolateManager.selectedIsolate.value!.id!;
+    final cpuSamples = await serviceManager.service!
+        .getCpuSamples(isolateId, startMicros, extentMicros);
```
kenzieschmoll, line 18: nit: trailing comma afer 'extendMicros'. I'm surprised the linter didn't catch this - have you merged master into your PR recently?

packages/devtools_app/test/performance_model_test.dart
```diff
@@ -43,7 +44,7 @@ void main() {
       expect(performanceData.cpuProfileData, isNull);
     });
 
-    test('to json', () {
+    test('to json', () async {
```
kenzieschmoll, line 47: this async should be able to be removed

packages/devtools_app/test/performance_model_test.dart
```diff
@@ -90,7 +91,10 @@ void main() {
               'vsyncOverhead': 1000
             },
           ],
-          PerformanceData.cpuProfileKey: goldenCpuProfileDataJson,
+          PerformanceData.cpuProfileKey: CpuProfileData.generateFromCpuSamples(
+            isolateId: goldenSamplesIsolate,
+            cpuSamples: CpuSamples.parse(goldenCpuSamplesJson)!,
+          ).toJson,
```
kenzieschmoll, line 97: isn't this equivalent to `goldenCpuProfileDataJson`? why not just continue using `goldenCpuProfileDataJson` directly?

packages/devtools_app/test/performance_model_test.dart
CoderDake: without the await this looks much cleaner now :D

packages/devtools_test/lib/src/mocks.dart
```diff
@@ -260,20 +261,24 @@ class FakeVmService extends Fake implements VmServiceWrapper {
...
-        cpuProfileData = cpuProfileData ??
-            CpuProfileData.parse(<String, dynamic>{
-              'type': '_CpuProfileTimeline',
+        cpuSamples = cpuSamples ??
+            CpuSamples.parse({
               'samplePeriod': 50,
+              'maxStackDepth': 12,
               'sampleCount': 0,
-              'stackDepth': 128,
+              'timeSpan': -1,
```
kenzieschmoll, line 272: why does this need to be -1?

packages/devtools_app/test/test_data/cpu_profile_test_data.dart
```diff
@@ -216,7 +542,8 @@ final Map<String, dynamic> goldenCpuProfileStackFrames =
               'path/to/flutter/packages/flutter/lib/src/rendering/stack.dart',
         },
         '140357727781376-17': {
-          'category': '[Stub] OneArgCheckInlineCache',
+          'category':
+              'Dart', // TODO: Double check this is getting done properly
```
kenzieschmoll, line 546: remove todo

packages/devtools_app/lib/src/screens/profiler/cpu_profile_service.dart
CoderDake: yeah just a couple hours ago. I'll give it another whirl though :D

packages/devtools_app/test/performance_model_test.dart
CoderDake: good catch. I forgot to switch this back, when I was fixing things up :)

packages/devtools_test/lib/src/mocks.dart
CoderDake: since it is deprecated and always returns -1 I set it to that. In hindsight I could probably leave it null though!

packages/devtools_app/lib/src/screens/profiler/cpu_profile_service.dart
CoderDake: It looks like this rule is only enforced when the declaration overflows a line
https://dart-lang.github.io/linter/lints/require_trailing_commas.html

packages/devtools_test/lib/src/mocks.dart
kenzieschmoll: leaving -1 sgtm then. Just add a quick comment explaining that

kenzieschmoll (approved):
one more comment about adding a comment explaining '-1', then lgtm

packages/devtools_test/lib/src/mocks.dart
CoderDake: I've removed it from the constructor since it is redundant, and only makes things more confusing :)

PR #3953, opened 2022-04-01
Only send the dwds event when the service is registered

kenzieschmoll:
This will prevent us from attempting to send this dwds event when the dwds service is not registered (non web apps and some special cases internally).

FYI @annagrin


elliette (approved):
Nice!

PR #3959, opened 2022-04-05
Add enhance tracing hints to the frame analysis view

kenzieschmoll:
Work towards https://github.com/flutter/devtools/issues/3485. This feature is still off by default, and this PR is work towards enabling an MVP.

Next steps: 
- rework this view so that Timeline Events and Frame Analysis are in two tabs
- show useful information for the selected phase (perhaps a profile-like view of timeline events for that phase)
- provide hints about expensive operations (in progress docs here: https://github.com/flutter/website/pull/6977)

Screen Shot 2022-04-05 at 10 05 23 AM

@InMatrix


packages/devtools_app/lib/src/screens/performance/performance_model.dart
```diff
@@ -1238,77 +1243,112 @@ class FrameAnalysis {
...
+  late FramePhase buildPhase = _generateBuildPhase();
+
+  FramePhase _generateBuildPhase() {
+    final uiEvent = frame.timelineEventData.uiEvent;
+    if (uiEvent == null) {
+      return FramePhase.build(events: <SyncTimelineEvent>[]);
+    }
+    final buildEvents = uiEvent
+        .childrenWithCondition(
+          (event) => event.name!.caseInsensitiveEquals(buildEventName),
```
elliette, line 1275: Can `name` be null?

packages/devtools_app/lib/src/screens/performance/performance_model.dart
```diff
@@ -1238,77 +1243,112 @@ class FrameAnalysis {
...
-    if (_paintEvent != null) return _paintEvent;
+  /// Data for the Paint phase of [frame].
+  ///
+  /// This is drawn from the "Paint" timeline event on the UI thread
+  late FramePhase paintPhase = _generatePaintPhase();
 
-    return _paintEvent =
-        frame.timelineEventData.uiEvent!.firstChildWithCondition(
+  FramePhase _generatePaintPhase() {
+    if (frame.timelineEventData.uiEvent == null) {
```
elliette, line 1313: Maybe pull out into a variable so the null assert isn't need

kenzieschmoll: done

packages/devtools_app/lib/src/screens/performance/performance_model.dart
kenzieschmoll: it shouldn't be for any child of uiEvent, but I added null checking to be safe.

elliette (approved):

PR #3960, opened 2022-04-05
Recover from missing trace events

kenzieschmoll:
Another case where we need to recover gracefully when we get bad data. See https://github.com/flutter/flutter/issues/101379 for details. This may be important enough to do a cherry pick release for, as it affects Track Widget Builds.

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/timeline_event_processor.dart
+++ b/packages/devtools_app/lib/src/screens/performance/timeline_event_processor.dart
@@ -342,9 +342,6 @@ class TimelineEventProcessor {
         // Animator::BeginFrame - DurationEnd ([event] - duplicate)
         // VSYNC - DurationEnd
         //
-        if (debugTimeline) {
-          debugFrameTracking
-              .writeln('Duplicate duration end event: $eventJson');
         debugTraceEventCallback(
           () => log(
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/screens/performance/timeline_event_processor.dart
```diff
@@ -366,40 +342,98 @@ class TimelineEventProcessor {
...
-        current = current.parent as SyncTimelineEvent?;
-        currentDurationEventNodes[eventThreadId] = current;
+          debugTraceEventCallback(() {
+            log(
+              'Missing DurationEnd event - adding fake end event $fakeEndEvent',
+            );
+          });
+
+          current.addEndEvent(fakeEndEvent);
+          current = current.parent as SyncTimelineEvent?;
```
bkonyi, line 431: Is this cast necessary?

bkonyi (approved):

packages/devtools_app/lib/src/screens/performance/timeline_event_processor.dart
kenzieschmoll: yeah. without it, the following analysis error is present: "A value of type 'TimelineEvent?' can't be assigned to a variable of type 'SyncTimelineEvent?'"

PR #3971, opened 2022-04-11
Migrate remaining in dev_tools lib to null safety.

polina-c:


packages/devtools_app/lib/src/app.dart
```diff
@@ -215,7 +213,7 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
                 .where((p) => embed && page != null ? p.screenId == page : true)
                 .where((p) => !hide.contains(p.screenId))
                 .toList();
-            if (tabs.isEmpty) return child;
+            if (tabs.isEmpty) return child!;
```
CoderDake, line 216: I don't know enough about ValueListenableBuilder to suggest what needs to be returned here. It looks like child may end up being null and causing an issue here though so we may need to find a way to return a blank widget or something agreeable to the builder.

packages/devtools_app/lib/src/app.dart
```diff
@@ -215,7 +213,7 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
                 .where((p) => embed && page != null ? p.screenId == page : true)
                 .where((p) => !hide.contains(p.screenId))
                 .toList();
-            if (tabs.isEmpty) return child;
+            if (tabs.isEmpty) return child ?? const SizedBox.shrink();
```
kenzieschmoll, line 216: we know child is not null because we specify it on line 236, so you can null assert child here

packages/devtools_app/lib/src/app.dart
```diff
@@ -376,24 +380,24 @@ class DevToolsScreen<C> {
   /// Defaults to false.
   final bool supportsOffline;
 
-  Provider<C> get controllerProvider {
+  Provider<C?> get controllerProvider {
```
kenzieschmoll, line 383: Looks like this can be non-nullable.

packages/devtools_app/lib/src/app.dart
```diff
@@ -313,15 +311,21 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
...
-          ),
-        ),
-      ),
+      builder: (context, child) {
+        Notifications? notifications;
+        if (child != null) {
+          notifications = Notifications(
+            child: ReleaseNotesViewer(
+              releaseNotesController: releaseNotesController,
+              child: child,
```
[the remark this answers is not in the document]
polina-c: done

packages/devtools_app/lib/src/app.dart
polina-c: done

kenzieschmoll (approved):

polina-c:
Got blocked: https://stackoverflow.com/questions/71847613/nullability-mismatch-in-simple-assignment-after-switching-to-sound-null-safety

polina-c:
unblocked and tested

PR #3983, opened 2022-04-15
Fix null safety errors across DevTools

kenzieschmoll:
Several DevTools workflows were broken due to errors with type casting and null safety.

### DIFF
--- a/packages/devtools_app/lib/src/primitives/enum_utils.dart
+++ b/packages/devtools_app/lib/src/primitives/enum_utils.dart
@@ -27,7 +27,8 @@ class EnumUtils<T> {
   final Map<String, T> _lookupTable = {};
   final Map<T, String> _reverseLookupTable = {};
 
-  T? enumEntry(String enumName) => _lookupTable[enumName];
+  T? enumEntry(String? enumName) =>
+      enumName != null ? _lookupTable[enumName] : null;
 
   String? name(T enumEntry) => _reverseLookupTable[enumEntry];
 }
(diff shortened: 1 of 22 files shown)


packages/devtools_app/lib/src/screens/debugger/variables.dart
```diff
@@ -58,13 +57,13 @@ class ExpandableVariable extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    if (variable == null) return const SizedBox();
+    final _variable = variable;
```
polina-c, line 60: Starting local variable with '_' looks confusing for me as normally '_' means visibility that does not apply to local variables. Can you name it variable_  or variableLocal instead?

polina-c (approved):

packages/devtools_app/lib/src/primitives/enum_utils.dart
```diff
@@ -27,7 +27,8 @@ class EnumUtils<T> {
   final Map<String, T> _lookupTable = {};
   final Map<T, String> _reverseLookupTable = {};
 
-  T? enumEntry(String enumName) => _lookupTable[enumName];
```
polina-c, line 30: All errors in this PR are runtime errors, that were not caught by tests. Would it be right to cover the fixes with tests? May be in a separate CL, to speed up the fix landing.

kenzieschmoll: added some tests and filed https://github.com/flutter/devtools/issues/3991 to track the meta issue of improving devtools testing

packages/devtools_app/lib/src/screens/debugger/variables.dart
kenzieschmoll: renamed.

Issue #3985, opened 2022-04-15
Clicking isolate selector shows menu item in the wrong place 


elliette:
Clicking on the isolate selector in the footer opens the menu item at the top left of DevTools: 

![Screen Shot 2022-04-15 at 2 31 04 PM](https://user-images.githubusercontent.com/21270878/163634808-4f683def-2b44-40a7-9a96-8ed63aafa4d9.png)

I see the following assertion failing:
```
Assertion failed:
file:///Users/elliottbrooks/dev/flutter/packages/flutter/lib/src/material/dropdown.dart:363:
16
menuLimits.top + menuLimits.height <= size.height
is not true
```

This is reproducible from master but not on DevTools `v.2.12.2`

jacob314:
That is ugly.

PR #3989, opened 2022-04-15
Isolate selector in footer bar is a `PopupMenu` instead of a `DropdownMenu`

elliette:
Fixes https://github.com/flutter/devtools/issues/3985

Also cleans up some unnecessary formatting in the footer (removes individual alignment of items, which was unnecessary because they end up being aligned correctly in their row with `spaceBetween`).

### DIFF
--- a/packages/devtools_app/lib/src/shared/status_line.dart
+++ b/packages/devtools_app/lib/src/shared/status_line.dart
@@ -220,7 +220,6 @@ class IsolateSelector extends StatelessWidget {
             },
           ).toList(),
         );
-        // );
       },
     );
   }
--- a/packages/devtools_app/lib/src/shared/status_line.dart
+++ b/packages/devtools_app/lib/src/shared/status_line.dart
@@ -15,6 +15,7 @@ import '../analytics/constants.dart' as analytics_constants;
 import '../info/info_controller.dart';
 import '../service/isolate_manager.dart';
 import '../service/service_manager.dart';
+import '../ui/utils.dart';
 import 'common_widgets.dart';
 import 'device_dialog.dart';
 import 'globals.dart';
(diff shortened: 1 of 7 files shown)


kenzieschmoll (commented):
have you checked these changes with a narrow screen width? I suspect that removing the Flexible and Expanded widgets may lead to renderflex overflow errors

packages/devtools_app/lib/src/shared/status_line.dart
```diff
@@ -35,78 +35,44 @@ class StatusLine extends StatelessWidget {
...
-    // Have an area for page specific help (always docked to the left).
-    children.add(
-      Expanded(
-        child: Align(
-          alignment: Alignment.centerLeft,
-          child: buildHelpUrlStatus(context, currentScreen, textTheme),
-        ),
-      ),
-    );
+        final children = <Widget>[];
```
kenzieschmoll, line 43: let's construct this list directly instead of calling children.add a bunch of times:

```
final children = [
  buildHelpUrlStatus(context, currentScreen, textTheme),
  const BulletSpacer(),
  if (showIsolateSelector)
    ...[
          const IsolateSelector(), 
          const BulletSpacer(),
       ],
   ...
];
```

We can also just build this directly as the children property in row. It actually makes the build method a little more readable.
```
Row(
  children: [ ... ],
)
```

elliette: Good point! Only change to your suggestion is I've pulled getting the children into a separate method

elliette:
> have you checked these changes with a narrow screen width? I suspect that removing the Flexible and Expanded widgets may lead to renderflex overflow errors

Good point! Changed so we only show the items that can fit.

Before (without this PR):
Screen Shot 2022-04-18 at 9 48 54 AM

Now:

Screen Shot 2022-04-18 at 11 58 14 AM

packages/devtools_app/lib/src/shared/status_line.dart
```diff
@@ -205,12 +158,13 @@ class StatusLine extends StatelessWidget {
                         Icons.info_outline,
                         size: actionsIconSize,
                       ),
-                      const SizedBox(width: denseSpacing),
-                      Text(
-                        description,
-                        style: textTheme.bodyText2,
-                        overflow: TextOverflow.clip,
-                      ),
+                      if (!isNarrow) const SizedBox(width: denseSpacing),
```
kenzieschmoll, line 161: write `if (!isNarrow) ...[ ... ],` instead so you can reuse the if statement

packages/devtools_app/lib/src/ui/utils.dart
```diff
@@ -221,3 +221,50 @@ class ThemedColor {
...
+  late MediaSize _height;
+  late MediaSize _width;
+
+  MediaSize _calculateWidth(BuildContext context) {
+    final width = MediaQuery.of(context).size.width;
+    if (width < 300) return MediaSize.xxs;
+    if (width >= 300 && width < 600) return MediaSize.xs;
+    if (width >= 600 && width < 900) return MediaSize.s;
+    if (width >= 900 && width < 1200) return MediaSize.m;
+    if (width >= 1200 && width < 1500)
```
kenzieschmoll, line 253: nit: we don't technically need to first check in all of these if statements. These are implied by the previous return. same comment below.

packages/devtools_app/lib/src/ui/utils.dart
```diff
@@ -221,3 +221,50 @@ class ThemedColor {
...
+enum MediaSize {
+  xxs,
+  xs,
+  s,
+  m,
+  l,
+  xl,
+}
+
+class ScreenSize {
```
kenzieschmoll, line 236: can you add tests for this class? you should be able to use the testWidgetsWithWindowSize helper to set screen sizes for the tests

packages/devtools_app/test/ui_utils_test.dart
```diff
@@ -214,3 +276,27 @@ void main() {
...
+  WidgetTester tester, {
+  required MediaSize width,
+  required MediaSize height,
+}) async {
+  await tester.pumpWidget(wrap(Container()));
+  final BuildContext context = tester.element(find.byType(Container));
+  expect(ScreenSize(context).width, equals(width));
+  expect(ScreenSize(context).height, equals(height));
+
+  // await tester.pumpWidget(
```
kenzieschmoll, line 290: should this be removed?

packages/devtools_app/lib/src/shared/status_line.dart
```diff
@@ -176,10 +186,18 @@ class StatusLine extends StatelessWidget {
           return child!;
         }
       },
-      child: Text(
-        'No client connection',
-        style: textTheme.bodyText2,
-      ),
+      child: isExtraNarrow
+          ? DevToolsTooltip(
+              message: 'No client connection',
```
kenzieschmoll, line 191: put 'No client connection' into a const since it is used twice

kenzieschmoll (approved):
one comment then lgtm

packages/devtools_app/lib/src/ui/utils.dart
elliette: Done, thanks!

packages/devtools_app/test/ui_utils_test.dart
elliette: Yes, thanks!

packages/devtools_app/lib/src/ui/utils.dart
elliette: Done!

packages/devtools_app/lib/src/shared/status_line.dart
elliette: Done, thanks!

PR #4006, opened 2022-04-20
Migrate tests to null safety.

polina-c:


packages/devtools_app/test/association_variable_test.dart
```diff
@@ -22,10 +19,11 @@ final libraryRef = LibraryRef(
 );
 
 void main() {
-  ServiceConnectionManager manager;
+  ServiceConnectionManager? manager;
```
kenzieschmoll, line 22: make this late and non-nullable and we can get rid of all the null asserts on `manager` later

packages/devtools_app/test/inspector_screen_test.dart
```diff
@@ -25,7 +23,7 @@ void main() {
   const screen = InspectorScreen();
 
   FakeServiceManager fakeServiceManager;
-  FakeServiceExtensionManager fakeExtensionManager;
+  FakeServiceExtensionManager? fakeExtensionManager;
```
kenzieschmoll, line 26: It is confusing to me why `FakeServiceManager fakeServiceManager;` isn't giving an analysis error here. If it isn't, do you know why that is?

Regardless, let's make both of these late

packages/devtools_app/test/memory_screen_test.dart
```diff
@@ -27,18 +26,17 @@ import 'test_data//memory_test_allocation_data.dart';
 import 'test_data/memory_test_data.dart';
 
 void main() {
-  MemoryScreen screen;
-  MemoryController controller;
+  late MemoryScreen screen;
+  late MemoryController controller;
   FakeServiceManager fakeServiceManager;
```
kenzieschmoll, line 31: make late

packages/devtools_app/test/network_screen_test.dart
```diff
@@ -14,16 +12,16 @@ import 'package:flutter_test/flutter_test.dart';
 import 'package:mockito/mockito.dart';
 
 void main() {
-  NetworkScreen screen;
+  late NetworkScreen screen;
   FakeServiceManager fakeServiceManager;
```
kenzieschmoll, line 16: make late

packages/devtools_app/test/scaffold_test.dart
```diff
@@ -21,18 +19,18 @@ import 'package:mockito/mockito.dart';
 
 void main() {
   group('DevToolsScaffold widget', () {
-    MockServiceManager mockServiceManager;
+    MockServiceManager? mockServiceManager;
```
kenzieschmoll, line 22: make late

packages/devtools_app/test/typed_data_variable_test.dart
```diff
@@ -69,7 +70,8 @@ void main() {
       ),
       isolateRef,
     );
-    when(manager.service.getObject(any, any, offset: 0, count: 4))
+
+    when(manager.service!.getObject('1', '123', offset: 0, count: 4))
```
kenzieschmoll, line 74: can you comment on whether '1' and '123' match some test data somewhere or if they are just arbitrary?

packages/devtools_app/test/association_variable_test.dart
```diff
@@ -88,7 +86,7 @@ void main() {
       ),
       isolateRef,
     );
-    when(manager.service.getObject(any, any, offset: 0, count: 2))
+    when(manager!.service!.getObject('433', '123', offset: 0, count: 2))
```
kenzieschmoll, line 89: can you comment on whether '433' and '123' match some test data somewhere or if they are just arbitrary?

packages/devtools_app/test/memory_screen_test.dart
polina-c: done

packages/devtools_app/test/network_screen_test.dart
polina-c: done

packages/devtools_app/test/typed_data_variable_test.dart
polina-c: I replaced them with constants to make them more trackable. 
I am not sure it makes sense to add comments, as there are many occurrences. Mockito documentation makes it clear that values should match the data. If it is not clear enough, it would be better to pull request the documentation, rather then comment it in code:

https://github.com/dart-lang/mockito/blob/master/NULL_SAFETY_README.md#mocking-members-with-non-nullable-type-variable-return-types

packages/devtools_app/test/association_variable_test.dart
polina-c: see comment above

packages/devtools_app/test/typed_data_variable_test.dart
kenzieschmoll: replacing with consts is good. Sometimes it is just helpful to have a simple comment "These values are arbitrary" where the consts are defined to help other people reading the code to know that they are not "special" values. In other words, that these values can be changed without breaking things.

kenzieschmoll (approved):

Issue #3966, opened 2022-04-08
Add a setting to enable `debugProfileBuildsEnabledUserWidgets` in DevTools


kenzieschmoll:
Right now, enabling "Track Widget Builds" in DevTools enables the [debugProfileBuildsEnabled](https://github.com/flutter/flutter/blob/35b18ba2e902dac1e16f557fa385069f1f5e1069/packages/flutter/lib/src/widgets/debug.dart#L119)  flag in the Flutter framework. There is another flag,  [debugProfileBuildsEnabledUserWidgets](https://github.com/flutter/flutter/blob/35b18ba2e902dac1e16f557fa385069f1f5e1069/packages/flutter/lib/src/widgets/debug.dart#L133), that only tracks user-created widget builds instead of every widget build.

We should either:
A) change our setting to enable [debugProfileBuildsEnabledUserWidgets](https://github.com/flutter/flutter/blob/35b18ba2e902dac1e16f557fa385069f1f5e1069/packages/flutter/lib/src/widgets/debug.dart#L133) instead of [debugProfileBuildsEnabled](https://github.com/flutter/flutter/blob/35b18ba2e902dac1e16f557fa385069f1f5e1069/packages/flutter/lib/src/widgets/debug.dart#L119), or
B) add an additional setting "Only track user-created widgets" that enables [debugProfileBuildsEnabledUserWidgets](https://github.com/flutter/flutter/blob/35b18ba2e902dac1e16f557fa385069f1f5e1069/packages/flutter/lib/src/widgets/debug.dart#L133). 

I think B is a better solution because there may be instances where a user wants to track the build of all widgets to look deeper.

PR #4010, opened 2022-04-21
Add setting for tracking user-created widget builds

kenzieschmoll:
twb-with-radio-button

CC @InMatrix @ayanogawa
CC @gaaclarke 

Fixes https://github.com/flutter/devtools/issues/3966


packages/devtools_app/lib/src/screens/performance/enhance_tracing.dart
```diff
@@ -0,0 +1,379 @@
...
+import '../../primitives/auto_dispose_mixin.dart';
+import '../../service/service_extension_manager.dart';
+import '../../service/service_extensions.dart' as extensions;
+import '../../shared/common_widgets.dart';
+import '../../shared/globals.dart';
+import '../../shared/theme.dart';
+import '../../ui/service_extension_widgets.dart';
+import 'performance_screen.dart';
+
+class EnhanceTracingButton extends StatelessWidget {
```
kenzieschmoll, line 18: this class was moved over from performance_screen.dart with a couple changes (marked below)

packages/devtools_app/lib/src/screens/performance/enhance_tracing.dart
```diff
@@ -0,0 +1,379 @@
...
+    final textStyle = theme.subtleTextStyle;
+    return ServiceExtensionCheckboxGroupButton(
+      title: title,
+      icon: icon,
+      tooltip: 'Add more detail to the Timeline trace',
+      minScreenWidthForTextBeforeScaling:
+          SecondaryPerformanceControls.minScreenWidthForTextBeforeScaling,
+      extensions: [
+        extensions.profileWidgetBuilds,
+        extensions.profileUserWidgetBuilds,
```
kenzieschmoll, line 37: this line was added

packages/devtools_app/lib/src/screens/performance/enhance_tracing.dart
```diff
@@ -0,0 +1,379 @@
...
+        extensions.profileRenderObjectLayouts,
+        extensions.profileRenderObjectPaints,
+      ],
+      customExtensionUi: {
+        extensions.profileWidgetBuilds.extension:
+            const TrackWidgetBuildsSetting(),
+        extensions.profileUserWidgetBuilds.extension: const SizedBox(),
+      },
```
kenzieschmoll, line 45: these lines were added

polina-c (approved):

CoderDake (commented):
I tried my best to look over this, but I think I'm still a bit too green to fully understand and approve this.
That being said I learned a lot reading it, and only found a couple of small things to comment on.

packages/devtools_app/lib/src/ui/service_extension_widgets.dart
```diff
@@ -760,12 +777,20 @@ class _ServiceExtensionCheckboxGroupOverlay extends StatelessWidget {
...
       ),
     );
   }
+
+  Widget _extensionSetting(ToggleableServiceExtensionDescription extension) {
+    assert(extensions.contains(extension));
+    final customUi = customExtensionUi[extension.extension];
+    return customUi != null
+        ? customUi
+        : ServiceExtensionCheckbox(serviceExtension: extension);
```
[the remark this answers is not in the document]
kenzieschmoll: done

packages/devtools_app/lib/src/screens/performance/enhance_tracing.dart
```diff
@@ -0,0 +1,379 @@
...
+  static const _scopeSelectorPadding = 32.0;
+
+  /// Service extensions for tracking widget builds.
+  ///
+  /// There is an extension for each value of [TrackWidgetBuildsScope], and the
+  /// extensions in this list are indexed according to their enum value index.
+  final _trackWidgetBuildsExtensions = [
+    extensions.profileWidgetBuilds,
+    extensions.profileUserWidgetBuilds,
+  ];
```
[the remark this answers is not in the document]
kenzieschmoll: Using a list can save space since with a map, we are keeping a set of keys in memory that we do not actually need. But for this case, since it is small, I have replaced with a map for improved readability.

InMatrix:
How expensive is it to enable TWB for the user code? I wonder if DevTools should just enable it as soon as the user opens the Performance page. Maybe this depends on project size?

gaaclarke:
@InMatrix I did a lot of work to try to make it faster to the point were we could enable it by default.  The problem is that anytime I made it faster, I also made the baseline faster.  The cost varies project to project but in our benchmarks we were seeing a 17% increase in build time with it on: https://github.com/flutter/flutter/pull/100926#issuecomment-1089442133

That makes it not a good candidate to always be on, but maybe you are right that it might make make sense if people are actively looking at performance.

PR #4015, opened 2022-04-22
Split debugger_screen_test.dart.

polina-c:
This PR is preparation to remove --no-sound-null-safety from bots.

Null safe mocks do not allow to invoke 'when' in test setup. 

See https://stackoverflow.com/questions/69810221/bad-state-cannot-call-when-within-a-stub-response-type-null-is-not-a-su/71971770#71971770

So, mocking should happen directly in main of the tests.
That's why it is more convenient to have tests that require different mocking, in different files.
Most our tests are already structured this way.

Splitting debugger_screen_test to meet this condition.

### DIFF
--- a/packages/devtools_app/test/debugger_codeview_test.dart
+++ b/packages/devtools_app/test/debugger_codeview_test.dart
@@ -1,0 +1,97 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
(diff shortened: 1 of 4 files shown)


kenzieschmoll:
> So, mocking should happen directly in main of the tests.
That's why it is more convenient to have tests that require different mocking, in different files.

I'm not going to bike shed too hard over tests, but this is exactly what testing `group`s can be used for. Each group can have its own `setUp` steps.

kenzieschmoll (approved):

Issue #3608, opened 2022-01-31
Toggle Widget Select Mode tooltip is hard to read in dark theme


DanTup:
In VS Code, this tooltip is black text on a dark background:

![Screenshot 2022-01-31 at 18 51 30](https://user-images.githubusercontent.com/1078012/151854639-dc3f84f0-6f58-4f6c-a090-7f2e1f36120b.png)

PR #4017, opened 2022-04-22
Fix tooltip colouring for "select widget mode"

CoderDake:
# Demo
[Demo of old vs new behaviour](https://threadit.app/thread/pbv0h2vmsdgfhzkgmhu8/message/r4mizeenfozhyn1cpyzaw42s?utm_medium=referral-link)

# Description
Fixes https://github.com/flutter/devtools/issues/3608 by fixing the select widget mode  tooltip. The tooltip colour text is now white during dark mode and black during light mode.

Another place I found that uses the `ServiceExtensionTooltip` is the "Show structured error" toggle. It seems to behave well with these new changes as well.
image

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -642,6 +642,7 @@ class DevToolsTooltip extends StatelessWidget {
     this.padding = const EdgeInsets.all(defaultSpacing),
     this.decoration,
     this.textStyle,
+    this.tooltipTextColor,
```
kenzieschmoll, line 645: instead of adding an extra parameter for the tooltipTextColor, this can be passed in as part of the text style.

packages/devtools_app/lib/src/ui/service_extension_widgets.dart
```diff
@@ -889,6 +889,9 @@ class ServiceExtensionTooltip extends StatelessWidget {
         borderRadius: BorderRadius.circular(defaultBorderRadius),
       ),
       textStyle: DefaultTextStyle.of(context).style,
```
kenzieschmoll, line 891: use the same text color as the other tooltips on the Inspector page: https://github.com/flutter/devtools/blob/374f37db70ab18acf1055807c4f59d14e85d24ce/packages/devtools_app/lib/src/ui/service_extension_widgets.dart#L919. 

You can pass in a TextStyle just like the text widget I linked passes in a TextStyle with this color:
`TextStyle(color: Theme.of(context).colorScheme.toggleButtonsTitle)`

packages/devtools_app/lib/src/shared/common_widgets.dart
CoderDake: unfortunately the text style gets overriden so I would need to refactor things a bit inside the Dev tool tip to make that work.

I originally wanted to change the richtext == null case to copy the style and override, but I wasn't sure if that was too heavy handed

kenzieschmoll: it's a bug that we are overriding textStyle when `richMessage == null`. Change that if statement to
`if (style == null && richMessage == null)` and this should fix the problem

kenzieschmoll (approved):

PR #4019, opened 2022-04-22
Fix cpu_profiler_test and others.

polina-c:


packages/devtools_test/lib/src/mocks/mocks.dart
```diff
@@ -121,7 +121,7 @@ class MockCpuSamples extends Mock implements CpuSamples {}
 // that will override the public getters on the class (e.g. isFlutterAppNow,
 // isProfileBuildNow, etc.). Do this after devtools_app is migrated to null
 // safety so that we can use null-safety here.
-class MockConnectedApp extends Mock implements ConnectedApp {}
+class MockConnectedAppLegacy extends Mock implements ConnectedApp {}
```
kenzieschmoll, line 124: can you add a comment describing why this is now legacy and also what the plan is moving forward to make this "non-legacy"

kenzieschmoll (approved):
one comment then lgtm

packages/devtools_test/lib/src/mocks/mocks.dart
polina-c: done

PR #4033, opened 2022-04-24
Split debugger_screen_test.

polina-c:
See https://github.com/flutter/devtools/pull/4015#issue-1211725157


packages/devtools_app/test/debugger_screen_variales_test.dart
```diff
@@ -0,0 +1,419 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
```
CoderDake, line 1: typo in the file name
rename the file from:
`test/debugger_screen_variales_test.dart`
to
`test/debugger_screen_variables_test.dart`

CoderDake (approved):
As far as I can tell the tests haven't been modified when splitting them into new files.
I tried using 
`git diff --color-moved=plain test/debugger_screen_test.dart [destination file]`
(I also tried using the zebra mode)
and it didn't seem to catch any major changes of the code after the move.

I wasn't able to tell if any tests were forgotten during the move, but as long as Polina is confident that everything was moved, then this should be good to go.

packages/devtools_app/test/debugger_screen_variales_test.dart
polina-c: done

PR #4034, opened 2022-04-24
Migrate some tests from unsound to sound null safety.

polina-c:


packages/devtools_app/test/resolved_uri_manager_test.dart
```diff
@@ -11,143 +11,142 @@ import 'package:mockito/mockito.dart';
...
+  });
+  when(service.onIsolateEvent).thenAnswer((_) {
+    return const Stream.empty();
+  });
+  when(service.onStdoutEvent).thenAnswer((_) {
+    return const Stream.empty();
+  });
+  when(service.onStderrEvent).thenAnswer((_) {
+    return const Stream.empty();
+  });
```
kenzieschmoll, line 33: These setup steps have come up in multiple tests. Would it be possible for the MockVmServiceWrapper class to do this when it is initialized?

kenzieschmoll (approved):
LGTM with one comment

packages/devtools_app/test/resolved_uri_manager_test.dart
polina-c: done

PR #4035, opened 2022-04-25
Split scaffold_test.

polina-c:
See https://github.com/flutter/devtools/pull/4015#issue-1211725157


packages/devtools_app/test/scaffold_debugging_controls_test.dart
```diff
@@ -0,0 +1,94 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
```
elliette, line 1: Even with splitting, I think we should change to 2022 (here and elsewhere)

elliette (approved):
One request but LGTM!

Issue #3719, opened 2022-02-24
"File a bug" missing in embedded Dart DevTools single-page view 


elliette:
When a panel in Dart DevTools is embedded in an IDE etc with the query parameters `embed=true` and `page=[[pageName]]`, the buttons in the top header are removed, including the "File a bug" button. 

If a single page is embedded, we should add the "File a bug" button somewhere else (maybe a footer?) to make sure users know where to file a bug. We should also verify there aren't any other features in the header that we want to add to the embedded view.

PR #4041, opened 2022-04-25
Include the footer when DevTools is embedded 

elliette:
Fixes https://github.com/flutter/devtools/issues/3719

Currently we hide both the header and footer when DevTools is embedded. This means that some of the actions only found in the header/footer are missing (eg, report a bug, select the isolate, etc). 

This change shows the DevTools footer in the embedded view. When embedded, the footer will include the "About DevTools" and "Report feedback" actions (which in non-embedded views are in the header). 

Embedded inspector view:

Screen Shot 2022-04-25 at 1 41 33 PM


elliette:
@DanTup @helin24 -will this cause any problems with embedding for the IDEs? Thanks!

DanTup:
It won't cause any issues for VS Code, although running the current code it seems to be chopped off at the bottom and I can't get to it (scrolling scrolls the tree if my mouse is over it, and if I move it over the footer there's no scrolling.. if I resize the window it just remains off-screen by the same amount):

Screenshot 2022-04-26 at 14 32 33

Apr-26-2022 14-34-11

elliette:
> running the current code it seems to be chopped off at the bottom and I can't get to it

Oh no! @DanTup - are there instructions for running my local DevTools from VSCode so I can try to reproduce? Thanks!

DanTup:
> @DanTup - are there instructions for running my local DevTools from VSCode so I can try to reproduce? Thanks!

There's not a _great_ way to do it right now.. I just modified the Dart-Code source to use a URL from a locally run server. Building a local SDK with the changes would work, but is probably a bit more time consuming.

It's probably useful to be able to swap this for testing though, so I'll add a new setting in the VS Code extension for it and make a build tomorrow you could use. (In the meantime, it's _possible_ if you put it inside an iframe that isn't the full height you might see the same behaviour, but I haven't tried this).

packages/devtools_app/lib/src/shared/about_dialog.dart
```diff
@@ -0,0 +1,92 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
```
elliette, line 1: This was moved over from app.dart with no changes

packages/devtools_app/lib/src/shared/report_feedback_button.dart
```diff
@@ -0,0 +1,43 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
```
elliette, line 1: This was moved over from app.dart with no changes

elliette:
> (In the meantime, it's _possible_ if you put it inside an iframe that isn't the full height you might see the same behaviour, but I haven't tried this).

I was able to reproduce the issue that way :) It **should** be fixed now, here is it in the IFRAME:

resize

DanTup (approved):

packages/devtools_app/lib/src/shared/status_line.dart
```diff
@@ -28,17 +30,33 @@ double get statusLineHeight => scaleByFontFactor(24.0);
...
+  EdgeInsets get padding => const EdgeInsets.fromLTRB(
+        defaultSpacing,
+        defaultSpacing,
+        defaultSpacing,
+        denseSpacing,
+      );
 
   @override
   Widget build(BuildContext context) {
+    final height = scaleByFontFactor(24.0) + padding.top + padding.bottom;
```
kenzieschmoll, line 48: lets use a named const instead of the raw number value

kenzieschmoll (approved):

packages/devtools_app/lib/src/shared/status_line.dart
elliette: Good point, done!

DanTup:
> @DanTup - are there instructions for running my local DevTools from VSCode so I can try to reproduce? Thanks!

FYI in the next Dart-Code release you can set the `"dart.customDevToolsUri"` setting like this:

```js
"dart.customDevToolsUri": "http://127.0.0.1:55431"
```

This will be used in place of the local spawned server when launching DevTools (both embedded and external). I had planned to let you provide the path to the DevTools entry script (like `tool/build_e2e.dart`) but it was a little clunky because it also requires a specific working directory and a `LOCAL_DART_SDK` env variable set so I decided to just keep it simple initially (this means while you can test changes work correctly in the embedded view, you cannot use this to test changes to the DevTools server - you'll need to build the SDK and ensure VS Code is using that SDK for that).

kenzieschmoll:
> you cannot use this to test changes to the DevTools server

Couldn't you just run `tool/build_e2e.dart` and use the uri that that outputs for the `"dart.customDevToolsUri"` setting? `build_e2e.dart` builds a local devtools app and launches it with a local server, so this URI is for a running instance of DevTools that contains both local app and server.

DanTup:
> Couldn't you just run `tool/build_e2e.dart` and use the uri that that outputs for the `"dart.customDevToolsUri"` setting?

That was my original plan, a `dart.customDevToolsPath` setting. However running `build_e2e` failed because it required a specific `cwd`. I added a bit of a hack that detected when it ended `tool/build_e2e.dart` and set the `cwd` to the path with that part chopped off. But then it failed because `LOCAL_DART_SDK` wasn't set. I was going to set that automatically based on the SDK selected in VS Code, but realised that might not always be right (for example I tend not to run VS Code with my checked out SDK because I don't compile it often).

Since it felt like it was getting a bit messy, I thought this may be simpler (at least for the requirement of just testing embedded). If it seems like there's some value in supporting running the server this way I can look again though (perhaps the setting can be an object with three keys, allowing you to set a script, `cwd` and `LOCAL_DART_SDK` all in one place?).

DanTup:
I made a start on the above, but found that `build_e2e` doesn't currently pass arguments like `--machine` through (and the `serve_local.dart` script in DDS also does not handle them) so it's not currently a drop-in replacement for the shipped `dart devtools` command.

I've filed https://github.com/Dart-Code/Dart-Code/issues/4080 about this and will see if I can tweak those scripts to work the same (I think it'd be better to keep the behaviour the same across the two so the testing is more accurate to what gets shipped than have Dart-Code have two paths for all of this).

kenzieschmoll:
> will see if I can tweak those scripts to work the same

Feel free to update those scripts (`build_e2e` and `serve_local.dart`) to handle the flags that we need (flags that the `dart devtools` command already handles).

PR #4046, opened 2022-04-26
Add raster metrics visualizer to performance page

kenzieschmoll:
raster-metrics-small-gif

We should still iterate on this UI with UXR (@InMatrix @ayanogawa). The goal is to get something landed here that we can gather initial feedback from and iterate upon. This feature is hidden behind a flag, so will not be enabled by default.

@iskakaushik @jacob314


packages/devtools_app/lib/src/screens/performance/raster_metrics.dart
```diff
@@ -0,0 +1,355 @@
...
+        ),
+      ),
+      actions: [
+        DialogCloseButton(),
+      ],
+    );
+  }
+}
+
+class RasterMetricsController {
```
elliette, line 236: This is just a curiosity - I know in other places we give the controller its own file. Is there a reason to include it here and not separate it out?

elliette (approved):

packages/devtools_app/lib/src/screens/performance/raster_metrics.dart
kenzieschmoll: I'll move it to a separate file in a follow up PR.

zanderso:
+1 great work!

InMatrix:
How to turn on the flag? I look forward to trying this out.

PR #4048, opened 2022-04-26
Test cleanup.

polina-c:
- [x] Delete some legacy mocks
 - [x] Delete test that exercises only mocks


packages/devtools_app/test/vm_service_private_test.dart
```diff
@@ -1,43 +0,0 @@
...
-    );
-    when(service.callMethod(argThat(equals('_collectAllGarbage')))).thenAnswer(
-      (_) => Future.value(Success()),
-    );
-
-    final fakeServiceManager = FakeServiceManager(service: service);
-    setGlobal(ServiceConnectionManager, fakeServiceManager);
-    VmServicePrivate.enablePrivateRpcs = false;
-
-    try {
```
CoderDake, line 29: Although this spec is a bit odd, I believe it still serves to tests that this exception is thrown
https://github.com/flutter/devtools/blob/374f37db70ab18acf1055807c4f59d14e85d24ce/packages/devtools_app/lib/src/service/vm_service_wrapper.dart#L909

So I think it is still providing value, although it may benefit us to rewrite it to be a bit cleaner in the future

polina-c: This test does not touch VmServiceWrapper. It works with MockVmServiceWrapper instead.

CoderDake: I chatted with @polina-c about this and she has mentioned that the test doesn't really work as depicted and isn't providing value any more. Her expert opinion is good enough for me :D

CoderDake (approved):

Issue #3925, opened 2022-03-25
Show release notes in the embedded DevTools


kenzieschmoll:
We need to make sure the release notes viewer works well for small screens.

CoderDake:
After reading through https://github.com/flutter/devtools/issues/3918 my understanding is that I will just make sure that the release notes pop up width is better handled in smaller screens
https://github.com/flutter/devtools/issues/3918

PR #4053, opened 2022-04-29
Release notes in small screens

CoderDake:
Fixes https://github.com/flutter/devtools/issues/3925

This PR Fixes the release notes popup so that it will display well in smaller screens. Demos have been provided to show how the popup behaves before and after

## Demos
[Release Notes Before Fix](https://threadit.app/thread/dj28yffbce3ozxjybsz8)
[Release Notes After Fix](https://threadit.app/thread/plng41ayc0xzqgpcie40)

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -826,6 +828,9 @@ class AreaPaneHeader extends StatelessWidget implements PreferredSizeWidget {
   }
 
   List<Widget> _buildActions() {
+    if (onlyShowRightActions) {
```
CoderDake, line 831: This configuration for _buildActions was added to make sure that there is maximum space left for the `title` in small screens. (otherwise there was just a big empty gap)

packages/devtools_app/test/common_widgets_test.dart
```diff
@@ -211,4 +211,87 @@ void main() {
       expect(findCheckboxValue(), isTrue);
     });
   });
+
+  group('AreaPaneHeader', () {
```
CoderDake, line 215: I would be interested if anyone has any recommendations for these specs. I feel like the way I've grouped them makes sense, but I wonder if there is a better way to do it.

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -843,6 +848,7 @@ class AreaPaneHeader extends StatelessWidget implements PreferredSizeWidget {
             scrollDirection: Axis.horizontal,
             children: scrollableCenterActions,
           ),
+          key: const Key('scrollableCenterActionsContainer'),
```
kenzieschmoll, line 851: it we are only adding this key for testing, we should remove it and find another way to check the center actions' existence in tests

packages/devtools_app/lib/src/shared/release_notes/release_notes.dart
```diff
@@ -59,9 +60,8 @@ class _ReleaseNotesViewerState extends State<ReleaseNotesViewer>
     visibilityController = longAnimationController(this);
     // Add [densePadding] to the end to account for the space between the
     // release notes viewer and the right edge of DevTools.
-    visibilityAnimation =
-        Tween<double>(begin: 0, end: viewerWidth + densePadding)
-            .animate(visibilityController);
+    visibilityAnimation = Tween<double>(begin: animationStartValue, end: 0)
```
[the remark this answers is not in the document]
CoderDake: yes it does 😄

packages/devtools_app/lib/src/shared/release_notes/release_notes.dart
```diff
@@ -38,7 +38,8 @@ class ReleaseNotesViewer extends StatefulWidget {
 
 class _ReleaseNotesViewerState extends State<ReleaseNotesViewer>
     with AutoDisposeMixin, SingleTickerProviderStateMixin {
-  static const viewerWidth = 600.0;
+  static const maxViewerWidth = 600.0;
+  static const animationStartValue = 1000.0;
```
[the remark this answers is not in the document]
CoderDake: honestly it is arbitrary. There is no specific reason for it being 1000, other than it seeming like a nice number.
I will experiment to see if I can just set the animation to go from 0.0 to 1.0 though. as this would simplify some math

packages/devtools_app/lib/src/shared/common_widgets.dart
CoderDake: since i want to test the lack of existence I think I need to keep a key so that I have something to test against.
From my experience in UI testing with html-ish stuff adding identifiers/keys was pretty standard practice. Do we have a different way of doing this in flutter?

packages/devtools_app/lib/src/shared/release_notes/release_notes.dart
```diff
@@ -10,6 +10,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter_markdown/flutter_markdown.dart';
 import 'package:http/http.dart' as http;
 
+import '../../../devtools.dart' as devtools;
```
kenzieschmoll, line 13: import the files you need directly instead of importing devtools.dart, which just exports a bunch of files

packages/devtools_app/test/common_widgets_test.dart
```diff
@@ -214,84 +214,79 @@ void main() {
...
-        await tester.pumpWidget(
-          wrap(
-            const AreaPaneHeader(
-              title: Text(titleText),
-              leftActions: [leftAction],
-              scrollableCenterActions: [centerAction],
-              onlyShowRightActions: true,
-            ),
+    testWidgets(
+        'does NOT show center actions if no left or center actions present',
```
kenzieschmoll, line 228: is this accurate? We always want to show center actions if center actions are present, even if there are no left actions

packages/devtools_app/lib/src/shared/release_notes/release_notes.dart
```diff
@@ -91,10 +89,26 @@ class _ReleaseNotesViewerState extends State<ReleaseNotesViewer>
...
+          LayoutBuilder(
+            builder: (context, constraints) {
+              final widthForSmallScreen =
+                  constraints.maxWidth - 2 * densePadding;
+              final width = min(
+                _ReleaseNotesViewerState.maxViewerWidth,
+                widthForSmallScreen,
+              );
+
+              return Stack(
```
kenzieschmoll, line 101: I don't think this needs to be in a stack

packages/devtools_app/lib/src/shared/release_notes/release_notes.dart
```diff
@@ -137,7 +155,12 @@ class ReleaseNotes extends AnimatedWidget {
         child: Column(
           children: [
             AreaPaneHeader(
-              title: const Text('What\'s new in DevTools?'),
+              title: const Expanded(
```
kenzieschmoll, line 158: do we need this to be in an `Expanded`? The actions that are added to the same row as `title` in `AreaPaneHeader` will be placed in `Expanded` widgets, and generally you only want one `Expanded` per `Row` or `Column`. Also, it is best practice to place `Expanded` and `Flexible` widgets directly in the `Row` or `Column` they'll be used in instead of defined somewhere outside of that context. This makes it clearer to understand how a flex layout is constructed.

packages/devtools_app/test/common_widgets_test.dart
```diff
@@ -211,4 +211,82 @@ void main() {
       expect(findCheckboxValue(), isTrue);
     });
   });
+
+  group('AreaPaneHeader', () {
```
kenzieschmoll, line 215: add one test that has left, right, and center actions all defined

packages/devtools_app/lib/src/shared/release_notes/release_notes.dart
CoderDake: In this case this file just contains the version of devtools and isn't the devtools package. So I think this one is different from importing the whole package

packages/devtools_app/lib/src/shared/release_notes/release_notes.dart
CoderDake: @kenzieschmoll I couldn't get this to work without a stack. It would bug out and just show a weird hazy overlay.
Do you know if I would have to do anything special to get this to work without a Stack?

packages/devtools_app/lib/src/shared/common_widgets.dart
CoderDake: ah thank you for sharing that video about keys. I had just assumed it was something akin to css ids/classes but see that it goes much deeper now.
I will approach this in the left right centre fashion in order to clean this up and make things less sketchy for testing too :D

packages/devtools_app/test/common_widgets_test.dart
CoderDake: in this case both center and left actions  are absent, so this case was checking to make sure the center actions did not  show. I'll clean up the wording and the tests when i add the left right and center  :)

packages/devtools_app/lib/src/shared/release_notes/release_notes.dart
CoderDake: I will look into this to verify :)

CoderDake: So in this case the ellipses won't show on the title unless that expansion is there. I will take a look to see if there is a  way that I can Rectify this issue in a way that satisfies the best practices you have listed

packages/devtools_app/lib/src/shared/release_notes/release_notes.dart
kenzieschmoll: oh whoops you're right. I was thinking it was `devtools_app.dart`.

kenzieschmoll (commented):
In this PR, let's also remove the `!isEmbedded()` check here: https://github.com/flutter/devtools/blob/374f37db70ab18acf1055807c4f59d14e85d24ce/packages/devtools_app/lib/src/shared/release_notes/release_notes.dart#L186.

Release notes were only disabled for embedded views because the viewer didn't look good for small screens. FYI @helin24 @DanTup with the next release of DevTools, release notes should start showing up in the IDEs

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -761,6 +761,18 @@ class ToolbarAction extends StatelessWidget {
   }
 }
 
+class AreaPaneHeaderActions extends StatelessWidget {
+  const AreaPaneHeaderActions(this.actions);
+
+  final List<Widget> actions;
+  @override
```
kenzieschmoll, line 768: nit: new line between class members

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -812,43 +820,21 @@ class AreaPaneHeader extends StatelessWidget implements PreferredSizeWidget {
...
+            Expanded(
+              child: DefaultTextStyle(
+                maxLines: maxLines,
+                overflow: TextOverflow.ellipsis,
+                style: theme.textTheme.subtitle2!,
+                child: title,
+              ),
             ),
-            ..._buildActions(),
+            AreaPaneHeaderActions(actions),
```
kenzieschmoll, line 831: Instead of creating another widget for this, just add the actions directly to this Row `...actions`. That gets rid of the nested row that is added by wrapping `actions` with `AreaPaneHeaderActions`

packages/devtools_app/test/shared/common_widgets_test.dart
```diff
@@ -211,4 +211,53 @@ void main() {
       expect(findCheckboxValue(), isTrue);
     });
   });
+
+  group('AreaPaneHeader', () {
+    const titleText = 'The title';
+    const centerActionText = 'The Center Action';
+    const centerActionContainerKey = Key('scrollableCenterActionsContainer');
+    const centerAction = Text(centerActionText);
```
kenzieschmoll, line 219: remove unused consts

packages/devtools_app/test/shared/common_widgets_test.dart
```diff
@@ -211,4 +211,53 @@ void main() {
...
+
+    setUp(() {
+      setGlobal(IdeTheme, IdeTheme());
+    });
```
kenzieschmoll, line 223: instead of doing this here, call `setGlobal(IdeTheme, IdeTheme());` in the top level main method. You can get rid of the other call at the top of the file too when you do this. You can also move `setGlobal(ServiceConnectionManager, FakeServiceManager());` to be called in top level main.

packages/devtools_app/test/shared/common_widgets_test.dart
```diff
@@ -211,4 +211,53 @@ void main() {
...
+      await tester.pumpWidget(
+        wrap(
+          const AreaPaneHeader(
+            title: Text(titleText),
+          ),
+        ),
+      );
+
+      final Row row = tester.widget(find.byType(Row)) as Row;
+      expect(
```
kenzieschmoll, line 236: also check for the presence of titleText

packages/devtools_app/test/shared/common_widgets_test.dart
```diff
@@ -211,4 +211,53 @@ void main() {
...
+        row.children.length,
+        equals(2),
+      );
+      expect(find.text(actionText), findsOneWidget);
```
kenzieschmoll, line 260: check for presence of title text here too

packages/devtools_app/test/shared/common_widgets_test.dart
CoderDake: hehe I beat you to this one 😄 
just pushed it up

kenzieschmoll (approved):

DanTup:
I can't see the video above, but if you need me to test in VS Code, let me know how to trigger it and I can have a look.

PR #4057, opened 2022-05-02
Refactor tests into subdirectories

kenzieschmoll:
no functional changes - just file moves


packages/devtools_app/test/app_size_tests/app_size_screen_test.dart
```diff
@@ -21,10 +21,10 @@ import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:mockito/mockito.dart';
 
-import 'test_data/app_size_test_data/new_v8.dart';
-import 'test_data/app_size_test_data/old_v8.dart';
-import 'test_data/app_size_test_data/sizes.dart';
-import 'test_data/app_size_test_data/unsupported_file.dart';
+import '../test_data/app_size_test_data/new_v8.dart';
```
polina-c, line 24: Do we have to postfix folder names with '_test' and '_test_data'?
It seems the type is clear from the context.

I am asking because shorter paths would be more convenient.

kenzieschmoll: renamed to avoid redundancy.

polina-c: Test folders are still postfixed: https://screenshot.googleplex.com/9XLKYmpXXgdqgQQ.png. It is redundancy, because the files are postfixed themselves.

kenzieschmoll: the intent was to differentiate top level folders in `test/` that actually contain tests from top-level folders in `test/` that do not (e.g. `fixtures`, `goldens`, `matchers`, etc.). But perhaps there is a cleaner way to do this without having redundant directory names. Removed "_test" from subdirectories.

polina-c (approved):

PR #4062, opened 2022-05-03
Add auto complete suggestions in eval text field

jankuss:
This PR adds the currently hovered auto complete item as a suggestion at the end of the eval text field.

## Before
image

## After

https://user-images.githubusercontent.com/1659532/166514287-e3123d72-c810-429d-9dd7-147bf1cda145.mov

This should improve the overall UX of the eval text field.

This fixes bullet point number 2 in #3095 

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -297,20 +297,24 @@ class AutoCompleteState extends State<AutoComplete> with AutoDisposeMixin {
...
     final areaHeight = offset.dy;
+
     final maxAreaForPopup = areaHeight - tileEntryHeight;
     // TODO(terry): Scrolling doesn't work so max popup height is also total
     //              matches to use.
-    topMatchesLimit = min(
-      defaultTopMatchesLimit,
-      (maxAreaForPopup / tileEntryHeight) - 1, // zero based.
-    ).truncate();
+    topMatchesLimit = tileEntryHeight > 0
```
elliette, line 305: What is this check for (when would the tileEntryHeight be 0?)

packages/devtools_app/lib/src/screens/debugger/evaluate.dart
```diff
@@ -176,6 +237,11 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
                   ),
                 );
               },
+              // Disable ligatures, so the suggestions of the auto complete work correcly.
```
elliette, line 240: What happens with ligatures enabled?

elliette:
This looks great, thank you! 🎉  🎉 

Would it be possible to update so that if the cursor is at the end of the text field and a user hits the right arrow key, the suggestion is entered?

kenzieschmoll:
> Would it be possible to update so that if the cursor is at the end of the text field and a user hits the right arrow key, the suggestion is entered?

Implementing the same for when a user hits 'tab' would also be great

packages/devtools_app/lib/src/screens/debugger/evaluate.dart
jankuss: This is what happens:

https://user-images.githubusercontent.com/1659532/166835194-ca3bef40-74be-4143-a39c-9d8c019ca49b.mov

T‌he ligature in this example is the `fi` of in the word `profile`. I'm not quite sure why this behaviour happens though, maybe do you have an idea?

T‌o build the content within the `TextField`, we use this code:
image

We have two `TextSpan`, the first displays the text `prof` (which is the actual typed text). The other `TextSpan` will display `ile`, which is the suggestion. The only reasonable cause I can come up with is that something goes wrong because the ligature is split between two different `TextSpan`.

I tried placing a "Zero Width Non-Joiner" (https://unicode-explorer.com/c/200C) character to prevent this, but this will break the caret position.

That's why the easiest solution to this is to disable ligatures for the text in the TextField.

packages/devtools_app/lib/src/ui/search.dart
jankuss: The `tileEntryHeight` will be zero, when there are no items in the auto complete popup. Interestingly, the logic im guarding with that check works fine in the actual application, it will just fail in tests 🤷 

Thinking about it, since we don't display any auto complete items in the popup when there are no matches, we could just `return` early and skip the unnecessary calculation which is causing issues.

Basically a early return along the lines of:
```
if (tileContents.isEmpty) return const SizedBox();
```

jankuss:
> This looks great, thank you! 🎉 🎉
> 
> Would it be possible to update so that if the cursor is at the end of the text field and a user hits the right arrow key, the suggestion is entered?

Will add it 👍 

> > Would it be possible to update so that if the cursor is at the end of the text field and a user hits the right arrow key, the suggestion is entered?
> 
> Implementing the same for when a user hits 'tab' would also be great

This was already possible before the PR

packages/devtools_app/lib/src/screens/debugger/evaluate.dart
elliette: Oh weird! Thanks for the explanation. Fine to keep as is with ligatures disabled

packages/devtools_app/lib/src/ui/search.dart
```diff
@@ -285,6 +285,10 @@ class AutoCompleteState extends State<AutoComplete> with AutoDisposeMixin {
         )
         .toList();
 
+    // When there are no tiles present, we don't need to display the
+    // auto correct list.
+    if (tileContents.isEmpty) return const SizedBox.shrink();
```
elliette, line 290: This seems like a good solution, thanks!

elliette (approved):
Thank you for adding this!

packages/devtools_app/test/eval_field_test.dart
```diff
@@ -0,0 +1,251 @@
+import 'package:devtools_app/devtools_app.dart';
```
kenzieschmoll, line 1: this file needs a copy right. You can copy from another test file and paste here, just make sure the year is 2022. Also, please move this test to the `test/debugger/` directory

packages/devtools_app/lib/src/screens/debugger/evaluate.dart
```diff
@@ -65,12 +74,63 @@ class _ExpressionEvalFieldState extends State<ExpressionEvalField>
...
+    final editingParts = _getCurrentEditingParts();
+
+    if (!_isRealVariableNameOrField(editingParts)) {
+      return;
+    }
+
+    _autoCompleteController.updateCurrentSuggestion(_activeWord);
+  }
+
+  EditingParts _getCurrentEditingParts() {
```
kenzieschmoll, line 103: nit: avoid starting method names with "get". Let's replace with `_currentEditingParts`

packages/devtools_app/test/eval_field_test.dart
jankuss: There is no `test/debugger/` directory, or do you mean that the test file should be prefixed with `debugger_`?

jankuss: I see - Added the copyright and moved the file

kenzieschmoll (approved):

PR #4088, opened 2022-05-11
Use AutomaticKeepAlive on the FlameChart widget

kenzieschmoll:
This wrapper can be used to prevent rebuilds between tab switches:
keep-alives

By wrapping a widget in `KeepAliveWrapper`, we can ensure its state is maintained when switching from one widget to another. This PR significantly improves the performance of switching tabs on the Performance page and on the Cpu Profiler page. We should consider using this elsewhere in DevTools when we have a tab view switching between views that are expensive to build / render.

### DIFF
--- a/packages/devtools_app/lib/src/charts/flame_chart.dart
+++ b/packages/devtools_app/lib/src/charts/flame_chart.dart
@@ -101,11 +101,7 @@ abstract class FlameChart<T, V> extends StatefulWidget {
 // like implementation).
 abstract class FlameChartState<T extends FlameChart,
         V extends FlameChartDataMixin<V>> extends State<T>
-    with
-        AutoDisposeMixin,
-        AutomaticKeepAliveClientMixin,
-        FlameChartColorMixin,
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -275,8 +279,13 @@ abstract class FlameChartState<T extends FlameChart,
     super.dispose();
   }
 
+  @override
+  bool wantKeepAlive = true;
```
polina-c, line 283: Do you want to add comment that explains why we want to keep it alive?

polina-c (approved):
One comment and lgtm

packages/devtools_app/lib/src/charts/flame_chart.dart
kenzieschmoll: added `KeepAliveWrapper` so that this can be used for any view, and added dart doc to that class.

PR #4089, opened 2022-05-12
Upgrade package:flutter_riverpod dependency to eliminate null-safety warnings

kenzieschmoll:


bkonyi:
LGTM, but it looks like you'll need to update the SDK lower bound to include the beta releases or update the  Dart version used by the CI to 2.17.

Issue #3948, opened 2022-04-01
Link to #devtools discord channel from DevTools help menu


kenzieschmoll:
channel: https://discord.com/channels/608014603317936148/958862085297672282

PR #4102, opened 2022-05-13
Add discord channel link to DevTools about dialog

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/3948

### DIFF
--- a/packages/devtools_app/lib/src/shared/about_dialog.dart
+++ b/packages/devtools_app/lib/src/shared/about_dialog.dart
@@ -16,9 +16,6 @@ import 'scaffold.dart';
 import 'theme.dart';
 
 class DevToolsAboutDialog extends StatelessWidget {
-  static const _discordChannelLink =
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/shared/about_dialog.dart
```diff
@@ -32,9 +35,30 @@ class DevToolsAboutDialog extends StatelessWidget {
             children: [
               const Text('Encountered an issue? Let us know at '),
               _createFeedbackLink(context),
-              const Text('.')
+              const Text(','),
             ],
           ),
+          Wrap(
```
polina-c, line 41: Do we want to have a method _createDiscordLink, in consistency with _createFeedbackLink?

polina-c (approved):

packages/devtools_app/lib/src/shared/about_dialog.dart
kenzieschmoll: refactored both to be stateless widgets.

PR #4104, opened 2022-05-15
Fix casting errors on memory screen.

polina-c:
Errors:
```
flutter: [ERROR]: type '_InternalLinkedHashMap<String, dynamic>' is not a subtype of type 'Map<String, Object>' in type cast
flutter: [ERROR]: LateInitializationError: Field 'processRss' has not been initialized.
```


jacob314 (approved):
lgtm

Issue #4105, opened 2022-05-16
Automatically include DevTools version in bug reports


kenzieschmoll:
When a user clicks the feedback button or link in devtools, the DevTools version should be automatically populated in the github issue or b/ bug.

PR #4112, opened 2022-05-17
Add devtools version and device info to github issue description

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/4105

Now, when a user clicks the report feedback button, the github issue will be pre-filled with helpful data:
Screen Shot 2022-05-17 at 11 37 11 AM


packages/devtools_app/test/shared/device_dialog_test.dart
```diff
@@ -88,6 +121,89 @@ void main() {
...
+      );
+
+      deviceDialog = DeviceDialog(
+        connectedApp: app,
+      );
+
+      await tester.pumpWidget(wrap(deviceDialog));
+      expect(find.text('Device Info'), findsOneWidget);
+
+      expect(find.text('Dart Version: '), findsOneWidget);
```
CoderDake, line 147: It feels a bit weird that we check for the line, but not necessarily that we are presenting the information that the feature is based around.
Would it be a big lift to check that we are presenting those values?

kenzieschmoll: Fixed 👍

CoderDake (approved):
Great additions to the specs!! Solid!

PR #4113, opened 2022-05-17
Add copy actions to network requests (Copy as URL, cURL)

### LINKED ISSUE #3042: feature request: copy http request "as curl"
Many (most?) HTTP inspection tools, proxies, etc have a "copy as curl" feature, eg. Charles Proxy:
charles-curl

and [Chrome DevTools](https://umaar.com/dev-tips/3-copy-as-curl/).

[This package ](https://pub.dev/packages/curl) could be a good starting point for this, though I believe it currently does not yet support query params.

Even round-tripping could be possible as @csells has [added Dart support for going from curl to Dart http client syntax](https://github.com/NickCarneiro/curlconverter/blob/master/generators/dart.js).

jankuss:
This PR adds the ability to copy HTTP requests in the Network screen.
Right now it is possible to Copy the URL, and to Copy as a cURL command.

https://user-images.githubusercontent.com/1659532/168923823-cff7eea5-a7eb-4333-a889-5e0146096aaa.mov

This closes #3042 

As per https://github.com/flutter/devtools/issues/3042#issuecomment-1127845746, we use the `Icons.more_vert` for the button which opens the menu.
The menu button is only shown when the request is selected.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/devtools_app/lib/src/shared/curl_command.dart
```diff
@@ -0,0 +1,132 @@
...
+
+    for (int index = 0; index < commandParts.length; index++) {
+      if (addSpace) {
+        commandString += ' ';
+      }
+      commandString += commandParts[index];
+
+      // Insert a line break if the index is contained in `lineBreakIndexes`.
+      // Don't insert when this is the last line.
+      if (lineBreakIndexes.contains(index) &&
```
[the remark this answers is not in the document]
jankuss: good idea 👍 
When building the string, we can just check if the previous `part` was a line break string (`\\\n`), and decide if we prefix a space or not based on that.

jankuss:
Thanks for the review :+1:

I have made adjustments based on your suggestions. I think the `CurlCommand` class is now easier to grasp.

The comments https://github.com/flutter/devtools/pull/4113#discussion_r876483581 and https://github.com/flutter/devtools/pull/4113#discussion_r876480806 clashed a little, since the `_headers()` and `_body()` couldn't be called in the factory because they were instance members. I solved this by marking them `static`, and passing the required information as parameters instead.

In this case when we reduce this even further, the properties `DartIOHttpRequestData data`, `bool followRedirects` and `bool multiline` are not needed anymore, as they are not accessed in anyway. The only thing that remains is the `List<String> commandParts`.

packages/devtools_app/test/http/curl_command_test.dart
```diff
@@ -0,0 +1,284 @@
...
+    return null;
+  }
+
+  @override
+  Future<void> getFullRequestData() async {
+    // Do nothing
+  }
+}
+
+DartIOHttpRequestData _getMockedDartIOHttpRequestData({
```
kenzieschmoll, line 239: nit: avoid starting method names with get. rename to _mockedDartIOHttpRequestData

packages/devtools_app/test/http/curl_command_test.dart
```diff
@@ -0,0 +1,284 @@
...
+
+      expect(
+        command.toString(),
+        "curl --location --request POST \'https://jsonplaceholder.typicode.com/posts\' \\\n--data-raw '    {\n      title: '\\''foo'\\'',\n      body: '\\''bar'\\'',\n      userId: 1,\n    }\n    '",
+      );
+    });
+  });
+}
+
+class _MockDartIOHttpRequestData extends DartIOHttpRequestData {
```
kenzieschmoll, line 211: nit: rename to _TestDartIOHttpRequestData since it is not truly a mock or a fake

packages/devtools_app/test/http/curl_command_test.dart
```diff
@@ -0,0 +1,284 @@
...
+  _MockDartIOHttpRequestData(
+    int timelineMicrosBase,
+    this._mockRequest,
+  ) : super(timelineMicrosBase, _mockRequest);
+
+  final HttpProfileRequest _mockRequest;
```
kenzieschmoll, line 217: nit: rename to _request for same reason as above

packages/devtools_app/lib/src/http/curl_command.dart
```diff
@@ -0,0 +1,126 @@
...
+  @override
+  String toString() {
+    return _buildCommandString(commandParts);
+  }
+
+  static List<String> _headers(
+    DartIOHttpRequestData data, {
+    required bool multiline,
+  }) {
+    final List<String> parts = [];
```
kenzieschmoll, line 50: nit: put type on RHS to match style of this codebase:
`final parts = <String>[];`

packages/devtools_app/lib/src/screens/network/network_screen.dart
```diff
@@ -408,6 +412,80 @@ class MethodColumn extends ColumnData<NetworkRequest> {
...
+    if (options.isEmpty || !isRowSelected) return const SizedBox.shrink();
+
+    return PopupMenuButton(
+      icon: const Icon(Icons.more_vert),
+      padding: const EdgeInsets.symmetric(horizontal: densePadding),
+      splashRadius: _actionSplashRadius,
+      tooltip: '',
+      itemBuilder: (context) {
+        return options;
+      },
```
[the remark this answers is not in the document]
jankuss: Maybe more like `ìtemBuilder: (context) => options`? In line 475 we are doing a check if the options returned from `_buildOptions` are empty, and only display the `PopupMenuButton` when we actually have items to display.

Otherwise we would be building the options two times.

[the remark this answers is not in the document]
kenzieschmoll: Ah you're right. `ìtemBuilder: (context) => options` sgtm

kenzieschmoll (approved):
LGTM. Thanks for your contribution of this feature!

Issue #3999, opened 2022-04-19
Add mixin for comparing enums with Dart 2.17


elliette:
Enhanced enums feature is being added to Dart 2.17.

This will allow us to add a mixin to compare enum values, eg:

```
mixin EnumIndexOrdering<T extends Enum> on Enum implements Comparable<T> {
  @override
  int compareTo(T other) => index - other.index;
}

enum MediaSize with EnumIndexOrdering {
  extraSmall,
  small,
  medium,
  large,
  extraLarge,
}

if (size > MediaSize.s) {
  ....
}
```

kenzieschmoll:
Now that DevTools is built from Flutter 2.13.0-0.3.pre (Dart 2.17.0), this should be possible

PR #4121, opened 2022-05-20
Add a mixin for comparing enums by index

elliette:
Closes https://github.com/flutter/devtools/issues/3999

Adds a new mixin for comparing enums by index, and uses the mixin for the `MediaSize` enum so that we can do comparisons like:

`if (screenWidth <= MediaSize.xs) ...`


packages/devtools_app/lib/src/primitives/enum_utils.dart
```diff
@@ -32,3 +32,24 @@ class EnumUtils<T> {
 
   String? name(T enumEntry) => _reverseLookupTable[enumEntry];
 }
+
+mixin EnumIndexOrdering<T extends Enum> on Enum implements Comparable<T> {
+  @override
+  int compareTo(T other) => index - other.index;
```
kenzieschmoll, line 38: is index an int? `=> index.compareTo(other.index)` might be more clear

packages/devtools_app/lib/src/shared/status_line.dart
```diff
@@ -67,22 +67,21 @@ class StatusLine extends StatelessWidget {
...
       ],
-      if (!isNarrow && pageStatus != null) ...[
+      if (screenWidth > MediaSize.xs && pageStatus != null) ...[
         pageStatus,
         const BulletSpacer(),
       ],
-      buildConnectionStatus(context, isExtraNarrow),
-      if (isEmbedded && !isExtraNarrow) ...[
+      buildConnectionStatus(context, screenWidth),
+      if (screenWidth > MediaSize.xxs && isEmbedded) ...[
```
kenzieschmoll, line 84: nit: we could put `screenWidth > MediaSize.xxs` since it is used twice

kenzieschmoll (approved):

packages/devtools_app/lib/src/primitives/enum_utils.dart
elliette: Done!

packages/devtools_app/lib/src/shared/status_line.dart
elliette: Good point, done!

PR #4123, opened 2022-05-20
optional_child_children

jacob314:
For discussion. Visualization of what DevTools code looks like if we omit `child` and `children` and automatically remove `[` and `]` for simple list literal cases.
We don't handle the 27% of list literals in DevTools that aren't simple.

### DIFF
--- a/packages/devtools_app/lib/src/analytics/prompt.dart
+++ b/packages/devtools_app/lib/src/analytics/prompt.dart
@@ -43,13 +43,13 @@ class _AnalyticsPromptState extends State<AnalyticsPrompt> {
       builder: (context, showPrompt, child) {
         return Column(
           crossAxisAlignment: CrossAxisAlignment.start,
-          children: [
+          [
             if (showPrompt) child!,
-            Expanded(child: widget.child),
+            Expanded(widget.child),
           ],
         );
       },
-      child: Card(
+      Card(
         shape: RoundedRectangleBorder(
           borderRadius: BorderRadius.circular(defaultBorderRadius),
           side: BorderSide(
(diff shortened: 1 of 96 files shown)


packages/devtools_app/lib/src/analytics/prompt.dart
```diff
@@ -58,35 +58,31 @@ class _AnalyticsPromptState extends State<AnalyticsPrompt> {
...
-                    style: textTheme.headline5,
-                  ),
-                  CircularIconButton(
-                    icon: Icons.close,
-                    onPressed: _controller!.hidePrompt,
-                    backgroundColor: theme.canvasColor,
-                    foregroundColor: theme.colorScheme.contrastForeground,
-                  ),
-                ],
+            Row(
```
kenzieschmoll, line 66: I think it is more clear to always show the list literal `[ ]`

packages/devtools_app/lib/src/analytics/prompt.dart
```diff
@@ -58,35 +58,31 @@ class _AnalyticsPromptState extends State<AnalyticsPrompt> {
...
+                onPressed: _controller!.hidePrompt,
+                backgroundColor: theme.canvasColor,
+                foregroundColor: theme.colorScheme.contrastForeground,
               ),
-              _analyticsDescription(textTheme),
-              const SizedBox(height: denseRowSpacing),
-              _actionButtons(),
-            ],
+            ),
+            const Padding(
```
kenzieschmoll, line 80: without the list literal, it makes it easy to mistake where one list ends and the next widget begins. In this case, CircularIconButton is the last child in Row.children, and Padding is the next child in Column.children, but it is difficult at a glance to see that since there is not a `],` breaking up the children of Row and Row's end parenthesis

packages/devtools_app/lib/src/app.dart
```diff
@@ -266,7 +266,7 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
       appSizePageId: (_, __, ___) {
         return DevToolsScaffold.withChild(
           key: const Key('appsize'),
-          child: _providedControllers(
+          _providedControllers(
```
kenzieschmoll, line 269: in the case of calling a method, I don't know whether `_providedControllers` returns a list or a single widget, so I don't know if this is a single child or a list of children that I can add to.

packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -1531,97 +1531,85 @@ class _FlameChartHelpDialog extends StatelessWidget {
...
+            'alt/option + scroll • ',
+            style: theme.fixedFontStyle,
+          ),
+          Text(
+            'scroll left / right • ',
+            style: theme.fixedFontStyle,
+          ),
+          Text(
+            'a / d • ',
+            style: theme.fixedFontStyle,
```
kenzieschmoll, line 1556: I miss the extra indent that separates the set of children from the other properties on the Column. Now the children are at the same indent as the widget properties.

PR #4124, opened 2022-05-20
enable sort_child_properties_last lint

jacob314:
used dart-fix to fix all violations.

### DIFF
--- a/packages/devtools_app/analysis_options.yaml
+++ b/packages/devtools_app/analysis_options.yaml
@@ -5,7 +5,3 @@ analyzer:
     - build/**
     - lib/generated_plugin_registrant.dart
     - macos/
-# Enable this option if you need to test whether the inspector_polyfill_script is compatible with
-# versions of flutter that require null safety.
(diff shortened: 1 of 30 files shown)


elliette (approved):

packages/devtools_app/analysis_options.yaml
```diff
@@ -5,7 +5,7 @@ analyzer:
...
     - lib/generated_plugin_registrant.dart
     - macos/
-# Enable this option if you need to test whether the inspector_polyfill_script is compatible with
-# versions of flutter that require null safety.
-#  enable-experiment:
-#    - non-nullable
+
+linter:
+  rules:
+    - sort_child_properties_last
```
kenzieschmoll, line 11: is there any reason this shouldn't be where all our other lint rules are defined (../../analysis_options.yaml) ?

jacob314: User error. Fixed.

kenzieschmoll (approved):

PR #4132, opened 2022-05-23
Reduce area of concern for low level memory libraries and move some to 'memory/primitives'.

polina-c:
1. memory_graph_model.dart and memory_timeline.dart stopped depending on memory_controller.dart.
2. lowest level classes are moved to the folder 'primitives'

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/memory_controller.dart
+++ b/packages/devtools_app/lib/src/screens/memory/memory_controller.dart
@@ -25,7 +25,7 @@ import '../../ui/search.dart';
 import 'memory_graph_model.dart';
 import 'memory_protocol.dart';
 import 'memory_snapshot_models.dart';
-import 'primitives/filtered_libraries.dart';
+import 'primitives/filter_config.dart';
 import 'primitives/memory_service.dart';
 import 'primitives/memory_timeline.dart';
 
--- a/packages/devtools_app/lib/src/screens/memory/memory_controller.dart
+++ b/packages/devtools_app/lib/src/screens/memory/memory_controller.dart
@@ -22,12 +22,12 @@ import '../../shared/table.dart';
 import '../../shared/table_data.dart';
 import '../../shared/utils.dart';
 import '../../ui/search.dart';
-import 'memory_filter.dart';
 import 'memory_graph_model.dart';
 import 'memory_protocol.dart';
 import 'memory_service.dart';
 import 'memory_snapshot_models.dart';
 import 'memory_timeline.dart';
+import 'primitives/filtered_libraries.dart';
 
 enum ChartType {
   DartHeaps,
--- a/packages/devtools_app/lib/src/screens/memory/memory_controller.dart
+++ b/packages/devtools_app/lib/src/screens/memory/memory_controller.dart
@@ -24,10 +24,10 @@ import '../../shared/utils.dart';
 import '../../ui/search.dart';
 import 'memory_graph_model.dart';
 import 'memory_protocol.dart';
-import 'memory_service.dart';
 import 'memory_snapshot_models.dart';
-import 'memory_timeline.dart';
 import 'primitives/filtered_libraries.dart';
+import 'primitives/memory_service.dart';
+import 'primitives/memory_timeline.dart';
 
 enum ChartType {
   DartHeaps,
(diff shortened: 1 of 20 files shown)


packages/devtools_app/lib/src/screens/memory/primitives/README.md
```diff
@@ -0,0 +1 @@
+This folder contains code that does not depend on other code in the folder `memory`.
```
kenzieschmoll, line 1: What is the value added by creating a subfolder `memory/primitives`?

packages/devtools_app/lib/src/screens/memory/primitives/filter_config.dart
```diff
@@ -0,0 +1,106 @@
...
+  /// Hide any library with no constructed class instances.
+  final ValueNotifier<bool> filterLibraryNoInstances;
+
+  /// State of filters used by filter dialog (create/modify) and used
+  /// by filtering in grouping.
+  final FilteredLibraries libraryFilters;
+}
+
+/// State of the libraries, wildcard included, filtered (shown or hidden).
+/// groupBy uses this class to determine is the library should be filtered.
```
kenzieschmoll, line 39: wrap groupBy in brackets if it is a method name `[groupBy]`. You may need to add the class name if the method is not part of this class `[MyClass.groupBy]`

packages/devtools_app/lib/src/screens/memory/primitives/filter_config.dart
```diff
@@ -0,0 +1,106 @@
...
+  void addFilter(String libraryUri) {
+    _filteredLibraries.add(libraryUri);
+  }
+
+  void removeFilter(String libraryUri) {
+    _filteredLibraries.remove(libraryUri);
+  }
+
+  // Keys in the libraries map is a normalized library name.
+  List<String> sort() => _filteredLibraries..sort();
```
kenzieschmoll, line 83: replace `..` with `.`

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
```diff
@@ -419,7 +419,7 @@ class MemoryController extends DisposableController
 
   /// Source of memory heap samples. False live data, True loaded from a
   /// memory_log file.
-  bool offline = false;
+  final offline = ValueNotifier<bool>(false);
```
kenzieschmoll, line 422: I'm not seeing anywhere where we are actually listening to this notifier. Are we anywhere? If that is the case, there is no added value gained by converting to a ValueNotifier instead of a bool.

packages/devtools_app/lib/src/screens/memory/primitives/README.md
polina-c: The value is the same as with `src/primitives`. People, who look on the code, want to understand the code structure, , i.e which code is high level and which is low level, faster. 
The folder `primitives` contains low level code, used by other code. It helps to understand the relations between the libraries.

polina-c: I also plan to create the folder `areas`, with subfolders `chart`, `analysis` and `allocations`, similar to the folder 'screen' in the root.
It will help to navigate the code, especially knowing that we will add new functionality and number of files will increase.

packages/devtools_app/lib/src/screens/memory/primitives/filter_config.dart
polina-c: The method `sort` does not return any value, but we need to return sorted list in this method. So, the cascade operator is needed here.

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
polina-c: It is passed to MemoryTimeline, that needs up-to-date value:
https://github.com/flutter/devtools/blob/35b74c3198bfaa7ceb386b5fd9acf2cc56067a87/packages/devtools_app/lib/src/screens/memory/memory_controller.dart#L239

packages/devtools_app/lib/src/screens/memory/primitives/README.md
polina-c: sounds good

kenzieschmoll (approved):
one comment about devtools_app.dart import, then lgtm

PR #4143, opened 2022-05-25
Add clarifying comment about android memory.

polina-c:


packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -406,6 +406,8 @@ class ServiceConnectionManager {
 
   Future<Response> get adbMemoryInfo async {
     return await _callServiceOnMainIsolate(
+      // This service is registered on the vm service by flutter tools:
```
kenzieschmoll, line 409: we actually already have this information in a comment where the flutterMemory service is defined

https://github.com/flutter/devtools/blob/8265435a4ada389031f3ba773db6221192f397b9/packages/devtools_shared/lib/src/service_utils.dart#L19

Issue #3557, opened 2021-12-23
Add option to copy file name, not just package name


elliette:
This is a customer request. We currently have a button to copy the file name, which is confusing because it actually copies the package name. We should rename the button, and ideally add another button to copy the actual file path. It would be great if we could also show the file path in the debugger view.

PR #4147, opened 2022-05-26
Add ability to copy file and package from debugger

CoderDake:
Fixes https://github.com/flutter/devtools/issues/3557

This PR:
- Renames the package copy option so that it reflects it's actual behaviour
- Adds the ability to copy the full file path

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -1126,19 +1126,41 @@ class ScriptPopupMenuOption {
...
-  copyScriptNameOption,
+  copyPackageOption,
+  copyFilePathOption,
   goToLineOption,
   openFileOption,
 ];
 
-final copyScriptNameOption = ScriptPopupMenuOption(
-  label: 'Copy filename',
+final copyPackageOption = ScriptPopupMenuOption(
```
kenzieschmoll, line 1135: nit: rename to copyPackagePathOption to be consistent with the title as well as copyFilePathOption below

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -1126,19 +1126,41 @@ class ScriptPopupMenuOption {
...
 );
 
+final copyFilePathOption = ScriptPopupMenuOption(
+  label: 'Copy file path',
+  icon: Icons.content_copy,
+  onSelected: (_, controller) async {
+    String? filePath;
+    final packagePath = controller.scriptLocation.value!.scriptRef.uri;
+    if (packagePath != null) {
+      final results = await serviceManager.service!.lookupResolvedPackageUris(
```
kenzieschmoll, line 1150: can we use ResolvedUriManager instead so that we can pull this from the cache if possible?

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -1126,19 +1126,41 @@ class ScriptPopupMenuOption {
...
+        serviceManager.isolateManager.selectedIsolate.value!.id!,
+        [packagePath],
+      );
+      if (results.uris != null) {
+        filePath = results.uris![0];
+      }
+    }
+    return Clipboard.setData(
+      ClipboardData(text: filePath),
```
polina-c, line 1159: What if filePath is null here? We do not want to copy null to clopboart, right?
I suggest to replace ifs with asserts if we know for sure packagePath and results.uris are not nulls.

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -1126,19 +1126,41 @@ class ScriptPopupMenuOption {
...
   onSelected: (_, controller) => Clipboard.setData(
     ClipboardData(text: controller.scriptLocation.value?.scriptRef.uri),
   ),
 );
 
+final copyFilePathOption = ScriptPopupMenuOption(
+  label: 'Copy file path',
+  icon: Icons.content_copy,
+  onSelected: (_, controller) async {
+    String? filePath;
```
polina-c, line 1147: How about moving the handler code to a method and creating test coverage for it?

packages/devtools_app/lib/src/screens/debugger/codeview.dart
CoderDake: I had thought about this but noticed the original copy didn't take it into account either. 
I purposely let this fall through to null so that we are no worse off than the original behaviour.

From my understanding asserts don't work in a fully built app, would those asserts just be there for our testing purposes?

packages/devtools_app/lib/src/screens/debugger/codeview.dart
CoderDake: I definitely like this idea.
I tried making a flutter test through the UI for this and it was a nightmare

packages/devtools_app/lib/src/screens/debugger/codeview.dart
kenzieschmoll: > From my understanding asserts don't work in a fully built app, would those asserts just be there for our testing purposes?

Correct. Asserts only run in debug mode.

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
```diff
@@ -10,43 +10,59 @@ class ResolvedUriManager {
   ///
   /// The map keys and values are as follows:
   /// `_isolateResolvedUrlMap[isolateId][resolvedUrl] = packageUrl`
```
kenzieschmoll, line 12: nit: update naming in comment

kenzieschmoll (approved):

packages/devtools_app/lib/src/service/resolved_uri_manager.dart
CoderDake: great catch!

PR #4150, opened 2022-05-27
Create `ProvidedControllerMixin` and use throughout DevTools

kenzieschmoll:
Deleted `PerformanceControllerMixin` and `MemoryControllerMixin` (fyi @polina-c) in favor of general `ProvidedControllerMixin` that can be used anywhere in DevTools. 

This reduces the boiler plate for using an inherited controller from 5 lines down to 2 lines.

Before:
```
ControllerType get controller => _controller!;
ControllerType? _controller;

final newController = Provider.of<ControllerType>(context);
if (newController == _controller) return;
_controller = newController;
```

After:
```
with ProvidedControllerMixin<ControllerType, WidgetType> // add mixin to state class
initController() // call in didChangeDependencies
```

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/memory_heap_treemap.dart
+++ b/packages/devtools_app/lib/src/screens/memory/memory_heap_treemap.dart
@@ -4,32 +4,27 @@
 
 import 'package:flutter/material.dart' hide TextStyle;
 import 'package:flutter/widgets.dart' hide TextStyle;
-import 'package:provider/provider.dart';
 
 import '../../charts/treemap.dart';
 import '../../primitives/auto_dispose_mixin.dart';
 import '../../shared/common_widgets.dart';
 import '../../shared/theme.dart';
+import '../../shared/utils.dart';
 import 'memory_controller.dart';
 import 'primitives/predefined_classes.dart';
 
 class MemoryHeapTreemap extends StatefulWidget {
-  const MemoryHeapTreemap(this.controller);
-
-  final MemoryController controller;
+  const MemoryHeapTreemap();
 
   @override
-  MemoryHeapTreemapState createState() => MemoryHeapTreemapState(controller);
+  MemoryHeapTreemapState createState() => MemoryHeapTreemapState();
 }
 
 class MemoryHeapTreemapState extends State<MemoryHeapTreemap>
-    with AutoDisposeMixin {
-  MemoryHeapTreemapState(this._controller);
-
+    with
+        AutoDisposeMixin,
+        ProvidedControllerMixin<MemoryController, MemoryHeapTreemap> {
   InstructionsSize? _sizes;
-
-  bool _initialized = false;
-  late MemoryController _controller;
 
   TreemapNode? root;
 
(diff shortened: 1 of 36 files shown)


packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -92,3 +93,31 @@ extension VmExtension on VM {
     ].join(' ');
   }
 }
+
+/// Mixin that provides a [controller] from package:provider for a State class.
+///
+/// [initController] must be called from [State.didChangeDependencies]. If
+/// [initController] returns false, [didChangeDependencies]
+mixin ProvidedControllerMixin<T, V extends StatefulWidget> on State<V> {
```
polina-c, line 101: Can it happen that one widget will need two controllers? If yes, what should we do with name conflict?

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -92,3 +93,31 @@ extension VmExtension on VM {
...
+  T? _controller;
+
+  /// Initializes [_controller] from package:provider.
+  ///
+  /// This method should be called in [didChangeDependencies]. Returns whether
+  /// or not a new controller was provided upon subsequent calls to
+  /// [initController].
+  ///
+  /// This method will commonly be used to return early from
+  /// [didChangeDependencies] when initialization code should not be ran again
```
CoderDake, line 113:
```suggestion
  /// [didChangeDependencies] when initialization code should not be run again
```

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -92,3 +93,31 @@ extension VmExtension on VM {
     ].join(' ');
   }
 }
+
+/// Mixin that provides a [controller] from package:provider for a State class.
+///
+/// [initController] must be called from [State.didChangeDependencies]. If
+/// [initController] returns false, [didChangeDependencies]
```
CoderDake, line 100: It feels like something might be missing here.
Does this comment need to be updated?

packages/devtools_app/lib/src/shared/utils.dart
kenzieschmoll: It could. For those cases we can either:
a) use `ProvidedControllerMixin` for one of the controllers (the primary one if that applies) and pull the other directly from provider via `Provider.of<ControllerType>(context)`
b) pull both controllers from provider via `Provider.of<ControllerType>(context)` and do not use `ProvidedControllerMixin` at all

packages/devtools_app/lib/src/screens/memory/memory_heap_treemap.dart
```diff
@@ -4,57 +4,49 @@
...
-  bool _initialized = false;
-  late MemoryController _controller;
-
   TreemapNode? root;
 
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
 
     // TODO(terry): Unable to short-circuit need to investigate why?
```
CoderDake, line 35: Is this TODO needed anymore?

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -92,3 +93,31 @@ extension VmExtension on VM {
...
+  /// This method should be called in [didChangeDependencies]. Returns whether
+  /// or not a new controller was provided upon subsequent calls to
+  /// [initController].
+  ///
+  /// This method will commonly be used to return early from
+  /// [didChangeDependencies] when initialization code should not be ran again
+  /// if the provided controller has not changed.
+  ///
+  /// E.g. `if (!initController()) return;`
+  bool initController() {
```
CoderDake, line 117: Would we ever run into issues with super.didChangeDependencies overriding each others controllers?

packages/devtools_app/lib/src/screens/memory/memory_heap_treemap.dart
kenzieschmoll: I am not sure. Leaving the TODO since this is out of scope of this PR.

packages/devtools_app/lib/src/shared/utils.dart
kenzieschmoll: not that I am aware of. It is uncommon for stateful widgets to use an extension pattern as opposed to a composition pattern. There is only one place in DevTools that I am aware of that has widgets and states extend other widgets and states, and they do not have conflicting controllers.

packages/devtools_app/lib/src/shared/utils.dart
kenzieschmoll: done

polina-c (approved):

Issue #2952, opened 2021-04-28
Use searchable, collapsible tree for network response tab


bryanhinkson:
I don't know if this is by design or not, but you are not able to select (and as a result copy) text in the network tab section.  This would especially be useful for responses.  There are times that I want to copy the text so I can format and search it for what I am looking for.  With some API's I am presented with a wall of plain text which is painful to read through.

I am not sure if that is a bug or a feature request but it would be really nice.  As long as I am requesting changes here are some other improvements that I think would make the Network experience better:

It would also be nice to be able to search the Network->response for text.

It would be really nice to be able to format that text into a collapsible JSON tree. (perhaps just for application/json responses) That would make it much easier to navigate through the responses.

These features are in the chrome dev tools experience and I believe that they would translate very well to the flutter dev tools as well.

Thanks for all you do!

kenzieschmoll:
These are great improvements - thank you for the feature requests!

kenzieschmoll:
What text in particular are you trying to select but are unable to?

kenzieschmoll:
some additional user feedback around the response tab: "i would like to have nicer formatting for responses. view response is too slow if the response body is too big"

PR #4167, opened 2022-06-03
Add `JsonViewer` widget and update formatting / highlighting of `ExpandableVariable`

bkonyi:
The `JsonViewer` widget wraps `ExpandableVariable`, displaying a tree of fake VM service `Instance`s built from a JSON object. These fake `Instance`s are stored in an instance of `JsonToServiceCache`, which implements `getObject(...)`, allowing for calls to `serviceManager.service.getObject(...)` to first check against the cache before speaking to the VM service.

This change also modifies the `ExpandableVariable` widget structure to condense `Map` entries into a single row, and also introduces some syntax highlighting.

`JsonViewer` in the network response tab:
image

Updated variable viewer in the debugger:
image

Fixes https://github.com/flutter/devtools/issues/2952.


packages/devtools_app/lib/src/shared/object_tree.dart
```diff
@@ -0,0 +1,655 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
```
bkonyi, line 1: The vast majority of this file is just a straight cut-paste from `debugger_model.dart`

packages/devtools_app/lib/src/shared/object_tree.dart
```diff
@@ -0,0 +1,655 @@
...
+
+  /// Creates a variable from a value that must be an InstanceRef or a primitive
+  /// type.
+  ///
+  /// [value] should typically be an [InstanceRef] but can also be a [Sentinel]
+  /// [ObjRef] or primitive type such as num or String.
+  factory DartObjectNode.fromValue({
+    String? name,
+    required Object? value,
+    bool artificialName = false,
```
bkonyi, line 439: The only changes in this code relate to this new `artificialName` property.

CoderDake (commented):
Still reviewing, but leaving a couple of comments for now

packages/devtools_app/lib/src/shared/object_tree.dart
CoderDake, line 439: Adding a description for `artificialName` might be helpful in the future adventures in this aread

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -284,6 +289,14 @@ class VmServiceWrapper implements VmService {
     int? offset,
     int? count,
   }) {
+    final cachedObj = fakeServiceCache.getObject(
```
kenzieschmoll, line 292: should we really be adding this logic in vm_service_wrapper? this seems like it should belong in `JsonToServiceCache` or some other helper / manager.

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1421,6 +1427,74 @@ class FormattedJson extends StatelessWidget {
   }
 }
 
+class JsonViewer extends StatefulWidget {
+  const JsonViewer({
+    required this.json,
+  });
+
+  final String json;
```
kenzieschmoll, line 1435: maybe rename to 'encodedJson'? when I see a var named `json` I immediately think it is a Map

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1421,6 +1427,74 @@ class FormattedJson extends StatelessWidget {
...
+      padding: const EdgeInsets.all(denseSpacing),
+      child: SingleChildScrollView(
+        child: FutureBuilder(
+          future: _initializeTree,
+          builder: (context, snapshot) {
+            if (snapshot.connectionState != ConnectionState.done)
+              return Container();
+            return ExpandableVariable(
+              variable: variable,
+              debuggerController: Provider.of<DebuggerController>(context),
```
kenzieschmoll, line 1489: pull the debugger controller in didChangeDependencies so we don't have to look it up from package:provider on each build. Better yet, use the new `ProvidedControllerMixin` on this state class.

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1421,6 +1427,74 @@ class FormattedJson extends StatelessWidget {
...
+      value: root,
+      artificialName: true,
+      isolateRef: IsolateRef(
+        id: 'fake-isolate',
+        number: 'fake-isolate',
+        name: 'local-cache',
+        isSystemIsolate: true,
+      ),
+    );
+    _initializeTree = buildVariablesTree(variable);
```
kenzieschmoll, line 1465: instead of adding a fakeServiceCache to the VmServiceWrapper, should we instead just manipulate `buildVariablesTree` to support passing in the fake dart object directly? And then in `buildVariablesTree` we can _not_ call `service.getObject` when the fake object has been passed in directly. Then we don't have to leak in to the vm service layer to implement this.

packages/devtools_app/lib/src/shared/object_tree.dart
CoderDake: @bkonyi could you give some context around the changes to _createVariablesForAssociations?
image

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
CoderDake: From what I understand, this is complicated by the fact that the widgets are hardcoded to depend on this global. So changing that dependency, (e.g. passing in the service manager instead of accessing it through the global, so that we could pass a fake one when we are using it for different purposes) would take a lot of work in refactoring.

bkonyi: I figured this would also give us support for adding data to the cache for use by other widgets with minimal overhead. `Map.contains` should be orders of magnitude less expensive than an actual service request, so there shouldn't be any noticeable performance difference.

packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
```diff
@@ -169,18 +167,17 @@ class HttpResponseView extends StatelessWidget {
     // timeline profiler since it's possible for response body data to get
     // dropped.
     final contentType = data.contentType;
+    final responseBody = data.responseBody;
```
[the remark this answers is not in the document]
bkonyi: I think I was under the assumption that it was valid to have a null `responseBody`, but that doesn't seem to be the case.

packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
```diff
@@ -169,18 +167,17 @@ class HttpResponseView extends StatelessWidget {
...
+    } else if (contentType != null &&
+        contentType.contains('json') &&
+        responseBody!.isNotEmpty) {
+      child = JsonViewer(responseBody);
     } else {
-      child = FormattedJson(
-        formattedString: data.responseBody,
-      );
+      // TODO(bkonyi): style
+      child = Text(responseBody!);
```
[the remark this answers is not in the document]
bkonyi: TBH, I'm not 100% sure. Most data has `\n` included, or it's JSON and we format it anyway. It's definitely worth looking into.

packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
```diff
@@ -169,18 +167,17 @@ class HttpResponseView extends StatelessWidget {
...
+    } else if (contentType != null &&
+        contentType.contains('json') &&
+        responseBody!.isNotEmpty) {
+      child = JsonViewer(json: responseBody);
     } else {
-      child = FormattedJson(
-        formattedString: data.responseBody,
-      );
+      // TODO(bkonyi): style
+      child = Text(responseBody!);
```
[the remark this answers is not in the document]
bkonyi: Yup, definitely.

packages/devtools_app/lib/src/shared/common_widgets.dart
bkonyi: Done.

packages/devtools_app/lib/src/shared/common_widgets.dart
bkonyi: As I mentioned in my comment above, I don't think there's much benefit to having fake objects treated differently. If there was a performance implication, I'd totally agree it's worth getting decoupled from the service layer, but service requests are significantly more expensive than a `Map.contains` call. Also, if we're still concerned by the `Map.contains` call, all of the IDs for objects stored in `JsonToServiceCache` have IDs starting with `json-cache-` so a simple string comparison could even be used to short circuit.

I just don't think it's worth the extra effort to add special cases to pass fake `Instance` objects around when we already have the ability to support accessing them at little to no cost via existing mechanisms.

packages/devtools_app/lib/src/shared/object_tree.dart
```diff
@@ -276,18 +276,48 @@ List<DartObjectNode> _createVariablesForAssociations(
 ) {
   final variables = <DartObjectNode>[];
   final associations = instance.associations ?? [];
+
+  // If the key type for the provided associations is not primitive, we want to
+  // allow for users to drill down into the key object's properties. If we're
+  // only dealing with primative types as keys, we can render a flatter
+  // representation.
+  final hasNonPrimitiveKey = associations.fold<bool>(
```
kenzieschmoll, line 284: should this be `hasPrimitiveKey`?

kenzieschmoll (approved):
lgtm with one comment

packages/devtools_app/lib/src/shared/object_tree.dart
bkonyi: Derp, yup. The coffee hadn't kicked in yet... :-P

xieenming:
Is there an option to switch between this new show format and the JSON show format.

The original JSON show format is like : 

```json
{
    "err": {
        "ec": 0,
        "em": "ok",
        "dm": "ok"
    },
    "ret": {
        "has_prev": false,
        "has_next": false,
        "limit": 100,
        "offset": 0,
        "data": [
            {
                "id": "box_1001",
                "services": []
            },
        ]
    }
}
```

PR #4169, opened 2022-06-03
Add missing `textSelectionColor` 

elliette:
We are using `theme.selectedTextStyle` in [`breakpoints.dart`](https://github.com/flutter/devtools/blob/278272dae6bd2d067426ff11cf7077a5ddfdbc74/packages/devtools_app/lib/src/screens/debugger/breakpoints.dart#L76-L79) and [`call_stack.dart`](https://github.com/flutter/devtools/blob/278272dae6bd2d067426ff11cf7077a5ddfdbc74/packages/devtools_app/lib/src/screens/debugger/call_stack.dart#L72) but its properties are not set and therefore are all `null`. One consequence of this is that the circles next to breakpoints don't show up when we select them (because we are passing in `null` as color). 

This PR explicitly sets the color property to fix the selected breakpoints UI.

Before:
Screen Shot 2022-06-03 at 3 30 56 PM
Screen Shot 2022-06-03 at 3 29 38 PM

After:
Screen Shot 2022-06-03 at 3 30 49 PM
Screen Shot 2022-06-03 at 3 29 45 PM

### DIFF
--- a/packages/devtools_app/lib/src/shared/theme.dart
+++ b/packages/devtools_app/lib/src/shared/theme.dart
@@ -54,6 +54,7 @@ ThemeData _darkTheme(IdeTheme ideTheme) {
     backgroundColor: background,
     indicatorColor: devtoolsBlue[400]!,
     selectedRowColor: devtoolsGrey[600]!,
+    textSelectionColor: Colors.black,
   );
 }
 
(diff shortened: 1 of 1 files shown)


kenzieschmoll (approved):
LGTM. One question though - have we manually verified that this doesn't negatively affect any other parts of DevTools that may have been inheriting textSelectionTheme?

elliette:
> LGTM. One question though - have we manually verified that this doesn't negatively affect any other parts of DevTools that may have been inheriting textSelectionTheme?

Looks like the only other place we're using it is in the call stack, which looks similar to breakpoints, so we should be good!

Screen Shot 2022-06-03 at 4 35 22 PM

PR #4173, opened 2022-06-07
Add isolate selection support to memory screen + other minor UI fixes

bkonyi:
While the live feed appears to display the sum of all memory consumption of each isolate, the heap snapshot mechanism only collects a snapshot for the currently selected isolate. Since there was no isolate picker enabled for the memory screen and no indication that snapshots are isolate dependent, there was no way to differentiate between snapshots and where they came from.

This change enables the isolate picker for the memory screen and displays the name of the active isolate corresponding to each snapshot to make it clearer that the snapshots correspond to individual isolates.

In addition, this change makes some minor UI tweaks, particularly with respect to how borders are drawn around the various components (e.g., there was no borders around widgets in a `Split`, making the actual footprint of the draggable bar unclear.

Finally, there's a fix for an issue where switching between the 'Analysis' and 'Allocations' tabs with classes selected for tracking would result in a null assertion being hit when trying to generate the 'Tracking' table. It seems that the `AllocationTableView` isn't kept
alive when switching tabs, causing a rebuild where we assumed some state was already initialized.

**Original Borders**
image

**Updated Borders + Isolate Selection**
image

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/memory_allocation_table_view.dart
+++ b/packages/devtools_app/lib/src/screens/memory/memory_allocation_table_view.dart
@@ -80,15 +80,6 @@ class AllocationTableViewState extends State<AllocationTableView>
     super.didChangeDependencies();
     if (!initController()) return;
 
-    // Make sure that we actually initialize the tracker tree rather than
-    // relying on `controller.updateClassStackTraces` to be updated.
-    // TODO: find out why AllocationTableView isn't persisting when switching
-    // between the 'Analysis' and 'Allocations' tabs.
-    trackerData.createTrackerTree(
-      controller.trackAllocations,
-      controller.allocationSamples,
-    );
-
     cancelListeners();
 
     // TODO(terry): setState should be called to set our state not change the
--- a/packages/devtools_app/lib/src/screens/memory/memory_allocation_table_view.dart
+++ b/packages/devtools_app/lib/src/screens/memory/memory_allocation_table_view.dart
@@ -10,6 +10,7 @@ import 'package:vm_service/vm_service.dart';
 
 import '../../primitives/auto_dispose_mixin.dart';
 import '../../primitives/utils.dart';
+import '../../shared/common_widgets.dart';
 import '../../shared/split.dart';
 import '../../shared/table.dart';
 import '../../shared/table_data.dart';
@@ -63,7 +64,6 @@ class AllocationTableViewState extends State<AllocationTableView>
   @override
   void initState() {
     super.initState();
-
     // Setup the columns.
     columns.addAll([
       FieldTrack(),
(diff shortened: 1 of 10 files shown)


packages/devtools_app/lib/src/screens/memory/memory_heap_tree_view.dart
```diff
@@ -1464,11 +1466,12 @@ class MemoryHeapTableState extends State<MemoryHeapTable>
...
       );
 
-      return controller.groupByTreeTable;
+      result = controller.groupByTreeTable;
     } else {
       // Nothing collected yet (snapshots/analyses) - return an empty area.
-      return const SizedBox();
+      result = const SizedBox();
     }
+    return OutlineDecoration(child: result);
```
kenzieschmoll, line 1474: does an OutlineDecoration around an empty SizedBox really do what we think it does? since the SizedBox has size 0,0 I would expect to see nothing.

packages/devtools_app/lib/src/screens/memory/memory_tracker_model.dart
```diff
@@ -141,7 +142,8 @@ class TreeTracker {
     Map<String, ClassRef> classesTracked,
     List<AllocationSamples> allocationSamples,
   ) {
-    final tree1Local = tree1 = TrackerClass('_ROOT_tracking');
+    tree1 ??= TrackerClass('_ROOT_tracking');
+    final tree1Local = tree1!;
```
kenzieschmoll, line 146: General dart null safety comment - the fact that we have to null assert here is surprising 🤔

packages/devtools_app/lib/src/screens/memory/memory_heap_tree_view.dart
bkonyi: Ah yeah, good point. Doesn't seem to make any difference.

packages/devtools_app/lib/src/screens/memory/memory_tracker_model.dart
bkonyi: `tree` isn't a local property so it could, in theory, be set to null again between the allocation and the assignment to `tree1Local` if there was any sort of asynchrony between those lines. There obviously can't be an asynchronous gap here in this case, but I'm guessing the CFE can't detect that for one reason or another.

packages/devtools_app/lib/src/screens/memory/memory_allocation_table_view.dart
```diff
@@ -80,6 +80,15 @@ class AllocationTableViewState extends State<AllocationTableView>
     super.didChangeDependencies();
     if (!initController()) return;
 
+    // Make sure that we actually initialize the tracker tree rather than
+    // relying on `controller.updateClassStackTraces` to be updated.
+    // TODO: find out why AllocationTableView isn't persisting when switching
```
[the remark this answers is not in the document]
bkonyi: Ah that fixes it, although I just noticed we're doing something weird with the state management for the table displaying the actual tracking information since it doesn't update when you add new classes from the list. I tried messing around with it a bit, and it seems like we're not calling `TreeTable._initData()` in `didUpdateWidget()` since the `collectionEquals(widget.dataRoots, oldWidget.dataRoots) == true)`. Not sure why `oldWidget.dataRoots.length` is the same as `widget.dataRoots.length` if we're rebuilding. We're going to want to look at that in a separate PR.

kenzieschmoll (approved):

Issue #3238, opened 2021-08-02
Poor performance on "Logging" tab when lots of events are emmited


nshahan:
Repro: 
1. Open the Flutter Gallery using Flutter Web `flutter run -d chrome`.
2. Navigate to a page with constant animation running. Ex: Material -> Progress indicators
3. Open DevTools and connect to the app.
4. Switch to the "Logging" tab in DevTools.

The DevTools app becomes unresponsive when events are flowing in for every animation frame.

I'm running flutter from master.

```
$ flutter --version
Flutter 2.5.0-2.0.pre.118 • channel master • git@github.com:nshahan/flutter.git
Framework • revision 22436cca0e (82 minutes ago) • 2021-08-02 14:40:04 -0700
Engine • revision 29f9df723f
Tools • Dart 2.14.0 (build 2.14.0-368.0.dev)
```

cc @kenzieschmoll

KlausJokisuo:
Any news on this? Performance is terrible on the logging tab. Also, the logging gets spammed with Flutter's log messages.

CoderDake:
I've found out that this issue can amended by setting an itemExtent on the ListView. I'm looking into adding this functionality into our table builder to ensure that the logger will have a much cleaner scrolling experience :)

CoderDake:
Master now has the changes that make the event log useable when there are lots of events :)

PR #4175, opened 2022-06-07
Fix Slow Event Log Scrolling for large # of events

CoderDake:
## Details

Fixes https://github.com/flutter/devtools/issues/3238

It is a known issue that ListItem.builder lists can have performance issues when scrolling large lists.
The reasoning for this is that the height of each item needs to be calculated when scrolling along a list. So the list being scrolled can't just skip to the new location.

If the height of each row is the same, this performance hit can be avoided by telling the ListView the height of each row ahead of time.
This is done by setting the [`itemExtent`](https://api.flutter.dev/flutter/widgets/ListView/itemExtent.html).

In our case the height of each row height is the same and should be `defaultRowHeight`

This also benefits the scrolling speed on the profiler page

### Original Behaviour
https://user-images.githubusercontent.com/1386322/172468485-5eb6fcd5-fbb5-4866-a1b8-abfe2ef6b50b.mp4

### Fixed Behaviour
https://user-images.githubusercontent.com/1386322/172468398-8d6969e9-23ba-44bf-b6b5-5f53912d5241.mp4

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/devtools_app/lib/src/shared/table.dart
+++ b/packages/devtools_app/lib/src/shared/table.dart
@@ -967,6 +967,7 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
                       child: ListView.builder(
                         controller: scrollController,
                         itemCount: widget.data.length,
+                        itemExtent: defaultRowHeight,
                         itemBuilder: _buildItem,
                       ),
                     ),
--- a/packages/devtools_app/lib/src/shared/table.dart
+++ b/packages/devtools_app/lib/src/shared/table.dart
@@ -235,6 +235,7 @@ class FlatTableState<T> extends State<FlatTable<T>>
           secondarySortColumn: widget.secondarySortColumn,
           onSortChanged: _sortDataAndUpdate,
           activeSearchMatchNotifier: widget.activeSearchMatchNotifier,
+          itemExtent: defaultRowHeight,
         );
       },
     );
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -815,14 +818,16 @@ class _Table<T> extends StatefulWidget {
   int get numSpacers => max(0, columns.length - 1);
 
   @override
-  _TableState<T> createState() => _TableState<T>();
+  _TableState<T> createState() => _TableState<T>(rowItemExtent: rowItemExtent);
 }
 
 class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
+  _TableState({this.rowItemExtent});
```
kenzieschmoll, line 825: never pass parameters into state objects. You can access the class widget class member from the state object via `widget.rowItemExtent`

CoderDake: this is a good life lesson thank you :)

kenzieschmoll (approved):

packages/devtools_app/lib/src/shared/table.dart
jacob314: I'm surprised the no-logic in createState lint didn't warn about this.

PR #4178, opened 2022-06-07
Removing Retry Logic 

### LINKED ISSUE #4177: Remove retry logic when dart-lang/webdev/issues/1439 has been fixed
See: https://github.com/flutter/devtools/blob/a5250b6495a3aef8bff593dbb0f7d7d1cae5a23a/packages/devtools_app/lib/src/screens/debugger/debugger_model.dart#L527-L529

Originally added as a patch to fix a bug in DWDS where an error is thrown if `getObject` is called with offset/count for an object that has no length. Created by @elliette

carolynqu:
Closes #4177

Removing `_getObjectWithRetry` after dart-lang/webdev/issues/1439 landed.


packages/devtools_app/lib/src/screens/debugger/debugger_model.dart
```diff
@@ -524,29 +524,6 @@ Future<void> buildVariablesTree(
...
-  String objectId,
-  DartObjectNode variable,
-) async {
-  final variableId = variable.ref!.isolateRef!.id!;
-  try {
-    final dynamic result = await serviceManager.service!.getObject(
-      variableId,
-      objectId,
-      offset: variable.offset,
-      count: variable.childCount,
```
kenzieschmoll, line 540: @elliette correct me if I am wrong, but we still want `offset` and `count` to be included in the call above don't we?

packages/devtools_app/lib/src/screens/debugger/debugger_model.dart
```diff
@@ -422,7 +422,7 @@ Future<void> buildVariablesTree(
   } else if (instanceRef != null && serviceManager.service != null) {
     try {
       final dynamic result =
-          await _getObjectWithRetry(instanceRef.id!, variable);
+          await serviceManager.service!.getObject(variable.ref!.isolateRef!.id!, instanceRef.id!);
```
kenzieschmoll, line 425: nit: pull `variable.ref!.isolateRef!.id!` out into a variable above like it was before, just to make this a little more readable

packages/devtools_app/lib/src/screens/debugger/debugger_model.dart
elliette: Yes, we still want to include offset and count :)

elliette (approved):
LGTM! Looks like there is a merge conflict to resolve

packages/devtools_app/lib/src/screens/debugger/debugger_model.dart
```diff
@@ -420,9 +420,11 @@ Future<void> buildVariablesTree(
       start += count;
     }
   } else if (instanceRef != null && serviceManager.service != null) {
+    final variableId = variable.ref!.isolateRef!.id!;
     try {
-      final dynamic result =
-          await _getObjectWithRetry(instanceRef.id!, variable);
+      final dynamic result = await serviceManager.service!.getObject(
+          variableId, instanceRef.id!,
+          offset: variable.offset, count: variable.childCount);
```
kenzieschmoll, line 427: nit: add a trailing comma after `variable.childCount` --> `variable.childCount,`. This will allow the dart formatter to format the code properly. Generally, whenever a method call has to break to a new line, you should add a trailing comma. We have a lint check for this enabled for the DevTools codebase, so when a trailing comma is needed, you should see a warning in the Dart Analysis (IntelliJ) or Problems (VS Code) section of your IDE.

kenzieschmoll (approved):

PR #4181, opened 2022-06-08
changed textspan to inlinespan in codeview

IslinaS:
Changed TextSpan to InlineSpan in codeview.dart so that type-casting isn't necessary.


packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -418,6 +418,8 @@ class _CodeViewState extends State<CodeView>
     );
   }
 
+  InlineSpan span(InlineSpan span) => span;
```
elliette, line 421: This shouldn't be necessary since it's just converting the inline span into an inline span

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -912,7 +914,7 @@ class _LineItemState extends State<LineItem>
   TextSpan searchAwareLineContents() {
     // TODO(elliette): Switch to using TextSpans instead of InlineSpans so
```
elliette, line 915: We can remove this TODO as well :)

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -254,19 +254,19 @@ class _CodeViewState extends State<CodeView>
 
         // Look for [TextSpan]s which only contain '\n' to manually break the
```
elliette, line 255: Let's update the comment to: "Look for [InlineSpan]s..."

PR #4190, opened 2022-06-13
Copy Network Reponse

CoderDake:
Adding the ability to copy network responses

## Things to look at 
Advice on the spec is welcome! I pulled together the spec a little clumsily, so there may be some ways to do it better.

## Demo

https://user-images.githubusercontent.com/1386322/173415944-09e4fb33-affb-461b-8f3f-519eacf6d720.mp4

### DIFF
--- a/packages/devtools_app/lib/src/screens/network/network_request_inspector.dart
+++ b/packages/devtools_app/lib/src/screens/network/network_request_inspector.dart
@@ -37,10 +37,11 @@ class NetworkRequestInspector extends StatelessWidget {
 
   final NetworkController controller;
 
-  DevToolsTab _buildTab(String tabName) {
+  DevToolsTab _buildTab({required String tabName, Widget? trailing}) {
     return DevToolsTab.create(
       tabName: tabName,
       gaPrefix: 'requestInspectorTab',
+      trailing: trailing,
     );
   }
 
@@ -50,15 +51,25 @@ class NetworkRequestInspector extends StatelessWidget {
       valueListenable: controller.selectedRequest,
       builder: (context, data, _) {
         final tabs = <DevToolsTab>[
-          _buildTab(NetworkRequestInspector._overviewTabTitle),
+          _buildTab(tabName: NetworkRequestInspector._overviewTabTitle),
           if (data is DartIOHttpRequestData) ...[
-            _buildTab(NetworkRequestInspector._headersTabTitle),
+            _buildTab(tabName: NetworkRequestInspector._headersTabTitle),
             if (data.requestBody != null)
-              _buildTab(NetworkRequestInspector._requestTabTitle),
+              _buildTab(tabName: NetworkRequestInspector._requestTabTitle),
             if (data.responseBody != null)
-              _buildTab(NetworkRequestInspector._responseTabTitle),
+              _buildTab(
+                tabName: NetworkRequestInspector._responseTabTitle,
+                trailing: Row(
+                  mainAxisAlignment: MainAxisAlignment.end,
+                  children: [
+                    CopyToClipboardControl(
+                      dataProvider: () => data.responseBody,
+                    )
+                  ],
+                ),
+              ),
             if (data.hasCookies)
-              _buildTab(NetworkRequestInspector._cookiesTabTitle),
+              _buildTab(tabName: NetworkRequestInspector._cookiesTabTitle),
           ],
         ];
         return Card(
@@ -84,6 +95,7 @@ class NetworkRequestInspector extends StatelessWidget {
                         if (data.hasCookies) HttpRequestCookiesView(data),
                       ],
                     ],
+
                     gaScreen: analytics_constants.network,
                     // TODO(kenz): Consider using the outlined style
                     outlined: false,
--- a/packages/devtools_app/test/network/network_request_inspector_test.dart
+++ b/packages/devtools_app/test/network/network_request_inspector_test.dart
@@ -1,0 +1,67 @@
+import 'package:devtools_app/devtools_app.dart';
+import 'package:devtools_app/src/screens/network/network_request_inspector.dart';
+import 'package:devtools_test/devtools_test.dart';
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+import 'package:provider/provider.dart';
+import 'package:vm_service/vm_service.dart';
+
+import '../fixtures/riverpod_app/lib/tester.dart';
+import 'utils/network_test_utils.dart';
+
+void main() {
+  group('NetworkRequestInspector', () {
+    late NetworkController controller;
+    SocketProfile socketProfile;
+    HttpProfile httpProfile;
+    FakeServiceManager fakeServiceManager;
+
+    setUpAll(() {
+      setGlobal(IdeTheme, IdeTheme());
+      socketProfile = loadSocketProfile();
+      httpProfile = loadHttpProfile();
+      fakeServiceManager = FakeServiceManager(
+        service: FakeServiceManager.createFakeService(
+          socketProfile: socketProfile,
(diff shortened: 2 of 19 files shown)


packages/devtools_app/lib/src/screens/network/network_request_inspector.dart
```diff
@@ -84,6 +95,7 @@ class NetworkRequestInspector extends StatelessWidget {
                         if (data.hasCookies) HttpRequestCookiesView(data),
                       ],
                     ],
+
```
CoderDake, line 98:
```suggestion
```

packages/devtools_app/lib/src/screens/network/network_request_inspector.dart
```diff
@@ -50,15 +51,25 @@ class NetworkRequestInspector extends StatelessWidget {
...
-            _buildTab(NetworkRequestInspector._headersTabTitle),
+            _buildTab(tabName: NetworkRequestInspector._headersTabTitle),
             if (data.requestBody != null)
-              _buildTab(NetworkRequestInspector._requestTabTitle),
+              _buildTab(tabName: NetworkRequestInspector._requestTabTitle),
             if (data.responseBody != null)
-              _buildTab(NetworkRequestInspector._responseTabTitle),
+              _buildTab(
+                tabName: NetworkRequestInspector._responseTabTitle,
+                trailing: Row(
```
kenzieschmoll, line 62: You shouldn't need to put CopyToClipboardControl in a Row. It should already be aligned right by the DevToolsTab widget.

packages/devtools_app/test/network/network_request_inspector_test.dart
```diff
@@ -0,0 +1,106 @@
+import 'package:devtools_app/devtools_app.dart';
```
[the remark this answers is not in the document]
CoderDake: oops ty :)

packages/devtools_app/test/network/network_request_inspector_test.dart
```diff
@@ -0,0 +1,106 @@
...
+          case 'Clipboard.getData':
+            return Future.value(<String, dynamic>{});
+          case 'Clipboard.hasStrings':
+            return Future.value(<String, dynamic>{'value': true});
+          default:
+            break;
+        }
+
+        return Future.value(true);
+      });
```
kenzieschmoll, line 49: since this is copied code from another test, put this into a shared file (perhaps test_utils.dart) so that it can be reused

packages/devtools_app/test/network/network_request_inspector_test.dart
```diff
@@ -0,0 +1,106 @@
...
+      await tester.pumpWidget(buildNetworkRequestInspector(controller));
+      await controller.networkService.refreshNetworkData();
+      final networkRequest = requestsNotifier.value.requests[5];
+
+      controller.selectRequest(networkRequest);
+      await tester.pumpAndSettle();
+
+      await tester.tap(find.text('Response'));
+      await tester.pumpAndSettle();
+      await tester.tap(find.byType(CopyToClipboardControl));
```
kenzieschmoll, line 65: before tapping the copy control, verify that the _clipboardContents are empty

packages/devtools_app/lib/src/screens/network/network_request_inspector.dart
CoderDake: image
No dice unfortunately :(

packages/devtools_app/test/network/network_request_inspector_test.dart
```diff
@@ -0,0 +1,106 @@
...
+      await tester.pumpAndSettle(const Duration(seconds: 1));
+    });
+  });
+}
+
+Widget buildNetworkRequestInspector(NetworkController controller) {
+  final networkRequestInspector = NetworkRequestInspector(controller);
+  final debuggerController = createMockDebuggerControllerWithDefaults();
+
+  final localizations = Localizations(
```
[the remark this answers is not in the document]
CoderDake: this is a huge tip!
I only learned about wrapWithControllers after I had reverse engineered all of the other context. So thank you for pointing out that wrapWithControllers actually addresses the whole context issue!

packages/devtools_app/test/network/network_request_inspector_test.dart
```diff
@@ -0,0 +1,78 @@
...
+import '../test_utils/test_utils.dart';
+import 'utils/network_test_utils.dart';
+
+void main() {
+  group('NetworkRequestInspector', () {
+    late NetworkController controller;
+    late SocketProfile socketProfile;
+    late HttpProfile httpProfile;
+    late FakeServiceManager fakeServiceManager;
+    late String _clipboardContents;
```
bkonyi, line 25: Nit: if we're always setting this to empty string in `setUp` anyway, we may as well initialize it here too and remove the `late`. It's not super important in this context, but using `late` does introduce additional initialization checks at each access, so if there's a simple way around using `late` you should try and avoid it.

packages/devtools_app/test/network/network_request_inspector_test.dart
```diff
@@ -0,0 +1,78 @@
...
+      await controller.startRecording();
+      await tester.pumpWidget(
+        wrapWithControllers(
+          NetworkRequestInspector(controller),
+          debugger: createMockDebuggerControllerWithDefaults(),
+        ),
+      );
+
+      await controller.networkService.refreshNetworkData();
+      final networkRequest = requestsNotifier.value.requests[5];
```
bkonyi, line 59: What's special about `requests[5]` here? Maybe leave a comment describing the contents.

packages/devtools_app/test/network/network_request_inspector_test.dart
```diff
@@ -0,0 +1,78 @@
...
+        ),
+      );
+      setGlobal(ServiceConnectionManager, fakeServiceManager);
+      controller = NetworkController();
+      setupClipboardCopyListener(clipboardContentsCallback: (contents) {
+        _clipboardContents = contents ?? '';
+      });
+    });
+
+    testWidgets('copy response body', (tester) async {
```
bkonyi, line 47: General nit: it's typically hard to get a good idea as to what's being done in these sorts of `flutter_driver` tests. Some high level comments describing what actions are being taken and what the test is supposed to be testing go a long way in improving readability and maintainability when someone has to figure out why the test is failing in the future.

bkonyi (approved):
LGTM with nits.

packages/devtools_app/test/network/network_request_inspector_test.dart
```diff
@@ -0,0 +1,80 @@
...
+      await tester.tap(find.byType(CopyToClipboardControl));
+
+      expect(_clipboardContents, isNotEmpty);
+      expect(
+        _clipboardContents,
+        equals((networkRequest as DartIOHttpRequestData).responseBody),
+      );
+
+      controller.stopRecording();
+      await tester.pumpAndSettle(const Duration(seconds: 1));
```
kenzieschmoll, line 77: why is this pumpAndSettle necessary?

kenzieschmoll (approved):
one comment then lgtm

packages/devtools_app/test/network/network_request_inspector_test.dart
CoderDake: That one is necessary for clearing out the timers. Without it then the network timers will remain after the spec is done

kenzieschmoll: This is surprising. controller.stopRecording() is synchronous, so I'm confused why we need to pump and settle.

bartcone-fulfilld:
Saw this in the release notes and yelled out loud. Thanks so much!

PR #4192, opened 2022-06-13
Add layer outlines to the raster metrics tab in the performance view

kenzieschmoll:
Without this outline, it is difficult to see where a layer begins and ends in the context of the entire screen image.
Screen Shot 2022-06-13 at 11 06 40 AM
@iskakaushik


iskakaushik (approved):
LGTM

PR #4195, opened 2022-06-17
Object inspector

migueltylee:
Implemented the object inspector feature under the VM tools tab, which will help developers obtain information related to the objects created in the Dart VM during program execution.
The object inspector contains a program explorer where the user can select different objects such as libraries, scripts, and classes; and displays the corresponding VM information on a HistoryViewport whose history is managed by an ObjRef HistoryManager.
This PR also includes the basic layout and VM data for the 'Class' objects on the viewport.


packages/devtools_app/.gitignore
```diff
@@ -8,7 +8,6 @@ doc/api/
 .flutter-plugins-dependencies
 android/
 ios/
-lib/generated_plugin_registrant.dart
```
kenzieschmoll, line 11: add this line back and remove the changes to lib/generated_plugin_restraint below

packages/devtools_app/lib/src/screens/vm_developer/vm_class_screen.dart
```diff
@@ -0,0 +1,122 @@
+import 'package:flutter/material.dart';
```
kenzieschmoll, line 1: add 2022 copyright

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
```diff
@@ -0,0 +1,123 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: nit 2022

packages/devtools_app/lib/src/screens/vm_developer/vm_class_screen_controller.dart
```diff
@@ -0,0 +1,58 @@
+import 'package:flutter/foundation.dart';
```
kenzieschmoll, line 1: here as well

packages/devtools_app/pubspec.yaml
```diff
@@ -21,6 +21,7 @@ dependencies:
   async: ^2.0.0
   codicon: ^1.0.0
   collection: ^1.15.0
+  dcdg: ^4.0.1
```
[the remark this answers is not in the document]
bkonyi: Nope, this is a UML generator package he was playing around with, so it can be removed.

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
```diff
@@ -0,0 +1,58 @@
...
+import 'package:flutter/material.dart';
+
+import '../../shared/split.dart';
+import '../debugger/program_explorer.dart';
+import '../debugger/program_explorer_controller.dart';
+import '../debugger/program_explorer_model.dart';
+import 'object_viewport.dart';
+import 'vm_developer_tools_screen.dart';
+
+/// Displays a program explorer used to select and display information about objects in the Dart VM
```
bkonyi, line 14: Nit: this should probably be a two line comment (I'm pretty sure it's > 80 characters long).

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
```diff
@@ -0,0 +1,58 @@
...
+      : super(
+          id,
+          title: 'Objects',
+          icon: Icons.data_object,
+        );
+  static const id = 'inspector-view';
+
+  final programExplorerController = ProgramExplorerController();
+
+  ObjectHistory objectHistory = ObjectHistory();
```
bkonyi, line 26: Nit: if we don't reassign to a property, we should make it final and drop the left hand side type.

```dart
final objectHistory = ObjectHistory();
```

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
```diff
@@ -0,0 +1,58 @@
...
+          objectHistory: objectHistory,
+        )
+      ],
+    );
+  }
+
+  void _onNodeSelected(VMServiceObjectNode node) {
+    final object = node.object;
+    if (object != null) objectHistory.pushEntry(object);
+    return;
```
bkonyi, line 56: Unnecessary return.

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
```diff
@@ -0,0 +1,58 @@
...
+          controller: programExplorerController,
+          objectHistory: objectHistory,
+        )
+      ],
+    );
+  }
+
+  void _onNodeSelected(VMServiceObjectNode node) {
+    final object = node.object;
+    if (object != null) objectHistory.pushEntry(object);
```
bkonyi, line 55: Nit: I try to avoid using single line statements like this since it makes it easier for bugs to sneak in later. Maybe change it to this:

```dart
if (object != null) {
  objectHistory.pushEntry(object);
}
```

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
```diff
@@ -0,0 +1,123 @@
...
+import '../../primitives/history_manager.dart';
+import '../../shared/common_widgets.dart';
+import '../../shared/history_viewport.dart';
+import '../../shared/utils.dart';
+import '../debugger/program_explorer_controller.dart';
+import 'vm_class_screen.dart';
+import 'vm_class_screen_controller.dart';
+import 'vm_developer_common_widgets.dart';
+
+/// Displays the VM information for the currently selected object in the program explorer.
```
bkonyi, line 19: Same comment as before, this should be on two lines.

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
```diff
@@ -0,0 +1,123 @@
...
+class ObjectViewport extends StatefulWidget {
```
bkonyi, line 20: I don't think this needs to be a `StatefulWidget` since we don't actually store any state in `_ObjectViewportState`.

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
```diff
@@ -0,0 +1,123 @@
...
+    return ValueListenableBuilder(
+      valueListenable: widget._refreshing,
+      builder: (context, refreshing, _) {
+        return HistoryViewport(
+          history: widget.objectHistory,
+          controls: [
+            ToolbarAction(
+              icon: Icons.refresh,
+              onPressed: () =>
+                  {widget._refreshing.value = !widget._refreshing.value},
```
bkonyi, line 53: While this does work, I don't think this is doing exactly what you think it is. If you're using `=>` notation, the right hand side expression is treated as a return value. In this case, the return value is a `Set` containing a single boolean value. You can drop the braces here.

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
```diff
@@ -0,0 +1,123 @@
...
+  _ObjectViewportState createState() => _ObjectViewportState();
+}
+
+class _ObjectViewportState extends State<ObjectViewport> with AutoDisposeMixin {
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder(
+      valueListenable: widget._refreshing,
+      builder: (context, refreshing, _) {
+        return HistoryViewport(
```
bkonyi, line 47: If you do `HistoryViewport<ObjRef>(...)` instead, you won't need to cast as `ObjRef` below.

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
```diff
@@ -0,0 +1,123 @@
...
+  bool get hasObjects => _openedObjects.isNotEmpty;
+
+  void pushEntry(ObjRef ref) async {
+    if (ref == current.value) return;
+
+    while (hasNext) {
+      pop();
+    }
+
+    //_openedObjects.remove(ref);
```
bkonyi, line 116: Is this line still needed?

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
```diff
@@ -0,0 +1,123 @@
...
+  if (objRef is ScriptRef) return 'Script @ ' + (objRef.uri ?? '<name>');
+  return '<unrecognized object>';
+}
+
+//Calls the object VM statistics card builder according to the VM Object type.
+Widget _buildObjectScreen(ObjRef objRef) {
+  if (objRef is ClassRef) {
+    final classController = ClassScreenController(objRef);
+    return VmClassScreen(
+      controller: classController,
```
bkonyi, line 89: Since it's not used anywhere else, you may as well just create the `ClassScreenController` inline:

```dart
if (objRef is ClassRef) {
  return VmClassScreen(
     controller: ClassScreenController(objRef),
  );
}
```

packages/devtools_app/lib/src/screens/vm_developer/vm_class_screen.dart
```diff
@@ -0,0 +1,122 @@
...
+  });
+
+  final ClassScreenController controller;
+
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder(
+      valueListenable: controller.refreshing,
+      builder: (context, refreshing, _) {
+        return Flexible(
```
bkonyi, line 21: @kenzieschmoll is `Flexible` the right thing here or should it be `Expanded`?

packages/devtools_app/lib/src/screens/vm_developer/vm_class_screen.dart
```diff
@@ -0,0 +1,122 @@
...
+          child: Row(
+            children: [
+              Flexible(
+                child: Column(
+                  children: [
+                    Flexible(
+                      flex: 10,
```
bkonyi, line 28: Why did we choose these flex values?

packages/devtools_app/lib/src/screens/vm_developer/vm_class_screen.dart
```diff
@@ -0,0 +1,122 @@
...
+            kbFractionDigits: 3,
+          ),
+        ),
+        const MapEntry('Reachable Size', 'TO-DO'),
+        const MapEntry('Retained Size', 'TO-DO'),
+        const MapEntry('Retaining path', 'TO-DO'),
+        const MapEntry('Inbound references', 'TO-DO'),
+        MapEntry(
+          'Library',
+          clazz?.library?.name == ''
```
bkonyi, line 78: Nit: instead of comparing against the empty string, you should do something like this:

```dart
clazz?.library?.name?.isEmpty ?? false
```

packages/devtools_app/lib/src/screens/vm_developer/vm_class_screen_controller.dart
```diff
@@ -0,0 +1,58 @@
...
+
+import '../../primitives/auto_dispose.dart';
+import '../../service/vm_service_wrapper.dart';
+import '../../shared/globals.dart';
+import '../debugger/debugger_model.dart';
+
+/// Auto DisposableController for fetching and displaying data on the class screen.
+class ClassScreenController extends DisposableController
+    with AutoDisposeControllerMixin {
+  ClassScreenController(ObjRef? objRef) {
```
bkonyi, line 12: I'd change the type here to `ClassRef` so that the cast has to happen outside of this object as this object will never be able to handle anything other than `ClassRef` inputs.

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_tools_screen.dart
```diff
@@ -60,15 +61,22 @@ class VMDeveloperToolsScreen extends Screen {
 }
 
 class VMDeveloperToolsScreenBody extends StatefulWidget {
-  const VMDeveloperToolsScreenBody();
+  const VMDeveloperToolsScreenBody({
+    this.displayObjectInspector = false,
```
bkonyi, line 65: Instead of making this a parameter, I'd just add a constant at the top of the file:

```dart
/// Set to true to enable the Objects tab.
const displayObjectInspector = false;
```

Then you should be able to make everything `const` further down:

```dart
static const List<VMDeveloperView> views = [
  VMStatisticsView(),
  IsolateStatisticsView(),
  if (displayObjectInspector) ObjectInspectorView(),
];
```

bkonyi (commented):
I've done an initial review pass. No major concerns, so we're off to a good start!

Also, would you mind adding some screenshots to the description? It will help reviewers get an idea of what UI changes are in the PR.

packages/devtools_app/lib/src/screens/vm_developer/vm_class_screen.dart
migueltylee: Because the upper VMInfoCard was overflowing with the default flex values, will make it scrollable soon.

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
```diff
@@ -0,0 +1,61 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final programExplorerKey = GlobalKey(debugLabel: 'programExplorerKey');
+    programExplorerController.initialize();
+    return Split(
+      axis: Axis.horizontal,
+      initialFractions: const [0.20, 0.80],
+      children: [
+        ProgramExplorer(
+          key: programExplorerKey,
```
kenzieschmoll, line 41: is this key necessary?

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
```diff
@@ -0,0 +1,118 @@
...
+
+  final ProgramExplorerController controller;
+  final ObjRef? initialObject;
+  final ObjectHistory objectHistory;
+  ValueListenable<bool> get refreshing => _refreshing;
+  final _refreshing = ValueNotifier<bool>(true);
+
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder(
```
kenzieschmoll, line 40: I don't think we need to use a ValueListenableBuilder here. We don't appear to be using the value of `refreshing` anywhere.

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
```diff
@@ -0,0 +1,118 @@
...
+      valueListenable: _refreshing,
+      builder: (context, refreshing, _) {
+        return HistoryViewport<ObjRef>(
+          history: objectHistory,
+          controls: [
+            ToolbarAction(
+              icon: Icons.refresh,
+              onPressed: () {
+                _refreshing.value = !_refreshing.value;
```
kenzieschmoll, line 49: This doesn't really appear to be doing anything at the moment. If we need to add a TODO for this PR, that is ok.
```
// TODO(mtaylee): hook up refresh functionality
onPressed: () {},
```

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
```diff
@@ -0,0 +1,118 @@
...
+    return const VMInfoCard(title: 'TO-DO: Display Field object data');
+  if (objRef is LibraryRef)
+    return const VMInfoCard(title: 'TO-DO: Display Library object data');
+  if (objRef is ScriptRef)
+    return const VMInfoCard(title: 'TO-DO: Display Script object data');
+  return const SizedBox.shrink();
+}
+
+/// Manages the history of selected ObjRefs to make them accessible on a HistoryViewport.
+class ObjectHistory extends HistoryManager<ObjRef> {
```
kenzieschmoll, line 100: I don't see where `hasObjects` and `openedObjects` are used in this PR. Can we remove them until we have a function for these two getters?

packages/devtools_app/lib/src/screens/vm_developer/vm_class_screen.dart
```diff
@@ -0,0 +1,126 @@
...
+import 'vm_class_screen_controller.dart';
+import 'vm_developer_common_widgets.dart';
+
+/// A screen on the object inspector historyViewport displaying information related to class objects in the Dart VM.
+class VmClassScreen extends StatelessWidget {
+  const VmClassScreen({
+    required this.controller,
+  });
+
+  final ClassScreenController controller;
```
kenzieschmoll, line 18: nit: rename to VmClassDisplayController

packages/devtools_app/lib/src/screens/vm_developer/vm_class_screen_controller.dart
```diff
@@ -0,0 +1,62 @@
...
+  InstanceSet? _instances;
+
+  String? get scriptUri => _scriptUri;
+  String? _scriptUri;
+
+  SourcePosition? get pos => _pos;
+  SourcePosition? _pos;
+
+  ValueListenable<bool> get refreshing => _refreshing;
+  final _refreshing = ValueNotifier<bool>(false);
```
kenzieschmoll, line 39: I don't think we are using this in the right way. Are you trying to use this ValueNotifier as a way to force the UI to rebuild when `refresh` is clicked? AFAICT we never actually do anything with the value of this notifier.

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_tools_screen.dart
```diff
@@ -62,13 +65,16 @@ class VMDeveloperToolsScreen extends Screen {
...
+    const VMStatisticsView(),
+    const IsolateStatisticsView(),
+    if (displayObjectInspector) ObjectInspectorView(),
   ];
 
   @override
-  _VMDeveloperToolsScreenState createState() => _VMDeveloperToolsScreenState();
+  _VMDeveloperToolsScreenState createState() {
+    return _VMDeveloperToolsScreenState();
+  }
```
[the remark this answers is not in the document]
bkonyi: I think Miguel was originally inserting `ObjectInspectorView()` into `views` here if `displayObjectInspector == true`. Probably just an artifact of that change.

packages/devtools_app/lib/src/screens/vm_developer/vm_class_screen_controller.dart
bkonyi: I do think we should keep this listenable, but we should be using it to determine whether or not we display a spinner. We could update `refresh()` below to do something like this:

```dart
Future<void> refresh() async {
  _refreshing.value = true;
  final isolateRef = ...
  // Refresh logic here.
  _refreshing.value = false;
}
```

And then use the `ValueListenableBuilder<bool>` to display the spinner when `_refreshing.value = true`.

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
```diff
@@ -84,9 +81,8 @@ String _getViewportTitle(ObjRef objRef) {
 //Calls the object VM statistics card builder according to the VM Object type.
 Widget _buildObjectScreen(ObjRef objRef) {
   if (objRef is ClassRef) {
-    final classController = ClassScreenController(objRef);
     return VmClassScreen(
-      controller: classController,
+      controller: ClassScreenController(objRef),
```
bkonyi, line 85: I'm wondering if it'd just be better to pass `objRef` directly into `VmClassScreen` and let it handle creating the controller internally since there's currently no need for the controller to be accessible outside of the screen.

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
```diff
@@ -0,0 +1,118 @@
...
+  final _openedObjects = <ObjRef>{};
+
+  bool get hasObjects => _openedObjects.isNotEmpty;
+
+  void pushEntry(ObjRef ref) async {
+    if (ref == current.value) return;
+
+    while (hasNext) {
+      pop();
+    }
```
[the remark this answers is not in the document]
migueltylee: When you go back through the object history, there will be objects ahead. If you then select another object, those objects are removed to have the newly selected object right after the current object.

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
```diff
@@ -0,0 +1,118 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder(
+      valueListenable: _refreshing,
+      builder: (context, refreshing, _) {
+        return HistoryViewport<ObjRef>(
+          history: objectHistory,
+          controls: [
+            ToolbarAction(
```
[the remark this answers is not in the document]
migueltylee: I had initially used it, but I thought the style was not appropiate. I think we want these to be only icons, as in the ones in the debugger viewport. I'll also be adding another icon button to show a popup with recently selected objects.

packages/devtools_app/lib/src/screens/vm_developer/vm_object_model.dart
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  Future<void> initialize();
+}
+
+class ClassObject extends VmObject {
+  ClassObject({required super.ref});
+
+  Class? get obj => _clazz;
+  Class? _clazz;
```
bkonyi, line 40: Can we make this non-nullable and late?

packages/devtools_app/lib/src/screens/vm_developer/vm_class_display.dart
```diff
@@ -0,0 +1,116 @@
...
+class VmClassDisplay extends StatelessWidget {
+  const VmClassDisplay({
+    required this.clazz,
+  });
+  final ClassObject clazz;
+
+  @override
+  Widget build(BuildContext context) {
+    return Column(
+      //mainAxisSize: MainAxisSize.min,
```
bkonyi, line 23: Old code?

packages/devtools_app/lib/src/screens/vm_developer/vm_class_display.dart
```diff
@@ -0,0 +1,116 @@
...
+        const MapEntry('Inbound references', 'TO-DO'),
+        MapEntry(
+          'Library',
+          clazz.obj?.library?.name?.isEmpty ?? false
+              ? clazz.script?.uri
+              : clazz.obj?.library?.name,
+        ),
+        MapEntry(
+          'Script',
+          (_fileName(clazz.script?.uri) ?? '') +
```
bkonyi, line 76: This can be done with string comprehension. If it's too cumbersome to do inline, just declare a local variable before the return above.

bkonyi (commented):
Almost there!

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
```diff
@@ -0,0 +1,60 @@
...
```
kenzieschmoll, line 42: @bkonyi I don't think we need a global key here. Is there a reason the key param is required from when you originally wrote the ProgramExplorer code?

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
```diff
@@ -0,0 +1,118 @@
...
+            return Expanded(child: child);
+          },
+        );
+      },
+    );
+  }
+}
+
+@visibleForTesting
+String getViewportTitle(VmObject? object) {
```
kenzieschmoll, line 67: nit: avoid starting method names with get: https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

Replace with _viewportTitle or _generateViewportTitle, or similar

packages/devtools_app/lib/src/screens/vm_developer/vm_class_display.dart
```diff
@@ -0,0 +1,116 @@
...
+            kbFractionDigits: 3,
+          ),
+        ),
+        const MapEntry('Reachable Size', 'TO-DO'),
+        const MapEntry('Retained Size', 'TO-DO'),
+        const MapEntry('Retaining path', 'TO-DO'),
+        const MapEntry('Inbound references', 'TO-DO'),
+        MapEntry(
+          'Library',
+          clazz.obj?.library?.name?.isEmpty ?? false
```
kenzieschmoll, line 70: can this just be `clazz.obj?.library.name.isEmpty ?? false`?

I'm surprised this isn't triggering an analyzer warning since this code is null safe

packages/devtools_app/lib/src/screens/vm_developer/vm_class_display.dart
```diff
@@ -0,0 +1,116 @@
...
+              ? clazz.script?.uri
+              : clazz.obj?.library?.name,
```
kenzieschmoll, line 72: same comment here

packages/devtools_app/lib/src/screens/vm_developer/vm_object_model.dart
```diff
@@ -0,0 +1,149 @@
...
+
+  Obj? _obj;
+
+  Future<void> fetchObject() async {
+    final isolateRef = serviceManager.isolateManager.selectedIsolate.value!;
+    _isolate = await _service.getIsolate(isolateRef.id!);
+    _obj = await _service.getObject(_isolate!.id!, ref.id!);
+  }
+
+  Future<void> initialize();
```
kenzieschmoll, line 33: if fetchObject should always be called as part of initialize, we can mark this method with @mustCallSuper instead of making it abstract

```
@mustCallSuper
Future<void> initialize() async {
  await fetchObject();
}
```

then overrides of this method don't have to all call fetchObject, which will let you eliminate several of the overrides below to `initialize`

packages/devtools_app/lib/src/screens/vm_developer/vm_object_model.dart
```diff
@@ -0,0 +1,149 @@
...
+      final token = _clazz!.location!.tokenPos!;
+      _script = await _service.getObject(
+        _isolate!.id!,
+        _clazz!.location!.script!.id!,
+      ) as Script;
+      _pos = SourcePosition.calculatePosition(
+        _script!,
+        token,
+      );
+    }
```
kenzieschmoll, line 70: as we implement the object inspector for other Vm object types, we should see if any of this can be generalized if it applies to other data types as well.

packages/devtools_app/lib/src/screens/vm_developer/vm_object_model.dart
```diff
@@ -0,0 +1,149 @@
...
+
+    _instances = await _service.getInstances(_isolate!.id!, ref.id!, 100);
+  }
+}
+
```
kenzieschmoll, line 75: nit: add a TODO here to indicate that all of the data classes below are incomplete.

kenzieschmoll (commented):
Some more comments, but overall looking good! Great work implementing the refactor changes we discussed offline.

After addressing these comments, the one thing left would be to add some tests :)

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
```diff
@@ -21,31 +21,44 @@ class ObjectInspectorView extends VMDeveloperView {
...
+class _ObjectInspectorView extends StatefulWidget {
   final controller = ObjectInspectorViewController();
 
+  final programExplorerController = ProgramExplorerController();
+
   @override
-  bool get showIsolateSelector => true;
+  State<StatefulWidget> createState() => _ObjectInspectorViewState();
 
+  State<StatefulWidget> initState() {
```
kenzieschmoll, line 41: initState is a lifecycle method on the State class (and it should return void). Move to _ObjectInspectorViewState below.

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
```diff
@@ -21,31 +21,44 @@ class ObjectInspectorView extends VMDeveloperView {
...
-  final programExplorerController = ProgramExplorerController();
+  @override
+  bool get showIsolateSelector => true;
 
+  @override
+  Widget build(BuildContext context) => _ObjectInspectorView();
+}
+
+class _ObjectInspectorView extends StatefulWidget {
   final controller = ObjectInspectorViewController();
```
kenzieschmoll, line 34: this controller should be initialized below in the State class's initState method.

bkonyi (commented):
LGTM once tests are done and @kenzieschmoll is happy with the state of things (I'll let her give explicit approval).

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
```diff
@@ -0,0 +1,77 @@
...
+  late final ObjectInspectorViewController controller;
+
+  late final ProgramExplorerController programExplorerController;
+
+  @override
+  void initState() {
+    super.initState();
+    controller = ObjectInspectorViewController();
+    programExplorerController = ProgramExplorerController();
+    programExplorerController.initialize();
```
kenzieschmoll, line 48: nit: this can be simplified to

```
programExplorerController = ProgramExplorerController()..initialize();
```

packages/devtools_app/test/shared/history_manager_test.dart
```diff
@@ -81,5 +81,57 @@ void main() {
...
+    test('replaceCurrent at the top of the stack', () {
+      history.push(ref1);
+      history.push(ref2);
+      history.push(ref3);
+      history.replaceCurrent(ref1);
+      expect(history.current.value, ref1);
+      expect(history.hasNext, false);
+      expect(history.hasPrevious, true);
+      history.moveBack();
+      expect(history.current.value, ref2);
```
kenzieschmoll, line 102: move back and check values all the way until `history.hasPrevious` is false

packages/devtools_app/test/shared/history_manager_test.dart
```diff
@@ -81,5 +81,57 @@ void main() {
...
+      history.replaceCurrent(ref1);
+      expect(history.current.value, ref1);
+      expect(history.hasNext, false);
+      expect(history.hasPrevious, false);
+    });
+
+    test('replaceCurrent at the top of the stack', () {
+      history.push(ref1);
+      history.push(ref2);
+      history.push(ref3);
```
kenzieschmoll, line 96: after this line, verify that the current value is ref3. That way, we know for sure that the replaceCurrent call changed the value

packages/devtools_app/test/shared/history_manager_test.dart
```diff
@@ -81,5 +81,57 @@ void main() {
...
+      history.moveBack();
+      expect(history.current.value, ref2);
+    });
+
+    test('replaceCurrent in the middle of the stack', () {
+      history.push(ref1);
+      history.push(ref2);
+      history.push(ref3);
+      history.moveBack();
+      history.replaceCurrent(ref3);
```
kenzieschmoll, line 110: same thing here. Verify the current value before calling replaceCurrent

packages/devtools_app/test/shared/history_manager_test.dart
```diff
@@ -81,5 +81,57 @@ void main() {
...
+      history.push(ref2);
+      history.push(ref3);
+      history.moveBack();
+      history.moveBack();
+      history.replaceCurrent(ref3);
+      expect(history.current.value, ref3);
+      expect(history.hasNext, true);
+      expect(history.hasPrevious, false);
+      history.moveForward();
+      expect(history.current.value, ref2);
```
kenzieschmoll, line 134: move all the way forward and check the last item in the stack too

packages/devtools_app/test/vm_developer/object_inspector/object_inspector_view_test.dart
```diff
@@ -0,0 +1,33 @@
...
+import 'package:mockito/mockito.dart';
+
+void main() {
+  final objectInspector = ObjectInspectorView();
+
+  final fakeServiceManager = FakeServiceManager();
+  when(fakeServiceManager.connectedApp!.isProfileBuildNow).thenReturn(false);
+  when(fakeServiceManager.connectedApp!.isDartWebAppNow).thenReturn(false);
+  setGlobal(ServiceConnectionManager, fakeServiceManager);
+  setGlobal(IdeTheme, IdeTheme());
```
kenzieschmoll, line 21: put all this into a `setUp` method instead. You'll need to define `objectInspector` and `fakeServiceManager` as late outside of the `setUp` method so that they can be accessed everywhere, but they should be initialized in `setUp`

packages/devtools_app/test/vm_developer/object_inspector/object_inspector_view_test.dart
```diff
@@ -0,0 +1,33 @@
...
+
+  testWidgets('builds screen', (WidgetTester tester) async {
+    await tester.pumpWidget(wrap(Builder(builder: objectInspector.build)));
+    expect(find.bySubtype<Split>(), findsNWidgets(2));
+    expect(find.bySubtype<ProgramExplorer>(), findsOneWidget);
+    expect(find.bySubtype<ObjectViewport>(), findsOneWidget);
```
kenzieschmoll, line 27: find by type instead of subtype

packages/devtools_app/test/vm_developer/object_inspector/object_viewport_test.dart
```diff
@@ -0,0 +1,276 @@
...
+  late TestObjectInspectorViewController testObjectInspectorViewController;
+
+  final mockClassObject = MockClassObject();
+
+  when(mockClassObject.name).thenReturn('FooClass');
+  when(mockClassObject.ref).thenReturn(fakeClassRef);
+  when(mockClassObject.obj).thenReturn(fakeClass);
+  when(mockClassObject.script).thenReturn(null);
+  when(mockClassObject.instances).thenReturn(null);
+  when(mockClassObject.pos).thenReturn(null);
```
kenzieschmoll, line 27: do this inside of setUp

packages/devtools_app/test/vm_developer/object_inspector/object_viewport_test.dart
```diff
@@ -0,0 +1,276 @@
...
+        .setCurrentObject(testFuncObject);
+    await tester.pumpWidget(
+      wrap(ObjectViewport(controller: testObjectInspectorViewController)),
+    );
+    expect(viewportTitle(testFuncObject), 'Function FooFunction');
+    expect(find.text('Function FooFunction'), findsOneWidget);
+    expect(find.bySubtype<VMInfoCard>(), findsOneWidget);
+  });
+
+  group('ObjectHistory', () {
```
kenzieschmoll, line 125: the only tests you need in this group are the ones that test the method you added to ObjectHistory (`pushEntry`). Some of these tests are duplicated and already tested in history_manager_test.dart, so no need to copy them here.

packages/devtools_app/test/vm_developer/object_inspector/vm_class_display_test.dart
```diff
@@ -0,0 +1,48 @@
...
+  final mockClassObject = MockClassObject();
+
+  fakeClass.size = 1024;
+
+  when(mockClassObject.name).thenReturn('FooClass');
+  when(mockClassObject.ref).thenReturn(fakeClassRef);
+  when(mockClassObject.obj).thenReturn(fakeClass);
+  when(mockClassObject.script).thenReturn(fakeScript);
+  when(mockClassObject.instances).thenReturn(fakeInstances);
+  when(mockClassObject.pos).thenReturn(fakePos);
```
kenzieschmoll, line 24: initialize mockClassObject and perform all this mocking setup inside the `setUp` method.

packages/devtools_app/test/vm_developer/vm_developer_test_utils.dart
```diff
@@ -0,0 +1,244 @@
...
+import 'package:devtools_app/src/screens/vm_developer/object_inspector_view_controller.dart';
+import 'package:devtools_app/src/screens/vm_developer/object_viewport.dart';
+import 'package:devtools_app/src/screens/vm_developer/vm_object_model.dart';
+import 'package:flutter/foundation.dart';
+import 'package:vm_service/vm_service.dart';
+
+final fakeLibRef =
+    TestLibraryRef(name: 'fooLib', uri: 'fooLib.dart', id: '1234');
+
+final fakeLib = Library(
```
kenzieschmoll, line 11: for any var that isn't technically a "fake" (i.e. any that are an instance of a real class from the vm service), do not include "fake" in the name. We can call these 'testLib', 'testClass' etc.

packages/devtools_app/test/vm_developer/vm_developer_test_utils.dart
```diff
@@ -0,0 +1,244 @@
...
+  @override
+  String? get name => 'FooInstance';
+}
+
+class TestInstanceRef extends InstanceRef {
+  TestInstanceRef(
+      {required super.kind,
+      required super.identityHashCode,
+      required super.classRef,
+      required super.id});
```
kenzieschmoll, line 240: add a trailing comma

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
migueltylee: Thanks for the tip

packages/devtools_app/test/vm_developer/object_inspector/object_viewport_test.dart
migueltylee: Done.
I did notice when testing pop, that it gives an error when popping an item when it is the only item in the stack. Because current does not go back to null, but instead tries to access element at index -1.

kenzieschmoll: In that case, we should add the pop test to history_manager_test.dart and fix that case so that we correctly set to null👍

migueltylee: Should that I fix that in a separate PR or on this one?

kenzieschmoll: Whichever you prefer. Normally I'd say separate PR, but I don't remind reviewing here to speed it along and avoid merge conflicts for you.

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
migueltylee: Good catch! Thank you

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
migueltylee: I think it was only there for testing the debugger... but not necessary anymore.

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
```diff
@@ -0,0 +1,118 @@
...
+                : _buildObjectScreen(currentObjRef);
+          },
+        );
+      },
+    );
+  }
+}
+
+String _getViewportTitle(ObjRef objRef) {
+  if (objRef is ClassRef) return objRef.type + ' ' + (objRef.name ?? '<name>');
```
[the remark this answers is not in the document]
migueltylee: Thanks for the tip!

packages/devtools_app/lib/src/screens/vm_developer/vm_object_model.dart
migueltylee: Done!

packages/devtools_app/lib/src/primitives/history_manager.dart
```diff
@@ -82,4 +82,9 @@ class HistoryManager<T> {
     _historyIndex = _history.length - 1;
     _current.value = _history[_historyIndex];
   }
+
+  /// Replaces the [current] item with a provided value.
+  void replaceCurrent(T value) {
+    _history[_historyIndex] = value;
```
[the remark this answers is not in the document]
migueltylee: Good catch!

packages/devtools_app/lib/src/screens/vm_developer/object_viewport.dart
migueltylee: Thanks for sharing!

packages/devtools_app/lib/src/screens/vm_developer/vm_class_display.dart
migueltylee: The obj is now non-nullable so I removed the '?'. The library and name fields are nullable so I need to access them with the '?' operator.

packages/devtools_app/test/vm_developer/vm_developer_test_utils.dart
```diff
@@ -0,0 +1,153 @@
+import 'package:devtools_app/src/screens/debugger/debugger_model.dart';
```
kenzieschmoll, line 1: add copyright

kenzieschmoll (approved):
one last comment then lgtm!

PR #4198, opened 2022-06-17
Some renamings in preferences.

polina-c:
Rename field preferences.inspectorPreferences to preferences.inspector for better conciseness. 
Rename toggle.. to set.. because toggle should not take parameter.
Rename 'value' to hoverEvalModeEnabledValue, because other values will be here.

### DIFF
--- a/packages/devtools_app/lib/src/shared/preferences.dart
+++ b/packages/devtools_app/lib/src/shared/preferences.dart
@@ -69,11 +69,11 @@ class InspectorPreferencesController {
   static const _hoverEvalModeStorageId = 'inspector.hoverEvalMode';
 
   Future<void> init() async {
-    String? value = await storage.getValue(_hoverEvalModeStorageId);
+    String? hoverEvalMode = await storage.getValue(_hoverEvalModeStorageId);
 
     // When embedded, default hoverEvalMode to off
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -87,7 +87,7 @@ class InspectorPreferencesController {
   }
 
   /// Change the value for the hover eval mode setting.
-  void toggleHoverEvalMode(bool enableHoverEvalMode) {
+  void setHoverEvalMode(bool enableHoverEvalMode) {
```
kenzieschmoll, line 90: let's change this back to `toggleHoverEvalMode`. Avoid starting method names with get and set.

kenzieschmoll (approved):
lgtm with one comment

packages/devtools_app/lib/src/shared/preferences.dart
polina-c: It is just about get, not set, both for getters and methods: https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

I would keep 'set' here. Toggle is confusing, because toggle is expected to read value, switch it and save.

Issue #4188, opened 2022-06-10
Persist memory screen settings.


polina-c:
Details:

Polina Cherkasova, 22 min
Is it expected that DevTools settings are reset every time I run the app from command line?

Kenzie Davisson, 16 min
Which settings are you referring to?

Kenzie Davisson, 10 min
If you are referring to the settings we store in preferences (theme, dense mode, vm developer mode), then these will persist in the following way:
1. When running on Desktop, these will always persist
2. When running on Web with the devtools server running, these will always persist
3. When running on Web without the devtools server running, these will not persist between runs, as they are stored in the Browser storage.

when you run flutter run -d target the devtools server is not running. If you want to run with the DevTools server, use the build_e2e.dart script under tools/

Polina Cherkasova, 7 min, Edited
i am running on macos and i always have to enable advanced settings for memory page to see garbage collection button

Kenzie Davisson, 4 min
those settings are not persisted in our preferences file. You can add a MemoryPreferencesController class to the preferences file to do this. See https://github.com/flutter/devtools/pull/4090/files for an example where Dan just recently did this to persist Inspector settings.

in general, we'll probably want to be careful that we don't persist all settings from all pages, as some settings may be related to the current data, and not to the tool (for example, certain filters, etc.)

polina-c:
inspector settings test: https://github.com/flutter/devtools/pull/4090/files#diff-f83bc9b9bce356789e0fbc423db7344bbd733560cf8acc2a99a483adafc8e906

PR #4090, opened 2022-05-12
Hover less

CoderDake:
Fixes https://github.com/flutter/devtools/issues/3921

This PR adds:
- The ability for the Flutter Inspector to have it's own preferences
- a setting that allows for  hover to be turned on and off
- Default setting for hover in embedded environments
- Some sweet tests to assert the above behaviour
- Added and used FlutterTestStorage to ensure that Storage state could not persist between test runs

## Demo

https://user-images.githubusercontent.com/1386322/170055479-86ebadc5-1697-4089-87a7-133364d5886c.mp4

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #4200, opened 2022-06-18
Make autosnapshotting configurable.

polina-c:


polina-c:
> Looks good! Can we add a test for the memory settings? Here's an example test group for the inspector settings: https://github.com/flutter/devtools/pull/4090/files#diff-f83bc9b9bce356789e0fbc423db7344bbd733560cf8acc2a99a483adafc8e906

Is it ok to postpone the test till having memory preferences persisted: https://github.com/flutter/devtools/issues/4188

kenzieschmoll (approved):

Issue #4081, opened 2022-05-06
Inspector goldens are unstable


kenzieschmoll:
The list of `dependencies` here does not have a stable order, which causes spurious failures in our golden testing. 
<img width="1398" alt="Screen Shot 2022-05-06 at 3 22 08 PM" src="https://user-images.githubusercontent.com/43759233/167225128-c5eaf8b6-ff1e-4269-8ade-be3c4dbd924d.png">

kenzieschmoll:
This can be closed once we bump our pinned flutter version to a version that includes https://github.com/flutter/flutter/commit/1b2ee411aa1389e74c7c2aada90bc468781532d2

CoderDake:
This issue was causing a lot of issues for merging, so this PR was created to temporarily avoid the check that flakes.
When we can pin the new flutter version we should remove the `ignoreGoldensForNow` variable and expose the code that was blocked by it

PR #4204, opened 2022-06-21
Comment out Flaky dependency tests for now

CoderDake:
Tests are a bit too flaky until we can pin to a new version in https://github.com/flutter/devtools/issues/4081

This PR makes the flakey expectations dead until that issue is resolved


packages/devtools_app/test/shared/table_test.dart
```diff
@@ -257,7 +257,7 @@ void main() {
         wrap(
           SizedBox(
             width: 200.0,
-            height: 200.0,
+            height: defaultRowHeight,
```
CoderDake, line 260: oops I'm not sure how this got in the diff. Resolving now

packages/devtools_app/test/inspector/inspector_service_test.dart
```diff
@@ -412,13 +417,16 @@ void main() async {
...
         ))!;
-        expect(
-          treeToDebugString(selection),
-          equalsIgnoringHashCodes(
-            'RenderParagraph#00000 relayoutBoundary=up2\n'
-            ' └─text: TextSpan\n',
-          ),
-        );
+        // ignore: dead_code
+        if (!ignoreGoldensForNow) {
```
kenzieschmoll, line 421: we shouldn't need to ignore this part. The part of the txt goldens that flake are lists of items that are not sorted the same way everytime, so this should be fine.

packages/devtools_app/test/inspector/inspector_service_test.dart
```diff
@@ -320,6 +320,8 @@ void main() async {
       });
 
       test('widget tree', () async {
+        // TODO(https://github.com/flutter/devtools/issues/4081): stop ignoring goldens when pinned to a new version with stable dependency orders
```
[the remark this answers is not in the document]
kenzieschmoll: linking to this issue is fine 👍

kenzieschmoll (approved):
one comment then lgtm

packages/devtools_app/test/inspector/inspector_service_test.dart
CoderDake: It fails with:
```
Expected: multi line description equals RenderParagraph#00000 relayoutBoundary=up2
           └─text: TextSpan
          
  Actual: 'RenderParagraph#7131e relayoutBoundary=up3\n'
            ' └─text: TextSpan\n'
            ''
   Which: expected normalized value
            'RenderParagraph#00000 relayoutBoundary=up2\n'
            ' └─text: TextSpan\n'
            ''
          but got
            'RenderParagraph#00000 relayoutBoundary=up3\n'
            ' └─text: TextSpan\n'
            ''
```
Do you know what that might be about?

kenzieschmoll: Is this only failing on master? something may have changed in the framework. For the other golden checks, we use `equalsGoldenIgnoringHashCodes`, which checks the `SHOULD_TEST_GOLDENS` environment variable in its `matches` method. This environment variable is only set to `true` for the bots where we use a pinned version of Flutter. 

We should create another helper method like `equalsGoldenIgnoringHashCodes` that takes a value instead of a file path, and checks `SHOULD_TEST_GOLDENS` before comparing. We have an existing matcher `_AlwaysTrueMatcher` that we can return if `SHOULD_TEST_GOLDENS` is false.

packages/devtools_app/test/matchers/matchers.dart
```diff
@@ -59,6 +59,78 @@ Matcher equalsGoldenIgnoringHashCodes(String path) {
   return _EqualsGoldenIgnoringHashCodes(path);
 }
 
+Matcher equalsGoldenValueIgnoringHashCodes(String path) {
```
kenzieschmoll, line 62: rename from path to value

kenzieschmoll (approved):

Issue #3951, opened 2022-04-01
leak-tracker: Detect memory leaks by identifying not disposed or not GCed objects.


polina-c:
Motivation: [go/flutter-devtools-survey-2022](https://goto.google.com/flutter-devtools-survey-2022)

polina-c:
- [ ] Investigation
    - [ ] prototype WeakMap     
    - [ ] understand Finalizer API
    - [ ] understand app to DevTools connection
    - [ ] design: https://docs.google.com/document/d/1Qcah_gb9b5aoxUdtX4-Nxp8zrsaJ9LmSvXYHi3MZ194/edit#

polina-c:
Drafts:
flutter: https://github.com/flutter/flutter/pull/103765
tool: https://github.com/polina-c/spikes/tree/master/memory_tools
dev_tools:

polina-c:
Options for Flutter specific leak tracker:

1.  package:leak_tracker_flutter published in [tools.dart.dev](http://tools.dart.dev/), and banking on 'flutter' in the name being clarifying enough
2. keeping leak_tracker_flutter in the same mono-repo, but publishing it out of the [flutter.dev](http://flutter.dev/) publisher
3. putting the code in the flutter/packages mono-repo and using the [flutter.dev](http://flutter.dev/) publisher

@devoncarew  leans towards the 2nd and 3rd options

stuartmorgan:
> 2\. keeping leak_tracker_flutter in the same mono-repo, but publishing it out of the [flutter.dev](http://flutter.dev/) publisher

It appears we did this; is there discussion/context somewhere on why that option was chosen? AFAIK this is the *only* `flutter.dev`-published package that is not in flutter/packages, which makes this pretty confusing.

polina-c:
> > 2. keeping leak_tracker_flutter in the same mono-repo, but publishing it out of the [flutter.dev](http://flutter.dev/) publisher
> 
> It appears we did this; is there discussion/context somewhere on why that option was chosen? AFAIK this is the _only_ `flutter.dev`-published package that is not in flutter/packages, which makes this pretty confusing.

Here is the issue with discussion: https://github.com/dart-lang/leak_tracker/issues/52#issuecomment-1666920653

Solido:
Documentation is explicit about pure Dart packages not being supported.
Will this land in the roadmap or is there a technical roadblock? Thanks!

polina-c:
[...]

PR #4211, opened 2022-06-22
Create tab 'Leaks'

polina-c:
https://github.com/flutter/devtools/issues/3951

### DIFF
--- a/packages/devtools_shared/lib/src/memory/leaks/constants.dart
+++ b/packages/devtools_shared/lib/src/memory/leaks/constants.dart
@@ -3,4 +3,4 @@
 // found in the LICENSE file.
 
 /// Name of extension that enables leak tracking for applications.
-const memoryLeakTrackingExtensionName = 'ext.flutter.memoryLeakTracking';
+const memoryLeakTrackingExtensionName = 'ext.dart.memoryLeakTracking';


packages/devtools_app/lib/src/screens/memory/memory_heap_tree_view.dart
```diff
@@ -454,6 +473,9 @@ class HeapTreeViewState extends State<HeapTree>
                     ],
                   ),
                 ),
+
+                // Leaks tab.
+                if (controller.shouldShowLeaksTab.value) const LeaksPane(),
```
kenzieschmoll, line 478: Wrap this in KeepAliveWrapper so that the LeaksPane is not completely destroyed and rebuilt on tab switches.

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
```diff
@@ -815,7 +818,16 @@ class MemoryController extends DisposableController
     // TODO(terry): Need an event on the controller for this too?
   }
 
-  void _handleConnectionStart(ServiceConnectionManager serviceManager) {
+  Future<void> _refreshShouldShowLeaksTab() async {
+    final isolateId = serviceManager.isolateManager.mainIsolate.value!.id!;
+    final Isolate isolate = await serviceManager.service!.getIsolate(isolateId);
+    _shouldShowLeaksTab.value =
+        isolate.extensionRPCs!.contains(leakTrackingExtensionName);
```
kenzieschmoll, line 825: use `serviceManager.serviceExtensionManager.hasServiceExtension(leakTrackingExtensionName)`
for this.

Example: https://github.com/flutter/devtools/blob/05d9d7482837bcce7659dbc0b1af7978500ba0c7/packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart#L174

packages/devtools_shared/lib/src/memory/leaks/constants.dart
```diff
@@ -0,0 +1,6 @@
+// Copyright 2021 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Name of extension that enables leak tracking for applications.
+const leakTrackingExtensionName = 'ext.memory-leak-tracking';
```
kenzieschmoll, line 6: add this to service_extensions.dart instead. This is where all our service extensions constants are defined. This extension name will need to be defined in both DevTools and the Framework.

Also, rename to 'ext.flutter.memoryLeakTracking' to be consistent with flutter service extension naming practices. We have some code that checks whether an extension is a flutter service extension by verifying that it starts with 'ext.flutter.'

packages/devtools_shared/lib/src/memory/leaks/constants.dart
```diff
@@ -3,4 +3,4 @@
 // found in the LICENSE file.
 
 /// Name of extension that enables leak tracking for applications.
-const leakTrackingExtensionName = 'ext.memory-leak-tracking';
+const memoryLeakTrackingExtensionName = 'ext.flutter.memory.leak-tracking';
```
kenzieschmoll, line 6: use camel case: `ext.flutter.memory.leakTracking`

Also I'd recommend `ext.flutter.memoryLeakTracking` instead of `ext.flutter.memory.leakTracking` to be consistent with the other service extensions registered in the flutter framework.

kenzieschmoll (approved):

PR #4214, opened 2022-06-23
Add support for column groups in tables

kenzieschmoll:
This will be used in new memory tooling that @bkonyi is working on. This screenshot is example usage of this feature and these groups will not actually be added to the logging page.

Screen Shot 2022-06-23 at 12 25 50 PM

### DIFF
--- a/packages/devtools_app/lib/src/shared/table.dart
+++ b/packages/devtools_app/lib/src/shared/table.dart
@@ -31,6 +31,10 @@ import 'utils.dart';
 /// height of 0 and a height of [defaultRowHeight].
 double get defaultRowHeight => scaleByFontFactor(32.0);
 
+const _columnGroupSpacing = 4.0;
+const _columnGroupSpacingWithPadding = _columnGroupSpacing + 2 * defaultSpacing;
+const _columnSpacing = defaultSpacing;
+
 typedef IndexedScrollableWidgetBuilder = Widget Function(
   BuildContext,
   LinkedScrollControllerGroup linkedScrollControllerGroup,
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -1503,6 +1618,100 @@ class _TableRowState<T> extends State<TableRow<T>>
...
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+      padding: const EdgeInsets.symmetric(
+        horizontal: _columnGroupSpacingWithPadding - _columnGroupSpacing,
+      ),
+      child: Container(
+        width: _columnGroupSpacing,
+        decoration: BoxDecoration(
+          gradient: LinearGradient(
```
bkonyi, line 1686: Oh cool, I didn't know this was this simple!

packages/devtools_app/lib/src/shared/table_data.dart
```diff
@@ -97,3 +98,12 @@ enum ColumnAlignment {
   right,
   center,
 }
+
+class ColumnGroup {
+  ColumnGroup({required this.name, required this.range});
+
+  final String name;
```
bkonyi, line 105: Nit: maybe `title`?

packages/devtools_app/lib/src/shared/table_data.dart
```diff
@@ -97,3 +98,12 @@ enum ColumnAlignment {
   right,
   center,
 }
+
+class ColumnGroup {
```
bkonyi, line 102: Nit: document what this class is for.

bkonyi (approved):
LGTM with nits.

Thanks for doing this!

packages/devtools_app/lib/src/shared/table_data.dart
kenzieschmoll: done

packages/devtools_app/lib/src/shared/table_data.dart
kenzieschmoll: done

PR #4215, opened 2022-06-25
Fixes 2 bugs that were happening in the VM tools tabs

migueltylee:
Fixes the following bugs:

* On isolate_statistics_view.dart one of the tables was overflowing: adjusted flex factors to better distribute the space available. Also added a scrollbar to the VMInfoList widget so that the widget does not overflow when resizing into a smaller window.
* On vm_statistics_view.dart a type error was occurring (expected int but received null) when calling vm?.nativeZoneMemoryUsage (line 193): made the nativeZoneMemoryUsage nullable on the vm_service_private_extensions.dart file.

### DIFF
--- a/packages/devtools_app/lib/src/screens/vm_developer/isolate_statistics_view.dart
+++ b/packages/devtools_app/lib/src/screens/vm_developer/isolate_statistics_view.dart
@@ -79,11 +79,13 @@ class IsolateStatisticsViewBody extends StatelessWidget {
           child: Column(
             children: [
               Flexible(
+                flex: 3,
                 child: GeneralIsolateStatisticsWidget(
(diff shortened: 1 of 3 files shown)


kenzieschmoll (approved):

kenzieschmoll:
WRT to tests, this code doesn't currently have any tests, so yes adding tests would be a good clean up. (We should file an issue for this). Likely out of scope for this PR though as we need to write tests for this entire screen in general. The VM developer page is designed mainly for VM developers on the Dart team, so at the time it was added, we prioritized getting features landed over adding tests.

Miguel is adding tests for the object inspector view he is building, so this will be well-tested, but the other features in this page should have tests backfilled at some point.

PR #4223, opened 2022-06-29
Receive leak summary reports.

polina-c:
This change makes the Memory page receiving periodic leak tracking summary.

See steps in https://github.com/flutter/devtools/blob/c2e2a65329b07b6d2db5364d6458ffbd25904649/packages/devtools_app/lib/src/screens/memory/panes/leaks/LEAK_TRACKING.md#detect-leaks-in-demo-app


packages/devtools_app/lib/src/screens/memory/panes/leaks/instrumentation/model.dart
```diff
@@ -0,0 +1,44 @@
+// Copyright 2021 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:collection/collection.dart';
+
+enum LeakType {
+  notDisposed,
+  notGCed,
+  gcedLate,
```
CoderDake, line 10:
```suggestion
  notGCed,
  gcedLate,
```
```suggestion
  notGarbageCollected,
  garbageCollectedLate,
```
avoiding acronyms may help with readability in the future

packages/devtools_app/lib/src/screens/memory/panes/leaks/instrumentation/model.dart
```diff
@@ -0,0 +1,44 @@
...
+
+  final Map<LeakType, int> totals;
+
+  bool get isEmpty => totals.values.sum == 0;
+
+  String toMessage() {
+    return '${totals.values.sum} memory leak(s): '
+        'not disposed: ${totals[LeakType.notDisposed]}, '
+        'not GCed: ${totals[LeakType.notGCed]}, '
+        'GCed late: ${totals[LeakType.gcedLate]}';
```
CoderDake, line 34: same about acronyms here

packages/devtools_app/lib/src/screens/memory/memory_heap_tree_view.dart
```diff
@@ -206,7 +208,9 @@ class HeapTreeViewState extends State<HeapTree>
         ),
     ];
 
+    _searchabeleTabs = {0, 1};
     _tabController = TabController(length: _tabs.length, vsync: this);
+    _tabController.addListener(() => _currentTab.value = _tabController.index);
```
kenzieschmoll, line 213: we may not need to add a listener for this. See this example where we can lookup the current tab directly without having to add a listener and a ValueListenableBuilder:

https://github.com/flutter/devtools/blob/063b56d62bdfad2b3b89ed487b8ccea373cf06c5/packages/devtools_app/lib/src/screens/app_size/app_size_screen.dart#L188

packages/devtools_app/lib/src/screens/memory/panes/leaks/LEAK_TRACKING.md
```diff
@@ -1,3 +1,37 @@
...
+
+## Understand concepts
+
+
+## Configure environment
+
+For Google3 applications, follow http://go/detect-memory-leaks-in-g3-flutter-app.
+
+Follow [standard Flutter installation process](https://docs.flutter.dev/get-started/install), but,
+instead of downloading or cloning official Flutter, clone `git@github.com:polina-c/flutter.git`,
```
kenzieschmoll, line 15: are your changes on a specific branch of your fork?

packages/devtools_app/lib/src/screens/memory/panes/leaks/instrumentation/model.dart
```diff
@@ -0,0 +1,44 @@
+// Copyright 2021 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: 2022

packages/devtools_app/lib/src/screens/memory/panes/leaks/instrumentation/model.dart
```diff
@@ -0,0 +1,44 @@
...
+
+import 'package:collection/collection.dart';
+
+enum LeakType {
+  notDisposed,
+  notGCed,
+  gcedLate,
+}
+
+LeakType _parseLeakType(String source) =>
```
kenzieschmoll, line 13: can we use enhanced enums to do this? If not we can create an extension method. See [example](https://github.com/flutter/devtools/blob/ecba4cd9dbdbe2e15ec0d299d24b8f23d9a151ab/packages/devtools_app/lib/src/screens/profiler/profile_granularity.dart#L29).

packages/devtools_app/lib/src/screens/memory/panes/leaks/instrumentation/model.dart
```diff
@@ -0,0 +1,44 @@
...
+  bool get isEmpty => totals.values.sum == 0;
+
+  String toMessage() {
+    return '${totals.values.sum} memory leak(s): '
+        'not disposed: ${totals[LeakType.notDisposed]}, '
+        'not GCed: ${totals[LeakType.notGCed]}, '
+        'GCed late: ${totals[LeakType.gcedLate]}';
+  }
+
+  bool equals(LeakSummary? other) {
```
kenzieschmoll, line 37: override == instead of creating a method `equals`

packages/devtools_app/lib/src/screens/memory/panes/leaks/leaks_pane.dart
```diff
@@ -1,4 +1,12 @@
 import 'package:flutter/material.dart';
+import 'package:intl/intl.dart';
+
+import '../../../../primitives/auto_dispose_mixin.dart';
+import '../../../../shared/globals.dart';
+import 'instrumentation/model.dart';
+
+final DateFormat _formatter = DateFormat.Hms();
+String _timeForConsole(DateTime time) => _formatter.format(time);
```
kenzieschmoll, line 9: can you use `formatDateTime` from utils.dart instead?

packages/devtools_app/lib/src/screens/memory/panes/leaks/leaks_pane.dart
```diff
@@ -7,13 +15,46 @@ class LeaksPane extends StatefulWidget {
...
 
-class _LeaksPaneState extends State<LeaksPane> {
+class _LeaksPaneState extends State<LeaksPane> with AutoDisposeMixin {
+  LeakSummary? _lastLeakSummary;
+  String _leakSummaryHistory = '';
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+    _subscribeForMemoryLeaksSummary();
```
kenzieschmoll, line 25: call cancelStreamSubscriptions() before calling _subscribeForMemoryLeaksSummary()

packages/devtools_shared/pubspec.yaml
```diff
@@ -17,3 +18,5 @@ dependencies:
 
 dev_dependencies:
   test: ^1.21.2
+  build_runner: ^2.0.4
+  mockito: ^5.1.0
```
kenzieschmoll, line 22: do we still need all these changes to the pubspec? they don't look related to any devtools_shared changes in this PR

packages/devtools_app/lib/src/screens/memory/panes/leaks/LEAK_TRACKING.md
```diff
@@ -1,3 +1,37 @@
-# Memory Leaks Tracking with DevTools
+# Memory leaks tracking with Dart DevTools
 
 This page and functionality are under construction. See https://github.com/flutter/devtools/issues/3951.
+
+[self-link](https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/memory/panes/leaks/LEAK_TRACKING.md)
+
+## Understand concepts
+
```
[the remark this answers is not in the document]
polina-c: Yes. The comment above states the page is 'under construction'.

packages/devtools_app/lib/src/screens/memory/panes/leaks/LEAK_TRACKING.md
polina-c: I use 'master'.

packages/devtools_app/lib/src/screens/memory/panes/leaks/LEAK_TRACKING.md
```diff
@@ -1,3 +1,37 @@
...
+For Google3 applications, follow http://go/detect-memory-leaks-in-g3-flutter-app.
+
+Follow [standard Flutter installation process](https://docs.flutter.dev/get-started/install), but,
+instead of downloading or cloning official Flutter, clone `git@github.com:polina-c/flutter.git`,
+and then never run `flutter upgrade` or `flutter channel`.
+
+## Detect leaks in demo app
+
+TODO: move the example to test/fixtures when it compiles with stable flutter.
```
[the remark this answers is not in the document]
polina-c: It is ok to show.

packages/devtools_app/lib/src/screens/memory/panes/leaks/LEAK_TRACKING.md
```diff
@@ -1,3 +1,37 @@
...
+resize the app window, to trigger GC events, and the message should show up:
+   
+```
+flutter: 1 memory leaks: not disposed: 1, not GCed: 0, GCed late: 0
+flutter: 3 memory leaks: not disposed: 1, not GCed: 2, GCed late: 0
+```
+
+## Detect leaks in your Flutter app
+
+
```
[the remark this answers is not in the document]
polina-c: Yes, as the document is under construction.

packages/devtools_app/lib/src/screens/memory/panes/leaks/instrumentation/model.dart
polina-c: 'GC' is the important concept of this feature (like html or http) that will be described in the central document. I prefer to leave the abbreviation in order to have shorter names. I added comments though. Thanks.

packages/devtools_app/lib/src/screens/memory/panes/leaks/instrumentation/model.dart
polina-c: I want the method to be private for the library, because it is not needed outside. If I make the method part of enum, it will become available for the library users.

packages/devtools_app/lib/src/screens/memory/panes/leaks/instrumentation/model.dart
polina-c: Replied above.

packages/devtools_app/lib/src/screens/memory/panes/leaks/instrumentation/model.dart
polina-c: I prefer 'equals', because the reader of the code that uses '==' will not know if '==' is overriden or not.

It would make sense to override '==' if we used the objects as map keys or something like this.

kenzieschmoll: We override `==` in many places in DevTools, so it would be completely reasonable to do this here, too. If you want `==` to continue to compare object identity, then let's rename this method to `matches`. It is confusing to have `equals` and `==` that do not do the same thing. Better to have the term "equals" mean true equality as determined by `==` than to have two different meanings.

packages/devtools_app/lib/src/screens/memory/memory_heap_tree_view.dart
kenzieschmoll: Interesting. May be a function of how / when we are rebuilding the tree.

In that case, put the listener into a private method `_onTabChanged` or similar so that you can add the listener with that method and remove the listener in dispose. The tab controller also needs to be disposed

```
_tabController
    ..removeListener(_onTabChanged)
    ..dispose();
```

packages/devtools_app/lib/src/screens/memory/memory_heap_tree_view.dart
```diff
@@ -206,7 +208,9 @@ class HeapTreeViewState extends State<HeapTree>
         ),
     ];
 
+    _searchabeleTabs = {0, 1};
```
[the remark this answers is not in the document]
polina-c: indeed. thanks

packages/devtools_app/lib/src/screens/memory/panes/leaks/instrumentation/model.dart
polina-c: After some thinking I realized there is no value in comparing LeakSummary by identity. Thanks.

polina-c: Then I've read the code if (oldSummary == newSummary) and realized that to verify this code I need to make sure '==' is overriden. So, created the method 'matches'.

packages/devtools_app/lib/src/screens/memory/panes/leaks/LEAK_TRACKING.md
polina-c: Actually, I use leak-tracking2. Added to the document.

CoderDake (approved):

PR #4226, opened 2022-06-29
Connect the TreeTable and TreeMap 

carolynqu:
**Previously**
When interacting with the TreeMap, the TreeTable would re-root and expand its nodes according to the selection. However, when interacting with the TreeTable, the TreeMap would not update with it. 

**Changes**
Added new function (onAnalysisRootChange) to handle changing analysisRoot when updating the table so that the map also updates. Removed TreeTable re-rooting according to analysisRoot and replacing the root to be originalRoot. Only allow the map to be updated by TreeTable and not Dominator Tree. 

**Next Steps**
Need to add a scrolling and expanding functionality for when the map is being interacted with. Add better checking to determine if analysisRoot should be updated.

### DIFF
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
@@ -51,6 +51,9 @@ class AppSizeController {
   ValueListenable<TreemapNode?> get analysisRoot => _analysisRoot;
   final _analysisRoot = ValueNotifier<TreemapNode?>(null);
 
+  ValueListenable<TreemapNode?> get originalRoot => _originalRoot;
+  final _originalRoot = ValueNotifier<TreemapNode?>(null);
+
   void changeAnalysisRoot(TreemapNode? newRoot) {
     _analysisRoot.value = newRoot;
     if (newRoot == null) return;
@@ -218,6 +221,7 @@ class AppSizeController {
 
     // Build a tree with [TreemapNode] from [processedJsonMap].
     final newRoot = generateTree(processedJson)!;
+    _originalRoot.value = newRoot;
 
     changeAnalysisRoot(newRoot);
 
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_screen.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_screen.dart
@@ -273,6 +273,7 @@ class _AnalysisViewState extends State<AnalysisView>
         AutoDisposeMixin,
         ProvidedControllerMixin<AppSizeController, AnalysisView> {
   TreemapNode? analysisRoot;
+  TreemapNode? originalRoot;
 
   @override
   void didChangeDependencies() {
@@ -280,9 +281,12 @@ class _AnalysisViewState extends State<AnalysisView>
     if (!initController()) return;
 
     analysisRoot = controller.analysisRoot.value;
+    originalRoot = controller.originalRoot.value;
+
     addAutoDisposeListener(controller.analysisRoot, () {
       setState(() {
         analysisRoot = controller.analysisRoot.value;
+        originalRoot = controller.originalRoot.value;
       });
     });
 
@@ -294,7 +298,7 @@ class _AnalysisViewState extends State<AnalysisView>
     return Column(
       children: [
         Expanded(
-          child: analysisRoot != null
+          child: analysisRoot != null && originalRoot != null
               ? _buildTreemapAndTableSplitView()
               : _buildImportFileView(),
         ),
@@ -325,7 +329,7 @@ class _AnalysisViewState extends State<AnalysisView>
                   children: [
                     Flexible(
                       child: AppSizeAnalysisTable(
-                        rootNode: analysisRoot!,
+                        rootNode: originalRoot!,
                         controller: controller,
                       ),
                     ),
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_screen.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_screen.dart
@@ -324,7 +324,10 @@ class _AnalysisViewState extends State<AnalysisView>
                 Row(
                   children: [
                     Flexible(
-                      child: AppSizeAnalysisTable(rootNode: analysisRoot!),
+                      child: AppSizeAnalysisTable(
+                        rootNode: analysisRoot!,
+                        controller: controller,
+                      ),
                     ),
                     if (analysisCallGraphRoot != null)
                       Flexible(
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
@@ -11,8 +11,11 @@ import '../../shared/table_data.dart';
 import '../../shared/utils.dart';
 import '../../ui/colors.dart';
 
+import 'app_size_controller.dart';
+
 class AppSizeAnalysisTable extends StatelessWidget {
-  factory AppSizeAnalysisTable({required TreemapNode rootNode}) {
+  factory AppSizeAnalysisTable(
+      {required TreemapNode rootNode, required AppSizeController controller}) {
     final treeColumn = _NameColumn(currentRootLevel: rootNode.level);
     final sizeColumn = _SizeColumn();
     final columns = List<ColumnData<TreemapNode>>.unmodifiable([
@@ -42,6 +45,8 @@ class AppSizeAnalysisTable extends StatelessWidget {
   final ColumnData<TreemapNode> sortColumn;
   final List<ColumnData<TreemapNode>> columns;
 
+  final AppSizeController controller;
+
   @override
   Widget build(BuildContext context) {
     return TreeTable<TreemapNode>(
@@ -51,6 +56,7 @@ class AppSizeAnalysisTable extends StatelessWidget {
       keyFactory: (node) => PageStorageKey<String>(node.name),
       sortColumn: sortColumn,
       sortDirection: SortDirection.descending,
+      onChange: controller.changeAnalysisRoot,
       autoExpandRoots: true,
     );
   }
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
@@ -29,6 +29,7 @@ class AppSizeAnalysisTable extends StatelessWidget {
       treeColumn,
       sizeColumn,
       columns,
+      controller,
     );
   }
 
@@ -37,6 +38,7 @@ class AppSizeAnalysisTable extends StatelessWidget {
     this.treeColumn,
     this.sortColumn,
     this.columns,
+    this.controller,
   );
 
   final TreemapNode rootNode;
@@ -56,7 +58,7 @@ class AppSizeAnalysisTable extends StatelessWidget {
       keyFactory: (node) => PageStorageKey<String>(node.name),
       sortColumn: sortColumn,
       sortDirection: SortDirection.descending,
-      onChange: controller.changeAnalysisRoot,
+      onAnalysisRootChange: controller.changeAnalysisRoot,
       autoExpandRoots: true,
     );
   }
--- a/packages/devtools_app/lib/src/shared/table.dart
+++ b/packages/devtools_app/lib/src/shared/table.dart
@@ -348,7 +348,7 @@ class TreeTable<T extends TreeNode<T>> extends StatefulWidget {
     required this.sortDirection,
     this.secondarySortColumn,
     this.selectionNotifier,
-    required this.onChange,
+    this.onAnalysisRootChange,
     this.autoExpandRoots = false,
   })  : assert(columns.contains(treeColumn)),
         assert(columns.contains(sortColumn)),
@@ -374,7 +374,7 @@ class TreeTable<T extends TreeNode<T>> extends StatefulWidget {
 
   final ValueNotifier<Selection<T>>? selectionNotifier;
 
-  final void Function(T? node) onChange;
+  final void Function(T? node)? onAnalysisRootChange;
 
   final bool autoExpandRoots;
 
@@ -489,9 +489,6 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
       nodeIndex: nodeIndex,
     );
 
-    widget.onChange(node);
-    //add changeAnalysisRoot (node)
-
     _toggleNode(node);
   }
 
@@ -499,6 +496,7 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     if (!node.isExpandable) {
       node.leaf();
       _updateItems();
+      widget.onAnalysisRootChange!(node);
       return;
     }
 
--- a/packages/devtools_app/lib/src/shared/table.dart
+++ b/packages/devtools_app/lib/src/shared/table.dart
@@ -348,6 +348,7 @@ class TreeTable<T extends TreeNode<T>> extends StatefulWidget {
     required this.sortDirection,
     this.secondarySortColumn,
     this.selectionNotifier,
+    required this.onChange,
     this.autoExpandRoots = false,
   })  : assert(columns.contains(treeColumn)),
         assert(columns.contains(sortColumn)),
@@ -372,6 +373,8 @@ class TreeTable<T extends TreeNode<T>> extends StatefulWidget {
   final ColumnData<T>? secondarySortColumn;
 
   final ValueNotifier<Selection<T>>? selectionNotifier;
+
+  final void Function(T? node) onChange;
 
   final bool autoExpandRoots;
 
@@ -485,6 +488,9 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
       node: node,
       nodeIndex: nodeIndex,
     );
+
+    widget.onChange(node);
+    //add changeAnalysisRoot (node)
 
     _toggleNode(node);
   }
--- a/packages/devtools_app/lib/src/shared/table.dart
+++ b/packages/devtools_app/lib/src/shared/table.dart
@@ -13,6 +13,7 @@ import '../primitives/flutter_widgets/linked_scroll_controller.dart';
 import '../primitives/listenable.dart';
 import '../primitives/trees.dart';
 import '../primitives/utils.dart';
+import '../screens/app_size/code_size_attribution.dart';
 import '../ui/colors.dart';
 import '../ui/search.dart';
 import 'collapsible_mixin.dart';
@@ -503,8 +504,11 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
   }
 
   void _toggleNode(T node) {
-    widget.onAnalysisRootChange!(node);
+    //TODO: not a great way to check for null
 
+    if (node is! DominatorTreeNode) {
+      widget.onAnalysisRootChange!(node);
+    }
     if (!node.isExpandable) {
       node.leaf();
       _updateItems();
--- a/packages/devtools_app/lib/src/shared/table.dart
+++ b/packages/devtools_app/lib/src/shared/table.dart
@@ -493,13 +493,14 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
   }
 
   void _toggleNode(T node) {
+    widget.onAnalysisRootChange!(node);
+
     if (!node.isExpandable) {
       node.leaf();
       _updateItems();
-      widget.onAnalysisRootChange!(node);
+
       return;
     }
-
     setState(() {
       if (!node.isExpandable) return;
       animatingNode = node;


packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -358,6 +359,7 @@ class TreeTable<T extends TreeNode<T>> extends StatefulWidget {
     required this.sortDirection,
     this.secondarySortColumn,
     this.selectionNotifier,
+    this.onAnalysisRootChange,
```
elliette, line 362: Let's rename this to something generic like onToggleNode

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -500,12 +504,17 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
   }
 
   void _toggleNode(T node) {
+    //TODO: not a great way to check for null
+
+    if (node is! DominatorTreeNode) {
```
elliette, line 509: Let's move the widget.onAnalysisRootChange to the function that is passed in called onToggleNode

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -51,6 +51,9 @@ class AppSizeController {
   ValueListenable<TreemapNode?> get analysisRoot => _analysisRoot;
   final _analysisRoot = ValueNotifier<TreemapNode?>(null);
 
+  ValueListenable<TreemapNode?> get originalRoot => _originalRoot;
+  final _originalRoot = ValueNotifier<TreemapNode?>(null);
```
elliette, line 55: Let's call this `appRoot`

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -218,6 +221,7 @@ class AppSizeController {
 
     // Build a tree with [TreemapNode] from [processedJsonMap].
     final newRoot = generateTree(processedJson)!;
```
elliette, line 223: This method is only invoked when we first load the JSON file into DevTools, right? If that's the case, let's rename `newRoot` to `appRoot` just to make it more clear.

packages/devtools_app/lib/src/screens/app_size/app_size_screen.dart
```diff
@@ -273,16 +273,20 @@ class _AnalysisViewState extends State<AnalysisView>
         AutoDisposeMixin,
         ProvidedControllerMixin<AppSizeController, AnalysisView> {
   TreemapNode? analysisRoot;
+  TreemapNode? originalRoot;
```
elliette, line 276: Again, let's rename `originalRoot` to `appRoot`

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -500,12 +504,17 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
...
+      widget.onAnalysisRootChange!(node);
+    }
     if (!node.isExpandable) {
       node.leaf();
       _updateItems();
+
```
elliette, line 515: Nit: remove line

kenzieschmoll:
High level: Could you provide a gif showing the new behavior? This tool is pretty useful for that: https://giphy.com/apps/giphycapture.

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
kenzieschmoll: We shouldn't need to track another notifier / listenable pair for this. The root should be accessible from the TreemapNode (see TreeNode.root). Take a look at the BreadcrumbNavigator we use in the app size page. The breadcrumb navigator is able to calculate the path from the current node to the root, so this should be possible without tracking another var.

packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
```diff
@@ -42,6 +49,7 @@ class AppSizeAnalysisTable extends StatelessWidget {
   final ColumnData<TreemapNode> sortColumn;
   final List<ColumnData<TreemapNode>> columns;
 
+  final AppSizeController controller;
   @override
```
kenzieschmoll, line 53: nit separate class members with new lines

packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
```diff
@@ -51,6 +59,7 @@ class AppSizeAnalysisTable extends StatelessWidget {
       keyFactory: (node) => PageStorageKey<String>(node.name),
       sortColumn: sortColumn,
       sortDirection: SortDirection.descending,
+      onToggleNode: (TreemapNode n) => controller.changeAnalysisRoot(n),
```
kenzieschmoll, line 62: we may be able to solve this in a simpler way by setting the selectionNotifier parameter to `controller.analysisRoot`.

in TreeTableState, we automatically expand the parents of the selected node: https://github.com/flutter/devtools/blob/d24491817645320c789a0db624364982982d5c37/packages/devtools_app/lib/src/shared/table.dart#L442

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -401,7 +401,7 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
   late List<bool> rootsExpanded;
   late FocusNode _focusNode;
 
-  late ValueNotifier<Selection<T>> selectionNotifier;
+  late ValueNotifier<Selection<T>>? selectionNotifier;
```
kenzieschmoll, line 404: why does this need to be nullable?

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -48,13 +49,15 @@ class AppSizeController {
   /// The node set as the analysis tab root.
   ///
   /// Used to build the treemap and the tree table for the analysis tab.
-  ValueListenable<TreemapNode?> get analysisRoot => _analysisRoot;
-  final _analysisRoot = ValueNotifier<TreemapNode?>(null);
+  ValueListenable<Selection<TreemapNode>> get analysisRoot => _analysisRoot;
+  final _analysisRoot =
+      ValueNotifier<Selection<TreemapNode>>(Selection<TreemapNode>());
```
kenzieschmoll, line 54: if we need to make analysisRoot accessible from the tree table, we can just get rid of the public listenable and make the notifier public:
```
final analysisRoot =
      ValueNotifier<Selection<TreemapNode>>(Selection<TreemapNode>());
```

then you won't have to cast later when you use `analysisRoot` as the `selection` notifier

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -438,7 +438,7 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
 
     addAutoDisposeListener(selectionNotifier, () {
       setState(() {
-        final node = selectionNotifier.value.node;
+        final node = selectionNotifier!.value.node;
```
kenzieschmoll, line 441: you should be able to remove all these null assertions `!` if `selectionNotifier` is non-nullable

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -48,13 +49,17 @@ class AppSizeController {
...
-  final _analysisRoot = ValueNotifier<TreemapNode?>(null);
+  final analysisRoot =
+      ValueNotifier<Selection<TreemapNode>>(Selection<TreemapNode>());
+
+  ValueListenable<int> get selectedNodeIndex => _selectedNodeIndex;
+  final _selectedNodeIndex = ValueNotifier<int>(0);
 
   void changeAnalysisRoot(TreemapNode? newRoot) {
-    _analysisRoot.value = newRoot;
     if (newRoot == null) return;
```
elliette, line 59: We should probably do the same thing we are doing on line 160 if `newRoot` is null (set the analysisRoot to be an empty selection)

packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
```diff
@@ -11,21 +11,28 @@ import '../../shared/table_data.dart';
...
+
 class AppSizeAnalysisTable extends StatelessWidget {
-  factory AppSizeAnalysisTable({required TreemapNode rootNode}) {
-    final treeColumn = _NameColumn(currentRootLevel: rootNode.level);
+  factory AppSizeAnalysisTable({
+    required TreemapNode rootNode,
+    required AppSizeController controller,
+  }) {
+    final treeColumn =
+        _NameColumn(currentRootLevel: rootNode.children[0].level);
```
elliette, line 22: I'm guessing this is part of the deferred loading changes? Deferred loading-specific logic should be removed from this PR

packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
```diff
@@ -42,16 +50,18 @@ class AppSizeAnalysisTable extends StatelessWidget {
   final ColumnData<TreemapNode> sortColumn;
   final List<ColumnData<TreemapNode>> columns;
 
+  final AppSizeController controller;
+
   @override
   Widget build(BuildContext context) {
     return TreeTable<TreemapNode>(
-      dataRoots: [rootNode],
+      dataRoots: rootNode.children,
```
elliette, line 58: Again, just rootNode

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -432,14 +458,48 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
...
     addAutoDisposeListener(selectionNotifier, () {
       setState(() {
         final node = selectionNotifier.value.node;
+        //offset will need to be modified with new UI table changes
         expandParents(node?.parent);
+
+        node?.select();
+
+        final searchCondition = (T n) => n.isSelected;
+        final int selectedDisplayRow = node!.childCountToMatchingNode(
```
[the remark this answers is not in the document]
kenzieschmoll: I'm referring to `node!.nodeIndex`, not `node!.index` (like here: https://github.com/flutter/devtools/blob/d24491817645320c789a0db624364982982d5c37/packages/devtools_app/lib/src/shared/table.dart#L864). Traversing the entire tree to find the node's index in the tree table list can be very expensive, especially for large trees, so it's better to cache this info. Which is what `nodeIndex` does IIRC.

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -48,12 +50,23 @@ class AppSizeController {
...
+      analysisRoot.value = Selection<TreemapNode>();
+      return;
+    }
+
+    final searchCondition = (TreemapNode n) => n == newRoot;
+    final int selectedRowIndex = newRoot.root.childCountToMatchingNode(
+      matchingNodeCondition: searchCondition,
+      includeCollapsedNodes: false,
+    );
+    final int newPos = selectedRowIndex * defaultRowHeight as int;
```
kenzieschmoll, line 67: this is calculating the pixel that the row starts at, and I think you just want the selected row index here index here. So this line is not needed, and below, you can just write 

`analysisRoot.value = Selection(node: newRoot, flatListIndex: selectedRowIndex);`

kenzieschmoll: also add `scrollIntoView: true` here when you are creating the instance of `Selection`

`Selection(node: newRoot, flatListIndex: selectedRowIndex, scrollIntoView: true)`

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -30,6 +31,7 @@ import 'utils.dart';
 /// When rows in the table expand or collapse, they will animate between a
 /// height of 0 and a height of [defaultRowHeight].
 double get defaultRowHeight => scaleByFontFactor(32.0);
+double get selectedNodePadding => defaultRowHeight * .5;
```
kenzieschmoll, line 34: I don't think we need this - doesn't look like it is getting used anywhere.

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -128,6 +132,24 @@ class FlatTableState<T> extends State<FlatTable<T>>
         !collectionEquals(widget.data, oldWidget.data)) {
       _initData();
     }
+
```
kenzieschmoll, line 135: can you try to remove this, and add this `addAutoDisposeListener` block back to _TableState, where the diff shows it being deleted? Now that we are computing the `flatListIndex` for the tree table's selected node, and setting `scrollIntoView: true`, the scrolling might work out of the box without having to make any changes to the table code.

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -432,14 +459,15 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
...
-
     cancelListeners();
 
     addAutoDisposeListener(selectionNotifier, () {
       setState(() {
         final node = selectionNotifier.value.node;
+        final index = selectionNotifier.value.flatListIndex as double;
+
         expandParents(node?.parent);
+        scrollToPosition(scrollController, index);
```
kenzieschmoll, line 470: we shouldn't need to add any of this if we add the `selectionNotifier` listener back to the base _TableState class.

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -801,7 +861,9 @@ class _Table<T> extends StatefulWidget {
     this.selectionNotifier,
     this.activeSearchMatchNotifier,
     this.rowItemExtent,
-  }) : super(key: key);
+    ScrollController? scrollController,
+  })  : scrollController = scrollController ?? ScrollController(),
```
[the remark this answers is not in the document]
kenzieschmoll: what I'm suggesting is that we should not be creating the `ScrollController()` here in the parameter list. Instead, in the State class' `initState` method (`_TableState.initState`), we should create the `ScrollController` if `widget.scrollController` is null.

inside _TableState:
```
late ScrollController scrollController; // this is what the code used to be
...
void initState() {
  ...
  scrollController = widget.scrollController ?? ScrollController();
}
```

However, with the changes I am suggesting I don't think we'll need to pass in the scroll controller at all. With setting `scrollIntoView: true` I suspect the scrolling will work out of the box without changes to the table code.

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -861,13 +866,12 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
...
         if (selection.scrollIntoView) {
-          final int selectedDisplayRow = (selection.node! as dynamic).nodeIndex;
+          final int selectedDisplayRow = selection.nodeIndex!;
           // TODO(terry): Optimize selecting row, if row's visible in
           //              the viewport just select otherwise jumpTo row.
           final newPos = selectedDisplayRow * defaultRowHeight;
 
-          // TODO(terry): Should animate factor out _moveSelection to reuse here.
-          scrollController.jumpTo(newPos);
+          scrollToPosition(scrollController, newPos);
```
kenzieschmoll, line 874: I'm fairly certain all other changes to table.dart outside of these two lines can be reverted to their previous state. Let's do that.

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -48,12 +50,23 @@ class AppSizeController {
...
+    }
+
+    final searchCondition = (TreemapNode n) => n == newRoot;
+    final int selectedRowIndex = newRoot.root.childCountToMatchingNode(
+      matchingNodeCondition: searchCondition,
+      includeCollapsedNodes: false,
+    );
+
+    analysisRoot.value = Selection(
+        node: newRoot, nodeIndex: selectedRowIndex, scrollIntoView: true);
```
kenzieschmoll, line 69: nit: trailing comma after `true`

packages/devtools_app/lib/src/ui/utils.dart
```diff
@@ -178,6 +178,24 @@ class _OffsetScrollbarState extends State<OffsetScrollbar> {
   }
 }
 
+void scrollToPosition(
+  ScrollController scrollController,
+  double position,
+) {
+  //TODO I would add padding here so if a row is half out of screen it will still jump
+  final scrollStart = scrollController.offset;
+  final scrollEnd = scrollStart + scrollController.position.extentInside;
```
kenzieschmoll, line 187: try using scrollController.position.maxScrollExtent instead

kenzieschmoll, line 187: nit: maybe rename to visibleStart and visibleEnd to indicate that these values are not the min / max scroll extents, but rather the visible range.

Bonus - use the Range class in DevTools so that you can use Range.contains to check if the position is inside the visibleRange

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -48,12 +50,23 @@ class AppSizeController {
...
   void changeAnalysisRoot(TreemapNode? newRoot) {
-    _analysisRoot.value = newRoot;
-    if (newRoot == null) return;
+    if (newRoot == null) {
+      analysisRoot.value = Selection<TreemapNode>();
+      return;
+    }
+
+    final searchCondition = (TreemapNode n) => n == newRoot;
+    final int selectedRowIndex = newRoot.root.childCountToMatchingNode(
```
elliette, line 63: I'm pretty sure `newRoot.root` can be changed to just `newRoot` (`newRoot` should be the root node and therefore shouldn't have any `parent` nodes)

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -48,12 +50,23 @@ class AppSizeController {
   /// The node set as the analysis tab root.
   ///
   /// Used to build the treemap and the tree table for the analysis tab.
-  ValueListenable<TreemapNode?> get analysisRoot => _analysisRoot;
-  final _analysisRoot = ValueNotifier<TreemapNode?>(null);
+  final analysisRoot =
```
elliette, line 53: Let's switch this back so that the `ValueNotifier` is still a private variable (begins with an underscore) and the `ValueListenable` is public (no underscore)

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
carolynqu: I believe newRoot is a node that is being passed as the selected node and not the root node (it's always changing).

elliette: Ah got it - let's maybe rename to `newAnalysisRoot` for clarity

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
carolynqu: SelectionNotifier takes a ValueNotifier so will be keeping analysisRoot as a ValueNotifier

carolynqu:
@polina-c This has some changes to the logic of flatTable scrolling that I see memory-heap-tree-view may use. Wanted to make sure that it doesn't break anything for memory. Let me know if any changes need to be made.

packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
```diff
@@ -42,16 +49,24 @@ class AppSizeAnalysisTable extends StatelessWidget {
...
   Widget build(BuildContext context) {
-    return TreeTable<TreemapNode>(
-      dataRoots: [rootNode],
-      columns: columns,
-      treeColumn: treeColumn,
-      keyFactory: (node) => PageStorageKey<String>(node.name),
-      sortColumn: sortColumn,
-      sortDirection: SortDirection.descending,
-      autoExpandRoots: true,
+    return ValueListenableBuilder<Selection<TreemapNode>>(
```
elliette, line 56: I'm guessing this `ValueListenableBuilder` can be removed now

packages/devtools_app/lib/src/ui/utils.dart
```diff
@@ -178,6 +182,27 @@ class _OffsetScrollbarState extends State<OffsetScrollbar> {
...
+  final extentVisible = Range(
+    scrollController.offset,
+    scrollController.offset + scrollController.position.extentInside,
+  );
+
+  if (!extentVisible.contains(position)) {
+    final positionToScrollTo = max(0.0, position - defaultRowHeight);
+
+    scrollController.animateTo(
+      //TODO: should be positionToScrollTo.clamp(0.0, scrollController.position.maxScrollExtent) but maxScrollExtent is not being updated
```
elliette, line 198: Open a bug and change TODO format to:

// TODO(your name): Description, link to issue

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -845,34 +843,28 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
...
-    cancelListeners();
+  void _addScrollListener(ValueNotifier<Selection<T>>? selectionNotifier) {
+    if (selectionNotifier != null) {
+      addAutoDisposeListener(selectionNotifier, () {
+        WidgetsBinding.instance.addPostFrameCallback((_) {
+          final Selection<T> selection = selectionNotifier.value;
+          if (selection.scrollIntoView) {
+            final int selectedDisplayRow = selection.nodeIndexCalculator != null
+                ? selection.nodeIndexCalculator!(selection.node!)
+                : selection.nodeIndex!;
```
elliette, line 859: Instead of null-asserting everywhere, let's pull these out into some variables, and only scroll if we have a node and a way of getting the node index:

```
if (scrollIntoView && node != null && (nodeIndexCalculator != null || nodeIndex != null)) {
}
```

elliette (approved):
LGTM with some suggestions!

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -8,19 +8,13 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart' hide TableRow;
 import 'package:flutter/services.dart';
 
-import '../primitives/auto_dispose_mixin.dart';
+import '../../devtools_app.dart';
```
kenzieschmoll, line 11: import libraries directly instead of importing `devtools_app.dart`, which exports a bunch of libraries in devtools

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -322,11 +316,13 @@ class Selection<T> {
   Selection({
     this.node,
     this.nodeIndex,
+    this.nodeIndexCalculator,
     this.scrollIntoView = false,
   });
 
   final T? node;
   final int? nodeIndex;
+  final int Function(T)? nodeIndexCalculator;
```
kenzieschmoll, line 325: do we need both of these? can you add a comment explaining which will be used if they are both specified? should we assert that only one of them is non-null?

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -845,34 +842,32 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
...
     _initSearchListener();
   }
 
-  @override
-  void didUpdateWidget(_Table<T> oldWidget) {
-    super.didUpdateWidget(oldWidget);
+  void _addScrollListener(ValueNotifier<Selection<T>>? selectionNotifier) {
+    if (selectionNotifier != null) {
+      addAutoDisposeListener(selectionNotifier, () {
+        WidgetsBinding.instance.addPostFrameCallback((_) {
```
kenzieschmoll, line 852: instead of putting this entire thing in a post frame callback, I would add a parameter `scrollInPostFrameCallback` to `scrollToPosition`, and when true, wrap only the call to `scrollController.animateTo(...)` in a post frame callback. 

Otherwise, all these calculations will be delayed until a frame has been drawn, and we don't need that - we just want the scroll to take place after a frame has been drawn

packages/devtools_app/lib/src/ui/utils.dart
```diff
@@ -178,6 +182,27 @@ class _OffsetScrollbarState extends State<OffsetScrollbar> {
   }
 }
 
+void scrollToPosition(
```
kenzieschmoll, line 185: nit: rename to `maybeScrollToPosition` or `scrollIfNotVisible`. And then add dart doc explaining what this util method does.

packages/devtools_app/lib/src/shared/table.dart
carolynqu: I believe we need the calculation to be delayed since we need to calculate the nodeIndex after the parents have been expanded.

packages/devtools_app/lib/src/shared/table.dart
carolynqu: Just added a TODO to address that having both is probably not necessary. And added assertion so that only one will be specified.

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -318,9 +325,16 @@ class Selection<T> {
     this.nodeIndex,
     this.nodeIndexCalculator,
     this.scrollIntoView = false,
-  });
+  }) : assert(nodeIndex != null || nodeIndexCalculator != null);
```
kenzieschmoll, line 328: this assert will still be true even if both nodeIndex and nodeIndexCalculator are specified. Change to `assert(nodeIndex == null || nodeIndexCalculator == null)` so that it will fail when both `nodeIndex` is non-null and `nodeIndexCalculator` is non-null, but will not fail when both are null.

packages/devtools_app/lib/src/ui/utils.dart
```diff
@@ -182,7 +182,7 @@ class _OffsetScrollbarState extends State<OffsetScrollbar> {
   }
 }
 
-void scrollToPosition(
+void maybeScrollToPosition(
```
kenzieschmoll, line 185: add dart doc above this method signature. Dart doc is with three slashes `///` and should describe what the method does, as opposed to comments that can be used in the body of a method. e.g.

```
/// Scrolls to [position] if [position] is not already visible in the scroll view.
```

packages/devtools_app/lib/src/ui/utils.dart
```diff
@@ -191,6 +191,7 @@ void scrollToPosition(
     scrollController.offset + scrollController.position.extentInside,
   );
 
+  // Scroll to position if not in view already
```
kenzieschmoll, line 194: if you add dart doc above (see earlier comment), you can remove this as it will become redundant.

kenzieschmoll (approved):

PR #4242, opened 2022-07-08
Change formatting for file name to make it sortable by time and the time recognizable by human

polina-c:


packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
```diff
@@ -104,8 +105,7 @@ abstract class ExportController {
     DateTime? time,
   }) {
     time ??= DateTime.now();
-    final timestamp =
-        '${time.year}_${time.month}_${time.day}-${time.microsecondsSinceEpoch}';
+    final timestamp = DateFormat('yy-MM-dd_HH:mm:ss.SSS').format(time);
```
kenzieschmoll, line 108: what about MM-DD-YY ? I would think this would be the most human readable format

packages/devtools_app/test/shared/import_export_test.dart
```diff
@@ -10,6 +10,20 @@ import 'package:devtools_test/devtools_test.dart';
 import 'package:flutter_test/flutter_test.dart';
 
 void main() async {
+  test('Filename is sortable by time', () async {
```
kenzieschmoll, line 13: to test that these are sortable, maybe create few different file names with `generateFileName`, sort them, and then verify the sorting is as you'd expect

packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
polina-c: 1. Europe uses DD-MM-YY, so, if you do not know format upfront, you cannot read it
2. We want the files to be sortable by time

packages/devtools_app/test/shared/import_export_test.dart
kenzieschmoll: Is a commit missing? I'm not seeing the change to this test to sort multiple filenames

kenzieschmoll: Oh whoops - disregard, I see it now.

kenzieschmoll (approved):

PR #4243, opened 2022-07-08
Create MemoryChartPaneController

polina-c:
Rename ChartControllers to MemoryChartPaneController and move to chart to make the class a placeholder for chart-related business logic.


packages/devtools_app/lib/src/screens/memory/memory_screen.dart
```diff
@@ -57,7 +58,7 @@ class MemoryBodyState extends State<MemoryBody>
         ProvidedControllerMixin<MemoryController, MemoryBody> {
   MemoryController get memoryController => controller;
 
-  late ChartControllers _chartControllers;
+  late MemoryChartPaneController _chartControllers;
```
kenzieschmoll, line 61: rename the variable to be singular instead of plural now that the class name has the singular form

kenzieschmoll (approved):
one comment then lgtm

Issue #3950, opened 2022-04-01
Provide the InspectorController through package:provider


kenzieschmoll:
right now, a new instance of InspectorController is created each time the inspector screen is opened (including tab switches). We should clean up this tech debt and provide the controller through package:provider like we do for other controllers.

PR #4244, opened 2022-07-08
Provide the InspectorController through package:provider

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/3950.

this change results in the InspectorController only being created once (like the other controllers in DevTools), instead of being recreated on each time the inspector screen is created and destroyed.


packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -112,6 +76,17 @@ class InspectorController extends DisposableController
       details = null;
     }
 
+    await serviceManager.onServiceAvailable;
```
jacob314, line 79: we should check whether this controller has been disposed after this async gap and bail on the rest of the init in that case.

jacob314 (approved):
lgtm_dog2

PR #4248, opened 2022-07-13
Debugger bugs

migueltylee:
Adds resetOutline() method to programExplorerController, so that the nodes in the outline tree are collapsed and deselected.
This PR fixes the following issues:
- when switching isolates the scriptsHistory remains the same, thus when navigating back to a script that is not present in the currently selected isolate produces the error: 
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: getObject: (-32602) Invalid params...]
and then the correct script is not displayed.
Fix: clear the scriptsHistory when switching isolates.
- When switching through the scripts in the Scripts HistoryViewport or selecting a different file, the selected node in the outline never unselects. Then, when going back and selecting another node in that same outline, there will be more than one node selected. Fix: call the resetOutline method when switching through the scripts in the HistoryViewport. 
Note: I think this can also be fixed by deselecting the node inside the SelectNode method (before line 164), but I also think it could be good to reset the outline since the debugger doesn't scroll to the selected outline object position when navigating through the scripts.


packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
```diff
@@ -182,6 +182,17 @@ class ProgramExplorerController extends DisposableController
     }
   }
 
+  void resetOutline() {
```
bkonyi, line 185: Can you add a comment describing what this method does?

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -74,6 +74,7 @@ class DebuggerController extends DisposableController
     }
     if (currentScriptRef.value != null) {
       await programExplorerController.selectScriptNode(currentScriptRef.value);
+      programExplorerController.resetOutline();
```
kenzieschmoll, line 77: do we want to reset the outline each time selectScriptNode is called? If that's the case, I'm wondering if we should put a call to resetOutline inside of selectScriptNode

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
```diff
@@ -182,6 +182,17 @@ class ProgramExplorerController extends DisposableController
...
+    _outlineSelection.value = null;
+
+    for (VMServiceObjectNode node in _outlineNodes.value) {
+      node.collapseCascading();
+      node.unselect();
```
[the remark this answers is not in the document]
migueltylee: Thanks!

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
migueltylee: Sure!

kenzieschmoll (approved):

PR #4251, opened 2022-07-13
Package Directory Interface

CoderDake:
Add frontend interface for package directory behaviour

[Most recent demo
](https://threadit.app/thread/a1aicelwkn63m52qxkv3)


packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -174,6 +176,13 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
         ),
       ],
     );
+    preferences.inspector.customPubRootDirectories.addListener(() {
+      _refreshInspector();
+    });
+    serviceManager.isolateManager.isolates.addListener(() {
+      _refreshInspector();
+      preferences.inspector.loadCustomPubRootDirectories();
+    });
```
kenzieschmoll, line 185: In general, we should never be doing anything with listeners inside of `build`. This code here would result in a new listener being called each time we build the inspector (which would be quite often). 

Any code like this should be in `initState` and should be added with `addAutoDisposeListener` - other wise the listener will never be disposed and this will lead to memory leaks.

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -341,21 +350,39 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+                  as ValueNotifier<bool?>,
+              title: 'Enable hover inspection',
+              description:
+                  'Hovering over any widget displays its properties and values.',
+              gaItem: analytics_constants.inspectorHoverEvalMode,
+            ),
+            const SizedBox(height: denseSpacing),
+            ...dialogSubHeader(Theme.of(context), 'Package Directories'),
+            Text(
+              'widgets in these directories will show up in your summary tree',
```
kenzieschmoll, line 379: r/widgets/Widgets and add a period at the end.

We should also add something like `(e.g. /absolute/path/to/myPackage)` to demonstrate what the path should look like.

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -515,3 +542,31 @@ class ErrorNavigator extends StatelessWidget {
     onSelectError(newIndex);
   }
 }
+
+class PubRootDirectorySection extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    preferences.inspector.loadCustomPubRootDirectories();
```
kenzieschmoll, line 549: same, we should not be doing this from build

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -515,3 +542,31 @@ class ErrorNavigator extends StatelessWidget {
...
+    return ValueListenableBuilder<List<IsolateRef?>>(
+      valueListenable: serviceManager.isolateManager.isolates,
```
kenzieschmoll, line 551: same question wrt whether we should be listening to the selected isolate or to all isolates

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -515,3 +542,31 @@ class ErrorNavigator extends StatelessWidget {
...
+      builder: (context, value, child) {
```
kenzieschmoll, line 552: if we never use value, replace with `_`

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -515,3 +542,31 @@ class ErrorNavigator extends StatelessWidget {
...
+        return Container(
+          width: double.maxFinite,
```
kenzieschmoll, line 554: why do we need this

packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
```diff
@@ -442,6 +452,35 @@ class InspectorService extends InspectorServiceBase {
     );
   }
 
+  Future<void> _addPubRootDirectories(List<String> pubDirectories) {
+    // No need to call this from a breakpoint.
```
kenzieschmoll, line 456: i don't understand this comment - what does it mean?

packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
```diff
@@ -442,6 +452,35 @@ class InspectorService extends InspectorServiceBase {
...
+    return invokeServiceMethodDaemonNoGroupArgs(
+      'removePubRootDirectories',
+      pubDirectories,
+    );
+  }
+
+  Future<List<String>?> getPubRootDirectories() async {
+    assert(useDaemonApi);
+    final response = await invokeServiceMethodDaemonNoGroupArgs(
+      'getPubRootDirectories',
```
kenzieschmoll, line 476: these will all fail on the bots if we do not bump DevTools to a version of Flutter that your flutter change is included in

packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
```diff
@@ -433,6 +433,16 @@ class InspectorService extends InspectorServiceBase {
     await _onRootDirectoriesChanged(rootDirectories);
```
kenzieschmoll, line 433: remove setPubRootDirectories and _setPubRootDirectories methods since they will no longer be in use

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -11,20 +11,14 @@ import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:vm_service/vm_service.dart';
 
+import '../../devtools_app.dart';
```
kenzieschmoll, line 14: import libraries directly instead of importing devtools_app.dart

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -254,6 +248,166 @@ class ClearButton extends IconLabelButton {
...
+    this.onEntryRemoved,
+    this.onRefresh,
+  });
+
+  final ValueListenable<List<String>> entries;
+  final ValueListenable<bool>? isRefreshing;
+  final Function(String)? onEntryAdded;
+  final Function(String)? onEntryRemoved;
+  final Function()? onRefresh;
+  final String label;
```
kenzieschmoll, line 266: nit: rename to textFieldLabel so that it is clear from the name what this label is used for

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -254,6 +248,166 @@ class ClearButton extends IconLabelButton {
...
+  @override
+  State<StatefulWidget> createState() => _EditableListState();
+}
+
+class _EditableListState extends State<EditableList> {
+  @override
+  void initState() {
+    super.initState();
+    textFieldController = TextEditingController();
+    widget.isRefreshing?.addListener(_stateRefresher);
```
kenzieschmoll, line 277: use addAutoDisposeListener (see AutoDisposeMixin). Then you don't have to manually dispose the listener. Another option, is to just use the DualValueListenableBuilder widget to listen to both `entries` and `isRefreshing` and then you don't have to manage adding listeners at all

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -254,6 +248,166 @@ class ClearButton extends IconLabelButton {
...
+      widget.onEntryAdded!(value);
+    }
+    textFieldFocusNode.requestFocus();
+  }
+
+  Widget _removeDirectoryButton(VoidCallback onPressed) {
+    return Container(
+      height: defaultIconSize + denseSpacing,
+      child: inputDecorationSuffixButton(Icons.delete, onPressed),
+    );
```
kenzieschmoll, line 308: why not just use IconButton?

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -254,6 +248,166 @@ class ClearButton extends IconLabelButton {
...
+  }
+
+  @override
+  Widget build(BuildContext context) {
```
kenzieschmoll, line 312: this build method is too long - let's break up into multiple widgets. It would also be reasonable to put this widget into its own file shared/editable_list.dart since it is relatively large.

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -64,11 +68,28 @@ class PreferencesController {
...
+  final _customPubRootDirectoriesAreBusy = ValueNotifier<bool>(false);
+  final _busyCounter = ValueNotifier<int>(0);
   static const _hoverEvalModeStorageId = 'inspector.hoverEvalMode';
+  static const _customPubRootDirectoriesStorageId =
+      'inspector.customPubRootDirectories';
 
   Future<void> init() async {
+    await initHoverEvalMode();
+    _initCustomPubRootListeners();
+    setGlobal(InspectorPreferencesController, this);
```
kenzieschmoll, line 89: why do we need InspectorPreferencesController to be a global when it is accessible from the preferences global? i think we can get rid of this and remove InspectorPreferencesController from globals

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -82,8 +103,84 @@ class InspectorPreferencesController {
...
 
-    setGlobal(InspectorPreferencesController, this);
+  void _initCustomPubRootListeners() {
+    _customPubRootDirectories.addListener(() {
+      storage.setValue(
+        _customPubRootDirectoriesStorageId,
+        jsonEncode(_customPubRootDirectories.value),
+      );
+    });
+    _busyCounter.addListener(() {
```
kenzieschmoll, line 115: these listeners will need to be disposed. Look at the AutoDisposeControllerMixin

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -174,6 +176,13 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
         ),
       ],
     );
+    preferences.inspector.customPubRootDirectories.addListener(() {
+      _refreshInspector();
+    });
+    serviceManager.isolateManager.isolates.addListener(() {
```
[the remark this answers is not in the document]
CoderDake: Great question, I found that just listening to the main isolate wasn't consistent enough to catch hot restarts. For some reason listening to all isolates worked better for me.

I can retest this with just the main isolate though :)

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
CoderDake: great catch on this!

packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
CoderDake: I have no idea either. It looks like i just copy and pasted it from the original.
I'll delete them :)

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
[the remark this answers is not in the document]
CoderDake: with this new implementation it looks like listening to just the mainIsolate is sufficient. So i have switched it :)

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
CoderDake: @kenzieschmoll it looks like we have a lint that forces a BuildContext to be named `context` should I override that lint when we don't use it?

kenzieschmoll: we can leave context, but you can change the other two params if they aren't used:
`builder: (context, _, __) {`

packages/devtools_app/lib/src/shared/editable_list.dart
```diff
@@ -0,0 +1,209 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return Flexible(
+      child: RoundedOutlinedBorder(
+        child: Scrollbar(
+          child: ListView.builder(
+            itemCount: widget.entries.value.length,
+            itemBuilder: (context, index) {
+              return Padding(
```
kenzieschmoll, line 179: store `widget.entries.value[index]` before the return here so that you can use it throughout.

packages/devtools_app/lib/src/shared/editable_list.dart
```diff
@@ -0,0 +1,209 @@
...
+  }
+}
+
+class _EditableListContentView extends StatelessWidget {
+  const _EditableListContentView({
+    Key? key,
+    required this.widget,
+  }) : super(key: key);
+
+  final EditableList widget;
```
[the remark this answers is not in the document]
CoderDake: This is the parent widget. So I'll name it to parent.

packages/devtools_app/lib/src/shared/editable_list.dart
```diff
@@ -0,0 +1,209 @@
...
+}
+
+class _EditableListActionBar extends StatelessWidget {
+  const _EditableListActionBar({
+    Key? key,
+    required this.textFieldFocusNode,
+    required this.textFieldController,
+    required this.widget,
+  }) : super(key: key);
+  void _addNewPubRootDirecory() {
```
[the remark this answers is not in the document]
CoderDake: :facepalm: ty for noticing that I accidentally left the name with pubrootdir

packages/devtools_app/test/shared/editable_list_test.dart
```diff
@@ -0,0 +1,135 @@
...
+        onEntryAdded: (e) {
+          entryThatWasRequestedForAdd = e;
+        },
+      );
+      await tester.pumpWidget(
+        wrap(widget),
+      );
+
+      final textField = find.byKey(widget.textFieldKey);
+      final addEntryButton = find.byKey(widget.addEntryButtonKey);
```
CoderDake, line 50: todo: make visiblefor testing classes instead of keys

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -18,25 +24,29 @@ class PreferencesController {
   ValueListenable<bool> get denseModeEnabled => _denseMode;
 
   InspectorPreferencesController get inspector => _inspector;
-  final _inspector = InspectorPreferencesController();
+
+  @visibleForTesting
+  set inspector(v) => _inspector = v;
```
CoderDake, line 29:
```suggestion
```
TODO: Remove this

packages/devtools_app/lib/src/shared/editable_list.dart
```diff
@@ -0,0 +1,284 @@
+import 'package:flutter/foundation.dart';
```
kenzieschmoll, line 1: add copyright

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -326,21 +336,42 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
-                as ValueNotifier<bool?>,
-            title: 'Enable hover inspection',
-            description:
-                'Hovering over any widget displays its properties and values.',
-            gaItem: analytics_constants.inspectorHoverEvalMode,
-          ),
-        ],
+      content: Container(
+        width: defaultDialogWidth,
+        height: scaleByFontFactor(400.0),
```
kenzieschmoll, line 344: pull this into a const

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -326,21 +336,42 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
...
+            Text(
+              'Widgets in these directories will show up in your summary tree.',
+              style: theme.subtleTextStyle,
+            ),
+            Text(
+              '(e.g. /absolute/path/to/myPackage)',
+              style: theme.subtleTextStyle,
+            ),
+            const SizedBox(height: denseSpacing),
+            PubRootDirectorySection(),
```
kenzieschmoll, line 372: wrap this in an Expanded, and remove the 200.0 height below

packages/devtools_app/lib/src/shared/editable_list.dart
```diff
@@ -0,0 +1,272 @@
...
+          children: [
+            EditableListActionBar(
+              textFieldFocusNode: textFieldFocusNode,
+              textFieldController: textFieldController,
+              isRefreshing: widget.isRefreshing,
+              textFieldLabel: widget.textFieldLabel,
+              onEntryAdded: widget.onEntryAdded,
+              onRefresh: widget.onRefresh,
+            ),
+            Flexible(
```
kenzieschmoll, line 67: nit: use Expanded instead here - this makes it clear that the intent is for this widget to take the remaining space in the column

packages/devtools_app/lib/src/shared/editable_list.dart
```diff
@@ -0,0 +1,272 @@
...
+
+class _EditableListContentView extends StatelessWidget {
+  const _EditableListContentView({
+    Key? key,
+    required this.entries,
+    required this.onEntryRemoved,
+  }) : super(key: key);
+
+  final ValueListenable<List<String>> entries;
+  final Function(String)? onEntryRemoved;
```
kenzieschmoll, line 169: It seems like this should be non-nullable. If we cannot remove things from the list, that eliminates a big function that makes the list "editable"

packages/devtools_app/test/shared/editable_list_test.dart
```diff
@@ -0,0 +1,142 @@
...
+        },
+      );
+      await tester.pumpWidget(
+        wrap(widget),
+      );
+      final row = find.byType(EditableListRow);
+      final removeButton = find.descendant(
+        of: row,
+        matching: find.byType(EditableListRemoveDirectoryButton),
+      );
```
kenzieschmoll, line 106: you can also just find.byType(EditableListRemoveDirectoryButton) directly instead of finding the row first and then grabbing the descendant

packages/devtools_app/lib/src/shared/editable_list.dart
```diff
@@ -0,0 +1,272 @@
...
+  const EditableList({
+    required this.entries,
+    required this.textFieldLabel,
+    this.isRefreshing,
+    this.onEntryAdded,
+    this.onEntryRemoved,
+    this.onRefresh,
+  });
+
+  final ValueListenable<List<String>> entries;
```
kenzieschmoll, line 22: ListValueNotifier

packages/devtools_app/lib/src/shared/editable_list.dart
```diff
@@ -0,0 +1,272 @@
...
+  final ValueListenable<bool>? isRefreshing;
+  final Function(String)? onEntryAdded;
+  final Function(String)? onEntryRemoved;
```
kenzieschmoll, line 25: on second thought, should we allow the EditableList to modify the ListValueNotifier directly, and then we wouldn't need to pass these callbacks through? Then on add and delete the editable list code could just call entried.add and entries.remove

CoderDake: I'm not sure that would be a better approach. 

My first concern is that making the widget the source of truth, and then having to interpret whether something was added or removed after the list is updated, would make the contents of the list unreliable.
This would be further complicated by the fact that letting the list in the widget be a source of truth, means that we have 2 sources of truth that we need to keep in sync. The directories on the service and the directories in the widget. We would need to have extra logic to handle the cases when they get out of sync, or ways for us to know they get out of sync.

kenzieschmoll: > making the widget the source of truth

The widget wouldn't be the source of truth, the ListValueNotifier would be the source of truth. The only change here is that the widget has write access to the ListValueNotifier instead of having to go through the controller as a middleman.

packages/devtools_app/test/shared/editable_list_test.dart
```diff
@@ -100,10 +100,7 @@ void main() {
         wrap(widget),
       );
       final row = find.byType(EditableListRow);
```
kenzieschmoll, line 102: looks like we don't need this anymore?

packages/devtools_app/lib/src/shared/editable_list.dart
```diff
@@ -5,24 +5,28 @@
...
     required this.entries,
     required this.textFieldLabel,
     this.isRefreshing,
-    this.onEntryAdded,
-    this.onEntryRemoved,
+    Function(String)? onEntryAdded,
+    Function(String)? onEntryRemoved,
     this.onRefresh,
-  });
+  }) {
```
kenzieschmoll, line 21: can you do this in the constructor directly instead of the method body?

`)) : this.onEntryAdded = onEntryAdded ?? (entry) => entries.add(entry), ...`

packages/devtools_app/lib/src/shared/editable_list.dart
```diff
@@ -5,24 +5,28 @@
...
+    this.onEntryRemoved = onEntryRemoved ?? (entry) => entries.remove(entry);
+  }
 
-  final ValueListenable<List<String>> entries;
+  final ListValueNotifier<String> entries;
   final ValueListenable<bool>? isRefreshing;
-  final Function(String)? onEntryAdded;
-  final Function(String)? onEntryRemoved;
+  late final Function(String)? onEntryAdded;
+  late final Function(String)? onEntryRemoved;
```
kenzieschmoll, line 29: these can be non-nullable now since there are default values (I guess if we are able to initialize in the constructor directly (see above)

kenzieschmoll (approved):
lgtm with a couple more comments

packages/devtools_app/lib/src/shared/editable_list.dart
CoderDake: This took a while for Ben and I to work out properly, this was the only way we could find to do this while maintaining `final` on the callbacks.
The reason it is tricky is because we are accessing entries in the callback body

packages/devtools_app/lib/src/shared/editable_list.dart
CoderDake: we chatted about this on Video chat and came a compromise where default values will be added on onAdded and onRemoved.
That way this common widget can be more easily dropped into place and interacted with, when there aren't complex async interactions for add and remove

packages/devtools_app/lib/src/shared/editable_list.dart
CoderDake: kk so i gave it another whirl and it turns out that the right incantation of parens let me move it into the constructor initializer list :)

CoderDake:
While doing a quick test I noticed that the "Widget Details" tab doesn't refresh when the node no longer exists.
This may not be a huge issue for now, so I've logged this issue as a follow up
https://github.com/flutter/devtools/issues/4279

kenzieschmoll (approved):

PR #4260, opened 2022-07-15
Show/edit list of snapshots.

polina-c:
https://user-images.githubusercontent.com/12115586/179445738-c234940c-6189-4da8-a37f-064ce0ecfdeb.mov

The diffing feature mimics UI of Chrome DevTools memory diffing. To check it:
1. Right click any page in chrome and select Inspect > Memory.

### DIFF
--- a/packages/devtools_app/lib/src/primitives/utils.dart
+++ b/packages/devtools_app/lib/src/primitives/utils.dart
@@ -1242,6 +1242,15 @@ class ListValueNotifier<T> extends ChangeNotifier
     _listChanged();
     return true;
   }
+
+  /// Removes a range of elements from the list.
+  ///
+  /// https://api.flutter.dev/flutter/dart-core/List/removeRange.html
+  void removeRange(int start, int end) {
+    _rawList = _rawList.toList();
+    _rawList.removeRange(start, end);
+    _listChanged();
+  }
 }
 
 /// Wrapper for a list that prevents any modification of the list's content.
--- a/packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
+++ b/packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
@@ -30,8 +30,8 @@ class _DiffPaneController {
   final selectedIndex = ValueNotifier<int>(0);
 
   /// If true, some process is going on.
-  final _isProcessing = ValueNotifier<bool>(false);
   ValueListenable<bool> get isProcessing => _isProcessing;
+  final _isProcessing = ValueNotifier<bool>(false);
 
   DiffListItem get selected => snapshots.value[selectedIndex.value];
 
@@ -183,7 +183,7 @@ class _SnapshotListTitle extends StatelessWidget {
               child: Text('Snapshots', overflow: TextOverflow.ellipsis),
             ),
             const SizedBox(width: denseRowSpacing),
-            const Text('\u{24D8}'),
+            const Text('ⓘ'),
             const SizedBox(width: denseRowSpacing),
           ],
           if (isProcessing) ...[
--- a/packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
+++ b/packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
@@ -143,8 +143,26 @@ class _SnapshotListTitle extends StatelessWidget {
       children: [
         _SelectionBox(selected: selected),
         const SizedBox(width: denseRowSpacing),
+        if (item.isProcessing) const _ProgressIndicator(),
+        if (item.isProcessing) const SizedBox(width: denseRowSpacing),
         Expanded(child: Text(itemLocal.name, overflow: TextOverflow.ellipsis)),
       ],
+    );
+  }
+}
+
+class _ProgressIndicator extends StatelessWidget {
+  const _ProgressIndicator({Key? key}) : super(key: key);
+  static const _progressIndicatorSize = 16.0;
+
+  @override
+  Widget build(BuildContext context) {
+    return SizedBox(
+      width: _progressIndicatorSize,
+      height: _progressIndicatorSize,
+      child: CircularProgressIndicator(
+        color: Theme.of(context).textTheme.bodyText1?.color,
+      ),
     );
   }
 }
@@ -194,20 +212,21 @@ class _ControlPane extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return ValueListenableBuilder<bool>(
-        valueListenable: controller.isProcessing,
-        builder: (_, isProcessing, __) => Row(
-              children: [
-                ToolbarAction(
-                  icon: Icons.fiber_manual_record,
-                  tooltip: 'Take heap snapshot',
-                  onPressed: isProcessing ? null : controller.takeSnapshot,
-                ),
-                ToolbarAction(
-                  icon: Icons.block,
-                  tooltip: 'Clear all snapshots',
-                  onPressed: isProcessing ? null : controller.clearSnapshots,
-                )
-              ],
-            ));
+      valueListenable: controller.isProcessing,
+      builder: (_, isProcessing, __) => Row(
+        children: [
+          ToolbarAction(
+            icon: Icons.fiber_manual_record,
+            tooltip: 'Take heap snapshot',
+            onPressed: isProcessing ? null : controller.takeSnapshot,
+          ),
+          ToolbarAction(
+            icon: Icons.block,
+            tooltip: 'Clear all snapshots',
+            onPressed: isProcessing ? null : controller.clearSnapshots,
+          )
+        ],
+      ),
+    );
   }
 }
--- a/packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
+++ b/packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
@@ -1,0 +1,21 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/material.dart';
+
+const shouldShowDiffPane = false;
+
+class DiffPane extends StatefulWidget {
+  const DiffPane({Key? key}) : super(key: key);
+
+  @override
+  State<DiffPane> createState() => _DiffPaneState();
+}
+
+class _DiffPaneState extends State<DiffPane> {
+  @override
+  Widget build(BuildContext context) {
+    return const Text('hello, I am diff pane');
+  }
+}
--- a/packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
+++ b/packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
@@ -167,7 +167,6 @@ class _SnapshotListTitle extends StatelessWidget {
       valueListenable: theItem.isProcessing,
       builder: (_, isProcessing, __) => Row(
         children: [
-          //_SelectionBox(selected: selected),
           const SizedBox(width: denseRowSpacing),
           if (theItem is SnapshotListItem)
             Expanded(
@@ -204,28 +203,6 @@ class _ProgressIndicator extends StatelessWidget {
     );
   }
 }
-
-// /// Blue or transparent square, to mark selected item.
-// class _SelectionBox extends StatelessWidget {
-//   const _SelectionBox({Key? key, required this.selected}) : super(key: key);
-//   static const _boxSize = Size(3.0, 40.0);
-//   final bool selected;
-//
-//   @override
-//   Widget build(BuildContext context) {
-//     return SizedBox(
-//       width: _boxSize.width,
-//       height: _boxSize.height,
-//       child: selected
-//           ? const DecoratedBox(
-//               decoration: BoxDecoration(
-//                 color: defaultSelectionColor,
-//               ),
-//             )
-//           : null,
-//     );
-//   }
-// }
 
 class _SnapshotListContent extends StatelessWidget {
   const _SnapshotListContent({Key? key, required this.item}) : super(key: key);
--- a/packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
+++ b/packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
@@ -3,6 +3,7 @@
 // found in the LICENSE file.
 
 import 'package:collection/collection.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 
 import '../../../../primitives/utils.dart';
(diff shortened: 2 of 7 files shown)


packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -4,9 +4,16 @@
...
+import '../../../../ui/colors.dart';
+import '../../primitives/memory_utils.dart';
+import 'model.dart';
+
+/// While this pane is under construction, we do not want our users to see it.
+///
 /// Flip this flag locally to test the pane and flip back before checking in.
-const shouldShowDiffPane = false;
+/// TODO: before removing this flag add widget/golden testing for the diff pane.
+const shouldShowDiffPane = true;
```
polina-c, line 16: I will flip the flag before submitting. I have test to guard it.

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -15,9 +23,270 @@ class DiffPane extends StatefulWidget {
...
+          ),
+        ],
+      ),
+    );
+  }
+}
+
+class _ProgressIndicator extends StatelessWidget {
+  const _ProgressIndicator({Key? key}) : super(key: key);
+  static const _progressIndicatorSize = 16.0;
```
CoderDake, line 188: does this need to be scaled for screen density?

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -15,9 +23,270 @@ class DiffPane extends StatefulWidget {
...
+  }
+
+  Future<void> clearSnapshots() async {
+    snapshots.removeRange(1, snapshots.length);
+    index.value = 0;
+    notifyListeners();
+  }
+
+  int _nextNameNumber() {
+    final numbers = snapshots.map((e) => e.nameNumber);
```
CoderDake, line 73: To avoid iterating through the list twice, to make a unique name, we could perhaps use uuid.v4()
https://pub.dev/packages/uuid

CoderDake: Alternatively, It looks like flutter has GlobalKey which would be unique across the entire app.
https://api.flutter.dev/flutter/widgets/GlobalKey-class.html

CoderDake (commented):
Looks good! I think the header could probably be dark like our other tables, that way it will be a bit more distinct from the contents below.

bkonyi:
We should make use of `OutlineDecoration` and `VerticalDivider` to highlight sections of the widgets. Maybe also consider using a `Card` for the list of snapshots + the information box.

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
polina-c: May be. I did not do this before. Can you point me to example, please?

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
polina-c: The numbers are visible, so we want the numbers to be user friendly, i,e. 1, 2, 3, 4....

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
CoderDake: I've notices this density being used in certain places
https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/inspector/inspector_breadcrumbs.dart#L36
I've also seen it used in conjunction with scaleByFontFactor
https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/inspector/inspector_tree.dart

I'm not really sure when or how to use these tbh but since a progress indicator might be something we want to be well proportioned, I figured it would be worth bringing it up.
Perhaps @kenzieschmoll could school us?

kenzieschmoll:
+1 to use OutlineDecoration around each item in splitter (list of snapshots and content view)

kenzieschmoll:
For selection indicator, we should be consistent with other lists in DevTools that have a selection indication. Why not highlight the whole row like we do in other screens?
Screen Shot 2022-07-18 at 1 14 09 PM
Screen Shot 2022-07-18 at 1 13 04 PM

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -2,11 +2,19 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
 
-/// Wile this pane is under construction, we do not want our users to see it.
+import '../../../../../devtools_app.dart';
```
kenzieschmoll, line 8: remove this import

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -15,9 +23,270 @@ class DiffPane extends StatefulWidget {
   State<DiffPane> createState() => _DiffPaneState();
 }
 
+/// Will notify if count of items or current index changed.
+class _DiffPaneController with ChangeNotifier {
```
kenzieschmoll, line 27: what is the benefit of mixing in ChangeNotifier instead of just having the controller class have ValueNotifier members? I'm not seeing anywhere where we call `diffPaneController.listen`, so what value is this adding?

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -15,9 +23,270 @@ class DiffPane extends StatefulWidget {
...
+    index.addListener(() => notifyListeners());
+  }
+
+  final scrollController = ScrollController();
+
+  /// The list contains one item that show information and all others
+  /// are snapshots.
+  final snapshots = <DiffListItem>[InformationListItem()];
+
+  final index = ValueNotifier<int>(0);
```
kenzieschmoll, line 38: nit: rename to `selectedIndex` so that it is clear what `index` we are referring to

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
kenzieschmoll: We can just use a `ListValueNotifier` for `snapshots` and then we are already using a `ValueNotifier` for `index` that other parts of the code can listen to. This will reduce an added layer of notifier complexity, and would be consistent with the rest of the DevTools codebase as a whole.

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -15,9 +23,270 @@ class DiffPane extends StatefulWidget {
...
+    isProcessing.value = false;
+  }
+
+  Future<void> clearSnapshots() async {
+    snapshots.removeRange(1, snapshots.length);
+    index.value = 0;
+    notifyListeners();
+  }
+
+  int _nextNameNumber() {
```
kenzieschmoll, line 72: nit: replace all instances of 'nameNumber' with 'displayNumber' for more clarity

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -15,9 +23,270 @@ class DiffPane extends StatefulWidget {
...
+      shrinkWrap: true,
+      itemCount: controller.snapshots.length,
+      itemBuilder: (context, index) {
+        return ListTile(
+          title: _SnapshotListTitle(
+            item: controller.snapshots[index],
+            selected: index == controller.index.value,
+          ),
+          onTap: () => controller.index.value = index,
+          key: GlobalObjectKey(controller.snapshots[index]),
```
kenzieschmoll, line 151: why do we need a global key? we shouldn't need a key here at all

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
kenzieschmoll: We use `scaleByFontFactor` to make DevTools respect the IDE / browser font size. We should use it here if we want it to scale. Here is an example of where we use `smallProgressSize`, which is a scaled value.

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -15,9 +23,270 @@ class DiffPane extends StatefulWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder<bool>(
+      valueListenable: controller.isProcessing,
+      builder: (_, isProcessing, __) {
+        final showTakeSnapshot = !isProcessing;
+        final showClearAll = !isProcessing & controller.hasSnapshots;
+        return Row(
+          children: [
+            ToolbarAction(
```
kenzieschmoll, line 254: since we have room for it, maybe we should add the text 'Take Heap Snapshot' or 'Take Snapshot' to make it more clear?

polina-c:
> Looks good! I think the header could probably be dark like our other tables, that way it will be a bit more distinct from the contents below.

I tried it. I feel it add noise to UI. It is actually not a header, but a special list item. If I highlight it, it looks as 'not clickable' (or clickable to sort the items) while it is clickable.
Chrome devtools does not make this item outstanding.

Screen Shot 2022-07-18 at 3 00 22 PM

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -15,9 +23,270 @@ class DiffPane extends StatefulWidget {
...
+  final snapshots = <DiffListItem>[InformationListItem()];
+
+  final index = ValueNotifier<int>(0);
+
+  /// If true, some process is going on.
+  final isProcessing = ValueNotifier<bool>(false);
+
+  DiffListItem get selected => snapshots[index.value];
+
+  bool get hasSnapshots => snapshots.length > 1;
```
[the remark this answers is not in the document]
polina-c: Because first item is information

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
polina-c: I feel it will make the UI more noisy, i.e. it will be harder to focus on important things.
The informational list item on the top will explain what to click to record snapshot (the icon will be included). Also, there is the tooltip.
User will need to learn it just once.

packages/devtools_app/lib/src/screens/memory/panes/diff/model.dart
```diff
@@ -0,0 +1,46 @@
...
+
+  /// Number, that, if shown in name, should be unique in the list.
+  ///
+  /// If the number is not shown, it should be 0.
+  int get nameNumber;
+
+  bool isProcessing = false;
+}
+
+class InformationListItem extends DiffListItem {
```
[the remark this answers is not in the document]
polina-c: discussed offline - we are mimicking Chrome DevTools, where information is one of the list items

polina-c:
> For selection indicator, we should be consistent with other lists in DevTools that have a selection indication. Why not highlight the whole row like we do in other screens? 


Implemented, but got minor issues. I am not sure what to do about them: 

https://user-images.githubusercontent.com/12115586/179645187-e020e348-eb8a-4a22-b895-d81ae384a950.mov

packages/devtools_app/lib/src/primitives/utils.dart
```diff
@@ -1242,6 +1242,15 @@ class ListValueNotifier<T> extends ChangeNotifier
     _listChanged();
     return true;
   }
+
+  /// Removes a range of elements from the list.
+  ///
+  /// https://api.flutter.dev/flutter/dart-core/List/removeRange.html
+  void removeRange(int start, int end) {
```
kenzieschmoll, line 1249: can we add a test for this in the existing group of ListValueNotifier tests in utils_test.dart?

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -16,139 +20,135 @@ import 'model.dart';
...
-  final snapshots = <DiffListItem>[InformationListItem()];
+  final snapshots = ListValueNotifier(<DiffListItem>[InformationListItem()]);
 
-  final index = ValueNotifier<int>(0);
+  final selectedIndex = ValueNotifier<int>(0);
 
   /// If true, some process is going on.
-  final isProcessing = ValueNotifier<bool>(false);
+  final _isProcessing = ValueNotifier<bool>(false);
+  ValueListenable<bool> get isProcessing => _isProcessing;
```
kenzieschmoll, line 34: style nit: order getters before the private value.

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -161,22 +161,35 @@ class _SnapshotListTitle extends StatelessWidget {
...
+          if (theItem is SnapshotListItem)
+            Expanded(
+              child: Text(theItem.name, overflow: TextOverflow.ellipsis),
+            ),
+          if (theItem is InformationListItem) ...[
+            const Expanded(
+              child: Text('Snapshots', overflow: TextOverflow.ellipsis),
+            ),
+            const SizedBox(width: denseRowSpacing),
+            const Text('\u{24D8}'),
```
kenzieschmoll, line 186: why are we using this instead of an Icon?

packages/devtools_app/lib/src/screens/memory/panes/diff/model.dart
```diff
@@ -2,45 +2,43 @@
...
+abstract class DiffListItem {
   /// Number, that, if shown in name, should be unique in the list.
   ///
   /// If the number is not shown, it should be 0.
-  int get nameNumber;
+  int get displayNumber;
 
-  bool isProcessing = false;
+  final _isProcessing = ValueNotifier<bool>(false);
+  ValueListenable<bool> get isProcessing => _isProcessing;
```
kenzieschmoll, line 15: style nit: order getter before private var

packages/devtools_app/lib/src/screens/memory/panes/diff/model.dart
```diff
@@ -2,45 +2,43 @@
...
-      notifyListeners();
+  SnapshotListItem(
+    Future<HeapSnapshotGraph?> graphReceiver,
+    this.displayNumber,
+    this._isolateName,
+  ) {
+    _isProcessing.value = true;
+    graphReceiver.whenComplete(() async {
+      _isProcessing.value = false;
+      graph = await graphReceiver;
```
kenzieschmoll, line 32: should the order of these be switched? is `_isProcessing` really false before `await graphReceiver` completes?

kenzieschmoll (commented):
a few more comments

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
polina-c: Looks much nicer and less widgets.

packages/devtools_app/lib/src/screens/memory/panes/diff/model.dart
polina-c: It actually already completed and this await will return momentarily, but, yes, it looks strange. Switched.

kenzieschmoll (approved):

PR #4263, opened 2022-07-19
Refactor notifications and chart pane.

polina-c:
Set notificationsService to global to enable access from controllers.
Move subscription, that handles change of data source, from ChartPane to MemoryController.
Separate notifications logic and UI.


packages/devtools_app/lib/src/shared/file_import.dart
```diff
@@ -7,13 +7,9 @@ import 'dart:convert';
 import 'package:file_selector/file_selector.dart';
 import 'package:flutter/material.dart';
 
+import '../../devtools_app.dart';
```
kenzieschmoll, line 10: import libraries directly

packages/devtools_app/lib/src/shared/scaffold.dart
```diff
@@ -162,7 +161,7 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
     super.didChangeDependencies();
 
     _importController = ImportController(
-      Notifications.of(context)!,
+      // Notifications.of(context)!,
```
kenzieschmoll, line 164: delete this line

packages/devtools_app/lib/src/app.dart
```diff
@@ -131,6 +131,8 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
     });
 
     releaseNotesController = ReleaseNotesController();
+
+    setGlobal(NotificationService, Notifications.of(context));
```
kenzieschmoll, line 135: at this point, should all uses of `Notifications.of(context)` be replaced with `notificationService`?

polina-c: refactored and replaced

packages/devtools_app/lib/src/primitives/notifications.dart
```diff
@@ -0,0 +1,56 @@
...
+
+  final String text;
+  final List<Widget> actions;
+  final Duration duration;
+}
+
+abstract class NotificationService {
+  /// Pushes a notification [message].
+  bool push(String message);
+  bool smartPush(
```
kenzieschmoll, line 25: new line between these methods and please add dart doc for `smartPush`.

and maybe rename to `pushNotification` so that its effect is a little more clear from the name it is from `smartPush`

packages/devtools_app/lib/src/framework/notifications.dart
```diff
@@ -110,41 +138,30 @@ class NotificationsState extends State<_NotificationsProvider>
...
       );
       _overlayEntry?.markNeedsBuild();
     });
-    return true;
   }
 
   /// Dismisses all notifications with a matching message.
-  void dismiss(String message) {
+  void _dismiss(String message) {
+    print('dismiss invoked');
```
kenzieschmoll, line 155: remove print

packages/devtools_app/lib/src/framework/notifications.dart
```diff
@@ -110,41 +138,30 @@ class NotificationsState extends State<_NotificationsProvider>
...
     bool didDismiss = false;
     // Make a copy so we do not remove a notification from [_notifications]
     // while iterating over it.
-    final notifications = List.from(_notifications);
+    final notifications = List.from(_notifications).cast<_Notification>();
     for (final notification in notifications) {
-      if (notification.tooltip == message) {
+      if (notification.message.text == message) {
+        print('dismiss invoked - notification found');
```
kenzieschmoll, line 162: remove

packages/devtools_app/lib/src/framework/notifications.dart
```diff
@@ -7,84 +7,100 @@ import 'dart:async';
...
-class _NotificationsProvider extends StatefulWidget {
-  const _NotificationsProvider({Key? key, required this.child})
-      : super(key: key);
+class NotificationController implements NotificationService {
+  final toPush = ValueNotifier<NotificationMessage>(NotificationMessage(''));
+  final toDismiss = ValueNotifier<NotificationMessage>(NotificationMessage(''));
 
-  final Widget child;
+  @visibleForTesting
+  final inProcess = <NotificationMessage>[];
```
kenzieschmoll, line 48: naming nit: maybe 'activeNotifications` or `visibleNotifications` for clarity?

packages/devtools_app/lib/src/framework/notifications.dart
```diff
@@ -7,84 +7,100 @@ import 'dart:async';
...
-    return provider?.data;
-  }
 }
 
-class _NotificationsProvider extends StatefulWidget {
-  const _NotificationsProvider({Key? key, required this.child})
-      : super(key: key);
+class NotificationController implements NotificationService {
+  final toPush = ValueNotifier<NotificationMessage>(NotificationMessage(''));
+  final toDismiss = ValueNotifier<NotificationMessage>(NotificationMessage(''));
```
kenzieschmoll, line 45: new line between class members

packages/devtools_app/lib/src/framework/notifications.dart
```diff
@@ -7,84 +7,100 @@ import 'dart:async';
...
-    final provider =
-        context.dependOnInheritedWidgetOfExactType<_InheritedNotifications>();
-    return provider?.data;
-  }
 }
 
-class _NotificationsProvider extends StatefulWidget {
-  const _NotificationsProvider({Key? key, required this.child})
-      : super(key: key);
+class NotificationController implements NotificationService {
```
kenzieschmoll, line 43: it looks like this is the only class that implements NotificationsService. Let's just combine the two classes then, as there isn't any added value by breaking this up into an abstract class and an implementation class. It is also a bit confusing that we have to cast `notificationService as NotificationController` in multiple places.

packages/devtools_app/lib/src/framework/notifications.dart
```diff
@@ -7,84 +7,100 @@ import 'dart:async';
...
-/// will allow you to write unit tests for the controller that consumes it
-/// instead of widget tests.
-abstract class NotificationService {
-  /// Pushes a notification [message].
-  void push(String message);
-}
-
 /// Manager for notifications in the app.
 ///
 /// Must be inside of an [Overlay].
```
kenzieschmoll, line 23: is this comment accurate? the Overlay is the top level widget in this build method, and it doesn't look like the `NotificationsView` widget is not inside an Overlay where it is used.

I'd also recommend combining the `NotificationsView` and the `_Notifications` widget since this one is so minimal and that would eliminate some unnecessary boilerplate. Especially since the `_NotificationsState` build method just returns `return widget.child;`

packages/devtools_app/lib/src/framework/notifications.dart
```diff
@@ -110,41 +138,30 @@ class NotificationsState extends State<_NotificationsProvider>
...
 
   /// Dismisses all notifications with a matching message.
-  void dismiss(String message) {
+  void _dismiss(String message) {
+    print('dismiss invoked');
     bool didDismiss = false;
     // Make a copy so we do not remove a notification from [_notifications]
     // while iterating over it.
-    final notifications = List.from(_notifications);
+    final notifications = List.from(_notifications).cast<_Notification>();
```
kenzieschmoll, line 159: can you add the type to List.from to get rid of the cast: `List<_Notifications>.from(_notifications)`

packages/devtools_app/lib/src/primitives/notifications.dart
```diff
@@ -0,0 +1,56 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: what is the benefit of splitting the notifications up into multiple libraries? When working in the codebase, I would find it a bit confusing that all the notifications code is not in the same place. We don't do any deferred loading or anything like that, so it is not a problem for us if files in other areas of DevTools need to import something from `framework/`

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
```diff
@@ -239,6 +239,19 @@ class MemoryController extends DisposableController
   MemoryController() {
     memoryTimeline = MemoryTimeline(offline);
     memoryLog = MemoryLog(this);
+
+    // Update the chart when the memorySource changes.
+    addAutoDisposeListener(memorySourceNotifier, () async {
```
kenzieschmoll, line 244: did this code accidentally slip into this PR? it seems unrelated from the rest

packages/devtools_app/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 48;
+	objectVersion = 54;
```
kenzieschmoll, line 6: are these `macos/` changes intentional?

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
polina-c: Now, when notifications can be accesses from controllers, this code can be moved from UI(chart) to the controller.
Actually this is why I started this PR, and it resulted in refactoring for notifications.

I can separate it, if it is needed for something.

packages/devtools_app/lib/src/primitives/notifications.dart
polina-c: There are two parts: interface and implementation.
Interface is used by screens and sits in `shared`. 
Implementation sits in `framework` and is used only by App. 

It is much easier to understand interface when it is clearly separated. Now shared/notifications.dart contains three classes intended to be used by screens. It is easy to observe and understand them. If implementation code was here, it would be harder.

Renamed implementation to notifications_view.dart to make it clearer.

packages/devtools_app/lib/src/framework/notifications.dart
polina-c: I tried it and it did not work for some reasons. I will postpone it as I want to commit this PR and switch to next tasks. Left the comment.

packages/devtools_app/macos/Runner.xcodeproj/project.pbxproj
polina-c: Seems to be result of too fresh version of flutter. Reverted.

packages/devtools_app/lib/src/framework/notifications_view.dart
```diff
@@ -40,53 +38,12 @@ class NotificationsView extends StatelessWidget {
...
-    toPush.dispose();
-    toDismiss.dispose();
-  }
-}
-
+// TODO(polina-c): it seems it would be better to combine the NotificationsView
+//  and the _Notifications widget since this one is so minimal and that would
+//  eliminate some unnecessary boilerplate. Especially since the
+//  _NotificationsState build method just returns return widget.child.
+// However, the direct merge of these classes does not work for some reasons.
```
kenzieschmoll, line 45: Oh I see the issue. We are calling `Overlay.of(context)` from lifecycle methods in `_NotificationsState`, which would fail if the `Overlay` widget is defined in `_NotificationsState.build` because there would be no Overlay in the tree at the time `Overlay.of(context)` is called. You can disregard then.

packages/devtools_app/lib/src/framework/notifications_view.dart
```diff
@@ -100,7 +57,7 @@ class _NotificationsState extends State<_Notifications> with AutoDisposeMixin {
   OverlayEntry? _overlayEntry;
 
   final List<_Notification> _notifications = [];
-  late NotificationController controller;
+  late NotificationService controller;
```
kenzieschmoll, line 60: nit: rename to service or notificationsService

packages/devtools_app/lib/src/primitives/notifications.dart
kenzieschmoll: > The fact that implementation sits in the folder framework that is forbidden for screens, creates some level of protection.

A screen can still import something from `framework/` - there is no rule against that. Can you elaborate on the "protection" you are referring to? What are you concerned about guarding against?

polina-c: It is just mental agreement about content of 'framework'. No formal protection.
We put things intended for screens to 'sharing', and things not intended for screens, to 'framework'.
For example, if some screen decides to host NotificationsView, things will mess up and may not be caught by a test. 
And we always can change our mind about what we want to share and move this thing from 'framework' to 'shared'.

packages/devtools_app/lib/src/framework/notifications_view.dart
polina-c: removed it

packages/devtools_app/lib/src/framework/notifications_view.dart
polina-c: added this to the comment

kenzieschmoll (approved):

PR #4278, opened 2022-07-21
Deferred Info UI TreeTable Update

carolynqu:
If app is deferred (contains a dummy root), TreeTable will hide the dummy root and display the main app and its deferred chunks on the same level. 

**Deferred App TreeTable View:**
Screen Shot 2022-07-21 at 1 58 38 PM

If app is not deferred, TreeTable will display the root and its children. 

**Non-Deferred App TreeTable View:**
Screen Shot 2022-07-21 at 1 57 06 PM

### DIFF
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
@@ -14,7 +14,6 @@ import '../../charts/treemap.dart';
 import '../../primitives/utils.dart';
 import '../../shared/table.dart';
 import 'app_size_screen.dart';
-import 'app_size_table.dart';
 
 //Temporary feature flag for deferred loading.
 const deferredLoadingSupport = true;
@@ -55,7 +54,7 @@ class AppSizeController {
   /// Used to build the treemap and the tree table for the analysis tab.
   final analysisRoot = ValueNotifier<Selection<TreemapNode>>(Selection.empty());
 
-  late final bool isDeferredApp;
+  late bool isDeferredApp;
 
   void changeAnalysisRoot(TreemapNode? newAnalysisRoot) {
     if (newAnalysisRoot == null) {
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
@@ -233,9 +233,10 @@ class AppSizeController {
 
     changeAnalysisJsonFile(jsonFile);
 
-    // Set name for root node.
-    if (processedJson['n'] == 'Dummy Root') isDeferredApp = true;
-    // processedJson['n'] = 'Root';
+    // Set deferred app flag or set name for root node.
+    processedJson['n'] == 'Dummy Root'
+        ? isDeferredApp = true
+        : processedJson['n'] = 'Root';
 
     // Build a tree with [TreemapNode] from [processedJsonMap].
     final newRoot = generateTree(processedJson)!;
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
@@ -240,7 +240,7 @@ class AppSizeController {
     isDeferredApp =
         deferredLoadingSupportEnabled && processedJson['n'] == 'ArtificialRoot';
 
-    //Set root name.
+    // Set root name.
     processedJson['n'] = isDeferredApp ? 'Entire app' : 'Root';
 
     // Build a tree with [TreemapNode] from [processedJsonMap].
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
@@ -234,10 +234,14 @@ class AppSizeController {
 
     changeAnalysisJsonFile(jsonFile);
 
-    // Set deferred app flag or set name for root node.
-    processedJson['n'] == 'Dummy Root'
-        ? isDeferredApp = true
-        : processedJson['n'] = 'Root';
+    // Set deferred app flag and root name.
+    if (processedJson['n'] == 'Dummy Root') {
+      isDeferredApp = true;
+      processedJson['n'] = 'Entire App';
+    } else {
+      isDeferredApp = false;
+      processedJson['n'] = 'Root';
+    }
 
     // Build a tree with [TreemapNode] from [processedJsonMap].
     final newRoot = generateTree(processedJson)!;
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
@@ -16,6 +16,9 @@ import '../../shared/table.dart';
 import 'app_size_screen.dart';
 import 'app_size_table.dart';
 
+//Temporary feature flag for deferred loading.
+const deferredLoadingSupport = true;
+
 enum DiffTreeType {
   increaseOnly,
   decreaseOnly,
@@ -234,8 +237,8 @@ class AppSizeController {
 
     changeAnalysisJsonFile(jsonFile);
 
-    // Set deferred app flag and root name.
-    if (processedJson['n'] == 'Dummy Root') {
+    // Set deferred app flag and root name if deferred app and support is on.
+    if (processedJson['n'] == 'Dummy Root' && deferredLoadingSupport) {
       isDeferredApp = true;
       processedJson['n'] = 'Entire App';
     } else {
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -230,7 +234,8 @@ class AppSizeController {
     changeAnalysisJsonFile(jsonFile);
 
     // Set name for root node.
-    processedJson['n'] = 'Root';
+    if (processedJson['n'] == 'Dummy Root') isDeferredApp = true;
+    // processedJson['n'] = 'Root';
```
elliette, line 238: I'm guessing we still want to rename this to 'Root' if it's not a deferred app

packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
```diff
@@ -13,19 +13,27 @@ import '../../ui/colors.dart';
 
 import 'app_size_controller.dart';
 
-const deferredLoadingSupport = false;
+//temporary feature flag for deferred loading
```
elliette, line 16: Nit: Space, capitalize, and add a period at the end of the comment

packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
```diff
@@ -13,19 +13,27 @@ import '../../ui/colors.dart';
...
 
 class AppSizeAnalysisTable extends StatelessWidget {
   factory AppSizeAnalysisTable({
     required TreemapNode rootNode,
     required AppSizeController controller,
   }) {
-    final treeColumn = _NameColumn(currentRootLevel: rootNode.level);
+    final treeColumn = _NameColumn(
+      currentRootLevel:
+          deferredLoadingSupport ? rootNode.children[0].level : rootNode.level,
```
elliette, line 26: For this (and below) we should be checking if the data we have is for a deferred app, not whether the feature is enabled (so `controller.isDeferredApp`);

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
elliette: Also we want to set `isDeferredApp` to false if we have no "dummy root". Let's also rename to `ArtificialRoot` instead of `Dummy`.

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -420,6 +420,8 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
     super.initState();
     _initData();
 
+    // print("data root ${dataRoots[0]} level ${dataRoots[0].index}");
```
elliette, line 423: Remove print

packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
elliette: Let's also rename this to `deferredLoadingSupportEnabled` for clarity

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -77,10 +79,12 @@ class AppSizeController {
 
   int nodeIndexCalculator(TreemapNode newAnalysisRoot) {
     final searchCondition = (TreemapNode n) => n == newAnalysisRoot;
-    return newAnalysisRoot.root.childCountToMatchingNode(
+    if (!newAnalysisRoot.root.isExpanded) newAnalysisRoot.root.expand();
+    final nodeIndex = newAnalysisRoot.root.childCountToMatchingNode(
       matchingNodeCondition: searchCondition,
       includeCollapsedNodes: false,
     );
+    return isDeferredApp ? nodeIndex - 1 : nodeIndex;
```
elliette, line 87: Let's also check if `deferredLoadingSupportEnabled` is true

elliette:
Can you add some demos to the PR description (importing deferred loading data vs normal app data)?

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
carolynqu: Should this check only be placed here if it's not present in all the other deferred loading support changes?

packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
```diff
@@ -13,19 +13,28 @@ import '../../ui/colors.dart';
 
 import 'app_size_controller.dart';
 
-const deferredLoadingSupport = false;
+//Temporary feature flag for deferred loading.
+const deferredLoadingSupport = true;
```
elliette: Let's move this to the `app_size_controller` file. That way, the `isDeferredApp` bool could return true if it is a `deferredApp` and the support is enabled. This way we don't have to check both everywhere.

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -15,6 +15,9 @@ import '../../primitives/utils.dart';
 import '../../shared/table.dart';
 import 'app_size_screen.dart';
 
+//Temporary feature flag for deferred loading.
```
elliette, line 18: Space between // and the comment

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -15,6 +15,9 @@ import '../../primitives/utils.dart';
...
+const deferredLoadingSupport = true;
```
elliette, line 19: Rename to `deferredLoadingSupportEnabled`

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -229,8 +236,14 @@ class AppSizeController {
 
     changeAnalysisJsonFile(jsonFile);
 
-    // Set name for root node.
-    processedJson['n'] = 'Root';
+    // Set deferred app flag and root name if deferred app and support is on.
+    if (processedJson['n'] == 'Dummy Root' && deferredLoadingSupport) {
```
elliette, line 240: Rename "Dummy Root" to "ArtificialRoot" (this is a better name for go/inclusive-code purposes)

elliette, line 240: This can be made more concise:

```
final isDeferredApp = deferredLoadingSupportEnabled && processedJson['n'] == 'ArtificialRoot`;
processedJson['n'] = isDeferredApp ? 'Entire app' : 'Root';
```

Also this way, instead of having to check `deferredLoadingSupport && isDeferredApp` everywhere, you only need to check `isDeferredApp`.

packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
```diff
@@ -13,19 +13,28 @@ import '../../ui/colors.dart';
 
 import 'app_size_controller.dart';
 
-const deferredLoadingSupport = false;
+//Temporary feature flag for deferred loading.
```
elliette, line 16: Remove comment and variable, we are now using the one in app_size_controller

packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
```diff
@@ -13,19 +13,28 @@ import '../../ui/colors.dart';
...
+bool get deferredLoadingSupport => true;
 
 class AppSizeAnalysisTable extends StatelessWidget {
   factory AppSizeAnalysisTable({
     required TreemapNode rootNode,
     required AppSizeController controller,
   }) {
-    final treeColumn = _NameColumn(currentRootLevel: rootNode.level);
+    final treeColumn = _NameColumn(
+      currentRootLevel: deferredLoadingSupport && controller.isDeferredApp
```
elliette, line 25: With the changes suggested above, we only need to check `controller.isDeferredApp`

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -15,6 +15,9 @@ import '../../primitives/utils.dart';
 import '../../shared/table.dart';
 import 'app_size_screen.dart';
 
+// Temporary feature flag for deferred loading.
+const deferredLoadingSupportEnabled = true;
```
elliette, line 19: Switch to false before submitting :)

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -229,8 +236,12 @@ class AppSizeController {
 
     changeAnalysisJsonFile(jsonFile);
 
-    // Set name for root node.
-    processedJson['n'] = 'Root';
+    // Set deferred app flag.
+    isDeferredApp =
+        deferredLoadingSupportEnabled && processedJson['n'] == 'ArtificialRoot';
+
+    //Set root name.
```
elliette, line 243: Nit: Space between // and comment

elliette (approved):
LGTM once the two open comments are addressed!

Issue #3561, opened 2022-01-04
Inspector tree search needs to scroll active search into view horizontally


kenzieschmoll:
When the widget tree pane is narrow, and active search match may not be in view horizontally. We should scroll to the active search match both vertically and horizontally:
![need-to-scroll-horizontally](https://user-images.githubusercontent.com/43759233/148128331-764af13f-7744-4592-a249-3adee68ba5f8.gif)

PR #4283, opened 2022-07-22
Fix inspector Auto Scroll

CoderDake:
Fixes https://github.com/flutter/devtools/issues/3561

This change fixes scrollToRect and related areas so that autoscrolling will always snap to the top left of the screen ( with 3 rows of padding).

If the focused node is already on the screen, no scrolling is done.


packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -918,39 +933,32 @@ class _InspectorTreeState extends State<InspectorTree>
   /// Placeholder viewport height to use if we don't yet know the real
   /// viewport height.
   static const _placeholderViewportHeight = 1000.0;
+  static const _placeholderViewportWidth = 1000.0;
```
kenzieschmoll, line 936: instead of creating a second var, just have one called `_placeholderViewportSize`

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -829,55 +830,69 @@ class _InspectorTreeState extends State<InspectorTree>
...
+    final rowIndex = treeController!.getRowIndex(rect.top);
+
+    final row = treeController!.getCachedRow(rowIndex);
+    final node = row?.node;
+    final textPreviewJson = node?.diagnostic?.json['textPreview'];
+    final textDescription = node?.diagnostic?.description;
+    final descriptions = <String>[
+      if (textDescription != null) textDescription,
+      if (textPreviewJson != null) '$textPreviewJson',
+    ];
```
kenzieschmoll, line 865: this seems fragile. What if we change how inspector nodes are displayed at some point? Is this display logic duplicated somewhere else that we could create a helper for?

CoderDake: Good call on this, it is definitely fragile.
I believe in order to achieve parity we would have to decouple this aspect.
https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/inspector/diagnostics.dart#L61

I can definitely look into what it would take to do so

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -331,8 +331,8 @@ class InspectorTreeController extends Object
     return (depth + 1) * columnWidth + horizontalPadding;
   }
 
-  double getRowY(int index) {
-    return rowHeight * index + verticalPadding;
+  double determineYForTopOfRow(int index) {
```
jacob314, line 334: nit: determine doesn't add any value. Maybe call this
rowY or rowYTop

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -761,6 +759,9 @@ class _InspectorTreeState extends State<InspectorTree>
 
   late DebuggerController _debuggerController;
 
+  /// When autoscrolling, the number of rows to pad the target location with.
+  final _scrollToTargetPaddingCount = 3;
```
jacob314, line 763: Add Row to this constant name as it is talking about a number of rows not a number of px.

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
jacob314: +1 on this being very fragile. Add a helper function + a unittest that fails if the predicted width and the actual width f rendering a diagnostic description go out of sync for some example input.

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -718,18 +718,6 @@ class InspectorController extends DisposableController
     // a node is selected, its properties will also be selected as by
     // convention properties are the first children of a node and properties
     // typically do not have children and are never expanded by default.
-    for (InspectorTreeNode child in node.children) {
```
jacob314, line 721: why was this logic removed? If the problem is it does bad things with too many properties, we can tweak the logic to only require some small number of children.

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -761,6 +762,9 @@ class _InspectorTreeState extends State<InspectorTree>
 
   late DebuggerController _debuggerController;
 
+  /// When autoscrolling, the number of rows to pad the target location with.
+  final _scrollRowPadding = 3;
```
jacob314, line 766: rather than hard coding, why not use min of 3 and the number of properties? Having all properties in view was nice.

packages/devtools_app/test/inspector/diagnostics_test.dart
```diff
@@ -109,5 +110,39 @@ void main() {
...
+        diagnosticWithoutService,
+      );
+      final nodeSize = tester.getSize(find.descendant(
+        of: find.byType(MouseRegion),
+        matching: find.byType(
+          RichText,
+        ),
+      ));
+      debugDumpApp();
+      expect(nodeSize.width, moreOrLessEquals(aproximateWidth, epsilon: 10.0));
```
jacob314, line 145: why do you need an epsilon? Can you fix the calculation so it is exactly right (at least until the icon size changes or something?

jacob314 (commented):
Looks good with a few questions.

packages/devtools_app/test/inspector/diagnostics_test.dart
CoderDake: We need an epsilon here because there isn't really a way for us to get this measurement, without changing the Remote Diagnostic Node widget builder.
The code the tries to find the width of the visible text on a node is an approximation since we don't have access to the widget at the time of checking for visibility.
The focus code and the build code could definitely be changed, but "close enough" seemed like a quick and easy way to get value for our users. :)

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
```diff
@@ -58,10 +59,56 @@ class DiagnosticsNodeDescription extends StatelessWidget {
...
+      }
+      spanWidth += calculateTextSpanWidth(TextSpan(text: name));
+    } else {
+      // When there is no name, an icon will be shown with the text spans.
+      spanWidth += approximateIconWidth;
+    }
+    return spanWidth;
+  }
+
+  static Iterable<TextSpan> buildDescriptionTextSpans(
```
kenzieschmoll, line 105: can we make these parameters named? easier to read at the call site

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
```diff
@@ -115,12 +162,18 @@ class DiagnosticsNodeDescription extends StatelessWidget {
     TextStyle textStyle,
     BuildContext context,
     ColorScheme colorScheme,
+    RemoteDiagnosticsNode? diagnostic,
```
kenzieschmoll, line 165: same here for all these params

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
```diff
@@ -58,10 +59,56 @@ class DiagnosticsNodeDescription extends StatelessWidget {
...
+      (sum, span) => sum + calculateTextSpanWidth(span),
+    );
+    String? name = diagnostic?.name;
+
+    // An Icon is approximately the width of 1 character
+    final approximateIconWidth = calculateTextSpanWidth(
+      const TextSpan(
+        text: 'A',
+      ),
+    );
```
kenzieschmoll, line 87: why not use the actual icon width here?

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
```diff
@@ -58,10 +59,56 @@ class DiagnosticsNodeDescription extends StatelessWidget {
...
+      // approximate name width.
+
+      if (diagnostic?.description != null) {
+        // If there is a description then a separator will show with the name.
+        name += ': ';
+      }
+      spanWidth += calculateTextSpanWidth(TextSpan(text: name));
+    } else {
+      // When there is no name, an icon will be shown with the text spans.
+      spanWidth += approximateIconWidth;
```
kenzieschmoll, line 100: if we only need `approximateIconWidth` in this else statement, initialize it here so we save a calculation in the if statement case

packages/devtools_app/test/inspector/diagnostics_test.dart
```diff
@@ -109,5 +111,121 @@ void main() {
...
+            .first
+            .widget as RichText;
+        final measuredTextWidth =
+            calculateTextSpanWidth(richTextWidget.text as TextSpan);
+
+        expect(
+          approximatedTextWidth,
+          moreOrLessEquals(
+            measuredTextWidth + measuredIconWidth,
+            epsilon: 5.0,
```
kenzieschmoll, line 225: put this epsilon into a const for use in the test

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
```diff
@@ -58,10 +59,56 @@ class DiagnosticsNodeDescription extends StatelessWidget {
...
 
-  Iterable<TextSpan> _buildDescriptionTextSpans(
+  /// Approximates the width of the elements inside a [RemoteDiagnosticsNode]
+  /// widget.
+  static double approximateNodeWidth(
+    RemoteDiagnosticsNode? diagnostic,
+  ) {
+    final spans = DiagnosticsNodeDescription.buildDescriptionTextSpans(
+      diagnostic?.description ?? '',
+      const TextStyle(),
```
kenzieschmoll, line 69: general concern for using a blank TextStyle is font sizing. Have you tried this with different sizes of text?

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
jacob314: Scrolling to the top left corner of the screen will be jarring for cases like going through a list of search results.
For example image stepping through search results and you get one that is slightly off the screen but now you scroll ALL the way up till it is at the top of the screen.
I would try to maintain the previous functionality which is pretty much industry standard for how you should scroll to bring some new content to be visible on the screen. Instead I would special case so we abandon making the parent visible when it is not possible as the list of properties is larger than the height of the screen.

packages/devtools_app/test/inspector/diagnostics_test.dart
jacob314: ok I'm fine with having this epsilon.

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
CoderDake: Good catch on this, it turns out it effects the sizing!

packages/devtools_app/lib/src/screens/inspector/diagnostics_node.dart
```diff
@@ -47,6 +47,10 @@ class RemoteDiagnosticsNode extends DiagnosticableTree {
     this.parent,
   );
 
+  /// Stores the [TextStyle] that was used when building the description. When
+  /// not set, then the description has not been built yet.
```
kenzieschmoll, line 51: blank line after first sentence in dart doc.

Also please describe why we need to store this value

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -934,12 +934,12 @@ class _InspectorTreeState extends State<InspectorTree>
     return initialX - columnWidth * padCount;
   }
 
-  /// Pad [initialY] with the vertical height of [padCount] rows.
+  /// Pad [initialY] so that a row would be placed in the vertical centre of
```
kenzieschmoll, line 937: sp: center

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -934,12 +934,12 @@ class _InspectorTreeState extends State<InspectorTree>
...
+  /// the screen.
   double _padTargetY({
     required double initialY,
-    int padCount = _scrollPadCount,
   }) {
-    return initialY - rowHeight * padCount;
+    return initialY - (safeViewportHeight / 2) + rowHeight / 2;
```
kenzieschmoll, line 942: nit: `initialY - (safeViewportHeight + rowHeight) / 2`

kenzieschmoll: nevermind - how you have it actually makes this more clear I think.

kenzieschmoll (approved):

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
CoderDake: We chatted about this during a 1:1 and decided together that scrolling to the center of the screen is a happy medium we can agree on.

CoderDake:
Tested on Macos and Web with:
- searching across the widgets
- keyboard navigation
- hanging widgets off of the screen and clicking on them
-

PR #4286, opened 2022-07-22
TreeMap opens directly to main root for deferred apps

carolynqu:
When opening App-Size panel, the TreeMap always opens to the entire app view. For deferred apps, Treemap should directly open up to the main app bundle directly.

### DIFF
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
@@ -16,7 +16,7 @@ import '../../shared/table.dart';
 import 'app_size_screen.dart';
 
 // Temporary feature flag for deferred loading.
-const deferredLoadingSupportEnabled = false;
+const deferredLoadingSupportEnabled = true;
 
 enum DiffTreeType {
   increaseOnly,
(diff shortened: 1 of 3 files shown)


elliette:
Instead of using a `postFrameCallback`, let's simply pass the actual root (instead of the `ArtificialRoot`) to the `changeAnalysisRoot` call when we first load the app size JSON file: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart#L249

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -246,7 +246,13 @@ class AppSizeController {
     // Build a tree with [TreemapNode] from [processedJsonMap].
     final newRoot = generateTree(processedJson)!;
 
-    changeAnalysisRoot(newRoot);
+    // Open Treemap to main app.
```
elliette, line 249: For clarity: let's rename `newRoot` to `jsonRoot`.  That way we can do something like this:

```
    // Determine the correct root node.
    final newRoot = isDeferredApp
        ? jsonRoot.childrenMap.values.firstWhere((node) => node.name == 'Root')
        : jsonRoot;
    changeAnalysisRoot(newRoot);
```

elliette:
Let's update the description to remove "To be completed"

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -16,7 +16,7 @@ import '../../shared/table.dart';
 import 'app_size_screen.dart';
 
 // Temporary feature flag for deferred loading.
-const deferredLoadingSupportEnabled = false;
+const deferredLoadingSupportEnabled = true;
```
elliette, line 19: Switch back to false

elliette (approved):
LGTM with one change requested!

PR #4288, opened 2022-07-22
Make UI jank hints aware of the enhance tracing state for a frame

kenzieschmoll:
For a frame with jank hints, we now change the enhance tracing hint dependent on whether the tracing was already enhanced for the selected frame.
Screen Shot 2022-07-22 at 11 07 19 AM

For a janky frame where enhance tracing was not already enabled, we will continue to point them to the Enhance Tracing feature.

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/panes/controls/enhance_tracing/enhance_tracing_controller.dart
+++ b/packages/devtools_app/lib/src/screens/performance/panes/controls/enhance_tracing/enhance_tracing_controller.dart
@@ -4,12 +4,10 @@
 
 import 'dart:async';
 
-import 'package:flutter/foundation.dart';
-
 import '../../../../../primitives/auto_dispose.dart';
 import '../../../../../service/service_extensions.dart' as extensions;
 import '../../../../../shared/globals.dart';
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/screens/performance/panes/controls/enhance_tracing/enhance_tracing_controller.dart
```diff
@@ -28,3 +70,31 @@ class EnhanceTracingController extends DisposableController
     super.dispose();
   }
 }
+
+@immutable
+class EnhanceTracingState {
```
polina-c, line 75: This class is a data structure, without logic. Should it go to model, to make the library 'controller' being focused on business logic, not on data structures?

polina-c (approved):

packages/devtools_app/lib/src/screens/performance/panes/controls/enhance_tracing/enhance_tracing_controller.dart
kenzieschmoll: done 👍

PR #4291, opened 2022-07-23
Make tabs on memory screen looking nicer.

polina-c:
1. Moved tab 'Profile' to the very left, to make tab difficulty increasing left to right
2. Applied Mindy's recommendation to convert icon '?' to link ant place immediately to the right of the buttons to which it apply
3. Added borders to the buttons on the pane 'Leaks'

Before:

Screen Shot 2022-07-23 at 8 22 58 AM

Screen Shot 2022-07-23 at 8 23 16 AM

After:

Screen Shot 2022-07-23 at 9 42 28 AM

Screen Shot 2022-07-23 at 9 41 23 AM


packages/devtools_app/lib/src/screens/memory/memory_heap_tree_view.dart
```diff
@@ -40,7 +40,7 @@ import 'primitives/memory_utils.dart';
 // TODO(bkonyi): enable new allocation profile table when we're ready to remove
 // the existing allocations table.
 @visibleForTesting
-bool enableNewAllocationProfileTable = false;
+bool enableNewAllocationProfileTable = true;
```
polina-c, line 43: I will flip it back before submitting.

packages/devtools_app/lib/src/screens/memory/primitives/ui.dart
```diff
@@ -0,0 +1,6 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Width of application when memory buttons loose their text.
```
bkonyi, line 5: Nit: loose -> lose

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -2150,3 +2150,31 @@ class _KeepAliveWrapperState extends State<KeepAliveWrapper>
     return widget.child;
   }
 }
+
+class DocumentationLink extends StatelessWidget {
```
bkonyi, line 2154: Nit: documentation for the `DocumentationLink`? ;)

packages/devtools_app/lib/src/analytics/constants.dart
```diff
@@ -139,6 +139,7 @@ const export = 'export';
 const expandAll = 'expandAll';
 const collapseAll = 'collapseAll';
 const documentationLink = 'documentationLink';
+String topicDocumentationLink(String topic) => '${topic}DocumentationLink';
```
bkonyi, line 142: It feels strange declaring this in the middle of a list of constants. Can it be put somewhere else?

bkonyi:
I'm not sure how I feel about using a `Link` style for linking to documentation. We only really use web-style link styling when explicitly linking to URLs and having this one link in a row of material widgets doesn't look right to me. I don't think it's worth blocking the whole PR on, but maybe Kenzie has an opinion on this.

packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
```diff
@@ -343,13 +344,8 @@ class _AllocationProfileTableControls extends StatelessWidget {
...
           allocationProfileController: allocationProfileController,
         ),
-        const Spacer(),
-        const InformationButton(
-          tooltip: 'View documentation for the allocation profile table.',
-          link: 'https://github.com/flutter/devtools/blob/master/'
-              'packages/devtools_app/lib/src/screens/memory/panes/'
-              'allocation_profile/ALLOCATION_PROFILE.md',
-        ),
+        const SizedBox(width: 2 * denseSpacing),
```
kenzieschmoll, line 347: use `defaultSpacing` instead

packages/devtools_app/lib/src/screens/memory/panes/leaks/leaks_pane.dart
```diff
@@ -252,12 +251,15 @@ class _LeaksPaneState extends State<LeaksPane> {
               controller: _leaksController.status,
               analysisStarter: Row(
                 children: [
-                  infoButton,
                   _AnalyzeButton(leaksController: _leaksController),
+                  const SizedBox(width: denseSpacing),
                   _ForceGCButton(leaksController: _leaksController),
+                  const SizedBox(width: 2 * denseSpacing),
```
kenzieschmoll, line 257: defaultSpacing

kenzieschmoll:
I agree with Ben wrt using the link instead of a button widget. If the goal is to make it clear that we are linking out to the browser, we could consider having the info button pop up an overlay with a brief description, and then that description could link out to to docs (similar to what we do for service extension buttons):
Screen Shot 2022-07-25 at 9 21 57 AM

melyndahoover:
I advised Polina to consider a link instead of the question button last week to help indicate the redirection to an outside page and to differentiate it from the "?" button on the top of the DevTools menu. However, I agree that a simple hyperlinked question-mark is still a bit unclear. If you keep the direct hyperlink, perhaps use "help" instead of a question mark. I also like Kenzie's idea of using a question mark button that triggers a pop up with more info and a hyperlink, as long as the behavior of these buttons is consistent throughout the DevTools.

polina-c:
> I advised Polina to consider a link instead of the question button last week to help indicate the redirection to an outside page and to differentiate it from the "?" button on the top of the DevTools menu. However, I agree that a simple hyperlinked question-mark is still a bit unclear. If you keep the direct hyperlink, perhaps use "help" instead of a question mark. I also like Kenzie's idea of using a question mark button that triggers a pop up with more info and a hyperlink, as long as the behavior of these buttons is consistent throughout the DevTools.

Adopted button with dialog from performance page.

packages/devtools_app/lib/src/analytics/constants.dart
polina-c: Separated documentation related items. It is good to have them together in spite of different types.

packages/devtools_app/lib/src/screens/memory/memory_heap_tree_view.dart
kenzieschmoll, line 43: flip this back to false before submitting

packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
```diff
@@ -412,3 +407,36 @@ class _RefreshOnGCToggleButton extends StatelessWidget {
     );
   }
 }
+
+class _ProfileHelpLink extends StatelessWidget {
+  const _ProfileHelpLink({Key? key}) : super(key: key);
+
+  static const _documentationTopic = 'profile';
```
kenzieschmoll, line 414: let's make this 'allocationProfile' to make it more specific. that will be easier to understand when looking at our analytics data since 'profile' by itself is pretty general

packages/devtools_app/lib/src/screens/memory/memory_heap_tree_view.dart
```diff
@@ -202,6 +202,12 @@ class HeapTreeViewState extends State<HeapTree>
 
   void _initTabs() {
     _tabs = [
+      if (enableNewAllocationProfileTable)
+        DevToolsTab.create(
+          key: dartHeapTableTabKey,
+          tabName: 'Profile',
```
kenzieschmoll, line 208: thoughts on renaming to 'Allocation Profile'? @polina-c @InMatrix

packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
```diff
@@ -412,3 +407,36 @@ class _RefreshOnGCToggleButton extends StatelessWidget {
...
+
+  @override
+  Widget build(BuildContext context) {
+    return HelpButtonWithDialog(
+      gaScreen: analytics_constants.memory,
+      gaSelection:
+          analytics_constants.topicDocumentationButton(_documentationTopic),
+      dialogTitle: 'Memory Allocation Profile Help',
+      child: Column(
+        crossAxisAlignment: CrossAxisAlignment.end,
```
kenzieschmoll, line 424: this seems odd. Why would we need to align to the end of the column? If anything, I'd think we would use start

packages/devtools_app/lib/src/screens/memory/memory_heap_tree_view.dart
polina-c: I believe tab names should be short for easiness of reference. We have context 'Memory'. So it is Memory Profile, that is the same as Memory Allocation Profile. How about taking this discussion out of this PR?

packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
```diff
@@ -412,3 +407,36 @@ class _RefreshOnGCToggleButton extends StatelessWidget {
...
+        children: [
+          const Text(
+              'The allocation profile tab displays information about currently\n'
+              'allocated objects in the Dart heap of the currently selected\n'
+              'isolate.'),
```
kenzieschmoll, line 429: @sfshaza2 for help with wording

packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
polina-c: The text itself is aligned to the left. Don't we want link 'more info' to be on the right?

packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
```diff
@@ -412,3 +407,36 @@ class _RefreshOnGCToggleButton extends StatelessWidget {
...
```
sfshaza2, line 429: I think you can lose both "current"s. If you prefer to keep one of them, I'd replace "selected" with "current".

```suggestion
              'The allocation profile tab displays information about\n'
              'allocated objects in the Dart heap of the selected\n'
              'isolate.'),
```

packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
kenzieschmoll: I see. Okay as long as the text is left aligned 👍  looks like the column width is the same width as the Text widget, so alignment of the Text widget within the column won't change wrt to cross axis alignment (unless there is a wider widget than the text widget added to the column at some point).

packages/devtools_app/lib/src/screens/memory/panes/leaks/leaks_pane.dart
```diff
@@ -273,6 +274,35 @@ class _LeaksPaneState extends State<LeaksPane> {
...
+    return HelpButtonWithDialog(
+      gaScreen: analytics_constants.memory,
+      gaSelection:
+          analytics_constants.topicDocumentationButton(_documentationTopic),
+      dialogTitle: 'Memory Leak Detection Help',
+      child: Column(
+        crossAxisAlignment: CrossAxisAlignment.end,
+        children: [
+          const Text('The memory leak detection tab helps to detect\n'
+              'and troubleshoot some types of memory leaks.'),
```
sfshaza2, line 293: What types? I'd prefer to change "some types" so that it's less ambiguous.

```suggestion
          const Text('Use the memory leak detection tab to detect\n'
              'and troubleshoot some types of memory leaks.'),
```

packages/devtools_app/lib/src/screens/memory/panes/README.md
```diff
@@ -0,0 +1,4 @@
+Panes in this folder should not depend on each other.
+
+If they need to share code, move the code to memory/primitives (if it is one library or less),
+or to memory/shared (if it is more than one library).
```
sfshaza2, line 4: How can it be in one library OR LESS?

```suggestion
If any panes need to share code, move the code to memory
or primitives if used in one library,
or to memory or shared
if used in more than one library.
```

packages/devtools_app/lib/src/screens/memory/primitives/ui.dart
```diff
@@ -0,0 +1,6 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Width of application when memory buttons lose their text.
```
sfshaza2, line 5: "Width of application"? What does that mean? Do you mean width of the screen?

sfshaza2 (commented):
A few suggested edits, otherwise lgtm

packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
polina-c: applied the Sham's suggestion

packages/devtools_app/lib/src/screens/memory/panes/README.md
polina-c: memory/primitives is a path that consists of two folders.  Added ticks.

packages/devtools_app/lib/src/screens/memory/primitives/ui.dart
polina-c: added 'window'

packages/devtools_app/lib/src/screens/memory/memory_heap_tree_view.dart
polina-c: it is protected by test, so no way I can forget it :)

kenzieschmoll (approved):

PR #4296, opened 2022-07-25
Deferred nodes are gray scaled on TreeMap

carolynqu:
On TreeMap UI, deferred nodes are gray scaled to help visibly distinguish between the main app vs deferred chunks. 

**Originals**
Screen Shot 2022-07-25 at 4 41 30 PM
Screen Shot 2022-07-25 at 4 41 49 PM

**New Deferred Graying**
Screen Shot 2022-07-25 at 4 58 01 PM
Screen Shot 2022-07-25 at 4 57 45 PM

### DIFF
--- a/packages/devtools_app/lib/src/charts/treemap.dart
+++ b/packages/devtools_app/lib/src/charts/treemap.dart
@@ -472,7 +472,9 @@ class _TreemapState extends State<Treemap> {
   Container buildCell() {
     return Container(
       decoration: BoxDecoration(
-        color: widget.rootNode!.displayColor,
+        color: widget.rootNode!.isDeferred
+            ? Colors.grey
+            : widget.rootNode!.displayColor,
         border: Border.all(color: Colors.black87),
       ),
       child: Center(
--- a/packages/devtools_app/lib/src/charts/treemap.dart
+++ b/packages/devtools_app/lib/src/charts/treemap.dart
@@ -605,10 +605,7 @@ class TreemapNode extends TreeNode<TreemapNode> {
 
   Color get displayColor {
     if (!showDiff) {
-      if (isDeferred)
-        return Colors.grey;
-      else
-        return mainUiColor;
+      return isDeferred ? treemapDeferredColor : mainUiColor;
     }
     if (byteSize < 0) {
       return treemapDecreaseColor;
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -489,11 +489,14 @@ class AppSizeController {
       childrenMap[child.name] = child;
     }
 
+    final bool isDeferred = treeJson['isDeferred'] != null;
```
elliette, line 492: Let's check if not null and also if true

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -745,7 +749,8 @@ class MultiCellPainter extends CustomPainter {
     );
 
     final rectPaint = Paint();
-    rectPaint.color = node.displayColor;
+    final paintColor = node.isDeferred ? Colors.grey : node.displayColor;
```
elliette, line 752: Let's actually move this logic into the `displayColor` getter itself

elliette: Also instead of `Colors.grey`, let's create a `treemapDeferredColor` variable. Let's try using the `Colors.indigo[100]` from the Material UI spec here: https://api.flutter.dev/flutter/material/Colors-class.html

We can list it along with the other treemap colors here: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/ui/colors.dart#L74

elliette:
Once you change the colors, can you add some screenshots of what it looks like in light and dark modes?

packages/devtools_app/lib/src/charts/treemap.dart
elliette: Unresolving to address the color change

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -591,18 +591,25 @@ class TreemapNode extends TreeNode<TreemapNode> {
...
 
   final bool showDiff;
+  final bool isDeferred;
 
   int get unsignedByteSize => byteSize.abs();
 
   Color get displayColor {
-    if (!showDiff) return mainUiColor;
+    if (!showDiff) {
+      if (isDeferred)
```
elliette, line 608: let's change this to a ternary: `return isDeferred ? treemapDeferredColor : mainUiColor`

elliette (approved):

PR #4297, opened 2022-07-26
Adding deferred indicator on the treemap

carolynqu:
Indicate that deferred chunks on the Treemap are deferred by adding a caption saying "_(Deferred)_"
Screen Shot 2022-07-27 at 2 03 05 PM

### DIFF
--- a/packages/devtools_app/lib/src/charts/treemap.dart
+++ b/packages/devtools_app/lib/src/charts/treemap.dart
@@ -478,8 +478,6 @@ class _TreemapState extends State<Treemap> {
       child: Center(
         child: widget.height > Treemap.minHeightToDisplayCellText
             ? buildNameAndSizeText(
-                textColor:
-                    widget.rootNode!.showDiff ? Colors.white : Colors.black,
                 oneLine: false,
               )
             : const SizedBox(),
@@ -499,9 +497,6 @@ class _TreemapState extends State<Treemap> {
             border: Border.all(color: Colors.black87),
           ),
           child: buildNameAndSizeText(
-            textColor:
-                (Theme.of(context).textTheme.bodyText2 ?? const TextStyle())
-                    .color,
             oneLine: true,
           ),
         ),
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -509,16 +509,18 @@ class _TreemapState extends State<Treemap> {
     }
   }
 
-  Text buildNameAndSizeText({
+  RichText buildNameAndSizeText({
```
elliette, line 512: Can we return a `TextSpan` instead of `RichText` (that way we don't have to wrap our `TextSpan` in `RichText`)?

carolynqu: RichText is being used to support buildCell and buildTitleText request for a widget.

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -760,12 +763,7 @@ class MultiCellPainter extends CustomPainter {
     if (positionedCell.width! > Treemap.minWidthToDisplayCellText &&
         positionedCell.height! > Treemap.minHeightToDisplayCellText) {
       final textPainter = TextPainter(
-        text: TextSpan(
-          text: node.displayText(oneLine: false),
-          style: TextStyle(
-            color: node.showDiff ? Colors.white : Colors.black,
-          ),
-        ),
+        text: node.displayText(),
```
elliette, line 766: Should we still be passing `oneLine: false` here?

packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
```diff
@@ -80,13 +80,14 @@ class _NameColumn extends TreeColumnData<TreemapNode> {
   final int currentRootLevel;
 
   @override
-  String getValue(TreemapNode dataObject) => dataObject.name;
+  String getValue(TreemapNode dataObject) => '${dataObject.name}';
```
elliette, line 83: `dataObject.name` should already be a string

elliette:
Can you add a few screenshots to the description?

packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -628,8 +635,19 @@ class TreemapNode extends TreeNode<TreemapNode> {
...
       }
     }
+
     final separator = oneLine ? ' ' : '\n';
-    return '$displayName$separator[${prettyByteSize()}]';
+
+    return TextSpan(
+      text: '$displayName$separator[${prettyByteSize()}]',
+      children: [
+        TextSpan(
```
elliette, line 644: Instead we can add a `if`statement to the list itself:

```
children: [
  if (caption != null) 
    TextSpan(text: caption, style: etc.),
]

elliette (approved):
LGTM with one suggestion

PR #4298, opened 2022-07-26
Improve `msText` util to support preventing zero values

kenzieschmoll:


packages/devtools_app/lib/src/primitives/utils.dart
```diff
@@ -122,9 +122,22 @@ String msText(
   Duration dur, {
   bool includeUnit = true,
   int fractionDigits = 1,
+  bool allowZeroValues = true,
```
polina-c, line 125: Can you add documentation to this method that will explain the method itself behavior in case of zeros?

polina-c (approved):

packages/devtools_app/lib/src/primitives/utils.dart
kenzieschmoll: done

PR #4299, opened 2022-07-26
Use Pie Charts for visualizing frame time in Frame Analysis tab

kenzieschmoll:
Here are some examples of what these Pie charts can look like:
Screen Shot 2022-07-26 at 1 17 18 PM
Screen Shot 2022-07-26 at 1 16 30 PM
Screen Shot 2022-07-26 at 1 14 31 PM

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/panes/frame_analysis/frame_analysis.dart
+++ b/packages/devtools_app/lib/src/screens/performance/panes/frame_analysis/frame_analysis.dart
@@ -4,14 +4,12 @@
 
 import 'package:flutter/material.dart';
 
-import '../../../../primitives/utils.dart';
 import '../../../../shared/common_widgets.dart';
 import '../../../../shared/theme.dart';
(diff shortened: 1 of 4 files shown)


InMatrix:
Is Build usually, if not always, the longest UI phase? If that's the case, what would an unusually long Build look like? I'm trying to understand what is the most important information to communicate here. 

In general, Pie chart has poor space efficiency, and it's usually for communicating proportional data, but the values in the screenshots are absolute values. Simple bar charts or just tables are probably sufficient here.

kenzieschmoll:
> Is Build usually, if not always, the longest UI phase? 

Not always. The layout or paint phase can take longer than build - just depends on what is being built/rendered in the frame.

> Simple bar charts or just tables are probably sufficient here.

Noted. Thank you for the feedback.

kenzieschmoll:
closing in favor of https://github.com/flutter/devtools/pull/4328

PR #4305, opened 2022-07-27
Adding deferred indicator on the treetable

carolynqu:
Indicate that deferred chunks on the Treetable are deferred by adding a caption saying "(Deferred)" to the side of the name.
Screen Shot 2022-07-27 at 4 29 17 PM

### DIFF
--- a/packages/devtools_app/lib/src/charts/treemap.dart
+++ b/packages/devtools_app/lib/src/charts/treemap.dart
@@ -478,8 +478,6 @@ class _TreemapState extends State<Treemap> {
       child: Center(
         child: widget.height > Treemap.minHeightToDisplayCellText
             ? buildNameAndSizeText(
-                textColor:
-                    widget.rootNode!.showDiff ? Colors.white : Colors.black,
                 oneLine: false,
               )
             : const SizedBox(),
@@ -499,9 +497,6 @@ class _TreemapState extends State<Treemap> {
             border: Border.all(color: Colors.black87),
           ),
           child: buildNameAndSizeText(
-            textColor:
-                (Theme.of(context).textTheme.bodyText2 ?? const TextStyle())
-                    .color,
             oneLine: true,
           ),
         ),
@@ -510,17 +505,9 @@ class _TreemapState extends State<Treemap> {
   }
 
   RichText buildNameAndSizeText({
-    required Color? textColor,
     required bool oneLine,
   }) {
     return RichText(text: widget.rootNode!.displayText(oneLine: oneLine));
-
-    // Text(
-    //   widget.rootNode!.displayText(oneLine: oneLine).toPlainText(),
-    //   style: TextStyle(color: textColor),
-    //   textAlign: TextAlign.center,
-    //   overflow: TextOverflow.ellipsis,
-    // );
   }
 
   Widget buildBreadcrumbNavigator() {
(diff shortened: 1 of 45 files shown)


packages/devtools_app/lib/src/charts/treemap.dart
```diff
@@ -641,11 +641,10 @@ class TreemapNode extends TreeNode<TreemapNode> {
     return TextSpan(
       text: '$displayName$separator[${prettyByteSize()}]',
       children: [
-        if (caption != null)
-          TextSpan(
-            text: '$separator$caption',
-            style: TextStyle(fontStyle: FontStyle.italic, color: textColor),
-          )
+        TextSpan(
```
elliette, line 644: Why were these changes introduced?

carolynqu: Oh whoops, it didn't update from the previous changes. Will revert!

packages/devtools_app/lib/src/shared/table_data.dart
```diff
@@ -61,6 +61,8 @@ abstract class ColumnData<T> {
   String getDisplayValue(T dataObject) =>
       getValue(dataObject)?.toString() ?? '';
 
+  String? getCaption(T dataObject) => '';
```
elliette, line 64: Switch to returning `null` by default instead of an empty string (that way the check `if (column.getCaption(node) != null)` will work as expected).

packages/devtools_app/test/app_size/app_size_screen_test.dart
```diff
@@ -170,8 +170,10 @@ void main() {
       expect(breadcrumbs.length, 1);
       expect(breadcrumbs.first.text, equals('Root [6.0 MB]'));
       expect(find.byType(BreadcrumbNavigator), findsOneWidget);
-      expect(find.text('package:flutter'), findsOneWidget);
-      expect(find.text('dart:core'), findsOneWidget);
+      String text = 'package:flutter';
+      expect(textFinder(text), findsOneWidget);
```
elliette, line 174: Once we change `getCaption` to return `null`, we should be able to search directly for the text with `find.richText(text)`, which should only return one widget (here and elsewhere).

packages/devtools_app/test/app_size/app_size_screen_test.dart
```diff
@@ -470,3 +479,7 @@ class AppSizeTestController extends AppSizeController {
     );
   }
 }
+
+Finder textFinder(String text) {
```
elliette, line 483: We can remove this and use `find.richText`

packages/devtools_app/test/app_size/app_size_screen_test.dart
```diff
@@ -170,8 +170,10 @@ void main() {
       expect(breadcrumbs.length, 1);
       expect(breadcrumbs.first.text, equals('Root [6.0 MB]'));
       expect(find.byType(BreadcrumbNavigator), findsOneWidget);
-      expect(find.text('package:flutter'), findsOneWidget);
-      expect(find.text('dart:core'), findsOneWidget);
+      String text = 'package:flutter';
```
elliette, line 173: Small nit: can we switch back to directly passing the raw string to the finder (instead of setting as a variable). Here and elsewhere.

elliette (approved):
LGTM with  one change, thanks!

PR #4306, opened 2022-07-27
Adding multi-package support for pubrootdirectories

CoderDake:
This PR turns on the pub root directory feature

It also adds the proper functionality that keeps the local storage of each application's  pubroot directories 
 seperate

https://user-images.githubusercontent.com/1386322/181389999-d804944d-2feb-4cc7-b413-355f748e4f8c.mp4


packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -114,25 +131,54 @@ class InspectorPreferencesController extends DisposableController
...
+    storage.setValue(
+      _customPubRootStorageId(),
+      jsonEncode(_customPubRootDirectories.value),
+    );
+  }
+
   Future<void> addPubRootDirectories(
     List<String> pubRootDirectories,
   ) async {
+    if (!serviceManager.hasConnection) return;
```
CoderDake, line 181: It might be worth adding an error state to the pubrootdirectories in the future. and set this instead of just shortcircuiting

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -86,19 +86,36 @@ class InspectorPreferencesController extends DisposableController
       serviceManager.inspectorService as InspectorService;
 
   final _hoverEvalMode = ValueNotifier<bool>(false);
-  final _customPubRootDirectories = ListValueNotifier<String>([]);
+  final ListValueNotifier<String> _customPubRootDirectories =
```
kenzieschmoll, line 89: nit: remove redundant type on LHS

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -86,19 +86,36 @@ class InspectorPreferencesController extends DisposableController
...
+    final isolateRef = serviceManager.isolateManager.mainIsolate.value!;
+    if (isolateRef.id != null) {
+      final isolate = await serviceManager.service?.getIsolate(isolateRef.id!);
+      final rootLibUri = Uri.parse(isolate?.rootLib?.uri ?? '');
+      final directorySegments = rootLibUri.pathSegments
+          .sublist(0, rootLibUri.pathSegments.length - 1);
+      final rootLibDirectory = rootLibUri.replace(
+        pathSegments: directorySegments,
+      );
+      _mainScriptDir = rootLibDirectory.path;
```
kenzieschmoll, line 109: was this code copied from somewhere in the debugger code by chance? if so, we should de-dupe and make a utility method

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -114,25 +131,54 @@ class InspectorPreferencesController extends DisposableController
     });
   }
 
-  void _initCustomPubRootListeners() {
-    // TODO(CoderDake): add _customPubRootDirectories listener back when
-    // finalizing https://github.com/flutter/devtools/issues/3941
-    /*
-    addAutoDisposeListener(_customPubRootDirectories, () {
```
kenzieschmoll, line 121: is it no longer necessary to listen to `_customPubRootDirectories`?

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -114,25 +131,54 @@ class InspectorPreferencesController extends DisposableController
...
+    if (isolateListener == null) {
+      isolateListener = () {
+        preferences.inspector.loadCustomPubRootDirectories();
+      };
+      addAutoDisposeListener(
+        serviceManager.isolateManager.mainIsolate,
+        isolateListener,
+      );
+    }
+  }
```
kenzieschmoll, line 169: if this should only happen once, let's add it in `_initCustomPubRootDirectories`

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -160,14 +207,24 @@ class InspectorPreferencesController extends DisposableController
...
+
+        _persistCustomPubRootDirectoriesToStorage();
       }
     });
   }
 
+  String _customPubRootStorageId() {
+    assert(_mainScriptDir != null);
+    final packageId = _mainScriptDir ?? '_fallback';
+    return '$_customPubRootDirectoriesStoragePrefix:$packageId';
```
kenzieschmoll, line 219: nit: maybe use `_` instead of `:` to be more dart-ey?

kenzieschmoll (commented):
should we add a simple test that uses the FlutterTestEnvironment to spin up a test app, and then verify `_customPubRootStorageId` generates the id we expect?

packages/devtools_app/lib/src/shared/preferences.dart
CoderDake: ours is pretty customized, the only part that would be similar is that we are getting the rootLib.uri
In this case I think it might be more trouble to try to make a utility method

packages/devtools_app/lib/src/shared/preferences.dart
CoderDake: I've added a much fancier and explicit mainIsolate behaviour for this now. I'd love to hear any feedback on it :D

packages/devtools_app/lib/src/shared/preferences.dart
CoderDake: Correct, this is now only done explicitly when we call _persistCustomPubRootDirectoriesToStorage.
The upside to this is that our values in storage will be more accurate, The downside is that we may redundantly save the same value a couple of times.

packages/devtools_app/lib/src/primitives/debouncer.dart
```diff
@@ -0,0 +1,18 @@
+import 'dart:async';
```
kenzieschmoll, line 1: copyright

packages/devtools_app/lib/src/primitives/debouncer.dart
```diff
@@ -0,0 +1,18 @@
...
+
+class Debouncer {
+  Debouncer({
+    required this.milliseconds,
+  });
+  final int milliseconds;
+  Timer? _timer;
+  void call(void Function() action) {
```
kenzieschmoll, line 9: new line between class members

packages/devtools_app/lib/src/primitives/debouncer.dart
CoderDake: 🤦 
thank you!

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -114,25 +129,71 @@ class InspectorPreferencesController extends DisposableController
...
+              serviceManager.isolateManager.mainIsolateDebuggerState;
+
+          if (debuggerState?.isPaused.value == false) {
+            // the isolate is already unpaused, we can try to load
+            // the directories
+            preferences.inspector.loadCustomPubRootDirectories();
+          } else {
+            late Function() pausedListener;
+
+            pausedListener = () {
```
kenzieschmoll, line 158: no need to make this late if we can just declare it directly

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -126,13 +130,10 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
       }
     });
     addAutoDisposeListener(preferences.inspector.customPubRootDirectories, () {
-      _refreshInspector();
-    });
-    addAutoDisposeListener(serviceManager.isolateManager.mainIsolate, () {
-      _refreshInspector();
-      preferences.inspector.loadCustomPubRootDirectories();
+      if (serviceManager.hasConnection) {
+        _customPubRootDirectoriesRefreshDebouncer(() => _refreshInspector());
```
kenzieschmoll, line 134: can we just `await Future.delayed(...)` if we want to call this after a delay, instead of creating a Debouncer class?

packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
```diff
@@ -460,10 +460,14 @@ class InspectorService extends InspectorServiceBase {
     final response = await invokeServiceMethodDaemonNoGroupArgs(
       'getPubRootDirectories',
     );
-
-    return (response as List<dynamic>)
-        .map<String>((e) => e.toString())
-        .toList();
+    if (response.runtimeType == List) {
```
kenzieschmoll, line 463: or just `if (response == nul) return [];`

packages/devtools_app/test/test_infra/flutter_test_environment.dart
```diff
@@ -115,7 +116,7 @@ class FlutterTestEnvironment {
       _service = _flutter!.vmService!;
 
       setGlobal(IdeTheme, IdeTheme());
-      final preferencesController = PreferencesController();
+      final preferencesController = FakePreferencesController();
```
kenzieschmoll, line 119: we may not always want to use fake preferences for all tests that use the flutter_test_environment. Maybe make this configurable if this ends up being the solution to the testing issues you were seeing

packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
CoderDake: I believe the response is `{result: null}` so I don't think this would work.

CoderDake:
A follow up issue has been created to make the inspector preferences cleaner and more testable
https://github.com/flutter/devtools/issues/4342

CoderDake:
Note: for now I've manually put the inspector preferences through the ringer. It behaves well and preserves pub root directories well between different packages

packages/devtools_test/lib/src/mocks/fake_preferences_controller.dart
```diff
@@ -0,0 +1,33 @@
+import 'package:devtools_app/devtools_app.dart';
```
CoderDake, line 1: copyright

packages/devtools_app/test/shared/preferences_controller_test.dart
```diff
@@ -112,5 +116,6 @@ void main() {
         );
       });
     });
+    // TODO: Add custom pubrootdir tests
```
kenzieschmoll, line 119: can you link to an issue for this?

packages/devtools_test/lib/src/mocks/fake_vm_service_wrapper.dart
```diff
@@ -204,7 +204,14 @@ class FakeVmServiceWrapper extends Fake implements VmServiceWrapper {
 
   @override
   Future<Isolate> getIsolate(String isolateId) {
-    return Future.value(MockIsolate());
+    return Future.value(
```
CoderDake, line 207: check if this is still needed

packages/devtools_app/test/test_infra/flutter_test_environment.dart
```diff
@@ -28,8 +29,11 @@ class FlutterTestEnvironment {
     this._runConfig, {
     this.testAppDirectory = 'test/fixtures/flutter_app',
     FlutterDriverFactory? flutterDriverFactory,
+    PreferencesController? preferencesController,
   })  : _flutterDriverFactory = flutterDriverFactory ?? defaultFlutterRunDriver,
-        _flutterExe = _parseFlutterExeFromEnv();
+        _flutterExe = _parseFlutterExeFromEnv(),
+        _preferencesController =
+            preferencesController ?? FakePreferencesController();
```
kenzieschmoll, line 36: I would do the opposite - use `PreferencesController()` unless `_preferencesController` has been specified (i.e. the prior behavior as to avoid unintended consequences)

packages/devtools_app/test/test_infra/flutter_test_environment.dart
```diff
@@ -28,8 +28,11 @@ class FlutterTestEnvironment {
     this._runConfig, {
     this.testAppDirectory = 'test/fixtures/flutter_app',
     FlutterDriverFactory? flutterDriverFactory,
+    PreferencesController? preferencesController,
```
kenzieschmoll, line 31: do we event need these changes anymore? I'm not seeing them being used anywhere

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -114,25 +129,71 @@ class InspectorPreferencesController extends DisposableController
...
+              if (debuggerState?.isPaused.value == false) {
+                preferences.inspector.loadCustomPubRootDirectories();
+
+                debuggerState?.isPaused.removeListener(pausedListener);
```
kenzieschmoll, line 162: we should override `dispose` on `InspectorPreferencesController` and remove this listener. That way, in case the app is never unpaused this listener will still be removed.

CoderDake: I think just adding autodispose for the listener will accomplish this.
It has the added benefit of still removing listeners that are around, even if the main isolate switches around a bunch too!

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -164,7 +164,7 @@ class InspectorPreferencesController extends DisposableController
             };
 
             // The isolate is still paused, listen for when it becomes unpaused.
-            debuggerState?.isPaused.addListener(pausedListener);
+            addAutoDisposeListener(debuggerState?.isPaused, pausedListener);
```
kenzieschmoll, line 167: just want to the disposal of this listener doesn't cause any issues when the disposal code path is hit after the listener has already been removed?

packages/devtools_app/test/test_infra/flutter_test_environment.dart
```diff
@@ -86,6 +86,8 @@ class FlutterTestEnvironment {
   // Switch this flag to false to debug issues with non-atomic test behavior.
   bool reuseTestEnvironment = true;
 
+  final PreferencesController _preferencesController = PreferencesController();
```
kenzieschmoll, line 89: is there a reason we want to initialize instead of during `setupEnvironment`? I suspect this will cause issues with starting with a stale preferences state when `setupEnvironment` is called.

CoderDake: No specific reason, I'll update it

packages/devtools_app/lib/src/shared/preferences.dart
CoderDake: The removeListener code only removes if it can find the listener, so I think this should be fine.
I'm also tested this manually and no errors came up :)

kenzieschmoll (approved):

PR #4308, opened 2022-07-28
Adding ArtificialRoot to non-deferred apps when using the DiffTable with a deferred app

carolynqu:
When using the DiffTable to compare a non-deferred and deferred app, the main app roots are not at the same location. So, the non-deferred app will add its own ArtificialRoot so that they can be accurately compared.


packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -279,8 +279,28 @@ class AppSizeController {
 
     Map<String, dynamic> diffMap;
     if (oldFile.isAnalyzeSizeFile && newFile.isAnalyzeSizeFile) {
+      var oldFileJson = oldFile.data as Map<String, dynamic>;
+      var newFileJson = newFile.data as Map<String, dynamic>;
+
+      if (!(oldFileJson['n'] == 'ArtificialRoot') ==
```
elliette, line 285: Let's pull the check for the `ArtificialRoot` into a helper function called `isDeferredFormat` which takes in the processed JSON and returns a bool. We can then use that here, and in the `isDeferredApp` check above.

elliette: Let's also assign it to a constant variable (eg `artificialRootName`, like this: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/debugger/debugger_screen.dart#L102) so that we don't have to use the raw string everywhere.

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -279,8 +279,28 @@ class AppSizeController {
...
+            'n': 'ArtificialRoot',
+            'children': [newFileJson]
+          };
+        }
+      }
+
       final oldApkProgramInfo = ProgramInfo();
-      final oldFileJson = oldFile.data as Map<String, dynamic>;
+
+      print(oldApkProgramInfo.root);
```
elliette, line 302: leftover?

PR #4313, opened 2022-07-28
Expand and Unexpand top nodes in Treetable

carolynqu:
Fixing: top nodes in treetable cannot be expanded or unexpanded when clicked on. 
bug


elliette (approved):

carolynqu:
Solution Options:
- Make it so that when changing DiffTreeTable type we also call initState and expand the roots then and remove the expand roots from didUpdateWidget
- Remove expandRoots from didUpdateWidget and instead place them in a location where DiffTreeTable is being modified.
- Investigate why root.isExpanded is changing from false to true when clicking on it. This causes expandRoot to expand the roots just after it has been collapsed by toggleNode which is called first.

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -417,6 +417,7 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
   @override
   void initState() {
     super.initState();
+
```
elliette, line 420: Let's remove newline

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -453,24 +454,26 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
...
 
-    if (widget.sortColumn != oldWidget.sortColumn ||
+    if (widget.sortColumn.title != oldWidget.sortColumn.title ||
         widget.sortDirection != oldWidget.sortDirection ||
         !collectionEquals(widget.dataRoots, oldWidget.dataRoots)) {
       _initData();
     }
     rootsExpanded =
         List.generate(dataRoots.length, (index) => dataRoots[index].isExpanded);
+
```
elliette, line 464: Remove new lien

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -453,24 +454,26 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
...
   void _initData() {
     dataRoots = List.generate(widget.dataRoots.length, (index) {
       final root = widget.dataRoots[index];
-      if (widget.autoExpandRoots) {
+      if (widget.autoExpandRoots && !root.isExpanded) {
         root.expand();
       }
       return root;
     });
+
```
elliette, line 476: Remove new line

elliette (approved):
LGTM once new lines are removed!

PR #4325, opened 2022-08-01
Extract some span_parser tests into "golden" tests

DanTup:
The goal here is to make it easier to extend the testing of the parsing with additional test cases without needing to construct a lot of Dart code for expectations in the tests. Instead, `.dart` files in `test_data/syntax_highlighting/` are run through the syntax highlighting and dumped as `.dart.golden` files that annotate the results for each line. Adding new tests just requires adding new files to that folder and running `flutter test test/shared/span_parser_test.dart --update-goldens`.

Some of the current results do not look like we'd expect (in particular, some overlaps in the `comments.dart` results), however this PR is intended only to record the current state of the parser and not make any changes/fixes. With some existing baselines we can ensure any changes to fix these issues are doing exactly (and only) as expected.

## Pre-launch Checklist

- [x] I signed the [CLA].

If you need help, consider asking for advice on the #hackers-new channel on [Discord].


packages/devtools_app/test/shared/span_parser_test.dart
```diff
@@ -192,1272 +40,196 @@ void spanTester({
...
-      spanTester(
-        span: spans[6],
-        scopes: [
-          'comment.line.double-slash.dart',
-        ],
-        line: 7,
-        column: 3,
-        length: 52,
-      );
+    group('golden', () {
```
kenzieschmoll, line 153: Just want to double check - these goldens are not platform specific, right? For our image goldens, we only test them on MacOs, but since these are text goldens, this shouldn't matter.

second question, can we use the `equalsGoldenIgnoringHashCodes` helper here? This is what the inspector uses for its text goldens.

DanTup: >  these goldens are not platform specific, right?

Correct. I'm normalizing the newlines when comparing to avoid any failures from different line endings (some Windows users check files out as-is on Windows, meaning the goldens would have `\n` and some have them translated to `\r\n`).

> can we use the `equalsGoldenIgnoringHashCodes` helper here?

It looks like that helper is mainly designed for regexing out any 5-digit hex values following `#` which doesn't apply here. It does also handle adding a suffix for the goldens folder, but I deliberately didn't put these inside `goldens` because everything in there seems to be Flutter-version specific (so we have `goldens` and `goldens_master`), but this is just internal code and not affected by Flutter version at all. We could use it, but we might end up having two copies of the same goldens to update. WDYT?

packages/devtools_app/test/test_data/syntax_highlighting/comments.dart
```diff
@@ -0,0 +1,44 @@
+/// Multiline dartdoc comment.
```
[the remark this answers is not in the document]
DanTup: Done! (it means they'll show up in the golden file since they'll be syntax highlighted, but I don't think that's a big deal).

packages/devtools_app/test/shared/span_parser_test.dart
kenzieschmoll: the `goldens_stable` folder is obsolete and something we should delete. For all the other goldens in DevTools, we only test against the version of flutter we pin our bots to (the latest beta). What do you think about putting the syntax_highlighing goldens under `goldens/syntax_highlighting/` so that all the goldens for DevTools are in a consistent place?

DanTup: > the `goldens_stable` folder is obsolete and something we should delete.

Ah, it's still set up in update_goldens by checking the Flutter version. Should that be gone too?

> What do you think about putting the syntax_highlighing goldens under `goldens/syntax_highlighting/` so that all the goldens for DevTools are in a consistent place?

If we're not maintaining two copies of goldens that seems reasonable to me (actually where I put them originally, then saw the suffix stuff). Although, it is nice to have the golden alongside the source file it goes with - but that might not make so much sense if inside `/goldens/`. WDYT?

kenzieschmoll: > Ah, it's still set up in update_goldens by checking the Flutter version. Should that be gone too?

Yes that logic should be changed. We should just make sure the the current flutter version matches the flutter version we pin to (flutter-version.txt) in the update_goldens script. We perform this check at the beginning of the build_release.sh script, so we should be able to copy from there, but this can be done in a separate PR.

I think it's okay to have the test data source files under `test_data/` and the goldens under `goldens/`. This would be consistent with most other tests in DevTools.

DanTup: > Yes that logic should be changed. We should just make sure the the current flutter version matches the flutter version we pin to (flutter-version.txt) in the update_goldens script

Ah yes, there's already a TODO explaining that, which I apparently missed.

> I think it's okay to have the test data source files under `test_data/` and the goldens under `goldens/`. This would be consistent with most other tests in DevTools.

sgtm, done!

kenzieschmoll (approved):

PR #4328, opened 2022-08-01
Rewrite FrameTimeVisualizer in performance page

kenzieschmoll:
- The frame time visualizer now has bars that are sized proportionately to the time a frame phase consumed (build, layout, paint, etc. time). 
Screen Shot 2022-08-01 at 3 42 55 PM

- The concept of selection was removed since it will not be used
- Added tooltips and polished the UI for narrow widths
adaptive-frame-time-viz


packages/devtools_app/lib/src/screens/performance/panes/frame_analysis/frame_analysis_model.dart
```diff
@@ -66,19 +58,30 @@ class FrameAnalysis {
...
-          event.name?.caseInsensitiveEquals(FramePhaseType.layout.eventName) ??
-          false,
-    );
-    return FramePhase.layout(
-      events: <SyncTimelineEvent>[
-        if (layoutEvent != null) layoutEvent as SyncTimelineEvent,
-      ],
-    );
+    return FramePhase.layout(events: <SyncTimelineEvent>[]);
   }
```
kenzieschmoll, line 85: do not review this part here - this should be reviewed in https://github.com/flutter/devtools/pull/4327, and then will merge here once that PR lands.

packages/devtools_app/lib/src/screens/performance/panes/frame_analysis/frame_time_visualizer.dart
```diff
@@ -0,0 +1,325 @@
...
+        content,
+      ],
+    );
+  }
+
+  /// Returns a list of adjusted widths for each block.
+  ///
+  /// The adjusted widths will ensure each block is at least
+  /// [_FramePhaseBlock.minBlockWidth] wide, and will modify surrounding block
+  /// widths to accomodate. f
```
CoderDake, line 194: Remove extra f here

packages/devtools_app/lib/src/screens/performance/panes/frame_analysis/frame_time_visualizer.dart
```diff
@@ -0,0 +1,325 @@
...
+    final adjustedBlockWidths = <double>[];
+    for (var i = 0; i < unadjustedBlockWidths.length; i++) {
+      final blockWidth = unadjustedBlockWidths[i];
+      var adjustedWidth = blockWidth;
+      if (i == widestBlockIndex) {
+        adjustedWidth -= adjustment;
+      }
+      adjustedWidth = math.max(adjustedWidth, _FramePhaseBlock.minBlockWidth);
+      adjustedBlockWidths.add(adjustedWidth);
+    }
```
CoderDake, line 227:
```
final adjustedBlockWidths = unadjustedBlockWidths.map(b => math.max(b,  _FramePhaseBlock.minBlockWidth));
adjustedBlockWidths[widestBlockIndex] -= adjustment; 
```
I think you can simplify this with a map and an index modification :)

packages/devtools_app/lib/src/screens/performance/panes/frame_analysis/frame_time_visualizer.dart
```diff
@@ -0,0 +1,325 @@
...
+        )
+        .toList();
+
+    var adjustment = 0.0;
+    var widestBlockIndex = 0;
+    for (var i = 0; i < unadjustedBlockWidths.length; i++) {
+      final unadjustedWidth = unadjustedBlockWidths[i];
+      final currentWidestBlock = unadjustedBlockWidths[widestBlockIndex];
+      if (unadjustedWidth > currentWidestBlock) {
+        widestBlockIndex = i;
```
CoderDake, line 211: if all blocks are less than the min width, then our widest block will end up being less than the min width. Is that ok based on the context with these block resizes?

kenzieschmoll: Below where we calculate the adjustedWidth, we adjust so that the width is at least `_FramePhaseBlock.minBlockWidth`:

`adjustedWidth = math.max(adjustedWidth, _FramePhaseBlock.minBlockWidth);`

So it won't end up being less than the min width after the adjustment. And if we overflow when all blocks are ` _FramePhaseBlock.minBlockWidth` wide, then so be it, because that means the screen is more narrow than we intend to support.

packages/devtools_app/lib/src/screens/performance/panes/frame_analysis/frame_time_visualizer.dart
kenzieschmoll: subtracting the adjustment needs to happen before we perform the `max` calculation so that we do not end up with an `adjustedWidth < _FramePhaseBlock.minBlockWidth` for the widest block index. 

I can simplify this though with another call to `max` when assigning `adjustedBlockWidths[widestBlockIndex]`.

packages/devtools_app/lib/src/screens/performance/panes/frame_analysis/frame_time_visualizer.dart
kenzieschmoll: done

CoderDake (approved):

PR #4329, opened 2022-08-02
add key and value types to VmInfoCard mapEntries

migueltylee:
Adds the type of MapEntry key and value to the VMInfoCard widget parameter 'rowKeyValues' - MapEntry<String,Widget>. It was previously not specified and thus, it was considered as dynamic by the compiler. 
Also added the implementation of PreferredSizeWidget to VMInfoCard.

**Issues fixed:**
The compiler now knows what type to expect for the key-value pairs in the VMInfoCard.

### DIFF
--- a/packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
+++ b/packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
@@ -21,7 +21,7 @@ import 'vm_service_private_extensions.dart';
 ///
 /// `table` is a widget (typically a table) that is to be displayed after the
 /// rows specified for `rowKeyValues`.
-class VMInfoCard extends StatelessWidget {
+class VMInfoCard extends StatelessWidget implements PreferredSizeWidget {
   const VMInfoCard({
     required this.title,
     this.rowKeyValues,
@@ -29,7 +29,7 @@ class VMInfoCard extends StatelessWidget {
   });
 
   final String title;
-  final List<MapEntry>? rowKeyValues;
+  final List<MapEntry<String, Widget>>? rowKeyValues;
   final Widget? table;
 
   @override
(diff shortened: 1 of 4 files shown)


bkonyi (commented):
Is there any reason we shouldn't just specify the types to be `MapEntry<String, Object?>` everywhere instead of requiring values to be a `Widget`?

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -42,6 +42,33 @@ class VMInfoCard extends StatelessWidget {
...
+    }
+    return Size.fromHeight(
+      areaPaneHeaderHeight +
+          (rowKeyValues?.length ?? 0) * defaultRowHeight +
+          defaultSpacing,
+    );
+  }
+}
+
+MapEntry<String, Widget> stringWidgetMapEntry(
```
kenzieschmoll, line 59: naming nit: maybe `_selectableTextMapEntry` instead of string widget

kenzieschmoll:
> Is there any reason we shouldn't just specify the types to be MapEntry<String, Object?> everywhere instead of requiring values to be a Widget?

We have cases where we want the value of each entry to be a widget. If we leave this as Object, then we have to do some forking in the build methods like `if (value is Widget)` use the widget and `if (value is String)` use some SelectableText, etc. 
Using type `Widget` explicitly avoids this sort of dynamic behavior. 

One thing we could consider that would be even better than passing around Widgets would be to pass in a builder: `MapEntry<String, Widget Function()>`. This way, where the builder would be used, we'd write `entry.value()` instead of `entry.value`, and the Widget would be constructed as part of the Flutter lifecycle instead of beforehand. It is generally a bad idea in Flutter to construct widgets outside of the `build` lifecycle method and to pass them around. Building a Widget outside of the `build` method could prevent things from updating properly when things like State or BuildContext change.

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -42,6 +42,33 @@ class VMInfoCard extends StatelessWidget {
...
+    if (table != null) {
+      return Size.infinite;
+    }
+    return Size.fromHeight(
+      areaPaneHeaderHeight +
+          (rowKeyValues?.length ?? 0) * defaultRowHeight +
+          defaultSpacing,
+    );
+  }
+}
```
kenzieschmoll, line 57: was this intended to be part of another PR? it seems unrelated to the rest of the changes

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
migueltylee: I'll change it. It still needs to be public though

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
migueltylee: It will be useful to have the VmInfoCard have a size, to use it directly in the object inspector and potentially other screens in the vm tools tab. I thought I would add that in this PR, since it is also a change to the VmInfoCard.

migueltylee:
Refactored the VmInfoCard widget to now take 'rowKeyValues' of type MapEntry<String, Widget Function(BuildContext)>, to ensure the widgets are constructed inside the VmInfoCard build method.

kenzieschmoll (approved):

PR #4332, opened 2022-08-02
Add a dropdown to App Size analysis tab

elliette:
For deferred apps, tab auto-opens to "Entire app" with a dropdown to switch between the entire app, the main unit, and the deferred units:

deferred-dropdown

@carolynqu is working on adding equivalent functionality to the diff tab.


packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -139,6 +146,22 @@ class AppSizeController {
   TreemapNode? _decreasedDiffTreeRoot;
   TreemapNode? _combinedDiffTreeRoot;
 
+  Map<String, dynamic>? get _activeAppUnit {
```
kenzieschmoll, line 149: maybe rename to `_dataForAppUnit` or `_activeAppData` since it doesn't return an instance of `AppUnit`, but rather the data associated with the `AppUnit`

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -238,22 +270,64 @@ class AppSizeController {
     changeAnalysisJsonFile(jsonFile);
 
     // Set deferred app flag.
-    isDeferredApp =
-        deferredLoadingSupportEnabled && processedJson['n'] == 'ArtificialRoot';
+    _isDeferredApp.value =
+        deferredLoadingSupportEnabled && _hasDeferredInfo(processedJson);
+
+    if (isDeferredApp.value) {
+      _deferredOnly = _extractDeferredUnits({...processedJson});
```
kenzieschmoll, line 277: can we just write `processedJson` instead of `{...processedJson}`?

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -238,22 +270,64 @@ class AppSizeController {
...
+      _loadApp(_activeAppUnit!);
+    } else {
+      _loadApp(processedJson);
+    }
+
+    _processingNotifier.value = false;
+  }
 
-    // Set root name.
-    processedJson['n'] = isDeferredApp ? 'Entire app' : 'Root';
```
kenzieschmoll, line 245: do we still need to set `processedJson['n']` to 'Root' somewhere for non-deferred apps?

packages/devtools_app/lib/src/screens/app_size/app_size_screen.dart
```diff
@@ -590,3 +629,17 @@ extension DiffTreeTypeExtension on DiffTreeType {
     }
   }
 }
+
+extension AppUnitExtension on AppUnit {
+  String get display {
```
kenzieschmoll, line 634: I think you can do this on the enum directly now with enhanced enums

kenzieschmoll (commented):
can we add a simple test to make sure that we are extracting the correct subsets of json for the different AppUnits?

packages/devtools_app/test/app_size/app_size_screen_test.dart
```diff
@@ -120,6 +128,34 @@ void main() {
...
+      // Verify the state of the splitter.
+      final splitFinder = find.byType(Split);
+      expect(splitFinder, findsOneWidget);
+      final Split splitter = tester.widget(splitFinder);
+      expect(splitter.initialFractions[0], equals(0.67));
+      expect(splitter.initialFractions[1], equals(0.33));
+
+      // Verify the dropdown for selecting deferred units exists.
+      final appUnitDropdownFinder =
+          find.byKey(AppSizeScreen.appUnitDropdownKey);
```
kenzieschmoll, line 156: I would also verify that this dropdown doesn't show up when we are on the diff tab

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -580,6 +582,11 @@ class AppSizeController {
       caption: isDeferred ? '(Deferred)' : null,
     )..addAllChildren(children);
   }
+
+  @visibleForTesting
+  void setDeferredLoadingSupportEnabled(bool enabled) {
```
kenzieschmoll, line 587: we don't need to add a method for this. You can just set `deferredLoadingSupportEnabled` to true directly from the test setup

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
elliette: Good point, switched to `_dataForAppUnit`

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
elliette: I'm using `{...processedJson}` to make a deep-copy of the `processedJson`, so that we aren't modifying the underlying data

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
elliette: Yes :) The test caught this too lol

packages/devtools_app/lib/src/screens/app_size/app_size_screen.dart
elliette: Good point, done!

packages/devtools_app/test/app_size/app_size_screen_test.dart
elliette: Done

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
elliette: Oooh how do you mean? It is currently a top level variable in the file

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
kenzieschmoll: You can also use `Map.from` which might make the intent more clear

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
kenzieschmoll: I mean in the test where you are currently calling `controller.setDeferredLoadingSupportEnabled(true);`, instead call `deferredLoadingSupportEnabled = true;`. We do this for several other feature flags in other screens.

elliette: Ah got it, done!

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
elliette: Done!

packages/devtools_app/test/app_size/app_size_screen_test.dart
```diff
@@ -120,6 +128,34 @@ void main() {
       expect(splitter.initialFractions[0], equals(0.67));
       expect(splitter.initialFractions[1], equals(0.33));
     });
+
+    testWidgetsWithWindowSize('builds deferred content', windowSize,
```
kenzieschmoll, line 132: can we add a test to verify we can change the app unit dropdown and that the data in the treemap / table changes when we do so?

elliette: Yes, added!

elliette:
> can we add a simple test to make sure that we are extracting the correct subsets of json for the different AppUnits?

Yes, done!

kenzieschmoll (approved):

PR #4337, opened 2022-08-03
Write tests for frame hints

kenzieschmoll:
This feature was previously untested, and is now tested with widget tests. The business logic for this feature is already tested in frame_analysis_model_test.dart, so we can safely use mocks for these UI tests.

### DIFF
--- a/packages/devtools_app/test/performance/frame_analysis/frame_hints_test.dart
+++ b/packages/devtools_app/test/performance/frame_analysis/frame_hints_test.dart
@@ -1,0 +1,326 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:devtools_app/src/config_specific/ide_theme/ide_theme.dart';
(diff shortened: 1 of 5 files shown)


packages/devtools_app/test/performance/frame_analysis/frame_hints_test.dart
```diff
@@ -0,0 +1,326 @@
...
+            frameAnalysis: frameAnalysis,
+            enhanceTracingController: mockEnhanceTracingController,
+          ),
+          performance: PerformanceController(),
+        ),
+      );
+      expect(find.byType(FrameHints), findsOneWidget);
+    }
+
+    void _mockFrameAnalysis({
```
polina-c, line 76: '_' in the beginning does not have effect here, because it is in method body

polina-c (approved):

packages/devtools_app/test/performance/frame_analysis/frame_hints_test.dart
kenzieschmoll: it doesn't do anything for privacy, but it does avoid a naming collision with the mock object `mockFrameAnalysis`

PR #4338, opened 2022-08-03
Adding ArtificialRoot to non-deferred apps for diff table

carolynqu:
When using the DiffTable to compare a non-deferred and deferred app, the main app roots are not at the same location. 

In order to get an accurate comparison between the two apps, the non-deferred app will add its own ArtificialRoot that deferred apps have.

### DIFF
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
@@ -457,7 +457,7 @@ class AppSizeController {
     changeOldDiffFile(oldFile);
     changeNewDiffFile(newFile);
 
-    diffMap['n'] = 'Root';
+    diffMap['n'] = isDeferredApp.value ? 'Entire App' : 'Root';
 
     // TODO(peterdjlee): Try to move the non-active tree generation to separate isolates.
     _combinedDiffTreeRoot = generateDiffTree(
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
@@ -17,7 +17,7 @@ import '../../ui/colors.dart';
 import 'app_size_screen.dart';
 
 // Temporary feature flag for deferred loading.
-bool deferredLoadingSupportEnabled = false;
+bool deferredLoadingSupportEnabled = true;
 
 enum DiffTreeType {
   increaseOnly,
@@ -585,7 +585,7 @@ class AppSizeController {
     }
 
     // If none of the children matched the diff tree type
-    if (totalByteSize == 0) {
+    if (totalByteSize == 0 && !isDeferredApp.value) {
       return null;
     } else {
       return _buildNode(
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
@@ -17,7 +17,7 @@ import '../../ui/colors.dart';
 import 'app_size_screen.dart';
 
 // Temporary feature flag for deferred loading.
-bool deferredLoadingSupportEnabled = true;
+bool deferredLoadingSupportEnabled = false;
 
 enum DiffTreeType {
   increaseOnly,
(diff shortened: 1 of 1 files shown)


elliette:
Let's change the check on line 588 from:

```
    if (totalByteSize == 0) {
      return null;
    }
```
 
to 
```
    if (totalByteSize == 0 && !isDeferredApp.value) {
      return null;
    } 
```

That way if the totalByteSize stays the same (for example, part of the main unit is moved to deferred units) we still build the diff view.

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -67,6 +67,8 @@ class AppSizeController {
   static const identicalFilesError =
       'Failed to load diff: OLD and NEW files are identical.';
 
+  static const deferredRootName = 'ArtificialRoot';
```
elliette, line 70: Let's rename this to `artificialRootNodeName`. Let's also add another constant `mainNodeName` which is equal to `Main`

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -380,8 +382,26 @@ class AppSizeController {
...
     Map<String, dynamic> diffMap;
     if (oldFile.isAnalyzeSizeFile && newFile.isAnalyzeSizeFile) {
+      var oldFileJson = oldFile.data as Map<String, dynamic>;
+      var newFileJson = newFile.data as Map<String, dynamic>;
+
+      if (_hasDeferredInfo(oldFileJson) || _hasDeferredInfo(newFileJson)) {
+        _isDeferredApp.value = deferredLoadingSupportEnabled;
+
+        if (!_hasDeferredInfo(oldFileJson)) {
+          oldFileJson = <String, dynamic>{
```
elliette, line 392: Let's pull this logic out into a private method (eg, `_wrapInArtificialRoot`) which takes the json file as a parameter. That way we don't have to repeat the code here and below. 

I think we also have to rename the first node from `Root` to `Main` before we wrap it, ie:

```
          oldFileJson['n'] = 'Main';
          oldFileJson = <String, dynamic>{
            'n': deferredRootName,
            'children': [oldFileJson]
          };
```

elliette:
> Let's change the check on line 588 from:
> 
> ```
>     if (totalByteSize == 0) {
>       return null;
>     }
> ```
> 
> to
> 
> ```
>     if (totalByteSize == 0 && !isDeferredApp.value) {
>       return null;
>     } 
> ```
> 
> That way if the totalByteSize stays the same (for example, part of the main unit is moved to deferred units) we still build the diff view.

Thinking about this more, we only want to show 0 byte size changes on the root node (not on the child nodes). Instead, let's add another parameter to both `generateDiffTree` and `_buildNodeWithChildren` (eg `skipNodesWithNoByteSizeChange`) which defaults to `true`. When we [generate the diff trees here](https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart#L444-L455)`skipNodesWithNoByteSizeChange` should be false if deferred app is true. For every recursive call on child nodes, `skipNodesWithNoByteSizeChange` should be the default value of true.

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -17,7 +17,7 @@ import '../../ui/colors.dart';
 import 'app_size_screen.dart';
 
 // Temporary feature flag for deferred loading.
-bool deferredLoadingSupportEnabled = false;
+bool deferredLoadingSupportEnabled = true;
```
elliette, line 20: Switch to false :)

elliette (approved):
LGTM once the trailing comma the linter is complaining about is added

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -67,6 +67,10 @@ class AppSizeController {
   static const identicalFilesError =
       'Failed to load diff: OLD and NEW files are identical.';
 
+  static const artificialRootNodeName = 'ArtificialRoot';
+
+  static const mainNodeName = 'Main';
```
kenzieschmoll, line 72: we should use the functionality @elliette added in https://github.com/flutter/devtools/pull/4346 once that lands.

PR #4340, opened 2022-08-04
Refactor RequestableSizeWidget

migueltylee:
Refactor of RequestableSizeWidget in vm_developer_common_widgets.dart. The widget now requires a ValueListenable<bool> fetching to know when the data is being fetched, and if that is the case display a circular progress indicator.

**Issues fixed:**
- The logic to display the circular progress indicator was previously handled outside the widget.
- The CircularProgressIndicator diameter was slightly bigger than the flex widget that contained it.

**Screenshots:**
before:
prev
after:
now


packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -158,39 +158,55 @@ class RequestDataButton extends IconLabelButton {
...
+        if (fetching) {
+          return const AspectRatio(
+            aspectRatio: 1,
+            child: Padding(
+              padding: EdgeInsets.all(densePadding),
+              child: CircularProgressIndicator(),
+            ),
+          );
+        } else {
+          final size = requestedSize.value;
```
bkonyi, line 188: Why does this need to be a `ValueListenable` if we're never listening, only accessing it's value?

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -158,39 +158,55 @@ class RequestDataButton extends IconLabelButton {
 
 /// Displays a RequestDataButton if [requestedSize] is null, otherwise displays
 /// the requestable size and a ToolbarRefresh button next to it,
-/// to request that size again if required.
+/// to request that size again if required. Displays a CircularProgressIndicator
```
kenzieschmoll, line 161: new line between first sentence of dart doc and remaining sentences

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
migueltylee: Because if it is not listenable, the value will remain the same after the first time it is passed into the widget, even if it changes in the VmObject it belongs to

migueltylee: I'm going to go with option 2. Thanks!

kenzieschmoll (approved):

bkonyi (approved):

PR #4346, opened 2022-08-04
App Size panel clean-up

elliette:
* Allow loading web apps in the app size panel
* Save commonly used string values as variables


packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -19,6 +19,12 @@ import 'app_size_screen.dart';
 // Temporary feature flag for deferred loading.
 bool deferredLoadingSupportEnabled = false;
 
+const artificialRootNodeName = 'ArtificialRoot';
+const entireAppNodeName = 'Entire App';
+const deferredNodeName = 'Deferred';
+const mainNodeName = 'Main';
+const rootNodeName = 'Root';
```
kenzieschmoll, line 26: should these be private?

kenzieschmoll (approved):
lgtm w/ one comment

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
elliette: Good point, done!

PR #4348, opened 2022-08-04
Adding drop down for Diff Table

carolynqu:
[to be done]

### DIFF
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
@@ -17,7 +17,7 @@ import '../../ui/colors.dart';
 import 'app_size_screen.dart';
 
 // Temporary feature flag for deferred loading.
-bool deferredLoadingSupportEnabled = false;
+bool deferredLoadingSupportEnabled = true;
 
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -61,6 +61,42 @@ enum AppUnit {
...
+  entireDecrease,
+  mainCombined,
+  mainIncrease,
+  mainDecrease,
+  deferredCombined,
+  deferredIncrease,
+  deferredDecrease,
+  entireCombined;
+
+  String get display {
```
elliette, line 75: This `display` getter can be removed - we aren't displaying the string anywhere so no need to have the getter

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -73,6 +109,8 @@ class AppSizeController {
   static const identicalFilesError =
       'Failed to load diff: OLD and NEW files are identical.';
 
+  static const deferredRootName = 'ArtificialRoot';
```
elliette, line 112: This should no longer be necessary with the changes in https://github.com/flutter/devtools/pull/4346

PR #4349, opened 2022-08-05
Script display initial implementation

migueltylee:
This PR adds the functionality to display data associated to Script VM objects within the object inspector tool.
This includes the general VM object information, as well as the uri, library, and load time.

**Screenshots for reference:**
script


packages/devtools_app/lib/src/screens/vm_developer/vm_object_model.dart
```diff
@@ -208,7 +209,18 @@ class ScriptObject extends VmObject {
...
+  String? get name => (obj.uri ?? scriptRef?.uri)?.split('/').last;
+
+  late final DateTime? loadTime;
+
+  @override
+  Future<void> initialize() async {
+    await super.initialize();
+
+    if (obj.loadTime != null) {
+      loadTime = DateTime.fromMillisecondsSinceEpoch(obj.loadTime!);
```
bkonyi, line 221: This should be relatively cheap to create, so I think we can just update `loadTime` to be a getter that does this operation. That way we can completely remove the need for an override of `initialize()`, at least for now.

packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
```diff
@@ -256,3 +256,9 @@ enum GuardClassKind {
     }
   }
 }
+
+/// An extension on [Script] which allows for access to VM internal fields.
+extension ScriptPrivateViewExtension on Script {
+  static const loadTimeKey = '_loadTime';
```
bkonyi, line 262: Nit: make this key private.

packages/devtools_app/lib/src/screens/vm_developer/vm_object_model.dart
```diff
@@ -208,7 +209,18 @@ class ScriptObject extends VmObject {
   SourceLocation? get _sourceLocation => null;
 
   @override
-  String? get name => null;
+  String? get name => (obj.uri ?? scriptRef?.uri)?.split('/').last;
```
kenzieschmoll, line 212: `.split('/').last` is used in several places in DevTools. Create a helper in utils.dart `fileNameFromUri(String? uri)` that returns `uri?.split('/').last`. And then file an issue to use this helper elsewhere in devtools (you can link to the TODO from a comment where you define the utility method).

packages/devtools_app/lib/src/screens/vm_developer/vm_object_model.dart
```diff
@@ -208,7 +209,18 @@ class ScriptObject extends VmObject {
...
+
+  late final DateTime? loadTime;
+
+  @override
+  Future<void> initialize() async {
+    await super.initialize();
+
+    if (obj.loadTime != null) {
```
kenzieschmoll, line 220: pull `obj.loadTime` into a local var so that it does not need to be null checked twice

packages/devtools_app/lib/src/screens/vm_developer/vm_script_display.dart
```diff
@@ -0,0 +1,42 @@
...
+    required this.script,
+  });
+
+  final ScriptObject script;
+
+  @override
+  Widget build(BuildContext context) => VmObjectDisplayBasicLayout(
+        object: script,
+        generalDataRows: _scriptDataRows(script),
+      );
```
kenzieschmoll, line 23: nit: prefer using brackets for returns that are more than one line, and especially for flutter build methods

packages/devtools_app/lib/src/screens/vm_developer/vm_object_model.dart
migueltylee: Done

packages/devtools_app/lib/src/screens/vm_developer/vm_object_model.dart
migueltylee: Done

packages/devtools_app/lib/src/screens/vm_developer/vm_script_display.dart
migueltylee: Done

packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
migueltylee: Done

packages/devtools_app/lib/src/screens/vm_developer/vm_object_model.dart
migueltylee: Done
https://github.com/flutter/devtools/issues/4360

packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
```diff
@@ -256,3 +256,9 @@ enum GuardClassKind {
     }
   }
 }
+
+/// An extension on [Script] which allows for access to VM internal fields.
+extension ScriptPrivateViewExtension on Script {
+  static const _loadTimeKey = '_loadTime';
+  int? get loadTime => json![_loadTimeKey];
```
bkonyi, line 263: This should never be null based on the implementation in the VM. We should be safe to add a null assertion here.

bkonyi (approved):

PR #4350, opened 2022-08-05
Library display initial implementation

migueltylee:
This PR adds the functionality to display VM data associated to Library VM objects within the object inspector tool.
This includes the general VM object information, as well as the library uri, VM name, and an expandable tile displaying a list of its dependencies.

**Screenshot for reference:**
lib


packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -139,6 +139,10 @@ List<Widget> _prettyRows(BuildContext context, List<Row> rows) {
   ];
 }
 
+List<Widget> prettyRows(BuildContext context, List<Row> rows) {
+  return <Widget>[..._prettyRows(context, rows)];
```
bkonyi, line 143: Why are we expanding a list into a new list here instead of just returning the list?

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
```diff
@@ -0,0 +1,116 @@
...
+    required this.dependencies,
+  });
+
+  final List<LibraryDependency> dependencies;
+
+  List<Row> dependencyRows(BuildContext context) {
+    final rows = <Row>[];
+
+    for (final dep in dependencies) {
+      final row = Row(
```
bkonyi, line 67: Nit: this can be done more concisely like this:

```dart
List<Row> dependencyRows(BuildContext context) {
  return <Row>[
    for (final dep in dependencies)
      Row(
        children: [
          ...
        ],
     ),
  ];
}
```

bkonyi: However, I'm not sure why we need rows with a single element here. Can this not just return a list of `SelectableText`?

packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
```diff
@@ -256,3 +256,8 @@ enum GuardClassKind {
     }
   }
 }
+
+/// An extension on [Library] which allows for access to VM internal fields.
+extension LibraryPrivateExtension on Library {
+  String? get vmName => json!['_vmName'];
```
bkonyi, line 262: Nit: put `_vmName` into a private variable and use that as the key.

packages/devtools_app/lib/src/screens/vm_developer/vm_object_model.dart
```diff
@@ -195,6 +195,15 @@ class LibraryObject extends VmObject {
 
   @override
   String? get name => obj.name;
+
+  late final String? vmName;
```
bkonyi, line 199: Make this a getter on `obj` and remove the `initialize()` override below.

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
migueltylee: Yeah... returning the list directly is a cleaner way. But, I'm going make the function `_prettyRows` public now

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
migueltylee: It could, but they would still have to be put in rows to show them in the expansion tile. Also, `_prettyRows` takes a list of Rows

packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
migueltylee: Making `_vmNameKey` a top level constant

packages/devtools_app/lib/src/screens/vm_developer/vm_object_model.dart
migueltylee: Done

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
```diff
@@ -0,0 +1,114 @@
...
+      ),
+      selectableTextBuilderMapEntry(
+        'VM Name',
+        library.vmName,
+      ),
+    ];
+  }
+}
+
+// An expandable tile displaying a list of library dependencies
```
kenzieschmoll, line 55: nit: use /// for dart doc

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
```diff
@@ -0,0 +1,114 @@
...
+class LibraryDependencies extends StatelessWidget {
+  const LibraryDependencies({
+    required this.dependencies,
+  });
+
+  final List<LibraryDependency> dependencies;
+
+  List<Row> dependencyRows(BuildContext context) {
+    return <Row>[
```
kenzieschmoll, line 64: put `Theme.of(context).fixedFontStyle` into a local var before returning this row so that we do not call `Theme.of(context)` on each iteration of the loop

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
```diff
@@ -0,0 +1,114 @@
...
+                _dependencyDescription(dep),
+                style: Theme.of(context).fixedFontStyle,
+              ),
+            ),
+          ],
+        )
+    ];
+  }
+
+  String _dependencyDescription(LibraryDependency dependency) {
```
kenzieschmoll, line 79: make this public (can be `@visibleForTesting`) and add a test for it

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
```diff
@@ -0,0 +1,114 @@
...
+    if (dependency.isImport != null) {
+      description.write(dependency.isImport! ? 'import ' : 'export ');
+    }
+
+    description.write(
+      dependency.target?.name ?? dependency.target?.uri ?? '<Library name>',
+    );
+
+    if (dependency.prefix != null) {
+      final prefix = dependency.prefix!;
```
kenzieschmoll, line 91: declare this var before the if check, and they you won't have to null assert it after checking that it is not null in the `if` statement.

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
migueltylee: Done

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
migueltylee: Done

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
migueltylee: Done

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
bkonyi: Are you sure they need to be in rows for the expansion tiles? As far as I can tell based on the code for `ExpansionTile`, it shouldn't matter.

However, you're right about `_prettyRows` only taking rows, so I guess this is fine.

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
```diff
@@ -61,46 +61,53 @@ class LibraryDependencies extends StatelessWidget {
...
+          ],
+        )
+    ];
   }
 
-  String _dependencyDescription(LibraryDependency dependency) {
+  @visibleForTesting
+  String dependencyDescription(LibraryDependency dependency) {
     final description = StringBuffer();
+    final void Function() addSpace =
```
bkonyi, line 84: Nit: drop the type on the LHS.

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
```diff
@@ -61,46 +61,53 @@ class LibraryDependencies extends StatelessWidget {
...
       dependency.target?.name ?? dependency.target?.uri ?? '<Library name>',
     );
 
-    if (dependency.prefix != null) {
-      final prefix = dependency.prefix!;
+    addSpace();
+
+    final prefix = dependency.prefix;
+
+    if (prefix != null) {
```
bkonyi, line 101: Nit: collapse this into a single conditional statement.

```dart
if (prefix != null && prefix.isNotEmpty) {
...
}
```

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
migueltylee: I made it a function now

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
migueltylee: Done

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
migueltylee: I've made the tests for the LibraryDependencies widget overall. If I only test  `dependencyDescription` then I would have to make an empty LibraryDependencies widget `final widget = LibraryDependencies(dependencies: [])` and then call `widget.dependencyDescription(dep...)`. 
Would you like me to change to this way or is it okay to keep it private and test for the LibraryDependencies widget?

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
migueltylee: You're right! The children are put in a Column within the ExpansionTile

bkonyi (approved):

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
kenzieschmoll: Ah I see, I didn't realize this was in a widget class at first. A better option would be to make this an extension method instead that you can easily test. So in this file you'd write

```dart
extension LibraryDependencyExtension on LibraryDependency {
  String get description {
    ...
  }
}
```

packages/devtools_app/test/vm_developer/object_inspector/object_viewport_test.dart
```diff
@@ -253,3 +262,29 @@ void main() {
     });
   });
 }
+
+// testWidgets('test for Field Object', (WidgetTester tester) async {
```
kenzieschmoll, line 266: are these tests supposed to be uncommented? or removed?

kenzieschmoll (approved):
lgtm with one question about some commented out code

packages/devtools_app/test/vm_developer/object_inspector/object_viewport_test.dart
migueltylee: Sorry I forgot to delete them 😅

PR #4351, opened 2022-08-05
Fix typos

kenzieschmoll:
These are issues for g3 common_typos 

@jacob314 @pq is a linter equivalent of this check something that we have currently or that is on the roadmap? I suspect this is something almost every package that rolls from github to g3 will run into. We've hit this issue quite often.


CoderDake (approved):

pq:
Spell-checking isn't something we've dug into, relying instead on other tools.

@DanTup: is there a spellchecker that plays nice w/ Dart that you're aware of?

DanTup:
Not that I've seen. There used to be some Microsoft-published spell checking extensions for VS Code, but they were deprecated:

https://github.com/microsoft/vscode-spell-check

It is something that I think about often though when I'm trying to keep my code and comments consistent with US spellings 😄

DanTup:
Oh, I remembered seeing something in the SDK when someone fixed some typos.. seems CFE-specific though:

https://github.com/dart-lang/sdk/blob/master/pkg/front_end/test/spell_checking_utils.dart

pq:
Thanks Danny!

PR #4354, opened 2022-08-05
Adding drop down to diff table

carolynqu:
For deferred apps, in the diff-table, the tab auto-opens to "Entire app" with a dropdown to switch between the entire app, the main unit, and the deferred units:

https://user-images.githubusercontent.com/67798872/183158040-548f1261-11b0-4201-aec1-811b114ca01a.mp4

### DIFF
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
@@ -17,7 +17,7 @@ import '../../ui/colors.dart';
 import 'app_size_screen.dart';
 
 // Temporary feature flag for deferred loading.
-bool deferredLoadingSupportEnabled = false;
+bool deferredLoadingSupportEnabled = true;
 
 const _artificialRootNodeName = 'ArtificialRoot';
 const _entireAppNodeName = 'Entire App';
@@ -59,6 +59,20 @@ enum AppUnit {
         return _entireAppNodeName;
     }
   }
+}
+
+class DiffTreeMap {
+  const DiffTreeMap({
+    required this.combined,
+    required this.increaseOnly,
+    required this.decreaseOnly,
+  });
+
+  final TreemapNode? combined;
+
+  final TreemapNode? increaseOnly;
+
+  final TreemapNode? decreaseOnly;
 }
 
 class AppSizeController {
@@ -252,9 +266,13 @@ class AppSizeController {
   ValueListenable<AppUnit> get selectedAppUnit => _selectedAppUnit;
   final _selectedAppUnit = ValueNotifier<AppUnit>(AppUnit.entireApp);
 
-  void changeSelectedAppUnit(AppUnit appUnit) {
+  void changeSelectedAppUnit(AppUnit appUnit, Key tabKey) {
     _selectedAppUnit.value = appUnit;
-    _loadApp(_dataForAppUnit!);
+    if (tabKey == AppSizeScreen.analysisTabKey) {
+      _loadApp(_dataForAppUnit!);
+    } else {
+      print('selected $appUnit in diff view');
+    }
   }
 
   /// Notifies that the json files are currently being processed.
@@ -343,7 +361,8 @@ class AppSizeController {
   ) {
     if (_hasDeferredInfo(jsonFile)) {
       jsonFile['children'] = _extractChildren(jsonFile)
-          .where((child) => child['isDeferred'] == true);
+          .where((child) => child['isDeferred'] == true)
+          .toList();
       jsonFile['n'] = _deferredNodeName;
     }
     return jsonFile;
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -458,25 +519,91 @@ class AppSizeController {
     diffMap['n'] = isDeferredApp.value ? _entireAppNodeName : _rootNodeName;
 
     // TODO(peterdjlee): Try to move the non-active tree generation to separate isolates.
-    _combinedDiffTreeRoot = generateDiffTree(
-      diffMap,
-      DiffTreeType.combined,
-      skipNodesWithNoByteSizeChange: !isDeferredApp.value,
+    // Entire app or root (for non-deferred):
+    final diffTreeMap = _generateDiffTrees(diffMap);
```
elliette, line 523: Let's actually set the `diffTreeMap` (and the `mainDiffTreeMap`, etc) as a top-level variable. So instead of:
```
    _combinedDiffTreeRoot
    _increasedDiffTreeRoot
    _decreasedDiffTreeRoot
  _mainCombinedDiffTreeRoot
   etc..
``` 
We only have:

```
_diffTreeMap
_mainDiffTreeMap
_deferredDiffTreeMap
```

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -161,21 +175,64 @@ class AppSizeController {
...
   TreemapNode? get _activeDiffRoot {
-    switch (_activeDiffTreeType.value) {
-      case DiffTreeType.increaseOnly:
-        return _increasedDiffTreeRoot;
-      case DiffTreeType.decreaseOnly:
-        return _decreasedDiffTreeRoot;
-      case DiffTreeType.combined:
-      default:
-        return _combinedDiffTreeRoot;
+    final isDeferred = _isDeferredApp.value;
```
elliette, line 178: If we follow the suggestion I made below, we can simplify this getter considerably. We can add another getter for the `_activeDiffMap` which returns either the `_diffMap`, `_mainDiffMap`, or `_deferredDiffMap`. Then, this `getter` only needs to return either `_activeDiffMap.combined`, `_activeDiffMap.increaseOnly` or `_activeDiffMap.decreaseOnly`

elliette: Looks better, but adding an `_activeDiffMap` getter (as described above) will clean this up even more!

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -160,21 +174,39 @@ class AppSizeController {
...
+      case AppUnit.entireApp:
+      default:
+        return _diffTreeMap;
+    }
+  }
+
   TreemapNode? get _activeDiffRoot {
-    switch (_activeDiffTreeType.value) {
+    final diffTreeType = _activeDiffTreeType.value;
+    final diffMap = _activeDiffMap;
```
elliette, line 193: Let's return `null` if `_activeDiffMap` is `null` (instead of null-asserting with `!`)

elliette:
Let's add a test case for the diff view with deferred units (it should be pretty similar to the test here: https://github.com/flutter/devtools/pull/4332)

packages/devtools_app/test/app_size/app_size_screen_test.dart
```diff
@@ -53,6 +55,18 @@ void main() {
...
   );
 
+  final diffDeferredAppFile = DevToolsJsonFile(
+    name: 'lib/src/app_size/stub_data/diff_deferred_app.dart',
+    lastModifiedTime: lastModifiedTime,
+    data: json.decode(diffDeferredApp),
+  );
+
+  final diffNonDeferredAppFile = DevToolsJsonFile(
+    name: 'lib/src/app_size/stub_data/diff_deferred_app.dart',
```
elliette, line 65: Should this be `diff_no_deferred_app`?

elliette (approved):

packages/devtools_app/lib/src/screens/app_size/app_size_screen.dart
```diff
@@ -208,9 +208,7 @@ class _AppSizeBodyState extends State<AppSizeBody>
                   ),
                   Row(
                     children: [
-                      if (currentTab.key == AppSizeScreen.analysisTabKey &&
-                          isDeferredApp)
-                        _buildAppUnitDropdown(),
+                      if (isDeferredApp) _buildAppUnitDropdown(currentTab.key!),
```
kenzieschmoll, line 211: add some spacing here:

```dart
if (isDeferredApp) ...[
  _buildAppUnitDropdown(currentTab.key!),
],
```

carolynqu: Added spacing so that it only impacts the visuals when both drop downs are there.

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -262,13 +262,15 @@ class AppSizeController {
     _diffCallGraphRoot.value = null;
     _oldDiffCallGraph = null;
     _newDiffCallGraph = null;
+    _isDeferredApp.value = false;
```
elliette, line 265: Let's split this into a separate PR so we can add a test for it at the same time

Issue #3896, opened 2022-03-18
Dart DevTools caching issues


elliette:
Spinning this out from https://github.com/flutter/flutter/issues/100247, for investigation into #2 as described in https://github.com/flutter/flutter/issues/100247#issuecomment-1071517924. See also https://github.com/flutter/flutter/issues/100247#issuecomment-1071160197.

elliette:
I haven’t been able to reproduce this issue since the first time (described [here](https://github.com/flutter/flutter/issues/100247#issuecomment-1071160197)), so I’m not sure exactly what happened. But I do have a few theories about things that could be causing issues with caching.

1. The `index.html` file has the response header `cache-control: max-age=900`. This means it is cached in the browser cache for 900 seconds/15 minutes (this is set [here](https://github.com/dart-lang/sdk/blob/167970dea7c12a996f61891b25fee42c6fee45d7/pkg/dds/lib/devtools_server.dart#L276-L280)). As mentioned above, it is the version number in `index.html` that determines whether or not the service worker should update. This means that a user could switch to a new Flutter/DevTools version, but if they recently had DevTools open the `index.html` would still be cached and therefore the service worker would not update as expected. We should set this to 0 to make sure we are always using the newest `index.html`.

2. Other assets (for example, `main.dart.js`) have a `last-modified` header, which is set by the StaticHandler that the DevTools server uses to serve web assets (see [here](https://github.com/dart-lang/shelf_static/blob/202ec1a53c9a830c17cf3b718d089cf7eba568ad/lib/src/static_handler.dart#L193)). The browser uses this to determine whether or not to use a copy cached in its cache, or request a new one. This means that the service worker might update, but the cache will still contain older assets because the browser doesn’t think they have been modified (this is probably often true, but I think it might make more sense to just request everything once the service worker updates to make sure we aren’t caching anything from the previous DevTools version). I think we should remove the `last-modified` header altogether.

3. Whenever a new service worker is activated, we have it claim all open DevTools tabs/windows (see [here](https://github.com/flutter/devtools/blob/master/packages/devtools_app/web/devtools_service_worker.js#L31-L34)). This was added so that if there is a new DevTools available, it is used everywhere (and replaces older DevTools). Thinking more about this, I don’t think this is what we want, because it causes the Service Worker to take over DevTools that are connected to apps running older versions of Flutter. Conversely, if a user downgrades their Flutter, you can have older versions of DevTools now connected to apps running on a newer version of Flutter.

[...]

PR #4356, opened 2022-08-08
Add some notes about configuring VS Code to run DevTools from local code

DanTup:
This adds some notes about running DevTools from local code in VS Code to test interactions with the server and embedded views.

It requires changes to the Dart extension that so far are only in the pre-release versions of the Dart extension (although I'd recommend any devs here use that version to help find any issues before they reach all users), v3.47:

Dart/Flutter pre-release versions

(note: although the screenshots shows both, there isn't currently a pre-release version of Flutter, the stable version is latest).

It would be useful to have someone other than me verify these steps work for them before landing (although be aware when testing this myself, I saw some [caching issues](https://github.com/flutter/devtools/issues/3896) that might complicated that).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,6 +84,26 @@ When you make changes on disk, you can hit `r` in your command-line to rebuild t
 refresh in your browser to see the changes. Hit `q` in the command line to terminate both the
 `flutter run` instance and the devtools server instance.
 
+## Development (VS Code Integration)
+
+To test integration with VS Code, you can instruct the Dart extension to run DevTools and the server from local code. You will need to have the Dart SDK source set up (see [dart-lang/sdk/CONTRIBUTING.md](https://github.com/dart-lang/sdk/blob/main/CONTRIBUTING.md#getting-the-code)) and you will need version v3.47 or newer of the Dart extension for VS Code.
+
+Open your VS Code settings (Run the **Preferences: Open User Settings (JSON)** command from the command palette (`F1`)) and add the following to your settings:
+
+```js
+"dart.customDevTools": {
+	"script": "/path/to/devtools/tool/build_e2e.dart",
+	"cwd": "/path/to/devtools",
+	"env": {
+		"LOCAL_DART_SDK": "/path/to/dart-sdk/sdk"
+	}
+},
+```
+
+This instructs VS Code to run the `tool/build_e2e.dart` script instead of running `dart devtools`. You must set the `cwd` and `LOCAL_DART_SDK` env variable correctly for the script to work.
+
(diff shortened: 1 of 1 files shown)


kenzieschmoll (commented):
When I did this, my VS Code seems to be stuck on this:
Screen Shot 2022-08-08 at 2 37 59 PM

DanTup:
@kenzieschmoll how long did you wait? It could take a little while for the build to complete (however long `build_e2e.dart` usually takes on your machine).

You can enabling logging by adding this to your _Workspace_ Settings (`F1` -> `Preferences: Open Workspace Settings (JSON)`) and reload (`F1` -> `Developer: Reload Window`) and then open the log file, which should show the output from the script as it runs.

```js
"dart.devToolsLogFile": "logs/devtools.txt"
```

(I presume you're on a recent DevTools and SDK checkout that include my changes from last week?)

DanTup:
I just noticed there's an existing "Developing with VS Code -> DevTools Server" section in this file, although I think it might be out of date (it says to run `packages/devtools/bin/devtools`). Should that section just be replaced by this new text? (I don't think there's a way to run with the server now without having the SDK?)

kenzieschmoll:
Yes anything that references `bin/devtools.dart` is obsolete. It could be work checking if the the `--debug` flag can still be passed to the server, though I personally haven't used that workflow in quite a while (year(s)?).

kenzieschmoll:
I upgraded to the latest Dart SDK with your changes and retried - it works now 👍 LGTM!

kenzieschmoll (approved):

DanTup:
> Yes anything that references `bin/devtools.dart` is obsolete.

Ok, shall I remove that section entirely, or there is still some value in having instructions for running the server manually without going through the VS Code integration?

> It could be work checking if the the `--debug` flag can still be passed to the server, though I personally haven't used that workflow in quite a while (year(s)?).

That flag still exists in the DDS server script, but as far as I can see the _only_ thing it does is wraps `bold()` around a message:

```dart
if (!machineMode && debugMode) {
  // Add bold to help find the correct url to open.
  message = ConsoleUtils.bold('$message\n');
}
```

I can't see any other use of it.

kenzieschmoll:
> Ok, shall I remove that section entirely, or there is still some value in having instructions for running the server manually without going through the VS Code integration?

Yeah, there is value in this. The instruction there should be simple - just running the e2e script locally.

DanTup:
It looks like there are some more things that need updating in that area.. the launch config (in `.vscode/launch.json`) is also trying to use old scripts that no longer exist.

If you're happy for me to land this change as-is, I'll go through that stuff and update it all for another PR.

kenzieschmoll (approved):
Good to land and the other stuff can be cleaned up in a follow up

PR #4359, opened 2022-08-09
`fileNameFromUri` utility method

migueltylee:
Adds the `fileNameFromUri` utility method to utils.dart, which returns the name of a file from an input URI or path string, by splitting it at the directory separators '/', and returning the last element.

**Related Issues:**
Use this method to fix issue #4360

### DIFF
--- a/packages/devtools_app/lib/src/primitives/utils.dart
+++ b/packages/devtools_app/lib/src/primitives/utils.dart
@@ -1460,6 +1460,7 @@ bool isPrimativeInstanceKind(String? kind) {
 
 // TODO(mtaylee): Prefer to use this helper method whenever a call to
 // .split('/').last is made on a String (usually on URIs).
+// See https://github.com/flutter/devtools/issues/4360.
 /// Returns the file name from a URI or path string, by splitting the [uri] at
 /// the directory separators '/', and returning the last element.
 String? fileNameFromUri(String? uri) => uri?.split('/').last;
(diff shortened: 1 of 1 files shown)


bkonyi (approved):
LGTM, but can you add a link to the issue in your TODO, just in case you don't get around to fixing it yourself...? :-)

packages/devtools_app/lib/src/primitives/utils.dart
```diff
@@ -1457,3 +1457,9 @@ bool isPrimativeInstanceKind(String? kind) {
       kind == InstanceKind.kNull ||
       kind == InstanceKind.kString;
 }
+
+// TODO(mtaylee): Prefer to use this helper method whenever a call to
+// .split('/').last is made on a String (usually on URIs).
+/// Returns the file name from a URI or path string, by splitting the [uri] at
+/// the directory separators '/', and returning the last element.
+String? fileNameFromUri(String? uri) => uri?.split('/').last;
```
kenzieschmoll, line 1465: @bkonyi will this work for a windows uri? Or does the VM send package uris always with `/` and never with `\` (event for a windows desktop app).

bkonyi: I'm 99% sure it will. I don't think the VM itself knows how to handle `\`'s internally, only in `dart:io`, so URIs from the service should be using forward slashes.

kenzieschmoll (approved):
lgtm, and +1 to Ben's request

PR #4303, opened 2022-07-27
Update to latest textmate grammar + support patterns in captures in parser

DanTup:
@bkonyi This updates to the [latest grammar](https://github.com/dart-lang/dart-syntax-highlight/blob/master/grammars/dart.json), which includes a few fixes and better highlighting for interpolated strings.

Before:

Screenshot 2022-07-27 at 16 56 09

After:

Screenshot 2022-07-27 at 16 47 27

However, the parser did not support having patterns nested in captures so I also had to make some changes there. I extracted a `PatternMatcher` that matches anything that has a set of `patterns` (and added this to `_Matcher.parse()`). and now specifically look for nested patterns in captures.

With these changes, I get the same output when running some tests (see https://github.com/dart-lang/dart-syntax-highlight/pull/28) as a NodeJS tool I was using to do the same. I have found some issues with the grammar (such has handling of nested type args), but I'd like to get and get some tests for the grammar (via https://github.com/dart-lang/dart-syntax-highlight/pull/28) before I start trying to fix that, to ensure I don't regress anything while trying to fix (since I don't expect the fix to be trivial).

Please review carefully, I'm a parser noob :-)

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #4363, opened 2022-08-09
Handle nested scopes better when parsing code for syntax highlighting

DanTup:
@bkonyi ~~I'm not ready to land this yet (I want to do some more manual testing, and compare the results to an npm tool that produces the same), but it ended up much larger than I'd planned so I thought it made sense to raise open for some feedback.~~

The goal here was to resolve some issues with overlapping/nested tokens. Previously we would produce nested tokens like this:

```
>  print('the value of \$i is $i');
#  ^^^^^ entity.name.function.dart
#        ^^^^^^^^^^^^^^^^^^^^^^^^ string.interpolated.single.dart
#                      ^^ string.interpolated.single.dart constant.character.escape.dart
#                              ^ string.interpolated.single.dart variable.parameter.dart
#                                 ^ punctuation.terminator.dart
```

The goal is to produce a simple flat list (which is consistent with another npm tool I use, allowing us to compare the output as an extra test). The new output for the above is:

```
>  print('the value of \$i is $i');
#  ^^^^^ entity.name.function.dart
#        ^^^^^^^^^^^^^^ string.interpolated.single.dart
#                      ^^ string.interpolated.single.dart constant.character.escape.dart
#                        ^^^^^^ string.interpolated.single.dart
#                              ^ string.interpolated.single.dart variable.parameter.dart
#                               ^ string.interpolated.single.dart
#                                 ^ punctuation.terminator.dart
```

Previously we would use the stack only to store the string scope names which made it difficult to produce the spans "in between" when popping the nested tokens. So I change this so the stack now drives all of the produced tokens. Instead of producing tokens directly, we always push/pop from the stack on boundaries (although there's an `add()` helper that just pushes and pops for when we want to produce an exact token). This allows us to produce those "in between" tokens (because any time we push a new scope, we can produce a token for the gap between the last token and the new token).

I added a new class to make it easier to pass around a "location" (an offset position, line number, and column number) because we now need to know the line/col for the _ends_ of tokens to support the functionality above (when we pop a token, we need to produce the next token starting at the previous tokens end).

I also moved the two remaining tests to use the "golden" format for convenience (since the results of one changed slightly with this change because it had nested tokens). I kept the source for them in the test file because our bots run `dart analyze` and the code is invalid, but the goldens now live in the `goldens` folder along with the rest.

I'll aim to test this more (and compare to the npm tool) tomorrow. If you can think of any cases this might not handle correctly, let me know and I can add additional tests (first in `master` so we can get the previous results for them, and then rebase this so we can see the impact of the changes).

### DIFF
--- a/packages/devtools_app/lib/src/screens/debugger/span_parser.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/span_parser.dart
@@ -605,7 +605,11 @@ class ScopeStack {
     assert(last.scope == scope);
     assert(last.location.position <= end.position);
 
-    _produceSpan(scopes, end: end);
+    // Only produce a span if the scope we're popping is not still in the stack
+    // somewhere.
+    if (!stack.any((item) => item.scope == scope)) {
+      _produceSpan(scopes, end: end);
+    }
   }
 
   void popAll(ScopeStackLocation location) {
--- a/packages/devtools_app/lib/src/screens/debugger/span_parser.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/span_parser.dart
@@ -106,7 +106,7 @@ class ScopeSpan {
   /// The one-based column number.
   int get column => startLocation.column + 1;
 
-  final List<String?> scopes;
+  final List<String> scopes;
 
   bool contains(int token) => (start <= token) && (token < end);
 
@@ -548,8 +548,8 @@ class _IncludeMatcher extends _Matcher {
 class ScopeStack {
   ScopeStack();
 
-  final Queue<ScopeStackItem> stack = Queue();
-  final List<ScopeSpan> spans = [];
+  final stack = Queue<ScopeStackItem>();
+  final spans = <ScopeSpan>[];
 
   /// Location where the next produced span should begin.
   ScopeStackLocation _nextLocation = ScopeStackLocation.zero;
--- a/packages/devtools_app/lib/src/screens/debugger/span_parser.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/span_parser.dart
@@ -12,14 +12,12 @@ abstract class SpanParser {
   /// Takes a TextMate [Grammar] and a [String] and outputs a list of
   /// [ScopeSpan]s corresponding to the parsed input.
   static List<ScopeSpan> parse(Grammar grammar, String src) {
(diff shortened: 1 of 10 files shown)


DanTup:
I've pushed some additional minor changes, and done some more manual tested in DevTools and comparisons to the NPM tool. With a little cleanup (because the NPM tool doesn't always do some things that we now do, such as avoiding adjacent identical tokens) the results for a some large folders from the analysis server are the same for both.

We're still on a slightly older version of the grammar here though, so I'm going to work on updating to the latest one (which may require [some further changes](https://github.com/flutter/devtools/pull/4303)) in a separate PR.

packages/devtools_app/lib/src/screens/debugger/span_parser.dart
```diff
@@ -85,61 +85,30 @@ class Grammar {
...
-  final int _start;
+  /// The one-based line number.
+  int get line => startLocation.line + 1;
 
-  int _end;
+  /// The one-based column number.
+  int get column => startLocation.column + 1;
 
-  final int? line;
+  final List<String?> scopes;
```
bkonyi, line 109: Can scope names be null?

packages/devtools_app/lib/src/screens/debugger/span_parser.dart
```diff
@@ -650,3 +542,201 @@ class _IncludeMatcher extends _Matcher {
     };
   }
 }
+
+/// Tracks the current scope stack, producing [ScopeSpan]s as the contents
+/// change.
+class ScopeStack {
+  ScopeStack();
+
+  final Queue<ScopeStackItem> stack = Queue();
```
bkonyi, line 551: Nits: drop the types on the LHS and use explicit types on the RHS instead for `stack` and `spans`.

bkonyi (approved):
Generally LGTM!

packages/devtools_app/lib/src/screens/debugger/span_parser.dart
DanTup: Doesn't seem like it, fixed!

packages/devtools_app/lib/src/screens/debugger/span_parser.dart
DanTup: Done!

Issue #4335, opened 2022-08-03
Rework `Allocations` Tab in Memory Page


bkonyi:
**Feature work:**

- [x] Reimplement allocation profile (e.g., breakdown memory + # of instances by class) (https://github.com/flutter/devtools/pull/4257)
- [x] Reimplement allocation tracing (e.g., record and display stack traces for allocations of specified classes)
    - [x] https://github.com/flutter/devtools/pull/4364

**Cleanup:**
- [x] Sort table by memory consumed rather than by class name
    - [x] In the tab Profile
    - [x] In the tab Tracing
- [ ] explain in documentation what is included in inner (shallow) size of an object
- [x] https://github.com/flutter/devtools/issues/4300
- [ ] Add button 'Deselect All' for tracing
- [x] https://github.com/flutter/devtools/issues/4484
- [ ] Support button 'Clear' and 'Export' on the control pane?

cc @polina-c

polina-c:
Related issue: https://github.com/flutter/devtools/issues/4314

InMatrix:
Is there a separate bug tracking documentation updates for the new memory screen? These two docs will become out of date once this change land in stable: 
- https://docs.flutter.dev/development/tools/devtools/memory
- https://github.com/flutter/devtools/blob/master/case_study/memory_leaks/images_1_null_safe/README.md

PR #4364, opened 2022-08-10
Initial implementation of `Tracing` tab

bkonyi:
This is an initial implementation of an allocation tracing tab for the memory screen. Current functionality includes:

- Enabling / disabling tracing for individual classes
- Refreshing tracing data
- Bottom-up tree view of allocation locations for traced classes, with inclusive and exclusive counts similar to those found on CPU profiles
- Expand / Collapse all support

Functionality that still needs to be implemented includes:

- Multi-isolate support
- Top-down tree view support
- Class list search filtering
- Improved table behavior when enabling / disabling tracing for classes. In particular, inserting a duplicate element into the list of classes when enabling tracing instead of moving the existing element to the top of the list.
- Support for serialization/deserialization

Tracking issue: https://github.com/flutter/devtools/issues/4335

**Tracing not enabled:**
image

**Tracing enabled with a selected allocation profile:**
image


bkonyi:
I still need to write tests, but figured this is worth an initial review pass.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
```diff
@@ -0,0 +1,160 @@
...
+          fixedWidthPx: scaleByFontFactor(55.0),
+          alignment: ColumnAlignment.left,
+        );
+
+  final AllocationProfileTracingViewController controller;
+
+  /// The `setState` function for [AllocationTracingTable], used to trigger a
+  /// rebuild of the table when allocation tracing is enabled or disabled as a
+  /// result of a checkbox state change in this column.
+  final Function(Function()) setState;
```
bkonyi, line 35: Not sure if this is a bit smelly, but it seemed like this simplest way to force the table to rebuild.

kenzieschmoll (commented):
partial review

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/ALLOCATION_TRACING.md
```diff
@@ -0,0 +1,17 @@
...
+tracing for specific types, which records the locations of
+allocations of instances of traced types within the
+currently selected isolate.
+            
+Allocation sites of traced types can be viewed by refreshing
+the tracing profile before selecting the traced type from the
+list, displaying a condensed view of locations where objects
+were allocated.
+
+// TODO(bkonyi): add more detailed instructions with screenshots.
```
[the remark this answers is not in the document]
polina-c: Would it be better to link the document from a less detailed page on the Flutter website?

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
```diff
@@ -0,0 +1,138 @@
...
+    TracedClass item, {
+    bool isRowSelected = false,
+    VoidCallback? onPressed,
+  }) {
+    final controller =
+        Provider.of<MemoryController>(context).allocationTracingController;
+    return Checkbox(
+      value: controller.isAllocationTracingEnabledForClass(item.cls),
+      onChanged: (value) async {
+        await controller.setAllocationTracingForClass(item.cls, value!);
```
polina-c, line 46: When I check the checkbox, the item disappears, that's confusing. Can we keep the item both on top and in the list with the selected sorting?

polina-c, line 46: After checking 'Text', and then finding it on top and selecting the row (for hello world flutter application on mac), got red screen with error: 

```
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
The following _CastError was thrown building ValueListenableBuilder<TracedClass?>(dirty, state:
_ValueListenableBuilderState<TracedClass?>#07e51):
Null check operator used on a null value

The relevant error-causing widget was:
  ValueListenableBuilder<TracedClass?>
  ValueListenableBuilder:file:///Users/polinach/_/devtools/packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocat
  ion_profile_tracing_tree.dart:29:12

#0      AllocationTracingTree.build.<anonymous closure>
(package:devtools_app/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart:46:57)
#1      _ValueListenableBuilderState.build (package:flutter/src/widgets/value_listenable_builder.dart:186:26)
#2      StatefulElement.build (package:flutter/src/widgets/framework.dart:4981:27)
#3      ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:4868:15)
#4      StatefulElement.performRebuild (package:flutter/src/widgets/framework.dart:5034:11)
#5      Element.rebuild (package:flutter/src/widgets/framework.dart:4595:5)
#6      BuildOwner.buildScope (package:flutter/src/widgets/framework.dart:2672:19)
#7      WidgetsBinding.drawFrame (package:flutter/src/widgets/binding.dart:882:21)
#8      RendererBinding._handlePersistentFrameCallback (package:flutter/src/rendering/binding.dart:374:5)
#9      SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1175:15)
#10     SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1104:9)
#11     SchedulerBinding._handleDrawFrame (package:flutter/src/scheduler/binding.dart:1015:5)
#15     _invoke (dart:ui/hooks.dart:148:10)
#16     PlatformDispatcher._drawFrame (dart:ui/platform_dispatcher.dart:316:5)
#17     _drawFrame (dart:ui/hooks.dart:113:31)
(elided 3 frames from dart:async)
```

bkonyi: This is something included in the list of functionality that still needs to be done.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
```diff
@@ -0,0 +1,138 @@
...
+
+  @override
+  Widget build(
+    BuildContext context,
+    TracedClass item, {
+    bool isRowSelected = false,
+    VoidCallback? onPressed,
+  }) {
+    final controller =
+        Provider.of<MemoryController>(context).allocationTracingController;
```
polina-c, line 42: Can you the class take the controller in constructor, like AllocationProfileTracingView does?
So that the class does not use MemoryController at all as it is too powerful and is not actually needed her.

I am working towards making all panes having their own controllers and not concerning about memory controller.

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -788,6 +795,13 @@ class TreeTableState<T extends TreeNode<T>> extends State<TreeTable<T>>
   }) {
     sortData(column, direction, secondarySortColumn: secondarySortColumn);
     _updateItems();
+    if (widget.onSortChanged != null) {
+      widget.onSortChanged!(
```
polina-c, line 799: You can do `widget.onSortChanged?.call(...)` instead.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
polina-c, line 46: When I click checkbox, nothing happens, but when I click class name, the checkbox get checked/unchecked.

polina-c, line 46: I selected `Text`, clicked button in application, and then clicked 'Refresh' on top. The panel on right did not refresh. However, when I unchecked/checked `Text`, the panel updated.

packages/devtools_app/lib/src/primitives/utils.dart
```diff
@@ -1217,6 +1217,19 @@ class ListValueNotifier<T> extends ChangeNotifier
     _listChanged();
   }
 
+  /// Replaces the first occurrence of [value] in this list.
+  ///
+  /// Runtime is O(n).
+  bool replace(T existing, T replacement) {
```
kenzieschmoll, line 1223: add a test for this by the other ListValueNotifier tests

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
```diff
@@ -0,0 +1,138 @@
...
+  _AllocationTracingTableClassName() : super.wide('Class');
+
+  @override
+  String? getValue(TracedClass stats) => stats.cls.name;
+
+  @override
+  bool get supportsSorting => true;
+}
+
+class _AllocationTracingTableInstances extends ColumnData<TracedClass> {
```
kenzieschmoll, line 72: nit on these column data class names. Since they are private, these column classes will only ever be used in this library, so can we get rid of the "_AllocationTracingTable" prefix in the name? e.g `_AllocationTracingTableClassName` --> `_ClassNameColumn`, `_AllocationTracingTableInstances` --> `_InstancesColumn`

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
```diff
@@ -0,0 +1,138 @@
...
+
+  final AllocationProfileTracingViewController controller;
+
+  @override
+  State<AllocationTracingTable> createState() => _AllocationTracingTableState();
+}
+
+class _AllocationTracingTableState extends State<AllocationTracingTable> {
+  SortDirection sortDirection = SortDirection.ascending;
+  ColumnData<TracedClass> secondarySortColumn = _classNameColumn;
```
kenzieschmoll, line 99: nit: sometimes it is helpful to have an `initState` method do the initialization so that it is clear what is stateful about this widget. At first glance, I looked at this and thought, "we aren't using any lifecycle methods besides `build`, so why does this need to be stateful"? And then looking closer I saw the call to `setState`. If these were late and then initialized in `initState` that would be more immediately clear.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart
```diff
@@ -0,0 +1,274 @@
...
+  const AllocationTracingTree({required this.controller});
+
+  final AllocationProfileTracingViewController controller;
+
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder<TracedClass?>(
+      valueListenable: controller.selectedTracedClass,
+      builder: (context, selection, _) {
+        Widget? errorColumn;
```
kenzieschmoll, line 32: instead of storing `errorColumn` in a var, just return from each if / else if statement since you are returning `errorColumn` anyway if non-null

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart
```diff
@@ -0,0 +1,274 @@
...
+          );
+        }
+        if (errorColumn != null) {
+          return errorColumn;
+        }
+        late Function(Function()) updateTreeStateCallback;
+        final theme = Theme.of(context);
+        return Column(
+          children: [
+            AreaPaneHeader(
```
kenzieschmoll, line 63: this build method is getting too large. Pull this AreaPaneHeader into a private widget `_AllocationTreeHeader` or similar.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart
```diff
@@ -0,0 +1,274 @@
...
+              tall: true,
+              needsTopBorder: false,
+              actions: [
+                ExpandAllButton(
+                  onPressed: () => updateTreeStateCallback(
+                    () {
+                      final data = controller.selectedTracedClassAllocationData;
+                      if (data == null) {
+                        return;
+                      }
```
kenzieschmoll, line 86: if `data == null` we should disable these buttons instead of having a pressed function that does nothing. Setting the `onPressed` param to null should disable the button

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart
```diff
@@ -0,0 +1,274 @@
...
+                      for (final root in data.bottomUpRoots) {
+                        root.expandCascading();
+                      }
+                    },
+                  ),
+                ),
+                const SizedBox(width: denseSpacing),
+                CollapseAllButton(
+                  onPressed: () => updateTreeStateCallback(
```
kenzieschmoll, line 95: same comment here

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart
```diff
@@ -0,0 +1,274 @@
...
+                    },
+                  ),
+                ),
+              ],
+            ),
+            Expanded(
+              child: StatefulBuilder(
+                builder: (context, setState) {
+                  updateTreeStateCallback = setState;
+                  return _AllocationProfileTracingCallTreeTable(
```
kenzieschmoll, line 113: should this class name have "BottomUp" in it instead of "CallTree"?

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart
```diff
@@ -0,0 +1,274 @@
...
+                },
+              ),
+            ),
+          ],
+        );
+      },
+    );
+  }
+
+  Widget _allocationTracingInstructions(BuildContext context) {
```
kenzieschmoll, line 128: make this a stateless widget class instead of a private method that returns a widget

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart
```diff
@@ -0,0 +1,274 @@
...
+  final ClassRef cls;
+  final List<CpuStackFrame> dataRoots;
+
+  @override
+  State<_AllocationProfileTracingCallTreeTable> createState() {
+    return _AllocationProfileTracingCallTreeTableState();
+  }
+}
+
+class _AllocationProfileTracingCallTreeTableState
```
kenzieschmoll, line 241: setState is never called in this State class - can it be stateless then? or are we missing a call to setState?

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_view.dart
```diff
@@ -0,0 +1,113 @@
...
+            'The allocation tracing tab allows for toggling allocation\n'
+            'tracing for specific types, which records the locations of\n'
+            'allocations of instances of traced types within the\n'
+            'currently selected isolate.\n'
+            '\n'
+            'Allocation sites of traced types can be viewed by refreshing\n'
+            'the tracing profile before selecting the traced type from the\n'
+            'list, displaying a condensed view of locations where objects\n'
+            'were allocated.',
+          ),
```
kenzieschmoll, line 100: CC @sfshaza2 for help with word smithing. Thanks!

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
bkonyi: I can't reproduce this.

bkonyi: Looks like there were a few state issues after I addressed the first set of comments. They should be fixed now.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart
bkonyi: Actually, this case should never happen since we won't be displaying this table if `selectedTracedClass == null`, so we should just null assert on `data`.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart
bkonyi: Missing a call to setState. Must have dropped it during some refactoring.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/ALLOCATION_TRACING.md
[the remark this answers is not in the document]
polina-c: @sfshaza2 , what do you think is better: to have everything on Flutter website or to have detailed docs here and link them from higher level pages on the website?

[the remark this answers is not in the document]
sfshaza2: I think it's better to have the docs on the Flutter website. It can potentially get a lot more visibility there.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_view.dart
sfshaza2: I'd be happy to wordsmith (it's quite a mouthful!!), but I'd love to first see a quick demo to better understand this. My calendar has space tomorrow!

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
```diff
@@ -0,0 +1,147 @@
...
+  const AllocationTracingTable({required this.controller});
+
+  final AllocationProfileTracingViewController controller;
+
+  @override
+  State<AllocationTracingTable> createState() => _AllocationTracingTableState();
+}
+
+class _AllocationTracingTableState extends State<AllocationTracingTable> {
+  late SortDirection sortDirection;
```
polina-c, line 97: I traced 'Text', collected 7 instances, then I unchecked it. 
I see number 7 in 'Instances' and I expect that the Text is not tracked any more, but the collected information preserved. 
But, when I select this row, I do not see the collected information and, when I order by number of instances, Text does not get to the top or bottom:
Screen Shot 2022-08-12 at 9 42 18 AM

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_view.dart
polina-c: @InMatrix  and Mindy could be also interested in demo.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
bkonyi: This should be fixed now.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
```diff
@@ -39,9 +38,9 @@ class _TraceCheckBoxColumn extends ColumnData<TracedClass>
     VoidCallback? onPressed,
   }) {
     final controller =
-        Provider.of<MemoryController>(context).allocationTracingController;
+        Provider.of<AllocationProfileTracingViewController>(context);
```
kenzieschmoll, line 41: There may be a performance impact for pulling this from the tree on each build. We may want to consider passing the controller in as a parameter to _TraceCheckBoxColumn for this reason.

packages/devtools_app/test/memory/allocation_tracing/allocation_profile_tracing_view_test.dart
```diff
@@ -0,0 +1,221 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:devtools_app/devtools_app.dart';
```
kenzieschmoll, line 8: import libraries directly

packages/devtools_app/test/memory/allocation_tracing/allocation_profile_tracing_view_test.dart
```diff
@@ -0,0 +1,221 @@
...
+    );
+
+    // Delay to ensure the memory profiler has collected data.
+    await tester.pumpAndSettle(const Duration(seconds: 1));
+    expect(find.byType(MemoryBody), findsOneWidget);
+  }
+
+  // Set a wide enough screen width that we do not run into overflow.
+  const windowSize = Size(2225.0, 1000.0);
+  setGlobal(NotificationService, NotificationService());
```
kenzieschmoll, line 69: is there a reason this is called outside of `setUp`?

packages/devtools_test/lib/src/mocks/fake_vm_service_wrapper.dart
```diff
@@ -22,30 +22,35 @@ class FakeVmServiceWrapper extends Fake implements VmServiceWrapper {
...
+        cpuSamples = cpuSamples ?? _defaultProfile,
+        allocationSamples = allocationSamples ?? _defaultProfile {
     _reverseResolvedUriMap = <String, String>{};
     if (_resolvedUriMap != null) {
       for (var e in _resolvedUriMap!.entries) {
         _reverseResolvedUriMap![e.value] = e.key;
       }
     }
   }
+  static final _defaultProfile = CpuSamples.parse({
```
kenzieschmoll, line 39: nit: new line above

kenzieschmoll (commented):
can you move `packages/devtools_app/test/test_data/allocation_trace.json` to `test_data/memory/allocation_tracing/`?

packages/devtools_app/test/memory/allocation_tracing/allocation_profile_tracing_view_test.dart
bkonyi: Because it was in `memory_screen_test.dart` and I copy-pasted the original test... :-)

packages/devtools_app/test/memory/allocation_tracing/allocation_profile_tracing_view_test.dart
bkonyi: I wish there was a way to prevent VSCode from auto-importing the root project library... is there any reason we export everything from `package:devtools_app/devtools_app.dart` anyway?

bkonyi:
> can you move `packages/devtools_app/test/test_data/allocation_trace.json` to `test_data/memory/allocation_tracing/`?

Done.

packages/devtools_app/test/memory/allocation_tracing/allocation_profile_tracing_view_test.dart
kenzieschmoll: I agree - is there a way to prohibit a specific import pattern in your IDE? @DanTup 

package:devtools_test imports this file. We also have g3 custom screens that import devtools_app.dart to use shared functionality.

kenzieschmoll (approved):

packages/devtools_app/test/memory/allocation_tracing/allocation_profile_tracing_view_test.dart
DanTup: > I agree - is there a way to prohibit a specific import pattern in your IDE? @DanTup

I'm not sure I understand the request here. Is the issue that the names could be imported from multiple files (eg. because of `exports`)? If so, the code completion should show each class name for each file that could be imported, and you can select which one you want:

Screenshot 2022-08-17 at 17 31 09
Screenshot 2022-08-17 at 17 31 14

I don't know of any way to set preferences around these though. Perhaps if the rules could be captured in a lint, then it could also be used to exclude some patterns from the completion/quick fixes?

PR #4370, opened 2022-08-12
Manage Android chart via settings and other improvements.

polina-c:
Remove settings for displaying units (as we are not using this setting in new Profile and plan to retire allocations).
Remove advanced memory settings as they affect only GC button and we agreed to show it to everybody. 
Remove android button from chart control pane as the chart can be turned on/off in settings.

Test coverage will be done after initial code review.


packages/devtools_app/lib/src/screens/memory/memory_controller.dart
```diff
@@ -887,6 +849,14 @@ class MemoryController extends DisposableController
         _memoryTracker = null;
       },
     );
+
+    bool _isAndroidChartVisible() =>
+        preferences.memory.androidCollectionEnabled.value &&
+        isConnectedDeviceAndroid;
+    isAndroidChartVisibleNotifier.value = _isAndroidChartVisible();
+    preferences.memory.androidCollectionEnabled.addListener(
```
CoderDake, line 857: we may want to use `autoDisposeStreamSubscription` so this automatically gets disposed when the memory controller is disposed

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -268,3 +279,57 @@ class InspectorPreferencesController extends DisposableController
...
+
+  ValueListenable<bool> get autoSnapshotEnabled => _autoSnapshotEnabled;
+  final _autoSnapshotEnabled = ValueNotifier<bool>(false);
+  static const _autoSnapshotEnabledStorageId = 'memory.autoSnapshotEnabled';
+
+  Future<void> init() async {
+    final androidCollectionEnabledValue =
+        await storage.getValue(_androidCollectionEnabledStorageId);
+    setAndroidCollectionEnabled(androidCollectionEnabledValue == 'true');
+    _androidCollectionEnabled.addListener(() {
```
CoderDake, line 298: we should probably auto dispose the listeners here too

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -48,13 +57,15 @@ class PreferencesController extends DisposableController
...
     await inspector.init();
+    await memory.init();
 
     setGlobal(PreferencesController, this);
   }
 
   @override
   void dispose() {
     inspector.dispose();
+    // memory.dispose();
```
CoderDake, line 68:
```suggestion
    memory.dispose();
```

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -268,3 +279,57 @@ class InspectorPreferencesController extends DisposableController
     _hoverEvalMode.value = enableHoverEvalMode;
   }
 }
+
+class MemoryPreferencesController {
```
CoderDake, line 283:
```suggestion
class MemoryPreferencesController extends DisposableController
    with AutoDisposeControllerMixin{
```

CoderDake (commented):
Looks pretty good so far. I think just that TODO for the tests needs to be addressed :)

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
```diff
@@ -567,20 +541,8 @@ class MemoryController extends DisposableController
...
-  bool get isAndroidChartVisible => _androidChartVisibleNotifier.value;
-
-  bool toggleAndroidChartVisibility() =>
-      _androidChartVisibleNotifier.value = !_androidChartVisibleNotifier.value;
-
-  bool get isAdvancedSettingsVisible => _advancedSettingsEnabled.value;
-
-  bool toggleAdvancedSettingsVisibility() =>
-      _advancedSettingsEnabled.value = !_advancedSettingsEnabled.value;
+  final ValueNotifier<bool> isAndroidChartVisibleNotifier =
```
kenzieschmoll, line 544: nit: remove type on LHS

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
kenzieschmoll, line 857: use addAutoDisposeListener for this

packages/devtools_app/lib/src/screens/memory/panes/chart/memory_android_chart.dart
```diff
@@ -160,23 +160,19 @@ class MemoryAndroidChartState extends State<MemoryAndroidChart>
...
   Widget build(BuildContext context) {
-    if (_chartController.timestamps.isNotEmpty) {
-      return Container(
-        height: defaultChartHeight,
-        child: Chart(_chartController),
-      );
-    }
-
-    return const SizedBox(width: denseSpacing);
+    return Container(
```
kenzieschmoll, line 172: use SizedBox instead of Container if we are just using the `height` property

packages/devtools_app/lib/src/screens/memory/panes/control/control_pane.dart
```diff
@@ -27,9 +27,6 @@ class _MemoryControlPaneState extends State<MemoryControlPane>
     with
         AutoDisposeMixin,
         ProvidedControllerMixin<MemoryController, MemoryControlPane> {
-  /// Updated when the MemoryController's _androidCollectionEnabled ValueNotifier changes.
-  bool _isAndroidCollection = MemoryController.androidADBDefault;
-
   @override
   Widget build(BuildContext context) {
```
kenzieschmoll, line 31: looks like this widget can now be stateless

packages/devtools_app/lib/src/screens/memory/panes/control/secondary_controls.dart
```diff
@@ -198,30 +195,18 @@ class _SecondaryControlsState extends State<SecondaryControls>
       children: [
         const MemorySourceDropdown(),
         const SizedBox(width: defaultSpacing),
-        if (controller.isConnectedDeviceAndroid ||
-            controller.isOfflineAndAndroidData)
-          ToggleAdbMemoryButton(
-            isAndroidCollection: widget.isAndroidCollection,
-          ),
         const SizedBox(width: denseSpacing),
```
kenzieschmoll, line 198: delete this SizedBox now since we have two in a row

packages/devtools_app/lib/src/screens/memory/panes/control/secondary_controls.dart
```diff
@@ -198,30 +195,18 @@ class _SecondaryControlsState extends State<SecondaryControls>
...
+        Row(
+          children: [
+            IconLabelButton(
+              onPressed: controller.isGcing ? null : _gc,
+              icon: Icons.delete,
+              label: 'GC',
+              minScreenWidthForTextBeforeScaling:
+                  primaryControlsMinVerboseWidth,
+            ),
+            const SizedBox(width: denseSpacing),
```
kenzieschmoll, line 208: pull these two children out into the parent row and delete this Row widget. Unnecessary nesting.

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -25,6 +28,12 @@ class PreferencesController extends DisposableController
 
   InspectorPreferencesController get inspector => _inspector;
 
+  MemoryPreferencesController get memory => _memory;
+  final _memory = MemoryPreferencesController();
+
+  MemoryPreferencesController get memoryPreferences => _memoryPreferences;
+  final _memoryPreferences = MemoryPreferencesController();
+
```
kenzieschmoll, line 36: delete this

packages/devtools_app/lib/src/shared/preferences.dart
kenzieschmoll: +1 use auto dispose mixin functionality

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -268,3 +279,57 @@ class InspectorPreferencesController extends DisposableController
...
+    _androidCollectionEnabled.value = enable;
+    if (enable) {
+      ga.select(
+        analytics_constants.memory,
+        analytics_constants.androidChart,
+      );
+    }
+  }
+
+  void setAutoSnapshotEnabled(bool enable) {
```
kenzieschmoll, line 326: should these methods be private? they don't appear to be used anywhere else

packages/devtools_app/lib/src/screens/memory/panes/control/settings_dialog.dart
```diff
@@ -10,6 +10,12 @@ import '../../../../shared/globals.dart';
 import '../../../../shared/theme.dart';
 import '../../memory_controller.dart';
 
+/// The dialog keys for testing purposes.
+class MemorySettingDialogKeys {
+  static const Key showAndroidChartCheckBox = ValueKey('showAndroidChart');
+  static const Key autoSnapshotCheckbox = ValueKey('autoSnapshotCheckbox');
```
kenzieschmoll, line 16: mark these as visibleForTesting

packages/devtools_app/lib/src/screens/memory/panes/control/settings_dialog.dart
```diff
@@ -10,6 +10,12 @@ import '../../../../shared/globals.dart';
 import '../../../../shared/theme.dart';
 import '../../memory_controller.dart';
 
+/// The dialog keys for testing purposes.
+class MemorySettingDialogKeys {
```
kenzieschmoll, line 14: these don't need to be in a class. There should be a lint triggering for this to avoid classes with only static members

packages/devtools_app/test/memory/control/settings_dialog_test.dart
```diff
@@ -0,0 +1,138 @@
...
+        memoryData: memoryJson,
+        allocationData: allocationJson,
+      ),
+    );
+    when(fakeServiceManager.connectedApp!.isDartWebAppNow).thenReturn(false);
+    when(fakeServiceManager.connectedApp!.isFlutterAppNow).thenReturn(true);
+    when(fakeServiceManager.connectedApp!.isDartCliAppNow).thenReturn(false);
+    when(fakeServiceManager.connectedApp!.isDebugFlutterAppNow)
+        .thenReturn(false);
+    when(fakeServiceManager.connectedApp!.isDartWebApp)
```
kenzieschmoll, line 49: use the mockConnectedApp helper here

packages/devtools_app/test/memory/control/settings_dialog_test.dart
```diff
@@ -0,0 +1,138 @@
...
+        .thenAnswer((_) => Future.value(false));
+    setGlobal(ServiceConnectionManager, fakeServiceManager);
+
+    controller.offline.value = true;
+    controller.memoryTimeline.offlineData.clear();
+    controller.memoryTimeline.offlineData.addAll(memoryJson.data);
```
kenzieschmoll, line 55: use .. notation
```
controller
  ..offline.value = true
  ..memoryTimeline.offlineData.clear()
  ...
```

packages/devtools_app/test/memory/control/settings_dialog_test.dart
```diff
@@ -0,0 +1,138 @@
...
+      setGlobal(
+        PreferencesController,
+        preferencesController = PreferencesController(),
+      );
+    });
+
+    testWidgetsWithWindowSize('settings update preferences', windowSize,
+        (WidgetTester tester) async {
+      await pumpMemoryScreen(tester);
+      _setUpServiceManagerForMemory();
```
kenzieschmoll, line 102: this should be done in `setUp`, and it appears that there is a bit of overlap between `_setUpServiceManagerForMemory` and the `setUp` method above. Please get rid of the overlap. Better yet, there doesn't appear to be a reason for having two methods at all, so doing all the set up tasks directly in the `setUp` test lifecycle method is best.

packages/devtools_app/test/memory/control/settings_dialog_test.dart
```diff
@@ -0,0 +1,138 @@
...
+      await tester.tap(find.byType(SettingsOutlinedButton));
+      await tester.pumpAndSettle();
+      await expectLater(
+        find.byType(MemorySettingsDialog),
+        matchesDevToolsGolden('../../goldens/settings_dialog_default.png'),
+      );
+
+      // Modify settings and check the changes are reflected in the controller.
+      expect(preferencesController.memory.autoSnapshotEnabled.value, isFalse);
+      expect(preferencesController.memory.autoSnapshotEnabled.value, isFalse);
```
kenzieschmoll, line 114: duplicated. Did you mean to check two different settings?

packages/devtools_app/test/memory/control/settings_dialog_test.dart
```diff
@@ -0,0 +1,138 @@
...
+      await tester
+          .tap(find.byKey(MemorySettingDialogKeys.showAndroidChartCheckBox));
+      await tester
+          .tap(find.byKey(MemorySettingDialogKeys.autoSnapshotCheckbox));
+      expect(preferencesController.memory.autoSnapshotEnabled.value, isTrue);
+      expect(preferencesController.memory.autoSnapshotEnabled.value, isTrue);
```
kenzieschmoll, line 120: same here

packages/devtools_app/test/memory/control/settings_dialog_test.dart
```diff
@@ -0,0 +1,138 @@
...
+      await tester.pumpAndSettle();
```
kenzieschmoll, line 121: call pumpAndSettle directly after the tap calls instead of after the expect statements

packages/devtools_app/test/shared/preferences_controller_test.dart
```diff
@@ -15,13 +15,18 @@ import '../test_infra/flutter_test_storage.dart';
 void main() {
   setGlobal(ServiceConnectionManager, FakeServiceManager());
 
-  group('PreferencesController', () {
+  group('$PreferencesController', () {
```
kenzieschmoll, line 18: why are we adding the `$` char in group names? This isn't done anywhere else in our test code, so let's remove to be consistent

packages/devtools_app/test/shared/preferences_controller_test.dart
```diff
@@ -119,4 +124,67 @@ void main() {
       // preferences testable, then test it
     });
   });
+
+  group('$MemoryPreferencesController', () {
+    late MemoryPreferencesController controller;
+    late FlutterTestStorage storage;
```
kenzieschmoll, line 130: we shouldn't need to store this var locally since we are setting a global. We should just be able to access it via the global var `storage`

packages/devtools_app/test/test_infra/flutter_test_storage.dart
```diff
@@ -8,15 +8,15 @@ import 'package:devtools_app/src/primitives/storage.dart';
 ///
 /// This ephemeral implementation is meant to help keep unit tests segregated
 class FlutterTestStorage implements Storage {
-  late final Map<String, dynamic> _values = {};
+  late final Map<String, dynamic> values = {};
```
kenzieschmoll, line 11: this doesn't need to be public since it can be accessed through getValue and setValue

polina-c: It is test storage, not prod storage and I want to be able to check number of values.

packages/devtools_app/test/shared/preferences_controller_test.dart
polina-c: With $, if the class is renamed, the group name will catch up the update. Yes, as any innovation, it is inconsistent with previous pattern. 
You think the benefit does not worth the break of consistency?

packages/devtools_app/lib/src/screens/memory/panes/control/settings_dialog.dart
polina-c: it is ok for enum-like classes: 
```
However, this isn’t a hard rule. With constants and enum-like types,
it may be natural to group them in a class:

class Color {
  static const red = '#f00';
  static const green = '#0f0';
  static const blue = '#00f';
  static const black = '#000';
  static const white = '#fff';
}
```

https://dart.dev/guides/language/effective-dart/design#avoid-defining-a-class-that-contains-only-static-members

and there is no lint

kenzieschmoll: The lint is here: https://dart-lang.github.io/linter/lints/avoid_classes_with_only_static_members.html This used to be enabled for our repo - I'm surprised this isn't triggering analysis warning. 

But okay to leave. In general, we do not want to get in the habit of adding keys just for testing, because keys can have a performance impact and other side effects - they do more than just give the widget an "id", so it's good to take caution. In some cases, it's okay to add them for testing if using finders to find a widget in tests is overly complicated without a key, but in general it's good to avoid.

packages/devtools_app/test/memory/control/settings_dialog_test.dart
```diff
@@ -17,52 +17,19 @@ import 'package:devtools_shared/devtools_shared.dart';
...
-  Future<void> pumpMemoryScreen(
-    WidgetTester tester, {
-    MemoryController? memoryController,
-  }) async {
+  Future<void> pumpMemoryScreen(WidgetTester tester) async {
     await tester.pumpWidget(
       wrapWithControllers(
         const MemoryBody(),
-        memory: controller = memoryController ?? MemoryController(),
+        memory: controller = controller,
```
kenzieschmoll, line 32: it looks like we are just assigning `controller` to itself, so this shouldn't be necessary.

packages/devtools_app/test/shared/preferences_controller_test.dart
kenzieschmoll: Ah, nice. Today I learned that you can use dart string interpolation on Classes. Good to leave.

kenzieschmoll (approved):
couple more comments, but generally lgtm

PR #4372, opened 2022-08-15
Fix getPubRootDirectories so it works over Dart Debug Extension

CoderDake:
I was able to determine that the pub root directories were not being returned to devtools because they were being returned as a `JSArray<dynamic>`. This PR removes the type check that was overly specific, and blocked the return of those values.

I chose not to try to catch any type casting exceptions yet, as I'm unsure of a specific case where we need this yet. So for now, if we get a non-list type, then this code may fail.

### DIFF
--- a/packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
+++ b/packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
@@ -461,9 +461,11 @@ class InspectorService extends InspectorServiceBase {
       'getPubRootDirectories',
     );
 
-    return (response as List<dynamic>)
-        .map<String>((e) => e.toString())
-        .toList();
+    if (response is! List<dynamic>) {
+      return [];
+    }
+
+    return response.map<String>((e) => e.toString()).toList();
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
```diff
@@ -461,8 +461,6 @@ class InspectorService extends InspectorServiceBase {
       'getPubRootDirectories',
     );
 
-    if (response.runtimeType != List) return [];
```
jacob314, line 464: You could use
```dart
if (response is! List<dynamic>) return []
```
You should also file a bug about the difference in VM behavior and Dart2JS behavior. I think the VM behavior is right and the type Object in Dart2Js should also be List.

CoderDake: You are right, looks like that works!

kenzieschmoll (approved):

jacob314 (approved):

PR #4374, opened 2022-08-16
InspectorService may not always be present

CoderDake:
While following the docs to roll devtools to g3, i noticed a failure in the preferences code. The inspector service was being called, but it was set to null.

This PR makes the getter optional to account for the fact that there won't always be an inspector, and as such we won't want to call those operations.

NOTE: I've tested these changes in g3 to make sure they fix the error.

### DIFF
--- a/packages/devtools_app/lib/src/shared/preferences.dart
+++ b/packages/devtools_app/lib/src/shared/preferences.dart
@@ -84,7 +84,7 @@ class InspectorPreferencesController extends DisposableController
   ValueListenable<bool> get isRefreshingCustomPubRootDirectories =>
       _customPubRootDirectoriesAreBusy;
   InspectorService? get _inspectorService =>
-      serviceManager.inspectorService as InspectorService;
+      serviceManager.inspectorService as InspectorService?;
 
   final _hoverEvalMode = ValueNotifier<bool>(false);
   final _customPubRootDirectories = ListValueNotifier<String>([]);
--- a/packages/devtools_app/lib/src/shared/preferences.dart
+++ b/packages/devtools_app/lib/src/shared/preferences.dart
@@ -82,7 +82,7 @@ class InspectorPreferencesController extends DisposableController
       _customPubRootDirectories;
   ValueListenable<bool> get isRefreshingCustomPubRootDirectories =>
       _customPubRootDirectoriesAreBusy;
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -195,7 +195,8 @@ class InspectorPreferencesController extends DisposableController
   ) async {
     if (!serviceManager.hasConnection) return;
     await _customPubRootDirectoryBusyTracker(() async {
-      await inspectorService.addPubRootDirectories(pubRootDirectories);
+      if (inspectorService == null) return;
```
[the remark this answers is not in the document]
CoderDake: this is a hot tip. I've always wondered why i see local vars like that

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -82,7 +82,7 @@ class InspectorPreferencesController extends DisposableController
       _customPubRootDirectories;
   ValueListenable<bool> get isRefreshingCustomPubRootDirectories =>
       _customPubRootDirectoriesAreBusy;
-  InspectorService get inspectorService =>
+  InspectorService? get inspectorService =>
```
jacob314: This won't change anything as the `as` will still throw if it is not null.
`InspectorService? get inspectorService => serviceManager.inspectorService;`
is probably what you want.

kenzieschmoll (approved):

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -82,8 +83,8 @@ class InspectorPreferencesController extends DisposableController
       _customPubRootDirectories;
   ValueListenable<bool> get isRefreshingCustomPubRootDirectories =>
       _customPubRootDirectoriesAreBusy;
-  InspectorService get inspectorService =>
-      serviceManager.inspectorService as InspectorService;
+  InspectorService? get _inspectorService =>
+      serviceManager.inspectorService as InspectorService?;
```
kenzieschmoll, line 87: do we still need the cast?

CoderDake: I've tried without it and it doesn't work because serviceManage.inspectorService is of type `InspectorServiceBase`

PR #4377, opened 2022-08-17
Add filtering support to Allocation Tracing tab

bkonyi:
**No filter applied:**
image

**Filtered:**
image

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
+++ b/packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
@@ -8,7 +8,9 @@ import '../../../../primitives/utils.dart';
 import '../../../../shared/common_widgets.dart';
 import '../../../../shared/table.dart';
 import '../../../../shared/table_data.dart';
+import '../../../../shared/theme.dart';
 import '../../../../shared/utils.dart';
+import '../../../../ui/icons.dart';
 import 'allocation_profile_tracing_view_controller.dart';
 
 /// The default width for columns containing *mostly* numeric data (e.g.,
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
```diff
@@ -117,29 +118,85 @@ class _AllocationTracingTableState extends State<AllocationTracingTable> {
...
+          padding: const EdgeInsets.only(bottom: densePadding),
+          child: ClassFilterTextField(
+            controller: widget.controller,
+          ),
+        ),
+        DualValueListenableBuilder<bool, List<TracedClass>>(
+          firstListenable: widget.controller.refreshing,
+          secondListenable: widget.controller.classList,
+          builder: (context, _, classList, __) {
+            return Expanded(
```
kenzieschmoll, line 133: put the Expanded widget around DualValueListenableBuilder so that it is clear that the Expanded is a direct child of the Column

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
```diff
@@ -117,29 +118,85 @@ class _AllocationTracingTableState extends State<AllocationTracingTable> {
...
+          // Flutter documentation. For more details, see:
+          // https://api.flutter.dev/flutter/material/InputDecoration/prefixIconConstraints.html
+          minHeight: 32,
+          maxHeight: 32,
+        ),
+        contentPadding: const EdgeInsets.all(denseSpacing),
+        border: const OutlineInputBorder(),
+        labelText: 'Class Filter',
+        hintText: 'Filter by class name',
+        suffixIcon: IconButton(
```
kenzieschmoll, line 187: can you use `suffix` and use the `clearInputButton` helper here instead?

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
```diff
@@ -117,29 +118,85 @@ class _AllocationTracingTableState extends State<AllocationTracingTable> {
...
+
+  final AllocationProfileTracingViewController controller;
+  final _textEditingController = TextEditingController();
+
+  @override
+  Widget build(BuildContext context) {
+    return TextField(
+      controller: _textEditingController,
+      decoration: InputDecoration(
+        constraints: const BoxConstraints(
```
kenzieschmoll, line 176: if you use the clearInputButton helper below I think you can get rid of these constraints

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_view_controller.dart
```diff
@@ -68,6 +68,9 @@ class AllocationProfileTracingViewController extends DisposableController
   /// The list of classes for the currently selected isolate.
   ValueListenable<List<TracedClass>> get classList => _classList;
   final _classList = ListValueNotifier<TracedClass>([]);
```
kenzieschmoll, line 70: let's rename to _filteredClassList. This would match naming convention for filtered data in other places

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_view_controller.dart
```diff
@@ -83,6 +86,18 @@ class AllocationProfileTracingViewController extends DisposableController
   final _tracedClasses = <String, TracedClass>{};
   final _tracedClassesProfiles = <String, CpuProfileData>{};
 
+  void updateClassFilter(String value) {
+    if (value.isEmpty && _currentFilter.isEmpty) return;
+    _currentFilter = value;
+    final updatedFilter = _unfilteredClassList
```
kenzieschmoll, line 92: we could be smarter here and only look through the filtered data when the previous value `_currentFilter` is a substring of the new value `value`

packages/devtools_app/test/memory/allocation_tracing/allocation_profile_tracing_view_test.dart
```diff
@@ -226,5 +241,105 @@ void main() {
...
+      });
+
+      testWidgetsWithWindowSize('persisted tracing state', windowSize,
+          (tester) async {
+        await pumpMemoryScreen(tester);
+
+        final controller = await navigateToAllocationTracing(tester);
+
+        final checkboxes = find.byType(Checkbox);
+        expect(checkboxes, findsNWidgets(originalClassCount));
```
kenzieschmoll, line 278: no need to have originalClassCount be a local if we only use it once.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
bkonyi: To be honest, I don't think it looks quite right as it's not correctly aligned and isn't visible unless the filter text field is focused. I think we should consider switching to this style, using `suffixIcon: clearInputButton(() {...})` where `clearInputButton` returns this widget.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
bkonyi: Removing these constraints results in an ~50% increase in the text field's height, even when using `clearInputButton`.

kenzieschmoll (approved):

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
```diff
@@ -117,29 +118,85 @@ class _AllocationTracingTableState extends State<AllocationTracingTable> {
...
+          // Forces the row and clear icon to be dense. Default is 48, as per
+          // Flutter documentation. For more details, see:
+          // https://api.flutter.dev/flutter/material/InputDecoration/prefixIconConstraints.html
+          minHeight: 32,
+          maxHeight: 32,
```
kenzieschmoll, line 181: let's put this into a const somewhere then like denseTextFieldHeight

kenzieschmoll (approved):

PR #4382, opened 2022-08-17
VM Developer Tools screen IndexedStack & other changes

migueltylee:
~~This PR removes the ValueListenableBuilder that was being used in the build method of `_VMDeveloperToolsScreenState` and uses the widget's state to switch between the VmDeveloperViews.
It also adds the `AutomaticKeepAliveClientMixin<VMDeveloperToolsScreenBody>` so that the screen stays alive when switching between the main DevTools tabs.~~

**UPDATE:**
This PR includes the `ObjectInspectorViewController` in the `VmDeveloperToolsController` to retain the data (mainly the ProgramExplorerController and the object history) when switching between the main DevTools tabs. It also adds an IndexedStack to maintain the state of the VmDeveloperView's that are accessed from the VmDeveloperToolsScreen `NavigationRail`.

Other minor changes are:
- Removing test classes that were no longer being used from vm_developer_test_utils.dart.
- In vm_object_model.dart: avoid making a getObject service call if the selected node from the ProgramExplorer already contains the full Obj (not an ObjRef).

### DIFF
--- a/packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view_controller.dart
+++ b/packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view_controller.dart
@@ -16,18 +16,6 @@ import 'vm_object_model.dart';
 /// the object history and the object viewport.
 class ObjectInspectorViewController extends DisposableController
     with AutoDisposeControllerMixin {
-  ObjectInspectorViewController() {
-    addAutoDisposeListener(
-      scriptManager.sortedScripts,
-      selectAndPushMainScript,
-    );
-
-    addAutoDisposeListener(
-      objectHistory.current,
-      _onCurrentObjectChanged,
-    );
-  }
-
   final programExplorerController =
       ProgramExplorerController(showCodeNodes: true);
 
@@ -40,10 +28,30 @@ class ObjectInspectorViewController extends DisposableController
   ValueListenable<bool> get refreshing => _refreshing;
   final _refreshing = ValueNotifier<bool>(false);
 
+  ValueListenable<bool> get initialized => _initialized;
+  final _initialized = ValueNotifier<bool>(false);
+
   void init() {
-    programExplorerController
-      ..initialize()
-      ..initListeners();
+    if (!initialized.value) {
+      programExplorerController
+        ..initialize()
+        ..initListeners();
+      initListeners();
+      selectAndPushMainScript();
+      _initialized.value = true;
+    }
+  }
+
+  void initListeners() {
+    addAutoDisposeListener(
+      scriptManager.sortedScripts,
+      selectAndPushMainScript,
+    );
+
+    addAutoDisposeListener(
+      objectHistory.current,
+      _onCurrentObjectChanged,
+    );
   }
 
   Future<void> _onCurrentObjectChanged() async {
--- a/packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view_controller.dart
+++ b/packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view_controller.dart
@@ -16,6 +16,18 @@ import 'vm_object_model.dart';
 /// the object history and the object viewport.
 class ObjectInspectorViewController extends DisposableController
     with AutoDisposeControllerMixin {
+  ObjectInspectorViewController() {
+    addAutoDisposeListener(
+      scriptManager.sortedScripts,
+      selectAndPushMainScript,
+    );
+
+    addAutoDisposeListener(
+      objectHistory.current,
+      _onCurrentObjectChanged,
+    );
+  }
+
   final programExplorerController =
       ProgramExplorerController(showCodeNodes: true);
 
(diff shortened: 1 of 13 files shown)


kenzieschmoll:
> It also adds the `AutomaticKeepAliveClientMixin<VMDeveloperToolsScreenBody>` so that the screen stays alive when switching between the main DevTools tabs.

I don't think we want to do this. This has some performance cost to it. For example, if we were to keep every devtools screen alive even when we are on other tabs, there would be active listeners firing and processing data that we don't need. This also has the potential cost of us bogging down the vm service with calls that we don't need to be making.

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_tools_screen.dart
```diff
@@ -78,47 +79,67 @@ class VMDeveloperToolsScreenBody extends StatefulWidget {
...
+  late List<Widget> _views;
+
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
     initController();
   }
 
+  void onDestinationSelected(int i) {
+    setState(() => _selectedIndex = i);
```
kenzieschmoll, line 95: calling setState has the same effect as wrapping the content in a ValueListenableBuilder (it rebuilds the whole widget). So logically, this change is a no-op.

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_tools_screen.dart
```diff
@@ -78,47 +79,67 @@ class VMDeveloperToolsScreenBody extends StatefulWidget {
...
+    controller.onSelectedIndex(i);
+  }
+
+  @override
+  void initState() {
+    super.initState();
+    _views = [
```
kenzieschmoll, line 102: we don't want to do this in initState. This means that the _views will only be built once, even if a rebuild is triggered further up the tree and we really want all of `VMDeveloperToolsScreenBody` to be rebuilt.

any time you are building widgets in initState in flutter is a code smell. Generally, never build a widget outside of the build lifecycle.

migueltylee: Yeah, you're right. I was following an example I found, my bad...

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_tools_screen.dart
```diff
@@ -112,8 +112,13 @@ class _VMDeveloperToolsScreenState extends State<VMDeveloperToolsScreenBody>
                 padding: const EdgeInsets.only(
                   left: defaultSpacing,
                 ),
-                child: VMDeveloperToolsScreenBody.views[selectedIndex]
-                    .build(context),
+                child: IndexedStack(
```
kenzieschmoll, line 115: Will the IndexedStack be necessary after you make the changes we discussed offline to change where the ObjectInspectorViewController is initialized?

migueltylee: Probably not, I'll remove it if it's not necessary

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_tools_screen.dart
migueltylee: Reverted

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_tools_screen.dart
migueltylee: Keeping the IndexedStack so the full state of the VmDeveloperViews persists (including the ExpansionTile states) when switching through the `NavigationRail`

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view_controller.dart
```diff
@@ -40,10 +28,30 @@ class ObjectInspectorViewController extends DisposableController
   ValueListenable<bool> get refreshing => _refreshing;
   final _refreshing = ValueNotifier<bool>(false);
 
+  ValueListenable<bool> get initialized => _initialized;
+  final _initialized = ValueNotifier<bool>(false);
```
kenzieschmoll, line 32: we aren't listening to this anywhere, so we can just use a private bool _initialzed instead of the notifier / listenable combo

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view_controller.dart
```diff
@@ -40,10 +28,30 @@ class ObjectInspectorViewController extends DisposableController
...
   void init() {
-    programExplorerController
-      ..initialize()
-      ..initListeners();
+    if (!initialized.value) {
+      programExplorerController
+        ..initialize()
+        ..initListeners();
+      initListeners();
+      selectAndPushMainScript();
```
kenzieschmoll, line 40: why did we not have to call selectAndPushMainScript in init before this change?

migueltylee: It had not beet added and the initial screen in the object viewport said 'No object selected', I thought it would be good to have the main script instead

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view_controller.dart
migueltylee: done

kenzieschmoll (approved):
LGTM once tests are passing 👍

PR #4383, opened 2022-08-18
Add support for pinning items to `FlatTable`

bkonyi:
This change adds support for table elements to request to be pinned to the top of a `FlatTable` by implementing `PinnableListEntry` and overriding the behavior of `bool get pinToTop`. Pinned elements can either be copied to the top of the table, leaving the original in place, or simply moved to the top of the table depending on the value of the `pinBehavior` property of `FlatTable`.

Pinned items will be placed at the top of the table until at most half of the visible table consists of pinned items, after which pinned items will be displayed in a scrollable list at the top of the table widget. Pinned items will also respect changes to sorting order, while still remaining at the top of the table.

This change also updates the `AllocationTracingTable` widget to support pinning traced classes to the top of the table using `pinOriginalToTop` pinning behavior.

**Example:**
Screen Recording 2022-08-17 at 4 33 30 PM


packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -132,7 +163,7 @@ class FlatTableState<T> extends State<FlatTable<T>>
   }
 
   void _initData() {
-    data = List.from(widget.data);
+    _originalData = UnmodifiableListView(List.from(widget.data));
```
kenzieschmoll, line 166: do we still need to initialize `data` here?

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -315,6 +349,23 @@ class FlatTableState<T> extends State<FlatTable<T>>
...
+        if (pinnableEntry.pinToTop) {
+          pinnedData.add(entry);
+        }
+        if (!pinnableEntry.pinToTop ||
+            widget.pinBehavior == FlatTablePinBehavior.pinCopyToTop) {
+          dataCopy.add(entry);
+        }
+      }
+      data.clear();
+      data.addAll(dataCopy);
```
kenzieschmoll, line 367: why not just do `data = List.from(dataCopy);`

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -944,12 +1001,23 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
     return tableWidth;
   }
 
+  Widget _buildItemPinned(BuildContext context, int index) {
+    return widget.rowBuilder(
+      context,
+      _linkedHorizontalScrollControllerGroup,
+      index,
+      widget.columnWidths,
+      true,
```
kenzieschmoll, line 1010: can we make these parameters required and named for better readability?

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -944,12 +1001,23 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
     return tableWidth;
   }
 
+  Widget _buildItemPinned(BuildContext context, int index) {
```
kenzieschmoll, line 1004: instead of creating a new method, just add an optional named parameter `bool pinned = false` to `_buildItem`

packages/devtools_app/lib/src/shared/table.dart
```diff
@@ -997,6 +1065,28 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
...
+                    child: ListView.builder(
+                      controller: pinnedScrollController,
+                      itemCount: widget.pinnedData.length,
+                      itemExtent: widget.rowItemExtent,
+                      itemBuilder: _buildItemPinned,
+                    ),
+                  ),
+                ),
+                const Divider(
+                  thickness: 5,
```
kenzieschmoll, line 1086: put this value into a named const

packages/devtools_app/lib/src/shared/table_data.dart
```diff
@@ -101,6 +101,11 @@ enum ColumnAlignment {
   center,
 }
 
+abstract class PinnableListEntry {
```
kenzieschmoll, line 104: should this be a Mixin instead?

packages/devtools_app/test/shared/table_test.dart
```diff
@@ -1120,6 +1262,22 @@ class TestData extends TreeNode<TestData> {
...
 }
 
+class PinnableTestData implements PinnableListEntry {
+  PinnableTestData({
+    required this.name,
+    required this.enabled,
+  });
+
+  final String name;
+  bool enabled;
```
kenzieschmoll, line 1272: should this be final?

packages/devtools_app/lib/src/shared/table.dart
bkonyi: May as well just do `data = dataCopy`. I think this is an artifact from a different approach I was taking before.

packages/devtools_app/lib/src/shared/table_data.dart
bkonyi: Yeah, that's what I had it as originally, but was fighting with the type check in `FlatTableState`. Done.

packages/devtools_app/test/shared/table_test.dart
bkonyi: Was planning on modifying it in place before but didn't end up doing that. Done.

packages/devtools_app/lib/src/shared/table.dart
bkonyi: No, since we always re-initialize it in `sortData` anyway.

packages/devtools_app/lib/src/screens/memory/memory_heap_tree_view.dart
```diff
@@ -41,7 +41,7 @@ import 'primitives/memory_utils.dart';
 // TODO(bkonyi): enable new allocation profile table when we're ready to remove
 // the existing allocations table.
 @visibleForTesting
-bool enableNewAllocationProfileTable = false;
+bool enableNewAllocationProfileTable = true;
```
kenzieschmoll, line 44: is this ready to flip on by default?

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
```diff
@@ -118,29 +118,45 @@ class _AllocationTracingTableState extends State<AllocationTracingTable> {
 
   @override
   Widget build(BuildContext context) {
```
kenzieschmoll, line 120: this looks broken

packages/devtools_app/lib/src/screens/memory/memory_heap_tree_view.dart
bkonyi: I can't wait to get rid of this flag so I can stop forgetting to flip it back... :-)

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
bkonyi: How so? Seems to work fine on my end.

kenzieschmoll (approved):

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
kenzieschmoll: weird. maybe git diff is struggling:
Screen Shot 2022-08-19 at 12 39 45 PM

PR #4387, opened 2022-08-19
Search FileExplorer and Outline nodes

migueltylee:
Implements the `searchFileExplorer` and `breadthFirstSearchObject` methods in the ProgramExplorerController.
The `searchFileExplorer` searches and returns the script or library node in the FileExplorer which is the source location of a target object.
The `breadthFirstSearchObject` performs a breath first search on a list of roots and returns the first node whose object is the same as the target object. It is used as a helper of `searchFileExplorer` or to search an object in the ProgramExplorer `outlineNodes.value`.

These two methods are currently used in the ObjectInspectorViewController `_onCurrentObjectChanged` method to find the corresponding FileExplorer and outline nodes if the current object does not have associated `scriptRefs` or `outlineNode` values.
Consider using these methods for pushing objects that are not accessed directly through the ProgramExplorer, but instead through a linking button or widget.

### DIFF
--- a/packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
@@ -302,73 +302,69 @@ class ProgramExplorerController extends DisposableController
 
   /// Searches and returns the script or library node in the FileExplorer
   /// which is the source location of the target [object].
-  Future<VMServiceObjectNode?> searchFileExplorer(ObjRef object) async {
+  Future<VMServiceObjectNode> searchFileExplorer(ObjRef object) async {
     final service = serviceManager.service!;
     final isolateId = serviceManager.isolateManager.selectedIsolate.value!.id!;
 
-    LibraryRef? targetLib;
     ScriptRef? targetScript;
-    VMServiceObjectNode? libNode;
 
-    VMServiceObjectNode? _searchRootObjectNodes(ObjRef obj) {
-      for (final rootNode in rootObjectNodes.value) {
-        if (rootNode.object?.id == obj.id) {
-          return rootNode;
-        }
-      }
-      return null;
+    // If `object` is a library, it will always be a root node and is simple to
+    // find.
+    if (object is LibraryRef) {
+      final result = _searchRootObjectNodes(object)!;
+      await result.populateLocation();
+      return result;
     }
 
-    if (object is LibraryRef) {
-      libNode = _searchRootObjectNodes(object);
-
-      libNode ??= breadthFirstSearchObject(object, rootObjectNodes.value);
-
-      return libNode;
-    } else if (object is ClassRef || object is FieldRef || object is FuncRef) {
-      final source = (object as dynamic).location as SourceLocation?;
-      targetScript = source?.script;
+    // Otherwise, we need to find the target script to determine the library
+    // the target node is listed under.
+    if (object is ClassRef) {
+      targetScript = object.location?.script;
+    } else if (object is FieldRef) {
+      targetScript = object.location?.script;
+    } else if (object is FuncRef) {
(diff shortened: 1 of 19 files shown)


packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
```diff
@@ -296,4 +296,102 @@ class ProgramExplorerController extends DisposableController
...
+  VMServiceObjectNode? breadthFirstSearchObject(
+    ObjRef obj,
+    List<VMServiceObjectNode> roots,
+  ) {
+    VMServiceObjectNode? match;
+
+    for (final root in roots) {
+      breadthFirstTraversal<VMServiceObjectNode>(
+        root,
+        action: (node) {
```
bkonyi, line 384: We can use `returnCondition` instead of `action` here and just do `returnCondition: (node) => node.object?.id == obj.id`. The result from `breadthFirstTraversal` will be the found node.

kenzieschmoll:
DBC: when selecting a link, does something need to be selected in the Outline tree?

bkonyi:
> DBC: when selecting a link, does something need to be selected in the Outline tree?

Not a DBC when you're already added as a reviewer... ;-)

We do the selection in the outline tree when we push the found object into the object viewport (see line 76 in `object_inspector_view_controller.dart`).

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
```diff
@@ -299,4 +299,91 @@ class ProgramExplorerController extends DisposableController
...
     }
   }
+
+  /// Searches and returns the script or library node in the FileExplorer
+  /// which is the source location of the target [object].
+  Future<VMServiceObjectNode> searchFileExplorer(ObjRef object) async {
+    final service = serviceManager.service!;
+    final isolateId = serviceManager.isolateManager.selectedIsolate.value!.id!;
+
+    ScriptRef? targetScript;
```
kenzieschmoll, line 309: nit: define this after the following if statement so that it is defined closer to where it is used.

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -609,6 +609,63 @@ class InboundReferencesWidget extends StatelessWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    String text = '';
+    if (textBuilder == null) {
+      if (object is LibraryRef) {
+        final lib = object as LibraryRef;
+        if (lib.uri!.startsWith('dart') || preferUri) {
+          text = lib.uri!;
+        } else {
+          text = lib.name!.isEmpty ? lib.uri! : lib.name!;
```
kenzieschmoll, line 634: nit: pull `lib.name` into local to avoid double null assert

packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
```diff
@@ -271,6 +271,12 @@ extension CodePrivateViewExtension on Code {
   Disassembly get disassembly => Disassembly.parse(json![_disassemblyKey]);
   set disassembly(Disassembly disassembly) =>
       json![_disassemblyKey] = disassembly.toJson();
+
+  /// Returns the function from which this code object was generated.
+  FuncRef? get function {
+    final functionJson = json!['function'] as Map<String, dynamic>;
```
kenzieschmoll, line 277: store 'function' in var _functionKey like what is done for `_disassemblyKey`

kenzieschmoll (approved):

bkonyi:
Closing in favor of https://github.com/flutter/devtools/pull/4466 due to the failing CLA check.

PR #4394, opened 2022-08-19
Comment out _updateLocalClasses code 

CoderDake:
The `_updateLocalClasses` code hits the inspector service with a lot of `getObject` calls. especially in larger projects.
`isLocalClass` isn't really fully set up in the UI Yet anyways, so we don't need to fetch all of the classes yet anyways.

I've left a TODO for our future selves when we come back to this and need to reenable the fetch.

In this PR I've also wrapped the `inferPubrootDirectory` code in a busy tracker so that the pub root directory spinner will spin when we are trying to infer pub root directories.

# Testing
I've tested this code internally and it has drastically reduce slowness on startup

### DIFF
--- a/packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
+++ b/packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
@@ -388,25 +388,26 @@ class InspectorService extends InspectorServiceBase {
   }
 
   Future<void> _updateLocalClasses() async {
-    localClasses.clear();
-    if (_rootDirectories.value.isNotEmpty) {
-      final isolate = inspectorLibrary.isolate!;
-      for (var libraryRef in isolate.libraries!) {
-        if (isLocalUri(libraryRef.uri!)) {
-          try {
-            final Library library = await inspectorLibrary.service
-                .getObject(isolate.id!, libraryRef.id!) as Library;
-            for (var classRef in library.classes!) {
-              localClasses[classRef.name!] = classRef;
-            }
-          } catch (e) {
-            // Workaround until https://github.com/flutter/devtools/issues/3110
-            // is fixed.
-            assert(serviceManager.connectedApp!.isDartWebAppNow!);
-          }
-        }
-      }
-    }
+    // TODO(https://github.com/flutter/devtools/issues/4393)
+    // localClasses.clear();
+    // if (_rootDirectories.value.isNotEmpty) {
+    //   final isolate = inspectorLibrary.isolate!;
+    //   for (var libraryRef in isolate.libraries!) {
+    //     if (isLocalUri(libraryRef.uri!)) {
+    //       try {
+    //         final Library library = await inspectorLibrary.service
+    //             .getObject(isolate.id!, libraryRef.id!) as Library;
+    //         for (var classRef in library.classes!) {
+    //           localClasses[classRef.name!] = classRef;
+    //         }
+    //       } catch (e) {
+    //         // Workaround until https://github.com/flutter/devtools/issues/3110
+    //         // is fixed.
+    //         assert(serviceManager.connectedApp!.isDartWebAppNow!);
+    //       }
+    //     }
+    //   }
+    // }
   }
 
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
```diff
@@ -388,25 +388,26 @@ class InspectorService extends InspectorServiceBase {
   }
 
   Future<void> _updateLocalClasses() async {
-    localClasses.clear();
-    if (_rootDirectories.value.isNotEmpty) {
-      final isolate = inspectorLibrary.isolate!;
-      for (var libraryRef in isolate.libraries!) {
-        if (isLocalUri(libraryRef.uri!)) {
-          try {
-            final Library library = await inspectorLibrary.service
```
bkonyi, line 397: Ah, this is probably the culprit for the worst of the slowdown. This can probably be updated to something like this:

```dart
final isolate = inspectorLibrary.isolate!;
final futures = <Future>[];
for (final libraryRef in isolate.libraries!) {
  if (isLocalUri(libraryRef.uri!)) {
    final future = inspectorLibrary.service.getObject(
        isolate.id!, libraryRef.id!
      ).then((result) {
        final library = result as Library;
        for (final classRef in library.classes!) {
          localClasses[classRef.name!] = classRef;
        }
      });
    futures.add(future);
  }
}
await Future.wait(futures);
```

This will make it so that you're not waiting on individual requests before requesting the next library, so we can really take advantage of parallelism.

packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
```diff
@@ -388,25 +388,26 @@ class InspectorService extends InspectorServiceBase {
...
-            }
-          } catch (e) {
-            // Workaround until https://github.com/flutter/devtools/issues/3110
-            // is fixed.
-            assert(serviceManager.connectedApp!.isDartWebAppNow!);
-          }
-        }
-      }
-    }
+    // TODO(https://github.com/flutter/devtools/issues/4393)
```
kenzieschmoll, line 391: can you reference this bug from the isLocalClass getter too? Just some someone in the future doesn't try to use isLocalClass and get confused why it's always false.

packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
CoderDake: I'll add this feedback to the issue.
I'm not sure it would 100% solve the issue as the network gets flooded with requests, so it may still slow things down

packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
CoderDake: great idea!

kenzieschmoll (approved):

packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
CoderDake: comment added :)

packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
bkonyi: Generally speaking, multiple outstanding network requests don't slow things down nearly as much as serializing them. This is a common pattern we use when we need to make a bunch of service requests (e.g., in the [`ProgramExplorer`](https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart#L224) we do this to request all fields and functions in libraries and classes).

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
jacob314: +1 to this pattern. Can't believe I wrote taht code. Wish I had an aggressive lint that said "no await within a for loop"

PR #4396, opened 2022-08-19
Add unsafe_html lint to DevTools

kenzieschmoll:
Ignoring the one exception case, but enabling this lint to catch other violations.


packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_controller_web.dart
```diff
@@ -42,6 +42,7 @@ class PerfettoController extends DisposableController
   void init() {
     _perfettoReady = Completer();
     _perfettoIFrame = html.IFrameElement()
+      // ignore: unsafe_html
```
jacob314, line 45: comment that this is safe because we have built the perfetto URL up ourself not received it from the user.

jacob314 (approved):
lgtm

CoderDake (approved):

PR #4403, opened 2022-08-23
Add support for a `Call Tree` view in the `Allocation Tracing` pane

bkonyi:
**Screenshot:**
image

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart
+++ b/packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart
@@ -11,6 +11,7 @@ import '../../../../shared/table.dart';
 import '../../../../shared/table_data.dart';
 import '../../../../shared/theme.dart';
 import '../../../../shared/utils.dart';
+import '../../../../ui/tab.dart';
 import '../../../profiler/cpu_profile_columns.dart';
 import '../../../profiler/cpu_profile_model.dart';
 import 'allocation_profile_tracing_view_controller.dart';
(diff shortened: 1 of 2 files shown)


polina-c:
It is not clear what numbers mean. How about replacing 'Traced allocations for: _Future' with 'Number of allocated instances of  _Future:'?

polina-c:
What is the motivation to add Call Tree? Did I miss some discussion?

From my perspective the buttons 'Bottom Up' and 'Call Tree' add unnecessary complexity to UI and to the code. I vote to have just 'Bottom Up' way as it is closer to what developers see in IDEs and console.

What is the value in 'Call Tree' that shows internal platform methods on the top, that are not interesting for developer.

bkonyi:
> What is the motivation to add Call Tree? Did I miss some discussion?

Maybe you just missed a [comment](https://github.com/flutter/devtools/pull/4403/files#diff-9affa23165d68f4775af61cd02c2fd40267c269ee1467f7d70f88f60466913a6L68), but I feel like this had come up before. Either way, it's a small change to add / remove, so we can always add it and then remove it when we do UX review (I think we're almost at the point where it's time to demo for @InMatrix et al).

> From my perspective the buttons 'Bottom Up' and 'Call Tree' add unnecessary complexity to UI and to the code. I vote to have just 'Bottom Up' way as it is closer to what developers see in IDEs and console.
> 
> What is the value in 'Call Tree' that shows internal platform methods on the top, that are not interesting for developer?

'Bottom Up' is more useful for determining _where_ the allocations were made, but that representation makes it more difficult to take more than a single allocation site into account when doing analysis. The 'Call Tree' format makes it easier to determine common allocation paths (e.g., two roots in the 'Bottom Up' view could result from a call to a common ancestor further down the stack and removing / reducing calls to that common ancestor could be a way to reduce allocations of the traced object type). I agree that the 'Bottom Up' is generally more useful and should be the view we default to, but I do think there is value in also providing a 'Call Tree' view as an option.

> And I see in 'Bottom Up' Exclusive is always equal to Inclusive. Should we have just one column named 'Instances'?

That sounds like a bug, actually. Exclusive should only equal Inclusive for all nodes iff all nodes are root nodes. Otherwise, Exclusive <= Inclusive.

InMatrix:
Just a reminder that UX review won't need a fully implemented prototype but an end-to-end usage scenario. Some parts of that scenario can be fake (e.g., static mocks, hard-coded data, placeholder description, etc). 

My drive-by comment is that the column headers can use some tooltips. I'm not sure what "Inclusive" and "Exclusive" mean, for example.

bkonyi:
> My drive-by comment is that the column headers can use some tooltips. I'm not sure what "Inclusive" and "Exclusive" mean, for example.

There's already tooltips for those columns... :-)

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart
```diff
@@ -145,11 +200,24 @@ class _AllocationProfileTracingTreeHeader extends StatelessWidget {
...
           onPressed: () => updateTreeStateCallback(
             () {
+              final isBottomUp = tabs[tabController.index] ==
+                  AllocationTracingTree._bottomUpTab;
               final data = controller.selectedTracedClassAllocationData!;
-              for (final root in data.bottomUpRoots) {
+              final roots =
+                  isBottomUp ? data.bottomUpRoots : data.callTreeRoots;
+              for (final root in roots) {
                 root.expandCascading();
```
kenzieschmoll, line 221: can you remind me again why we are expanding all the roots? Expanding all call tree roots could prevent the user from seeing more than the first root's call path.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart
```diff
@@ -159,8 +227,12 @@ class _AllocationProfileTracingTreeHeader extends StatelessWidget {
         CollapseAllButton(
           onPressed: () => updateTreeStateCallback(
             () {
+              final isBottomUp = tabs[tabController.index] ==
+                  AllocationTracingTree._bottomUpTab;
               final data = controller.selectedTracedClassAllocationData!;
-              for (final root in data.bottomUpRoots) {
+              final roots =
```
kenzieschmoll, line 233: create a helper _currentDataRoots that can be used both the expand and collapse button onPressed functions

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart
bkonyi: We don't do this by default, only when the user presses the `Expand All` button.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_tree.dart
bkonyi: Done.

kenzieschmoll (approved):

Issue #4300, opened 2022-07-26
Memory > Profile sorting does not persist.


polina-c:
If I change default sorting, and table refreshes, manually or automatically, the data are sorted with default sorting.

PR #4405, opened 2022-08-23
Persist sorting order for Memory Profile table

bkonyi:
Also changes default sort column to total size.

Fixes https://github.com/flutter/devtools/issues/4300

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
+++ b/packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
@@ -23,7 +23,7 @@ import 'allocation_profile_table_view_controller.dart';
 
 /// The default width for columns containing *mostly* numeric data (e.g.,
 /// instances, memory).
-const _defaultNumberFieldWidth = 80.0;
+const _defaultNumberFieldWidth = 90.0;
 
 class _FieldClassName extends ColumnData<ClassHeapStats> {
   _FieldClassName()
@@ -208,7 +208,7 @@ class AllocationProfileTableViewState
         ),
         Expanded(
           child: _AllocationProfileTable(
-            allocationProfileController: widget.controller,
+            controller: widget.controller,
           ),
         ),
       ],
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
```diff
@@ -234,10 +234,14 @@ class _AllocationProfileTable extends StatelessWidget {
     ),
   ];
 
+  static final _initialSortColumn = _FieldSizeColumn(
+    heap: _HeapGeneration.total,
+  );
+
   static final _columns = [
     _FieldClassName(),
```
polina-c, line 242: Add postfix 'Column' for consistency to this class?

packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
```diff
@@ -265,12 +269,28 @@ class _AllocationProfileTable extends StatelessWidget {
...
+  final AllocationProfileTableViewController controller;
+
+  @override
+  State<_AllocationProfileTable> createState() =>
+      _AllocationProfileTableState();
+}
+
+class _AllocationProfileTableState extends State<_AllocationProfileTable> {
+  late SortDirection sortDirection;
+  late ColumnData<ClassHeapStats> sortColumn;
```
polina-c, line 281: Should these two fields be private?

bkonyi: They definitely can be, I'm just a little bit looser on naming for members of private classes I guess... :-) Done.

packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
```diff
@@ -265,12 +269,28 @@ class _AllocationProfileTable extends StatelessWidget {
...
+  final AllocationProfileTableViewController controller;
+
+  @override
+  State<_AllocationProfileTable> createState() =>
+      _AllocationProfileTableState();
+}
+
+class _AllocationProfileTableState extends State<_AllocationProfileTable> {
+  late SortDirection _sortDirection;
+  late ColumnData<ClassHeapStats> _sortColumn;
```
kenzieschmoll, line 281: can you link to a bug to store these values in a TableController? or if one does not exist, can you create one? I believe this is now the second case where we are storing these value for a Table in the widget that creates the Table

kenzieschmoll (approved):

PR #4406, opened 2022-08-23
Create helper for bumping dev version

CoderDake:
I've refactored `tool/update_version.dart` so that it's code can be shared with `tool/update_dev_version.dart`.

`tool/update_dev_version.dart` can now be run to bump the dev version.

## Testing
I've manually checked that:
- `-dev.0` is added if no dev version is present
- the `-dev.#` version is incremented when one is present.


tool/update_dev_version.dart
```diff
@@ -0,0 +1,39 @@
+import 'update_version.dart' as update_version;
```
bkonyi, line 1: Add a copyright header to this file.

CoderDake: 🤦

elliette:
Let's update the documentation as well: https://github.com/flutter/devtools/blob/master/tool/README.md#update-the-devtools-version-number

Is there a reason to have this as a separate script instead of passing `dev` as an argument to `update_version.dart` (eg, `dart tool/update_version.dart dev` would update the dev version)?

CoderDake:
@elliette no specific reason other than that is the way jacob brought it up.
The flag definitely sounds a bit cleaner, so I'll quickly do that.

elliette (approved):
This will be very helpful 🎉

tool/update_version.dart
```diff
@@ -5,80 +5,90 @@
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:args/command_runner.dart';
+
 // This script must be executed from the top level devtools/ directory.
 // TODO(kenz): If changes are made to this script, first consider refactoring to
 // use https://github.com/dart-lang/pubspec_parse.
```
CoderDake, line 12:
```suggestion
// use https://github.com/dart-lang/pubspec_parse.

```

Issue #4407, opened 2022-08-23
Clicking on file search results does not select file on Flutter master branch


elliette:
Clicking on a file in the file search results should open and select a file. When on the Flutter master branch, clicking one of the results causes the file search to lose focus, therefore closing the file search dropdown (and not selecting the file). 

I can reproduce this on the Flutter master channel, but not when synced to the [`3.3.0-0.4.pre` beta tag](https://github.com/flutter/flutter/releases/tag/3.3.0-0.4.pre) (from 4 days ago). This seems like it's a regression introduced by a recent Flutter change.

![Screen Shot 2022-08-23 at 12 33 39 PM](https://user-images.githubusercontent.com/21270878/186250128-b7af2fe7-d215-4d97-80de-a4a92aaf3c3f.png)

elliette:
Actually, looks like there are a quite a few changes between the `3.3.0-0.4.pre` beta tag and master: https://github.com/flutter/flutter/compare/3.3.0-0.4.pre...master

elliette:
Looks like this was introduced in https://github.com/flutter/flutter/pull/107262 When using Flutter at the commit right before that commit, clicking a file works. When using Flutter at that commit, clicking a file doesn't work. 

@gspencergoog is it expected that clients will need to change how they are handling focus for things to continue working after https://github.com/flutter/flutter/pull/107262? For context, a lot of the focus logic for DevTools' file search was added in https://github.com/flutter/devtools/pull/3612. 

As a side note, I'm surprised this wasn't caught by this test case: https://github.com/flutter/devtools/blob/master/packages/devtools_app/test/debugger/file_search_test.dart#L35-L76 I will need to do some investigation to figure out why this wasn't caught.

elliette:
Actually thinking about this more:

I wonder if the issue is due to https://github.com/flutter/devtools/pull/3602, which was a workaround we implemented in DevTools to resolve https://github.com/flutter/flutter/issues/97389

It seems like  https://github.com/flutter/flutter/issues/97389 might be fixed by https://github.com/flutter/flutter/pull/107262. @gspencergoog is that correct? Thanks!

gspencergoog:
> It seems like https://github.com/flutter/flutter/issues/97389 might be fixed by https://github.com/flutter/flutter/pull/107262. @gspencergoog is that correct? Thanks!

[...]

PR #4409, opened 2022-08-23
Clicking on file name selects the file in file search autocomplete

elliette:
Fixes https://github.com/flutter/devtools/issues/4407, see discussion on the bug. 

TLDR; recent change to Flutter framework broke the ability to select a file search autocomplete result by clicking on it. Wrapping the results in the new `TextFieldTapRegion` widget tells the framework to keep the focus on the file search text field, and therefore not close the autocomplete dropdown when a result in the dropdown is clicked.


elliette:
Tests failing because we are on an older version of Flutter, opened https://github.com/flutter/devtools/pull/4410

CoderDake (approved):
Lgtm when we can get our checks green

kenzieschmoll (approved):

Issue #1033, opened 2019-09-10
Audit bottom up total time / self time logic - are they always the same?


escamoteur:
Hi,

just observed that the total time in the BottomUp View doesn't match that of the CPU Flame chart. On a closer look it turns out that currently on the bottom up view the total time is always equal to the Self Time. Maybe I understand the graph just wrong but what is it good ?

Cheers
Thomas

kenzieschmoll:
Total time for the Bottom Up view is expected to be different than the CPU flame chart. The definitions for total / self time are slightly different for [Bottom Up](https://flutter.dev/docs/development/tools/devtools/performance#bottom-up) and [Call Tree](https://flutter.dev/docs/development/tools/devtools/performance#call-tree).

I'd encourage you to read the definitions at the provided links to understand the differences. Thanks!

escamoteur:
I read it and this here is stated:

```
Total time
Time the method spent executing its own code as well as the code for its callee.
```
Which I read (maybe a language issue) its the time of the method plus the time of all the methods that method calls. 
Am I wrong.

Besides that if both times are always the same, it might make sense not to show both columns.

kenzieschmoll:
You are correct. One small clarification though - for a the bottom up chart, each "caller" will only have _one_ "callee" (the node up one level in the tree hierarchy). 
```
Call Tree
A
 |__ B
       |__ C
```
```
Bottom Up
C
 |__ B
       |__ A
```
So for the bottom up chart, the total time for node A will be the time spent executing A(caller) as well as the time spent executing B (the callee).

Self time and Total time should not always be the same. I'll reopen this issue to track auditing that to make sure that we do not have a bug in our calculation logic.

escamoteur:
So does the total time of B consist of the self time of B + the self time of C but does not contain any other methods B might call?

I guess I'm not sure what the value of the bottom up view is, perhaps this could be explained a bit more in the docs.

Here is another example where self and total time is always the same

![image](https://user-images.githubusercontent.com/16635729/64680275-c3a6cd80-d47d-11e9-8057-f07be734d8e6.png)

kenzieschmoll:
Correct. In the bottom up chart, we care more about patterns at the bottom of the call tree (or the top of the call stack if you want to view it that way). 

```
Call Tree
A
 |__ B
       |__ C
D
 |__ E
       |__ C
F
|__ B
      |__C
```
```
Bottom Up
C
 |__ B
 |     |__ A
 |     |__ F
[...]

PR #4413, opened 2022-08-25
Fix misc bugs in CPU profile trees

bkonyi:
Fixed invalid time ranges for filtered CPU profiles and removed unused user tag logic.

### DIFF
--- a/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
+++ b/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
@@ -160,8 +160,11 @@ class CpuProfileData {
       // times wouldn't be helpful.
       time: TimeRange()
         ..start = const Duration()
-        ..end =
-            Duration(microseconds: (newSampleCount * microsPerSample).round()),
+        ..end = Duration(
+          microseconds: microsPerSample.isInfinite
+              ? 0
+              : (newSampleCount * microsPerSample).round(),
+        ),
     );
 
     // Use a SplayTreeMap so that map iteration will be in sorted key order.
@@ -258,7 +261,9 @@ class CpuProfileData {
       time: TimeRange()
         ..start = const Duration()
         ..end = Duration(
-          microseconds: (filteredCpuSamples.length * microsPerSample).round(),
+          microseconds: microsPerSample.isInfinite
+              ? 0
+              : (filteredCpuSamples.length * microsPerSample).round(),
         ),
     );
 
--- a/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
+++ b/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
@@ -705,7 +705,7 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
   /// profile, changing the exclusive counts.
   ///
   /// When [copySampleCounts] is true, inclusive sample counts are also reset
-  /// by default, unless [resetInclusiveSampleCount] is also set to false. 
+  /// by default, unless [resetInclusiveSampleCount] is also set to false.
   /// Inclusive sample counts should only be copied as part of a deep copy of
   /// a tree.
   @override
--- a/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
+++ b/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
@@ -176,7 +176,7 @@ class CpuProfileData {
 
       while (currentStackFrame != null) {
         stackFramesWithTag[currentId!] = currentStackFrame.shallowCopy(
-          copyExclusiveSampleCount: false,
+          copySampleCounts: false,
           profileMetaData: metaData,
         );
         final parentId = currentStackFrame.parentId;
@@ -246,8 +246,20 @@ class CpuProfileData {
       return null;
     }
 
+    final originalTime = originalData.profileMetaData.time!.duration;
+    final microsPerSample =
+        originalTime.inMicroseconds / originalData.profileMetaData.sampleCount;
     final updatedMetaData = originalData.profileMetaData.copyWith(
       sampleCount: filteredCpuSamples.length,
+      // The start time is zero because only `TimeRange.duration` will matter
+      // for this profile data, and the samples included in this data could be
+      // sparse over the original profile's time range, so true start and end
+      // times wouldn't be helpful.
+      time: TimeRange()
+        ..start = const Duration()
+        ..end = Duration(
+          microseconds: (filteredCpuSamples.length * microsPerSample).round(),
+        ),
     );
 
     void walkAndFilter(CpuStackFrame stackFrame) {
@@ -256,7 +268,7 @@ class CpuProfileData {
         final filteredParentId = filteredParentStackFrameId(parent);
         filteredStackFrames[stackFrame.id] = stackFrame.shallowCopy(
           parentId: filteredParentId,
-          copyExclusiveSampleCount: false,
+          copySampleCounts: false,
           profileMetaData: updatedMetaData,
         );
         if (filteredParentId != null) {
@@ -685,24 +697,12 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
     ].join(' - ');
   }
 
-  /// [copyExclusiveSampleCount] and [copyInclusiveSampleCount] control whether
-  /// or not the resulting [CpuStackFrame] will have [exclusiveSampleCount] and
-  /// [inclusiveSampleCount] initialized.
+  /// [copySampleCounts] control whether or not the resulting [CpuStackFrame]
+  /// will have [exclusiveSampleCount] and [inclusiveSampleCount] initialized.
   ///
-  /// By default, a shallow copy will copy the exclusive sample count from the
-  /// original [CpuStackFrame], but will set `_inclusiveSampleCount` to null,
-  /// causing it to be re-initialized the first time [inclusiveSampleCount] is
-  /// accessed. This is the default behavior as, in contexts where we're
-  /// creating a non-filtered copy of the tree (i.e., the tree contains the
-  /// same number of samples), the exclusive count for each node will always be
-  /// the same, whereas the inclusive count may change based on how the tree is
-  /// rooted (i.e., call tree vs bottom up view).
-  ///
-  /// Exclusive sample counts should only be reset when building a filtered
-  /// view of the full set of samples, as some stacks may no longer be included
-  /// in the profile, changing the exclusive counts.
-  ///
-  /// Inclusive sample counts should only be copied when performing a deep copy.
+  /// Sample counts should only be reset when building a filtered view of the
+  /// full set of samples, as some stacks may no longer be included in the
+  /// profile, changing the exclusive counts.
   @override
   CpuStackFrame shallowCopy({
     String? id,
@@ -714,8 +714,7 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
     String? parentId,
     int? sourceLine,
     CpuProfileMetaData? profileMetaData,
-    bool copyExclusiveSampleCount = true,
-    bool copyInclusiveSampleCount = false,
+    bool copySampleCounts = true,
   }) {
     final copy = CpuStackFrame._(
       id: id ?? this.id,
@@ -728,11 +727,11 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
       parentId: parentId ?? this.parentId,
       profileMetaData: profileMetaData ?? this.profileMetaData,
     );
-    copy
-      ..exclusiveSampleCount =
-          copyExclusiveSampleCount ? exclusiveSampleCount : 0
-      ..inclusiveSampleCount =
-          copyInclusiveSampleCount ? inclusiveSampleCount : null;
+    if (copySampleCounts) {
+      copy
+        ..exclusiveSampleCount = exclusiveSampleCount
+        ..inclusiveSampleCount = inclusiveSampleCount;
+    }
     return copy;
   }
 
@@ -741,7 +740,7 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
   /// The returned copy stack frame will have a null parent.
   @override
   CpuStackFrame deepCopy() {
-    final copy = shallowCopy(copyInclusiveSampleCount: true);
+    final copy = shallowCopy();
     for (CpuStackFrame child in children) {
       copy.addChild(child.deepCopy());
     }
--- a/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
+++ b/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
@@ -704,9 +704,10 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
   /// full set of samples, as some stacks may no longer be included in the
   /// profile, changing the exclusive counts.
   ///
-  /// Inclusive sample counts are also reset by default, unless
-  /// [resetInclusiveSampleCounts] is set to false. Inclusive sample counts
-  /// should only be copied as part of a deep copy of a tree.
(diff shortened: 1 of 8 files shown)


CoderDake (approved):
I don't know much about the guts of this profiling stuff yet, but these code looks fine to me. Especially if it gets us closer to the expected representation that observatory has.

kenzieschmoll:
Did you compare to other profiling tools like Chrome DevTools? Or verify against a CPU profile and not just an allocation tracing profile? Since this affects more than just allocation tracing, I want to make sure we cover our bases here.

bkonyi:
> Did you compare to other profiling tools like Chrome DevTools?

Chrome DevTools appears to be consistent with Observatory's bottom-up representation.

> Or verify against a CPU profile and not just an allocation tracing profile? Since this affects more than just allocation tracing, I want to make sure we cover our bases here.

I actually noticed this first in the CPU profile, and it looks like there's an outstanding issue for this: https://github.com/flutter/devtools/issues/1033.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -729,8 +728,7 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
     if (copySampleCountsAndTags) {
       copy
         ..exclusiveSampleCount = exclusiveSampleCount
-        ..inclusiveSampleCount =
-            resetInclusiveSampleCount ? null : inclusiveSampleCount;
```
kenzieschmoll, line 733: removing this makes me a bit nervous. Anywhere `shallowCopy` was used where `resetInclusiveSampleCount: false` was not specified will now be affected. For example, `CpuStackFrame.shallowCopy` is used in the filtering code (filtering by user tag or source url). I'd be a bit nervous that these code paths are affected by this change. Have you tried filtering by a user tag or by the filter categories (hide native, hide flutter, etc.)?

kenzieschmoll (commented):
Can you take a look at the tooltips for Total time and Self time columns for the Call tree and Bottom up tables? I'd like to get your input on whether these are really communication the right thing (especially for bottom up).

polina-c:
Now I see ComponentElement.performRebuild creates 13 inclusive instances of Text.
While I see only 12 exclusive in the call tree. How can it be?
(i used counter application)


Screen Shot 2022-09-02 at 1 15 28 PM

polina-c:
Do you think fractions for percentage can be important for user?
Would it be better to save screen real estate showing whole numbers (like '22%')?

polina-c:
Feature request that is nice to have and can be outside of this PR.
It would be nice if filter accepted syntax to search whole word. We can take standard search UX: when I type `"Text"`, I will get exact 'Text' classes, not 100 classes that has 'Text' in name.
I expect users who need it to just try it and discover.

bkonyi:
> Now I see ComponentElement.performRebuild creates 13 inclusive instances of Text. While I see only 12 exclusive in the call tree. How can it be? (i used counter application)
> 
> Screen Shot 2022-09-02 at 1 15 28 PM

There's 13 instances allocated, just one was allocated in a call to `_TooltipOverlay.build` from `ComponentElement.performRebuild`.

bkonyi:
> Feature request that is nice to have and can be outside of this PR. It would be nice if filter accepted syntax to search whole word. We can take standard search UX: when I type `"Text"`, I will get exact 'Text' classes, not 100 classes that has 'Text' in name. I expect users who need it to just try it and discover.

That sounds like a good idea, but I'm not sure I'll have cycles for it. Maybe file an issue for later work? :-)

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
bkonyi: I just checked, I don't see any regressions with this patch vs the current release.

bkonyi:
> Can you take a look at the tooltips for Total time and Self time columns for the Call tree and Bottom up tables? I'd like to get your input on whether these are really communication the right thing (especially for bottom up).

Self-time is the same metric for call tree and bottom up views, and that can be described as "the percentage of samples collected where this is the top frame". Total time just outlines the percentage of samples collected where a given frame was on the stack.

In general, `sum(selfTimes) == 1.0` and `sum(totalTimes) >= 1.0`.

polina-c:
> > Now I see ComponentElement.performRebuild creates 13 inclusive instances of Text. While I see only 12 exclusive in the call tree. How can it be? (i used counter application)
> > Screen Shot 2022-09-02 at 1 15 28 PM
> 
> There's 13 instances allocated, just one was allocated in a call to `_TooltipOverlay.build` from `ComponentElement.performRebuild`.

I see

polina-c:
> > Do you think fractions for percentage can be important for user? Would it be better to save screen real estate showing whole numbers (like '22%')?
> 
> I don't think dropping fractions of a percent is a good idea, it'll just result in confusing values being displayed (e.g., percentages won't add up to or will overshoot 100%, 0.5% would either be displayed as 0% or 1%, neither of which are correct) and remove precision for very little gain.

With fractions you still can have 0.00%, that will be 'incorrect'. But:

1. Do not people normally expect that numbers are rounded up when they see percents?
2. We show number of instances anyway, so there will not be confusion

My suggestion is that difference in less than 1% is not interesting for memory analysis/optimization, so the space taken by three extra symbols on each row does not add value, but adds screen noise and extra space.
To communicate that numbers are rounded up (if it is needed) we may add '~' in front.

Outlining the options:
1. 86.05%
2. 86%
3. ~86%

@melyndahoover, what do you think?

bkonyi:
> > > Do you think fractions for percentage can be important for user? Would it be better to save screen real estate showing whole numbers (like '22%')?
> > 
> > 
> > I don't think dropping fractions of a percent is a good idea, it'll just result in confusing values being displayed (e.g., percentages won't add up to or will overshoot 100%, 0.5% would either be displayed as 0% or 1%, neither of which are correct) and remove precision for very little gain.
> 
> With fractions you still can have 0.00%, that will be 'incorrect'. But:
> 
> 1. Do not people normally expect that numbers are rounded, when they see percents?

Sure, but rounding to zero decimal points causes significantly more information to be lost than rounding to two decimal points.

> 2. We show number of instances anyway, so there will not be confusion

Only for the allocation tracing tree, not for the CPU Profiler view which shares a lot of the same logic.

> My suggestion is that difference in less than 1% is not interesting for memory analysis/optimization, so the space taken by three extra symbols on each row does not add value, but adds screen noise and extra space. To communicate that numbers are rounded up (if it is needed) we may add '~' in front.
> 
> Outlining the options:
> 
> 1. 86.05%
> 2. 86%
> 3. ~86%

I really don't think it's worth losing precision for the sake of a few pixels when we can spend some time to better optimize the tree column size calculation, which currently results in the tree column taking up way more space than it needs.

kenzieschmoll:
FWIW, the `msText` util has functionality `allowRoundingToZero` that, when false, will output something like '< 0.01 ms'. I am in favor of keeping the precision we currently have (2 decimal places), but we could at least do something like this for the '0.00' percentages.

+1 to Ben's comment that we should focus space saving efforts on optimizing the column size calculations.

melyndahoover:
I think it is important to think about the use case and when/if it is important for the user to have higher precision in this number. If the user just needs a quick gut check, then rounded percentages are probably fine. What kind of use cases would require high precision values here?

polina-c:
> FWIW, the `msText` util has functionality `allowRoundingToZero` that, when false, will output something like '< 0.01 ms'. I am in favor of keeping the precision we currently have (2 decimal places), but we could at least do something like this for the '0.00' percentages.
> 
> +1 to Ben's comment that we should focus space saving efforts on optimizing the column size calculations.

I agree that optimizing the column size is not important in this case. I care more about about the screen noisiness for the user.

polina-c:
> > > > Do you think fractions for percentage can be important for user? Would it be better to save screen real estate showing whole numbers (like '22%')?
> > > 
> > > 
> > > I don't think dropping fractions of a percent is a good idea, it'll just result in confusing values being displayed (e.g., percentages won't add up to or will overshoot 100%, 0.5% would either be displayed as 0% or 1%, neither of which are correct) and remove precision for very little gain.
> > 
> > 
> > With fractions you still can have 0.00%, that will be 'incorrect'. But:
> > 
> > 1. Do not people normally expect that numbers are rounded, when they see percents?
> 
> Sure, but rounding to zero decimal points causes significantly more information to be lost than rounding to two decimal points.
> 
> > 2. We show number of instances anyway, so there will not be confusion
> 
> Only for the allocation tracing tree, not for the CPU Profiler view which shares a lot of the same logic.
> 
> > My suggestion is that difference in less than 1% is not interesting for memory analysis/optimization, so the space taken by three extra symbols on each row does not add value, but adds screen noise and extra space. To communicate that numbers are rounded up (if it is needed) we may add '~' in front.
> > Outlining the options:
> > 
> > 1. 86.05%
> > 2. 86%
> > 3. ~86%
> 
> I really don't think it's worth losing precision for the sake of a few pixels when we can spend some time to better optimize the tree column size calculation, which currently results in the tree column taking up way more space than it needs.

Yes, we do not want to loose anything to save pixels. I care about amount of noise for user on the screen.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -245,13 +246,18 @@ class CpuProfileData {
       return null;
     }
 
+    final updatedMetaData = originalData.profileMetaData.copyWith(
+      sampleCount: filteredCpuSamples.length,
```
kenzieschmoll, line 250: do we also need to update the time here like we did for the filter by user tag case?

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -712,7 +696,8 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
     String? parentId,
     int? sourceLine,
     CpuProfileMetaData? profileMetaData,
-    bool copySampleCountsAndTags = true,
+    bool copyExclusiveSampleCount = true,
+    bool copyInclusiveSampleCount = false,
```
kenzieschmoll, line 700: can both of these be false by default?

polina-c:
Yesterday on demo I saw two views: bottom up and call tree. When I load the PR and run DevTools, I see just one option. Is it under flag or setting?

polina-c:
If I expand the methods clicking them, source file names are still here, but if i click 'expand all', the sources disappear.

polina-c:
I get this if i connect to the app with vm developer mode off, open memory page, and then turn the vm developer mode 
on:

stack trace: https://paste.googleplex.com/6504044402049024

Screen Shot 2022-09-08 at 9 43 48 AM

polina-c:
I see full path in tooltip, but only part of it in the table:
Screen Shot 2022-09-08 at 9 39 18 AM

May be use Text.overflow.ellipsis with button 'Copy' like GitHub, so that user can locate the file in IDE:

Screen Shot 2022-09-08 at 9 52 30 AM

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
kenzieschmoll: can we add some descriptive dart doc about these two parameters explaining when they should be true and false? just to try to avoid any confusion or breakages in the future

bkonyi:
> Yesterday on demo I saw two views: bottom up and call tree. When I load the PR and run DevTools, I see just one option. Is it under flag or setting?

It sounds like you're on an old revision as the call tree change landed a week or two ago.

bkonyi:
> If I expand the methods clicking them, source file names are still here, but if i click 'expand all', the sources disappear.

I can't reproduce this, and I'm guessing it's just going off screen due to the buggy tree column width calculations.

bkonyi:
> I get this if i connect to the app with vm developer mode off, open memory page, and then turn the vm developer mode on:
> 
> stack trace: https://paste.googleplex.com/6504044402049024
> 
> Screen Shot 2022-09-08 at 9 43 48 AM
> 
> Can you create a test for this scenario?

This was an old exception that we used to throw. I think you're on a pretty old DevTools revision as this was fixed already.

bkonyi:
> I see full path in tooltip, but only part of it in the table: Screen Shot 2022-09-08 at 9 39 18 AM
> 
> May be use Text.overflow.ellipsis with button 'Copy' like GitHub:
> 
> Screen Shot 2022-09-08 at 9 52 30 AM

Yeah, this is off screen but can be scrolled to. Nothing's actually cut off here.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
bkonyi: As discussed offline, `copyExclusiveSampleCount` can't be false by default since the bottom-up tree builder in `profiler_utils.dart` expects a `TreeNode` with the default `shallowCopy()` signature.

I'll add some documentation here explaining them.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
bkonyi: Done.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -160,7 +176,8 @@ class CpuProfileData {
 
       while (currentStackFrame != null) {
         stackFramesWithTag[currentId!] = currentStackFrame.shallowCopy(
-          copySampleCountsAndTags: false,
+          copySampleCounts: false,
+          profileMetaData: metaData,
```
kenzieschmoll, line 180: I thought we don't need each stack frame to store a copy of the metadata anymore?

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -701,6 +697,16 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
...
   }
 
+  /// [copySampleCounts] control whether or not the resulting [CpuStackFrame]
+  /// will have [exclusiveSampleCount] and [inclusiveSampleCount] initialized.
+  ///
+  /// Sample counts should only be reset when building a filtered view of the
+  /// full set of samples, as some stacks may no longer be included in the
+  /// profile, changing the exclusive counts.
+  ///
+  /// Inclusive sample counts are also reset by default, unless
```
kenzieschmoll, line 707: maybe add here: "When [copySampleCounts] is true, inclusive counts are reset by default..." to make it clear that the `resetInclusiveSampleCounts` param only does something if `copySampleCounts` is true.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -712,8 +718,8 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
     String? parentId,
     int? sourceLine,
     CpuProfileMetaData? profileMetaData,
-    bool copySampleCountsAndTags = true,
-    bool resetInclusiveSampleCount = true,
+    bool copySampleCounts = true,
+    bool resetInclusiveSampleCounts = true,
```
kenzieschmoll, line 722: nit: rename back to `resetInclusiveSampleCount` since it is a singular value `CpuStackFrame.inclusiveSampleCount`

kenzieschmoll, line 722: nit: rename back to `resetInclusiveSampleCount` since it is a singular value `CpuStackFrame.inclusiveSampleCount`

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
bkonyi: We do, unless we want to do some refactoring to remove its usage in `ProfilableDataMixin`.

kenzieschmoll (approved):

PR #4415, opened 2022-08-26
Accessibility and LICENSE Fixes

CoderDake:
The License changes are required to be compliant with the licenses that Perfetto uses. 
It looks like that license is BSD-0.

I would like to make sure that this is an acceptable change, since I'm not too familiar with how we deal with Licensing changes in devtools yet.

The accessibility changes were found using internal tools, so I've fixed them here.


packages/devtools_app/LICENSE
```diff
@@ -25,3 +25,11 @@ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
 THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH
```
CoderDake, line 29: Before merging, this needs to be discussed with the team

vsmenon: lgtm

bkonyi (approved):

Issue #3986, opened 2022-04-15
Inspector throws exception if app is paused


kenzieschmoll:
Repro steps:
1. Connect to app
2. Go to devtools debugger and click 'Pause' to pause the app
3. Go back to Inspector page and observe exception

```
flutter: [ERROR]: Unrecognized error: [ErrorRef id: objects/1087, kind: UnhandledException, message: Unhandled exception:
Id does not exist.
#0      WidgetInspectorService.toObject (package:flutter/src/widgets/widget_inspector.dart:1300:7)
#1      WidgetInspectorService._getSelectedSummaryWidget (package:flutter/src/widgets/widget_inspector.dart:1887:48)
#2      WidgetInspectorService.getSelectedSummaryWidget (package:flutter/src/widgets/widget_inspector.dart:1880:28)
#3      Eval ()
] - WidgetInspectorService.instance.getSelectedSummaryWidget('null', 'selection_6')
flutter: [ERROR]: #0      EvalOnDartLibrary._eval (package:devtools_app/src/shared/eval_on_dart_library.dart:214:9)
<asynchronous suspension>
#1      EvalOnDartLibrary.addRequest.wrappedRequest (package:devtools_app/src/shared/eval_on_dart_library.dart:570:23)
<asynchronous suspension>
```

PR #4418, opened 2022-08-29
Fix inspector exception when app paused

CoderDake:
Fixes https://github.com/flutter/devtools/issues/3986

Trying to add a client while the app is paused was causing errors to be thrown.

This PR adds makes it so that the client is only added after the app is unpaused.

To accomplish this, new functionality was added to auto_dispose to add a type of listener that can be disposed early. This was the 2nd/3rd time that I've run into cases like this, so I figured that adding some functionality to help with this would be beneficial :)


packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -773,6 +775,12 @@ class _InspectorTreeState extends State<InspectorTree>
     }
     _focusNode = FocusNode(debugLabel: 'inspector-tree');
     autoDisposeFocusNode(_focusNode);
+    addConditionalAutoDisposeListener(
```
polina-c, line 778: Instead of using auto_dispose_mixin, you can just listen to _doneBindingToController and dispose the controller in the listener.
Will it be more readable and take less code then?

CoderDake: I chatted with Kenzie about this a bit and we decided to pivot the approach a bit so that we can still make things a bit easier for the caller. i've tweaked the helper so that it will make it much easier for us to run once now :D

packages/devtools_app/lib/src/primitives/auto_dispose.dart
```diff
@@ -75,6 +76,35 @@ class Disposer {
...
+  /// Any listeners set by [callOnceWhenReady] will auto dispose, or be removed after the callback is run.
+  void callOnceWhenReady<T>({
+    required VoidCallback callback,
+    required ValueListenable<T> trigger,
+    required bool Function(T triggerValue) readyWhen,
+  }) {
+    if (readyWhen(trigger.value)) {
+      callback();
+    } else {
+      // do the stuff to add the listener and remove it when appropriate
```
kenzieschmoll, line 97: is this comment still necessary?

packages/devtools_app/lib/src/primitives/auto_dispose.dart
```diff
@@ -75,6 +76,35 @@ class Disposer {
...
+      VoidCallback? earlyDisposeCallback;
```
kenzieschmoll, line 98: nit: rename to `listener` or `triggerListener`. I think the "early dispose" part is a bit confusing.

packages/devtools_app/lib/src/primitives/auto_dispose.dart
```diff
@@ -75,6 +76,35 @@ class Disposer {
...
+      earlyDisposeCallback = () {
+        if (readyWhen(trigger.value)) {
+          callback();
+          trigger.removeListener(earlyDisposeCallback!);
```
kenzieschmoll, line 102: shouldn't we also be removing this listener from the delegate here?

kenzieschmoll: or in other words, removing it from trigger from listenables and removing the callback from listeners

packages/devtools_app/test/shared/auto_dispose_mixin_test.dart
```diff
@@ -105,19 +103,131 @@ void main() {
...
 
-      // ignore: invalid_use_of_protected_member
       expect(notifier.hasListeners, isFalse);
       notifier.value = 21;
       expect(values.length, equals(3));
       expect(values.last, equals(19));
     });
+
+    group('callOnceWhenReady', () {
+      for (bool readyWhen in [false, true]) {
```
kenzieschmoll, line 119: nit: maybe rename to 'ready' since 'readyWhen' sounds like a function

packages/devtools_app/lib/src/primitives/auto_dispose.dart
CoderDake: 🤦 nope

packages/devtools_app/lib/src/primitives/auto_dispose.dart
CoderDake: ah yes you are spot on! then we will be full clean

packages/devtools_app/test/shared/auto_dispose_mixin_test.dart
```diff
@@ -105,19 +103,131 @@ void main() {
...
+            disposer.callOnceWhenReady(
+              trigger: trigger,
+              readyWhen: (triggerValue) => triggerValue == readyWhen,
+              callback: () {
+                counter++;
+                callbackEntries.add(counter);
+              },
+            );
+
+            expect(callbackEntries, equals([]));
```
kenzieschmoll, line 137: use isEmpty instead of equals([])

packages/devtools_app/test/shared/auto_dispose_mixin_test.dart
```diff
@@ -105,19 +103,131 @@ void main() {
...
       expect(notifier.hasListeners, isFalse);
       notifier.value = 21;
       expect(values.length, equals(3));
       expect(values.last, equals(19));
     });
+
+    group('callOnceWhenReady', () {
+      for (bool readyWhen in [false, true]) {
+        group('readyWhen=$readyWhen', () {
+          testWidgets('triggers callback and cancels listeners when ready ',
```
kenzieschmoll, line 121: I don't think these need to be widget tests. These can just be normal dart tests like above

packages/devtools_app/test/shared/auto_dispose_mixin_test.dart
```diff
@@ -105,19 +103,131 @@ void main() {
...
+              (WidgetTester tester) async {
+            final disposer = Disposer();
+            final trigger = ValueNotifier<bool?>(!readyWhen);
+            final callbackEntries = <int>[];
+            int counter = 0;
```
kenzieschmoll, line 126: we never actually check the value of counter. Either only check the length of callbackEntries or only check the value of counter. If we aren't ever checking the actual values in callbackEntries, then I'd recommend we just check the value of counter

packages/devtools_app/test/shared/auto_dispose_mixin_test.dart
```diff
@@ -105,19 +105,135 @@ void main() {
...
-      // ignore: invalid_use_of_protected_member
       expect(notifier.hasListeners, isFalse);
       notifier.value = 21;
       expect(values.length, equals(3));
       expect(values.last, equals(19));
     });
+
+    group('callOnceWhenReady', () {
+      for (bool isReady in [false, true]) {
+        group('readyWhen=$isReady', () {
```
kenzieschmoll, line 122: nit: change group name to 'isReady=$isReady'

kenzieschmoll (approved):
looks great!

polina-c (approved):

packages/devtools_app/test/shared/auto_dispose_mixin_test.dart
CoderDake: :facepalm:

PR #4422, opened 2022-08-30
Revert perfetto bundle

CoderDake:
The perfetto bundle changes have caused a chain of license issues that are blocking internal rolls.
I'm reverting for now until we can get more time to look at these.


bkonyi (approved):
RSLGTM

Issue #4227, opened 2022-06-29
Write tests for the ProgramExplorer


kenzieschmoll:

srawlins:
Comprehensive tests for ProgramExplorer were added in PR #4424 (see packages/devtools_app/test/screens/debugger/program_explorer_test.dart). Closing this issue as completed.

If there is more work to do here, please let us know by filing a new issue with up to date information. Thanks!

PR #4424, opened 2022-08-30
Improve testing of `ProgramExplorer`

bkonyi:
Adds a `FakeProgramExplorerController` that retains the vast majority of the functionality of `ProgramExplorerController` while allowing for fake data to be injected, removing all service requests.

Makes progress on https://github.com/flutter/devtools/issues/4227.

### DIFF
--- a/packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
@@ -21,9 +21,8 @@ class ProgramExplorerController extends DisposableController
   ProgramExplorerController({
     this.showCodeNodes = false,
     ListValueNotifier<VMServiceObjectNode>? rootObjectNodesOverride,
-  })  : 
-        _rootObjectNodes =
-            rootObjectNodesOverride ?? ListValueNotifier<VMServiceObjectNode>([]);
+  }) : _rootObjectNodes = rootObjectNodesOverride ??
+            ListValueNotifier<VMServiceObjectNode>([]);
 
   /// The outline view nodes for the currently selected library.
   ValueListenable<List<VMServiceObjectNode>> get outlineNodes => _outlineNodes;
--- a/packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
@@ -20,9 +20,7 @@ class ProgramExplorerController extends DisposableController
   /// outline view.
   ProgramExplorerController({
     this.showCodeNodes = false,
-    ListValueNotifier<VMServiceObjectNode>? rootObjectNodesOverride,
-  }) : _rootObjectNodes = rootObjectNodesOverride ??
-            ListValueNotifier<VMServiceObjectNode>([]);
+  });
 
   /// The outline view nodes for the currently selected library.
   ValueListenable<List<VMServiceObjectNode>> get outlineNodes => _outlineNodes;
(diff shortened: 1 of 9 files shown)


packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
```diff
@@ -20,7 +20,9 @@ class ProgramExplorerController extends DisposableController
   /// outline view.
   ProgramExplorerController({
     this.showCodeNodes = false,
-  });
+    ListValueNotifier<VMServiceObjectNode>? rootObjectNodesOverride,
+  }) : _rootObjectNodes = rootObjectNodesOverride ??
+            ListValueNotifier<VMServiceObjectNode>([]);
```
kenzieschmoll, line 25: Is there a better way we can do this that isn't changing the constructor for the sake of a test? For example, maybe creating a visibleForTesting getter for _rootObjectNodes that can be overridden in a TestProgramExplorerController class?

packages/devtools_app/test/debugger/program_explorer_test.dart
```diff
@@ -4,55 +4,166 @@
...
+            await initializeProgramExplorer(tester);
+        final libNode = programExplorerController.rootObjectNodes.value.first;
+
+        // No node has been selected yet, so the outline should be empty.
+        expect(programExplorerController.outlineNodes.value.isEmpty, true);
+        expect(programExplorerController.scriptSelection, isNull);
+        expect(programExplorerController.outlineSelection.value, isNull);
+
+        // Select the library node and ensure the outline is populated.
+        await programExplorerController.selectNode(libNode);
```
kenzieschmoll, line 147: since these are widget tests, we should be selecting via `tester.tap`, not via the controller. Otherwise, this is just testing the controller and not the UI

packages/devtools_app/test/test_data/vm_service_object_tree.dart
```diff
@@ -0,0 +1,82 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: put this file in a subfolder under test_data. Perhaps test_data/debugger/

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
bkonyi: You can't override a private property/method if you're not in the same library, so we'd have to make `_rootObjectNodes` public at the very least.

packages/devtools_app/test/debugger/program_explorer_test.dart
bkonyi: Ah good point. Done.

packages/devtools_app/test/test_data/vm_service_object_tree.dart
bkonyi: Done.

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
kenzieschmoll: There is already a public getter `rootObjectNodes`. This could be overriden on the TestProgramExplorer class.

kenzieschmoll (approved):

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
kenzieschmoll: Okay nevermind. I see the issue. I think making rootObjectNodesInternal public and visible for testing is a better solution than changing the constructor. lgtm.

PR #4427, opened 2022-08-31
Update to latest textmate grammar and support nested captures in parser

DanTup:
This updates the textmate grammar to the latest version from https://github.com/dart-lang/dart-syntax-highlight/blob/master/grammars/dart.json, and makes some parser changes to support some of the nested captures it includes.

Previously, the parser had a `topLevelMatchers` collection that was a set of `pattern` matchers. I've extracted that to its own `Matcher` (`_PatternMatcher`) so it can be reused, and then updated `_applyCapture` to use it if a capture being parsed uses `patterns`.

The resulting changes can be seen in the `.golden` files, which are that expressions in string interpolation are now parsed recursively so they can contain syntax highlighting within (which fixes some issues where interpolation expressions contained the same quotes that the string used because we didn't always pick the correct closing quote).

Before:
Screenshot 2022-08-31 at 18 35 01

After:
Screenshot 2022-08-31 at 18 36 23

@bkonyi


packages/devtools_app/lib/src/screens/debugger/span_parser.dart
```diff
@@ -212,44 +200,47 @@ abstract class _Matcher {
...
+      final match = scanner.substring(start, end);
+      for (int i = 0; i <= lastMatch.groupCount; ++i) {
+        // Skip if we don't have a scope or nested patterns for this capture.
+        if (!captures.containsKey(i.toString())) continue;
+
+        final captureText = lastMatch.group(i);
+        if (captureText == null || captureText.isEmpty) continue;
+
+        final startOffset = match.indexOf(captureText);
+        final capture = captures[i.toString()] as Map<String, Object?>;
```
bkonyi, line 224: Nit: We should either make `captures` a `Map<String, Object?>` or stick with `Map<String, dynamic>`.

bkonyi (approved):
Generally LGTM!

packages/devtools_app/lib/src/screens/debugger/span_parser.dart
DanTup: Any preference? I tend to prefer using `Object?` over `dynamic` because it's harder to accidentally write code that doesn't give analysis errors, but I see there's quite a lot of `dynamic`s already here (I'm happy to make them all `Object?` if that suits you).

FWIW, I don't think it will avoid the cast on this line though, as even as `Map<String, Object?>` the values are still `Object?` and we know they're `Map`s that we need to read.

bkonyi: Yeah, I know we wouldn't get rid of the cast, but then we'd at least be consistent in forcing casts and performing null checks. Before null-safety, I always used `dynamic`, but since that drops the requirement for null checks and can introduce runtime errors, I think we should probably try and use `Object?` instead where possible.

I definitely thought `captures` was a `List<Map<String, dynamic>>`, not a `Map<String, dynamic>`. In that case, can we change it to a `Map<String, Map<String, Object?>>` to be explicit?

DanTup:
Pushed a fix to update the syntax highlight tests I missed.

packages/devtools_app/lib/src/screens/debugger/span_parser.dart
DanTup: sgtm! I've also pushed a change that changes all `dynamic` to `Object?` and adds any necessary casts as a result, PTAL!

bkonyi: LGTM!

DanTup:
@bkonyi there are some failures in the dart2js bot:

```
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following assertion was thrown while running async test code:
Golden "../goldens/memory_heap_tree.png": Pixel test failed, 0.02% diff detected.
Failure feedback can be found at
/Users/runner/work/devtools/devtools/packages/devtools_app/test/memory/failures
```

Looking at the golden at https://github.com/flutter/devtools/blob/master/packages/devtools_app/test/goldens/memory_heap_android.png, it doesn't seem like this page has any syntax highlighting on it - do you know if this test has been flaky or it's something I may have caused?

DanTup:
Oh, I see the same failure on some other recent PRs, so perhaps unrelated. There was a stable release of Flutter recently, but I think DevTools in pinned to a specific version?

bkonyi:
Yeah, I think a golden needs to be updated. Your tests all seem to be passing though, so this is safe to land once someone fixes the tests.

DanTup:
Green now after merging in Kenzie's fix. Thanks!

Issue #3777, opened 2022-03-04
Network Tab is not showing any requests


Lennart-Franck:
### Specs
Flutter Version: 2.10.3
Flutter Channel: stable
DevTools Version: 2.9.2+1

### Problem
I wanted to look up the requests made by my application in the Network tab of the DevTools. The requests in my application are with Firebase and also with the dart http packet. I can understand that the DevTools may can not show the Firebase request but  the usual http requests are also not being shown. 

The network tab is only showing the loading indicator shown in the picture below. 
![image](https://user-images.githubusercontent.com/69119365/156772901-088d7806-f647-4ca7-a7fc-d41bb361a7a9.png)

kenzieschmoll:
CC @bkonyi 

Thanks for the issue. Can you please provide a small repro case so that we can reproduce this issue?

rmRizki:
This issue is also happening to me when I'm doing hot restart

tienlam79:
any update?

PrimeTimeTran:
Is the Network tab able to show requests for assets? 
My situation is when I load an asset on one screen and load the same asset on a different screen, I see different videos. I've confirmed the IDs are the same yet get different results. Strange.

https://gifyu.com/image/SKV9Q

ShafiMunshi:
Use physical device instead of emulator. It's shows all the network connection in devTools

PR #4438, opened 2022-09-02
Fix Network recordings after hot restart

CoderDake:
Fixes https://github.com/flutter/devtools/issues/3777

On hot restart, network recordings would stop working.

calling `startRecording` after the mainIsolate changes helps fix this issue.

### DIFF
--- a/packages/devtools_app/lib/src/screens/network/network_screen.dart
+++ b/packages/devtools_app/lib/src/screens/network/network_screen.dart
@@ -13,6 +13,7 @@ import '../../http/http_request_data.dart';
 import '../../primitives/auto_dispose_mixin.dart';
 import '../../primitives/utils.dart';
 import '../../shared/common_widgets.dart';
+import '../../shared/globals.dart';
 import '../../shared/screen.dart';
 import '../../shared/split.dart';
 import '../../shared/table.dart';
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/screens/network/network_screen.dart
```diff
@@ -137,6 +138,11 @@ class _NetworkScreenBodyState extends State<NetworkScreenBody>
 
   @override
   Widget build(BuildContext context) {
+    addAutoDisposeListener(serviceManager.isolateManager.mainIsolate, () {
```
kenzieschmoll, line 141: we should never be adding listeners or doing business logic like this in `build`. This would add a listener each time the widget is rebuilt, which is not what we want.

we should do this in didChangeDependencies after we make the initial call to controller.startRecording.

CoderDake: After discussing this with Kenzie I've added this to the `didChangeDependencies` as well as added a call to `cancelListeners` to avoid adding duplicates :D

kenzieschmoll (approved):

Gantushig77:
Thank you! 🫡

PR #4447, opened 2022-09-07
Add "Code Preview" section to Object Inspector views

bkonyi:
Adds support for displaying code associated with a given VM service object. Dart code for only the given object will be displayed (e.g., a subset of the containing script's source for functions, fields, and classes), but will display the original range of line numbers.

In addition, users can use the "Code Preview" view to view and set breakpoints on executable lines, just as if they were interacting with the code view in the debugger.

image

### DIFF
--- a/packages/devtools_app/lib/src/primitives/utils.dart
+++ b/packages/devtools_app/lib/src/primitives/utils.dart
@@ -801,6 +801,7 @@ class LineRange {
   @override
   int get hashCode => Object.hash(begin, end);
 }
+
 extension SortDirectionExtension on SortDirection {
   SortDirection reverse() {
     return this == SortDirection.ascending
--- a/packages/devtools_app/lib/src/primitives/utils.dart
+++ b/packages/devtools_app/lib/src/primitives/utils.dart
@@ -778,19 +778,29 @@ enum SortDirection {
   descending,
 }
 
-class LineRange extends Range {
-  const LineRange(super.begin, super.end);
+/// A Range-like class that works for inclusive ranges of lines in source code.
+class LineRange {
+  const LineRange(this.begin, this.end) : assert(begin <= end);
 
-  @override
-  int get begin => super.begin.toInt();
+  final int begin;
+  final int end;
 
-  @override
-  int get end => super.end.toInt();
-
-  @override
   int get size => end - begin + 1;
-}
 
+  bool contains(num target) => target >= begin && target <= end;
+
+  @override
+  String toString() => 'LineRange($begin, $end)';
+
+  @override
+  bool operator ==(other) {
+    if (other is! LineRange) return false;
+    return begin == other.begin && end == other.end;
+  }
+
+  @override
+  int get hashCode => Object.hash(begin, end);
+}
 extension SortDirectionExtension on SortDirection {
   SortDirection reverse() {
     return this == SortDirection.ascending
--- a/packages/devtools_app/lib/src/primitives/utils.dart
+++ b/packages/devtools_app/lib/src/primitives/utils.dart
@@ -778,6 +778,19 @@ enum SortDirection {
   descending,
 }
 
+class LineRange extends Range {
+  const LineRange(super.begin, super.end);
+
+  @override
+  int get begin => super.begin.toInt();
+
+  @override
+  int get end => super.end.toInt();
+
+  @override
+  int get size => end - begin + 1;
+}
+
 extension SortDirectionExtension on SortDirection {
   SortDirection reverse() {
     return this == SortDirection.ascending
--- a/packages/devtools_app/lib/src/screens/debugger/breakpoint_manager.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/breakpoint_manager.dart
@@ -16,11 +16,9 @@ class BreakpointManager extends Disposer {
 
   final bool initialSwitchToIsolate;
 
-  VmServiceWrapper get _service {
-    return serviceManager.service!;
-  }
+  VmServiceWrapper get _service => serviceManager.service!;
 
-  final Map<String, List<SourcePosition>> _breakPositionsMap = {};
+  final _breakPositionsMap = <String, List<SourcePosition>>{};
 
   final _breakpoints = ValueNotifier<List<Breakpoint>>([]);
 
--- a/packages/devtools_app/lib/src/screens/debugger/breakpoint_manager.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/breakpoint_manager.dart
@@ -12,7 +12,6 @@ import '../../shared/globals.dart';
 import 'debugger_model.dart';
 
 class BreakpointManager extends Disposer {
-
   BreakpointManager({this.initialSwitchToIsolate = true});
 
   final bool initialSwitchToIsolate;
--- a/packages/devtools_app/lib/src/screens/debugger/breakpoint_manager.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/breakpoint_manager.dart
@@ -1,0 +1,292 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:collection/collection.dart';
+import 'package:flutter/foundation.dart';
+import 'package:vm_service/vm_service.dart';
+
+import '../../primitives/auto_dispose.dart';
+import '../../service/vm_service_wrapper.dart';
+import '../../shared/globals.dart';
+import 'debugger_model.dart';
+
+class BreakpointManager extends Disposer {
+
+  BreakpointManager({this.initialSwitchToIsolate = true});
+
+  final bool initialSwitchToIsolate;
+
+  VmServiceWrapper get _service {
+    return serviceManager.service!;
+  }
+
+  final Map<String, List<SourcePosition>> _breakPositionsMap = {};
+
+  final _breakpoints = ValueNotifier<List<Breakpoint>>([]);
+
+  ValueListenable<List<Breakpoint>> get breakpoints => _breakpoints;
+
+  final _breakpointsWithLocation =
+      ValueNotifier<List<BreakpointAndSourcePosition>>([]);
+
+  ValueListenable<List<BreakpointAndSourcePosition>>
+      get breakpointsWithLocation => _breakpointsWithLocation;
+
+  IsolateRef? _isolateRef;
+
+  String get _isolateRefId => _isolateRef?.id ?? '';
+
+  void initialize() {
+    final isolate = serviceManager.isolateManager.selectedIsolate.value;
+    if (initialSwitchToIsolate && isolate != null) {
+      switchToIsolate(serviceManager.isolateManager.selectedIsolate.value);
+    }
+
+    addAutoDisposeListener(serviceManager.isolateManager.selectedIsolate, () {
+      switchToIsolate(serviceManager.isolateManager.selectedIsolate.value);
+    });
+    autoDisposeStreamSubscription(
+      _service.onDebugEvent.listen(_handleDebugEvent),
+    );
+    autoDisposeStreamSubscription(
+      _service.onIsolateEvent.listen(_handleIsolateEvent),
+    );
+  }
+
+  void switchToIsolate(IsolateRef? isolateRef) async {
+    _isolateRef = isolateRef;
+
+    if (isolateRef == null) {
+      _breakpoints.value.clear();
+      _breakpointsWithLocation.value.clear();
+      return;
+    }
+
+    final isolate = await _service.getIsolate(_isolateRefId);
+    if (isolate.id != _isolateRefId) {
+      // Current request is obsolete.
+      return;
+    }
+
+    _breakpoints.value = isolate.breakpoints ?? [];
+
+    // Build _breakpointsWithLocation from _breakpoints.
+    // ignore: unawaited_futures
+    Future.wait(
+      _breakpoints.value.map(breakpointManager.createBreakpointWithLocation),
+    ).then((list) {
+      if (isolate.id != _isolateRefId) {
+        // Current request is obsolete.
+        return;
+      }
+      _breakpointsWithLocation.value = list.toList()..sort();
+    });
+  }
+
+  void clearCache() {
+    _breakPositionsMap.clear();
+    _breakpoints.value = [];
+    _breakpointsWithLocation.value = [];
+  }
+
+  Future<void> clearBreakpoints() async {
+    final breakpoints = _breakpoints.value.toList();
+    await Future.forEach(breakpoints, (Breakpoint breakpoint) {
+      return removeBreakpoint(breakpoint);
+    });
+  }
+
+  Future<Breakpoint> addBreakpoint(String scriptId, int line) =>
+      _service.addBreakpoint(_isolateRefId, scriptId, line);
+
+  Future<void> removeBreakpoint(Breakpoint breakpoint) =>
+      _service.removeBreakpoint(_isolateRefId, breakpoint.id!);
+
+  Future<void> toggleBreakpoint(ScriptRef script, int line) async {
+    if (serviceManager.isolateManager.selectedIsolate.value == null) {
+      // Can't toggle breakpoints if we don't have an isolate.
+      return;
+    }
+    // The VM doesn't support debugging for system isolates and will crash on
+    // a failed assert in debug mode. Disable the toggle breakpoint
+    // functionality for system isolates.
+    if (serviceManager.isolateManager.selectedIsolate.value!.isSystemIsolate!) {
+      return;
+    }
+
+    final bp = breakpointsWithLocation.value.firstWhereOrNull((bp) {
+      return bp.scriptRef == script && bp.line == line;
+    });
+
+    if (bp != null) {
+      await removeBreakpoint(bp.breakpoint);
+    } else {
+      try {
+        await addBreakpoint(script.id!, line);
+      } catch (_) {
+        // ignore errors setting breakpoints
+      }
+    }
+  }
+
+  void _updateAfterIsolateReload(
+    Event reloadEvent,
+  ) async {
+    // TODO(devoncarew): We need to coordinate this with other debugger clients
+    // as well as pause before re-setting the breakpoints.
+    // Refresh the list of scripts.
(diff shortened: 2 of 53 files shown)


bkonyi:
This change will still need some tests, but can probably go through an initial review pass.

jacob314:
minor nit: code preview should show the file name for added context on where the code snippet is from.

packages/devtools_app/lib/src/primitives/utils.dart
```diff
@@ -778,6 +778,19 @@ enum SortDirection {
...
+  const LineRange(super.begin, super.end);
+
+  @override
+  int get begin => super.begin.toInt();
+
+  @override
+  int get end => super.end.toInt();
+
+  @override
+  int get size => end - begin + 1;
```
jacob314, line 791: why is this Range now inclusive instead of being exclusive like I assume the normal Range class is?

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -295,123 +311,133 @@ class _CodeViewState extends State<CodeView>
...
+            onSelected: (scriptRef) {
+              widget.controller.showScriptLocation(ScriptLocation(scriptRef));
+            },
+            enabled: widget.controller.scriptsHistory.hasScripts,
+          ),
+        ],
+        contentBuilder: contentBuilder,
+      );
+    }
+    // TODO(bkonyi): this is a hack to deal with some Expanded elements
```
jacob314, line 436: @kenzieschmoll is there an easier way to do this?

bkonyi: I totally forgot I had this here. I still need to investigate this before this is ready to actually land.

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view_controller.dart
```diff
@@ -31,6 +32,8 @@ class ObjectInspectorViewController extends DisposableController
   final programExplorerController =
       ProgramExplorerController(showCodeNodes: true);
 
+  final debuggerController = DebuggerController();
```
jacob314, line 35: shouldn't we reuse the same debuggerController as the debugger page using the provider? Otherwise I imagine things like setting and getting breakpoints could be strangely inconsistent.

packages/devtools_app/lib/src/primitives/utils.dart
bkonyi: Basically for the `lineCount` argument to `Gutter`:

```dart
lineCount: widget.lineRange?.size ?? lines.length,
```

Maybe we should just make `LineRange` completely disjoint from `Range`?

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -731,3 +734,89 @@ List<MapEntry<String, WidgetBuilder>> vmObjectGeneralDataRows(
...
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder<ParsedScript?>(
+      valueListenable: widget.debuggerController.currentParsedScript,
+      builder: (context, currentParsedScript, _) {
+        LineRange? lineRange;
+        if (currentParsedScript != null) {
+          final obj = widget.object;
+          SourceLocation? location;
+          if (obj is ClassRef || obj is FuncRef || obj is FieldRef) {
+            final dynObj = obj as dynamic;
```
jacob314, line 778: avoid `as dynamic`

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view_controller.dart
bkonyi: Nope, because then updating the current script in the debugger on either screen will result in a change on the other. Setting and getting breakpoints seems to work fine since the controller is wired in such a way that it will react to new breakpoints being added / removed by other clients. Seemed to work fine with manual testing.

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
bkonyi: Yeah, I typically do, but the alternative isn't great either:

```dart
if (obj is ClassRef) {
  location = object.location;
} else if (obj is FuncRef) {
  location = object.location;
} ...
```

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -249,7 +262,10 @@ class _CodeViewState extends State<CodeView>
     final scriptSource = parsedScript?.script.source;
     if (script != null && scriptSource != null) {
       if (scriptSource.length < 500000) {
-        final highlighted = script.highlighter.highlight(context);
+        final highlighted = script.highlighter.highlight(
+          context,
+          lineRange: widget.lineRange,
```
jacob314, line 267: Maybe add a TODO to switch from having lineRange to highlightedLineRange where text outside the highlighted line range has a greyed background and selection is auto scrolled to the highlighted range. that would be great for the inspector use case. Note we already scroll to lines on hitting a breakpoint which could interact with this in tricky ways.

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view_controller.dart
bkonyi: Ideally, we'd split `DebuggerController` into two components: `DebuggerController` and `CodeViewController`, where `CodeViewController` is responsible for responsible for script state and jumping to locations in the script. That way, `DebuggerController` would only be responsible for debugger state and could be reused across the application.

jacob314: Yeah please split into a separate CodeViewController.  That should be pretty easy.
FWIW, I'd be ok with landing even if you cause the script to change on the other page. That is actually a bit of a feature as well as a bug. E.g. I was viewing this file in the program explorer and now I see it in the debugger page... convenient.

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
jacob314: even though that has duplicate code I think it is better and may even trigger lints to leverage patterns in the future.

packages/devtools_app/lib/src/primitives/utils.dart
jacob314: Making it disjoint sounds fine to me. I would also consider making the end value just be exclusive rather than inclusive as that is the default expectation for ranges.

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view_controller.dart
bkonyi: Eh, I'm not convinced that changing the script in a separate tab as a result of operations performed in another is good UX unless we have an explicit way to jump to the other tab displaying the updated script. These kind of surprises will cause more confusion than anything.

bkonyi:
I went to break out the code view related logic from `DebuggerController`, and ended up shaving the yak... I pulled all the breakpoint logic into a global `BreakpointManager` so the code view can set breakpoints without requiring a full `DebuggerController`, which meant I had to update a bunch of tests as well.

This is a big PR now, so I can try and split this changes out from the "Code Preview" change if you'd like.

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view_controller.dart
jacob314: ok fine to leave as a todo and add a CodeViewController later

bkonyi: ... too late :-)

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -108,31 +124,17 @@ class _CodeViewState extends State<CodeView>
     }
 
     addAutoDisposeListener(
-      widget.controller.scriptLocation,
+      widget.codeViewController.scriptLocation,
       _handleScriptLocationChanged,
     );
   }
 
-  @override
```
jacob314, line 116: why did didUpdateWidget go away?

packages/devtools_app/lib/src/screens/debugger/codeview_controller.dart
```diff
@@ -0,0 +1,266 @@
...
+import '../../config_specific/logger/logger.dart';
+import '../../primitives/auto_dispose.dart';
+import '../../primitives/history_manager.dart';
+import '../../shared/globals.dart';
+import '../../ui/search.dart';
+import 'debugger_model.dart';
+import 'program_explorer_controller.dart';
+import 'syntax_highlighter.dart';
+
+class CodeViewController extends DisposableController
```
jacob314, line 17: Thank you so much for shaving the yack and refactoring this.

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view_controller.dart
```diff
@@ -47,6 +50,8 @@ class ObjectInspectorViewController extends DisposableController
       programExplorerController
         ..initialize()
         ..initListeners();
+      // TODO(bkonyi):
+//      codeViewController.initialize();
```
jacob314, line 54: why is this commented out?

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -20,32 +20,104 @@ import '../../shared/dialogs.dart';
...
+          );
+        } else {
+          return child!;
+        }
+      },
+      child: DualValueListenableBuilder<ScriptRef?, ParsedScript?>(
+        firstListenable: controller.currentScriptRef,
+        secondListenable: controller.currentParsedScript,
+        builder: (context, scriptRef, parsedScript, _) {
+          // TODO(bkonyi)
```
jacob314: why is this commented out?

jacob314 (commented):
Ok reviewed everything. A few minor comments and then good to land.

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view_controller.dart
bkonyi: Just old code, `CodeViewController` doesn't have an `initialize()` method.

packages/devtools_app/lib/src/screens/debugger/codeview.dart
bkonyi: I had moved it at some point, but then changed my mind. Guess I forgot to add it back.

packages/devtools_app/lib/src/screens/debugger/codeview.dart
bkonyi: Wasn't sure where it should live, but I've decided against creating a `CodeArea` widget just for the debugger screen, so this just goes back to where it was originally.

jacob314:
lgtm_dog2

jacob314 (approved):

packages/devtools_app/lib/src/primitives/utils.dart
```diff
@@ -778,6 +778,30 @@ enum SortDirection {
   descending,
 }
 
+/// A Range-like class that works for inclusive ranges of lines in source code.
+class LineRange {
```
kenzieschmoll, line 782: can we just modify the existing Range class to support this?

e.g. add a `bool inclusive` field and create a constructor `Range.inclusive`

packages/devtools_app/lib/src/screens/debugger/breakpoint_manager.dart
```diff
@@ -0,0 +1,291 @@
...
+import 'debugger_model.dart';
+
+class BreakpointManager extends Disposer {
+  BreakpointManager({this.initialSwitchToIsolate = true});
+
+  final bool initialSwitchToIsolate;
+
+  VmServiceWrapper get _service {
+    return serviceManager.service!;
+  }
```
kenzieschmoll, line 21: nit: can we use => since it's a one-liner

packages/devtools_app/lib/src/screens/debugger/breakpoint_manager.dart
```diff
@@ -0,0 +1,291 @@
...
+
+  final Map<String, List<SourcePosition>> _breakPositionsMap = {};
```
kenzieschmoll, line 23: nit: move type to RHS

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -295,127 +318,137 @@ class _CodeViewState extends State<CodeView>
...
+                  children: [
+                    ValueListenableBuilder<List<BreakpointAndSourcePosition>>(
+                      valueListenable:
+                          breakpointManager.breakpointsWithLocation,
+                      builder: (context, breakpoints, _) {
+                        return Gutter(
+                          gutterWidth: gutterWidth,
+                          scrollController: gutterController,
+                          lineCount: widget.lineRange?.size ?? lines.length,
+                          lineOffset: (widget.lineRange?.begin ?? 1) - 1,
```
kenzieschmoll, line 353: what if widget.lineRange.begin = 0?

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -450,10 +484,11 @@ class _CodeViewState extends State<CodeView>
     return Center(
       child: ElevatedButton(
         autofocus: true,
-        onPressed: () => widget.controller.toggleFileOpenerVisibility(true),
+        onPressed: () =>
+            widget.codeViewController.toggleFileOpenerVisibility(true),
         child: Text(
           'Open a file ($openFileKeySetDescription)',
-          style: theme.textTheme.subtitle1,
+          style: theme.textTheme.titleMedium,
```
kenzieschmoll, line 491: you're probably on a newer version of flutter locally than our bots run on. I'd checkout flutter  3.3.0-0.5.pre and see if this is still valid

packages/devtools_app/lib/src/screens/debugger/codeview_controller.dart
```diff
@@ -0,0 +1,266 @@
...
+    with AutoDisposeControllerMixin, SearchControllerMixin<SourceToken> {
+  CodeViewController() {
+    _scriptHistoryListener = () {
+      final currentScriptValue = scriptsHistory.current.value;
+      if (currentScriptValue != null)
+        _showScriptLocation(ScriptLocation(currentScriptValue));
+    };
+    scriptsHistory.current.addListener(_scriptHistoryListener);
```
kenzieschmoll, line 25: should we be removing this listener in dispose?

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -31,25 +27,20 @@ final _log = DebugTimingLogger('debugger', mute: true);
...
     }
-    _scriptHistoryListener = () {
-      final currentScriptValue = scriptsHistory.current.value;
-      if (currentScriptValue != null)
-        _showScriptLocation(ScriptLocation(currentScriptValue));
-    };
-    scriptsHistory.current.addListener(_scriptHistoryListener);
   }
 
+  final debuggerCodeViewController = CodeViewController();
```
kenzieschmoll, line 42: nit: maybe just CodeViewController? DebuggerController.debuggerCodeViewController is a bit redundant

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -796,3 +800,102 @@ List<MapEntry<String, WidgetBuilder>> vmObjectGeneralDataRows(
       ),
   ];
 }
+
+/// Creates a simple [CodeView] which displays the code relevant to [object] in
+/// [script]. If [object] is synthetic and doesn't have actual token positions,
```
kenzieschmoll, line 805: dart doc nit: new line after first sentence

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -796,3 +800,102 @@ List<MapEntry<String, WidgetBuilder>> vmObjectGeneralDataRows(
...
+
+class _ObjectInspectorCodeViewState extends State<ObjectInspectorCodeView> {
+  @override
+  void didChangeDependencies() async {
+    super.didChangeDependencies();
+    if (widget.script != widget.codeViewController.currentScriptRef.value) {
+      widget.codeViewController.resetScriptLocation(
+        ScriptLocation(widget.script),
+      );
+    }
```
kenzieschmoll, line 831: should this be in didUpdateWidget as well?

packages/devtools_app/lib/src/screens/vm_developer/vm_library_display.dart
```diff
@@ -14,22 +16,35 @@ import 'vm_object_model.dart';
...
+                LibraryDependencies(dependencies: dependencies)
+            ],
+          ),
+        ),
+        ObjectInspectorCodeView(
+          codeViewController: controller.codeViewController,
+          script: library.scriptRef!,
+          object: library.obj,
+        ),
       ],
```
kenzieschmoll, line 48: this code seems to be duplicated for each object type. can we create a helper widget that creates the Split and the ObjectInspectorCodeView and takes a parameter for the first child in the split widget?

kenzieschmoll (commented):
Can we add / update any of the existing VM developer screen tests for the new functionality in this PR?

packages/devtools_app/lib/src/screens/debugger/codeview.dart
bkonyi: This shouldn't happen as line numbers always start at 1.

packages/devtools_app/lib/src/screens/debugger/codeview.dart
bkonyi: Ah yeah, good point. I saw the deprecation, figured I'd fix it without thinking.

bkonyi:
> Can we add / update any of the existing VM developer screen tests for the new functionality in this PR?

I was planning on adding proper testing in a follow up PR since this one is already huge.

PR #4448, opened 2022-09-07
Fix `CodeView`'s scrollable area

bkonyi:
Currently, the `CodeView`'s scrollable area is only as wide as the longest line in the currently selected script, resulting in arguably unexpected behavior when trying to scroll while hovering over the right side of the code area.

This change makes each line in the `CodeView` take up the maximum amount of horizontal space possible, making the scrollable area consist of the entire code viewport.

### DIFF
--- a/packages/devtools_app/lib/src/screens/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/codeview.dart
@@ -362,10 +362,6 @@ class _CodeViewState extends State<CodeView>
                         Expanded(
                           child: LayoutBuilder(
                             builder: (context, constraints) {
-                              final double fileWidth = calculateTextSpanWidth(
-                                findLongestTextSpan(lines),
-                              );
-
                               return Scrollbar(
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -376,7 +372,7 @@ class _CodeViewState extends State<CodeView>
                                   controller: horizontalController,
                                   child: SizedBox(
                                     height: constraints.maxHeight,
-                                    width: fileWidth,
+                                    width: constraints.maxWidth,
```
kenzieschmoll, line 375: I think this should be the max of fileWidth and constraints.maxWidth so that we cover both the case where the file is wider than the view and more narrow than the view

bkonyi: Ah yes, that makes more sense.

kenzieschmoll (approved):

PR #4463, opened 2022-09-11
Adopt stager for DevTools.

polina-c:


.gitignore
```diff
@@ -26,3 +26,4 @@ Podfile.lock
 **/generated/not_namespaced_r_class_sources/
 **/build/
 packages/devtools_test/lib/src/mocks/generated.mocks.dart
+**/*.stager_app.dart
```
kenzieschmoll, line 29: is there a reason we don't want this checked in?

polina-c: they are generated
but we can check them in...
It may be convenient
What do you think?

kenzieschmoll: if we don't check them in, what does it look like for someone who doesn't have the generated files on their machine? should we add generating these stager_app files to the generate code scripts we have?

polina-c: I thought it is already here, but just realized it is for other package. Added.

kenzieschmoll (approved):

PR #4471, opened 2022-09-13
Use a broadcast stream controller for the enhance tracing button

kenzieschmoll:
This fixes a bug where opening the Performance page, opening a different page, and then re-opening a performance page would throw an exception for the StreamController's stream being listened to twice. Using a broadcast stream allows us to add multiple listeners.

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/panes/controls/enhance_tracing/enhance_tracing_controller.dart
+++ b/packages/devtools_app/lib/src/screens/performance/panes/controls/enhance_tracing/enhance_tracing_controller.dart
@@ -18,7 +18,7 @@ final enhanceTracingExtensions = [
 
 class EnhanceTracingController extends DisposableController
     with AutoDisposeControllerMixin {
-  final showMenuStreamController = StreamController<void>();
+  final showMenuStreamController = StreamController<void>.broadcast();
 
   late EnhanceTracingState tracingState;
 


packages/devtools_app/lib/src/screens/performance/panes/controls/enhance_tracing/enhance_tracing_controller.dart
```diff
@@ -18,7 +18,7 @@ final enhanceTracingExtensions = [
 
 class EnhanceTracingController extends DisposableController
     with AutoDisposeControllerMixin {
-  final showMenuStreamController = StreamController<void>();
+  final showMenuStreamController = StreamController<void>.broadcast();
```
jacob314, line 21: any reason we are using streams here instead of ValueListener and friends?

jacob314 (approved):
lgtm

packages/devtools_app/lib/src/screens/performance/panes/controls/enhance_tracing/enhance_tracing_controller.dart
kenzieschmoll: In this instance the stream controller is used to dispatch an event that another stateful widget is listening to (ServiceExtensionCheckboxGroupButton), and upon receiving an event, the listener inserts an overlay. There isn't really any 'value' being passed around, so it doesn't fit the ValueNotifier / ValueListenable pattern very well. I suppose we could pass around a bool value, but that also doesn't exactly fit this use case since we are inserting an overlay and not just changing some part of what we build.

PR #4473, opened 2022-09-14
First chunk on snapshots visualization.

polina-c:


packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -0,0 +1,73 @@
...
+import '../../../../../primitives/utils.dart';
+import '../../../../../shared/common_widgets.dart';
+import '../../../primitives/memory_utils.dart';
+import '../../../shared/heap/model.dart';
+import 'model.dart';
+
+class DiffPaneController {
+  DiffPaneController(this.snapshotTaker);
+
+  final listScrollController = ScrollController();
```
kenzieschmoll, line 18: scroll controllers should be created in widget lifecycle methods, preferably initState

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -2,119 +2,69 @@
...
-    snapshots.removeRange(selectedIndex.value, selectedIndex.value + 1);
-    selectedIndex.value = selectedIndex.value - 1;
-  }
-}
+bool shouldShowDiffPane = true;
 
 class DiffPane extends StatefulWidget {
-  const DiffPane({Key? key}) : super(key: key);
+  const DiffPane({Key? key, required this.snapshotTaker}) : super(key: key);
+  final SnapshotTaker snapshotTaker;
```
kenzieschmoll, line 24: nit: new line between constructor and class members

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_control.dart
```diff
@@ -0,0 +1,31 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder<bool>(
+      valueListenable: controller.isProcessing,
+      builder: (_, isProcessing, __) => Row(
+        children: [
+          ToolbarAction(
+            icon: Icons.clear,
+            tooltip: 'Delete snapshot',
```
kenzieschmoll, line 24: instead of having this delete control on the snapshot itself, what about having a little trash can icon that appears over a snapshot list item when hovered? This would be make it more clear that the delete action actually takes a snapshot out of the list completely, and doesn't just close the table view.

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -6,59 +6,69 @@ import 'package:flutter/material.dart';
...
+import 'widgets/snapshot_control.dart';
 import 'widgets/snapshot_list.dart';
 import 'widgets/snapshot_view.dart';
 
 /// While this pane is under construction, we do not want our users to see it.
 ///
 /// Flip this flag locally to test the pane and flip back before checking in.
 /// TODO: before removing this flag add widget/golden testing for the diff pane.
-bool shouldShowDiffPane = false;
+bool shouldShowDiffPane = true;
```
polina-c, line 20: Will not forget to turn it off, because it is covered by test.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
polina-c: The button that invokes the method to add snapshot lives in _ListControlPane, while the list that should be scrolled lives in _SnapshotListItems. Both are stateless. 
Should I convert list widget to stateful and add notifier to the controller, to listen for signal to scroll to bottom?

Sounds as unnecessary complexity.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_control.dart
polina-c: Yes, it would be great and I tried this path. I did not find a simple way to highlight 'x' on hover.
I would leave it to fix-it-friday.

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
```diff
@@ -254,6 +256,8 @@ class MemoryController extends DisposableController
     });
   }
 
+  late SnapshotTaker snapshotTaker;
```
kenzieschmoll, line 259: add dart doc

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -9,10 +9,15 @@ import 'package:flutter/material.dart';
...
 import '../../../../../shared/common_widgets.dart';
 import '../../../primitives/memory_utils.dart';
+import '../../../shared/heap/model.dart';
 import 'model.dart';
 
 class DiffPaneController {
-  final scrollController = ScrollController();
+  DiffPaneController(this.snapshotTaker);
+
+  final listScrollController = ScrollController();
```
kenzieschmoll, line 18: see my comments on https://github.com/flutter/devtools/pull/4475 about where this should be created and used.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -61,6 +66,8 @@ class DiffPaneController {
   void deleteCurrentSnapshot() {
     assert(selected is SnapshotListItem);
     snapshots.removeRange(selectedIndex.value, selectedIndex.value + 1);
+    // We must change the selectedIndex, because otherwise the content will
+    // not be re-rendered.
     selectedIndex.value = selectedIndex.value - 1;
```
kenzieschmoll, line 71: this is not safe as is. what if 0 is the selected index? I know right now this is the "information" item, but if that changes and we choose to show instructions / info in a different way, we will have broken behavior. Use `math.max(selectedIndex.value - 1, 0)` instead.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_list.dart
```diff
@@ -39,19 +39,19 @@ class _ListControlPane extends StatelessWidget {
     return ValueListenableBuilder<bool>(
       valueListenable: controller.isProcessing,
       builder: (_, isProcessing, __) {
-        final showTakeSnapshot = !isProcessing;
-        final showClearAll = !isProcessing & controller.hasSnapshots;
+        final takeSnapshotEnabled = !isProcessing;
+        final clearAllEnabled = !isProcessing & controller.hasSnapshots;
```
kenzieschmoll, line 43: && instead of & ?

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_view.dart
```diff
@@ -13,6 +18,173 @@ class SnapshotView extends StatelessWidget {
...
+
+        final stats = item.stats;
+        if (stats == null) {
+          return const Center(child: Text('Could not take snapshot.'));
+        }
+
+        return _StatsTable(
+          data: stats,
+          sorting: item.sorting,
+          key: ObjectKey(item),
```
kenzieschmoll, line 34: nit: order key first. Can you also add a comment explaining why we need a key here?

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_view.dart
```diff
@@ -13,6 +18,173 @@ class SnapshotView extends StatelessWidget {
...
+  @override
+  String getTooltip(HeapStatsRecord record) => record.fullClassName;
+}
+
+class _InstanceColumn extends ColumnData<HeapStatsRecord> {
+  _InstanceColumn()
+      : super(
+          'Non GC-able\nInstances',
+          titleTooltip: 'Number of instances of the class, '
+              'that have retaining path from the root.',
```
kenzieschmoll, line 65: nit: that have _a_ retaining path

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_view.dart
```diff
@@ -13,6 +18,173 @@ class SnapshotView extends StatelessWidget {
...
+
+  @override
+  bool get numeric => true;
+}
+
+class _ShallowSizeColumn extends ColumnData<HeapStatsRecord> {
+  _ShallowSizeColumn()
+      : super(
+          'Shallow\nSize',
+          titleTooltip: 'Total shallow Dart (not native) size of objects.',
```
kenzieschmoll, line 84: could be good to also explain what shallow means in this tooltip

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_view.dart
```diff
@@ -13,6 +18,173 @@ class SnapshotView extends StatelessWidget {
...
+        kbFractionDigits: 1,
+      )!;
+}
+
+class _RetainedSizeColumn extends ColumnData<HeapStatsRecord> {
+  _RetainedSizeColumn()
+      : super(
+          'Retained\nSize',
+          titleTooltip: 'Total size of objects plus objects they retain, '
+              'taking to account only the shortest retaining path for the referenced objects.',
```
kenzieschmoll, line 111: nit: taking _into_ account

does this size exclude native as well? if so, we should mention that here to match what is done for shallow size

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_view.dart
```diff
@@ -13,6 +18,173 @@ class SnapshotView extends StatelessWidget {
...
+}
+
+class _StatsTable extends StatefulWidget {
+  const _StatsTable({
+    Key? key,
+    required this.data,
+    required this.sorting,
+  }) : super(key: key);
+  final List<HeapStatsRecord> data;
+  final ColumnSorting sorting;
```
kenzieschmoll, line 140: new lines between

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_view.dart
```diff
@@ -13,6 +18,173 @@ class SnapshotView extends StatelessWidget {
...
+  @override
+  State<_StatsTable> createState() => _StatsTableState();
+}
+
+class _StatsTableState extends State<_StatsTable> {
+  late final List<ColumnData<HeapStatsRecord>> _columns;
+
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
```
kenzieschmoll, line 151: this can all be done in initState. We only need to use didChangeDependencies instead of initState when we need to be able to access `context`

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_view.dart
```diff
@@ -13,6 +18,173 @@ class SnapshotView extends StatelessWidget {
...
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return FlatTable<HeapStatsRecord>(
+      columns: _columns,
+      data: widget.data,
+      keyFactory: (e) => Key(e.fullClassName),
+      onItemSelected: (r) {},
```
kenzieschmoll, line 176: should we at least be highlighting the selected row onPressed?

packages/devtools_app/lib/src/screens/memory/shared/heap/heap_analyzer.dart
```diff
@@ -2,23 +2,53 @@
...
+  assert(_verifyHeapIntegrity(heap));
+}
+
+List<HeapStatsRecord> heapStats(AdaptedHeap? heap) {
+  final result = <String, HeapStatsRecord>{};
+  if (heap == null) return [];
+  if (!heap.isSpanningTreeBuilt) buildSpanningTree(heap);
+  for (var object in heap.objects) {
+    // We do not show objects that will be garbage collected soon.
+    if (object.retainedSize == null || object.isSentinel) continue;
```
kenzieschmoll, line 24: how accurate is this? should we be trying to force a GC before we take a snapshot to make sure?

packages/devtools_app/lib/src/screens/memory/shared/heap/heap_analyzer.dart
```diff
@@ -64,5 +117,29 @@ bool _canRetain(String klass, String library) {
...
+      (object.retainedSize == null) == (object.retainer == null),
+      'retainedSize = ${object.retainedSize}, retainer = ${object.retainer}',
+    );
+    if (object.retainer != null) totalReachableSize += object.shallowSize;
+  }
+
+  assert(
+    heap.root.retainedSize == totalReachableSize,
+    '${heap.root.retainedSize} not equal to $totalReachableSize',
+  );
```
kenzieschmoll, line 143: if this should all only be done in debug mode, just wrap the entire method body in an assert

packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
```diff
@@ -109,35 +124,64 @@ class AdaptedHeapObject {
...
+  int retainedSize = 0;
+  int instanceCount = 0;
+
+  String get fullClassName => _fullClassName(library, className);
+}
+
+String _fullClassName(String library, String className) =>
+    library.isNotEmpty ? '$library/$className' : className;
+
+class SnapshotTaker {
```
kenzieschmoll, line 181: can you add dart doc to this class to explain why it is necessary

packages/devtools_app/test/memory/diff/diff_pane_test.dart
```diff
@@ -2,11 +2,86 @@
...
+        expect(find.text('main-${i + 1}'), findsOneWidget);
+      }
+      await expectLater(
+        finder,
+        matchesDevToolsGolden('../../goldens/memory_diff_three_snapshots.png'),
+      );
+
+      // Delete and take a snapshot.
+      await tester.tap(find.byTooltip('Delete snapshot'));
+      await tester.pumpAndSettle();
```
kenzieschmoll, line 70: at this point can we check that a snapshot has been successfully deleted?

packages/devtools_app/test/scenes/memory/offline.dart
```diff
@@ -77,3 +80,63 @@ class MemoryOfflineScene extends Scene {
...
+  @override
+  Future<AdaptedHeap?> take() async {
+    // This delay is needed for UI to start showing the progress indicator.
+    await Future.delayed(const Duration(milliseconds: 100));
+
+    // Return null if it is the first time.
+    if (firstTime) {
+      firstTime = false;
+      return null;
+    }
```
kenzieschmoll, line 98: why do we want to return null for first time? maybe add comment

packages/devtools_app/test/scenes/memory/offline.dart
```diff
@@ -4,6 +4,7 @@ import 'package:devtools_app/src/screens/memory/memory_controller.dart';
 import 'package:devtools_app/src/screens/memory/memory_heap_tree_view.dart';
```
kenzieschmoll, line 4: not from this PR, but can we rename this file to be more reflective of what this file contains? at this point it contains a Scene and helpers. Maybe scene.dart, default_scene.dart, common.dart, shared.dart? etc.

packages/devtools_app/lib/src/screens/memory/shared/heap/heap_analyzer.dart
polina-c: It is accurate (if there is no bugs in gc and snapshot). Even if we force GC, it will be different command (take profile) and new GCable objects can show up. 
Forcing GC affects application, that I would avoid if possible, and it is possible in this case.

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
```diff
@@ -256,8 +258,7 @@ class MemoryController extends DisposableController
     });
   }
 
-  /// Memory snapshot taker, that is convinient to mock for testing.
-  late SnapshotTaker snapshotTaker;
+  late DiffPaneController diffPaneController;
```
kenzieschmoll, line 261: make this late final and add a comment to explain why this is late (I think this is for testing convenience right?)

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -6,50 +6,35 @@ import 'package:flutter/material.dart';
...
+  const DiffPane({Key? key, required this.controller}) : super(key: key);
 
-  final SnapshotTaker snapshotTaker;
+  final DiffPaneController controller;
 
   @override
   State<DiffPane> createState() => _DiffPaneState();
 }
 
 class _DiffPaneState extends State<DiffPane> {
```
kenzieschmoll, line 24: looks like this widget can become stateless now

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_list.dart
```diff
@@ -128,16 +128,19 @@ class _SnapshotListItemsState extends State<_SnapshotListItems>
...
+    addAutoDisposeListener(widget.controller.selectedIndex, () async {
+      setState(() {});
+
       final newSnapshotsLength = widget.controller.snapshots.value.length;
       if (newSnapshotsLength > _currentSnapshotsLength) {
-        // If new snapshot added, scroll to bottom.
-        setState(() => _scrollController.autoScrollToBottom());
+        _currentSnapshotsLength = newSnapshotsLength;
+        await _scrollController.autoScrollToBottom();
+        setState(() {});
```
kenzieschmoll, line 141: couple things here:
1) avoid empty setState calls if you can. Whatever values need to be updated or actions that need to be taken should take place inside the setState callback. 
2) we are potentially calling setState twice here. Instead, place all this code inside a single setState callback. 
```
setState(() {
  final newSnapshotsLength = widget.controller.snapshots.value.length;
  if (newSnapshotsLength > _currentSnapshotsLength) {
    _currentSnapshotsLength = newSnapshotsLength;
    await _scrollController.autoScrollToBottom();
  }
});
```

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_view.dart
```diff
@@ -135,8 +139,12 @@ class _StatsTable extends StatefulWidget {
     Key? key,
     required this.data,
     required this.sorting,
+    required this.selectedRecord,
   }) : super(key: key);
+
   final List<HeapStatsRecord> data;
+  final ValueNotifier<HeapStatsRecord?> selectedRecord;
```
kenzieschmoll, line 146: nit: new line

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1427,6 +1427,7 @@ extension ColorExtension on Color {
 
 /// Gets an alternating color to use for indexed UI elements.
 Color alternatingColorForIndex(int index, ColorScheme colorScheme) {
+  // First row should be 0, not 1, to contrast with header.
```
kenzieschmoll, line 1430: is this comment now obsolete?

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_view.dart
```diff
@@ -13,6 +18,185 @@ class SnapshotView extends StatelessWidget {
...
+
+class _ShallowSizeColumn extends ColumnData<HeapStatsRecord> {
+  _ShallowSizeColumn()
+      : super(
+          'Shallow\n Dart Size',
+          titleTooltip: 'Total shallow size of the instances.\n'
+              'Shallow size of an object is size of this object plus\n'
+              'the references it holds to other Dart objects in its fields\n'
+              '(this does not include the size of the fields\n'
+              ' - just the size of the reference)',
```
kenzieschmoll, line 90: nit: references to match plural fields

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_list.dart
```diff
@@ -114,53 +114,47 @@ class _SnapshotListItems extends StatefulWidget {
...
+    addAutoDisposeListener(widget.controller.selectedIndex);
 
-      final newSnapshotsLength = widget.controller.snapshots.value.length;
-      if (newSnapshotsLength > _currentSnapshotsLength) {
-        _currentSnapshotsLength = newSnapshotsLength;
-        await _scrollController.autoScrollToBottom();
-        setState(() {});
-      }
+    addAutoDisposeListener(widget.controller.itemAdded, () async {
+      await _scrollController.autoScrollToBottom();
```
kenzieschmoll, line 134: we don't need another notifier for this. We are already notifying when an item is added because the widget.controller.snapshots notifier will be triggered.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -17,26 +17,33 @@ class DiffPaneController {
...
 
+  final _snapshots = ListValueNotifier(<DiffListItem>[InformationListItem()]);
+
   /// The list contains one item that show information and all others
   /// are snapshots.
-  final snapshots = ListValueNotifier(<DiffListItem>[InformationListItem()]);
+  ValueListenable<List<DiffListItem>> get snapshots => _snapshots;
+
+  final _selectedIndex = ValueNotifier<int>(0);
+  ValueListenable<int> get selectedIndex => _selectedIndex;
```
kenzieschmoll, line 27: always order getter before private field. Order should be

getter (if present)
field
setter (if present)

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -17,26 +17,33 @@ class DiffPaneController {
...
 
-  final selectedIndex = ValueNotifier<int>(0);
+  final _isProcessing = ValueNotifier<bool>(false);
 
   /// If true, some process is going on.
   ValueListenable<bool> get isProcessing => _isProcessing;
```
kenzieschmoll, line 32: same comment here

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_list.dart
```diff
@@ -114,53 +114,47 @@ class _SnapshotListItems extends StatefulWidget {
...
   final _headerHeight = 1.20 * defaultRowHeight;
-  late ScrollController _scrollController;
-  late int _currentSnapshotsLength;
+  late final ScrollController _scrollController;
 
   @override
   void initState() {
-    super.initState();
     _scrollController = ScrollController();
+    super.initState();
```
kenzieschmoll, line 122: super.initState should always be called first

kenzieschmoll (approved):

PR #4475, opened 2022-09-14
Refactor diff tab code to prepare for more code coming.

polina-c:


packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -0,0 +1,66 @@
...
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+
+import '../../../../../primitives/utils.dart';
+import '../../../../../shared/common_widgets.dart';
+import '../../../primitives/memory_utils.dart';
+import 'model.dart';
+
+class DiffPaneController {
+  final scrollController = ScrollController();
```
kenzieschmoll, line 15: can we create this scroll controller in the widget lifecycle methods? the FlatTable should have autoscroll to bottom functionality you can leverage

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -123,162 +65,53 @@ class _DiffPaneState extends State<DiffPane> {
...
+//     final itemLocal = item;
+//     if (itemLocal is InformationListItem) {
+//       return const Text('Introduction to snapshot diffing will be here.');
+//     }
+//     if (itemLocal is SnapshotListItem) {
+//       return Text('Content of ${itemLocal.name} will be here.');
+//     }
+//     throw 'Unexpected type of the item: ${itemLocal.runtimeType}';
+//   }
+// }
```
kenzieschmoll, line 83: should this be removed?

kenzieschmoll (approved):
couple comments but lgtm

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -82,39 +26,37 @@ class DiffPane extends StatefulWidget {
...
             OutlineDecoration(
-              child: Column(
-                children: [
-                  _ListControlPane(controller: controller),
-                  Expanded(
-                    child: _SnapshotList(controller: controller),
-                  ),
-                ],
-              ),
+              child: SnapshotList(controller: controller),
```
kenzieschmoll, line 56: is it possible that the snapshots in the list will change and the selectedIndex stays the same? I think this only works right now because we change the index each time the list of snapshots changes, but if those were ever decoupled I think this would break. 

When we discussed this earlier, I was saying that the list content only needs to rebuild when the selected index changes - not when the list of snapshots changes. But the SnapshotList itself probably needs to be notified of both. You could put a ValueListenableBuilder inside SnapshotList.build that listens for the list of snapshots.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
polina-c: I use ListView, that does not have this feature

kenzieschmoll: either way, we should be creating ScrollControllers within the widget lifecycle, there is still a way to scroll to the bottom without putting this scroll controller in business logic. We could add a listener to the snapshot list, and scroll to the bottom whenever a new snapshot has been added.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_list.dart
```diff
@@ -132,6 +100,43 @@ class _SnapshotListTitle extends StatelessWidget {
...
+
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder<List<DiffListItem>>(
+      valueListenable: controller.snapshots,
+      builder: (_, list, __) {
+        return ListView.builder(
+          controller: controller.scrollController,
+          shrinkWrap: true,
+          itemCount: controller.snapshots.value.length,
```
kenzieschmoll, line 117: use `list.length` here

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_list.dart
```diff
@@ -132,6 +100,43 @@ class _SnapshotListTitle extends StatelessWidget {
   }
 }
 
+class _SnapshotListItems extends StatelessWidget {
```
kenzieschmoll, line 103: this widget should be stateful and this is where we should be creating the scroll controller and listening to the snapshot list.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_list.dart
```diff
@@ -132,6 +100,43 @@ class _SnapshotListTitle extends StatelessWidget {
...
+      valueListenable: controller.snapshots,
+      builder: (_, list, __) {
+        return ListView.builder(
+          controller: controller.scrollController,
+          shrinkWrap: true,
+          itemCount: controller.snapshots.value.length,
+          itemBuilder: (context, index) {
+            return Container(
+              height: headerHeight,
+              color: controller.selectedIndex.value == index
```
kenzieschmoll, line 121: I'm not seeing where this widget is rebuilding when `controller.selectedIndex.value` changes now? _SnapshotListItems needs to rebuild when the snapshot list changes and when the index changes. Let's make this widget stateful, and add listeners for both `controller.snapshots` and `controller.selectedIndex`, calling setState when either change and also scrolling to the bottom of the list when the snapshots list changes

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -30,58 +30,39 @@ class _DiffPaneState extends State<DiffPane> {
 
   @override
   Widget build(BuildContext context) {
-    return ValueListenableBuilder<int>(
+    final Widget listContent = ValueListenableBuilder<int>(
```
kenzieschmoll, line 33: nit: rename to snapshotView since this is not actually content in a list - it's content for a selected snapshot item in a list.

Issue #4452, opened 2022-09-08
Decouple widget eval ui from service requests



CoderDake:
Right now the widget eval ui is couple to the async requests that it depends on 
This results in a non responsive UI experience since a UI event is blocked waiting on data.

To improve responsiveness the UI should be unblocked so that the card can pop up immediately and it's pending status can be properly displayed to the user.

Issue #4450, opened 2022-09-08
Debounce Inspector Hover triggers


CoderDake:
While hovering, a lot of inspector screens can pop up if a lot of hovers are triggered and take a long time each.

We need to find a way to debounce those hover dialog popups so we don't tax the system as hard, and to improve the user experience

CoderDake:
This code will be fixed in https://github.com/flutter/devtools/pull/4488 as the ability to stop work on stale cards is addressed there.

PR #4488, opened 2022-09-16
Big Fix for HoverCard Behaviour

CoderDake:
Fixes https://github.com/flutter/devtools/issues/4452
Fixes https://github.com/flutter/devtools/issues/4450
Fixes https://github.com/flutter/devtools/issues/4499

This PR fixes a whole bunch of issues related to hover cards, and how we interact with them.

* Fix `HoverCardTooltip` so that it can never be blocked, waiting on an async operation.
* Added `HoverCardTooltipController` which controll a screen's hover state
    * This ensures we will only ever have 1 hovercard on the screen at a time (fixing some other bugs where many hovers would come up at once)
    * Each screen's controller gets one of these so that all overs on that screen will benefit from this.
* Added the concept of `isHoverStale` to our hover helpers.
    * This allows `HoverCardData` generating callback to exit early if it's hover card is no longer visible/relevant
    * this will reduce the number of our service calls a lot since stale cards will no longer try to keep getting data.
* Created `HoverCardTooltip.sync` constructor for Hover behaviour that can be generated synchronously. No spinner will be created for these cards since their data is immediately available.
*


packages/devtools_app/lib/devtools_app.dart
```diff
@@ -80,5 +80,6 @@ export 'src/shared/theme.dart';
 export 'src/shared/utils.dart';
 export 'src/shared/version.dart';
 export 'src/ui/filter.dart';
+export 'src/ui/hover.dart';
```
CoderDake, line 83: I had to do this to be able to mock for tests, is this ok?

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -42,6 +43,8 @@ class DebuggerController extends DisposableController
   final codeViewController = CodeViewController();
 
   bool _firstDebuggerScreenLoaded = false;
+  final HoverCardTooltipController hoverCardTooltipController =
```
jacob314, line 46: nit: remove type on LHS here for consistency with the code 2 lines above.

packages/devtools_app/lib/src/screens/debugger/variables.dart
```diff
@@ -114,6 +114,7 @@ Widget displayProvider(
       multiline: true,
       style: theme.fixedFontStyle,
       debuggerController: controller,
+      hoverCardTooltipController: controller.hoverCardTooltipController,
```
jacob314, line 117: nit: consider just reading this off the debuggerController if continues to be something that is hung off the debuggerController.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
jacob314: higher level, it is a little odd this is hung off the debuggerController. Maybethis should be just more a separate top level service associated with hovercards not the debugger as we don't really want multiple hover card tips regardless of whether the debugger is present and a single screen better not diverge on which controller it uses for this.

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -246,6 +247,9 @@ class InspectorController extends DisposableController
   /// inspector tree load timing.
   bool firstInspectorTreeLoadCompleted = false;
 
+  final HoverCardTooltipController hoverCardTooltipController =
```
jacob314, line 250: this shouldn't be on the inspector controller. Lets crease a separate top level controller with a provider instead. could be bugs on the inspector page as sometimes it is going through the debugger controller and sometimes the inspector one.

packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -193,6 +192,8 @@ class HoverCard {
           title: title,
         );
 
+  int? freshnessToken;
```
jacob314, line 195: make this private if possible.

packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -215,29 +221,124 @@ class HoverCard {
...
+
+  /// Helper for incrementing the freshness token.
+  void _invalidateFreshnessToken() {
+    _newFreshnessToken();
+  }
+
+  /// Check's that a [hoverCard.freshnessToken] is the same as the active
+  /// freshness token.
+  bool isHoverCardStillFresh(HoverCard hoverCard) {
+    return hoverCard.freshnessToken == _freshnessToken;
```
jacob314, line 279: instead of using tokens, you can just have a WeakRef<HoverCard> for the currently active HoverCard. I am assuming you aren't storing the active HoverCard due to concerns about leaking memory if it was stored here.

jacob314 (commented):
high level: hoverCardTipController should get its own provider that gets injected either with each screen or above all the screens in the tree. That should eliminate all of the flowing of HoverCardTipController through the code. This CL is passing the HoverCardTip Controller a very large # of places but it is only used in a few places so it will be easier to just lookup the controller when it is needed. The controller is also only needed when a hover related action is triggered so looking it up with Provider would definitely be performant.

packages/devtools_app/lib/src/screens/inspector/diagnostics.dart
```diff
@@ -194,6 +193,7 @@ class DiagnosticsNodeDescription extends StatelessWidget {
         required event,
         required isHoverStale,
       }) async {
+        await Future.delayed(const Duration(seconds: 5));
```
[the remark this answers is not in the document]
CoderDake: sorry i just accidentally committed this

packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -224,59 +226,41 @@ class HoverCard {
...
-  void removeHoverCard() {
-    _invalidateFreshnessToken();
-    _hoverCard?.remove();
-    _hoverCard = null;
-  }
-
-  /// Create increment the freshness token and return the new value.
-  int _newFreshnessToken() {
-    return ++_freshnessToken;
+  void removeHoverCard(HoverCard card) {
```
jacob314, line 251: method comment is now misleading.

packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -350,13 +330,23 @@ class _HoverCardTooltipState extends State<HoverCardTooltip> {
   /// A timer that removes a [HoverCard] when completed.
   Timer? _removeTimer;
 
+  HoverCard? _currentHoverCard;
+
+  HoverCardTooltipController? _hoverCardTooltipController;
```
jacob314, line 335: consider making this late as that seems like how you are using it.

packages/devtools_app/lib/src/app.dart
```diff
@@ -95,6 +96,9 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
   bool get denseModeEnabled => _denseModeEnabled;
   bool _denseModeEnabled = false;
 
+  final HoverCardTooltipController hoverCardTooltipController =
```
kenzieschmoll, line 99: nit: remove type on LHS

packages/devtools_app/lib/src/app.dart
```diff
@@ -324,10 +328,13 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
       builder: (context, child) {
         return Provider<AnalyticsController>.value(
           value: widget.analyticsController,
-          child: NotificationsView(
-            child: ReleaseNotesViewer(
-              releaseNotesController: releaseNotesController,
-              child: child,
+          child: Provider<HoverCardTooltipController>.value(
```
kenzieschmoll, line 331: you may be able to use MultiProvider here to condense this

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -879,7 +883,8 @@ class _LineItemState extends State<LineItem>
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
     final darkTheme = theme.brightness == Brightness.dark;
-
+    _hoverCardTooltipController ??=
+        Provider.of<HoverCardTooltipController>(context);
```
kenzieschmoll, line 887: do this in didChangeDependencies so it only has to be looked up once

packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -7,10 +7,10 @@ import 'dart:math' as math;
 
 import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
+import 'package:provider/provider.dart';
 
+import '../../devtools_app.dart';
```
kenzieschmoll, line 12: import the libraries you need directly

packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -203,8 +206,13 @@ class HoverCard {
   ///
   /// The HoverCard will not be removed if the mouse is currently inside the
   /// widget.
-  void maybeRemove() {
-    if (!_hasMouseEntered) remove();
+  /// Returns whether or not the HoverCard was removed
```
kenzieschmoll, line 209: nit: period at end of sentence

packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -215,23 +223,93 @@ class HoverCard {
   }
 }
 
+/// Ensures that only one [HoverCard] is ever displayed at a time.
+class HoverCardTooltipController {
```
kenzieschmoll, line 227: should this be 'HoverCartController' instead of HoverCardTooltipController? adding 'tooltip' kind of makes it sound like this controls tooltips on the hover cards, not the cards themselves

packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -215,23 +223,93 @@ class HoverCard {
...
+  /// The card that is currenty being displayed
```
kenzieschmoll, line 228: nit: period at end of sentence

packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -215,23 +223,93 @@ class HoverCard {
...
+  HoverCard? _currentHoverCard;
+
+  /// Sets [hoverCard] as the most recently displayed [HoverCard].
+  ///
+  /// This methods sets a unique [hoverCard.freshnessToken], as a way to track whether
```
kenzieschmoll, line 233: this comment seems obsolete. Also, maybe rename to 'show' since 'set' is kind of a loaded word? (setters)

packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -215,23 +223,93 @@ class HoverCard {
...
 /// A hover card based tooltip.
 class HoverCardTooltip extends StatefulWidget {
+  /// A [HoverCardTooltip] that generates it's [HoverCardData] asynchronously.
+  ///
+  /// [asyncGenerateHoverCardData] is used to generate the data that will
+  /// display in the final [HoverCard]. While that data is being generated,
+  /// a [HoverCard] with a spinner will show. If any [HoverCardData] returned
+  /// from [asyncGenerateHoverCardData] the spinner [HoverCard] will be replaced
+  /// with one containing the generated [HoverCardData].
   const HoverCardTooltip({
```
kenzieschmoll, line 273: maybe we should make this HoverCardTooltip.async so it is immediately clear which constructor to use based on your use case?

packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -249,16 +327,24 @@ class _HoverCardTooltipState extends State<HoverCardTooltip> {
   /// A timer that removes a [HoverCard] when completed.
   Timer? _removeTimer;
 
-  /// Displays the evaluation result of a source code item.
-  HoverCard? _hoverCard;
+  HoverCard? _currentHoverCard;
```
kenzieschmoll, line 330: do we need to store this here too since it is also stored on HoverCardTooltipController?

packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -314,13 +444,19 @@ class _HoverCardTooltipState extends State<HoverCardTooltip> {
...
+      _hoverCardTooltipController.removeHoverCard(_currentHoverCard!);
+    }
     widget.disposable?.dispose();
     super.dispose();
   }
 
   @override
   Widget build(BuildContext context) {
+    _hoverCardTooltipController =
+        Provider.of<HoverCardTooltipController>(context);
```
kenzieschmoll, line 459: do this in didChangeDependencies so we only have to look it up once

packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -447,16 +446,20 @@ class _HoverCardTooltipState extends State<HoverCardTooltip> {
...
+      _hoverCardController.removeHoverCard(_currentHoverCard!);
     }
     widget.disposable?.dispose();
     super.dispose();
   }
 
+  @override
+  void didChangeDependencies() {
+    _hoverCardController = Provider.of<HoverCardController>(context);
+    super.didChangeDependencies();
```
kenzieschmoll, line 458: call to super first

kenzieschmoll (approved):

packages/devtools_app/lib/src/ui/hover.dart
CoderDake: ouu interesting.
Screen Shot 2022-09-22 at 1 27 48 PM

The default template puts the TODO at the top, so I had put it above the super purposely.

kenzieschmoll: @DanTup this would be a good polish item for VS code. To add the TODO where the implementation is supposed to go.

DanTup: Yeah, seems like this should probably be the other way around. Although, there's also an issue open about removing TODOs at https://github.com/dart-lang/sdk/issues/49331. I'll ping on there and see if there are opinions of changing this order vs just removing (the change will be in the Dart analysis server rather than VS Code, as it handles most code edits/generation like this).

PR #4491, opened 2022-09-19
Add offline support for the Raster stats data in the performance page

kenzieschmoll:
This PR
- Breaks the former RasterMetricsController class into two classes (RasterStats, the data model, and RasterStatsController, the controller the UI interacts with)
- Includes RasterStats data with the existing PerformanceData class that is exported and imported for the Performance page
- Renames all uses of "Raster Metrics" to "Raster Stats"


packages/devtools_app/lib/src/screens/performance/panes/raster_stats/raster_stats_model.dart
```diff
@@ -0,0 +1,178 @@
...
+
+  double get percentRenderingTimeAsDouble =>
+      duration.inMicroseconds / totalRenderingDuration!.inMicroseconds;
+
+  String get percentRenderingTimeDisplay =>
+      percent2(percentRenderingTimeAsDouble);
+
+  String get displayName => 'Layer $id';
+
+  Map<String, Object?> get json => {
```
CoderDake, line 169: Instead of manually generating serializers, would we be open to using a serialization package?
https://pub.dev/packages/json_serializable

packages/devtools_app/lib/src/screens/performance/performance_controller.dart
```diff
@@ -611,6 +611,10 @@ class PerformanceController extends DisposableController
     List<TraceEvent> threadNameEvents, {
     bool isInitialUpdate = false,
   }) {
+    // This can happen if there is a race between this method being called and
+    // losing connection to the app.
+    if (serviceManager.connectedApp == null) return;
```
CoderDake, line 616: Does this squelch something that we would want to reflect?
Perhaps is there an error logging message we could send, or and assertion we could set in case this is ever happening regularly and causing problems?

packages/devtools_app/lib/src/screens/performance/panes/raster_stats/raster_stats_model.dart
kenzieschmoll: We could certainly look into it. We'll want to make we can use a serialization package in g3 first before making a switch. I'd prefer to do this as part of a larger overhaul, and not as part of this PR, since we are already doing manual serialization in a lot of places in DevTools.

packages/devtools_app/lib/src/screens/performance/panes/raster_stats/raster_stats_model.dart
```diff
@@ -0,0 +1,178 @@
...
+
+  static const _frameWidthKey = 'frame_width';
+
+  static const _frameHeightKey = 'frame_height';
+
+  final List<LayerSnapshot> layerSnapshots;
+
+  final Size? originalFrameSize;
+
+  final Duration totalRasterTime;
```
CoderDake, line 85: looks like this could be a getter

packages/devtools_app/lib/src/screens/performance/performance_controller.dart
kenzieschmoll: `updateThreadIds` is called from `_pullTraceEventsFromVmTimeline`, which is called using a RateLimiter and polling timer. When we lose connection from the app, we cancel the timer and rate limiter
https://github.com/flutter/devtools/blob/0e3f7e76f9aa1b769267adac524dbf3752393547/packages/devtools_app/lib/src/screens/performance/performance_controller.dart#L241-L244

But the call to `_pullTraceEventsFromVmTimeline` (which calls `updateThreadIds`) could already be in progress. So adding a check that we are still connected here prevents us from calling `serviceManager.connectedApp!` on a null value and unnecessarily continuing on with the rest of the method.

CoderDake (approved):

packages/devtools_app/lib/src/screens/performance/panes/raster_stats/raster_stats_model.dart
kenzieschmoll: Created https://github.com/flutter/devtools/issues/4493 to track.

Issue #4497, opened 2022-09-20
Performance screen scrollbar sometimes gets out of sync with some content.


jacob314:
Demo showing the labels getting out of sync with the chart bars:
![unlinked_scroll_bug](https://user-images.githubusercontent.com/1226812/191384365-ccddae04-ae2c-44a0-918e-36059675f07f.gif)
This is likely some weird edge case with linked scroll controller. We don't really need it for this UI and could use a single ListView.

PR #4503, opened 2022-09-21
Simplify frames chart to use a single ListView instead of multiple ListViews

jacob314:
This fixes https://github.com/flutter/devtools/issues/4497
by eliminating an unnecessary use of LinkedScrollController.

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/flutter_frames_chart.dart
+++ b/packages/devtools_app/lib/src/screens/performance/flutter_frames_chart.dart
@@ -388,9 +388,8 @@ class FlutterFramesChartItem extends StatelessWidget {
             children: [
               content,
               Container(
-                padding: EdgeInsets.only(top: defaultChartHeight),
-                height: defaultChartHeight +
-                    _FlutterFramesChartState._frameNumberSectionHeight,
+                margin: EdgeInsets.only(top: defaultChartHeight),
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/screens/performance/flutter_frames_chart.dart
```diff
@@ -361,48 +335,71 @@ class FlutterFramesChartItem extends StatelessWidget {
...
+        ),
       ),
     );
+    return index % 2 == 0
+        ? Stack(
+            children: [
+              content,
+              Container(
+                padding: EdgeInsets.only(top: defaultChartHeight),
+                height: defaultChartHeight +
```
kenzieschmoll, line 392: giving this container `height: _FlutterFramesChartState._frameNumberSectionHeight` and using `margin:defaultChartHeight` instead of padding may make this more clear.

Or instead of using margin, you could wrap in a Positioned widget and place directly at the offset you want it in the stack

kenzieschmoll (approved):
lgtm with one comment

packages/devtools_app/lib/src/screens/performance/flutter_frames_chart.dart
jacob314: I like it! done.

PR #4504, opened 2022-09-21
Save last VM Service URI on landing screen in debug mode only.

jacob314:
I found it annoying to have to paste the VM Service URI of the sample app I was using into DevTools after hot restarting or re-running devtools.

Only for debug mode, I've tweaked DevTools to save the last VM Service URI used. This will save us time as long as most of the time when debugging devtools we just want to connect to the previous sample app. I haven't enabled this in profile or release builds as I don't think the previously used VM Service URI is useful unless you are iterating on the DevTools app.

The extra package added as a dependency already exists in g3 and should be dead code that is compiled out of release builds.

### DIFF
--- a/packages/devtools_app/pubspec.yaml
+++ b/packages/devtools_app/pubspec.yaml
@@ -43,6 +43,8 @@ dependencies:
   mime: ^1.0.0
   path: ^1.8.0
   provider: ^6.0.2
+  # Only used for debug mode logic.
+  shared_preferences: ^2.0.15
   sse: ^4.0.0
   stack_trace: ^1.10.0
   string_scanner: ^1.1.0
--- a/packages/devtools_app/lib/src/framework/landing_screen.dart
+++ b/packages/devtools_app/lib/src/framework/landing_screen.dart
@@ -5,6 +5,7 @@
 import 'package:devtools_shared/devtools_shared.dart';
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
+import 'package:shared_preferences/shared_preferences.dart';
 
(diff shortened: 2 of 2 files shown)


packages/devtools_app/pubspec.yaml
```diff
@@ -43,6 +43,8 @@ dependencies:
   mime: ^1.0.0
   path: ^1.8.0
   provider: ^6.0.2
+  # Only used for debug mode logic.
+  shared_preferences: ^2.0.15
```
kenzieschmoll, line 47: couple questions 
- can this be moved to dev_dependencies?
- will this persist only for a single instance of the macos runner or chrome? or does it persist if you do a full stop app and re-run?

kenzieschmoll (commented):
this looks really useful!

kenzieschmoll (approved):

bkonyi (approved):
I was thinking of doing something similar just this morning!

packages/devtools_app/pubspec.yaml
jacob314: I don't think it can be moved to dev dependencies as it is part of the main package and I don't know an easy way to move it outside the main package short of having a wacky debugging entry point for DevTools outside the lib directory.

jacob314: persists even if you do a full-stop and re-run. personally I found that useful as sometimes Flutter gets in a bad state or I'm paranoid it is in a bad state so I full stop. We could store the time-stamp of the last run and only reuse the setting if it is less than an hour old or something to reduce the # of annoying cases.

PR #4505, opened 2022-09-21
Fix exception in flame chart when widget became unmounted before callback is completed

jacob314:


packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -561,11 +561,11 @@ abstract class FlameChartState<T extends FlameChart,
     // had time to update their scroll extents. Otherwise, we can hit a race
     // where are trying to scroll to an offset that is beyond what the scroll
     // controller thinks its max scroll extent is.
-    WidgetsBinding.instance.addPostFrameCallback((_) async {
-      await Future.wait([
```
kenzieschmoll, line 565: do we need to remove the await here? this code is a bit fragile, so I just want to make sure we've checked that this doesn't have unexpected consequences

jacob314: the await did absolutely nothing because FrameCallback is a void function
https://api.flutter.dev/flutter/scheduler/FrameCallback.html so the value returned (in this case a Future) was just ignored.

kenzieschmoll (approved):

PR #4513, opened 2022-09-22
Add support for inspecting objects in an isolate's object store

bkonyi:
The object store is used to provide easy and cheap access to important VM objects within the VM. Examples of objects stored in the object store include:

  - Code stubs (e.g., allocation paths, async/await machinery)
  - References to core classes (e.g., `Null`, `Object`, `Never`)
  - Common type arguments (e.g., `<String, dynamic>`)
  - References to frequently accessed fields / functions (e.g., `_objectEquals()`, `_Enum._name`)
  - Cached, per-isolate data (e.g., library URI mappings)

The object store is considered one of the GC roots by the VM's garbage collector, meaning that objects in the store will be considered live, even if they're not referenced anywhere else in the program.

The `Object Store` tab under the `Objects` VM tooling tab allows for users to inspect elements contained within the object store, where they may not necessarily be reachable from the program explorer. Tapping a link to an object store entry will display the entry in the object viewport and, if present in the program explorer, mark the relevant `VMServiceObjectNode`s as selected. 

image


packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
```diff
@@ -46,18 +52,59 @@ class _ObjectInspectorViewState extends State<_ObjectInspectorView> {
         Provider.of<VMDeveloperToolsController>(context);
     controller = vmDeveloperToolsController.objectInspectorViewController!
       ..init();
+    tabController = TabController(
```
kenzieschmoll, line 55: dispose this in dispse

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
```diff
@@ -46,18 +52,59 @@ class _ObjectInspectorViewState extends State<_ObjectInspectorView> {
...
+      length: 2,
+      vsync: this,
+      // Disable the animation.
+      animationDuration: const Duration(),
```
kenzieschmoll, line 59: what's the reasoning for disabling this?

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
```diff
@@ -46,18 +52,59 @@ class _ObjectInspectorViewState extends State<_ObjectInspectorView> {
...
-          controller: controller.programExplorerController,
-          onNodeSelected: _onNodeSelected,
-          title: 'Program Explorer',
+        Column(
+          children: [
+            Container(
+              color: theme.titleSolidBackgroundColor,
+              child: OutlineDecoration(
+                showBottom: false,
+                child: TabBar(
```
kenzieschmoll, line 76: if we want to track analytics for this, we should to use `AnalyticsTabbedView`.

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
```diff
@@ -46,18 +52,59 @@ class _ObjectInspectorViewState extends State<_ObjectInspectorView> {
...
+                  controller: tabController,
+                  tabs: [
+                    Tab(
+                      height: defaultRowHeight,
+                      text: 'Program Explorer',
+                    ),
+                    Tab(
```
kenzieschmoll, line 83: DevToolsTab provides helpers to respect font size changes. Use that if possible.

packages/devtools_app/lib/src/screens/vm_developer/object_store.dart
```diff
@@ -0,0 +1,92 @@
...
+  @override
+  bool get includeHeader => true;
+
+  @override
+  String getValue(ObjectStoreEntry dataObject) {
+    return dataObject.key;
+  }
+}
+
+class _Object extends ColumnData<ObjectStoreEntry> implements ColumnRenderer {
```
kenzieschmoll, line 28: nit: name _EntryColumn and _ObjectColumn to make it clear that these are column classes and not data classes

packages/devtools_app/lib/src/screens/vm_developer/object_store_controller.dart
```diff
@@ -0,0 +1,27 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: 2022

packages/devtools_app/lib/src/screens/vm_developer/object_store_controller.dart
```diff
@@ -0,0 +1,27 @@
...
+// Use of this source code is governed by a BSD-style license that can be
```
kenzieschmoll, line 2: can we create a subfolder in vm_developer for the object_inspector? there are starting to be quite a few files for this tool

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -765,6 +765,16 @@ class ToolbarAction extends StatelessWidget {
   }
 }
 
+class BlankHeader extends StatelessWidget implements PreferredSizeWidget {
```
kenzieschmoll, line 768: can you add some dart doc here to describe why we need a widget for this when `SizedBox()` would be the equivalent? at first glance I think it is because we need it to be a subclass of PreferredSizeWidget, but some dart doc could help

packages/devtools_app/lib/src/screens/vm_developer/object_inspector_view.dart
bkonyi: It just looks really out of place and makes navigating between the program explorer and object store look slow.

kenzieschmoll (approved):

Issue #4484, opened 2022-09-15
Memory > Profile does not sort by class name.


polina-c:
<img width="410" alt="Screen Shot 2022-09-15 at 4 10 14 PM" src="https://user-images.githubusercontent.com/12115586/190524411-50cdf9b1-976e-4539-8935-d25efd88c756.png">

PR #4523, opened 2022-09-26
Enable new "Profile" and "Allocation Tracing" sub-tabs in Memory screen

bkonyi:
Also removes the old "Allocations" tool and its relevant code.


kenzieschmoll (approved):

polina-c:
Please rename columns to be consistent with diff tab:
Size / Internal / External - > Total Shallow / Shallow Dart / Shallow Native

For total shallow reference this tooltip: https://github.com/flutter/devtools/blob/b8672b0af21378bcd9d3b23b397b6f598713a0f4/packages/devtools_app/lib/src/screens/memory/shared/heap/primitives.dart#L5

polina-c:
The bug is still here: https://github.com/flutter/devtools/issues/4484
Can you fix it?

polina-c:
And sorting by instances does not seem to work too:
Screen Shot 2022-09-26 at 2 45 03 PM

polina-c:
Tooltips for columns are the same in every group. We need to either to add tooltips for column groups (seems to be cleaner) or to have little different tooltips for the columns themselves.

bkonyi:
@polina-c I'm going to follow up on this change with additional PRs that clean things up. Didn't want to put them all in one PR.

bkonyi:
> The bug is still here: #4484 Can you fix it?

> And sorting by instances does not seem to work too: Screen Shot 2022-09-26 at 2 45 03 PM

Both of these are fixed by https://github.com/flutter/devtools/pull/4532.

bkonyi:
> Please rename columns to be consistent with diff tab: Size / Internal / External - > Total Shallow / Shallow Dart / Shallow Native
> 
> And in the documentation: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/ALLOCATION_PROFILE.md
> 
> For total shallow reference this tooltip:
> 
> https://github.com/flutter/devtools/blob/b8672b0af21378bcd9d3b23b397b6f598713a0f4/packages/devtools_app/lib/src/screens/memory/shared/heap/primitives.dart#L5

"Shallow Native" isn't necessarily correct as it's non-Dart memory and only as accurate as the values provided by the embedder (see the [specification here](https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#memoryusage)). There's no requirement for the provided value to be deep/shallow, or for the embedder to provide any value (e.g., the Flutter engine no longer reports image data sizes via this field as it manages the lifecycle of images independently from the VM's garbage collector).

I'm also not convinced that it's necessary to change the headers to be this verbose when we can just update the tooltips to better explain that these are shallow memory values, especially when there's no retained memory column in the same context. I propose that we use the following instead:

- "Size" -> "Total Size" Tooltip: "The sum of the object type's total shallow memory consumption in the Dart heap and associated external (e.g., non-Dart heap) allocations"
- "Internal" -> "Dart Heap" Tooltip: "$shallowSizeColumnTooltip"
- "External" -> "External" Tooltip: "Non-Dart heap allocated memory associated with an object type"

kenzieschmoll:
> "Size" -> "Total Size" Tooltip: "The sum of the object type's total shallow memory consumption in the Dart heap and associated external (e.g., non-Dart heap) allocations"
"Internal" -> "Dart Heap" Tooltip: "$shallowSizeColumnTooltip"
"External" -> "External" Tooltip: "Non-Dart heap allocated memory associated with an object type"

SGTM - does "Dart Heap" + "External" for an object = "Total Size"?

Issue #4524, opened 2022-09-26
File opener and search are broken


elliette:
Previously, you could open the file search modal by clicking on the file name or using cmd/ctrl + p. Looks like this functionality was accidentally removed with https://github.com/flutter/devtools/pull/4447, if `enableFileExplorer` is `true`, we never show the file opener: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/debugger/codeview.dart#L232-L23 4. 

We should also add a test to catch this in the future.

elliette:
Looks like search is also broken. I think the ternaries are backwards.

PR #4525, opened 2022-09-26
Fixes broken file opener and search

elliette:
Fixes https://github.com/flutter/devtools/issues/4524

The ternaries are backwards, so we weren't showing the file opener / file search if they were enabled. 

Will add test cases for both in follow-up PR so that we don't accidentally disable them in the future.

### DIFF
--- a/packages/devtools_app/macos/Runner.xcodeproj/project.pbxproj
+++ b/packages/devtools_app/macos/Runner.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 48;
+	objectVersion = 54;
 	objects = {
 
 /* Begin PBXAggregateTarget section */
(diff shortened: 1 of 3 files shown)


bkonyi:
shame

LGTM!

packages/devtools_app/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -192,7 +192,7 @@
 			isa = PBXProject;
 			attributes = {
 				LastSwiftUpdateCheck = 0920;
-				LastUpgradeCheck = 0930;
```
bkonyi, line 195: Do we want to commit these files?

bkonyi (approved):

packages/devtools_app/macos/Runner.xcodeproj/project.pbxproj
elliette: Good point, removed!

PR #4529, opened 2022-09-26
Upgrade tests deps in refresh

CoderDake:


bkonyi (approved):
Remove the redundant `pub get` since `pub upgrade` does that already.

PR #4530, opened 2022-09-26
Prepare dev release for g3 roll

elliette:
I would like to roll DevTools into g3 to fix https://github.com/flutter/devtools/issues/4524

Following instructions to prepare dev release here: https://github.com/flutter/devtools/tree/master/tool#prepare-the-release

Ran `dart tool/update_version.dart auto --type dev`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 2.17.0-dev.0
+TODO: update changelog
+
 ## 2.17.0
 * Prepare for 2.17.0 release [#4428](https://github.com/flutter/devtools/pull/4428)
 * Revert perfetto bundle [#4422](https://github.com/flutter/devtools/pull/4422)
--- a/packages/devtools_app/lib/devtools.dart
+++ b/packages/devtools_app/lib/devtools.dart
@@ -7,4 +7,4 @@
 // that updates all versions for DevTools.
 // Note: a regexp in tools/update_version.dart matches the following line so
 // if you change it you must also modify tools/update_version.dart.
-const String version = '2.17.0';
+const String version = '2.17.0-dev.0';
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 2.17.0-dev.0
+TODO: update changelog
```
elliette, line 2: What should the CHANGELOG look like? Should I remove the TODO? Or generate the changelog (in which case we should update instructions in https://github.com/flutter/devtools/tree/master/tool#prepare-the-release)

kenzieschmoll: I think TODO is fine for dev release. The generate changelog tool isn't smart enough to use releases that aren't formally tagged on our repo as a 'before' data for the range of commits. Plus, the release will be going out in a couple of days so this changelog will get updated then.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 2.17.0-dev.0
```
kenzieschmoll, line 1: this version should be 2.18.0-dev.0, as 2.17.0-dev.0 is actually before 2.17.0 in semantic versioning

kenzieschmoll (approved):
lgtm once version is changed

CHANGELOG.md
CoderDake: perhaps we should just leave a bullet mentioning it is a dev release, so there aren't any notes needed?
Leaving todo's implies that something needs to be done and might come up later

CHANGELOG.md
CoderDake: This might need to be updated in our version bumping script.
if we wanted to make a dev version of `2.17.0` what would you expect the dev version to be?

elliette: Discussed offline! By default we will have the `dev` version be preparatory for next `minor` release

CHANGELOG.md
elliette: Added that to https://github.com/flutter/devtools/pull/4536 :)

PR #4535, opened 2022-09-27
Add support for user-specified export locations of allocation profile CSVs

bkonyi:
image

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
+++ b/packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
@@ -3,6 +3,7 @@
 // found in the LICENSE file.
 
 import 'package:flutter/material.dart';
+import 'package:path/path.dart' as path;
 import 'package:vm_service/vm_service.dart';
 
 import '../../../../analytics/constants.dart' as analytics_constants;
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
```diff
@@ -398,10 +399,19 @@ class _ExportAllocationProfileButton extends StatelessWidget {
           tooltip: 'Download allocation profile data in CSV format',
           onPressed: currentAllocationProfile == null
               ? null
-              : () {
-                  final file = allocationProfileController
+              : () async {
+                  final loc = await allocationProfileController
```
kenzieschmoll, line 403: nit: use 'location' instead of abbreviating (I initially thought "loc" = "line of code")

packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view_controller.dart
```diff
@@ -66,7 +67,15 @@ class AllocationProfileTableViewController extends DisposableController
   /// Converts the current [AllocationProfile] to CSV format and downloads it.
   ///
   /// The returned string is the name of the downloaded CSV file.
-  String downloadMemoryTableCsv(AllocationProfile profile) {
+  Future<String?> downloadMemoryTableCsv(AllocationProfile profile) async {
+    final path = await getSavePath(
```
kenzieschmoll, line 71: can you file an issue to upstream this user-specified location option into export controller directly? That way other screens can leverage this functionality.

kenzieschmoll (commented):
did you manually verify this works well on Desktop and Web?

bkonyi:
> did you manually verify this works well on Desktop and Web?

Good call, I hadn't on web and it doesn't seem to work. Will need to follow up on this one, but I don't think it's critical for the next release.

kenzieschmoll:
> Good call, I hadn't on web and it doesn't seem to work. Will need to follow up on this one, but I don't think it's critical for the next release.

Agreed it is not critical. We probably need to add web and desktop implementations (IIRC there is a file_selector_web plugin we use?), so this is another reason to upstream this functionality into the import_export code, which already has platform specific files. We may also want to consider having the user set a common setting for download location for all of DevTools instead of requiring them to specify a download location for each download on each page.

bkonyi:
It looks like it's not even possible to open a `Save As` dialog on web, so `getSavePath()` always returns an empty string in that context, so I'm going to abandon this change.

PR #4536, opened 2022-09-27
Update version tool logic for `dev` versions

elliette:
* By default,`dart tool/update_version.dart auto --type dev` creates a `dev` version for the next `minor` version
* The `devType` arg can be provided to create a `dev` version for the next `major` or `patch` version instead (eg, ` dart tool/update_version.dart auto --type dev --devType patch`)
* CHANGELOG for dev versions shows:
```
## 2.18.0-dev.0
* Dev version
```


tool/update_version.dart
```diff
@@ -288,19 +289,34 @@ class AutoUpdateCommand extends Command {
...
+        abbr: 'd',
+        allowed: ['patch', 'minor', 'major'],
+        allowedHelp: {
+          'patch': 'bumps the version to the next patch value',
+          'minor': 'bumps the version to the next minor value',
+          'major': 'bumps the version to the next major value',
+        },
+        mandatory: false,
+        defaultsTo: 'minor',
+        help: '''Bumps the preparory devtools version by the selected type.
```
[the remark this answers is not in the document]
elliette: Oops! It was indeed, thanks!

tool/update_version.dart
```diff
@@ -288,19 +289,34 @@ class AutoUpdateCommand extends Command {
         },
         mandatory: true,
         help: 'Bumps the devtools version by the selected type.');
+
+    argParser.addOption('devType',
```
CoderDake, line 297: What would you think about us changing 'type' to a multioption instead.
So instead  of doing `--type dev --devType patch` we could do `--type dev,patch`  (as a multioption would allow multiple values to be entered)

then we could have a logic flow on the inside.
Starting version: `2.17.0`
- `--type dev`
   - Fails with message "no previous dev version defined, please also add patch,minor, or major, to your type selection" (or something to this effect)
- `--type dev,patch`
    - updates to `2.17.1-dev.0`

Starting version: `2.17.0-dev.3`
- `--type dev`
    - updates to `2.17.0-dev.4`
- `--type dev,patch`
   - updates to `2.17.1-dev.0`

CoderDake: perhaps for the error message when no type is passed along with dev, it could say
`Current version 2.17.0 has no dev pre-release. Please use '--type dev,patch' OR  '--type dev,minor', OR  '--type dev,major' to specify which version you would like to bump

elliette: Sorry, this fell by the wayside! Made the proposed changes, with one change: if only `dev` is specified, we default to a `minor` pre-release (instead of showing an error message) since that is the default. We can still specify `patch` or `major` versions with `dev,patch` or `dev,major`

kenzieschmoll (approved):

Issue #4537, opened 2022-09-27
[Allocation Tracing] Add tests for multi-isolate support


bkonyi:

srawlins:
Tests covering multi-isolate support for allocation tracing now exist in [`tracing_pane_controller_test.dart`](https://github.com/flutter/devtools/blob/master/packages/devtools_app/test/screens/memory/tracing/tracing_pane_controller_test.dart).

`TracePaneController` maintains a `stateForIsolate` map keyed by isolate id along with the currently selected isolate, and the tests exercise construction with multiple isolates, construction with an invalid selected isolate, and serialization / deserialization round trips of that multi-isolate state. Closing this as done.

If there is more work to do here, please let us know by filing a new issue with up to date information. Thanks!

PR #4538, opened 2022-09-27
Add multi-isolate support for `Allocation Tracing` tab

bkonyi:
This change introduces support for allocation tracing on multiple isolates. Each isolate has the following allocation tracing state that is saved and restored when the current isolate is changed:

- The set of traced classes
- The most recently requested allocation traces for traced classes
- The state of the last class filter applied
- The current selection in the class table

Multi-isolate allocation tracing tests will be included in a follow-up
PR (tracking issue: https://github.com/flutter/devtools/issues/4537).


packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_view_controller.dart
```diff
@@ -56,81 +57,59 @@ class TracedClass with PinnableListEntry {
...
+      tracedClasses[cls.id!] = TracedClass(cls: cls);
     }
-    _filteredClassList.addAll(_tracedClasses.values);
-    _unfilteredClassList.addAll(_tracedClasses.values);
-
-    await refresh();
-    _initializing.value = false;
+    _filteredClassList
+      ..clear()
+      ..addAll(tracedClasses.values);
```
kenzieschmoll, line 106: use replaceAll so we don't notify twice

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_view_controller.dart
```diff
@@ -216,9 +209,90 @@ class AllocationProfileTracingViewController extends DisposableController
...
+
+    await updateState();
+    await refresh();
+
+    _initializing.value = false;
+  }
+
+  @override
+  void dispose() {
+    super.dispose();
```
kenzieschmoll, line 276: super.dispose should be called last

kenzieschmoll (approved):
lgtm with a couple comments

Issue #4541, opened 2022-09-28
New `ObjectInspectorViewController` preventing Dart Devtools roll into google3


elliette:
When running Dart DevTools in google3 with the recent changes on head, I see the following error: `Uncaught TypeError: Cannot read properties of null (reading '_sortedScripts')`

This is coming from the `ObjectInspectorViewController`, which is being instantiated for ACX apps because:

1.  We unconditionally instantiate the `VMDeveloperToolsController` here: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/app.dart#L569
2. The `VMDeveloperToolsController` instantiates the `ObjectInspectorViewController` : https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/vm_developer/vm_developer_tools_controller.dart#L14 (this was enabled in https://github.com/flutter/devtools/pull/4514)
3. `scriptManager` is `null` in the `ObjectInspectorViewController`: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/vm_developer/object_inspector/object_inspector_view_controller.dart#L23

A simple fix is to provide the `VMDeveloperToolsController` with `createController`, but I see that was removed to fix a null-pointer exception in https://github.com/flutter/devtools/pull/2905

PR #4542, opened 2022-09-28
Provide `VMDeveloperToolsController` with `createController`

elliette:
Null-pointer exception mentioned in https://github.com/flutter/devtools/pull/2905 is no longer an issue. 

Also fixes a null-cast error (not sure if this is related to the `createController` changes)

Fixes  https://github.com/flutter/devtools/issues/4541


bkonyi (approved):
Thanks for fixing!

Issue #4544, opened 2022-09-28
Add analytics to track “broken” inspector state


CoderDake:
- empty tree (one already kind of exists for this state - clicking refresh on an empty tree)
- when a tree loads with one node [root] this is an indication that there are not pub roots set
- ask Jacob for others?
- see how badly things go if you’re paused at a breakpoint
- When firstInspectorTreeLoadCompleted is done check if root is there and tree empty, log error them

PR #4553, opened 2022-09-29
Empty tree root analytics

CoderDake:
Fixes https://github.com/flutter/devtools/issues/4544

Add analytics events for:
- every time the inspector tree is loaded and is empty
- any time the user refreshes the tree and the tree comes back empty


CoderDake:
Closing for now as this needs more discussion

CoderDake:
[Request Changes] Send the message after the inferpubroot directory. THat will be a bit more accurate

PR #4554, opened 2022-09-29
fix analyzer errors

CoderDake:
https://github.com/flutter/flutter/commit/2051f09c68de6f0a6f435457b18ac63df7d08858 replaces `of` with `maybeOf`  and `of` now returns a non-optional value.

Updating  our code since the analyzer is preventing a roll to g3


CoderDake:
This won't be needed until our bot version catches up with the updates

PR #4566, opened 2022-09-30
Rebuild Indicator UI support

jacob314:
This adds rebuild indicator support which is disabled behind a flag until at least Flutter support for a couple new features needed for rebuild indicators is added.
There is a stager offline sample demonstrating rebuilds from a modified version of Flutter.
I've also deleted the CPU profiler window at the bottom of the performance page as it made performance painful as well as being confusing if you weren't on the timeline events tab as that was the only tab synced with it. Rebuild indicators are only available in debug mode currently but supporting in profile mode is possible.

I noticed a bug in the logic to track whether the performance page was currently processing so I also tossed in a fix for that on this CL. See FutureWorkTracker which avoids edge cases not handled in the previous performance controller impl.

Rebuilds are shown in two places in the performance page:
Screen Shot 2022-09-30 at 3 42 20 PM
Screen Shot 2022-09-30 at 3 41 10 PM


packages/devtools_app/lib/src/screens/performance/performance_controller.dart
```diff
@@ -773,12 +811,14 @@ class PerformanceController extends DisposableController
 
     // Process trace events [processTraceEventsHelper] and time the operation
     // for analytics.
-    await ga.timeAsync(
-      analytics_constants.performance,
-      analytics_constants.traceEventProcessingTime,
-      asyncOperation: processTraceEventsHelper,
-      screenMetricsProvider: () => PerformanceScreenMetrics(
-        traceEventCount: processingTraceCount,
+    unawaited(
```
kenzieschmoll, line 814: why do we no longer want to await `processTraceEventsHelper`?

packages/devtools_app/lib/src/screens/performance/performance_controller.dart
```diff
@@ -949,8 +989,9 @@ class PerformanceController extends DisposableController
     firstWellFormedFrameMicros = null;
     _selectedTimelineEventNotifier.value = null;
     _selectedFrameNotifier.value = null;
-    _processing.value = false;
+    _workTracker.clear();
     serviceManager.errorBadgeManager.clearErrors(PerformanceScreen.id);
+    data?.rebuildCountModel.clearAllCounts();
```
kenzieschmoll, line 994: this is redundant since we already call `data?.clear()` above, so we can remove this

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
```diff
@@ -20,112 +21,297 @@ import '../../primitives/utils.dart';
...
+    _locations.notifyListeners();
+  }
+}
+
+/// Mapping from ids to [Location] objects.
+///
+/// Some Locations may be unresolved with ids but no valid path while valid
+/// paths are being fetched from the running Flutter application.
+class LocationMap {
+  LocationMap();
```
kenzieschmoll, line 59: can we remove this empty constructor?

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
```diff
@@ -20,112 +21,297 @@ import '../../primitives/utils.dart';
...
+
+  LocationMap.fromJson(Map<String, dynamic> json) {
+    processLocationMap(json);
+  }
```
kenzieschmoll, line 63: can this be a factory constructor or a static method?

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
```diff
@@ -20,112 +21,297 @@ import '../../primitives/utils.dart';
...
+        ids.add(location.id);
+        lines.add(location.line!);
+        columns.add(location.column!);
+        names.add(location.name!);
+      }
+      json[path] = {
+        'ids': ids,
+        'lines': lines,
+        'columns': columns,
+        'names': names,
```
kenzieschmoll, line 114: pull these strings out into consts to be used here and below

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
```diff
@@ -20,112 +21,297 @@ import '../../primitives/utils.dart';
...
-            id: id,
-            path: file,
-            line: line,
-            column: entries[i + 2],
-            name: '$shortName:$line',
-          );
-          _locationMap[id] = location;
-          _locations.add(location);
-        }
+  RebuildCountModel.fromJson(Map<String, dynamic> json) {
```
kenzieschmoll, line 174: same comment here: can we remove empty constructor and can we make fromJson a facotry constructor or a static method

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
```diff
@@ -20,112 +21,297 @@ import '../../primitives/utils.dart';
...
+        (json['frames'] as List<dynamic>).cast<Map<String, dynamic>>();
+    frames.forEach(processRebuildsForFrame);
+  }
+
+  // Maximum number of historic frames to keep rebuild counts to ensure memory
+  // usage from rebuild counts is not excessive.
+  static const int rebuildFrameCacheSize = 10000;
+
+  /// Source of truth for all resolution fo location ids to [Location] objects.
+  final LocationMap locationMap = LocationMap();
```
kenzieschmoll, line 187: nit: remove type on LHS

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
```diff
@@ -20,112 +21,297 @@ import '../../primitives/utils.dart';
...
+
+  /// Map from frame id to list of rebuilds for the frame.
+  final _rebuildsForFrame = <int, List<RebuildLocation>>{};
+
+  final _stats = _RebuildCountStats();
+  ValueListenable<List<RebuildLocation>> get locationStats => _stats.locations;
+
+  List<RebuildLocation>? get rebuildsForLastFrame =>
+      _rebuildsForFrame.values.lastOrNull;
```
kenzieschmoll, line 196: is .values guaranteed to be in the order that items were added to the map? is this the order we want?

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
```diff
@@ -20,112 +21,297 @@ import '../../primitives/utils.dart';
...
+    processRebuildsForFrame(json);
+
+    // We've updated the build counts and possibly the locations.
+    _stats.dataChanged();
+  }
+
+  void clearFromRestart() {
+    _rebuildsForFrame.clear();
+    locationMap.clear();
+    _stats.clear();
```
kenzieschmoll, line 239: nit: rewrite as follows to reduce duplication
```
locationMap.clear();
clearAllCounts();
```

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
```diff
@@ -20,112 +21,297 @@ import '../../primitives/utils.dart';
...
-    _locations.notifyListeners();
-  }
+  bool get isResolved => path != null;
 }
 
 class RebuildLocation {
-  RebuildLocation({this.id, this.path, this.line, this.column, this.name});
+  RebuildLocation(this.location, {this.buildCount = 0});
+  Location location;
+  int buildCount;
```
kenzieschmoll, line 284: should these be final?

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
```diff
@@ -20,112 +21,297 @@ import '../../primitives/utils.dart';
...
 
-  void clearFromReload() {
-    // TODO(devoncarew): We need to call this when we see a hot reload or restart.
+class Location {
+  Location({required this.id, this.path, this.line, this.column, this.name});
 
-    _locationMap.clear();
-    _locations.clear();
-  }
+  int id;
```
kenzieschmoll, line 270: should this be final?

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
```diff
@@ -20,112 +21,297 @@ import '../../primitives/utils.dart';
...
+  int buildCount;
+}
 
-  final int? id;
-  final String? path;
-  final int? line;
-  final int? column;
-  final String? name;
+/// Helper class to merge together rebuild metrics from multiple sources.
+/// For example, rebuild metrics on a per frame basis and total metrics since
```
kenzieschmoll, line 288: nit: new line after first sentence of dart doc

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
```diff
@@ -20,112 +21,297 @@ import '../../primitives/utils.dart';
...
+/// the last route change.
+class RebuildLocationMetrics {
+  RebuildLocationMetrics(this.location, {required int numMetrics})
+      : buildCounts = List<int>.filled(numMetrics, 0);
+  Location location;
+  List<int> buildCounts;
```
kenzieschmoll, line 294: final?

packages/devtools_app/lib/src/screens/performance/tabbed_performance_view.dart
```diff
@@ -82,8 +87,20 @@ class _TabbedPerformanceViewState extends State<TabbedPerformanceView>
       ),
     );
 
+    final rebuildMetrics = RebuildMetricsView(
```
kenzieschmoll, line 90: super nit: should we rename to rebuildStats and RebuildStatsView to have consistent naming with feature flag and other names?

packages/devtools_app/lib/src/screens/performance/tabbed_performance_view.dart
```diff
@@ -119,22 +132,6 @@ class _TabbedPerformanceViewState extends State<TabbedPerformanceView>
     final hasData = data != null && !data.isEmpty;
     final searchFieldEnabled = hasData && !widget.processing;
     return [
-      _buildTab(
-        tabName: 'Timeline Events',
```
kenzieschmoll, line 123: moving this tab to no longer be the default seems like a change we should consult UXR on @InMatrix. I would thing at this point, our users expect to see the timeline trace when they open the perf page. I guess the question we need to ask is, what do we want our users to be looking at by default?

packages/devtools_app/lib/src/shared/future_work_tracker.dart
```diff
@@ -0,0 +1,45 @@
...
+
+import 'package:flutter/foundation.dart';
+
+/// Class that tracks whether work defined by when futures complete is still in
+/// progress.
+///
+/// Work is added by calling [track] and completed when the [Future]s tracked
+/// complete or when [clear] is called.
+class FutureWorkTracker {
+  final Set<Future<void>> _inProgress = <Future<void>>{};
```
kenzieschmoll, line 15: remove type on LHS

packages/devtools_app/lib/src/shared/future_work_tracker.dart
```diff
@@ -0,0 +1,45 @@
...
+
+  /// ValueNotifier that returns where any of the futures added since last
```
kenzieschmoll, line 17: nit: r/where/whether

packages/devtools_app/lib/src/screens/performance/panes/frame_analysis/frame_analysis.dart
```diff
@@ -48,6 +56,36 @@ class FlutterFrameAnalysisView extends StatelessWidget {
...
+              final rebuilds =
+                  rebuildCountModel.rebuildsForFrame(frameAnalysis.frame.id);
+              if (!state.enabled && rebuilds == null) {
+                // No need to explain why there are no rebuilds as rebuild
+                // tracking is disabled.
+                return const SizedBox();
+              }
+              if (rebuilds == null) {
+                return const Text(
+                    'Rebuild information not available for frame');
```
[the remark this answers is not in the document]
CoderDake: nit: nit: `for this frame.`

packages/devtools_app/lib/src/screens/performance/panes/controls/enhance_tracing/enhance_tracing_controller.dart
```diff
@@ -10,6 +10,7 @@ import '../../../../../shared/globals.dart';
 import 'enhance_tracing_model.dart';
 
 final enhanceTracingExtensions = [
+  extensions.trackRebuildWidgets,
```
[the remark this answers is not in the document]
jacob314: Added to more debugging options for now. Also fine with just surfacing it on the rebuilds tab.

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
```diff
@@ -2,15 +2,16 @@
 // Use of this source code is governed by a BSD-style license that can be
```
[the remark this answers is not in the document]
jacob314: done

packages/devtools_app/lib/src/screens/performance/performance_screen.dart
```diff
@@ -163,23 +161,7 @@ class PerformanceScreenBodyState extends State<PerformanceScreenBody>
...
-              ? tabbedPerformanceView
-              : Split(
-                  axis: Axis.vertical,
-                  initialFractions: const [0.7, 0.3],
-                  children: [
-                    tabbedPerformanceView,
-                    ValueListenableBuilder<TimelineEvent?>(
-                      valueListenable: controller.selectedTimelineEvent,
-                      builder: (context, selectedEvent, _) {
-                        return EventDetails(selectedEvent);
```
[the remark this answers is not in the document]
jacob314: done. moved to the timeline events tab where the selected timeline event makes more sense.

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
jacob314: needed due to named constructor.

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
jacob314: what is the advantage of using a factory constructor? this is a little more terse for this case.

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
jacob314: done

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
jacob314: done

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
jacob314: making it a factory doesn't remove the need for an empty constructor and I think a static method makes this uglier.

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
jacob314: yep. because of the kind of map used.

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
jacob314: done although reversed order to be idiomatic so future changes to clearAllCounts don't run into trouble because the locationMap was cleared first.

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
jacob314: done here and elsewhere. A few things were final but I'd missed marking them. @srawlins this is a case where I'd love whole package level lints to warn me that I had fields that could be final but I'd forgotten to mark final.

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
```diff
@@ -20,112 +21,297 @@ import '../../primitives/utils.dart';
...
+      final columns = entries['columns']!.cast<int>();
+      final names = entries['names']!.cast<String>();
+
+      for (var i = 0; i < ids.length; i++) {
+        final id = ids[i];
+        final existing = _locationMap[id];
+        if (existing != null) {
+          if (existing.path == null) {
+            // We added a placeholder entry to the _locationsMap before.
+            // Update it to reflect the actual location.
```
CoderDake, line 135: this comment seems out of place, does it belong here?

packages/devtools_app/lib/src/screens/performance/tabbed_performance_view.dart
```diff
@@ -64,11 +65,15 @@ class _TabbedPerformanceViewState extends State<TabbedPerformanceView>
...
     late Widget frameAnalysisView;
+    if (controller.data == null) {
+      return const SizedBox();
+    }
     final selectedFrame = _selectedFlutterFrame;
     if (selectedFrame != null) {
       frameAnalysisView = FlutterFrameAnalysisView(
         frameAnalysis: selectedFrame.frameAnalysis,
         enhanceTracingController: controller.enhanceTracingController,
+        rebuildCountModel: controller.data!.rebuildCountModel,
```
CoderDake, line 76: The 
```
if (controller.data == null) {
      return const SizedBox();
    }
    ```
    Code makes it seem like `controller.data` may be null sometimes.
    It seems like `!` here may end up crashing for us sometime as a result.

File: packages/devtools_app/test/scenes/performance/default.dart [hunk 24]
```diff
@@ -0,0 +1,61 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be found
+// in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:devtools_app/devtools_app.dart';
```
CoderDake, line 9: extra devtools_app import?

packages/devtools_app/lib/src/screens/performance/panes/frame_analysis/frame_analysis.dart
```diff
@@ -48,6 +56,36 @@ class FlutterFrameAnalysisView extends StatelessWidget {
...
+                serviceManager.serviceExtensionManager.getServiceExtensionState(
+              extensions.trackRebuildWidgets.extension,
+            ),
+            builder: (context, state, _) {
+              final rebuilds =
+                  rebuildCountModel.rebuildsForFrame(frameAnalysis.frame.id);
+              if (!state.enabled && rebuilds == null) {
+                // No need to explain why there are no rebuilds as rebuild
+                // tracking is disabled.
+                return const SizedBox();
```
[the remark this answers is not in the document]
InMatrix: Do we need to communicate any side-effect of turning this option on? Does this make the app slow?

packages/devtools_app/lib/src/screens/performance/tabbed_performance_view.dart
InMatrix: My general thought is that tabs offering higher level information should be shown first. One question that can potentially help us figure out how prominent the Timeline Events tab still needs to be is: _what kinds of performance problems users still have to use the Timeline Events tab, instead of any of the other three tabs, to understand?_

InMatrix:
What is the "Overall" column of the Rebuild Stats tab showing? Are those counts since the last reload or app start?

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
jacob314: done

packages/devtools_app/lib/src/screens/performance/tabbed_performance_view.dart
jacob314: done. removed all uses of the word metric[s].

packages/devtools_app/lib/src/screens/performance/tabbed_performance_view.dart
jacob314: I think users should only go to the timeline tab after they have verified their problem cannot be solved by another tab.
For example, if you have a raster issue, use the frame analysis tab before even thinking about looking at the timeline tab. If you have a build problem, make sure you aren't just building things you didn't expect before drilling into the timeline tab.

packages/devtools_app/lib/src/shared/future_work_tracker.dart
jacob314: done

packages/devtools_app/lib/src/screens/performance/panes/widget_rebuild.dart
```diff
@@ -0,0 +1,260 @@
...
+  void didUpdateWidget(covariant RebuildMetricsView oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    cancelListeners();
+    _initListeners();
+  }
+
+  void _initListeners() {
+    addAutoDisposeListener(widget.model.locationStats, computeMetrics);
+    addAutoDisposeListener(widget.selectedFrame, computeMetrics);
+    computeMetrics();
```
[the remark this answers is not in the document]
jacob314: this needs to be called in both didUpdateWidget and initState as didUpdateWidget may not be called. @kenzieschmoll do we have a good convention for the name of the private helper typically called from both initState and didUpdateWidget.

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
```diff
@@ -20,112 +21,297 @@ import '../../primitives/utils.dart';
...
-        final ids = entries['ids']!.cast<int>();
-        final lines = entries['lines']!.cast<int>();
-        final columns = entries['columns']!.cast<int>();
-        final names = entries['names']!.cast<String>();
-
-        for (var i = 0; i < ids.length; i++) {
-          final location = RebuildLocation(
-            id: ids[i],
-            path: file,
+  void incrementCount(Location location, int count) {
```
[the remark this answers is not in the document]
jacob314: Added a comment. It is a little low value but maybe it is helpful.

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
jacob314: cleaned a little. It belongs there but was unclear.

packages/devtools_app/lib/src/screens/performance/tabbed_performance_view.dart
jacob314: cleaned up code so no `!` were required. that helped fix one case that might have NPE

packages/devtools_app/test/scenes/performance/default.dart
jacob314: it is needed. generally I'd just trust our lints for unneeded imports to do their job and not comment on extra imports in a code review. I tested and in this case a bunch of symbols are unresolved if this import is removed.

packages/devtools_app/lib/src/screens/performance/panes/widget_rebuild.dart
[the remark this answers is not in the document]
kenzieschmoll: in several places, we just use `_init`. You could rename `_initListeners` to `_init` here to be more broad. Also, we need to call `cancelListeners` before the calls to `addAutoDisposeListener`

packages/devtools_app/test/scenes/performance/default.dart
kenzieschmoll: I think Dan was referring to the fact that we try to import files directly instead of importing devtools_app.dart, which exports a ton of files in DevTools

packages/devtools_app/lib/src/screens/performance/panes/widget_rebuild.dart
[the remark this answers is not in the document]
jacob314: I'm calling cancelListeners but only in the didUpdateWidget case as the cancel is not needed in initState.

packages/devtools_app/lib/src/screens/performance/panes/frame_analysis/frame_analysis.dart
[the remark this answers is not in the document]
jacob314: Updated the message.
The performance impact is small enough and only debug mode so probably not. This is considerably cheaper than other widget tracing (less chatty protocol) and debug mode only.

packages/devtools_app/lib/src/screens/performance/panes/frame_analysis/frame_analysis.dart
```diff
@@ -48,6 +56,36 @@ class FlutterFrameAnalysisView extends StatelessWidget {
...
+              }
+              if (rebuilds == null) {
+                return const Text(
+                    'Rebuild information not available for frame');
+              } else if (rebuilds.isEmpty) {
+                return const Text(
+                  'No rebuilds of widgets directly created by the project occurred.',
```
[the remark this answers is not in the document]
jacob314: done.

packages/devtools_app/lib/src/screens/performance/panes/frame_analysis/frame_analysis.dart
```diff
@@ -48,6 +56,36 @@ class FlutterFrameAnalysisView extends StatelessWidget {
...
+                );
+              } else {
+                return Expanded(
```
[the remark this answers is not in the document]
jacob314: refactored this a bit. unfortunately this can't really go in a scrollview because the rebuild table also needs to scroll and making that all work well seems more trouble than it is worth. I've added a todo to worry about this if we get more frame hints.

packages/devtools_app/lib/src/screens/performance/panes/widget_rebuild.dart
```diff
@@ -0,0 +1,260 @@
...
+      [_widgetColumn, ..._metricsColumns, _locationColumn];
+
+  @override
+  Widget build(BuildContext context) {
+    final borderSide = defaultBorderSide(Theme.of(context));
+    return Container(
+      decoration: BoxDecoration(
+        border: Border(right: borderSide),
+      ),
+      child: FlatTable<RebuildLocationMetrics>(
```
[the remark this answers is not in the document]
jacob314: merged.

packages/devtools_app/lib/src/screens/performance/panes/widget_rebuild.dart
```diff
@@ -0,0 +1,260 @@
...
+        _columnCache[name] = cached;
+      }
+      columns.add(cached);
+    }
+    return columns;
+  }
+
+  final ColumnData<RebuildLocationMetrics> _widgetColumn = _WidgetColumn();
+
+  final ColumnData<RebuildLocationMetrics> _locationColumn = _LocationColumn();
```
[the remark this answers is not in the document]
jacob314: done

packages/devtools_app/lib/src/screens/performance/panes/widget_rebuild.dart
```diff
@@ -0,0 +1,260 @@
...
+  final List<String> metricNames;
+  final List<RebuildLocationMetrics> metrics;
+
+  @override
+  State<RebuildTable> createState() => _RebuildTableState();
+}
+
+class _RebuildTableState extends State<RebuildTable> {
+  SortDirection sortDirection = SortDirection.descending;
+  int sortMetricColumnIndex = 0;
```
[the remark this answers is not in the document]
jacob314: done

packages/devtools_app/lib/src/screens/performance/panes/widget_rebuild.dart
```diff
@@ -0,0 +1,260 @@
...
+  final ValueListenable<FlutterFrame?> selectedFrame;
+
+  @override
+  State<RebuildMetricsView> createState() => _RebuildMetricsViewState();
+}
+
+class _RebuildMetricsViewState extends State<RebuildMetricsView>
+    with AutoDisposeMixin {
+  List<String> metricNames = const [];
+  List<RebuildLocationMetrics> metrics = const [];
```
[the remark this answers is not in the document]
jacob314: done

packages/devtools_app/lib/src/screens/performance/panes/frame_analysis/frame_analysis.dart
```diff
@@ -48,6 +56,36 @@ class FlutterFrameAnalysisView extends StatelessWidget {
           Expanded(
             child: FrameTimeVisualizer(frameAnalysis: frameAnalysis),
           ),
+          ValueListenableBuilder<ServiceExtensionState>(
```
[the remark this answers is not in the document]
jacob314: maybe not. this is a bit redundant. leaving in for now unless you have a strong preference to remove it.

packages/devtools_app/lib/src/screens/performance/performance_controller.dart
```diff
@@ -229,8 +234,47 @@ class PerformanceController extends DisposableController
               frame.enhanceTracingState = enhanceTracingController.tracingState;
             }
             addFrame(frame);
-          } else if (event.extensionKind == 'Flutter.RebuiltWidgets') {
-            rebuildCountModel.processRebuildEvent(event.extensionData!.data);
+          } else if (event.extensionKind == 'Flutter.RebuiltWidgets' &&
+              FeatureFlags.widgetRebuildstats) {
+            if (_currentRebuildWidgetsIsolate != event.isolate) {
+              data!.rebuildCountModel.clearFromRestart();
+              _currentRebuildWidgetsIsolate = event.isolate;
```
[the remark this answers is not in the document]
jacob314: doh. done

packages/devtools_app/lib/src/screens/performance/performance_controller.dart
```diff
@@ -229,8 +234,47 @@ class PerformanceController extends DisposableController
               frame.enhanceTracingState = enhanceTracingController.tracingState;
             }
             addFrame(frame);
-          } else if (event.extensionKind == 'Flutter.RebuiltWidgets') {
-            rebuildCountModel.processRebuildEvent(event.extensionData!.data);
+          } else if (event.extensionKind == 'Flutter.RebuiltWidgets' &&
+              FeatureFlags.widgetRebuildstats) {
+            if (_currentRebuildWidgetsIsolate != event.isolate) {
+              data!.rebuildCountModel.clearFromRestart();
```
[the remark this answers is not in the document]
jacob314: done

packages/devtools_app/lib/src/screens/performance/performance_controller.dart
```diff
@@ -229,8 +234,47 @@ class PerformanceController extends DisposableController
...
+            // TODO(jacobr): need to make sure we don't get events from before
+            // the last hot restart. Their data would be bogus.
+            data!.rebuildCountModel
+                .processRebuildEvent(event.extensionData!.data);
+            if (data!.rebuildCountModel.locationMap.locationsResolved.value ==
+                    false &&
+                !_fetchMissingLocationsStarted) {
+              // Some locations are missing. This can occur if rebuilds were
+              // enabled before DevTools connected. Rebuild events only occur
+              // lcoations that had yet to be sent in an event.
```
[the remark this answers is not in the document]
jacob314: rephrased. not that happy with the comment.

packages/devtools_app/lib/src/screens/performance/performance_controller.dart
```diff
@@ -229,8 +234,47 @@ class PerformanceController extends DisposableController
...
+              _fetchMissingLocationsStarted = true;
```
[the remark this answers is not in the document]
jacob314: unfortunately that makes it sound like fetching missing locations has completed which is not the case.

packages/devtools_app/lib/src/screens/performance/performance_controller.dart
```diff
@@ -229,8 +234,47 @@ class PerformanceController extends DisposableController
...
+            }
+            _currentRebuildWidgetsIsolate = event.isolate;
+            // TODO(jacobr): need to make sure we don't get events from before
+            // the last hot restart. Their data would be bogus.
+            data!.rebuildCountModel
+                .processRebuildEvent(event.extensionData!.data);
+            if (data!.rebuildCountModel.locationMap.locationsResolved.value ==
+                    false &&
+                !_fetchMissingLocationsStarted) {
+              // Some locations are missing. This can occur if rebuilds were
```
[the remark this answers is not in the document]
jacob314: done

packages/devtools_app/lib/src/screens/performance/performance_controller.dart
jacob314: the thing that is unawaited is the ga.timeAsync call.
I don't think we should slow down the UI waiting for the analytics event to complete.

packages/devtools_app/lib/src/screens/performance/performance_controller.dart
jacob314: done

packages/devtools_app/lib/src/screens/performance/rebuild_counts.dart
jacob314: Done

packages/devtools_app/lib/src/shared/future_work_tracker.dart
jacob314: done

packages/devtools_app/test/scenes/performance/default.dart
jacob314: I'd like to push back on this. in test code I don't see anything wrong with just importing devtools_app. maybe if you were lib/src/ui/common importing devtools_app would be ugly but it seems like the cleanest way to surface this from a test.

jacob314:
PTAL

packages/devtools_app/lib/src/screens/performance/panes/rebuild_stats/rebuild_counts.dart
```diff
@@ -0,0 +1,325 @@
...
+  List<RebuildLocation>? rebuildsForFrame(int frameNumber) {
+    return _rebuildsForFrame[frameNumber];
+  }
+
+  Map<String, dynamic>? toJson() {
+    if (_rebuildsForFrame.isEmpty) {
+      // No need to encode data unless there were actually rebuilds reported.
+      return null;
+    }
+    final frames = <dynamic>[];
```
kenzieschmoll, line 216: should this have type `<Map<String, dynamic>>` instead of `<dynamic>`

packages/devtools_app/lib/src/screens/performance/performance_controller.dart
kenzieschmoll: if we don't await the call to `ga.timeAsync`, then we inherently will not await the `asyncOperation` `processTraceEventsHelper`, right? We need to await `processTraceEventsHelper`, which is called in the body of the `timeAsync` method.

packages/devtools_app/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -419,7 +419,10 @@
 					"$(PROJECT_DIR)/Flutter/ephemeral",
 				);
 				INFOPLIST_FILE = Info.plist;
-				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks";
+				LD_RUNPATH_SEARCH_PATHS = (
```
kenzieschmoll, line 422: are the changes to this file intentional?

kenzieschmoll (approved):
lgtm with a few comments. I am concerned about removing the await on the call to ga.timeAsync

kenzieschmoll:
Breaking out a couple pieces of this PR in https://github.com/flutter/devtools/pull/4629 so that I can build on top of them (since jacob is OOO for a few weeks). So FYI Jacob you'll want to merge master once that lands.

packages/devtools_app/lib/src/screens/performance/performance_controller.dart
jacob314: dropped this from the PR.

jacob314:
PTAL

packages/devtools_app/lib/src/screens/performance/performance_controller.dart
```diff
@@ -100,14 +106,57 @@ class PerformanceController extends DisposableController
...
+                !_fetchMissingLocationsStarted) {
+              _fetchMissingRebuildLocations();
+            }
           }
         }),
       );
     }
   }
 
+  void _fetchMissingRebuildLocations() async {
```
kenzieschmoll, line 131: can this be moved into the rebuild model? After the refactor I did recently, we are trying to keep code related to a single feature in that feature's controller. Maybe there is a use for a WidgetRebuildsController?

jacob314: Sending a quick followup CL to move.

packages/devtools_app/lib/src/screens/performance/performance_model.dart
```diff
@@ -278,14 +287,15 @@ class TimelineRowData {
...
-    TimelineEvent? selectedEvent,
-    double? displayRefreshRate,
-    CpuProfileData? cpuProfileData,
-    RasterStats? rasterStats,
+    required int? selectedFrameId,
+    required TimelineEvent? selectedEvent,
+    required double? displayRefreshRate,
+    required CpuProfileData? cpuProfileData,
+    required RasterStats? rasterStats,
+    required RebuildCountModel? rebuildCountModel,
```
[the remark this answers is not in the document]
jacob314: For methods like this making the parameters required avoids bugs.
For example, yes selectedFrameId can be null but it is a bug if you are reading this data from JSON and neglect to consider specifying the selectedFrameId

PR #4578, opened 2022-10-04
Use PointerInterceptor in dialogs to allow clicks over an iFrame

kenzieschmoll:
This allows dialogs to accept mouse events when the dialog is rendered over the top of the Perfetto iFrame.


CoderDake (approved):
NICE FIX!

PR #4579, opened 2022-10-04
Add Use Legacy Trace Viewer setting to the Performance page

kenzieschmoll:
Toggling this setting will switch between the legacy trace viewer and the perfetto trace viewer

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/performance_screen.dart
+++ b/packages/devtools_app/lib/src/screens/performance/performance_screen.dart
@@ -164,7 +164,7 @@ class PerformanceScreenBodyState extends State<PerformanceScreenBody>
             },
           ),
         Expanded(
-          child: FeatureFlags.embeddedPerfetto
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/screens/performance/performance_screen.dart
```diff
@@ -164,7 +164,8 @@ class PerformanceScreenBodyState extends State<PerformanceScreenBody>
             },
           ),
         Expanded(
-          child: FeatureFlags.embeddedPerfetto
+          child: FeatureFlags.embeddedPerfetto &&
```
CoderDake, line 167: can `useLegacyTraceViewer change here? Should we perhaps add a listenable builder to ensure a refresh
?

kenzieschmoll: done

CoderDake (approved):

PR #4582, opened 2022-10-05
Build and test DevTools from the latest flutter candidate branch

kenzieschmoll:
This PR:
- removes the `pinned` and `master` bot flavors
- removes obsolete job "packages"
- builds and tests DevTools against the latest flutter candidate branch (this version is created by the flutter --> g3 autoroller, so this version is guaranteed to be in g3
- adds a script to generate the latest flutter candidate branch
- modifies the update_flutter_sdk.sh script to use the latest flutter candidate branch and adds an option '--local' so a developer can update their local flutter sdk checkout at the same time.

Clean up
- Moves some code to `devtools_shared` so that it can be reused by the `tool/` package
- Removes some stale code related to updating goldens for the stable branch as well as a version validation command that is no longer necessary
- Migrate devtools_app code away from deprecated members


tool/lib/commands/latest_flutter_candidate.dart
```diff
@@ -0,0 +1,277 @@
...
+
+    // TODO(kenz): consider traversing pages properly instead of using a while
+    // loop.
+    // See https://docs.github.com/en/rest/guides/traversing-with-pagination
+    const maxPerPage = 100;
+    var requestPage = 0;
+    bool lastPageReceived = false;
+    while (!lastPageReceived) {
+      final uri = Uri.https(
+        '${auth}api.github.com',
```
CoderDake, line 43: i feel like this auth would need to be in a bearer token header

tool/build_release.sh
```diff
@@ -17,16 +17,16 @@ DEVTOOLS_DIR="${TOOL_DIR}/.."
 FLUTTER_DIR="`pwd`/flutter-sdk"
 PATH="$FLUTTER_DIR/bin":$PATH
 
-REQUIRED_FLUTTER_VERSION=$(<"${DEVTOOLS_DIR}/flutter-version.txt")
+REQUIRED_FLUTTER_VERSION=`dart $TOOL_DIR/bin/repo_tool.dart latest-flutter-candidate | tail -n 1`
```
CoderDake, line 20: is the `|tail - n 1 necessary` here? it looks like the command would only ever print one line

tool/lib/commands/latest_flutter_candidate.dart
```diff
@@ -0,0 +1,277 @@
...
+    // 2.15.0-233.0.dev (dev) (Mon Oct 18 14:06:26 2021 -0700) on "ios_x64"
+    // 2.15.0-178.1.beta
+    // 2.6.0-12.0.pre.443
+    //
+    // So split on the spaces to the version, and then on the dash char to
+    // separate the main semantic version from the pre release version.
+    final splitOnSpaces = versionString.split(' ');
+    final version = splitOnSpaces.first;
+    final splitOnDash = version.split('-');
+    assert(splitOnDash.length <= 2, 'version: $version');
```
CoderDake, line 133: it might be way more simple to parse this using a regex.
https://github.com/flutter/devtools/blob/5ccb2bd3594456efd37a36b58ea793695875c567/tool/update_version.dart#L63-L70
is a good example of how to pull things apart.
I believe the regex for a candidate branch would be something like:
final semVerMatch = RegExp(r'^flutter-(?<major>\d+)\.(?<minor>\d+)-candidate\.(?<candidate_minor>\d+)')
      .firstMatch(version);

CoderDake (commented):
I'm still reviewing some stuff here, but pushing up some comments early

tool/lib/commands/latest_flutter_candidate.dart
CoderDake: It looks like i missunderstood this. I thought this was trying to parse candidate branches based on context. I see that is done elsewhere now though

CoderDake: that being said this could definitly benefit from some regex named grouping. right now it is a bit hard to make sense of :sweat_smile:

tool/build_release.sh
kenzieschmoll: the command may have to call `flutter pub get`, in which case the stdout will get clogged up with all that output

tool/lib/commands/latest_flutter_candidate.dart
kenzieschmoll: I'm going to hold off on this for now. This code was copied from the devtools_app repo (I have a comment above to consider placing this in devtools_shared), but any fixes to the semantic version code should be applied to the original file (or a shared location in the future).

tool/lib/commands/latest_flutter_candidate.dart
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  final x = int.parse(match.group(1)!);
+  final y = int.parse(match.group(2)!);
+  final z = int.parse(match.group(3)!);
+
+  return SemanticVersion(major: x, minor: y, patch: z);
+}
+
+// TODO(kenz): consider placing this class in devtools_shared, as this is
```
CoderDake, line 101: definitely agree with this. similar behaviour is in lots of places now

tool/lib/commands/latest_flutter_candidate.dart
```diff
@@ -0,0 +1,277 @@
...
+      }
+
+      requestPage++;
+      if (branches.length < maxPerPage) {
+        lastPageReceived = true;
+      }
+    }
+
+    if (latestBranchName == null) {
+      print('Something went wrong. Could not find the latest candidate branch:'
```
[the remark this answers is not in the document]
kenzieschmoll: done

tool/lib/commands/latest_flutter_candidate.dart
kenzieschmoll: moved this to devtools_shared

tool/lib/commands/latest_flutter_candidate.dart
kenzieschmoll: you're right! this is what was causing the test flakes

.github/workflows/build.yaml
```diff
@@ -13,6 +13,9 @@ on:
 # Declare default permissions as read only.
 permissions: read-all
 
+env:
+  AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
CoderDake, line 17: NICE!

packages/devtools_shared/lib/src/utils/semantic_version.dart
```diff
@@ -0,0 +1,165 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:math' as math;
+
+import 'compare.dart';
+
+class SemanticVersion with CompareMixin {
```
CoderDake, line 9: ty for moving this to a common location!

tool/bots.sh
```diff
@@ -27,26 +27,29 @@ function flutter {
...
 export PATH=`pwd`/flutter-sdk/bin/cache/dart-sdk/bin:`pwd`/flutter-sdk/bin:`pwd`/bin:$PATH
 
+# Look up the latest flutter candidate (this is the latest flutter version in g3)
+# TODO(kenz): if we re-write this logic as a shell script, we won't have to incurr the cost
+# of building flutter tool twice. We need flutter and dart to use this repo_tool, so we have
+# to checkout flutter before we can call this script in its current form.
+echo "Looking up the latest Flutter candidate branch"
+pushd packages/devtools_app
+repo_tool latest-flutter-candidate --githubToken=$AUTH_TOKEN
+LATEST_FLUTTER_CANDIDATE=`repo_tool latest-flutter-candidate --githubToken=$AUTH_TOKEN | tail -n 1`
```
kenzieschmoll, line 46: will remove this duplicate call before submitting

tool/bots.sh
```diff
@@ -42,8 +42,8 @@ export PATH=`pwd`/flutter-sdk/bin/cache/dart-sdk/bin:`pwd`/flutter-sdk/bin:`pwd`
 # to checkout flutter before we can call this script in its current form.
 echo "Looking up the latest Flutter candidate branch"
 pushd packages/devtools_app
-LATEST_FLUTTER_CANDIDATE=`repo_tool latest-flutter-candidate | tail -n 1`
-echo "Latest Flutter candidate branch: $LATEST_FLUTTER_CANDIDATE"
+repo_tool latest-flutter-candidate --githubToken=$AUTH_TOKEN
```
CoderDake, line 45:
```suggestion
```
probably don't need this extra line now

CoderDake (approved):

PR #4584, opened 2022-10-06
Add flag to turn on experimental features in compile time.

polina-c:


packages/devtools_app/lib/src/primitives/feature_flags.dart
```diff
@@ -2,15 +2,27 @@
...
+///
+/// By default, the constant is false.
+/// To enable it, pass the compilation flag
+/// `--dart-define=enable_experiments=true`.
+///
+/// To enable the flag in debug configuration of VSCode, add value:
+///   "args": [
+///     "--dart-define=enable_experiments=true"
+///   ]
+const bool _kEnableExperiments = bool.fromEnvironment('enable_experiments');
```
kenzieschmoll, line 15: would this enable an end user to hack these experiments on? end user being someone running a flutter app and debugging with devtools

polina-c: It is compile time flag. If end user compiles devtools, then yes, otherwise - no.

Do we want them to hack?

kenzieschmoll: No we don't. Would there ever be an occasion where we don't want all the experiments enabled at once?

polina-c: Or we can give user build that was compiled with this flag.

kenzieschmoll (approved):

packages/devtools_app/lib/src/primitives/feature_flags.dart
polina-c: > No we don't. Would there ever be an occasion where we don't want all the experiments enabled at once?

I do not think it can happen. If it does, we will adjust the code.

PR #4585, opened 2022-10-06
Include the full path for Perfetto assets

kenzieschmoll:
The asset variant support does work for this use case on the latest flutter version.

### DIFF
--- a/tool/update_perfetto_assets.dart
+++ b/tool/update_perfetto_assets.dart
@@ -1,0 +1,73 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:io';
+
+void main(List<String> args) {
+  final mainDevToolsDirectory = Directory.current;
+  if (!mainDevToolsDirectory.path.endsWith('/devtools')) {
+    throw Exception(
+      'Please execute this script from your top level '
+      '\'devtools/\' directory. Running the `update_perfetto.sh` script will'
+      'automatically do this.',
+    );
+  }
+
+  final perfettoDistDir = Directory.fromUri(
+    Uri.parse(
+        '${mainDevToolsDirectory.path}/third_party/packages/perfetto_compiled/lib/dist'),
+  );
+
+  // Find the new perfetto version number.
+  String newVersionNumber = '';
+  final versionRegExp = RegExp(r'v[0-9]+[.][0-9]+-[0-9a-f]+');
+  final entities = perfettoDistDir.listSync();
+  for (FileSystemEntity entity in entities) {
+    final path = entity.path;
+    final match = versionRegExp.firstMatch(path);
+    if (match != null) {
+      newVersionNumber = path.split('/').last;
+      print('New Perfetto version: $newVersionNumber');
+      break;
+    }
+  }
+
+  if (newVersionNumber.isEmpty) {
+    throw Exception(
+      'Error updating Perfetto assets: could not find Perfetto version number '
+      'from entities: ${entities.map((e) => e.path).toList()}',
+    );
+  }
+
+  final pubspec = File.fromUri(
+    Uri.parse(
+        '${mainDevToolsDirectory.path}/packages/devtools_app/pubspec.yaml'),
+  );
+
+  // TODO(kenz): Ensure the pubspec.yaml contains an entry for each file in
+  // [perfettoDistDir].
+
+  final perfettoAssetRegExp = RegExp(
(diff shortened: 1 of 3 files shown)


tool/update_perfetto_assets.dart
```diff
@@ -0,0 +1,73 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:io';
+
+void main(List<String> args) {
+  final mainDevToolsDirectory = Directory.current;
+  if (!mainDevToolsDirectory.path.endsWith('/devtools')) {
```
CoderDake, line 9: To make this more flexible what if we require a DEVTOOLS_DIR env var?

That way we can do fun stuff like adding the tools dir to our PATHs and not have to worry about our location:D

tool/update_perfetto_assets.dart
```diff
@@ -0,0 +1,73 @@
...
+  final perfettoDistDir = Directory.fromUri(
+    Uri.parse(
+        '${mainDevToolsDirectory.path}/third_party/packages/perfetto_compiled/lib/dist'),
+  );
+
+  // Find the new perfetto version number.
+  String newVersionNumber = '';
+  final versionRegExp = RegExp(r'v[0-9]+[.][0-9]+-[0-9a-f]+');
+  final entities = perfettoDistDir.listSync();
+  for (FileSystemEntity entity in entities) {
```
CoderDake, line 26: Something about this feels like it may cause issues in the future. Is there a specific file we could focus on that would for sure have the version?

tool/update_perfetto_assets.dart
```diff
@@ -0,0 +1,73 @@
...
+
+  final perfettoAssetRegExp = RegExp(
+      r'.*packages\/perfetto_compiled\/dist\/v[0-9]+[.][0-9]+-[0-9a-f]+\/');
+  bool isPerfettoAssetLine(String line) => perfettoAssetRegExp.hasMatch(line);
+
+  final lines = pubspec.readAsLinesSync();
+  for (int i = 0; i < lines.length; i++) {
+    final line = lines[i];
+    if (isPerfettoAssetLine(line)) {
+      final parts = line.split('/').map(
```
CoderDake, line 59: to avoid manually pulling this apart, you can use name groups in the regexp
```
  final perfettoAssetRegExp = RegExp(
      r'(?<prefix>^.*packages\/perfetto_compiled\/dist\/)(?<version>v\d+[.]\d+-[0-9a-fA-F]+)(?<suffix>\/.*$)');
  bool isPerfettoAssetLine(String line) => perfettoAssetRegExp.hasMatch(line);
  final lines = pubspec.readAsLinesSync();
  for (int i = 0; i < lines.length; i++) {
    final line = lines[i];
    final match = perfettoAssetRegExp.firstMatch(line);
    if (match != null) {
      final prefix = match.namedGroup('prefix')!;
      final suffix = match.namedGroup('suffix')!;
      lines[i] = prefix + newVersionNumber + suffix;
    }
  }
  ```

File: tool/update_perfetto_assets.dart (hunk 2 above)
Comment (kenzieschmoll) [reply tool/update_perfetto_assets.dart:26]: It's a single folder:
Screen Shot 2022-10-06 at 11 28 11 AM

File: tool/update_perfetto_assets.dart (hunk 3 above)
Comment (kenzieschmoll) [reply tool/update_perfetto_assets.dart:59]: excellent!

File: tool/update_perfetto_assets.dart [hunk 4]
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  final perfettoDistDir = Directory.fromUri(
+    Uri.parse(
+        '${mainDevToolsDirectory.path}/third_party/packages/perfetto_compiled/lib/dist'),
+  );
+
+  // Find the new perfetto version number.
+  String newVersionNumber = '';
+  final versionRegExp = RegExp(r'v[0-9]+[.][0-9]+-[0-9a-f]+');
```
CoderDake, line 24:
```suggestion
  final versionRegExp = RegExp(r'v\d+[.]\d+-[0-9a-fA-F]+');
```

tool/update_perfetto_assets.dart
CoderDake: ahhhh it is a filename! Sounds good 😄

tool/update_perfetto_assets.dart
kenzieschmoll: discussed offline - we can optimize this in the future if needed. This script should never really be called by a human anyway - it should be called by the update_perfetto.sh script

CoderDake (approved):

PR #4586, opened 2022-10-06
CSV and UX for memory diffing.

polina-c:
1. Implement CSV download
2. Some UX improvements based on  [go/devtools-memory-diff-ux-research](https://goto.google.com/devtools-memory-diff-ux-research) and on common sense


packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
```diff
@@ -126,6 +126,20 @@ abstract class ExportController {
     return '${prefix}_$timestamp$postfix.$type';
   }
 
+  /// Downloads a file with [contents] and pushes notification about success.
+  void downloadAndNotify(
```
kenzieschmoll, line 130: instead of creating a new method, can we add a parameter 'notify' to downloadFile, and then conditionally call `notificationService.push(successfulExportMessage(file));`?

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_details.dart
```diff
@@ -119,34 +122,49 @@ class _RetainedSizeColumn extends ColumnData<StatsByPathEntry> {
...
-  static final _shallowSizeColumn = _ShallowSizeColumn();
-  static late final List<ColumnData<StatsByPathEntry>> _columns =
-      <ColumnData<StatsByPathEntry>>[
-    _RetainingPathColumn(),
-    _InstanceColumn(),
-    _shallowSizeColumn,
-    _RetainedSizeColumn(),
-  ];
+  static final _columns = <String, _RetainingPathTableColumns>{};
+  static _RetainingPathTableColumns _obtainColumns(
```
kenzieschmoll, line 151: naming nit: r/obtain/generate

packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -245,16 +245,17 @@ class FlatTableState<T> extends State<FlatTable<T>> with AutoDisposeMixin {
     );
   }
 
-  Widget _buildRow({
+  Widget? _buildRow({
```
kenzieschmoll, line 248: this doesn't feel safe. We shouldn't be returning null for a row in the table. This looks like a bandaid for some exception rather than a fix for the root cause. `index` should never be greater than `data.length` so whatever is causing that to happen is the bug we need to fix.

packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
polina-c: downloadFile is overriden for each platform, so created one more method to avoid duplicating this logic many times.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_details.dart
polina-c: I agree 'obtain' is not the best name, however 'generate' sounds confusing to me as it implies the method will always create new columns, while it is not true. Renamed to '_columns'.

packages/devtools_app/lib/src/shared/table/table.dart
polina-c: As the flutter's TableView is on the way, are you sure we want to heavily invest into maintaining our table?

This is caused by shrinking the dataset:
1. Scroll to the end in the table of retaining paths
2. Switch to other class or snapshot and scroll to the end again
3. Eventually you will get to smaller list of the retaining paths, that will cause the error

polina-c: removed nullability as discussed offline

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -100,6 +102,28 @@ class DiffPaneController extends DisposableController {
...
+  void downloadCurrentItemToCsv() {
+    final classes = derived.heapClasses.value!;
+    final item = core.selectedItem as SnapshotInstanceItem;
+    final diffWith = item.diffWith.value;
+
+    late String filePrefix;
+    if (diffWith == null) {
+      filePrefix = item.name;
+    } else {
+      filePrefix = '${item.name}-${diffWith.name}';
```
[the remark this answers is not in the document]
polina-c: It is just prefix. The download functionality will add timestamp.

packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
kenzieschmoll: You can put this in the superclass method and then annotate with mustCallSuper so that subclasses have to call `super.downloadFile`

packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -32,7 +32,7 @@ import 'table_data.dart';
 /// height of 0 and a height of [defaultRowHeight].
 double get defaultRowHeight => scaleByFontFactor(32.0);
 
-typedef IndexedScrollableWidgetBuilder = Widget Function({
+typedef IndexedScrollableWidgetBuilder = Widget? Function({
```
kenzieschmoll, line 35: remove nullability

packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -901,7 +901,7 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
     return tableWidth;
   }
 
-  Widget _buildItem(BuildContext context, int index, {bool isPinned = false}) {
+  Widget? _buildItem(BuildContext context, int index, {bool isPinned = false}) {
```
kenzieschmoll, line 904: remove nullability

kenzieschmoll (approved):

packages/devtools_app/lib/src/config_specific/import_export/import_export.dart
polina-c: Then we will need to also add 'mustBeOverridden' and it will be not very clear what is already implemented in super and what is not yet.

I separated methods instead: 
- saveFile just saves file and is different for each platform
- downloadFile wraps saveFile with additional logic that is platform agnostic

PR #4589, opened 2022-10-07
Add case study for memory diffing.

polina-c:


kenzieschmoll (approved):
looks good - one question is do we really need all the platforms? (web, linux, windows, etc.). If not we can remove the ones we don't care about

polina-c:
> looks good - one question is do we really need all the platforms? (web, linux, windows, etc.). If not we can remove the ones we don't care about

I think we may want to experiment with other platforms in future. So, I would keep all of them for now as they do not hurt. If they start bringing maintenance toil, we will delete them.

PR #4599, opened 2022-10-11
Copyable classes and paths in diff.

polina-c:


packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -2501,3 +2502,40 @@ class ToCsvButton extends StatelessWidget {
     );
   }
 }
+
+class CopyButton extends StatelessWidget {
```
kenzieschmoll, line 2506: we already have a common widget for the copy control. `CopyToClipboardControl` displays a notification upon success too. See if we can reuse that.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_control_pane.dart
```diff
@@ -36,17 +37,39 @@ class SnapshotControlPane extends StatelessWidget {
...
+                    ValueListenableBuilder<StatsByPathEntry?>(
+                      valueListenable: controller.derived.selectedPathEntry,
+                      builder: (_, pathEntry, ___) => CopyButton(
+                        tooltip: 'Copy the selected retaining path.',
+                        contentBuilder: pathEntry == null
+                            ? null
+                            : () => pathEntry.key.asLongString(),
+                        label: 'Path',
+                      ),
+                    ),
```
kenzieschmoll, line 72: instead of adding these copy controls to the button bar, I think we should mimc what is done on the network page. This will help with consistency and will make the action bar less busy.
Screen Shot 2022-10-10 at 9 29 26 PM
Screen Shot 2022-10-10 at 9 29 31 PM

PR #4600, opened 2022-10-11
Tooltips for group columns.

polina-c:


packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -1521,7 +1521,9 @@ class _TableRowState<T> extends State<TableRow<T>>
           switch (displayTypeForIndex) {
             case _TableRowPartDisplayType.column:
               final current = columnIndexTracker;
-              columnIndexTracker++;
+              if (columnIndexTracker + 1 < widget.columns.length) {
+                columnIndexTracker++;
+              }
```
kenzieschmoll, line 1526: why would we ever have more items in `rowDisplayParts` with type `_TableRowPartDisplayType.column` than we have columns? that seems like a bug with how `rowDisplayParts` is being generated

packages/devtools_app/lib/src/shared/table/table_data.dart
```diff
@@ -113,12 +113,14 @@ mixin PinnableListEntry {
 /// will be drawn between groups and an additional header row will be added to
 /// the table to display the column group titles.
 class ColumnGroup {
-  ColumnGroup({required this.title, required this.range});
+  ColumnGroup({this.tooltip, required this.title, required this.range});
```
kenzieschmoll, line 116: nit: put non-required item at the end of the parameter list

packages/devtools_app/lib/src/shared/table/table.dart
polina-c: it seems to be side effect of scroll overflow

kenzieschmoll: I dug into this a bit. Fix for this bug is here: https://github.com/flutter/devtools/pull/4606

kenzieschmoll (approved):
you'll have to merge and resolve the conflicts (i.e. remove this PR's changes to columnIndexTracker), but LGTM after that

PR #4608, opened 2022-10-14
Make class name copyable by adding button to the cell.

polina-c:
Replacement for https://github.com/flutter/devtools/pull/4599.

Screen Shot 2022-10-13 at 11 57 40 PM

Response for the [comment](https://github.com/flutter/devtools/pull/4599#discussion_r991740650): context menu does not look discoverable, because most texts/tables in DevTools do not support the context menu.


packages/devtools_app/lib/src/screens/memory/shared/widgets/heap_class_view.dart
```diff
@@ -0,0 +1,34 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/widgets.dart';
+
+import '../../../../../devtools_app.dart';
+import '../heap/model.dart';
+
+class HeapClassView extends StatelessWidget {
```
kenzieschmoll, line 10: Does this need to be in its own file? I worry about starting to feel too java-like where every class is in its own file and things start to feel "overly" abstracted. If the concern is that this class is used in multiple places, maybe it makes sense to create a `shared_widgets` file in the memory code where all widgets can go that are shared in multiple panes

kenzieschmoll:
Can we make the copy icon show up only for the selected row? Having a copy icon in each row is a little busy

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
```diff
@@ -39,7 +41,16 @@ class _ClassNameColumn extends ColumnData<DiffClassStats> {
   bool get supportsSorting => true;
 
   @override
-  String getTooltip(DiffClassStats classStats) => classStats.heapClass.fullName;
+  String getTooltip(DiffClassStats classStats) => '';
```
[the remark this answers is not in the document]
polina-c: We are removing the tooltip, because it is provided by [HeapClassView].
Added comment.

polina-c:
> Can we make the copy icon show up only for the selected row? Having a copy icon in each row is a little busy

good idea
done

packages/devtools_app/lib/src/screens/memory/shared/shared_memory_widgets.dart
```diff
@@ -4,13 +4,18 @@
...
-import '../heap/model.dart';
+import '../../../../devtools_app.dart';
+import 'heap/model.dart';
 
 class HeapClassView extends StatelessWidget {
-  const HeapClassView({super.key, required this.theClass});
+  const HeapClassView({
+    super.key,
+    required this.theClass,
+    this.showCopyButton = true,
```
kenzieschmoll, line 14: nit: the default should probably be false here

kenzieschmoll (commented):
comment regarding the golden image update: can you make one of the classes selected in the test so that we can see the copy icon only on the selected row?

kenzieschmoll (approved):
2 comments - then lgtm

polina-c:
> comment regarding the golden image update: can you make one of the classes selected in the test so that we can see the copy icon only on the selected row?

good idea. done

PR #4613, opened 2022-10-16
Steps to run the app.

polina-c:


tool/update_flutter_sdk.sh
```diff
@@ -4,21 +4,30 @@
...
 # Any subsequent commands failure will cause this script to exit immediately
+set -e
+
+GLOBAL_PATH=$1
+UPDATE_LOCALLY=$2
+
+# We need path to `dart` to be included.
+# Global path is not available here, so we are passing it as parameter.
+# https://unix.stackexchange.com/questions/704630/path-in-script-different-than-actual-path
+export PATH="$GLOBAL_PATH:$PATH"
```
kenzieschmoll, line 16: I'm confused why this is not required on my machine, but is on yours? We should have some more members of the team try this update_flutter_sdk script, because maybe there is an environment setup step that we should specify in our contributing instructions instead of making users of this script have to pass in more parameters.

polina-c: I use zsh that is default on mac. Do you use bash or zsh?

CoderDake: Polina and I looked into this and it looks like this may have just been a side effect of using export, and the PATH of her terminal getting a bit messed up as a result.

CoderDake (approved):

PR #4616, opened 2022-10-18
Visualize retaining path.

polina-c:


polina-c:
https://user-images.githubusercontent.com/12115586/196534926-f1ce02a4-2405-4b98-8ef3-40c51057df5c.mov

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_details/class_details.dart
```diff
@@ -28,10 +35,35 @@ class HeapClassDetails extends StatelessWidget {
...
+          path: selection.key,
+          controller: pathController,
+        );
+      },
+    );
+
+    return Split(
+      axis: Axis.horizontal,
+      initialFractions: const [0.7, 0.3],
+      children: [area1, area2],
```
bkonyi, line 66: Nit: rename these variables to something more descriptive. Maybe `retainingPathView` and `statsView`?

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_details/path.dart
```diff
@@ -0,0 +1,108 @@
...
+      children: [
+        const SizedBox(height: densePadding),
+        _PathControlPane(
+          controller: controller,
+          path: path,
+        ),
+        Expanded(
+          child: Padding(
+            padding:
+                const EdgeInsets.only(top: densePadding, left: densePadding),
```
bkonyi, line 35: Nit: trailing comma after `densePadding`.

packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
```diff
@@ -126,10 +126,69 @@ class ClassOnlyHeapPath {
             heapPath.objects.map((o) => o.heapClass).toList(growable: false);
   final List<HeapClassName> classes;
 
-  String asShortString() => classes.map((e) => e.className).join('/');
+  String asShortString({String? delimiter, bool inverted = false}) => _asString(
```
bkonyi, line 129: Nit: consider using `toShortString()` and `toLongString()` instead.

packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
```diff
@@ -126,10 +126,69 @@ class ClassOnlyHeapPath {
...
+        ),
+        inverted: inverted,
+      );
+
+  String asLongString({
+    String? delimiter,
+    bool inverted = false,
+    bool hideStandard = false,
+  }) {
+    late List<String> data;
```
bkonyi, line 144: Does this need to be late? I would expect static analysis to have noticed that this is initialized unconditionally below.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/simple_controllers.dart
```diff
@@ -0,0 +1,10 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/material.dart';
+
+class RetainingPathController {
+  final hideStandard = ValueNotifier<bool>(true);
```
bkonyi, line 8: Nit: I generally try and avoid exposing `ValueNotifier`s directly. If these are just meant to be toggled, maybe use `ValueListenable` getters and add `toggleHideStandard()` and `toggleInvert()` methods?

bkonyi:
> Screen.Recording.2022-10-18.at.1.15.10.PM.mov

Thanks for the video! I do still think that the set of retaining paths for an instance might be better visualized as a tree, but I haven't had a chance to play around with your implementation quite yet.

packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
polina-c: wow! indeed
thanks, i did not expect it to be possible :)

polina-c:
> > Screen.Recording.2022-10-18.at.1.15.10.PM.mov
> 
> Thanks for the video! I do still think that the set of retaining paths for an instance might be better visualized as a tree, but I haven't had a chance to play around with your implementation quite yet.

You can add this suggestion to go/ux-improvements and link a doc or issue with details/screenshots. I hope to make the process around the doc alive at some moment.

PR #4617, opened 2022-10-18
Update manual test steps to remove UX friction points.

polina-c:


CONTRIBUTING.md
```diff
@@ -227,15 +227,17 @@ To explore DevTools with all experimental features enabled:
 
 1. [Configure](https://docs.flutter.dev/get-started/install) Dart or Flutter.
 
-2. Start DevTools:
+2. Create an empty folder and `cd` to it in terminal. 
+
+3. Start DevTools:
 ```
-git clone git@github.com:flutter/devtools.git
+git clone https://github.com/flutter/devtools.git
```
CoderDake, line 234: It might be worth leaving both options since both are vlid
```suggestion
- Clone with SSH
    - git clone git@github.com:flutter/devtools.git
- OR clone with https
    - git clone https://github.com/flutter/devtools.git
```

CONTRIBUTING.md
```diff
@@ -227,15 +227,17 @@ To explore DevTools with all experimental features enabled:
 
 1. [Configure](https://docs.flutter.dev/get-started/install) Dart or Flutter.
 
-2. Start DevTools:
+2. Create an empty folder and `cd` to it in terminal.
```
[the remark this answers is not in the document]
polina-c: I want to keep the steps simple and the scrip runnable without editing.
Professional engineers, who understand the script, will know how to deviate from it.
Updated instruction a little to add more flexibility.

CONTRIBUTING.md
polina-c: I want the script to be simple and not requiring choices and edits, so that UX people could get devtools up and running fast.
Professional engineers will know how to deviate if they want.
I am choosing http, because SSH requires configuration.

CONTRIBUTING.md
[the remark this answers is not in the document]
polina-c: Yes, the reason is potential name conflict. I faced it when I tested the script with Mindy.

CONTRIBUTING.md
```diff
@@ -221,21 +221,39 @@ To update goldens for all tests, run:
...
+
+```
+git pull origin master
+./tool/update_flutter_sdk.sh
+cd packages/devtools_app
+flutter pub upgrade
+../../tool/flutter-sdk/bin/flutter run -d chrome --dart-define=enable_experiments=true
+```
+
+2. Paste URL of your application (for example [Gallery](#connect-to-application)) to the connection textbox.
```
CoderDake, line 256: The `#connect-to-application` section mentioned here seems to be out of date since it mentions an observatory url.
We will probably want to update it to reflect the VM Service URL
Also the `#developer` section below it makes reference to the observatory URL so we should fix that too

CoderDake (commented):
The refresh section is a good addition, but I think it needs to be a bit more explicit. I've added a part that makes sure they are on the correct branch, before they perform the other  parts of the checkout at setup.

This will make sure that the reader won't get in a state where they are on the wrong branch or still have local changes while testing.

I've also left some suggestions about splitting up into a couple of sections to help the reader understand what is being accomplished.

CONTRIBUTING.md
```diff
@@ -227,15 +227,17 @@ To explore DevTools with all experimental features enabled:
 
 1. [Configure](https://docs.flutter.dev/get-started/install) Dart or Flutter.
 
-2. Start DevTools:
+2. In your terminal `cd` to a folder where you want to clone devtools, and that does not have subfolder `devtools` yet. It is ok if you already have devtools cloned somewhere else.
+
```
kenzieschmoll, line 233: why are we telling people to clone DevTools again? They should use their existing clone from the tip of master branch if they already have DevTools checked out. We should only tell a user to clone DevTools if they do not already have a local checkout

CONTRIBUTING.md
```diff
@@ -221,21 +221,43 @@ To update goldens for all tests, run:
...
+### Setup and start
+
 1. [Configure](https://docs.flutter.dev/get-started/install) Dart or Flutter.
 
-2. Start DevTools:
+2. In your terminal `cd` to a folder where you want to clone devtools, and that does not have subfolder `devtools` yet. 
+
+3. Start DevTools:
+```
+git clone https://github.com/flutter/devtools.git;
```
kenzieschmoll, line 237: I do not think we should point users to use https. We should default to SSH and then have a section about HTTPS if they do not have an SSH key configured.

CONTRIBUTING.md
```diff
@@ -221,21 +221,43 @@ To update goldens for all tests, run:
...
+git reset --hard origin/master;
+./tool/update_flutter_sdk.sh;
+cd packages/devtools_app;
+flutter pub upgrade;
 ../../tool/flutter-sdk/bin/flutter run -d chrome --dart-define=enable_experiments=true
 ```
 
-3. Paste URL of your application (for example [Gallery](#connect-to-application)) to the connection textbox.
+3. Paste the `VM Service URL` of a running application 
+    - For an example of running an application and grabbing the VM Service URL see the  [Connect To Application](#connect-to-application)) section.
```
kenzieschmoll, line 263: remove extra space after "see the"

CONTRIBUTING.md
polina-c: This script target users like Mindy. I want to give them a fastest way to kick off devtools. They do not have SSH configured.

CONTRIBUTING.md
```diff
@@ -221,21 +221,46 @@ To update goldens for all tests, run:
...
+### Refresh and start
+
+1. `cd` to the `devtools` directory created in the [Setup and start](#setup-and-start) section.
+
+2. Refresh and run DevTools (it will delete all your local changes!):
+
+```bash
+# Checkout the master branch and ensure it is at the most recent change:
+git checkout master;
+git reset --hard origin/master;
```
CoderDake, line 254:
```suggestion
git reset --hard origin/master; # this line will remove all local changes or commits on your branch
```

CoderDake (approved):

CONTRIBUTING.md
polina-c: Discussed offline: we use the same language as flutter console.

CONTRIBUTING.md
polina-c: Created other section for refresh.

PR #4620, opened 2022-10-19
Fix scrolling failure for the table.

polina-c:


packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -811,6 +806,25 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
...
+
+  @override
+  void didUpdateWidget(covariant _Table<T> oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    if (_data == oldWidget.tableController.tableData.value.data) return;
+    _initData();
+  }
+
+  void _initData() {
+    cancelListeners();
```
kenzieschmoll, line 819: cancelListeners doesn't need to be called in initState. Instead of including this line in the helper, add it to didUpdateWidget directly.

packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -811,6 +806,25 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
     scrollController = widget.tableController.verticalScrollController!;
 
     pinnedScrollController = ScrollController();
+  }
+
+  @override
+  void didUpdateWidget(covariant _Table<T> oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    if (_data == oldWidget.tableController.tableData.value.data) return;
```
kenzieschmoll, line 814: we need to check that `widget.tableController.tableData == oldWidget.tableController.tableData` here. We aren't checking for equality of the data, but rather equality of the ValueNotifier.

Also this early return isn't exactly accurate. This only matters for the `widget.tableController.tableData` listener. But the tableData notifier could be unchanged in theory and `widget.selectionNotifier` could have been changed, which would mean we'd need to call `_initScrollOnSelectionListener(widget.selectionNotifier);` still. 

For simplicity, we can change this if check to check all three conditions:

```
if (widget.tableController.tableData != oldWidget.tableController.tableData || widget.selectionNotifier != oldWidget.selectionNotifier || widget.activeSearchMatchNotifier != oldWidget.activeSearchMatchNotifier) {
  cancelListeners();
  // init all three listeners
}
```

packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -811,6 +806,31 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
...
+        widget.selectionNotifier != oldWidget.selectionNotifier ||
+        widget.activeSearchMatchNotifier != oldWidget.activeSearchMatchNotifier;
+
+    if (dataChanged) {
+      cancelListeners();
+      _initData();
+    }
+  }
+
+  void _initData() {
```
kenzieschmoll, line 826: nit: rename to _init, _initHelper, _initDataAndAddListeners since this is doing more than just initializing the data.

packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -811,6 +806,31 @@ class _TableState<T> extends State<_Table<T>> with AutoDisposeMixin {
     scrollController = widget.tableController.verticalScrollController!;
 
     pinnedScrollController = ScrollController();
+  }
+
+  @override
+  void didUpdateWidget(covariant _Table<T> oldWidget) {
+    super.didUpdateWidget(oldWidget);
+
+    final dataChanged = widget.tableController.tableData !=
```
kenzieschmoll, line 815: nit: rename to notifiersChanged since we are actually checking all the notifiers we listen to and not just table data

kenzieschmoll (approved):
a couple naming nits, but LGTM

PR #4621, opened 2022-10-19
Codeowners changes

CoderDake:
- Set @flutter/devtools-reviewers  as a codeowner of the whole repository, so that we can rotate on reviews
- Add specific owners for different sections of our code
    - We can make these sections more specific as necessary


kenzieschmoll (approved):
will be interesting to see this in use!

Issue #3861, opened 2022-03-15
DevTools CPU profiler pane functionality should be on-par with Observatory


bkonyi:
**[See Observatory -> DevTools Tracking Issue](https://github.com/flutter/devtools/issues/3858).**

Performance tooling functionality from Observatory that’s not present in DevTools:

- [x] Changing between code/function CPU profiles ([PR](https://github.com/flutter/devtools/pull/4622))
- [x] Call tree filtering
- [x] Tag ordering (VM, User, None) ([PR](https://github.com/flutter/devtools/pull/4693))
- [ ] Last refresh time
- [x] profiler stats (@kenzieschmoll) 
- [x] Go-to source from tree entry
- [x] Method table view (tracked at https://github.com/flutter/devtools/issues/841) (@kenzieschmoll)
- [x] closure display is incomplete (https://github.com/flutter/devtools/issues/5354)

**Observatory CPU Profile View**

![image](https://user-images.githubusercontent.com/24210656/158471844-6dc12d45-d801-4769-929e-8e86c59260f5.png)

**Observatory CPU Profile Method Table View**

![image](https://user-images.githubusercontent.com/24210656/158471712-91a253f8-9d27-4c92-9f44-3ed9cd79e3de.png)

kenzieschmoll:
tracking the last refresh time data at https://github.com/flutter/devtools/issues/5382. This is non critical and marked as such. Closing this issue now that the method table has been added.

PR #4622, opened 2022-10-19
Add "Code view" support to CPU Profiler

bkonyi:
The current CPU profiler view renders CPU profiles in terms of Dart functions. This is useful for the general Dart developer who is interested in determining where time is being spent during execution, but does not necessarily provide an accurate picture when profiling performance of generated code. This is because compiler developers are interested in the amount of time spent executing portions of generated code, which can include inlined functions.

This PR adds an option in the CPU Profiler tab to display the CPU profile without expanding inlined functions. When the "Code" view is enabled, instead of building the CPU profile with a function tree the tree is instead built based on `Code` objects from the VM.

This functionality is only enabled when VM developer mode is turned on, and the extra data required to compute the code-based CPU profile is only requested from the VM in this mode.

Related issue: https://github.com/flutter/devtools/issues/3861

image

### DIFF
--- a/packages/devtools_app/lib/src/screens/profiler/cpu_profile_controller.dart
+++ b/packages/devtools_app/lib/src/screens/profiler/cpu_profile_controller.dart
@@ -19,6 +19,16 @@ import 'cpu_profile_service.dart';
 import 'cpu_profile_transformer.dart';
 import 'profiler_screen.dart';
 
+enum CpuProfilerViewType {
+  function,
+  code;
+
+  @override
+  String toString() {
+    return this == function ? 'Function' : 'Code';
+  }
+}
+
 class CpuProfilerController
     with
         SearchControllerMixin<CpuStackFrame>,
@@ -84,8 +94,17 @@ class CpuProfilerController
   final _userTagFilter = ValueNotifier<String>(userTagNone);
 
   Iterable<String> get userTags =>
-      cpuProfileStore.lookupProfile(label: userTagNone)?.userTags ??
+      cpuProfileStore
+          .lookupProfile(
+            label: userTagNone,
+          )
+          ?.functionProfile
+          .userTags ??
       const <String>[];
+
+  ValueListenable<CpuProfilerViewType> get viewType => _viewType;
+  final _viewType =
+      ValueNotifier<CpuProfilerViewType>(CpuProfilerViewType.function);
 
   bool get isToggleFilterActive =>
       toggleFilters.any((filter) => filter.enabled.value);
@@ -153,19 +172,22 @@ class CpuProfilerController
 
     _processingNotifier.value = true;
 
-    var cpuProfileData = baseStateCpuProfileData;
+    var cpuProfiles = CpuProfilePair(
+      functionProfile: baseStateCpuProfileData,
+      codeProfile: null,
+    );
 
     _dataNotifier.value = null;
 
     Future<void> pullAndProcessHelper() async {
       // TODO(kenz): add a cancel button to the processing UI in case pulling a
       // large payload from the vm service takes a long time.
-      cpuProfileData = await serviceManager.service!.getCpuProfile(
+      cpuProfiles = await serviceManager.service!.getCpuProfile(
         startMicros: startMicros,
         extentMicros: extentMicros,
       );
       await processAndSetData(
-        cpuProfileData,
+        cpuProfiles,
         processId: processId,
         storeAsUserTagNone: true,
         shouldApplyFilters: true,
@@ -181,8 +203,8 @@ class CpuProfilerController
         analytics_constants.cpuProfileProcessingTime,
         asyncOperation: pullAndProcessHelper,
         screenMetricsProvider: () => ProfilerScreenMetrics(
-          cpuSampleCount: cpuProfileData.profileMetaData.sampleCount,
-          cpuStackDepth: cpuProfileData.profileMetaData.stackDepth,
+          cpuSampleCount: cpuProfiles.profileMetaData.sampleCount,
+          cpuStackDepth: cpuProfiles.profileMetaData.stackDepth,
         ),
       );
     } else {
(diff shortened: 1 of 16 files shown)


packages/devtools_app/lib/src/screens/profiler/cpu_profile_controller.dart
```diff
@@ -384,53 +446,74 @@ class CpuProfilerController
...
+        data,
+        label: tag,
+      );
     }
 
     data = applyToggleFilters(data);
     if (!data.processed) {
-      await transformer.processData(
-        data,
+      await data.processData(
```
kenzieschmoll, line 495: nit: rename to process now that it is being called on the data itself. `data.processData` is a bit redundant

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -15,8 +15,70 @@ import '../../primitives/utils.dart';
...
+      return functionProfile;
+    }
+    return codeProfile!;
+  }
+
+  CpuProfileData? getInactive(CpuProfilerViewType activeType) {
+    if (activeType == CpuProfilerViewType.function) {
+      return codeProfile;
+    }
+    return functionProfile;
```
kenzieschmoll, line 68: nit: these could both be ternary returns to make these functions more terse

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -916,35 +999,55 @@ class CpuProfileStore {
...
+      cpuSamples,
+      kRootIndex,
+      buildCodeTree,
+    );
     _CpuProfileTimelineTree current;
     // TODO(bkonyi): handle truncated?
-    for (final sample in cpuSamples.samples ?? []) {
+    for (final sample in cpuSamples.samples ?? <vm_service.CpuSample>[]) {
       current = root;
+      final stack = buildCodeTree ? sample.codeStack : sample.stack!;
```
kenzieschmoll, line 1014: if buildCodeTree is true, do we need to still build the function tree as well? or is it an either / or situation depending on the value of buildCodeTree?

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -916,35 +999,55 @@ class CpuProfileStore {
...
+  dynamic get _function {
+    if (isCodeTree) {
+      return samples.codes[index].code!.function!;
+    }
+    return samples.functions![index].function;
+  }
+
+  String? get name {
+    if (isCodeTree) {
+      return samples.codes[index].code!.name;
```
kenzieschmoll, line 1043: should we add a helper `get _code => samples.codes[index].code;` since this is used in a couple places?

packages/devtools_app/lib/src/screens/profiler/cpu_profile_service.dart
```diff
@@ -19,10 +20,42 @@ extension CpuProfilerExtension on VmServiceWrapper {
...
+        cpuSample.setCodeStack(rawSample[kCodeStack].cast<int>());
+      }
+    }
+
+    final functionProfile = await CpuProfileData.generateFromCpuSamples(
       isolateId: isolateId,
       cpuSamples: cpuSamples,
     );
+    CpuProfileData? codeProfile;
+    if (buildCodeProfile) {
```
kenzieschmoll, line 48: should we only be doing this in vm developer mode?

packages/devtools_app/lib/src/screens/profiler/cpu_profiler.dart
```diff
@@ -179,6 +179,21 @@ class _CpuProfilerState extends State<CpuProfiler>
               const SizedBox(width: denseSpacing),
               UserTagDropdown(widget.controller),
               const SizedBox(width: denseSpacing),
+              ValueListenableBuilder<bool>(
+                valueListenable: preferences.vmDeveloperModeEnabled,
+                builder: (context, vmDeveloperModeEnabled, _) {
+                  if (!vmDeveloperModeEnabled) {
+                    return Container();
```
kenzieschmoll, line 186: use SizedBox() instead

packages/devtools_app/lib/src/screens/profiler/cpu_profiler.dart
```diff
@@ -179,6 +179,21 @@ class _CpuProfilerState extends State<CpuProfiler>
...
+                  }
+                  return Row(
+                    children: [
+                      ModeDropdown(widget.controller),
+                      const SizedBox(width: denseSpacing),
```
kenzieschmoll, line 191: wrap ModeDropdown in a Padding widget with right padding instead of using a Row here. That will make it more clear what the sized box is for.

packages/devtools_app/lib/src/screens/profiler/cpu_profiler.dart
```diff
@@ -179,6 +179,21 @@ class _CpuProfilerState extends State<CpuProfiler>
...
+                    ],
+                  );
+                },
+              ),
+              const SizedBox(width: denseSpacing),
```
kenzieschmoll, line 196: remove this sized box. If vm developer mode is not enabled, this would result in double padding

packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
```diff
@@ -395,3 +398,84 @@ class ObjectStore {
...
+    this.kind,
+    this.inclusiveTicks,
+    this.exclusiveTicks,
+    this.code,
+  });
+
+  ProfileCode._fromJson(Map<String, dynamic> json) {
+    kind = json['kind'] ?? '';
+    inclusiveTicks = json['inclusiveTicks'] ?? -1;
+    exclusiveTicks = json['exclusiveTicks'] ?? -1;
```
kenzieschmoll, line 417: are these values ever going to be used in a summation calculation? if so, having them be -1 may mess up calculations

kenzieschmoll (commented):
done with initial review. Add some tests to verify we only work with code profiles when vm developer mode is enabled

packages/devtools_app/lib/src/screens/profiler/cpu_profile_service.dart
bkonyi: This is only done in VM developer mode as `getCpuSamples` in `vm_service_wrapper.dart` only requests the `CpuSamples` object with code profile support when VM developer mode is enabled. I can add a comment here explaining that.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
bkonyi: This is an either or situation. We'll build two `_CpuProfileTimelineTree` instances in the process of building two `CpuProfileData` instances when VM developer mode is enabled. This seemed like the least intrusive way to share the logic for function and code profiles since it doesn't involve hacking at `CpuProfileData` to have it conditionally contain two profile types.

packages/devtools_app/lib/src/screens/profiler/cpu_profiler.dart
bkonyi: Must have missed this one. Good catch!

packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
bkonyi: No, these are here simply because `ProfileFunction` in `package:vm_service` has this structure.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_controller.dart
```diff
@@ -507,11 +507,8 @@ class CpuProfilerController
 
   void updateView(CpuProfilerViewType view) {
     _viewType.value = view;
-    _dataNotifier.value = cpuProfileStore
-        .lookupProfile(
-          label: _wrapWithFilterSuffix(userTagNone),
-        )
-        ?.getActive(view);
+    _dataNotifier.value =
+        cpuProfileStore.lookupProfile(label: userTagNone)?.getActive(view);
```
kenzieschmoll, line 511: why are we no longer including the filter suffix?

bkonyi: `loadProcessedData` doesn't seem to include the filter suffix and was causing tests to fail. Should we just add the filter suffix there as well?

kenzieschmoll (approved):

packages/devtools_app/lib/src/screens/profiler/cpu_profile_controller.dart
kenzieschmoll: yes I think we need a filter suffix in both places.

kenzieschmoll: well I guess really we need to understand how tests were passing before this PR? and what about this PR changed things so that they fail?

bkonyi: Sorry, I wasn't clear. It was causing the new tests to fail since we didn't have a filter suffix in `loadProcessedData`.

PR #4627, opened 2022-10-20
Fix Debugger HoverCards

CoderDake:
> Make sure to turn off white space diff when reviewing this one
## Problem:
- debugger hovercards pop up but then don't disappear afterwards

## Solution:
Upgrade the debugger hovercards to use the new and improved HoverCardTooltips that automatically handle appearing and disappearing

### Extra Bug Fix:
- if null was returned from `asyncGenerateHoverCardData` then the spinner wasn't disappearing afterwards. So I've fixed hoverCardTooltip so that it removes the spinner when null is returned.

## Testing
- Compared the way that hovercards looked and behaved on master to the new hovercards.
    - they seem to have the same 
- Checked widget inspector hovering to ensure it is still working well


polina-c (approved):

elliette (approved):
LGTM, thanks!

PR #4628, opened 2022-10-20
Create class filter for diff.

polina-c:
Reviewed help text with writers: [go/help-for-class-filters](https://goto.google.com/help-for-class-filters)

https://user-images.githubusercontent.com/12115586/197637545-446ff62c-50cb-4236-9f1f-cae931345091.mov


case_study/memory_leaks/leaking_counter_1/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 51;
+	objectVersion = 54;
```
kenzieschmoll, line 6: accidental changes to this file?

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/utils.dart
```diff
@@ -38,3 +39,22 @@ String classesToCsv(HeapClasses classes) {
 
   return csvBuffer.toString();
 }
+
+String? _cashedIsolateId;
+String? _cashedRootPackage;
+
+/// Returns root library or empty string.
+Future<String?> tryToDetectRootPackage() async {
```
kenzieschmoll, line 47: is this code that we can share with the debugger page?

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/utils.dart
```diff
@@ -38,3 +39,22 @@ String classesToCsv(HeapClasses classes) {
 
   return csvBuffer.toString();
 }
+
+String? _cashedIsolateId;
+String? _cashedRootPackage;
```
kenzieschmoll, line 44: sp: cashed/cached

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_filter_dialog.dart
```diff
@@ -0,0 +1,159 @@
...
+    _type = filter.filterType;
+    _except.text = filter.except;
+    _only.text = filter.only ?? _rootPackage;
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+    final textFieldLeftPadding = scaleByFontFactor(40.0);
+    final itemSpacing = scaleByFontFactor(28.0);
```
kenzieschmoll, line 64: can we use shared values from theme.dart instead of defining our own here? For example we should use `defaultSpacing` instead of 28.0 to maintain consistency

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_filter_dialog.dart
```diff
@@ -0,0 +1,159 @@
...
+
+    void onTypeChanged(ClassFilterType? type) => setState(() => _type = type!);
+
+    return DevToolsDialog(
+      title: dialogTitleText(theme, 'Filter Classes and Packages'),
+      content: !_initialized
+          ? Progress(size: 30.0)
```
kenzieschmoll, line 71: let's use `CenteredCircularProgressIndicator` here instead so that we use the default size for `CircularProgressIndicator` and do not have to specify a value here

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_filter_dialog.dart
```diff
@@ -0,0 +1,159 @@
...
+                children: [
+                  IconLabelButton(
+                    tooltip: _showHelp ? 'Hide help' : 'Show help',
+                    icon: _showHelp
+                        ? Icons.keyboard_arrow_up
+                        : Icons.keyboard_arrow_down,
+                    label: 'Help',
+                    onPressed: () => setState(() => _showHelp = !_showHelp),
+                  ),
+                  if (_showHelp) ...[
```
kenzieschmoll, line 86: instead of conditionally showing this help information, why don't we always show it in a subtle text style? This is one less click / thing for the user to figure out, and they'll likely need instructions and examples anyway. See example from other Filter dialogs in DevTools:
Screen Shot 2022-10-24 at 4 23 08 PM

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_filter_dialog.dart
```diff
@@ -0,0 +1,159 @@
...
+                    const SizedBox(height: denseSpacing),
+                    const Text('Choose and customize the filter.\n'
+                        'List full or partial class names separated by new lines. For example:\n\n'
+                        '  package:myPackage/src/myFolder/myLibrary.dart/MyClass\n'
+                        '  MyClass\n'
+                        '  package:myPackage/src/\n\n'
+                        'Specify:\n'
+                        '  - ${ClassFilter.coreLibrariesAlias} for core classes without package prefix\n'
+                        '  - ${ClassFilter.dartAndFlutterLibrariesAlias} for "dart." and "package:" libraries published by Dart and Flutter orgs.'),
+                  ],
```
kenzieschmoll, line 96: additionally, pull all this help info into a private widget inside this file. This build method is getting a bit long

case_study/memory_leaks/leaking_counter_1/macos/Runner.xcodeproj/project.pbxproj
polina-c: This one is needed change: this case study is outdated and needed to test the diff.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_filter_dialog.dart
```diff
@@ -0,0 +1,159 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+    final textFieldLeftPadding = scaleByFontFactor(40.0);
+    final itemSpacing = scaleByFontFactor(28.0);
+
+    void onTypeChanged(ClassFilterType? type) => setState(() => _type = type!);
+
+    return DevToolsDialog(
+      title: dialogTitleText(theme, 'Filter Classes and Packages'),
```
kenzieschmoll, line 69: pull the code from here https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/ui/filter.dart#L134
out into a shared widget in filter.dart `FilterDialogTitle`, and use that here.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_filter_dialog.dart
```diff
@@ -0,0 +1,159 @@
...
+                  ),
+                ],
+              ),
+            ),
+      actions: [
+        DialogTextButton(
+          onPressed: () =>
+              setState(() => _loadStateFromFilter(ClassFilter.empty())),
+          child: const Text('Reset Defaults'),
+        ),
```
kenzieschmoll, line 142: remove this once you use the new common widget I suggested `FilterDialogTitle` (see previous comment)

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_filter_dialog.dart
```diff
@@ -0,0 +1,159 @@
...
+            final newFilter = ClassFilter(
+              filterType: _type,
+              except: _except.text,
+              only: _only.text,
+            );
+            if (newFilter.equals(widget.classFilter.value)) return;
+            widget.classFilter.value = newFilter;
+          },
+        ),
+        const DialogCancelButton(),
```
kenzieschmoll, line 155: Use 'Apply' and 'Cancel' to match other filter dialogs. In filter.dart, pull the pair into a shared widget or a helper method that can be used in filter.dart and in this file

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_filter_dialog.dart
```diff
@@ -0,0 +1,159 @@
...
+        ),
+        DialogCloseButton(
+          label: 'OK',
+          onClose: () {
+            final newFilter = ClassFilter(
+              filterType: _type,
+              except: _except.text,
+              only: _only.text,
+            );
+            if (newFilter.equals(widget.classFilter.value)) return;
```
kenzieschmoll, line 151: this is redundant so we can remove. ValueNotifier will not notify when the value is the same

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
```diff
@@ -199,15 +199,23 @@ class ClassesTableDiff extends StatelessWidget {
...
-      dataKey: 'ClassesTableDiff',
-      keyFactory: (e) => Key(e.heapClass.fullName),
-      selectionNotifier: selection,
-      defaultSortColumn: _retainedSizeDeltaColumn,
-      defaultSortDirection: SortDirection.descending,
+    return ValueListenableBuilder(
+      valueListenable: classes.filter,
+      builder: (_, __, ___) {
+        final data = classes.filtered();
+        final dataKey = 'ClassesTableSingle-$data';
```
kenzieschmoll, line 206: what is an example of what $data yields? if it returns something like `InstanceOf List<DiffClassStats`, then this won't be unique

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_single.dart
```diff
@@ -127,14 +127,21 @@ class ClassesTableSingle extends StatelessWidget {
...
   Widget build(BuildContext context) {
-    return FlatTable<SingleClassStats>(
-      columns: _columns,
-      data: classes.classes,
-      dataKey: 'ClassesTableSingle',
-      keyFactory: (e) => Key(e.heapClass.fullName),
-      selectionNotifier: selection,
-      defaultSortColumn: _shallowSizeColumn,
-      defaultSortDirection: SortDirection.descending,
+    return ValueListenableBuilder(
```
kenzieschmoll, line 130: instead of notifying on the filter everywhere, we should have a different value `filteredData` on `classes` that these table's use.  We do not want to be calling `filtered` in a build method as that is potentially expensive business logic.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_control_pane.dart
```diff
@@ -7,18 +7,20 @@ import 'package:flutter/material.dart';
...
 class SnapshotControlPane extends StatelessWidget {
   const SnapshotControlPane({Key? key, required this.controller})
       : super(key: key);
 
   final DiffPaneController controller;
-  static const _classFilterWidth = 200.0;
 
   @override
   Widget build(BuildContext context) {
+    final theFilter = controller.core.classFilter;
```
kenzieschmoll, line 23: nit: `filter` would be more terse as `the` doesn't really add any value here

packages/devtools_app/lib/src/screens/memory/primitives/class_name.dart
```diff
@@ -0,0 +1,265 @@
...
+    if (library.startsWith('dart.')) return true;
+
+    const packagePrefix = 'package:';
+    if (!library.startsWith(packagePrefix)) return false;
+    final slashIndex = library.indexOf('/');
+    if (slashIndex == -1) return false;
+    final packageName = library.substring(packagePrefix.length, slashIndex);
+
+    return _dartAndFlutterPackages.contains(packageName);
+  }
```
kenzieschmoll, line 89: pull the prefixes from here into a shared location and use in both places: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart#L629-L639

packages/devtools_app/lib/src/screens/memory/shared/heap/heap.dart
```diff
@@ -29,17 +35,62 @@ class AdaptedHeap {
...
-  Iterable<ClassStats> get classStatsList;
+abstract class HeapClasses<T extends ClassStats> with Sealable {
+  HeapClasses(this.filter);
+
+  List<T> get classStatsList;
+
+  final ValueListenable<ClassFilter> filter;
+}
+
+mixin Filterable<T extends ClassStats> on HeapClasses<T> {
```
kenzieschmoll, line 50: we should try to reuse the existing filter framework in devtools instead of reinventing the wheel here. Look at `filter.dart`. We should be able to use `FilterControllerMixin`

packages/devtools_app/lib/src/screens/memory/primitives/class_name.dart
```diff
@@ -0,0 +1,265 @@
...
+    }
+    return other is HeapClassName && other.fullName == fullName;
+  }
+
+  @override
+  int get hashCode => fullName.hashCode;
+}
+
+/// Packages that are published by dart.dev or flutter.dev.
+const _dartAndFlutterPackages = {
```
kenzieschmoll, line 104: this will get out of date quickly. Do you have a proposed maintenance plan for this? Additionally, do we really need to get this granular for hiding dart and flutter packages? I think users would expect that "hide dart" hides anything from dart core (dart:package_name excluding dart:ui), and "hide flutter" hides anything from the flutter framework and flutter engine ('package:flutter/...', 'flutter::...', and 'dart:ui')

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -2475,11 +2475,16 @@ class Progress extends StatelessWidget {
...
   Widget build(BuildContext context) {
+    // SizedBox is not always enough to constrain the indicator. Sometimes it should be
+    // accompanied with `Center`.
+    // https://stackoverflow.com/questions/51901379/how-to-set-size-to-circularprogressindicator
     return SizedBox(
       width: size,
       height: size,
-      child: CircularProgressIndicator(
-        color: Theme.of(context).textTheme.bodyLarge?.color,
+      child: Center(
```
kenzieschmoll, line 2484: we should delete the `Progress` widget, and add an optional size parameter to `CenteredCircularProgressIndicator`, since now these two helper widgets are almost identical

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -2508,3 +2513,33 @@ class ToCsvButton extends StatelessWidget {
...
+  Widget build(BuildContext context) {
+    return Row(
+      children: [
+        Radio<T>(
+          value: itemValue,
+          groupValue: groupValue,
+          materialTapTargetSize: MaterialTapTargetSize.shrinkWrap,
+          onChanged: onChanged,
+        ),
+        Text(label),
```
kenzieschmoll, line 2541: This should likely be in an expanded to adapt to all layout sizes where this will be used. And give the Text an overflow style of ellipsis

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/utils.dart
polina-c: reused in inspector. Debugger already has isolate instance that it uses for other goals, so it is better as it is.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_filter_dialog.dart
polina-c: used defaultSpacing

But, I cannot replace textFieldLeftPadding as it should be of width of the radio button.
I think it is too specific for this dialog to share.

packages/devtools_app/lib/src/screens/memory/shared/heap/heap.dart
polina-c: I reviewed it. It makes sense to reuse UI, but, as there is a lot of specifics in diffing, it seems reuse of data processing will bring more complexity, than value.

Here is refactoring needed to reuse UI: https://github.com/flutter/devtools/pull/4638

CoderDake (commented):
Going to review some more after my lunch too

packages/devtools_app/lib/src/screens/memory/shared/heap/class_filter.dart
```diff
@@ -0,0 +1,137 @@
...
+        return stringToSet(except);
+      case ClassFilterType.only:
+        return stringToSet(only);
+    }
+  }
+
+  late final Set<String> filters = _filtersAsSet();
+
+  /// Task to be applied when filter changed.
+  FilteringTask task({required ClassFilter? previous}) {
```
CoderDake, line 85: could the previous filter logic be kept in FilteringTask instead?

packages/devtools_app/lib/src/screens/memory/shared/heap/heap.dart
```diff
@@ -29,17 +35,62 @@ class AdaptedHeap {
...
+  ClassFilter? _appliedFilter;
+  List<T>? _filtered;
+
+  List<T> filtered() {
+    final newFilter = filter.value;
+    final oldFilter = _appliedFilter;
+    final oldFiltered = _filtered;
+    _appliedFilter = newFilter;
+    if ((oldFilter == null) != (oldFiltered == null)) {
+      throw StateError('nullness should match');
```
CoderDake, line 60: this error message might need a bit more context.

packages/devtools_app/lib/src/screens/memory/shared/heap/heap.dart
```diff
@@ -29,17 +35,62 @@ class AdaptedHeap {
...
+      return _filtered = classStatsList;
+    }
+
+    final Iterable<T> dataToFilter;
+    if (task == FilteringTask.refilter) {
+      dataToFilter = classStatsList;
+    } else if (task == FilteringTask.reuse) {
+      dataToFilter = oldFiltered!;
+    } else {
+      throw StateError('Unexpected value: $task.');
```
CoderDake, line 81:
```suggestion
      throw StateError('Unexpected task: $task.');
```

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -159,3 +159,19 @@ mixin ProvidedControllerMixin<T, V extends StatefulWidget> on State<V> {
     return true;
   }
 }
+
+String? _cachedIsolateId;
+String? _cachedRootPackage;
```
CoderDake, line 164: Would these cached values be frequently garbage collected since they wouldn't have a permanent instance that they live in?

also how often do isolateId's conflict between applications. Would we ever run into a case where we grab the rootPackage of another application since their isolateIds were identical?

CoderDake: Another thing I'm thinking of, should we be persisting these in a dictionary incase we reconnect to the application and want to grab this again?

packages/devtools_app/lib/src/screens/memory/shared/heap/heap.dart
polina-c: I assume the person who investigate it will read the code. And code is very descriptive here.
Do see cases when code will not be available for the investigator?

packages/devtools_app/lib/src/shared/utils.dart
polina-c: Globals are never garbage collected.
I am ok to re-query the package in case of reconnection, as it is rare case and is not worth code complexity IMHO.
Moved it to serviceManager ant protected from conflict. Thank you.

packages/devtools_app/lib/src/screens/memory/primitives/class_name.dart
polina-c: On the tab 'Analysis' we already filter out 'collection' and some other non 'package:flutter/' package.
And Dan confirmed it is how it should be.

My hypothesis is:
1. In memory troubleshooting user will want to focus on non-Google packages, because the Google ones are highly trusted.
2. We do not guarantee to exclude all Google-created package. And it is not a big deal if we miss some of them. User can always add it to the list and file a bug if it bothers them. 
3. We can create a test that pulls html content of the pub.dev pages and verifies all of them are listed. It will be more convenient to find an API to pull the list in a structured way though. But, again, it is nice to have.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -258,8 +264,19 @@ class DerivedData extends DisposableController with AutoDisposeControllerMixin {
...
+    if (classStats == null) return null;
+    if (_core.classFilter.value.apply(classStats.heapClass)) return classStats;
+    return null;
+  }
+
+  bool _updatingValues = false;
+
   /// Updates fields in this instance based on the values in [core].
   void _updateValues() {
+    assert(!_updatingValues);
```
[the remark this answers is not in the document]
polina-c: Started using in a widget and added comment why it is important.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_details/paths.dart
```diff
@@ -8,15 +8,15 @@ import '../../../../../../primitives/utils.dart';
...
 import '../../../../shared/heap/heap.dart';
-import '../../../../shared/heap/primitives.dart';
 
 class _RetainingPathColumn extends ColumnData<StatsByPathEntry> {
   _RetainingPathColumn(String className)
       : super.wide(
           'Shortest Retaining Path for Instances of $className',
-          titleTooltip: 'Class names of objects that retain'
-              '\nthe instances from garbage collection.',
+          titleTooltip: 'The shortest of sequences of objects that retain'
```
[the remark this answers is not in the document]
polina-c: May be the phrasing should be improved other way though.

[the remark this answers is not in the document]
kenzieschmoll: I would say "The shortest sequence" instead of "The shortest of sequences" as these are equivalent and the former is easier to understand and more terse.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/utils.dart
```diff
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import '../../../../../shared/globals.dart';
+import '../../../../../../devtools_app.dart';
```
kenzieschmoll, line 5: import files directly

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_filter_dialog.dart
```diff
@@ -59,101 +64,76 @@ class _ClassFilterDialogState extends State<ClassFilterDialog> {
...
+            groupValue: _type,
+            onChanged: onTypeChanged,
+          ),
+          const SizedBox(height: defaultSpacing),
+          RadioButton<ClassFilterType>(
+            label: 'Show all classes except:',
+            itemValue: ClassFilterType.except,
+            groupValue: _type,
+            onChanged: onTypeChanged,
+          ),
```
kenzieschmoll, line 102: Create a helper method that takes a label and a ClassFilterType to reduce duplicated code between all these RadioButtons

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
```diff
@@ -199,23 +199,17 @@ class ClassesTableDiff extends StatelessWidget {
...
-          columnGroups: _columnGroups,
-          data: data,
-          dataKey: dataKey,
-          keyFactory: (e) => Key(dataKey),
-          selectionNotifier: selection,
-          defaultSortColumn: _retainedSizeDeltaColumn,
-          defaultSortDirection: SortDirection.descending,
-        );
-      },
+    final dataKey = 'ClassesTableSingle-${identityHashCode(classes)}';
```
kenzieschmoll, line 202: do we want to be uniquely identifying this table by the data set? I would think we'd want to preserve the sorting and sort directions for ClassesTableDiff no matter what the data passed to it is.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_single.dart
```diff
@@ -127,21 +127,15 @@ class ClassesTableSingle extends StatelessWidget {
...
-          columns: _columns,
-          data: data,
-          dataKey: dataKey,
-          keyFactory: (e) => Key(dataKey),
-          selectionNotifier: selection,
-          defaultSortColumn: _shallowSizeColumn,
-          defaultSortDirection: SortDirection.descending,
-        );
-      },
+    final dataKey = 'ClassesTableSingle-${identityHashCode(classes)}';
```
kenzieschmoll, line 130: same question here. We probably want the same key for all table of this type 'ClassesTableSingle'

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_view.dart
```diff
@@ -20,34 +20,43 @@ class SnapshotView extends StatelessWidget {
...
+          classTable = ClassesTableSingle(
+            classes: singleClasses,
             selection: controller.derived.selectedSingleClassStats,
           );
-        } else if (classes is DiffHeapClasses) {
-          table1 = ClassesTableDiff(
-            classes: classes,
+        } else if (diffClasses != null) {
+          classTable = ClassesTableDiff(
+            classes: controller.derived.diffClassesToShow.value!,
```
kenzieschmoll, line 52: use `diffClasses` instead here

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -163,6 +163,24 @@ class ServiceConnectionManager {
     return listenable;
   }
 
+  String? _cachedRootLibKey;
+  String? _cachedRootLibValue;
+
+  /// Returns root library or empty string.
```
kenzieschmoll, line 169: looks like it returns root library or null if it cannot detect it. Update comment to match

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -163,6 +163,24 @@ class ServiceConnectionManager {
...
+  Future<String?> tryToDetectRootLib() async {
```
kenzieschmoll, line 170: will we ever try to detect the root library for an isolate that is not the main isolate? Caching the value of the root library and only grabbing it for the main isolate will be an issue if that is the case

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -176,6 +194,9 @@ class ServiceConnectionManager {
       _serviceAvailable = Completer();
     }
 
+    _cachedRootLibKey = null;
+    _cachedRootLibValue = null;
```
kenzieschmoll, line 198: we should null these out in vmServiceClosed, not vmServiceOpened. We can assert that they are null here if we want to be extra careful

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_filter_dialog.dart
```diff
@@ -0,0 +1,139 @@
...
+          setState(() => _loadStateFromFilter(ClassFilter.empty())),
+      onApply: () {
+        final newFilter = ClassFilter(
+          filterType: _type,
+          except: _except.text,
+          only: _only.text,
+        );
+        widget.onChanged(newFilter);
+      },
+      onCancel: null,
```
kenzieschmoll, line 85: Can we make onCancel non-required instead of forcing an explicit passing of null? This is pretty standard for Flutter

packages/devtools_app/lib/src/screens/memory/primitives/class_name.dart
```diff
@@ -0,0 +1,285 @@
...
+            theClass?.libraryName,
+            theClass?.libraryUri.toString(),
+          ),
+          className: theClass?.name ?? '',
+        );
+
+  static String _library(String? libName, String? libUrl) {
+    libName ??= '';
+    if (libName.isNotEmpty) return libName;
+    return libUrl ?? '';
```
kenzieschmoll, line 41: this could be a one liner
`return libName ?? (libUrl ?? '');`

packages/devtools_app/lib/src/screens/memory/primitives/class_name.dart
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  final String className;
+  final String library;
+
+  String get fullName => library.isNotEmpty ? '$library/$className' : className;
+
+  bool get isSentinel => className == 'Sentinel' && library.isEmpty;
+
+  /// Detects if a class can retain an object from garbage collection.
```
kenzieschmoll, line 51: nit: can we phrase this comment in terms of the getter name? This dart doc would make me think the opposite, that a weak entry retains an object from garbage collection. How about:

/// Whether a class can hold a reference to an object without preventing garbage collection.

packages/devtools_app/lib/src/screens/memory/primitives/class_name.dart
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  /// True, if the library is a core library.
+  ///
+  /// I.e. if the library name is empty or does not have prefix
+  /// `dart:` or `package:`.
+  bool get isCore =>
+      library.isEmpty ||
+      (!library.startsWith(PackagePrefixes.dart) &&
+          !library.startsWith(PackagePrefixes.genericDartPackage));
```
kenzieschmoll, line 81: wouldn't this also pick up any arbitrary package that is outside of user code? Why would we want to filter that out when the user is trying to filter out dart core libraries?

packages/devtools_app/lib/src/screens/memory/primitives/class_name.dart
```diff
@@ -0,0 +1,285 @@
...
+    // (1) name collision or (2) bug in this code.
+    // Throwing exception in debug mode to verify option #2.
+    // TODO(polina-c): create a way for users to add their weak classes
+    // or detect weak references automatically, without hard coding
+    // class names.
+    assert(false, 'Unexpected library for $className: $library.');
+    return false;
+  }
+
+  /// True, if the library is a core library.
```
kenzieschmoll, line 74: can you add some explanation for what a "core" library is?

packages/devtools_app/lib/src/screens/memory/primitives/class_name.dart
```diff
@@ -0,0 +1,285 @@
...
+  /// `dart:` or `package:`.
+  bool get isCore =>
+      library.isEmpty ||
+      (!library.startsWith(PackagePrefixes.dart) &&
+          !library.startsWith(PackagePrefixes.genericDartPackage));
+
+  /// True, if the package has prefix `dart:` or has perfix `package:` and is
+  /// published by Dart or Flutter org.
+  bool get isDartOrFlutter {
+    if (library.startsWith(PackagePrefixes.dart)) return true;
```
kenzieschmoll, line 86: can we also add a shortcut check for flutter libraries so we don't have to perform List.contains for every Flutter lib?

packages/devtools_app/lib/src/screens/memory/shared/heap/class_filter.dart
```diff
@@ -0,0 +1,140 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/material.dart';
+
+import '../../primitives/class_name.dart';
+
+enum ClassFilterType {
+  all,
```
kenzieschmoll, line 10: can this be 'showAll', or 'none'? 'all' by itself makes it seem like we are filtering all classes instead of showing all (i.e. no filter)

packages/devtools_app/lib/src/screens/memory/shared/heap/class_filter.dart
```diff
@@ -0,0 +1,140 @@
...
+    return displayString;
+  }
+
+  bool equals(ClassFilter value) {
+    return value.filterType == filterType &&
+        value.except == except &&
+        value.only == only;
+  }
+
+  Set<String> _filtersAsSet() {
```
kenzieschmoll, line 71: one thing to be aware of here is that users may expect to be able to enter a comma separated list instead of adding a new line for each filter item

packages/devtools_app/lib/src/screens/memory/shared/heap/class_filter.dart
```diff
@@ -0,0 +1,140 @@
...
+      return FilteringTask.reuse;
+    }
+
+    return FilteringTask.refilter;
+  }
+
+  String get displayString {
+    switch (filterType) {
+      case ClassFilterType.all:
+        return 'Show all classes.';
```
kenzieschmoll, line 113: remove the period to be consistent with the display of the other two options, which don't end in periods

packages/devtools_app/lib/src/screens/memory/shared/heap/heap.dart
```diff
@@ -33,12 +35,52 @@ class AdaptedHeap {
   }
 }
 
-abstract class HeapClasses with Sealable {
-  Iterable<ClassStats> get classStatsList;
+abstract class HeapClasses<T extends ClassStats> with Sealable {
+  List<T> get classStatsList;
+}
+
+mixin Filterable<T extends ClassStats> on HeapClasses<T> {
```
kenzieschmoll, line 42: make this private since this filtering mixin is specific to the memory code and shouldn't be used in other places in Devtools

packages/devtools_app/test/memory/diff/controller/filter_test.dart
```diff
@@ -0,0 +1,34 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:devtools_app/src/screens/memory/shared/heap/class_filter.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  group('$ClassFilter', () {
+    test('task for ${ClassFilterType.except}', () {
```
kenzieschmoll, line 10: add a class for the all case so that we can verify the data does not change

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
polina-c: Added TODO to address it later.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_single.dart
polina-c: Added TODO.

packages/devtools_app/lib/src/service/service_manager.dart
polina-c: renamed to clarify that we use main isolate

packages/devtools_app/lib/src/screens/memory/primitives/class_name.dart
polina-c: If libName is empty, I want to return libUrl
Reduced to two lines

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_single.dart
kenzieschmoll: I expect this will completely break the sorting behavior you wanted (maintaining sort direction / column between snapshots) in the current state.

packages/devtools_app/lib/src/screens/memory/primitives/class_name.dart
polina-c: Added. It is something very internal. Deeper, than 'dart:'
```
/// True, if the library does not belong to a package.
  ///
  /// I.e. if the library does not have prefix
  /// `dart:` or `package:`.
  /// Examples of such classes: Code, Function, Class, Field,
  /// number_symbols/NumberSymbols, vector_math_64/Matrix4.
  bool get isPackageless =>
```

packages/devtools_app/lib/src/screens/memory/shared/heap/class_filter.dart
polina-c: added support for comma

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_single.dart
polina-c: Yes, I just do not want to work on this in this PR.

packages/devtools_app/lib/src/screens/memory/shared/heap/heap.dart
polina-c: It is used outside the library. Renamed to FilterableHeapClasses to imply it is memory specific.

packages/devtools_app/lib/src/screens/memory/shared/heap/class_filter.dart
```diff
@@ -69,11 +69,16 @@ class ClassFilter {
...
 
   Set<String> _filtersAsSet() {
-    Set<String> stringToSet(String? s) =>
-        s == null ? {} : s.split('\n').where((e) => e.isNotEmpty).toSet();
+    Set<String> stringToSet(String? s) => s == null
+        ? {}
+        : s
+            .split(RegExp(',|\n'))
+            .where((e) => e.isNotEmpty)
+            .map((e) => e.trim())
```
kenzieschmoll, line 77: we should `trim` before we call `where` so that we do not end up with empty strings in the set

packages/devtools_app/lib/src/screens/memory/primitives/class_name.dart
```diff
@@ -0,0 +1,289 @@
...
+    if (library.startsWith(PackagePrefixes.flutterPackage)) return true;
+
+    if (!library.startsWith(PackagePrefixes.genericDartPackage)) return false;
+
+    final slashIndex = library.indexOf('/');
+    if (slashIndex == -1) return false;
+    final packageName = library.substring(
+      PackagePrefixes.genericDartPackage.length,
+      slashIndex,
+    );
```
kenzieschmoll, line 98: You could shorten this up by using `library.split('/').last`

kenzieschmoll (approved):
couple more comments then lgtm

packages/devtools_app/lib/src/screens/memory/primitives/class_name.dart
polina-c: Not really: I need a piece between ':' and '\' and the fastest way to do it is `substring`.

PR #4629, opened 2022-10-20
Reorder performance view tabs and hide Event details for most tabs

kenzieschmoll:
Before tab order: 
Timeline Events - Frame Analysis - Raster Stats

After tab order:
Frame Analysis - Raster Stats - Timeline Events

This PR also includes the change to only show the Event Details view for the Timeline Events tab (which is really the only place it makes sense).

This breaks out some of the changes from https://github.com/flutter/devtools/pull/4566 (which is already approved) into a separate PR. Landing these separately since @jacob314 is out for a few weeks, and I'd like to use these changes in another PR I'm working on.


CoderDake (approved):
Seems about right to me :D
Extra points for splitting a bit off of 4566

PR #4633, opened 2022-10-21
Delete inspector polyfill logic

kenzieschmoll:
Fixes https://github.com/flutter/devtools/pull/4633. Requires https://github.com/flutter/flutter/pull/113861 to land and roll into a flutter-candidate-* branch before we can land this PR.


CoderDake (approved):

jacob314:
Can this be landed now? Would be great to get rid of the polyfill now that our version is tied to Flutter.
One thing we might want to do is actively fail at startup if you happen to use DevTools with an incompatible version of Flutter. That could happen if someone launched DevTools from one Flutter SDK and connected to a Flutter app from a different Flutter SDK. With this and similar changes we will start mysteriously failing when this happens. Users will assume DevTools is flaky rather than understanding they used an incompatible Flutter unless we warn them.

kenzieschmoll:
yes this can be landed now. I've been waiting for https://github.com/flutter/flutter/pull/113861 to roll into google3, which looks like just happened friday afternoon.

jacob314:
Fyi @annagrin and @elliette. This should significantly speedup Flutter Web inspector startup as we won't be doing evals as part of the initial load.

PR #4635, opened 2022-10-21
Introduce _kEnableBeta.

polina-c:


packages/devtools_app/lib/src/primitives/feature_flags.dart
```diff
@@ -14,12 +14,17 @@
 ///   ]
 const bool _kEnableExperiments = bool.fromEnvironment('enable_experiments');
 
+/// If true, features, ready for beta testing, will be on.
+///
+/// See [_kEnableExperiments] documentation  for usage.
+const bool _kEnableBeta = bool.fromEnvironment('enable_beta');
```
kenzieschmoll: instead of using an environment variable, can we just default this value to !isExternalBuild? g3 users won't be running DevTools from the source code where they can pass in arguments

packages/devtools_app/lib/src/primitives/feature_flags.dart
```diff
@@ -12,14 +12,20 @@
...
 ///     "--dart-define=enable_experiments=true"
 ///   ]
-const bool _kEnableExperiments = bool.fromEnvironment('enable_experiments');
+const bool _kEnableExperiments = bool.fromEnvironment('enable_experiments') || _kEnableBeta;
+
+/// If true, features, ready for beta testing, will be on.
+///
+/// See [_kEnableExperiments] documentation  for usage.
+const bool _kEnableBeta =
+    bool.fromEnvironment('enable_beta');
```
kenzieschmoll, line 21: you can also add `!kReleaseMode || ...` so that when developing DevTools in debug or profile mode, this feature will be turned on by default, but it will not be turned on in release builds until we remove the beta flag

packages/devtools_app/lib/src/shared/feature_flags.dart
```diff
@@ -12,14 +16,22 @@
...
-const bool _kEnableExperiments = bool.fromEnvironment('enable_experiments');
+
+bool _kEnableExperiments =
+    const bool.fromEnvironment('enable_experiments') || !kReleaseMode;
+
+/// If true, features, ready for beta testing, will be on.
+///
+/// Always true when [_kEnableExperiments] is true.
+/// See [_kEnableExperiments] documentation  for usage.
+late final bool _kEnableBeta = !isExternalBuild;
```
kenzieschmoll, line 27: I think this be `const bool.fromEnvironment('enable_experiments') || !isExternalBuild` according to the dart doc

polina-c: Fixed. Thank you. It cannot be constant though.

packages/devtools_app/lib/src/shared/globals.dart
```diff
@@ -15,6 +17,22 @@ import 'framework_controller.dart';
 import 'preferences.dart';
 import 'survey.dart';
 
+/// Whether this DevTools build is external.
+bool get isExternalBuild => _isExternalBuild;
+bool _isExternalBuild = false;
+
+/// Flag the build as external.
+set isExternalBuild(bool value) {
+  if (kReleaseMode && _isExternalBuild && !value) {
```
kenzieschmoll, line 26: we should actually never be changing the value of isExternalBuild, regardless of what run mode we are in. So can we change this back to:

`late final bool isExternalBuild;`

packages/devtools_app/lib/src/shared/globals.dart
```diff
@@ -15,6 +17,22 @@ import 'framework_controller.dart';
...
+    throw StateError(
+      'isExternalBuild should not switch from true to false in release mode.',
+    );
+  }
+  _isExternalBuild = value;
+}
+
+bool isTestEnvironment = true;
```
kenzieschmoll, line 34: this does not seem right to have a variable isTestEnvironment that is true outside of tests.

kenzieschmoll: we should get rid of this and just set the `FeatureFlags.myFlag` value to what we want from the tests. As part of our common setup methods we can add a method `turnOffAllFeatureFlags` or something like that.

I'm trying to think if there is some way we can detect whether we are in a test or not, but I can't think of anything that is more lightweight than my suggestion currently.

packages/devtools_app/lib/src/shared/globals.dart
polina-c: It should be false for test. Should we add it to all tests as setup?

packages/devtools_app/lib/src/shared/globals.dart
polina-c: Yes, it tests all features should be off by default, except cases when we want to test the feature under construction. 

The question is - how to make all features off in tests, if we want them on when the app starts in debug mode.
May be the application should check it and, if it is not release mode, turn them on?

kenzieschmoll: > The question is - how to make all features off in tests

For this, my suggestion is to add a testing helper method `turnOffAllFeatureFlags`. It is expected to need to do some setup steps for tests. For example, in several tests we have to set globals like `IdeTheme` that are expected to be non-null. Setting all the feature flags to false for tests (other than when we want to run tests with a flag on) will allow us to keep the value of `kEnableBeta = !isExternalBuild && !kReleaseMode` for the default (non-test) case.

packages/devtools_app/lib/src/shared/globals.dart
kenzieschmoll: why should this be false for tests? the only place this should be false is in google3, where we don't run tests.

kenzieschmoll: I agree that we want to make sure experimental features are off in tests unless we are intentionally testing them. But flipping `isExternalBuild` to false is a hacky way to do this that could have other side effects in the code. Our tests test the "external" build, so this will always be true outside of google3.

packages/devtools_app/lib/src/app.dart
```diff
@@ -60,16 +61,32 @@ const debugEnableSampleScreen = false;
...
 const showVmDeveloperMode = false;
 
-/// Whether this DevTools build is external.
-bool isExternalBuild = true;
+/// If true, features under construction will be enabled.
+///
+/// By default, the features are off in release mode.
+/// To enable them pass the compilation flag
+/// `--dart-define=enable_experiments=true`.
+bool _kEnableExperiments = const bool.fromEnvironment('enable_experiments');
```
kenzieschmoll, line 69: I suggest we move this and the enableNonProdFeatures code back to feature flags. It logically makes sense to be in that file since that is where we define feature visibility.

Make the `FeatureFlags.enableNonProdFeatures` method do what the `_enableNonProdFeatures` method does here, which is check for release mode and experiment status.

packages/devtools_app/lib/src/shared/globals.dart
```diff
@@ -15,6 +15,13 @@ import 'framework_controller.dart';
 import 'preferences.dart';
 import 'survey.dart';
 
+/// Whether this DevTools build is external.
+bool get isExternalBuild => _isExternalBuild;
+bool _isExternalBuild = true;
+
+/// Flag the build as external.
+void setExternalBuild() => _isExternalBuild = true;
```
kenzieschmoll, line 23: my comment still stands about making this `late final isExternalBuild`. We never want this value changed after the initial assignment.

packages/devtools_app/lib/src/shared/globals.dart
polina-c: Good point. Set it to true by default.

packages/devtools_app/lib/src/shared/globals.dart
polina-c: The problem is that test environment never initializes this flag and thus fails with non initialized late field.
We do not have a shared test configuration point, right?

polina-c: I can add flag like 'bool _sealed' and set it in setExternalBuild, though. How does it sound?

PR #4636, opened 2022-10-22
Show async-loading debugger panel hovercards after 100ms

elliette:
Currently in the debugger panel, when hovering on a symbol, we immediately show a hovercard with a spinner. Once the expression evaluation completes, we remove the spinner and show the result, or remove the hovercard entirely if the result is null. This means that a hovercard can appear with a spinner, only to be removed right after. 

This PR adds a heuristic to try to prevent that. We're assuming that when expression evaluation returns with null it does so fast-ish, and so we only show the hovercard with a spinner if we haven't received a result after 100ms.

I still need to add tests, but would appreciate a review on the implementation first :)

### DIFF
--- a/packages/devtools_app/lib/src/ui/hover.dart
+++ b/packages/devtools_app/lib/src/ui/hover.dart
@@ -402,13 +402,12 @@ class _HoverCardTooltipState extends State<HoverCardTooltip> {
           spinnerHoverCard != null &&
           !_hoverCardController.isHoverCardStillActive(spinnerHoverCard),
     );
+    final hoverCardDataCompleter = _hoverCardDataCompleter(hoverCardDataFuture);
     // If we have set the async hover card to show up only after a timeout,
     // then race the timeout against generating the hover card data. If
     // generating the data completes first, immediately show the hover card
     // (or return early if there is no data).
     if (asyncTimeout != null) {
-      final hoverCardDataCompleter =
-          _hoverCardDataCompleter(hoverCardDataFuture);
       await Future.any([
         _timeoutCompleter(asyncTimeout).future,
         hoverCardDataCompleter.future,
@@ -441,7 +440,7 @@ class _HoverCardTooltipState extends State<HoverCardTooltip> {
     );
 
     // The spinner is showing, we can now generate the HoverCardData
-    final hoverCardData = await hoverCardDataFuture;
+    final hoverCardData = await hoverCardDataCompleter.future;
 
     if (!_hoverCardController.isHoverCardStillActive(spinnerHoverCard)) {
       // The hovercard became stale while fetching it's data. So it should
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -358,74 +368,139 @@ class _HoverCardTooltipState extends State<HoverCardTooltip> {
...
-      } else {
-        assert(widget.generateHoverCardData != null);
+    _setHoverCard(
+      spinnerHoverCard,
+    );
 
-        hoverCardData = widget.generateHoverCardData!(event);
-      }
+    // The spinner is showing, we can now generate the HoverCardData
+    final hoverCardData = await hoverCardDataFuture;
```
CoderDake, line 444: is it possible to wait on our completer instead? Something about this makes it seem like we would be fetching the data a 2nd time by calling this again

packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -358,74 +368,139 @@ class _HoverCardTooltipState extends State<HoverCardTooltip> {
...
+      isHoverStale: () =>
+          spinnerHoverCard != null &&
+          !_hoverCardController.isHoverCardStillActive(spinnerHoverCard),
+    );
+    // If we have set the async hover card to show up only after a timeout,
+    // then race the timeout against generating the hover card data. If
+    // generating the data completes first, immediately show the hover card
+    // (or return early if there is no data).
+    if (asyncTimeout != null) {
+      final hoverCardDataCompleter =
```
CoderDake, line 410: Elliott this is such a great use of completers and future.any! Such a slick way of approaching this problem!

packages/devtools_app/lib/src/ui/hover.dart
```diff
@@ -358,74 +368,139 @@ class _HoverCardTooltipState extends State<HoverCardTooltip> {
...
+    if (!_hoverCardController.isHoverCardStillActive(spinnerHoverCard)) {
+      // The hovercard became stale while fetching it's data. So it should
+      // no longer be shown.
+      return;
+    }
+    if (hoverCardData == null) {
+      // No data was provided so remove the spinner
+      _removeHoverCard(spinnerHoverCard);
+      return;
+    }
```
CoderDake, line 455: I think we might need to post the hoverCard data in a hovercard here
hoverCardData
otherwise if we wait long enough to get the spinner, the official hovercard won't ever show

packages/devtools_app/lib/src/ui/hover.dart
elliette: Good point! Switched to using the completer

packages/devtools_app/lib/src/ui/hover.dart
elliette: Oops! Thanks for catching that

CoderDake (approved):

PR #4638, opened 2022-10-25
Decouple filter UI from data processing.

polina-c:
Separate dialog DialogWithDefaults, to make the filtering UI available for reuse.

### DIFF
--- a/packages/devtools_app/lib/src/shared/dialogs.dart
+++ b/packages/devtools_app/lib/src/shared/dialogs.dart
@@ -5,13 +5,21 @@
 import 'package:flutter/material.dart';
 import 'package:pointer_interceptor/pointer_interceptor.dart';
 
+import '../ui/label.dart';
 import 'common_widgets.dart';
 import 'theme.dart';
 
 const dialogDefaultContext = 'dialog';
 
-Text dialogTitleText(ThemeData theme, String text) {
-  return Text(text, style: theme.textTheme.titleLarge);
+class DialogTitleText extends StatelessWidget {
+  const DialogTitleText({super.key, required this.text, required this.theme});
+
+  final ThemeData theme;
+  final String text;
+
+  @override
+  Widget build(BuildContext context) =>
+      Text(text, style: theme.textTheme.titleLarge);
 }
 
 List<Widget> dialogSubHeader(ThemeData theme, String titleText) {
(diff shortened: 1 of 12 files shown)


packages/devtools_app/lib/src/shared/dialogs.dart
```diff
@@ -5,13 +5,21 @@
...
 import 'theme.dart';
 
 const dialogDefaultContext = 'dialog';
 
-Text dialogTitleText(ThemeData theme, String text) {
-  return Text(text, style: theme.textTheme.titleLarge);
+class DialogTitleText extends StatelessWidget {
+  const DialogTitleText({super.key, required this.text, required this.theme});
+
+  final ThemeData theme;
```
kenzieschmoll, line 17: we shouldn't be passing the theme in. We can access the theme in the build method directly.

packages/devtools_app/lib/src/shared/dialogs.dart
```diff
@@ -21,6 +29,94 @@ List<Widget> dialogSubHeader(ThemeData theme, String titleText) {
   ];
 }
 
+/// A standardized dialog with help text and buttons `Reset to default`,
+/// `APPLY` and `CANCEL`.
+class DialogWithDefaults extends StatelessWidget {
```
kenzieschmoll, line 34: rename to FilterDialogWithDefaults since this combination of actions (apply + cancel + reset to default) only apply to filter dialogs, not all dialogs

packages/devtools_app/lib/src/shared/dialogs.dart
```diff
@@ -21,6 +29,94 @@ List<Widget> dialogSubHeader(ThemeData theme, String titleText) {
...
+      ),
+      actions: [
+        DialogApplyButton(onPressed: onApply ?? () {}),
+        DialogCancelButton(cancelAction: onCancel),
+      ],
+    );
+  }
+}
+
+class _DialogWithDefaultsTitle extends StatelessWidget {
```
kenzieschmoll, line 84: rename to _FilterDialogTitle, since again, the reset action only applies to filter dialogs

packages/devtools_app/lib/src/shared/dialogs.dart
```diff
@@ -21,6 +29,94 @@ List<Widget> dialogSubHeader(ThemeData theme, String titleText) {
...
+        padding: const EdgeInsets.symmetric(
+          horizontal: defaultSpacing,
+        ),
+        width: dialogWidth ?? defaultDialogWidth,
+        child: Column(
+          mainAxisSize: MainAxisSize.min,
+          crossAxisAlignment: CrossAxisAlignment.start,
+          children: [
+            child,
+            if (helpText != null) ...[
```
kenzieschmoll, line 68: the help text should probably be at the top of the dialog, not bottom. swap order with child

packages/devtools_app/lib/src/shared/dialogs.dart
```diff
@@ -21,6 +29,94 @@ List<Widget> dialogSubHeader(ThemeData theme, String titleText) {
...
+              const SizedBox(height: defaultSpacing),
+              DialogHelpText(helpText: helpText!),
+              const SizedBox(height: defaultSpacing),
```
kenzieschmoll, line 71: nit: add this to the Padding around the Column instead of having a spacer as the last column child. Makes it more clear that this spacing is for padding and not for the column content

packages/devtools_app/lib/src/shared/dialogs.dart
kenzieschmoll: Oh actually I see why you did this now, since this was the existing behavior for the filter dialogs. Will this order apply to the new filter dialog you are adding? if not, we should consider passing in `children` instead of `child`. That way the layout order is more flexible

packages/devtools_app/lib/src/shared/dialogs.dart
polina-c: Right now yes, but I can imagine some settings dialog that would looks this way too.

kenzieschmoll: Good point. Maybe something like 'StatefulDialogWithDefaults'?

polina-c: Us discussed offline renamed to StateUpdateDialog

packages/devtools_app/lib/src/shared/dialogs.dart
polina-c: Yes, i am ok to have help at the bottom. 
From UX perspective it may be even better as the user first want to be puzzled in order to be interested in help.

packages/devtools_app/lib/src/shared/dialogs.dart
```diff
@@ -54,23 +53,25 @@ class DialogWithDefaults extends StatelessWidget {
...
-          mainAxisSize: MainAxisSize.min,
-          crossAxisAlignment: CrossAxisAlignment.start,
-          children: [
-            child,
-            if (helpText != null) ...[
-              const SizedBox(height: defaultSpacing),
-              DialogHelpText(helpText: helpText!),
-              const SizedBox(height: defaultSpacing),
+        child: Padding(
+          padding: const EdgeInsets.only(bottom: defaultSpacing),
```
kenzieschmoll, line 63: we don't need an extra padding widget here. Use the padding property on Container above

kenzieschmoll (approved):
lgtm with one cleanup

PR #4669, opened 2022-10-31
Make snapshot creation for diffing async.

polina-c:


packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
```diff
@@ -43,12 +59,26 @@ class AdaptedHeapData {
     );
   }
 
-  factory AdaptedHeapData.fromHeapSnapshot(HeapSnapshotGraph graph) =>
-      AdaptedHeapData(
-        graph.objects
-            .map((e) => AdaptedHeapObject.fromHeapSnapshotObject(e))
-            .toList(),
-      );
+  factory AdaptedHeapData.fromHeapSnapshot(HeapSnapshotGraph graph) {
```
bkonyi, line 62: Why not just make this a static method to avoid passing `Future`s around? Something like this:

```dart
static Future<AdaptedHeapData> fromHeapSnapshot(HeapSnapshotGraph graph) async {
  Future<List<AdaptedHeapObject>> objectsInitializer() async {...}
  // Private constructor with non-future argument
  return AdaptedHeapData._(await objectsInitializer());
} 
```

Then this code below:

```dart
final result = AdaptedHeapData.fromHeapSnapshot(snapshot);
await result.init();
```

can just be:

```dart
final result = await AdaptedHeapData.fromHeapSnapshot(snapshot);
```

polina-c: thanks
created another PR: https://github.com/flutter/devtools/pull/4670

PR #4676, opened 2022-10-31
Make snapshot creation for diffing async.

polina-c:
Replacement for https://github.com/flutter/devtools/pull/4669.

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
+++ b/packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
@@ -54,8 +54,6 @@ class AdaptedHeapData {
 
     final objects = await Future.wait(listOfFutures);
 
-    // The future is passed as an argument intentionally.
-    // ignore: discarded_futures
     return AdaptedHeapData(objects);
   }
 
--- a/packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
+++ b/packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
@@ -4,6 +4,7 @@
 
 import 'package:vm_service/vm_service.dart';
 
+import '../../../../primitives/utils.dart';
 import '../../primitives/class_name.dart';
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
```diff
@@ -43,12 +44,20 @@ class AdaptedHeapData {
...
+  ) async {
+    final listOfFutures = graph.objects.map((e) async {
+      await delayForBatchProcessing();
+      return AdaptedHeapObject.fromHeapSnapshotObject(e);
+    }).toList();
+
+    final objects = await Future.wait(listOfFutures);
+
+    // The future is passed as an argument intentionally.
+    // ignore: discarded_futures
```
bkonyi, line 58: Is this ignore necessary? `objects` isn't an instance of `Future` at this point.

bkonyi (approved):

packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
```diff
@@ -43,12 +44,18 @@ class AdaptedHeapData {
...
-      AdaptedHeapData(
-        graph.objects
-            .map((e) => AdaptedHeapObject.fromHeapSnapshotObject(e))
-            .toList(),
-      );
+  static Future<AdaptedHeapData> fromHeapSnapshot(
+    HeapSnapshotGraph graph,
+  ) async {
+    final listOfFutures = graph.objects.map((e) async {
+      await delayForBatchProcessing();
```
kenzieschmoll, line 51: how many objects will there be? Do we really want to delay for every object or do we want to delay for every N objects? 

Here's an example from the CPU profiler where we calculate the batch size: https://github.com/kenzieschmoll/devtools/blob/scroll-to-time-range/packages/devtools_app/lib/src/screens/profiler/cpu_profile_transformer.dart#L52

PR #4677, opened 2022-10-31
Move button Clear to chart area.

polina-c:
The tabs that need to be cleared have their own buttons 'clear'.
It does not seem to be ever needed to clear all areas at once. 

Screen Shot 2022-10-31 at 4 39 59 PM


bkonyi (approved):

kenzieschmoll:
I would move the clear button to be directly to the right of pause and resume. We can put pause and resume in a button group so they do not have a space between them, and then I'd follow that button group with `SizedBox(width: defaultSpacing), ClearButton()`

polina-c:
> I would move the clear button to be directly to the right of pause and resume. We can put pause and resume in a button group so they do not have a space between them, and then I'd follow that button group with `SizedBox(width: defaultSpacing), ClearButton()`
> 
> EDIT: I thought we had a nice utility for a button group but it doesn't look like we do, so I'd start with just moving the clear button to be by the pause / resume buttons and we can worry about creating a group utility later.

We have PauseResumeButtonGroup and it is used here. I made it defaultSpacing before the clear button:
Screen Shot 2022-11-01 at 11 12 18 AM

packages/devtools_app/lib/src/screens/memory/panes/chart/chart_control_pane.dart
```diff
@@ -83,22 +83,51 @@ class _ChartControlPaneState extends State<ChartControlPane>
...
+    ga.select(analytics_constants.memory, analytics_constants.clear);
+
+    controller.memoryTimeline.reset();
+
+    // Clear all analysis and snapshots collected too.
+    controller.clearAllSnapshots();
+    controller.classRoot = null;
+    controller.topNode = null;
+    controller.selectedSnapshotTimestamp = null;
+    controller.selectedLeaf = null;
```
kenzieschmoll, line 96: do we still need all these? It looks like we are now having each memory tab below handle its own clearing?

polina-c: I will clean it up together with analysis tab.

PR #4681, opened 2022-11-01
Revert async calculations.

polina-c:
Async calculations how they are implemented now take 10 times (5 sec) longer than sync. 
We will do it smarter if we get signal it impacts  users.

For now reverting.

Details: https://github.com/flutter/devtools/pull/4680

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
+++ b/packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
@@ -4,7 +4,6 @@
 
 import 'package:vm_service/vm_service.dart';
 
-import '../../../../primitives/utils.dart';
 import '../../primitives/class_name.dart';
 import '../../primitives/memory_utils.dart';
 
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
```diff
@@ -44,16 +43,13 @@ class AdaptedHeapData {
     );
   }
 
-  static Future<AdaptedHeapData> fromHeapSnapshot(
+  static AdaptedHeapData fromHeapSnapshot(
     HeapSnapshotGraph graph,
-  ) async {
-    final listOfFutures = graph.objects.map((e) async {
-      await delayForBatchProcessing();
```
bkonyi, line 51: I think this might have been causing some of your issues. Can you try not awaiting this and see if that improves performance at all? Something like this:

```dart
final listOfFutures = graph.objects.map((e) {
  return delayForBatchProcessing().then(
    (_) async => await AdaptedHeapData.fromHeapSnapshot(graph),
  );
});
await Future.wait(listOfFutures);
```

bkonyi (approved):

packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
polina-c: tried. did not help :(

kenzieschmoll: I'm fairly certain this is because we are awaiting a delay for every object in the graph. We should not do this. See my comment here: https://github.com/flutter/devtools/pull/4676#discussion_r1010542209

PR #4683, opened 2022-11-01
Small refactoring to the performance page

kenzieschmoll:
this will help with the diff on another PR I'm working on. Also fixes some deprecation errors at HEAD


packages/devtools_app/lib/src/screens/provider/instance_viewer/instance_providers.dart
```diff
@@ -257,7 +257,7 @@ Setter? _parseSetter({
       if (field.isFinal) return null;
       return mutate;
     },
-    orElse: () => throw FallThroughError(),
+    orElse: () => throw Exception(),
```
CoderDake, line 260: Is there a message we could pass to exception that could help with context, if this ever gets hit?

kenzieschmoll: Will address in a follow up

PR #4693, opened 2022-11-02
Add support for grouping samples by tag

bkonyi:
This PR adds support for grouping CPU samples by user or VM tags. This gives developers the ability to see the full CPU profile broken down by user or VM tags all at once rather than having to inspect samples from one tag at a time.

This change also removes the root node from the displayed CPU profile tree, meaning that the 'all' node will no longer be displayed.

**Group by User Tag (Bottom Up):**
image

**Group by User Tag (Call Tree):**
image

**Group by VM Tag (Bottom Up):**
image

**Group by VM Tag (Call Tree):**
image

cc @aam


kenzieschmoll:
It looks like this change removes the Filter by tag feature? I still think it is useful to have both. Grouping and filtering are two different functions. Even if we move the Filter by tag feature into the filter dialog, that is better than completely removing it.

bkonyi:
> It looks like this change removes the Filter by tag feature? I still think it is useful to have both. Grouping and filtering are two different functions. Even if we move the Filter by tag feature into the filter dialog, that is better than completely removing it.

This doesn't remove `Filter by tag`, the dropdown is used for both `Filter by tag` and `Group by tag` since they're all related.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -78,6 +78,24 @@ class CpuProfilePair {
     return CpuProfilePair(functionProfile: function, codeProfile: code);
   }
 
+  factory CpuProfilePair.withTagRoots(
+    CpuProfilePair original,
+    CpuProfilerTagType tag,
```
kenzieschmoll, line 83: nit: rename to tagType so that it is clear this is a type and not a user / vm tag

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -78,6 +78,24 @@ class CpuProfilePair {
...
+  ) {
+    final function = CpuProfileData.withTagRoots(
+      original.functionProfile,
+      tag,
+    );
+    CpuProfileData? code;
+    if (original.codeProfile != null) {
+      code = CpuProfileData.withTagRoots(
+        original.codeProfile!,
```
kenzieschmoll, line 92: nit, put codeProfile into a local to avoid null asserting twice

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -250,20 +261,138 @@ class CpuProfileData {
...
+    int nextId = 1;
+
+    for (final tagProfileEntry in tagProfiles.entries) {
+      final tag = tagProfileEntry.key;
+      final tagProfile = tagProfileEntry.value;
+      if (tagProfile.cpuSamples.isEmpty) {
+        continue;
+      }
+      final isolateId = tagProfile.cpuSamples.first.leafId.split('-').first;
+      final tagId = '$isolateId-${nextId++}';
```
kenzieschmoll, line 306: could this stack frame id already be in use? this could cause collisions in the map.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -250,20 +261,138 @@ class CpuProfileData {
...
+        sourceLine: null,
+        parentId: null,
+        profileMetaData: metaData,
+        isTag: true,
+      );
+      final idMapping = <String, String>{
+        rootId: tagId,
+      };
+
+      String? getId(String? id) {
```
kenzieschmoll, line 323: this helper name is a little confusing. I wouldn't expect a method 'getXYZ' to have a side effect of storing into a map

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -573,19 +705,28 @@ class CpuProfileData {
...
   static const userTagKey = 'userTag';
+  static const vmTagKey = 'vmTag';
 
   final Map<String, CpuStackFrame> stackFrames;
 
   final List<CpuSampleEvent> cpuSamples;
 
   final CpuProfileMetaData profileMetaData;
 
+  /// `true` if the CpuProfileData has tag-based roots. This value is used
```
kenzieschmoll, line 716: dart doc nit: new line after first sentence

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -618,7 +760,23 @@ class CpuProfileData {
...
+    }
+    final tags = <String>{};
+    for (final cpuSample in cpuSamples) {
+      final tag = cpuSample.vmTag;
+      if (tag != null) {
+        tags.add(tag);
+      }
+    }
+    _vmTags = tags;
+    return _vmTags!;
```
kenzieschmoll, line 775: `return _vmTags = tags;`?

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -775,9 +956,12 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
   @override
   String get displayName => name;
 
+  final bool isTag;
```
kenzieschmoll, line 959: can you add some dart doc here to describe that when this is true, this is a synthetic stack frame that is not actually part of the call stack tree

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
bkonyi: No, we're regenerating stack frame IDs for all the sub-profiles using this single counter, so IDs will remain unique.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_controller.dart
```diff
@@ -29,6 +29,11 @@ enum CpuProfilerViewType {
   }
 }
 
+enum CpuProfilerTagType {
+  user,
+  vm;
```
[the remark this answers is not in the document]
bkonyi: I was originally using some enhanced enum functionality similar to `CpuProfilerViewType` above, but it can just be a comma now.

packages/devtools_app/lib/src/screens/profiler/cpu_profiler.dart
```diff
@@ -421,31 +421,47 @@ class UserTagDropdown extends StatelessWidget {
...
+                    // case.
+                    if (!(userTags.length == 1 &&
+                        userTags.first == UserTag.defaultTag.label)) ...[
+                      for (final tag in userTags)
+                        _buildMenuItem(
+                          display: '$filterByTag $tag',
+                          value: tag,
+                        ),
+                      _buildMenuItem(
+                        display: 'Group by: User Tag',
```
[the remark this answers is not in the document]
kenzieschmoll: Why don't we leave it as is for now and then check our analytics. The only other alternative I can think of is to put the filter by usertag dropdown inside the filter dialog, but I don't want to reduce the discoverability of this feature either.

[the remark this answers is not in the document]
bkonyi: SGTM

kenzieschmoll (approved):

PR #4695, opened 2022-11-02
Add analytics to snapshot adaptation. 

polina-c:


packages/devtools_app/lib/src/screens/memory/primitives/simple_elements.dart
```diff
@@ -17,3 +17,49 @@ const String nonGcableInstancesColumnTooltip =
     'Number of instances of the class,\n'
     'that are reachable, i.e. have a retaining path from the root\n'
     'and therefore can’t be garbage collected.';
+
+/// Analytic time constants specific for memory screen.
+class MemoryTimeAnalytics {
```
kenzieschmoll, line 22: please move both of these constants to analytics_constants.dart. This is where all constants for our analytics live, so let's keep them all in the same place.

packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
```diff
@@ -271,7 +275,16 @@ class SnapshotTaker {
   Future<AdaptedHeapData?> take() async {
     final snapshot = await snapshotMemory();
     if (snapshot == null) return null;
-    return AdaptedHeapData.fromHeapSnapshot(snapshot);
+    late final AdaptedHeapData result;
+    ga.timeSync(
+      memory,
```
kenzieschmoll, line 280: use `MemoryScreen.id` instead

kenzieschmoll: oh, nevermind - I see now that this is what `memory` is equal to. Can you instead add a prefix for the constants.dart import?

`import '../../../../analytics/constants.dart' as analytics_constants;`

This is what we do throughout devtools and it helps to make it more clear where these constants are coming from.

packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
```diff
@@ -288,3 +300,11 @@ mixin Sealable {
   bool get isSealed => _isSealed;
   bool _isSealed = false;
 }
+
+class _SnapshotAnalyticsMetrics extends ScreenAnalyticsMetrics {
+  _SnapshotAnalyticsMetrics({
+    required this.numberOfObjects,
+  });
+
+  final int? numberOfObjects;
```
kenzieschmoll, line 309: make this non-nullable

kenzieschmoll (approved):
lgtm with one comment

PR #4699, opened 2022-11-03
Fix potential race condition when expanding variables

CoderDake:
These buildVariablesTree call weren't getting waited on. So if they tool a little longer, the variable would expand, but it's data wouldn't refresh on the screen properly.

Open question: @kenzieschmoll should we add analytics timers here to see how long variable expansion takes?

### DIFF
--- a/packages/devtools_app/lib/src/screens/debugger/variables.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/variables.dart
@@ -37,10 +37,10 @@ class Variables extends StatelessWidget {
     );
   }
 
-  void onItemPressed(DartObjectNode v, DebuggerController controller) {
+  void onItemPressed(DartObjectNode v, DebuggerController controller) async {
     // On expansion, lazily build the variables tree for performance reasons.
     if (v.isExpanded) {
-      v.children.forEach(buildVariablesTree);
+      await Future.wait(v.children.map(buildVariablesTree));
     }
   }
 }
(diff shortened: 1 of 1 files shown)


elliette (approved):

packages/devtools_app/lib/src/screens/debugger/variables.dart
```diff
@@ -37,10 +37,10 @@ class Variables extends StatelessWidget {
     );
   }
 
-  void onItemPressed(DartObjectNode v, DebuggerController controller) {
+  void onItemPressed(DartObjectNode v, DebuggerController controller) async {
```
annagrin, line 40: Can this method return `Future<void>`? This way we can get warnings if the caller does not await on it (btw there is a warning for this as well: https://dart-lang.github.io/linter/lints/avoid_void_async.html)

packages/devtools_app/lib/src/screens/debugger/variables.dart
```diff
@@ -76,10 +76,10 @@ class ExpandableVariable extends StatelessWidget {
...
```
annagrin, line 79: Same as above.

annagrin (approved):
LGTM with comments.

kenzieschmoll:
> Open question: @kenzieschmoll should we add analytics timers here to see how long variable expansion takes?

This can be done as part of a different PR. I do think we should track analytics to see how long this operation is taking, but we'll have to be careful about this so that we don't end up sending a bunch of events due to recursion.

kenzieschmoll (approved):
+1 to Anna's comments but LGTM

annagrin (approved):
LGTM!

PR #4701, opened 2022-11-04
Add TimeRange.offset utility method.

kenzieschmoll:
This will be used in a later PR.


packages/devtools_app/test/primitives/utils_test.dart
```diff
@@ -337,6 +337,46 @@ void main() {
         expect((TimeRange()..end = Duration.zero).isWellFormed, isFalse);
         expect((TimeRange()..start = Duration.zero).isWellFormed, isFalse);
       });
+
+      test('offset', () {
```
CoderDake, line 341: nit: this could be a group called `offset`, then it could be split into 4 individual unit tests

CoderDake (approved):

packages/devtools_app/test/primitives/utils_test.dart
kenzieschmoll: done

PR #4703, opened 2022-11-04
Cache the flutter-sdk for testing

CoderDake:
This PR updates our build.yml workflow so that each of the jobs cache the flutter-sdk and the flutter build artifacts.

For example this is the macos job before the cache is populated
https://github.com/flutter/devtools/actions/runs/3411571892/jobs/5675938782

This is the macos job after the cache is populated
https://github.com/flutter/devtools/actions/runs/3411987066/jobs/5676899259

Details about the cache loading step can be seen in the "Cache Generated Files" step
Screen Shot 2022-11-07 at 2 05 48 PM

on the actions page, our caches can be seen:
Screen Shot 2022-11-07 at 1 58 00 PM


tool/bots.sh
```diff
@@ -28,7 +28,15 @@ function flutter {
...
 # Get Flutter.
 echo "Cloning the Flutter master branch"
-git clone https://github.com/flutter/flutter.git ./flutter-sdk
+if [ -d "./flutter-sdk" ]; then
+    # the flutter-sdk dir exists so just make sure it is up to date
+    pushd ./flutter-sdk
+    git fetch --all
+    popd
+else
+    # the flutter-sdk dir DOES NOT exists so clone it
```
CoderDake, line 37:
```suggestion
    # the flutter-sdk dir DOES NOT exist so clone it
```

CoderDake:
Talked with Kenzie about this and we are going to try to tighten up the cache usage even more here.
the plan will be to add a build step before`main` `test_ddc` and `test_dart2js` so they can all benefit from flutter already being built for them.
That way we don't need to do the same work 3 times.

.github/workflows/build.yaml
```diff
@@ -15,23 +15,86 @@ permissions: read-all
 
 env:
   AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
kenzieschmoll, line 18: why do we need two of these? can we just use AUTH_TOKEN?

.github/workflows/build.yaml
```diff
@@ -15,23 +15,86 @@ permissions: read-all
...
+  ACTIONS_RUNNER_DEBUG: true
```
kenzieschmoll, line 19: should this be false?

.github/workflows/build.yaml
```diff
@@ -15,23 +15,86 @@ permissions: read-all
...
+          key: flutter-sdk-${{ runner.os }}-${{ steps.flutter-candidate.outputs.FLUTTER_CANDIDATE }}
+
+      - if: ${{ steps.cache-flutter.outputs.cache-hit != 'true' }}
+        name: Clone Flutter SDK if none cached
+        run: |
+          git clone https://github.com/flutter/flutter.git ./flutter-sdk
+          cd flutter-sdk
+          git checkout $LATEST_FLUTTER_CANDIDATE
+        env:
+          LATEST_FLUTTER_CANDIDATE: ${{ steps.flutter-candidate.outputs.FLUTTER_CANDIDATE }}
```
kenzieschmoll, line 54: I see ${{ steps.flutter-candidate.outputs.FLUTTER_CANDIDATE }} being used a few times, are we duplicating this, or in other words can it be put in a local var?

.github/workflows/build.yaml
CoderDake: GH_TOKEN works automatically with the `gh api` cli stuff

I didn't want to touch the AUTH_TOKEN since I didn't really know where it is used.

.github/workflows/build.yaml
CoderDake: ty for noticing this :)

tool/latest_flutter_candidate.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Copyright 2022 The Chromium Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+RESPONSE=$(gh api --paginate /repos/flutter/flutter/branches)
```
kenzieschmoll, line 7: should we delete the latest_flutter_candidate.dart file in this PR too? It is in the repo_tool. We have other scripts that are using it (update_flutter_sdk.sh) that should use this new bash script instead.

.github/workflows/build.yaml
kenzieschmoll: it used to be used in bots.sh, but that usage is removed in this PR, so we can delete the AUTH_TOKEN one then

tool/latest_flutter_candidate.sh
CoderDake: I was thinking this could be done seperately

.github/workflows/build.yaml
CoderDake: unfortunately I don't think there is a way to make this less verbose.

tool/latest_flutter_candidate.sh
CoderDake: Looking at it now though and it looks simple enough though. I will add it now :)

kenzieschmoll: Feel free to file an issue and we can do in a follow up PR

kenzieschmoll (approved):

PR #4704, opened 2022-11-04
Add missing dart doc for performance code

kenzieschmoll:
just docs and cleanup

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/panes/flutter_frames/flutter_frames_controller.dart
+++ b/packages/devtools_app/lib/src/screens/performance/panes/flutter_frames/flutter_frames_controller.dart
@@ -37,9 +37,15 @@ class FlutterFramesController extends PerformanceFeatureController {
   ValueListenable<bool> get recordingFrames => _recordingFrames;
   final _recordingFrames = ValueNotifier<bool>(true);
 
+  /// Whether the main 'Performance' tab should be badged with a count of the
+  /// janky Flutter frames present in the Flutter frames chart.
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/screens/performance/panes/flutter_frames/flutter_frames_controller.dart
```diff
@@ -37,9 +37,15 @@ class FlutterFramesController extends PerformanceFeatureController {
   ValueListenable<bool> get recordingFrames => _recordingFrames;
   final _recordingFrames = ValueNotifier<bool>(true);
 
+  /// Whether the main 'Performance' tab should be badged with a count of the
+  /// janky Flutter frames present in the Flutter frames chart.
   ValueListenable<bool> get badgeTabForJankyFrames => _badgeTabForJankyFrames;
   final _badgeTabForJankyFrames = ValueNotifier<bool>(false);
 
+  /// The display refresh rate for the connected device.
```
polina-c, line 45: 'The display value for refresh rate'

polina-c (approved):
One comment then lgtm.

packages/devtools_app/lib/src/screens/performance/panes/flutter_frames/flutter_frames_controller.dart
kenzieschmoll: "display refresh rate" is actually the term for this value. Display refresh rate is something like 60 frames per second, 120 frames per second, etc.

Issue #4645, opened 2022-10-27
Import/Export (or Save/Open) should be discoverable and consistent


polina-c:
Problems:

1. It is now called Import/Export and should be Open/Save
2. Drag and drop for import is not discoverable
3. Memory page has different implementation

Mindy: 
needs afforances for drag and drop and import button. "Export" might be the wrong word here if the main user journey is to save and reload states within the DevTools. "Save" and "Open" might be more appropriate. Even better if on save the user can name the file and the directory is shown so they know where to find it for sharing with others. -Mindy

[Mindy / Polina (2022-10-21 11:39 GMT-7)](http://go/mindy-polina-2022-10-21-11:39-gmt-7) at 4:15

kenzieschmoll:
Switching to Open / Save SGTM

polina-c:
How about discoverability of 'Open' with drag and drop? Do we want to continue supporting it? If yes, we need to figure out how to make it discoverable.
But may be it will be better to reserve drag and drop for other more important future features that may take advantage of it?

kenzieschmoll:
It doesn't cost us any overhead to leave the Drag and drop functionality there. We can leave it here for users who are used to it and would not like if it was ripped out, and we can point all new users to use the Save / Open buttons.

polina-c:
And, we can suggest to drag and drop in the tooltip for the button 'Open'.

bkonyi:
> we already have the file selector plugin hooked up in other parts of DevTools.

I didn't think we had a file selector plugin that worked on web, or am I mistaken?

polina-c:
Hm, I did not know this. If there is no one, when user clocks "Open", we can open a dialog that say something like "To open a saved screen, drag and drop the file to any place on the screen or this dialog." 
The dialog will close on 'OK' and on drag and drop.

@melyndahoover , what do you think?

polina-c:
We have it. See for example App Size > Import File

MaryaBelanger:
> Save the screen state to the folder Download, to open or share later.

Change to:

> Save the screen state to the Downloads folder

Save's original is a little long for a tooltip. Plus the buttons are logically paired (and I'm assuming graphically paired as well), so Open's tooltip provides the same information

melyndahoover:
Marya's suggestion for the save tooltip and Polina's suggestion for the open tooltip sound good to me.

polina-c:
This is how return back to online mode is implemented in CPU Profiler. 
[...]

PR #4705, opened 2022-11-04
Configure GA for memory screen.

polina-c:


packages/devtools_app/lib/src/screens/memory/panes/control/settings_dialog.dart
```diff
@@ -35,7 +35,8 @@ class MemorySettingsDialog extends StatelessWidget {
           children: [
             CheckboxSetting(
               notifier: preferences.memory.androidCollectionEnabled,
-              title: 'Show Android memory chart',
+              title:
+                  'Show Android memory chart in addition to Dart memory chart',
```
bkonyi, line 39: Is this a tooltip?

bkonyi (approved):
Everything LGTM from a code point of view, but I'd feel more comfortable letting @kenzieschmoll verify to make sure the analytics changes are good.

packages/devtools_app/lib/src/screens/memory/panes/control/settings_dialog.dart
polina-c: Nope, it is label for the checkbox in the settings dialog.

packages/devtools_app/lib/src/analytics/constants.dart
```diff
@@ -158,47 +139,54 @@ String topicDocumentationButton(String topic) => '${topic}DocumentationButton';
...
 
 /// Analytic event constants specific for memory screen.
-class MemoryEventAnalytics {
+class MemoryEvent {
   static const gc = 'gc';
   static const settings = 'settings';
+  static const autoSnapshot = 'autoSnapshot';
 
+  static const chartLegend = 'chartLegend';
+  static const chartAndroid = 'chartAndroid';
```
kenzieschmoll, line 155: renaming the variable is fine, but we should not change the String. This will create a new event in our analytics and will prevent us from being able to look at historical data before this change.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -305,9 +318,7 @@ class DerivedData extends DisposableController with AutoDisposeControllerMixin {
 
   /// Updates fields in this instance based on the values in [core].
   void _updateValues() {
-    // Make sure the method does not trigger itself recursively.
-    assert(!_updatingValues);
-    _updatingValues = true;
+    _startUpdatingValues();
```
kenzieschmoll, line 321: instead of using timeStart and timeEnd, you could use timeSync, and place the whole body of this method in the syncOperation function parameter. This will be less added lines of code and will reduce some duplication
```
ga.timeSync(
  analytics_constants.memory,     
  analytics_constants.MemoryTime.updateValues,
  syncOperation: () {
    ... the previous body of this entire method
  },
);
```

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/heap_diff.dart
```diff
@@ -17,10 +20,34 @@ class HeapDiffStore {
...
+  ga.timeSync(
+    analytics_constants.memory,
+    analytics_constants.MemoryTime.calculateDiff,
+    syncOperation: () => result = DiffHeapClasses(couple),
+    screenMetricsProvider: () => _DiffAnalyticsMetrics(
+      objectsBefore: couple.older.data.objects.length,
+      objectsAfter: couple.younger.data.objects.length,
+    ),
+  );
+  return result;
```
kenzieschmoll, line 38: if the catch block in timeSync is hit, this will throw a late initialization error. Should we let result be nullable to handle this? or rethrow?

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/heap_diff.dart
```diff
@@ -17,10 +20,34 @@ class HeapDiffStore {
...
+}
+
+class _DiffAnalyticsMetrics extends ScreenAnalyticsMetrics {
```
kenzieschmoll, line 41: any custom metrics added in a ScreenAnalyticsMetrics subclass will need to be added to the _gtagEvent, _gtagException, and the GtagEventDevTools class in _analytics_web.dart. These will also need to be added in the Analytics360 console.

For all these reasons, we should make sure any custom metrics we want to add really will add meaning to our analysis of the timed operation.

kenzieschmoll: It would also be preferable if we didn't have multiple ScreenAnalyticsMetrics subclasses for a single screen. For example, the only existing subclasses we currently have are PerformanceScreenMetrics and ProfilerScreenMetrics, so ideally we could also have MemoryScreenMetrics that hold all the custom metrics we need for the Memory screen.

packages/devtools_app/lib/src/analytics/constants.dart
```diff
@@ -95,24 +94,6 @@ const profileAppStartUp = 'profileAppStartUp';
...
 const cpuProfileProcessingTime = 'cpuProfileProcessingTime';
 
-// Memory UX actions:
-const gc = 'gc';
-const memoryLegend = 'memoryLegend';
-const memorySettings = 'memorySettings';
-const androidChart = 'androidChart';
-const autoSnapshot = 'autoSnapshot';
-const showChart = 'showChart';
-const hideChart = 'hideChart';
```
kenzieschmoll, line 105: We shouldn't change the strings here (I see we are now using 'chartExpand' and 'chartCollapse'). Changing the strings will prevent us from being able to compare current data with historical data in our analytics

packages/devtools_app/lib/src/analytics/constants.dart
```diff
@@ -158,47 +139,54 @@ String topicDocumentationButton(String topic) => '${topic}DocumentationButton';
...
+  static const diffClassSingleSelect = 'diffClassSingleSelect';
   static const diffPathSelect = 'diffPathSelect';
+  static const diffClassDiffCopy = 'diffClassDiffCopy';
+  static const diffClassSingleCopy = 'diffClassSingleCopy';
   static const diffPathCopy = 'diffPathCopy';
   static const diffPathFilter = 'diffPathFilter';
-  static const diffPathUnfilter = 'diffPathUnfilter';
-  static const diffPathRevert = 'diffPathRevert';
-  static const diffPathUnrevert = 'diffPathUnrevert';
+  static const diffPathInvert = 'diffPathRevert';
```
kenzieschmoll, line 188: change string to match var name 'diffPathInvert' (changing this string is okay since the feature sending this event hasn't been released yet, so we don't have existing data using 'diffPathRevert')

packages/devtools_app/lib/src/screens/memory/panes/allocation_profile/allocation_profile_table_view.dart
```diff
@@ -338,7 +339,13 @@ class _AllocationProfileTableControls extends StatelessWidget {
         ),
         const SizedBox(width: denseSpacing),
         RefreshButton.icon(
-          onPressed: allocationProfileController.refresh,
+          onPressed: () async {
+            ga.select(
+              analytics_constants.memory,
+              analytics_constants.MemoryEvent.gc,
```
kenzieschmoll, line 345: We should not be re-using the same event twice. This event is already used for the GC button.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
```diff
@@ -121,7 +127,13 @@ class _AllocationTracingTableState extends State<AllocationTracingTable> {
           child: DevToolsClearableTextField(
             labelText: 'Class Filter',
             hintText: 'Filter by class name',
-            onChanged: widget.controller.updateClassFilter,
+            onChanged: (value) {
+              ga.select(
+                analytics_constants.memory,
+                analytics_constants.MemoryEvent.tracingClassFilter,
```
kenzieschmoll, line 133: will this be triggered on every update to the text field? We probably don't want to send an analytics event for every character change to the text field. Can we send this in onSubmitted or onEditingComplete instead? You'll have to play with this a bit to make sure we send the event at the right time

packages/devtools_app/lib/src/screens/memory/panes/control/source_dropdown.dart
```diff
@@ -67,10 +65,6 @@ class _MemorySourceDropdownState extends State<MemorySourceDropdown>
       value: controller.memorySource,
       onChanged: (String? newValue) {
         setState(() {
-          ga.select(
-            analytics_constants.memory,
-            analytics_constants.sourcesDropDown,
-          );
```
kenzieschmoll, line 73: are we removing this because we are removing the SourcesDropdown feature?

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
polina-c: I thought about . There are other non-functional operations to be done at start and at end. The current structure seems to be most readable for me. Yes, at cost of code duplicates.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_control_pane.dart
```diff
@@ -80,6 +85,34 @@ class SnapshotControlPane extends StatelessWidget {
...
+
+  @override
+  Widget build(BuildContext context) {
+    return FilterButton(
+      onPressed: () async {
+        ga.select(
+          analytics_constants.memory,
+          analytics_constants.MemoryEvent.diffSnapshotFilter,
+        );
+        await showDialog(
```
kenzieschmoll, line 102: this should be unawaited

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_list.dart
```diff
@@ -39,19 +41,28 @@ class _ListControlPane extends StatelessWidget {
...
               icon: Icons.block,
               tooltip: 'Clear all snapshots',
-              onPressed: clearAllEnabled ? controller.clearSnapshots : null,
+              onPressed: clearAllEnabled
+                  ? () async {
+                      ga.select(
+                        analytics_constants.memory,
+                        analytics_constants.MemoryEvent.diffClearSnapshots,
+                      );
+                      await controller.clearSnapshots();
```
kenzieschmoll, line 63: should this be awaited? I don't think we want to block the UI on this finishing.

packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
```diff
@@ -274,19 +274,23 @@ class SnapshotTaker {
...
   }
 }
 
+AdaptedHeapData _adaptSnapshotGaWrapper(HeapSnapshotGraph graph) {
+  late final AdaptedHeapData result;
+  ga.timeSync(
+    analytics_constants.memory,
+    analytics_constants.MemoryTime.adaptSnapshot,
+    syncOperation: () => result = AdaptedHeapData.fromHeapSnapshot(graph),
+    screenMetricsProvider: () => _SnapshotAnalyticsMetrics(
```
kenzieschmoll, line 287: same comment as before about _SnapshotAnalyticsMetrics (we should create MemoryScreenMetrics to hold all our custom metrics, and then we need to add handling in _analytics_web.dart)

packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
```diff
@@ -274,19 +274,23 @@ class SnapshotTaker {
...
+      numberOfObjects: graph.objects.length,
+    ),
+  );
+  return result;
```
kenzieschmoll, line 291: also same comment about whether we should add handling for the case when the timeSync catch block is hit

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/heap_diff.dart
kenzieschmoll: I'm not referring to ProcessCancelledException. If _any_ exception is thrown by the function passed to syncOperation, the catch block will be hit.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/heap_diff.dart
polina-c: It looks to be cleaner to put the subclasses to the folder 'analytics', the same way as constants, so that analytics_web does not have to reference individual screen folders.

I will move other subclasses later, in separate PRs.

polina-c: is there a guidance to add this to the Analytics360 console?

kenzieschmoll: I think I may be the only one with Admin access to our console, so if you can make the changes to the analytics_web file, I will update the analytics 360 console when those land. The order matters here (you'll see other comments like "// metric5, // metric6", so once the changes are in the order should not be changed.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
polina-c: Added throttling to handle this.

packages/devtools_app/lib/src/screens/memory/panes/control/source_dropdown.dart
polina-c: Yes, as we agreed in https://github.com/flutter/devtools/issues/4645, we will redesign the feature. I do not see value in maintaining analytics for it.

packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
polina-c: added rethrow to tymeSync

packages/devtools_app/lib/src/analytics/_analytics_stub.dart
```diff
@@ -42,11 +41,7 @@ void timeSync(
 }) {
   // Execute the operation here so that the desktop app still functions without
   // the real analytics call.
-  try {
-    syncOperation();
-  } on ProcessCancelledException catch (_) {
-    // Do nothing for instances of [ProcessCancelledException].
-  }
+  syncOperation();
```
kenzieschmoll, line 44: we need to catch and rethrow here and below if we want the consumer of timeSync and timeAsync to be able to handle the exceptions

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
```diff
@@ -128,10 +141,7 @@ class _AllocationTracingTableState extends State<AllocationTracingTable> {
             labelText: 'Class Filter',
             hintText: 'Filter by class name',
             onChanged: (value) {
-              ga.select(
-                analytics_constants.memory,
-                analytics_constants.MemoryEvent.tracingClassFilter,
-              );
+              _sendFilterEditGaEvent();
```
kenzieschmoll, line 144: is there a reason we can't use onSubmitted or onEditingComplete? Seems a bit hack to throttle when we can use one of the text field lifecycle methods that indicates "the user is done editing"

packages/devtools_app/lib/src/screens/profiler/cpu_profile_controller.dart
```diff
@@ -207,15 +207,19 @@ class CpuProfilerController
...
+        await ga.timeAsync(
+          analyticsScreenId!,
+          analytics_constants.cpuProfileProcessingTime,
+          asyncOperation: pullAndProcessHelper,
+          screenMetricsProvider: () => ProfilerScreenMetrics(
+            cpuSampleCount: cpuProfiles.profileMetaData.sampleCount,
+            cpuStackDepth: cpuProfiles.profileMetaData.stackDepth,
+          ),
+        );
+      } on ProcessCancelledException catch (_) {
```
kenzieschmoll, line 220: we are now handling ProcessCancelledException in both the try and the if block. Condense so that we are only handling it once. (wrap the try around the if/else)

packages/devtools_app/lib/src/analytics/constants.dart
kenzieschmoll: we are still changing the analytics for hiding and showing the chart, which we should not do for existing analytics. Can we switch back to using showChart and hideChart?

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -42,6 +45,17 @@ class DiffPaneController extends DisposableController {
   // is taken, and is used to assign a unique id to each [SnapshotListItem].
   int _snapshotId = 0;
 
+  VoidCallback? takeSnapshotHandler(String gaEvent) {
```
kenzieschmoll, line 48: instead of adding an additional handler, can we just add an optional parameter `gaEvent` to takeSnapshot and send the analytics event from there?

Then from the widget code:
`onPressed: () => controller.takeSnapshot(myGaEvent)`

I'm not seeing extra value by adding an additional method, plus this adds risk that takeSnapshot can be called from somewhere else and analytics will not be recorded.

packages/devtools_app/lib/src/analytics/_analytics_stub.dart
polina-c: discussed offline and agreed it is ok to do nothing

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
polina-c: both are designed to handle ui forms or dialogs that at some moment are 'done'. they need special signal that editing is done. it is not our case. 

throttling seems to be simplest and most reliable way to handle ga for filter

https://stackoverflow.com/questions/63690311/flutter-textfield-difference-between-onedittingcomplete-and-onsubmitted

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
polina-c: In both cases the handler should be null if snapshot taking is already in process. I do not want this logic to be duplicated.

PR #4712, opened 2022-11-08
Test class filter.

polina-c:


packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_control_pane.dart
```diff
@@ -87,6 +87,9 @@ class SnapshotControlPane extends StatelessWidget {
   }
 }
 
+@visibleForTesting
+const classFilterButtonKey = Key('classFilterButton');
```
bkonyi, line 91: From what I recall, we typically try and avoid using `Key`s just for testing (cc @kenzieschmoll)

polina-c: replaced with type

packages/devtools_app/test/memory/diff/widgets/class_filter_test.dart
```diff
@@ -0,0 +1,141 @@
...
+  WidgetTester tester,
+) async {
+  await _checkDataGolden(from, tester);
+
+  // Open dialog.
+  await tester.tap(find.byType(ClassFilterButton));
+  await _checkFilterGolden(from, tester);
+
+  // Select new filter.
+  await tester.tap(find.byKey(Key(to.toString())));
```
bkonyi, line 134: We don't need to find this by key, we can just do something like this:

```dart
await tester.tap(find.text(to.toString());
```

polina-c: The problem is that tap by text does not work. You need to tab the radio button directly.

bkonyi (approved):

packages/devtools_app/test/memory/diff/widgets/class_filter_test.dart
kenzieschmoll: You can find all the radio keys and then grab the one you want by index. Key's can have unintended side effects and should not be used just for testing where-ever possible. Please change.

PR #4713, opened 2022-11-08
Add diff case for class filter test.

polina-c:


packages/devtools_app/test/memory/diff/widgets/class_filter_test.dart
```diff
@@ -13,20 +14,42 @@ import 'package:flutter_test/flutter_test.dart';
...
+        scene.diffController.derived.selectedItem.value as SnapshotInstanceItem,
+        scene.diffController.core.snapshots.value[1] as SnapshotInstanceItem,
+      );
+    }
+
+    await tester.pumpWidget(scene.build());
+    await expectLater(
+      find.byType(SnapshotInstanceItemPane),
+      matchesDevToolsGolden(
+        '../../../goldens/memory_diff_snapshot_scene_${test.name}.png',
```
CoderDake, line 50:
```suggestion
        test.path,
```
Might be worth generating the path in _FilterTest

packages/devtools_app/test/memory/diff/widgets/class_filter_test.dart
```diff
@@ -41,79 +64,85 @@ void main() {
...
-    await _checkDataGolden(ClassFilterType.showAll, tester);
-  });
+      // Close dialog.
+      await tester.tap(find.text('APPLY'));
+      await _checkDataGolden(ClassFilterType.showAll, tester, t);
+    });
+  }
 }
 
+/// If type is null, fileter is [_customFilter].
```
CoderDake, line 121:
```suggestion
/// If type is null, filter is [_customFilter].
```

CoderDake (approved):

packages/devtools_app/test/memory/diff/widgets/class_filter_test.dart
```diff
@@ -13,20 +14,42 @@ import 'package:flutter_test/flutter_test.dart';
 import '../../../matchers/matchers.dart';
 import '../../../scenes/memory/diff_snapshot.dart';
 
+final _customFilter =
+    ClassFilter(filterType: ClassFilterType.only, except: '', only: '');
```
bkonyi, line 18: Nit: trailing comma

packages/devtools_app/test/memory/diff/widgets/class_filter_test.dart
```diff
@@ -41,79 +64,85 @@ void main() {
...
-  ClassFilterType type,
+  ClassFilterType? type,
   WidgetTester tester,
 ) async {
   await tester.pumpAndSettle();
   await expectLater(
     find.byType(ClassFilterDialog),
     matchesDevToolsGolden(
-      '../../../goldens/memory_diff_filter_dialog_${type.name}.png',
+      '../../../goldens/memory_diff_filter_dialog_${type?.name ?? "custom"}.png',
```
bkonyi, line 130: Nit: we can still use `'custom'` here since it's within a `${...}`.

packages/devtools_app/test/memory/diff/widgets/class_filter_test.dart
```diff
@@ -13,20 +14,42 @@ import 'package:flutter_test/flutter_test.dart';
 import '../../../matchers/matchers.dart';
 import '../../../scenes/memory/diff_snapshot.dart';
 
+final _customFilter =
+    ClassFilter(filterType: ClassFilterType.only, except: '', only: '');
+
+class _FilterTest {
```
bkonyi, line 20: This can probably be changed to an enum:

```dart
enum FilterTestType {
  diff,
  single,
}
```

And then instead of creating `_tests` below, you can just do `for (test in FilterTestType.values) {...}`, using `test.name` to get `'diff'` or `'single'`.

packages/devtools_app/test/memory/diff/widgets/class_filter_test.dart
```diff
@@ -41,79 +64,85 @@ void main() {
     scene.tearDown();
   });
 
-  testWidgetsWithWindowSize('$ClassFilterDialog filters classes', windowSize,
-      (WidgetTester tester) async {
-    await pumpScene(tester);
+  for (var t in _tests) {
```
bkonyi, line 67: Uber nit:

```dart
for (final test in _tests) {
  ...
}
```

packages/devtools_app/test/memory/diff/widgets/class_filter_test.dart
polina-c: Good idea. However, _FilterTest turned out to be convenient place to generate paths for tests, so I will stay with it.

bkonyi: This can still be an enum:

```dart
enum FilterTestType {
  diff,
  single;

  String get sceneGolden => '../../../goldens/memory_diff_snapshot_scene_$name.png';
  // ...
}

```

PR #4714, opened 2022-11-08
Enable memory diffing and retire the Analysis tab.

polina-c:
Should not be merged while https://github.com/dart-lang/sdk/issues/50226 and https://github.com/flutter/devtools/pull/4713 are active.

Release notes wording: go/wording-for-memory-diff-release-notes

Screen Shot 2022-11-08 at 2 26 54 PM


packages/devtools_app/test/scenes/memory/diff_snapshot.dart
```diff
@@ -58,9 +55,7 @@ class DiffSnapshotScene extends Scene {
   @override
   String get title => '$DiffSnapshotScene';
 
-  void tearDown() {
-    FeatureFlags.memoryDiffing = false;
-  }
+  void tearDown() {}
```
bkonyi, line 58: This isn't needed anymore.

bkonyi (approved):

packages/devtools_app/test/scenes/memory/diff_snapshot.dart
polina-c: I want it to be invoked by consumers in case we will want to put something here.

PR #4715, opened 2022-11-08
Hook up 'isActiveFeature' functionality in the Performance page

kenzieschmoll:
This allows the Perfetto controller to delay interacting with the iFrame until the view is active.

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/performance_controller.dart
+++ b/packages/devtools_app/lib/src/screens/performance/performance_controller.dart
@@ -197,6 +197,7 @@ abstract class PerformanceFeatureController extends DisposableController {
 
   PerformanceData? get data => performanceController.data;
 
+  /// Whether this feature is active and visible to the user.
   bool get isActiveFeature => _isActiveFeature;
   bool _isActiveFeature = false;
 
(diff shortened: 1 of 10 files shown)


packages/devtools_app/lib/src/screens/performance/performance_controller.dart
```diff
@@ -195,7 +197,17 @@ abstract class PerformanceFeatureController extends DisposableController {
 
   PerformanceData? get data => performanceController.data;
 
-  bool isActiveFeature = false;
+  bool get isActiveFeature => _isActiveFeature;
```
polina-c, line 200: Can you add dart doc for the getter?

packages/devtools_app/lib/src/screens/performance/performance_controller.dart
```diff
@@ -195,7 +197,17 @@ abstract class PerformanceFeatureController extends DisposableController {
...
+  bool _isActiveFeature = false;
+
+  Future<void> setIsActiveFeature(bool value) async {
```
polina-c, line 203: Would it benefit from some test coverage?

polina-c (approved):
Couple comments then lgtm.

PR #4720, opened 2022-11-10
Macos seperate goldens

CoderDake:
This change makes sure that only tests containing `matchesDevToolsGolden` will be run.

This halves the amount of time using the slow macos runners, and allows us to have sub 15 minute build times for our workflows :D


tool/bots.sh
```diff
@@ -124,13 +124,19 @@ elif [[ "$BOT" == "test_ddc" || "$BOT" == "test_dart2js" ]]; then
     fi
     echo "USE_WEBDEV_RELEASE = $USE_WEBDEV_RELEASE"
 
+    FILES="test/"
+    if [ "$ONLY_GOLDEN" = "true" ]; then
+        # Set the test files to only those containing golden test
+        FILES=$(grep -rl "matchesDevToolsGolden" test | grep "_test.dart$" | tr '\n' ' ')
```
kenzieschmoll, line 130: There is a chance that someone could add a test that uses matchesGolden instead of matchesDevToolsGolden even though all golden checks should go through matchesDevToolsGolden. So we can do one of two things:

1) add a check on this job to make sure that no test files use 'matchesGolden'
2) check for 'matchesDevToolsGolden' and 'matchesGolden' in this search function

CoderDake: kk I've added a 2nd match for us to filter on 😄

tool/bots.sh
```diff
@@ -127,7 +127,7 @@ elif [[ "$BOT" == "test_ddc" || "$BOT" == "test_dart2js" ]]; then
     FILES="test/"
     if [ "$ONLY_GOLDEN" = "true" ]; then
         # Set the test files to only those containing golden test
-        FILES=$(grep -rl "matchesDevToolsGolden" test | grep "_test.dart$" | tr '\n' ' ')
+        FILES=$(grep -rl "matchesDevToolsGolden\|matchesGolden" test | grep "_test.dart$" | tr '\n' ' ')
```
kenzieschmoll, line 130: sorry I misled you - it is actually `matchesGoldenFile`, not `matchesGolden`. matchesDevToolsGolden calls matchesGoldenFile

kenzieschmoll (approved):

Issue #3874, opened 2022-03-16
Improve readability of deep profiles


bkonyi:
**[See Observatory -> DevTools Tracking Issue](https://github.com/flutter/devtools/issues/3858).**

It's currently very difficult to keep track of which sample frame node is a parent of other nodes when stacks are very deep (e.g., Flutter widget trees):

![image](https://user-images.githubusercontent.com/24210656/158641398-09255af1-eeaa-4a06-9856-c8c4d03d1710.png)

We should consider adding some sort of guidelines (not necessarily always visible / emphasized), similar to those found in Observatory's CPU profile page:

![image](https://user-images.githubusercontent.com/24210656/158641630-70107c57-daff-4afa-9575-4faa4e6a8540.png)

PR #4722, opened 2022-11-11
Add support for displaying guidelines in `TreeTable` columns

bkonyi:
This change allows for users to enable guidelines in the CPU profiler's `TreeTable`, making it easier to read deep profiles. This setting is persisted across sessions.

Fixes https://github.com/flutter/devtools/issues/3874

**Guidelines Disabled:**
image

**Dark Theme:**
image

**Light Theme:**
image


packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -1597,6 +1620,38 @@ class _TableRowState<T> extends State<TableRow<T>>
   bool shouldShow() => widget.isShown;
 }
 
+class _RowGuidelinePainter extends CustomPainter {
+  _RowGuidelinePainter(this.level, this.colorScheme);
+
+  final int level;
+  final ColorScheme colorScheme;
```
kenzieschmoll, line 1627: nit: pass in the list of colors directly since the ColorScheme isn't used for anything else

kenzieschmoll (commented):
can we also add a test to the CPU profiler code that verifies guidelines can be added and removed?

bkonyi:
> can we also add a test to the CPU profiler code that verifies guidelines can be added and removed?

Done.

kenzieschmoll (approved):

PR #4723, opened 2022-11-11
Calculate the width of a table's columns more accurately

bkonyi:
This change updates the column width calculations for `TreeTable` to waste less space in tree columns. Previously, we were using a hard coded value of 500 plus the indent of the deepest level to determine the overall width of the tree column, which often left a significant amount of empty space.

Column width calculations for tree columns now assume that the contents of each node will be text with monospaced fonts and uses the length of the maximum displayed string plus its indent levels.

Fixes https://github.com/flutter/devtools/issues/2047

**Before:**
image

**After:**
image

### DIFF
--- a/packages/devtools_app/lib/src/shared/table/column_widths.dart
+++ b/packages/devtools_app/lib/src/shared/table/column_widths.dart
@@ -7,6 +7,7 @@ import 'dart:math';
 import '../../primitives/trees.dart';
 import '../../primitives/utils.dart';
 import '../theme.dart';
+import '../utils.dart';
 import 'table_controller.dart';
 import 'table_data.dart';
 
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/shared/table/column_widths.dart
```diff
@@ -105,42 +106,32 @@ extension FlatColumnWidthExtension<T> on FlatTableController<T> {
...
-        // with the longest text, but because of variable-width fonts, this is
-        // not always true.
-        width += _defaultColumnWidth;
+        // Note: this is assuming that we're always using a monospace font in
+        // our tree tables. This will almost certainly cause overflows if we
+        // try to use non-monospace fonts.
+        //
+        // `fontFactor` was determined through trial and error to roughly
+        // approximate the width of a single character.
+        const fontFactor = 9.0;
```
kenzieschmoll, line 124: nit: rename to characterWidth - the `scaleByFontFactor` method takes the original width and performs the scaling of that width by `ideTheme.fontSizeFactor`

packages/devtools_app/test/shared/table_test.dart
```diff
@@ -962,6 +962,52 @@ void main() {
...
+        'All work',
+        'and no play',
+        'makes Ben',
+        'a dull boy',
+        // ignore: no_adjacent_strings_in_list
+        'The quick brown fox jumped over the lazy dog, although the fox '
+        "couldn't jump very high and the dog was very, very small, so it really"
+        " wasn't much of an achievement on the fox's part, so I'm not sure why "
+        "we're even talking about it."
+      ];
```
kenzieschmoll, line 977: This has now become the best easter egg in our codebase.

packages/devtools_app/test/shared/table_test.dart
```diff
@@ -962,6 +962,52 @@ void main() {
...
+      await tester.pumpWidget(
+        wrap(
+          SizedBox(
+            width: 200.0,
+            height: 200.0,
+            child: table,
+          ),
+        ),
+      );
+      expect(find.byWidget(table), findsOneWidget);
```
kenzieschmoll, line 1008: can you add a comment above this line that says, "if the table text overflowed, an exception would have been thrown."

kenzieschmoll (approved):
couple comments then lgtm

packages/devtools_app/test/shared/table_test.dart
bkonyi: I'm glad you appreciate it! 🤣

Issue #4728, opened 2022-11-11
Unconstrain our archive dependency


CoderDake:
An issue with archive is causing our build to break, see: https://github.com/brendan-duncan/archive/issues/221 

When that issue is resolved we should be able to remove the `archive: <3.3.3`  constraint that was  added in: https://github.com/flutter/devtools/compare/master...CoderDake:archieve?expand=1

kenzieschmoll:
This is still happening with package:archive 3.3.4. Constraint added back in https://github.com/flutter/devtools/pull/4751.

parlough:
Fixed by https://github.com/flutter/devtools/commit/d457d49002c97f4320f2e5adbbe6272eb6f2035e

PR #4730, opened 2022-11-12
Remove `package:archive` constraint

elliette:
The newest version of `package:archive` has been released, therefore it should be safe to remove the constraint.

See: https://github.com/brendan-duncan/archive/issues/221


kenzieschmoll (approved):

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/4728

PR #4735, opened 2022-11-14
Fix some offline import issues with the Performance page.

kenzieschmoll:


packages/devtools_app/lib/src/screens/performance/tabbed_performance_view.dart
```diff
@@ -77,12 +77,24 @@ class _TabbedPerformanceViewState extends State<TabbedPerformanceView>
...
+
+    var showFrameAnalysis = isFlutterApp;
+    var showRasterStats = isFlutterApp;
+    var showRebuildStats = FeatureFlags.widgetRebuildstats && isFlutterApp;
+    final offlineData = controller.offlinePerformanceData;
+    if (isOffline && offlineData != null) {
+      showFrameAnalysis = showFrameAnalysis && offlineData.frames.isNotEmpty;
+      showRasterStats = showRasterStats && offlineData.rasterStats != null;
+      showRebuildStats = showRebuildStats &&
+          offlineData.rebuildCountModel.locationStats.value.isNotEmpty;
```
kenzieschmoll, line 91: @jacob314 can you verify whether this is the proper "empty" state to check here?

packages/devtools_app/lib/src/screens/performance/panes/raster_stats/raster_stats_model.dart
```diff
@@ -15,21 +15,16 @@ class RasterStats {
...
-      (json[_frameWidthKey] as num).toDouble(),
-      (json[_frameHeightKey] as num).toDouble(),
-    );
-
+    Size? originalFrameSize;
+    final originalWidth = json[_frameWidthKey] as num?;
+    final originalHeight = json[_frameHeightKey] as num?;
+    if (originalHeight != null && originalWidth != null) {
+      originalFrameSize = Size(
+        (json[_frameWidthKey] as num).toDouble(),
```
jacob314, line 24: use originalWidth and originalHeight on lines 24 and 25 to avoid the `as` checks.

packages/devtools_app/lib/src/screens/performance/tabbed_performance_view.dart
jacob314: I would add a rebuildCountModel.isNotEmpty getter.
_rebuildsForFrame.isNotEmpty is probably the cleanest implementation.

kenzieschmoll: Is it possible that we have rebuilds that are not associated with a flutter frame?

jacob314: no. all rebuilds have an associated frame. It is plausible in the future that the aggregate counts could however be zero even though there are rebuilds for a frame as a user could clear the total aggregate counts (e.g. by manually clicking the clear/reset button in the UI).

PR #4744, opened 2022-11-15
Fix beta testing instructions.

polina-c:
1. In some shells # and .git are interpreted as mistake
2. ';' looks weird for some users and is not necessary
3. We do not need flags to enable beta features, as they are enabled in non-release mode by default now
4. Added instructions to install jq

### DIFF
--- a/BETA_TESTING.md
+++ b/BETA_TESTING.md
@@ -10,18 +10,15 @@ If you want to make code changes, follow [contributing guidance](https://github.
 
 2. In your terminal `cd` to a folder where you want to clone devtools, and that does not have subfolder `devtools` yet.
 
-3. Start DevTools:
+3. Clone the repo, get needed Flutter version to local folder and start DevTools:
 
 ```bash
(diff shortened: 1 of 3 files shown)


BETA_TESTING.md
```diff
@@ -10,18 +10,15 @@ If you want to make code changes, follow [contributing guidance](https://github.
...
+git clone https://github.com/flutter/devtools.git
 
-# Get local Flutter of the correct version:
-./devtools/tool/update_flutter_sdk.sh;
+./devtools/tool/update_flutter_sdk.sh
 
-# Start the application:
-cd devtools/packages/devtools_app;
-../../tool/flutter-sdk/bin/flutter run -d chrome --dart-define=enable_experiments=true;
+cd devtools/packages/devtools_app
```
CoderDake, line 20: nit: leaving the semi colons in allows the reader to copy and paste the whole block if they would like to

CoderDake (approved):

PR #4760, opened 2022-11-16
Update Debugger screen to support more VM service types

bkonyi:
The variables viewer now has support for displaying and inspecting the following types:

  - `Expandos`
  - `Types`
  - `TypeArguments`
  - `Parameters`
  - `Closures` + closure `Context`s
  - `WeakProperty`
  - `Function`
  - `FunctionType`
  - `ReceivePort`
  - `Closure`
  - `RegExp`

### DIFF
--- a/packages/devtools_app/lib/src/shared/object_tree.dart
+++ b/packages/devtools_app/lib/src/shared/object_tree.dart
@@ -120,7 +120,7 @@ Future<void> buildVariablesTree(
         count: variable.childCount,
       );
       if (result is Instance) {
-        switch (result.kind!) {
+        switch (result.kind) {
           case InstanceKind.kMap:
             variable.addAllChildren(
               _createVariablesForAssociations(result, isolateRef),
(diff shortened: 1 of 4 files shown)


elliette:
FYI it looks like this should fix https://github.com/flutter/devtools/issues/4793 🎉

packages/devtools_app/lib/src/shared/object_tree.dart
```diff
@@ -92,45 +93,106 @@ Future<void> buildVariablesTree(
...
-          );
-        } else if (result.elements != null) {
-          variable
-              .addAllChildren(_createVariablesForElements(result, isolateRef));
-        } else if (result.bytes != null) {
-          variable.addAllChildren(_createVariablesForBytes(result, isolateRef));
-          // Check fields last, as all instanceRefs may have a non-null fields
-          // with no entries.
-        } else if (result.fields != null) {
+        switch (result.kind!) {
```
elliette, line 123: I'm nervous that this will throw if `kind` is `null`.  Do we need the `!`? I think if we remove it we should hit the `default` case instead of throwing an error.

bkonyi: `kind` should never be null if `result is Instance`, but I don't think the assertion is necessary if we'll just fall through to the default case anyway.

PR #4765, opened 2022-11-16
UX improvements for memory screen.

polina-c:
See http://go/ux-improvements-devtools

1. Rename tabs to 'Profile, Diff, Trace'
2. Allocation tracing: rename to 'Delta' with tooltip ‘Number of instances, allocated after the class was selected for tracing’
3. remove the class at the beginning of the path, because it is always the same class, that is selected in class table
4. no-data messages: align top-left instead of center, as it is how desktop apps should be structured
5. display Context as "Closure Context", to avoid confusion with BuildContext

Before:

Screenshot 2022-11-16 at 4 36 51 PM

After:
Screenshot 2022-11-16 at 4 19 03 PM


kenzieschmoll:
DBC - can we make sure the text on the selected row is white? The class column has the wrong selected text color

polina-c:
> 

done

kenzieschmoll:
> no-data messages: align top-left instead of center, as it is how desktop apps should be structured

can you give a screenshot of this? Also was this discussed? We have centered "empty" messages and "instructions" across DevTools, so we should be consistent.

packages/devtools_app/test/memory/allocation_tracing/allocation_profile_tracing_view_test.dart
```diff
@@ -142,9 +142,9 @@ void main() {
       final refresh = find.text('Refresh');
       expect(refresh, findsOneWidget);
 
-      expect(find.text('Trace'), findsOneWidget);
+      expect(find.text('Trace'), findsNWidgets(2));
```
kenzieschmoll, line 145: can you add a comment describing why there are two?

polina-c:
> > no-data messages: align top-left instead of center, as it is how desktop apps should be structured
> 
> can you give a screenshot of this? Also was this discussed? We have centered "empty" messages and "instructions" across DevTools, so we should be consistent.


It was not discussed. If it is a big change we may want to postpone it to other PR.

In general I believe the principle should be: the no-content message should be aligned the same way as content. 

I noticed when I work with the memory screen, I do not see the message in the center, because my eyes search for the content at the top left.

Screenshot 2022-11-17 at 2 01 17 PM


Screenshot 2022-11-17 at 2 01 58 PM


Screenshot 2022-11-17 at 2 02 19 PM

packages/devtools_app/test/memory/allocation_tracing/allocation_profile_tracing_view_test.dart
polina-c: done

kenzieschmoll:
> It was not discussed. If it is a big change we may want to postpone it to other PR.

I recommend we move to another PR. IMO, the help text looks better centered, but let's consult UX.

polina-c:
> > It was not discussed. If it is a big change we may want to postpone it to other PR.
> 
> I recommend we move to another PR. IMO, the help text looks better centered, but let's consult UX.

Reverted

packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
```diff
@@ -190,11 +191,24 @@ class ClassOnlyHeapPath {
...
+    var result = data.join(delimiter);
+    if (!skipObject) return result;
+
+    if (inverted) {
+      result = '$delimiter$result'.trim();
+    } else {
+      result = '$result$delimiter'.trim();
+    }
+
+    return result;
```
kenzieschmoll, line 210: This method can be simplified:
```dart
data = data.joinWith(delimiter);
if (inverted) data = data.reversed;
if (skipObject) data.removeAt(inverted ? 0 : data.length -1);
return data;
```

kenzieschmoll (approved):
One comment and then lgtm

PR #4766, opened 2022-11-17
Hook up scroll-to-time-range functionality for Perfetto trace viewer

kenzieschmoll:
This PR hooks up the scroll to time range functionality and does some refactoring to support testing on the web. We cannot generate a mock for a class that uses conditional imports.

There is now an interface class PerfettoController and two impl classes PerfetoControllerImpl (one for desktop and one for web).

perfetto-scroll-to-frame

Tests will be written as a follow up.

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/perfetto_controller.dart
+++ b/packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/perfetto_controller.dart
@@ -6,7 +6,6 @@ import '../../../../../primitives/auto_dispose.dart';
 import '../../../../../primitives/trace_event.dart';
 import '../../../../../primitives/utils.dart';
 import '../../../performance_controller.dart';
-import '../timeline_events_controller.dart';
 import '_perfetto_controller_desktop.dart'
     if (dart.library.html) '_perfetto_controller_web.dart';
 import 'perfetto_event_processor.dart';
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/perfetto_controller.dart
```diff
@@ -0,0 +1,39 @@
...
+}
+
+abstract class PerfettoController extends DisposableController {
+  PerfettoController(this.performanceController) {
+    processor = PerfettoEventProcessor(performanceController);
+  }
+
+  String get viewId => '';
+
+  final PerformanceController performanceController;
```
polina-c, line 26: Does PerfettoController really need the entire PerformanceController?

Would it be better to narrow it to some sub-functionality of the PerformanceController, to make the PerfettoController's contract clearer for the code reader?

packages/devtools_app/lib/src/screens/performance/panes/timeline_events/timeline_events_controller.dart
```diff
@@ -327,6 +325,33 @@ class TimelineEventsController extends PerformanceFeatureController
     });
   }
 
+  Future<void> _perfettoSelectFrame(FlutterFrame frame) async {
+    if (!offlineController.offlineMode.value) {
```
polina-c, line 329: Would early return make it more readable?

if (offlineController.offlineMode.value) return;

kenzieschmoll: no because there is a call to `perfettoController.scrollToTimeRange` at the end of the method that needs to be hit regardless of whether we are in offline mode or not

packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/perfetto_controller.dart
kenzieschmoll: changed to reference TimelineEventsController instead

polina-c (approved):

PR #4767, opened 2022-11-17
Add config for tests.

polina-c:


packages/.vscode/launch.json
```diff
@@ -4,6 +4,12 @@
     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
     "version": "0.2.0",
     "configurations": [
+        {
+            "name": "opened test or devtools",
```
kenzieschmoll, line 8: I don't understand what this is for? Is this for running a test or for running DevTools? Why both?

polina-c: vscode will try to guess what you want. If a test is opened, it will run the test. Otherwise, devtools.

DanTup: Because here's no `"program"` defined here, it will cause the Dart plugin to infer what you wanted to do (the same as if you press `F5` without any launch.json at all). It will look at your open file and if it's runnable (eg. is it inside a `test`, `integration_test`, `bin` or `tool` folder) it will run it. If not, it will then try some default locations (`lib/main.dart`, `bin/main.dart`) depending on the project type.

I usually have an entry like this (although without a `cwd`) where I have `launch.json`s (I just call it "Dart") as a way of selecting a context-sensitive "just do the likely thing" config for general work and pressing F5, and use the others for when I want to do something specific regardless of my active file/project.

kenzieschmoll (approved):

PR #4774, opened 2022-11-18
Decouple screen ids and metrics from screen main widgets.

polina-c:


packages/devtools_app/lib/src/primitives/simple_items.dart
```diff
@@ -28,7 +28,16 @@ class PackagePrefixes {
...
+  static const simple = 'simple';
+
+  static const appSize = 'app-size';
+  static const debugger = 'debugger';
+  static const provider = 'provider';
   static const inspector = 'inspector';
+  static const vmTools = 'vm-tools';
+  static const isolateStatistics = 'isolate-statistics';
+  static const objectInspectorView = 'object-inspector-view';
+  static const vmStatistics = 'vm-statistics';
```
kenzieschmoll, line 40: can we create a separate class (or maybe this class should really be an enhanced enum?) for the vm developer views? these aren't really "screens" in the sense like all the other "screens", which are top level devtools tabs. The vmTools entry should still be in ScreenIds since it is a top level Devtools tab

polina-c: true
reverted

kenzieschmoll (approved):

PR #4811, opened 2022-11-18
Make chart legend visible by default.

polina-c:
To unblock UXR, make chart legend visible by default and move from overlay to the screen surface.
See: http://go/ux-improvements-devtools, row 5

https://user-images.githubusercontent.com/12115586/202816716-ef201e64-45b6-4f2d-ae2c-b404baac7ecf.mov


kenzieschmoll:
I worry that this takes up too much vertical space. Is there any way the "Memory Legend" and "Android Legend" sections could be split into two columns like the "Events Legend" section?

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
```diff
@@ -140,9 +76,9 @@ class MemoryController extends DisposableController
   final _shouldShowLeaksTab = ValueNotifier<bool>(false);
   ValueListenable<bool> get shouldShowLeaksTab => _shouldShowLeaksTab;
 
-  ValueListenable get legendVisibleNotifier => _legendVisibleNotifier;
+  ValueListenable<bool> get legendVisibleNotifier => _legendVisibleNotifier;
 
-  final _legendVisibleNotifier = ValueNotifier<bool>(false);
+  final _legendVisibleNotifier = ValueNotifier<bool>(true);
```
kenzieschmoll, line 81: should true really be the default. That will make the UI very busy IMO

packages/devtools_app/lib/src/screens/memory/panes/chart/chart_pane.dart
```diff
@@ -230,6 +232,22 @@ class _MemoryChartPaneState extends State<MemoryChartPane>
...
               ),
+              // The legend.
+              DualValueListenableBuilder<bool, bool>(
+                firstListenable: controller.legendVisibleNotifier,
+                secondListenable: controller.isAndroidChartVisibleNotifier,
+                builder: (_, isLegendVisible, isAndroidChartVisible, __) {
+                  if (!isLegendVisible) return const SizedBox.shrink();
+                  return Padding(
+                    padding: const EdgeInsets.only(right: denseSpacing),
+                    child: MemoryChartWidget(
```
kenzieschmoll, line 243: rename to MemoryLegend

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
polina-c: requested by Mindy, because of some UXR

@melyndahoover , can you add details?

packages/devtools_app/lib/src/screens/memory/panes/chart/legend.dart
```diff
@@ -14,6 +14,102 @@ import 'memory_charts.dart';
...
+    final events = eventLegendContent(colorScheme.isLight);
+    legendRows.add(
+      Container(
+        padding: legendTitlePadding,
+        child: Text('Events Legend', style: legendHeading),
+      ),
+    );
+
+    var iterator = events.entries.iterator;
+    while (iterator.moveNext()) {
```
kenzieschmoll, line 44: nit: use a for loop instead

packages/devtools_app/lib/src/screens/memory/panes/chart/legend.dart
```diff
@@ -14,6 +14,102 @@ import 'memory_charts.dart';
...
+    final vms = vmLegendContent(chartController.vm);
+    legendRows.add(
+      Container(
+        padding: legendTitlePadding,
+        child: Text('Memory Legend', style: legendHeading),
+      ),
+    );
+
+    iterator = vms.entries.iterator;
+    while (iterator.moveNext()) {
```
kenzieschmoll, line 65: same here and elsewhere

packages/devtools_app/lib/src/screens/memory/panes/chart/legend.dart
```diff
@@ -14,6 +14,102 @@ import 'memory_charts.dart';
...
+            entry1: legendEntry,
+            chartController: chartController,
+          ),
+        );
+      }
+    }
+
+    return Container(
+      width: legendWidth,
+      height: isAndroidVisible ? legendHeight2Charts : legendHeight1Chart,
```
kenzieschmoll, line 98: we shouldn't need to specify the height here. If should match the height of its content

packages/devtools_app/lib/src/screens/memory/panes/chart/legend.dart
```diff
@@ -14,6 +14,102 @@ import 'memory_charts.dart';
...
+      padding: const EdgeInsets.fromLTRB(0, 5, 5, 8),
```
kenzieschmoll, line 99: use padding consts from theme.dart. denseSpacing, densePadding, etc.

Also, we shouldn't need bottom padding here

packages/devtools_app/lib/src/screens/memory/panes/chart/legend.dart
```diff
@@ -14,6 +14,102 @@ import 'memory_charts.dart';
...
+      decoration: BoxDecoration(
+        color: colorScheme.defaultBackgroundColor,
+        border: Border.all(color: theme.focusColor),
+        borderRadius: BorderRadius.circular(10.0),
```
kenzieschmoll, line 103: use a default radius value from theme.dart for consistency

polina-c:
> I worry that this takes up too much vertical space. Is there any way the "Memory Legend" and "Android Legend" sections could be split into two columns like the "Events Legend" section?

It increases height of the chart area by just ~5 pixels. Not big deal assuming legend can be hidden.

packages/devtools_app/lib/src/screens/memory/panes/chart/primitives.dart
```diff
@@ -0,0 +1,68 @@
...
+
+enum ChartType {
+  DartHeaps,
+  AndroidHeaps,
+}
+
+/// Automatic pruning of collected memory statistics (plotted) full data is
+/// still retained. Default is the best view each tick is 10 pixels, the
+/// width of an event symbol e.g., snapshot, monitor, etc.
+enum ChartInterval {
```
kenzieschmoll, line 13: make this an enhanced enum so that we don't have to have methods `chartDuration` and `displayDuration`. See this recent PR for an example of this: https://github.com/flutter/devtools/pull/4803/files#diff-6f087d6cdc6c21288a3321f6fbf556393aee9f2bcf167ab040d21fd44887d4c3

packages/devtools_app/lib/src/screens/memory/panes/chart/legend.dart
polina-c: not in this loop as it is doing iterator.moveNext() in the body

packages/devtools_app/lib/src/screens/memory/panes/chart/legend.dart
polina-c: without height the behavior is weird:  

https://user-images.githubusercontent.com/12115586/202821852-ab5aa99a-a2ee-4355-8820-53f85af66256.mov

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
melyndahoover: This was a suggestion based on an informal cognitive walk through. Essentially, it is not very discoverable if it is hidden by default. But perhaps there is a compromise. From the few examples I have seen of people using this tool, it seems that the majority of the markings defined in the legend aren't used very often. Is there a common subset that could be displayed by default perhaps? Also, it lookes like there are some items with missing symbols under "memory legend" are those necessary to have?

polina-c: @kenzieschmoll , what you think about removing items without symbols? I did not notice them till Mindy's comment. Sonds as a good option for me.

packages/devtools_app/lib/src/screens/memory/panes/chart/legend.dart
kenzieschmoll: Okay, not a huge deal to leave. But I suspect there were some errors in the console when you were seeing this. Might be an unbounded scrollview error or something like that.

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
kenzieschmoll: These aren't missing a symbol intentionally. There must be an error in how the legend is be rendered, which we should fix. Every item that is broken has the parameter 'dashed: true'.

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
```diff
@@ -136,7 +136,7 @@ class MemoryController extends DisposableController
   /// Default is to display default tick width based on width of chart of the collected
   /// data in the chart.
   final _displayIntervalNotifier =
-      ValueNotifier<ChartInterval>(ChartInterval.Default);
+      ValueNotifier<ChartInterval>(ChartInterval.theDefault);
```
kenzieschmoll, line 139: why "the" default and not just default?

packages/devtools_app/lib/src/screens/memory/panes/chart/legend.dart
kenzieschmoll: Oh can you try to add MainAxisSize.min to the Column below? That should allow you to get rid of the explicit height

kenzieschmoll (approved):
two more comments then lgtm

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
polina-c: because 'default' is keyword

packages/devtools_app/lib/src/screens/memory/panes/chart/legend.dart
polina-c: Did not help:
```
FlutterError (RenderFlex children have non-zero flex but incoming height constraints are unbounded.
When a column is in a parent that does not provide a finite height constraint, for example if it is in a vertical scrollable, it will try to shrink-wrap its children along the vertical axis. Setting a flex on a child (e.g. using Expanded) indicates that the child is to expand to fill the remaining space in the vertical direction.
These two directives are mutually exclusive. If a parent is to shrink-wrap its child, the child cannot simultaneously expand to fit its parent.
Consider setting mainAxisSize to MainAxisSize.min and using FlexFit.loose fits for the flexible children (using Flexible rather than Expanded). This will allow the flexible children to size themselves to less than the infinite remaining space they would otherwise be forced to take, and then will cause the RenderFlex to shrink-wrap the children rather than expanding to fit the maximum constraints provided by the parent.
If this message did not help you determine the problem, consider using debugDumpRenderTree():
  https://flutter.dev/debugging/#rendering-layer
  http://api.flutter.dev/flutter/rendering/debugDumpRenderTree.html
The affected RenderFlex is:
  RenderFlex#721b6 relayoutBoundary=up11 NEEDS-LAYOUT NEEDS-PAINT NEEDS-COMPOSITING-BITS-UPDATE(creator: Column ← Padding ← DecoratedBox ← ConstrainedBox ← Container ← MemoryChartLegend ← Padding ← DualValueListenableBuilder<bool, bool> ← Row ← Semantics ← _FocusMarker ← Focus ← ⋯, parentData: offset=Offset(1.0, 5.0) (can use size), constraints: BoxConstraints(w=194.0, 0.0<=h<=Infinity), size: Size(194.0, 194.0), direction: vertical, mainAxisAlignment: start, mainAxisSize: min, crossAxisAlignment: start, textDirection: ltr, verticalDirection: down)
The creator information is set to:
  Column ← Padding ← DecoratedBox ← ConstrainedBox ← Container ← MemoryChartLegend ← Padding ← DualValueListenableBuilder<bool, bool> ← Row ← Semantics ← _FocusMarker ← Focus ← ⋯
The nearest ancestor providing an unbounded width constraint is: RenderFlex#d32b0 relayoutBoundary=up4 NEEDS-LAYOUT NEEDS-PAINT NEEDS-COMPOSITING-BITS-UPDATE:
  needs compositing
  creator: Column-[GlobalKey#7ff88 Chart Hover] ← MemoryBody ← Expanded ← Column ← BannerMessages ← Expanded ← Column ← ValueListenableBuilder<bool> ← AnalyticsPrompt ← _FocusMarker ← Semantics ← FocusScope ← ⋯
  parentData: offset=Offset(0.0, 0.0); flex=1; fit=FlexFit.tight (can use size)
  constraints: BoxConstraints(0.0<=w<=834.0, h=615.0)
  size: Size(834.0, 615.0)
  direction: vertical
  mainAxisAlignment: start
  mainAxisSize: max
  crossAxisAlignment: center
  verticalDirection: down
See also: https://flutter.dev/layout/
If none of the above helps enough to fix this problem, please don't hesitate to file a bug:
  https://github.com/flutter/flutter/issues/new?template=2_bug.md)
```

Issue #4786, opened 2022-11-18
Bad column width calculation for "Method" column in CPU profile.


kenzieschmoll:
This happened when switching from a fully expanded profile to "Group by: User Tag"
![Screen Shot 2022-11-18 at 9 51 05 AM](https://user-images.githubusercontent.com/43759233/202770299-d0a07da8-1085-49ec-8301-2da2d049c1c6.png)

PR #4818, opened 2022-11-21
Fix column width calculations

bkonyi:
We weren't taking the expansion caret into account, causing the largest node to be slightly cut off.

Fixes https://github.com/flutter/devtools/issues/4786


bkonyi:
It looks like the test I added when I made this change doesn't actually catch overflows since we have `TextOverflow.ellipsis` set. Do you have any suggestions on how to modify that test?

kenzieschmoll:
> Do you have any suggestions on how to modify that test?

You could find all the text widgets and verify none of them contain the substring '...'

packages/devtools_app/lib/src/shared/table/column_widths.dart
```diff
@@ -121,17 +121,22 @@ extension TreeColumnWidthExtension<T extends TreeNode<T>>
...
           if (width > longestWidth) {
             longestWidth = width;
           }
         }
-        width = longestWidth;
+        // Add two characters to account for expansion caret and ensure we
+        // always display the full column title.
+        width = max(
+              max(longestWidth, column.minWidthPx ?? 0),
+              column.title.length,
```
kenzieschmoll, line 137: should this be `column.title.length *  characterWidth`?

packages/devtools_app/test/shared/table_test.dart
```diff
@@ -1006,7 +1006,34 @@ void main() {
           ),
         ),
       );
+
       expect(find.byWidget(table), findsOneWidget);
+      expect(
```
kenzieschmoll, line 1011: just to make sure, does this part of the test you added fail without the fix you added in this PR? Might be good to add a comment "Regression test for (<github issue link>)" to make that clear

packages/devtools_app/lib/src/shared/table/column_widths.dart
```diff
@@ -121,17 +121,22 @@ extension TreeColumnWidthExtension<T extends TreeNode<T>>
...
           }
         }
-        width = longestWidth;
+        // Add two characters to account for expansion caret and ensure we
+        // always display the full column title.
+        width = max(
+              max(longestWidth, column.minWidthPx ?? 0),
+              column.title.length * characterWidth,
+            ) +
+            characterWidth * 2;
```
kenzieschmoll, line 139: do we need to estimate here or do we know the exact size of the expansion caret?

kenzieschmoll (approved):
a couple more comments, but overall lgtm

packages/devtools_app/lib/src/shared/table/column_widths.dart
bkonyi: This is the exact size (one character for `^` and one for the whitespace).

Issue #4793, opened 2022-11-18
Expanding a list in a variable hovercard shows blank grey box


elliette:
Some examples:

![Screenshot 2022-11-18 at 10 26 57 AM](https://user-images.githubusercontent.com/21270878/202776988-51aeab49-8223-4f31-9171-ca09ced8762e.png)

![Screenshot 2022-11-18 at 10 27 43 AM](https://user-images.githubusercontent.com/21270878/202777001-7c57a5a0-92ac-41c9-8a09-fcf3ffc8b25b.png)

elliette:
Note: does not seem to be the case for all lists

![Screenshot 2022-11-18 at 10 43 33 AM](https://user-images.githubusercontent.com/21270878/202779517-7a13a7be-f2cd-4057-a806-819a7ab4539f.png)

kenzieschmoll:
were there any exceptions in DevTools when you saw the grey box? The grey box usually comes up when there is a flutter error. I suspect if you run DevTools in debug mode you'd see a red error box instead of grey

PR #4825, opened 2022-11-21
Handle possible `null` values when expanding variables 

elliette:
Fixes https://github.com/flutter/devtools/issues/4793

A few errors were being thrown if variables contained any `null` values, which caused the expanded variable list to be a grey box. Now, we show `"Unknown"` if we are missing a the variable display name. 

Screenshot 2022-11-21 at 2 19 35 PM

Before:

Screenshot 2022-11-18 at 10 33 32 AM


elliette:
Closing in favor of https://github.com/flutter/devtools/pull/4760, which fixes the same issue.

PR #4827, opened 2022-11-22
Polish Perfetto iFrame lifecycle

kenzieschmoll:
This PR fixes the lifecycle of the Perfetto iFrame so that code that needs to be called in the widget lifecycle is called at the appropriate times. Before this PR, the Perfetto widget would not properly load data when navigating away from the Performance page and back to it. By making the Perfetto widget stateful and loading the proper data in initState, this problem is solved. 

All of the "deleted" code in `_perfetto_controller_web.dart` was moved over to `_perfetto_web.dart` into the `_PerfettoViewController` class.


packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_web.dart
```diff
@@ -2,25 +2,305 @@
...
   }
+
+  @override
+  void dispose() {
+    _viewController.dispose();
+    super.dispose();
+  }
+}
+
+class _PerfettoViewController extends DisposableController
```
kenzieschmoll, line 87: the code in this class was all moved from PerfettoController. The one area that should be looked as new is the init method - it was partially moved from PerfettoController.

polina-c (approved):

PR #4828, opened 2022-11-22
Make feature flags more readable/supportable.

polina-c:


packages/devtools_app/lib/src/primitives/feature_flags.dart
```diff
@@ -18,7 +18,9 @@ import '../shared/globals.dart';
 ///   ]
 const bool _kEnableExperiments = bool.fromEnvironment('enable_experiments');
 
+@visibleForTesting
```
CoderDake, line 21: since this isn't hidden is `visibleForTesting` necessary here?

polina-c: visibleForTesting protects public members from use in non-test environment: https://api.flutter.dev/flutter/meta/visibleForTesting-constant.html

It cannot make public members private.

packages/devtools_app/lib/src/primitives/feature_flags.dart
```diff
@@ -18,7 +18,9 @@ import '../shared/globals.dart';
...
 bool enableExperiments = _kEnableExperiments;
```
kenzieschmoll, line 22: why can't we just do this:
`final bool enableExperiments = _kEnableExperiments || !kReleaseMode;`

Then we don't need a method `setExperimentsEnabled` at all

polina-c: It is for protection.

enableExperiments is private and protected from switching back and force.
It can be flipped only once to true.

kenzieschmoll: we don't need it to be switched at all though. We don't need to enable experiments from main.dart, we can just set the value from the start in feature_flags.dart. The following code would still allow us to disable experiments for tests if that is a concern:

```
@visibleForTesting
bool enableExperiments = _kEnableExperiments || !kReleaseMode;
```

polina-c: How can we disable experiments for all tests? As I know there is no method like 'setupForAllTestLibraries'. And we do not want to update every single test, right?

kenzieschmoll: It's a tradeoff we have to consider here. I'm not a huge fan of making our production code messier for the sake of tests, but the tradeoff would be to manually disable experiments for tests that shouldn't have them on, which doesn't scale well. I also don't know of a flutter test configuration where we could set a value like this for all tests. Let's leave it as you have it.

kenzieschmoll (approved):

packages/devtools_app/lib/src/primitives/feature_flags.dart
kenzieschmoll: After discussing this a little bit since we made the original change, I think option 2 is our best step forward. In an offline discussion with @jacob314, he also voiced concerns that having experiments on by default could prevent us from catching issues that only show up when experiments are disabled, which is what our users will see.

kenzieschmoll: Implemented here: https://github.com/flutter/devtools/pull/4861

PR #4837, opened 2022-11-28
Add initial support for non-query param navigation state

bkonyi:
This change adds support for providing additional state when navigating through DevTools via `DevToolsRouterDelegate` without polluting the query parameters with extra entries. This state is included in the navigation history.

Controllers can mixin the new `RouteStateHandlerMixin` and override `onRouteStateUpdate(DevToolsNavigationState)` to handle changes in the current DevTools navigation state. To handle state changes, `subscribeToRouterEvents` must first be called on the controller.

This change also adds support for jumping to a source location in the debugger from URIs in the CPU profiler.

### DIFF
--- a/packages/devtools_app/lib/src/screens/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/codeview.dart
@@ -168,56 +168,65 @@ class _CodeViewState extends State<CodeView>
       return;
     }
 
-    if (!verticalController.hasAttachedControllers) {
-      // TODO(devoncarew): I'm uncertain why this occurs.
-      log('LinkedScrollControllerGroup has no attached controllers');
-      return;
-    }
-    final line = widget.codeViewController.scriptLocation.value?.location?.line;
-    if (line == null) {
-      // Don't scroll to top if we're just rebuilding the code view for the
-      // same script.
-      if (_lastScriptRef?.uri != scriptRef?.uri) {
-        // Default to scrolling to the top of the script.
+    void updateScrollPositionImpl() {
+      if (!verticalController.hasAttachedControllers) {
+        // TODO(devoncarew): I'm uncertain why this occurs.
+        log('LinkedScrollControllerGroup has no attached controllers');
+        return;
+      }
+      final line =
+          widget.codeViewController.scriptLocation.value?.location?.line;
+      if (line == null) {
+        // Don't scroll to top if we're just rebuilding the code view for the
+        // same script.
+        if (_lastScriptRef?.uri != scriptRef?.uri) {
+          // Default to scrolling to the top of the script.
+          if (animate) {
+            unawaited(
+              verticalController.animateTo(
+                0,
+                duration: longDuration,
+                curve: defaultCurve,
+              ),
+            );
+          } else {
+            verticalController.jumpTo(0);
+          }
+          _lastScriptRef = scriptRef;
+        }
+        return;
+      }
+
+      final position = verticalController.position;
+      final extent = position.extentInside;
+
+      // TODO(devoncarew): Adjust this so we don't scroll if we're already in the
+      // middle third of the screen.
+      final lineCount = parsedScript?.lineCount;
+      if (lineCount != null && lineCount * CodeView.rowHeight > extent) {
+        final lineIndex = line - 1;
+        final scrollPosition = lineIndex * CodeView.rowHeight -
+            ((extent - CodeView.rowHeight) / 2);
         if (animate) {
           unawaited(
             verticalController.animateTo(
-              0,
+              scrollPosition,
               duration: longDuration,
               curve: defaultCurve,
             ),
           );
         } else {
-          verticalController.jumpTo(0);
+          verticalController.jumpTo(scrollPosition);
         }
-        _lastScriptRef = scriptRef;
       }
-      return;
+      _lastScriptRef = scriptRef;
     }
 
-    final position = verticalController.position;
-    final extent = position.extentInside;
-
-    // TODO(devoncarew): Adjust this so we don't scroll if we're already in the
-    // middle third of the screen.
-    final lineCount = parsedScript?.lineCount;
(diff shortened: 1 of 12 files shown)


packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -785,12 +793,11 @@ class _LinesState extends State<Lines> with AutoDisposeMixin {
...
-          valueListenable: widget
-              .codeViewController.programExplorerController.outlineSelection,
-          builder: (context, outlineNode, _) {
-            final isFocusedLine =
-                (outlineNode?.location?.location?.line ?? -1) == lineNum;
+        return ValueListenableBuilder<int>(
+          valueListenable: widget.codeViewController.focusLine,
+          builder: (context, focusLine, _) {
+            final isFocusedLine = focusLine == lineNum;
+            //(outlineNode?.location?.location?.line ?? -1) == lineNum;
```
kenzieschmoll, line 800: remove this

packages/devtools_app/lib/src/screens/debugger/codeview_controller.dart
```diff
@@ -346,3 +385,40 @@ class ParsedScript {
...
+class CodeViewSourceLocationNavigationState extends DevToolsNavigationState {
+  CodeViewSourceLocationNavigationState({
+    required ScriptRef script,
+    required int line,
+  }) : super(
+          kind: type,
+          state: <String, String?>{
+            'scriptId': script.id,
+            'uri': script.uri,
+            'line': line.toString(),
```
kenzieschmoll, line 400: put these map keys into static consts and use them here and below

packages/devtools_app/lib/src/shared/routing.dart
```diff
@@ -185,6 +215,7 @@ class DevToolsRouterDelegate extends RouterDelegate<DevToolsRouteConfiguration>
     routes
       ..clear()
       ..add(configuration);
+    notifyListeners();
```
kenzieschmoll, line 218: this will have side effects in `setNewRoutePath` where `_replaceStack` was also called but without calling `notifyListeners` after.

packages/devtools_app/lib/src/shared/routing.dart
```diff
@@ -205,8 +236,13 @@ class DevToolsRouterDelegate extends RouterDelegate<DevToolsRouteConfiguration>
 
     final currentPage = currentConfiguration!.page;
     final newArgs = {...currentConfiguration!.args, ...argUpdates};
-    _replaceStack(DevToolsRouteConfiguration(currentPage, newArgs));
-    notifyListeners();
+    _replaceStack(
+      DevToolsRouteConfiguration(
+        currentPage,
+        newArgs,
+        currentConfiguration!.state,
```
kenzieschmoll, line 243: pull currentConfifuration out into a var before currentPage so it only has to be null asserted once

packages/devtools_app/lib/src/shared/routing.dart
```diff
@@ -215,4 +251,75 @@ class DevToolsRouterDelegate extends RouterDelegate<DevToolsRouteConfiguration>
         {...currentConfiguration!.args, ...?changes},
         {...currentConfiguration!.args},
       );
+
+  /// Checks whether applying [changes] over the current routes state will result
```
kenzieschmoll, line 255: super nit: r/routes/route's

packages/devtools_app/lib/src/shared/routing.dart
```diff
@@ -215,4 +251,75 @@ class DevToolsRouterDelegate extends RouterDelegate<DevToolsRouteConfiguration>
...
+  /// in any changes.
+  bool _changesState(DevToolsNavigationState? changes) {
+    return currentConfiguration!.state?.hasChanges(changes) ?? false;
+  }
+}
+
+class DevToolsNavigationState {
```
kenzieschmoll, line 262: add dart doc

packages/devtools_app/lib/src/shared/routing.dart
```diff
@@ -215,4 +251,75 @@ class DevToolsRouterDelegate extends RouterDelegate<DevToolsRouteConfiguration>
...
+    required Map<String, String?> state,
+  }) : _state = {
+          '_kind': kind,
+          ...state,
+        };
+
+  factory DevToolsNavigationState.fromJson(Map<String, dynamic> json) =>
+      DevToolsNavigationState._(json.cast<String, String?>());
+
+  DevToolsNavigationState._(this._state) : kind = _state['_kind']!;
```
kenzieschmoll, line 274: pull '_kind' into const. use here and above

packages/devtools_app/lib/src/shared/routing.dart
```diff
@@ -215,4 +251,75 @@ class DevToolsRouterDelegate extends RouterDelegate<DevToolsRouteConfiguration>
...
+    final state = _delegate?.currentConfiguration?.state;
+    if (state == null) return;
+    onRouteStateUpdate(state);
+  }
+
+  /// Perform operations based on changes in navigation state.
+  ///
+  /// This method is only invoked if [subscribeToRouterEvents] has been called on
+  /// this instance with a valid [DevToolsRouterDelegate].
+  void onRouteStateUpdate(DevToolsNavigationState state) {}
```
kenzieschmoll, line 324: can this be abstract? or another option, can we annotate with mustCallSuper and add an assert in here that `subscribeToRouterEvents` has been invoked for this instance

packages/devtools_app/lib/src/shared/routing.dart
bkonyi: Thought this was necessary, but just double checked and it seems like it isn't. Reverted the `notifyListeners` changes.

bkonyi:
Filed https://github.com/flutter/devtools/issues/4902 for future integration testing.

Issue #4597, opened 2022-10-08
Memory > Allocation Tracing: checkboxes are not clickable in profile mode


polina-c:
For profile mode we need to show empty tab with message like 'Run application in debug mode to enable allocation tracing'.

```
#0      _MemoryChartPaneState.didChangeDependencies.<anonymous closure> (package:devtools_app/src/screens/memory/panes/chart/chart_pane.dart:119:46)
package:devtools_app/src/screens/memory/panes/chart/chart_pane.dart:119
#3      ChartState.build.<anonymous closure>.<anonymous closure> (package:devtools_app/src/charts/chart.dart:105:36)
package:devtools_app/src/charts/chart.dart:105
#3      VmServiceWrapper.setTraceClassAllocation (package:devtools_app/src/service/vm_service_wrapper.dart:660:18)
#4      AllocationProfileTracingIsolateState.setAllocationTracingForClass (package:devtools_app/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_view_controller.dart:173:21)
#5      AllocationProfileTracingViewController.setAllocationTracingForClass (package:devtools_app/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_view_controller.dart:316:33)
#6      _TraceCheckBoxColumn.build.<anonymous closure> (package:devtools_app/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart:46:26)
#3      VmServiceWrapper.setTraceClassAllocation (package:devtools_app/src/service/vm_service_wrapper.dart:660:18)
#4      AllocationProfileTracingIsolateState.setAllocationTracingForClass (package:devtools_app/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_view_controller.dart:173:21)
#5      AllocationProfileTracingViewController.setAllocationTracingForClass (package:devtools_app/src/screens/memory/panes/allocation_tracing/allocation_profile_tracing_view_controller.dart:316:33)
#6      _TraceCheckBoxColumn.build.<anonymous closure> (package:devtools_app/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart:46:26)
#3      VmServiceWrapper.setTraceClassAllocation (package:devtools_app/src/service/vm_service_wrapper.dart:660:18)
```

polina-c:
@bkonyi

bkonyi:
[...]

PR #4839, opened 2022-11-28
Disable allocation tracing in profile mode

bkonyi:
Dart AOT applications don't currently support allocation tracing. This changes hides the "Trace" tab in the memory screen when connected to a profile mode Flutter application or an AOT Dart application.

Temporary fix for https://github.com/flutter/devtools/issues/4597

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/memory_tabs.dart
+++ b/packages/devtools_app/lib/src/screens/memory/memory_tabs.dart
@@ -6,6 +6,7 @@ import 'package:flutter/material.dart';
 
 import '../../primitives/auto_dispose_mixin.dart';
 import '../../shared/common_widgets.dart';
+import '../../shared/globals.dart';
 import '../../shared/theme.dart';
 import '../../shared/utils.dart';
 import '../../ui/search.dart';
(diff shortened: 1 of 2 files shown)


kenzieschmoll (approved):

packages/devtools_app/lib/src/screens/memory/memory_tabs.dart
```diff
@@ -107,7 +110,7 @@ class _MemoryTabsState extends State<MemoryTabs>
   @override
   Widget build(BuildContext context) {
     final themeData = Theme.of(context);
-
+    final isProfile = serviceManager.connectedApp?.isProfileBuildNow ?? false;
```
polina-c, line 113: How about making isProfile a getter, and maybe rename it to shouldShowTrace, so that we do not have logic duplicates?

polina-c (approved):
Couple comments and then LGTM

packages/devtools_app/lib/src/screens/memory/memory_tabs.dart
polina-c: If isProfile accidentally provide different values for both invocations, we will have UI failure, because number for tab headers will not match number of tabs.

Lets calculate it in initTabs and just use the value in build.

polina-c:
We want users to know that it is temporary that tracing is disabled for profile. If they do not know it, they will remember it and never try again to use it for profile.

How about showing message on the tab instead of hiding the tab?

PR #4841, opened 2022-11-28
Prevent _gtagEvent from being called in a non-release builds

kenzieschmoll:
`_gtagEvent` cannot be called in a test environment, as it requires the index.html file to be present (where the js function `getDevToolsPropertyID` is registered).

### DIFF
--- a/packages/devtools_app/lib/src/analytics/_analytics_web.dart
+++ b/packages/devtools_app/lib/src/analytics/_analytics_web.dart
@@ -10,7 +10,6 @@ library gtags;
 import 'dart:async';
 import 'dart:html';
 
-import 'package:flutter/foundation.dart';
 import 'package:js/js.dart';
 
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/analytics/_analytics_web.dart
```diff
@@ -383,14 +393,16 @@ void screen(
...
 ]) {
-  GTag.event(
-    screenName,
-    _gtagEvent(
-      event_category: analytics_constants.screenViewEvent,
-      value: value,
-      send_to: () => gaDevToolsPropertyId(),
-    ),
-  );
+  if (kReleaseMode || _debugAnalytics) {
```
polina-c, line 396: How about creating getter _analyticsEnabled, to avoid logic duplicates?

polina-c (approved):

PR #4846, opened 2022-11-29
Handle breaking changes in package:vm_service v10

derekxu16:
The `InboundReference` and `RetainingObject` types will be changed in `package:vm_service` version 10.0.0 (see [this CL](https://dart-review.googlesource.com/c/sdk/+/268521/14/runtime/vm/service/service.md)). This handles those changes. The also handles the optional parameters added to `getInstances` in `package:vm_service` version 10.1.0.

### DIFF
--- a/packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
+++ b/packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
@@ -520,7 +520,10 @@ class _RetainingObjectDescription extends StatelessWidget {
       const TextSpan(text: 'Retained by '),
     ];
 
-    if (object.parentField != null) {
+    if (object.parentField is int) {
+      assert((object.value as InstanceRef).kind == InstanceKind.kRecord);
+      entries.add(TextSpan(text: '\$${object.parentField} of '));
+    } else if (object.parentField != null) {
       entries.add(TextSpan(text: '${object.parentField} of '));
     }
 
@@ -553,13 +556,12 @@ class _RetainingObjectDescription extends StatelessWidget {
         ).buildTextSpan(context),
       );
     } else {
-      entries.addAll([
-        const TextSpan(text: 'of '),
+      entries.add(
         VmServiceObjectLink(
           object: object.value,
           onTap: onTap,
         ).buildTextSpan(context),
-      ]);
+      );
     }
     return SelectableText.rich(
       TextSpan(children: entries),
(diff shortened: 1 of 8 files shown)


packages/devtools_app/test/vm_developer/vm_developer_test_utils.dart
```diff
@@ -96,6 +96,14 @@ final testRetainingObjects = [
     parentListIndex: 1,
     parentField: 'fooParentField',
```
derekxu16, line 97: This test is used to check which property takes priority between `parentListIndex` and `parentField`, but it should be impossible for both of these properties ot appear on the same `RetainingObject`. Do we want to change this?

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -553,13 +556,12 @@ class _RetainingObjectDescription extends StatelessWidget {
         ).buildTextSpan(context),
       );
     } else {
-      entries.addAll([
-        const TextSpan(text: 'of '),
```
bkonyi, line 557: Why was this removed?

packages/devtools_app/pubspec.yaml
```diff
@@ -57,7 +57,7 @@ dependencies:
   string_scanner: ^1.1.0
   url_launcher: ^6.1.0
   url_launcher_web: ^2.0.6
-  vm_service: ^9.0.0
+  vm_service: '>=10.1.0'
```
bkonyi, line 60: This can be `^10.1.0`

packages/devtools_app/test/vm_developer/vm_developer_test_utils.dart
bkonyi: Yeah, if it doesn't make sense anymore, we should update the test.

packages/devtools_test/pubspec.yaml
```diff
@@ -27,7 +27,7 @@ dependencies:
   mockito: ^5.0.9
   path: ^1.8.0
   provider: ^6.0.2
-  vm_service: ^9.0.0
+  vm_service: '^10.1.0'
```
bkonyi, line 30: These quotations aren't necessary.

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
derekxu16: It creates two "of"s in a row, because these lines `entries` always ends with "of" after these lines https://github.com/flutter/devtools/pull/4846/files#diff-8483d57bb25f0c712d2792fc77a3f949a5f9be04fbb779e1e0e966f3a9fd3a64R523-R528

bkonyi (approved):

PR #4849, opened 2022-11-30
Update latest flutter candidate script to use simpler commands

CoderDake:
This was discussed/requested[ here](https://dart-review.git.corp.google.com/c/recipes/+/265320/1..2/recipes/devtools/package_devtools.py#b38)
 
 This change makes the script less dependant on non-standard commands, so it can be used in more environments


tool/latest_flutter_candidate.sh
```diff
@@ -7,25 +7,10 @@
...
-
-LATEST_VERSION=$(echo "$VERSIONS" |  sort --version-sort | tail  -n 1 )
-
-if [ -z ${LATEST_VERSION+x} ]; then
-    echo "Unable to get Latest flutter candidate version"
-    exit 1
-fi
-
-LATEST_FLUTTER_CANDIDATE="flutter-$LATEST_VERSION"
+LATEST_FLUTTER_CANDIDATE=$(git ls-remote https://dart.googlesource.com/external/github.com/flutter/flutter.git \
```
kenzieschmoll, line 10: this outputs refs/heads/ .... and we will need it to output flutter-version-candidate for our purposes. We may need two scripts: one that does this that the auto roller can use and another that uses this and strips refs/heads/ from the output for our purposes. Either that, or our other scripts that use this need to strip refs/heads/ from the output before they use it

CoderDake: I've update this so that it mimicks the old behaviour.
Do you know if it would be ok for the autoroller to just append refs/head on the branch name if it is needed?

kenzieschmoll: @athomas

athomas: Yes, that is fine.

tool/test/release_notes/release_note_classes_test.dart
```diff
@@ -0,0 +1,15 @@
+// import 'package:devtools_repo/release_note_classes.dart';
```
athomas, line 1: Is this file meant to be here?

kenzieschmoll: remove this file then LGTM

kenzieschmoll (approved):

tool/test/release_notes/release_note_classes_test.dart
CoderDake: this was accidentally added from another branch i was working on. Sorry for the confusion

tool/latest_flutter_candidate.sh
CoderDake: @athomas @kenzieschmoll I've updated the remote that we are touching to `git@github.com:flutter/flutter.git`
it looks like the other remote we were pointing to didn't have all of the refs for the candidate branches, so it was grabbing old candidate versions

tool/latest_flutter_candidate.sh
```diff
@@ -7,7 +7,7 @@
 # Any subsequent commands failure will cause this script to exit immediately
 set -e
 
-LATEST_FLUTTER_CANDIDATE=$(git ls-remote https://dart.googlesource.com/external/github.com/flutter/flutter.git \
+LATEST_FLUTTER_CANDIDATE=$(git ls-remote --heads --sort='-v:refname' git@github.com:flutter/flutter.git \
```
kenzieschmoll, line 10: will we run into rate limiting since we are accessing github now?

tool/latest_flutter_candidate.sh
CoderDake: ugh it looks like in order to use the flutter/flutter remote we'll have to:
 - generate a private key for the flutter/flutter repo, 
 - save it to the devtools repo as a secret token
 - use that token to auth against the flutter/flutter repo so that our `git ls-remote` command will have permission to fetch the refs

 Alternatively we could find out how to make sure the refs are up to date on https://dart.googlesource.com/external/github.com/flutter/flutter.git
 
 @kenzieschmoll @athomas do either of you have opinions about these options?

kenzieschmoll: > Alternatively we could find out how to make sure the refs are up to date on https://dart.googlesource.com/external/github.com/flutter/flutter.git

@athomas is there expected latency between flutter/flutter on github and flutter/flutter on dart.googlesource.com ?

athomas: Under normal circumstances, the latency should be very low. It's a mirror, so it's not instant.

The Dart mirror might be special in some way, I can check if we're still mirroring everything (we had some hacks around `main` branches before, not sure if those are still there). An alternative would be https://flutter.googlesource.com/mirrors/flutter/ which might be a more exact mirror.

kenzieschmoll (approved):

kenzieschmoll:
@athomas once this lands, can we push https://dart-review.git.corp.google.com/c/recipes/+/265320 forward?

PR #4890, opened 2022-12-02
Code metric fixes

jacob314:
Experiment applying most of the fixes suggested by the default dart_code_metrics configuration.

Feedback appreciated. Most of the switches to conditionals seem like an improvement but a few might be going overboard.

### DIFF
--- a/packages/devtools_app/lib/src/screens/debugger/call_stack.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/call_stack.dart
@@ -112,17 +112,15 @@ class _CallStackState extends State<CallStack>
       ),
     );
 
-    if (isAsyncBreak) {
-      return result;
-    } else {
-      return DevToolsTooltip(
-        message: locationDescription == null
-            ? frameDescription
-            : '$frameDescription $locationDescription',
-        waitDuration: tooltipWaitLong,
-        child: result,
-      );
-    }
+    return isAsyncBreak
+        ? result
+        : DevToolsTooltip(
+            message: locationDescription == null
+                ? frameDescription
+                : '$frameDescription $locationDescription',
+            waitDuration: tooltipWaitLong,
+            child: result,
+          );
   }
 
   Future<void> _onStackFrameSelected(StackFrameAndSourcePosition frame) async {
--- a/packages/devtools_app/lib/src/screens/debugger/debugger_model.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/debugger_model.dart
@@ -437,13 +437,11 @@ class ScriptRefUtils {
       ];
     }
 
-    if (parts.length > 1) {
-      return [
-        parts.first,
-        parts.sublist(1).join('/'),
-      ];
-    } else {
-      return parts;
-    }
+    return parts.length > 1
+        ? [
+            parts.first,
+            parts.sublist(1).join('/'),
+          ]
+        : parts;
   }
 }
--- a/packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
+++ b/packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
@@ -65,16 +65,14 @@ class InspectorController extends DisposableController
       onExpand: _onExpand,
       onClientActiveChange: _onClientChange,
     );
-    if (isSummaryTree) {
-      details = InspectorController(
-        inspectorTree: detailsTree!,
-        treeType: treeType,
-        parent: this,
-        isSummaryTree: false,
-      );
-    } else {
-      details = null;
-    }
+    details = isSummaryTree
+        ? InspectorController(
+            inspectorTree: detailsTree!,
+            treeType: treeType,
+            parent: this,
+            isSummaryTree: false,
+          )
+        : null;
 
     await serviceManager.onServiceAvailable;
 
@@ -807,7 +805,7 @@ class InspectorController extends DisposableController
   }
 
   void selectionChanged() {
-    if (visibleToUser == false) {
+    if (!visibleToUser) {
       return;
     }
 
--- a/packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
+++ b/packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
@@ -122,13 +122,11 @@ abstract class InspectorServiceBase extends DisposableController
   }
 
   Future<bool> invokeBoolServiceMethodNoArgs(String methodName) async {
-    if (useDaemonApi) {
-      return await invokeServiceMethodDaemonNoGroupArgs(methodName) == true;
-    } else {
-      return (await invokeServiceMethodObservatoryNoGroup(methodName))
-              ?.valueAsString ==
-          'true';
-    }
+    return useDaemonApi
+        ? await invokeServiceMethodDaemonNoGroupArgs(methodName) == true
+        : (await invokeServiceMethodObservatoryNoGroup(methodName))
+                ?.valueAsString ==
+            'true';
   }
 
   Future<Object?> invokeServiceMethodDaemonNoGroupArgs(
@@ -638,13 +636,11 @@ abstract class ObjectGroupBase implements Disposable {
     String methodName,
   ) async {
     if (disposed) return null;
-    if (useDaemonApi) {
-      return parseDiagnosticsNodeDaemon(invokeServiceMethodDaemon(methodName));
-    } else {
-      return parseDiagnosticsNodeObservatory(
-        invokeServiceMethodObservatory(methodName),
-      );
-    }
+    return useDaemonApi
+        ? parseDiagnosticsNodeDaemon(invokeServiceMethodDaemon(methodName))
+        : parseDiagnosticsNodeObservatory(
+            invokeServiceMethodObservatory(methodName),
+          );
   }
 
   Future<RemoteDiagnosticsNode?> invokeServiceMethodReturningNodeInspectorRef(
@@ -652,15 +648,13 @@ abstract class ObjectGroupBase implements Disposable {
     InspectorInstanceRef? ref,
   ) async {
     if (disposed) return null;
-    if (useDaemonApi) {
-      return parseDiagnosticsNodeDaemon(
-        invokeServiceMethodDaemonInspectorRef(methodName, ref),
-      );
-    } else {
-      return parseDiagnosticsNodeObservatory(
-        invokeServiceMethodObservatoryInspectorRef(methodName, ref),
-      );
-    }
+    return useDaemonApi
+        ? parseDiagnosticsNodeDaemon(
+            invokeServiceMethodDaemonInspectorRef(methodName, ref),
+          )
+        : parseDiagnosticsNodeObservatory(
+            invokeServiceMethodObservatoryInspectorRef(methodName, ref),
+          );
   }
 
   Future<RemoteDiagnosticsNode?> invokeServiceMethodWithArgReturningNode(
@@ -668,15 +662,13 @@ abstract class ObjectGroupBase implements Disposable {
     String arg,
   ) async {
     if (disposed) return null;
-    if (useDaemonApi) {
-      return parseDiagnosticsNodeDaemon(
-        invokeServiceMethodDaemonArg(methodName, arg, groupName),
-      );
-    } else {
-      return parseDiagnosticsNodeObservatory(
-        invokeServiceMethodObservatoryWithGroupName1(methodName, arg),
-      );
-    }
+    return useDaemonApi
+        ? parseDiagnosticsNodeDaemon(
+            invokeServiceMethodDaemonArg(methodName, arg, groupName),
+          )
+        : parseDiagnosticsNodeObservatory(
+            invokeServiceMethodObservatoryWithGroupName1(methodName, arg),
+          );
   }
 
   Future<void> invokeVoidServiceMethodInspectorRef(
@@ -1049,19 +1041,17 @@ abstract class ObjectGroupBase implements Disposable {
     bool isProperty,
   ) async {
     if (disposed) return const [];
-    if (useDaemonApi) {
-      return parseDiagnosticsNodesDaemon(
-        invokeServiceMethodDaemonInspectorRef(methodName, instanceRef),
-        parent,
-        isProperty,
-      );
-    } else {
-      return parseDiagnosticsNodesObservatory(
-        invokeServiceMethodObservatoryInspectorRef(methodName, instanceRef),
-        parent,
-        isProperty,
-      );
-    }
+    return useDaemonApi
+        ? parseDiagnosticsNodesDaemon(
+            invokeServiceMethodDaemonInspectorRef(methodName, instanceRef),
+            parent,
+            isProperty,
+          )
+        : parseDiagnosticsNodesObservatory(
+            invokeServiceMethodObservatoryInspectorRef(methodName, instanceRef),
+            parent,
+            isProperty,
+          );
   }
 
   /// Evaluate an expression where `object` references the [inspectorRef] or
@@ -1246,12 +1236,10 @@ class ObjectGroup extends ObjectGroupBase {
     }
     if (disposed) return null;
 
-    if (newSelection != null &&
-        newSelection.dartDiagnosticRef == previousSelectionRef) {
-      return previousSelection;
-    } else {
-      return newSelection;
-    }
+    return newSelection != null &&
+            newSelection.dartDiagnosticRef == previousSelectionRef
+        ? previousSelection
+        : newSelection;
   }
 
   Future<bool> setSelectionInspector(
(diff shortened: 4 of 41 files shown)


packages/devtools_app/lib/src/http/_http_date.dart
```diff
@@ -221,11 +221,7 @@ class HttpDate {
 
     int expectNum(String separator) {
       int pos;
-      if (separator.length > 0) {
-        pos = date.indexOf(separator, index);
-      } else {
-        pos = date.length;
-      }
+      pos = separator.length > 0 ? date.indexOf(separator, index) : date.length;
```
jacob314, line 224: changes like this seem like an improvement. Fyi @pq have we thought about adding a lint for preferring conditionals?
The one in Dart Code Metrics goes a bit overboard. I'd imagine if we added it to the core linter it would only fire for short expressions like this and not the large ones that code metrics suggests.

packages/devtools_app/lib/src/screens/debugger/call_stack.dart
```diff
@@ -63,40 +63,42 @@ class _CallStackState extends State<CallStack>
...
-    if (asyncMarker) {
-      child = Row(
-        children: [
-          const SizedBox(width: defaultSpacing, child: Divider()),
-          Padding(
-            padding: const EdgeInsets.symmetric(horizontal: densePadding),
-            child: Text(
-              frameDescription,
-              style: selected ? theme.selectedTextStyle : theme.subtleTextStyle,
+    child = asyncMarker
```
jacob314, line 66: @kenzieschmoll what do you think about cases like this? Is this cleaner or more obfuscated?

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
```diff
@@ -807,7 +805,7 @@ class InspectorController extends DisposableController
   }
 
   void selectionChanged() {
-    if (visibleToUser == false) {
+    if (!visibleToUser) {
```
jacob314, line 808: Fyi @pq. This is a lint fix from Dart code metrics that I'd be comfortable having in the recommended set if we were able to run Dart code metrics without significant performance overhead.

packages/devtools_app/lib/src/screens/inspector/inspector_service.dart
```diff
@@ -638,45 +636,39 @@ abstract class ObjectGroupBase implements Disposable {
...
-    if (useDaemonApi) {
-      return parseDiagnosticsNodeDaemon(
-        invokeServiceMethodDaemonInspectorRef(methodName, ref),
-      );
-    } else {
-      return parseDiagnosticsNodeObservatory(
-        invokeServiceMethodObservatoryInspectorRef(methodName, ref),
-      );
-    }
+    return useDaemonApi
```
jacob314, line 651: this was a larger block but I still think this was a little bit of an improvement over the original code.

packages/devtools_app/lib/src/shared/theme.dart
```diff
@@ -297,9 +297,7 @@ extension DevToolsColorScheme on ColorScheme {
 
   Color get toggleButtonsFillSelected => devtoolsBlue[400]!;
 
-  Color get grey => isLight
-      ? const Color.fromARGB(255, 128, 128, 128)
-      : const Color.fromARGB(255, 128, 128, 128);
+  Color get grey => const Color.fromARGB(255, 128, 128, 128);
```
jacob314, line 300: Case the linter caught we had a useless check for `isLight`

packages/devtools_app/lib/src/ui/utils.dart
```diff
@@ -288,9 +286,7 @@ class ScreenSize {
     if (height < 450) return MediaSize.xs;
     if (height < 600) return MediaSize.s;
     if (height < 750) return MediaSize.m;
-    if (height < 900)
-      return MediaSize.l;
-    else
-      return MediaSize.xl;
+    if (height < 900) return MediaSize.l;
```
jacob314, line 289: Fyi @pq this sort of case is probably one we would want to start suggesting and quick fixing to Patterns.

packages/devtools_app/lib/src/charts/flame_chart.dart
```diff
@@ -379,11 +379,7 @@ abstract class FlameChartState<T extends FlameChart,
   }
 
   void _handleKeyEvent(RawKeyEvent event) {
-    if (event.isAltPressed) {
-      _altKeyPressed = true;
-    } else {
-      _altKeyPressed = false;
-    }
+    _altKeyPressed = event.isAltPressed ? true : false;
```
bwilkerson, line 382: This should be further simplified to `_altKeyPressed = event.isAltPressed;` (as the branches just return the value of the condition).

packages/devtools_app/lib/src/screens/debugger/call_stack.dart
bwilkerson: I find it harder to read when one or more of the three operands need to be split across multiple lines.

packages/devtools_app/lib/src/framework/status_line.dart
```diff
@@ -98,29 +98,27 @@ class StatusLine extends StatelessWidget {
...
-                : 'flutter.dev/devtools/$docPageId',
-            url: 'https://flutter.dev/devtools/$docPageId',
-            gaScreenName: currentScreen.screenId,
-            gaSelectedItemDescription: analytics_constants.documentationLink,
-          ),
-          context: context,
-        ),
-      );
-    } else {
-      // Use a placeholder for pages with no explicit documentation.
```
kenzieschmoll, line 116: was the comment automatically removed? we should probably leave this even with the ternary operator

packages/devtools_app/lib/src/screens/debugger/call_stack.dart
kenzieschmoll: I do think the if / else block is a little bit easier to read here given how many lines each operand is.

packages/devtools_app/lib/src/screens/debugger/debugger_model.dart
```diff
@@ -391,11 +391,13 @@ class FileNode extends TreeNode<FileNode> {
...
-    } else {
-      return node.scriptRef == null ? false : scriptRef == node.scriptRef;
-    }
+    return scriptRef == null
+        ? node.scriptRef != null
+            ? false
+            : name == node.name
+        : node.scriptRef == null
+            ? false
+            : scriptRef == node.scriptRef;
```
kenzieschmoll, line 400: the nested ternary operators makes this a bit difficult to read

packages/devtools_app/lib/src/screens/memory/panes/chart/chart_pane.dart
```diff
@@ -457,25 +455,15 @@ class _MemoryChartPaneState extends State<MemoryChartPane>
...
-      }
+      traceColor = colorPatch != null
+          ? dashed
+              ? createDashWidget(colorPatch)
+              : createSolidLine(colorPatch)
+          : image == null
+              ? const SizedBox()
+              : scaleImage
+                  ? Image(image: AssetImage(image), width: 20, height: 10)
+                  : Image(image: AssetImage(image));
```
kenzieschmoll, line 466: triple nest 😱

packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
```diff
@@ -320,11 +318,9 @@ extension FieldPrivateViewExtension on Field {
...
     }
 
-    if (guardClassType == '@Class' || guardClassType == 'Class') {
-      return true;
-    } else {
-      return false;
-    }
+    return guardClassType == '@Class' || guardClassType == 'Class'
+        ? true
+        : false;
```
kenzieschmoll, line 323: could be further simplified to `return guardClassType == '@Class' || guardClassType == 'Class';`

packages/devtools_app/lib/src/screens/debugger/call_stack.dart
jacob314: Agreed. Reverted this one

packages/devtools_app/lib/src/charts/flame_chart.dart
jacob314: done. filed https://github.com/dart-lang/linter/issues/3888 to track this as this really obvious unnecessary code has been committed over 200X in g3.

packages/devtools_app/lib/src/screens/debugger/call_stack.dart
```diff
@@ -112,17 +114,15 @@ class _CallStackState extends State<CallStack>
...
-      return DevToolsTooltip(
-        message: locationDescription == null
-            ? frameDescription
-            : '$frameDescription $locationDescription',
-        waitDuration: tooltipWaitLong,
-        child: result,
-      );
-    }
+    return isAsyncBreak
+        ? result
```
jacob314, line 118: @kenzieschmoll what do you think of this case is. Brian thought it is a bit harder to read but I could go either way on this one.

packages/devtools_app/lib/src/screens/debugger/debugger_model.dart
jacob314: agreed this is overly clever. reverted.

packages/devtools_app/lib/src/screens/memory/panes/chart/chart_pane.dart
jacob314: yeah. if we were to add a lint about ternary operators, it would never suggest more than one nesting level and never suggest large blocks.

packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
jacob314: Fixed and filed https://github.com/dart-lang/linter/issues/3888

jacob314 (commented):
PTAL

packages/devtools_app/lib/src/http/_http_date.dart
kevmoo: what about `isNotEmpty` for `seperator`?

incendial:
@jacob314 hi, if you have any feedback about the package (aside the comments above which I agree with, especially `? true : false` cases), please share! Also interested in the rules configuration you used, not sure what you meant by `default dart_code_metrics configuration.`.

> This is a lint fix from Dart code metrics that I'd be comfortable having in the recommended set if we were able to run Dart code metrics without significant performance overhead.

This is interesting 🙂 . I'm working on a DCM version without the plugins API (and thus not listed in pubspec at all) and the overall performance is very good (VS Code integration is ready, now finishing the IntelliJ part). I'd be happy to discuss this idea in more details to understand what can be done from our side to enable DCM for the Flutter teams.

jacob314:
> @jacob314 hi, if you have any feedback about the package (aside the comments above which I agree with, especially `? true : false` cases), please share! Also interested in the rules configuration you used, not sure what you meant by `default dart_code_metrics configuration.`.
> 
> > This is a lint fix from Dart code metrics that I'd be comfortable having in the recommended set if we were able to run Dart code metrics without significant performance overhead.
> 
> This is interesting 🙂 . I'm working on a DCM version without the plugins API (and thus not listed in pubspec at all) and the overall performance is very good (VS Code integration is ready, now finishing the IntelliJ part). I'd be happy to discuss this idea in more details to understand what can be done from our side to enable DCM for the Flutter teams.

By default configuration I mean the basic configuration suggested by https://dartcodemetrics.dev/docs/getting-started/configuration
I'm very interested in hearing more about the DCM version without the plugins API as well. In particular I'm interested in what memory usage and performance you are seeing when you avoid the plugins API.

General feedback: all of the lints I tried were ones I would have liked to have  turned on by default except for `prefer-conditional-expressions` which would have too many cases I'd need to ignore to enable.
I'd love a more restricted version of `prefer-conditional-expressions` that doesn't ever prefer double nested conditional expressions and doesn't prefer conditional expressions if either clause of the conditional is more than a couple lines long.  The `prefer-conditonal-expressions` lint also caught a bunch of silly redundant logic in our code so I'd be really excited to enable it if it could be made more restricted.
I think the restrictions are needed   due to it being harder to tell what the block boundaries are for a conditional expression than an `if`, `else` resulting in code that was harder to read as seen in this CL.
The other piece of feedback is the quick fix for `prefer-conditional-expression` corrupted by code a few times. I would suggest a bit more error checking to verify that the quick fix only does safe code transformations. 
It is preferable for a quick-fix to sometimes be unable to fix code than to introduce errors in the code. I can try to give a repro if it helps. In addition to sometimes corrupting logic, the fixes also removed comments which was a more minor but annoying case when trying to bulk cleanup issues.

packages/devtools_app/lib/src/http/_http_date.dart
jacob314: I'd also be happy with a lint + fix for isEmpty and isNotEmpty.  Seems like a nice simple lint to nudge people to use the Dart core libraries. Looks like our users use `.isNotEmpty` about 10X as often then `.length > 0` so the lint shouldn't be contreversial.

kevmoo: We HAVE that lint. Somewhere...

kevmoo: https://dart-lang.github.io/linter/lints/prefer_is_not_empty.html – HA! We should expand this to String!

kevmoo: https://github.com/dart-lang/linter/issues/3889

packages/devtools_app/lib/src/screens/debugger/call_stack.dart
kenzieschmoll: I also could go either way. I think I prefer the if / else here but the ternary isn't terrible since of of the options is a one-liner.

packages/devtools_app/lib/src/http/_http_date.dart
pq: Related: [`avoid_bool_literals_in_conditional_expressions`](https://dart-lang.github.io/linter/lints/avoid_bool_literals_in_conditional_expressions.html).

packages/devtools_app/lib/src/screens/inspector/inspector_controller.dart
pq: Right, that's:

https://dartcodemetrics.dev/docs/rules/common/no-boolean-literal-compare

I'm a little surprised that was never defined in the core linter...

packages/devtools_app/lib/src/ui/utils.dart
pq: See: https://github.com/dart-lang/linter/issues/3673

packages/devtools_app/lib/src/screens/memory/panes/chart/legend.dart
```diff
@@ -167,19 +167,13 @@ class LegendRow extends StatelessWidget {
...
-      }
-
-      return traceColor;
+      return color != null
+          ? dashedLine as bool
+              ? createDashWidget(color)
+              : createSolidLine(color)
+          : image == null
+              ? const SizedBox()
+              : Image(image: AssetImage(image));
```
kenzieschmoll, line 176: can we revert the double ternary?

kenzieschmoll (approved):
one comment then lgtm

packages/devtools_app/lib/src/screens/memory/panes/chart/legend.dart
jacob314: Done. Reverted this file.

incendial:
> By default configuration I mean the basic configuration suggested by https://dartcodemetrics.dev/docs/getting-started/configuration

Oh, I see. I guess we need to update it coz the list of rules is far more bigger right now 🙂 (you can find the whole list in presets https://github.com/dart-code-checker/dart-code-metrics/tree/master/lib/presets)

> I'm very interested in hearing more about the DCM version without the plugins API as well. In particular I'm interested in what memory usage and performance you are seeing when you avoid the plugins API.

**VS Code:** the overall memory consumption for so called DCM analysis-server is about one and a half times less than for the Dart analysis-server

In numbers (for monorepo of like 6 packages):
 - DCM analysis server - ~260 MB
 - Dart - ~370 MB
Screenshot 2022-12-09 at 12 59 49

Just for the DCM repo https://github.com/dart-code-checker/dart-code-metrics:
 - DCM analysis server - ~ 200 MB
 - Dart - ~370 MB
 - Dart + plugin - ~1,1 GB
Screenshot 2022-12-09 at 13 08 55

Meaning that the DCM + Dart consume about twice less as Dart + plugin 😁.

**Please note:**  the measurements above taken after both processes were running for some time. The measurement on a start gives an even bigger difference (like 3 times).

Additionally, there is no "new plugin instance for each package" problem, meaning that it scales pretty good. And the overall responsiveness is better (if both the extension and the plugin are enabled, I see the highlight from the extension and then from the plugin in a noticeable amount of time).

**IntelliJ:** I haven't yet measured the IntelliJ difference, since the implementation is still in progress, but hopefully soon I'll be able to measure. Overall performance for the issues highlight is good, IMO.

> I'd love a more restricted version of prefer-conditional-expressions that doesn't ever prefer double nested conditional expressions and doesn't prefer conditional expressions if either clause of the conditional is more than a couple lines long. The prefer-conditonal-expressions lint also caught a bunch of silly redundant logic in our code so I'd be really excited to enable it if it could be made more restricted.

Thanks, noted! We have a separate rule for limiting the nesting level https://dartcodemetrics.dev/docs/rules/common/avoid-nested-conditional-expressions, but idea to make `prefer-conditonal-expressions` aware of nesting sound solid!

> The other piece of feedback is the quick fix for prefer-conditional-expression corrupted by code a few times. I would suggest a bit more error checking to verify that the quick fix only does safe code transformations.

An example would be great, tbh (so I'd be able to fix it). But I can agree that the quick fixes we have don't check for the output validity and more about "this should work" type of a result.

>  I can try to give a repro if it helps. In addition to sometimes corrupting logic,

Would be amazing!

> the fixes also removed comments which was a more minor but annoying case when trying to bulk cleanup issues.

Definitely a bug, thank you!

PR #4892, opened 2022-12-05
Fix legend to show dash line.

polina-c:
Before, dashed line was appearing on hot reload and then disappearing:

https://user-images.githubusercontent.com/12115586/205525186-e83254d2-f07d-4ab8-a7ea-67d063cd1a20.mov

After: 

Screenshot 2022-12-04 at 4 18 22 PM

Filed https://github.com/flutter/flutter/issues/116494


packages/devtools_app/lib/src/screens/memory/primitives/painting.dart
```diff
@@ -4,30 +4,23 @@
...
 
-const _totalDashWidth = 15.0;
+// TODO(polina-c): use the same constants as for dash in the chart.
 const _dashHeight = 2.0;
-const _dashWidth = 4.0;
+const _dashWidth = 5.0;
 const _spaceBetweenDash = 3.0;
 
 Widget createDashWidget(Color color) {
-  return Container(
```
bkonyi, line 13: Do you know why we had a `painter` and `foregroundPainter`? Can you try removing one of these painters and see if you have the same issue?

polina-c: I tried this and many other things. It is bug in Flutter, they could repro it (see https://github.com/flutter/flutter/issues/116494).

bkonyi: Okay, just wanted to make sure!

bkonyi (approved):

Issue #1987, opened 2020-05-19
Re-write integration tests for Flutter app


kenzieschmoll:
We previously had integration tests for the html app. We disabled these tests but kept the files for reference during test re-write. We should write versions of these tests for the Flutter app.

kenzieschmoll:
Required pieces before integration test framework is complete

- [ ] Run a real flutter app to connect DevTools to in the integration tests
- [ ] Run integration tests on bots

PR #4901, opened 2022-12-07
Add skeleton for integration test setup

kenzieschmoll:
Work towards https://github.com/flutter/devtools/issues/1987.


kenzieschmoll:
There is some code on this PR that we may want to reference later, but we can close it for now.

PR #4904, opened 2022-12-08
IDG: Add first elements of discoverability & extensibility API

lukechurch:
For discussion @kenzieschmoll  @InMatrix 

This PR is contains the first part of the infrastructure for supporting Interactive Debugging Guides, per the high level design documentation we talked about earlier.

This follows an approach of constructing a facade over the existing pages with elements that are safe for an extensibility system to use. It's more work on the page authors to maintain than a dynamic search scheme would have been, but allows for static analysis to determine breaking changes between pages and guides, without creating excessive coupling between the pages and all potential guides. It might also serve as the beginning of a wider extensibility story. 

If you're comfortable with this facade approach this would then be followed by PRs containing:

- The IDG engine
- Sample guides
- The IDG UI components

Thanks in advance for taking a look.

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/memory_controller_discoverable.dart
+++ b/packages/devtools_app/lib/src/screens/memory/memory_controller_discoverable.dart
@@ -1,0 +1,41 @@
+import 'package:flutter/foundation.dart';
+
+import '../../../devtools_app.dart';
+import '../../analytics/constants.dart' as analytics_constants;
+import '../../extensibility/discoverable.dart';
+import 'memory_tabs.dart';
+
+class DiscoverableMemoryPage extends DiscoverablePage {
+  DiscoverableMemoryPage(this.controller) : super() {
+    discoverableApp.memoryPage = this;
+  }
+
+  final MemoryController controller;
+
+  static String get id => MemoryScreen.id;
+
+  // Events
+  static const memorySnapshotTaken = 'mem-snapshot-done';
+
+  // Actions
+  void takeSnapshotAction() {
+    final takeSnapshot = controller.diffPaneController.takeSnapshotHandler(
+      analytics_constants.MemoryEvent.diffTakeSnapshotControlPane,
+    );
+    if (takeSnapshot == null)
+      print("takeSnapshotHandler returned null, can't take snapshot");
+    else
+      takeSnapshot();
+  }
+
+  void changeTabAction(Key tab) {
+    controller.currentTab.value = tab;
+  }
+
+  static const dartHeapTableProfileTab =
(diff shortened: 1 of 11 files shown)


packages/devtools_app/lib/src/ui/tab.dart
```diff
@@ -174,6 +201,7 @@ class _AnalyticsTabbedViewState extends State<AnalyticsTabbedView>
 
   @override
   void dispose() {
+    widget.selectedTabNotifier?.removeListener(_onChangeTab);
```
jacob314, line 204: use auto dispose mixin here and for the other listeners in this class.

packages/devtools_app/lib/src/screens/performance/performance_controller_discoverable.dart
```diff
@@ -0,0 +1,21 @@
+import '../../../devtools_app.dart';
+import '../../extensibility/discoverable.dart';
+
+class DiscoverablePerformancePage extends DiscoverablePage {
```
jacob314, line 4: is there a reason this can't be part of the PerformanceScreen class itself?

packages/devtools_app/lib/src/screens/memory/memory_controller_discoverable.dart
```diff
@@ -0,0 +1,41 @@
...
+  DiscoverableMemoryPage(this.controller) : super() {
+    discoverableApp.memoryPage = this;
+  }
+
+  final MemoryController controller;
+
+  static String get id => MemoryScreen.id;
+
+  // Events
+  static const memorySnapshotTaken = 'mem-snapshot-done';
```
jacob314, line 18: could be nice to have a single enum of all the events so it is easier to track them across all screens.
The enum could have comments about what page they are for.

InMatrix:
Thanks for taking a look @jacob314. I'd like to add some context about this PR after talking with @lukechurch and @kenzieschmoll separately. This PR is intended to demonstrate how we might add the capability of highlighting specific UI elements of DevTools from a program decoupled from DevTools. The first use case is likely to provide some sort of onboarding tour of features for perf debugging tabs. 

Luke plans on sending a few follow-up PRs to demonstrate how this capability can be used in interactive tutorials that can drive DevTools. These parts are more experimental and will require updated docs and case studies for perf debugging tools to have value. For now, the PRs will serve the purpose of a technical feasibility evaluation, so we can have a better idea about whether this is something we want to productionize when the tutorial content is ready.

InMatrix:
I believe this PR has been replaced by https://github.com/flutter/devtools/pull/5469. I'll close it for now.

Issue #4902, opened 2022-12-07
Add integration tests for router state based logic


bkonyi:
We need to be able to run browser based integration tests to fully test navigation state logic (e.g., jump to source history).

PR #4906, opened 2022-12-08
Add support for Debugger screen navigation history

bkonyi:
https://user-images.githubusercontent.com/24210656/206500549-450b876d-0033-4302-bad3-8f556d8d90f1.mov

Note: this change will eventually need to be tested by browser integration tests (see https://github.com/flutter/devtools/issues/4902).


packages/devtools_app/lib/src/shared/routing.dart
```diff
@@ -297,6 +323,20 @@ class DevToolsNavigationState {
     );
   }
 
+  /// Creates a new [DevToolsNavigationState] by merging this instance with
+  /// [other].
+  ///
+  /// State contained in [other] will take presidence over state contanied in
```
kenzieschmoll, line 329: typo: contained

packages/devtools_app/lib/src/screens/debugger/debugger_screen.dart
```diff
@@ -199,9 +200,13 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
   void _onNodeSelected(VMServiceObjectNode? node) {
     final location = node?.location;
     if (location != null) {
-      controller.codeViewController.showScriptLocation(
-        location,
-        focusLine: true,
+      final routerDelegate = DevToolsRouterDelegate.of(context);
+      routerDelegate.updateStateIfNotCurrent(
```
kenzieschmoll, line 204: just piecing this new system together: does the debugger have to listen for router updates somewhere?

kenzieschmoll (commented):
I know we can't add a full integration test yet, but can you verify that the router state is updated when new nodes are clicked?

also, are routes popped off of the stack when you click back in the browser?

so for node selections 1 --> 2 --> 3, then back, then to 4, would the stack look like 1 --> 2 --> 4?

packages/devtools_app/lib/src/screens/debugger/debugger_screen.dart
bkonyi: It can, but in this situation, it doesn't need to since the `CodeViewController` is handling the state update. For example, if we wanted to support history for debugger frame selection (not that I think we should do this), the `DebuggerController` would need to be able to listen to router updates to determine if it should update the currently selected frame.

bkonyi:
> I know we can't add a full integration test yet, but can you verify that the router state is updated when new nodes are clicked?

Yes, we should be able to verify that the currently selected node matches the current state in the router.

> also, are routes popped off of the stack when you click back in the browser?
> 
> so for node selections 1 --> 2 --> 3, then back, then to 4, would the stack look like 1 --> 2 --> 4?

That's correct. We're using the standard browser history behavior here.

packages/devtools_app/lib/src/screens/debugger/codeview_controller.dart
```diff
@@ -50,8 +51,22 @@ class CodeViewController extends DisposableController
         {
           final processedState =
               CodeViewSourceLocationNavigationState._fromState(state);
-          // TODO(bkonyi): investigate delay in scrolling to source location.
+          final object = processedState.object;
```
elliette, line 54: Consider pulling body of the `case` clause into a separate method for clarity, eg. `_updateCodeViewSource`

packages/devtools_app/lib/src/screens/debugger/codeview_controller.dart
```diff
@@ -456,8 +484,21 @@ class CodeViewSourceLocationNavigationState extends DevToolsNavigationState {
 
   int get line => int.parse(state[_kLine]!);
 
+  ObjRef? get object {
+    final obj = state[_kObject];
+    if (obj == null) {
+      return null;
+    }
+    return createServiceObject(json.decode(obj), const []) as ObjRef?;
```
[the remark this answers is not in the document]
bkonyi: There's probably more efficient ways to do this, but we'd need to create a generalized cache for VM service objects which would probably open a whole other can of worms given that object IDs expire. Given this operation requires user interaction, I wouldn't be too concerned about some extra time serializing / deserializing VM service objects.

packages/devtools_app/NEXT_RELEASE_NOTES.md
```diff
@@ -1,38 +1,27 @@
 This is draft for future release notes, that are going to land on
 [the Flutter website](https://docs.flutter.dev/development/tools/devtools/release-notes).
 
-# DevTools 2.21.0 release notes
+# DevTools 2.22.0 release notes
```
kenzieschmoll, line 4: this shouldn't be changed to 2.22.0 - all these release notes are for 2.21.0 which has not been released yet, so just add your entry to the debugger section and leave all the rest

packages/devtools_app/lib/src/screens/debugger/debugger_screen.dart
```diff
@@ -203,11 +215,17 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
 
   void _onNodeSelected(VMServiceObjectNode? node) {
     final location = node?.location;
-    if (location != null) {
-      controller.codeViewController.showScriptLocation(
-        location,
-        focusLine: true,
-      );
+    if (node != null && location != null) {
```
kenzieschmoll, line 218: the check to node != null is redundant.

packages/devtools_app/lib/src/screens/debugger/debugger_screen.dart
```diff
@@ -203,11 +215,17 @@ class DebuggerScreenBodyState extends State<DebuggerScreenBody>
...
+      final routerDelegate = DevToolsRouterDelegate.of(context);
+      Router.navigate(context, () {
+        routerDelegate.updateStateIfChanged(
+          CodeViewSourceLocationNavigationState(
+            script: location.scriptRef,
+            line: location.location?.line ?? -1,
```
kenzieschmoll, line 224: should the default be -1 or 0?

packages/devtools_app/lib/src/shared/routing.dart
```diff
@@ -245,6 +247,54 @@ class DevToolsRouterDelegate extends RouterDelegate<DevToolsRouteConfiguration>
...
+      DevToolsRouteConfiguration(
+        currentConfig.page,
+        currentConfig.args,
+        state,
+      ),
+    );
+
+    final path = '/${currentConfig.page}';
+    // Create a new map in case the one we were given was unmodifiable.
+    final params = {...currentConfig.args};
```
kenzieschmoll, line 262: nit: Map.of would be a bit cleaner

packages/devtools_app/lib/src/shared/routing.dart
```diff
@@ -245,6 +247,54 @@ class DevToolsRouterDelegate extends RouterDelegate<DevToolsRouteConfiguration>
...
+      state: state,
+      replace: true,
+    );
+  }
+
+  /// Updates state for the current page.
+  ///
+  /// Existing state will be preserved unless overwritten by [stateUpdate].
+  void updateStateIfChanged(
+    DevToolsNavigationState stateUpdate,
```
kenzieschmoll, line 279: super nit: could this fit on one line?

kenzieschmoll (commented):
can we test the routing.dart changes in routing_test.dart?

packages/devtools_app/NEXT_RELEASE_NOTES.md
CoderDake: Our update helper will reset this file anytime we do a major or minor version bump. We will probably need to undo the changes to this file so we don't remove the release notes

packages/devtools_app/lib/src/screens/debugger/debugger_screen.dart
bkonyi: 0 is probably more correct.

bkonyi:
> can we test the routing.dart changes in routing_test.dart?

Done.

packages/devtools_app/lib/src/shared/routing.dart
```diff
@@ -259,9 +259,8 @@ class DevToolsRouterDelegate extends RouterDelegate<DevToolsRouteConfiguration>
 
     final path = '/${currentConfig.page}';
     // Create a new map in case the one we were given was unmodifiable.
-    final params = {...currentConfig.args};
+    final params = Map.of(currentConfig.args);
     params.removeWhere((key, value) => value == null);
-    await SystemNavigator.selectMultiEntryHistory();
```
kenzieschmoll: was this remove intentional?

kenzieschmoll (approved):

packages/devtools_app/lib/src/shared/routing.dart
bkonyi: Yes, this isn't necessary.

PR #4907, opened 2022-12-08
Reorganize code.

polina-c:
1. Move folders with code that is used by screens and does not depend on screens, to `shared`.
2. Rename alias `analytics_constants` to `gac`, to make code conciser.

Will update README.md for `shared` and `primitives` in next PR.

Screenshot 2022-12-08 at 9 23 22 AM


packages/devtools_app/lib/src/screens/memory/shared/shared_memory_widgets.dart
```diff
@@ -5,7 +5,7 @@
 import 'package:flutter/widgets.dart';
 
 import '../../../../devtools_app.dart';
```
kenzieschmoll, line 7: remove the devtools_app import

kenzieschmoll (approved):

packages/devtools_app/lib/src/screens/memory/shared/shared_memory_widgets.dart
polina-c: will do in next PR. I really want to avoid conflicts :)

PR #4908, opened 2022-12-08
Daily Dev Bumps

CoderDake:
# Description
This workflow will run a dev bump every day  at midnight.

For now, it uses my email as the author. In the future we will want to create a github devtools account that can be the author instead. This issue is tracked here: https://github.com/flutter/devtools/issues/4949

# Example Auto Commit

The following change is what an auto dev bump will look like
https://github.com/flutter/devtools/commit/76aff6b892ca980f5808f2ccadc77425d40ab0d5

**QUESTION:** Should we remove the changelog entry when doing the auto commits?

# Example Failure
Doing a dev bump on a release version will fail. This is to prevent the workflow from bumping the version while we are doing a release.
Screen Shot 2022-12-14 at 9 45 23 AM


kenzieschmoll:
I don't think we want to add a new CHANGELOG entry for each bump. We should either modify the existing dev entry, or ignore the CHANGELOG update all together.

tool/update_version.dart
```diff
@@ -33,7 +34,9 @@ void main(List<String> args) async {
 }
 
 Future<void> performTheVersionUpdate(
-    {required String currentVersion, required String newVersion}) async {
+    {required String currentVersion,
+    required String newVersion,
+    bool modifyChangeLog = true}) async {
```
kenzieschmoll, line 39: nit: tc

tool/update_version.dart
```diff
@@ -150,6 +155,7 @@ void writeVersionToVersionFile(File versionFile, String version) {
 }
 
 void writeVersionToChangelog(File changelog, String version) {
+  //TODO Do not do this for devv file
```
kenzieschmoll, line 158: is this todo necessary?

CoderDake: oops this is extraneous

kenzieschmoll (approved):

PR #4910, opened 2022-12-08
Update-version script changes

CoderDake:
To simplify the update-version script, each type only modifies  bumps itself.

a `--dry-run` flag has been added to print information about what kind of update will happen
a `--type release` has been added which strips the any pre release versions
  
Screen Shot 2022-12-09 at 4 03 15 PM


kenzieschmoll:
How will a version be bumped from release to dev? e.g. 2.20.0 --> 2.21.0-dev.0

tool/update_version.dart
```diff
@@ -284,43 +293,43 @@ class AutoUpdateCommand extends Command {
...
-          'patch': 'bumps the version to the next patch value',
-          'minor': 'bumps the version to the next minor value',
-          'major': 'bumps the version to the next major value',
+              'bumps the version to the next dev pre-release value (minor by default).',
+          'patch':
+              'bumps the version to the next patch value, and sets the dev version to 0.',
+          'minor':
+              'bumps the version to the next minor value, and sets the dev version to 0.',
+          'major':
+              'bumps the version to the next major value, and sets the dev version to 0.',
```
kenzieschmoll, line 306: it would be helpful if each of these help texts had an example like 2.20.0 --> 2.21.0

kenzieschmoll, line 306: the fact that these set the dev version also is a bit confusing.

Issue #4923, opened 2022-12-09
Show text in "Show chart" button on memory page


kenzieschmoll:
We could also consider changing this to "Stats chart" or "Live memory stats" and then just have the arrow up or down indicate expanded collapsed state (like how we have "Flutter frames" with caret up or caret down on the performance screen).

![Screenshot 2022-12-09 at 10 57 03 AM](https://user-images.githubusercontent.com/43759233/206775591-abd9650e-4d84-4199-a33d-9b1f1c71fea1.png)

There is plenty of horizontal space to show the text on this button and that would make it more clear what the button is for

PR #4928, opened 2022-12-09
Stop changing tooltip on state change.

polina-c:
fixes https://github.com/flutter/devtools/issues/4923


packages/devtools_app/lib/src/screens/memory/panes/control/primary_controls.dart
```diff
@@ -19,6 +19,7 @@ class PrimaryControls extends StatelessWidget {
       showChart: preferences.memory.showChart,
       onPressed: (show) => preferences.memory.showChart.value = show,
       minScreenWidthForTextBeforeScaling: memoryControlsMinVerboseWidth,
```
kenzieschmoll, line 21: can we also add "Memory usage" as a label here? We can make the tooltip more verbose: "Toggle the visibility of the Memory usage chart". Same comment below for the tooltip on the Performance screen "Toggle the visibility of the Flutter frames chart"

kenzieschmoll (approved):
one fix required, then lgtm

packages/devtools_app/lib/src/screens/performance/performance_screen.dart
```diff
@@ -199,12 +199,12 @@ class _PrimaryControls extends StatelessWidget {
...
         if (serviceManager.connectedApp!.isFlutterAppNow!) ...[
-          ChartVisibilityButton(
-            showChart:
-                controller.flutterFramesController.showFlutterFramesChart,
+          VisibilityButton(
+            show: controller.flutterFramesController.showFlutterFramesChart,
             onPressed:
                 controller.flutterFramesController.toggleShowFlutterFrames,
             label: 'Flutter frames',
+            tooltip: 'Toggle visibility of the Flutter frames',
```
kenzieschmoll, line 207: r/Flutter frames/Flutter frames chart

PR #4931, opened 2022-12-10
More fixes from enabling all of code metrics

jacob314:
I'm still not done but this is a good sampling to get feedback.

Fyi @pq , @incendial.

Some of the  dead code caught was pretty amazing. Also helped clean up a bunch of really dumb async code.

```
#    - avoid-global-state   <- probably could be enabled as a follow up although most of our use cases are reasonable for just setting bool flags.
#    - avoid-ignoring-return-values. <-- too many false positives. wish this one could be enabled.
#    - avoid-late-keyword <-- generally we think late is ok or at least unavoidable.
#    - avoid-non-null-assertion <-- unrealistically strict for devtools.
#    - avoid-passing-async-when-sync-expected TODO(jacobr): consider re-enabliing.
# This one seems interesting but has too many false positives. Gave it a try including kitchen sink adding parameters to ignore. Generally it didn't seem to catch that many real issues but had tons of false positives.
# - no-equal-arguments:
#    ignored-parameters:
#      - height
#      - width
#      - double-literal-format
#      - defaultSortColumn
#      - left
#      - right
#      - top
#      - bottom
#      - bottomLeft
#      - topLeft
#      - enabledBorder
#   - no-magic-number
# - prefer-match-file-name
# - prefer-static-class
# - avoid-returning-widgets  This one is nice but has a lot of false positives. Would be great to figure out how to make it more tolerant of reasonable patterns where a separate widget would just make things ugly.
# - prefer-single-widget-per-file. <-- just didn't match our style.
```

### DIFF
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -112,21 +112,21 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
     });
 
     _isDarkThemeEnabled = preferences.darkModeTheme.value;
-    preferences.darkModeTheme.addListener(() {
+    addAutoDisposeListener(preferences.darkModeTheme, () {
       setState(() {
         _isDarkThemeEnabled = preferences.darkModeTheme.value;
       });
     });
 
     _vmDeveloperModeEnabled = preferences.vmDeveloperModeEnabled.value;
-    preferences.vmDeveloperModeEnabled.addListener(() {
+    addAutoDisposeListener(preferences.vmDeveloperModeEnabled, () {
       setState(() {
         _vmDeveloperModeEnabled = preferences.vmDeveloperModeEnabled.value;
       });
     });
 
     _denseModeEnabled = preferences.denseModeEnabled.value;
-    preferences.denseModeEnabled.addListener(() {
+    addAutoDisposeListener(preferences.denseModeEnabled, () {
       setState(() {
         _denseModeEnabled = preferences.denseModeEnabled.value;
       });
@@ -191,10 +191,10 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
   }
 
   Widget _buildTabbedPage(
-    BuildContext context,
+    BuildContext _,
     String? page,
     Map<String, String?> params,
-    DevToolsNavigationState? state,
+    DevToolsNavigationState? __,
   ) {
     final vmServiceUri = params['uri'];
 
@@ -234,7 +234,8 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
                 .where((p) => !hide.contains(p.screenId))
                 .toList();
             if (screens.isEmpty) return child ?? const SizedBox.shrink();
-            return _providedControllers(
+            return MultiProvider(
+              providers: _providedControllers(),
               child: DevToolsScaffold(
                 embed: embed,
                 ideTheme: ideTheme,
@@ -277,8 +278,8 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
         return DevToolsScaffold.withChild(
           key: UniqueKey(),
           ideTheme: ideTheme,
-          child: _providedControllers(
-            offline: true,
+          child: MultiProvider(
+            providers: _providedControllers(offline: true),
             child: SnapshotScreenBody(snapshotArgs, _screens),
           ),
         );
@@ -292,7 +293,8 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
             ReportFeedbackButton(),
             OpenAboutAction(),
           ],
-          child: _providedControllers(
+          child: MultiProvider(
+            providers: _providedControllers(),
             child: const AppSizeBody(),
           ),
         );
@@ -308,18 +310,13 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
 
   List<Screen> _visibleScreens() => _screens.where(shouldShowScreen).toList();
 
-  Widget _providedControllers({required Widget child, bool offline = false}) {
-    final _providers = widget.screens
+  List<Provider> _providedControllers({bool offline = false}) {
+    return widget.screens
         .where(
           (s) => s.providesController && (offline ? s.supportsOffline : true),
         )
         .map((s) => s.controllerProvider(routerDelegate))
         .toList();
-
-    return MultiProvider(
-      providers: _providers,
-      child: child,
-    );
   }
 
   @override
@@ -456,7 +453,7 @@ class OpenSettingsAction extends StatelessWidget {
     return DevToolsTooltip(
       message: 'Settings',
       child: InkWell(
-        onTap: () async {
+        onTap: () {
           unawaited(
             showDialog(
               context: context,
@@ -505,7 +502,8 @@ class SettingsDialog extends StatelessWidget {
             CheckboxSetting(
               label: const Text('Enable analytics'),
               listenable: analyticsController.analyticsEnabled,
-              toggle: analyticsController.toggleAnalyticsEnabled,
+              toggle: (enable) =>
+                  unawaited(analyticsController.toggleAnalyticsEnabled(enable)),
               gaItem: gac.analytics,
             ),
           CheckboxSetting(
@@ -537,7 +535,7 @@ class CheckboxSetting extends StatelessWidget {
 
   final ValueListenable<bool> listenable;
 
-  final Function(bool) toggle;
+  final void Function(bool) toggle;
 
   final String gaItem;
 
--- a/packages/devtools_app/lib/src/framework/framework_core.dart
+++ b/packages/devtools_app/lib/src/framework/framework_core.dart
@@ -19,8 +19,9 @@ import '../shared/primitives/utils.dart';
 import '../shared/scripts/script_manager.dart';
 import '../shared/survey.dart';
 
-typedef ErrorReporter = void Function(String title, dynamic error);
+typedef ErrorReporter = void Function(String title, Object error);
 
+// TODO(jacobr): refactor this class to not use static members.
 // ignore: avoid_classes_with_only_static_members
 class FrameworkCore {
   static void initGlobals() {
--- a/packages/devtools_app/lib/src/framework/landing_screen.dart
+++ b/packages/devtools_app/lib/src/framework/landing_screen.dart
@@ -130,29 +130,9 @@ class _ConnectDialogState extends State<ConnectDialog>
 
   @override
   Widget build(BuildContext context) {
-    return LandingScreenSection(
-      title: 'Connect',
-      child: Column(
-        crossAxisAlignment: CrossAxisAlignment.start,
-        children: [
-          Text(
-            'Connect to a Running App',
-            style: Theme.of(context).textTheme.titleMedium,
-          ),
-          const SizedBox(height: denseRowSpacing),
-          Text(
-            'Enter a URL to a running Dart or Flutter application',
-            style: Theme.of(context).textTheme.bodySmall,
-          ),
-          const Padding(padding: EdgeInsets.only(top: 20.0)),
-          _buildConnectInput(),
-        ],
-      ),
-    );
-  }
+    final textTheme = Theme.of(context).textTheme;
 
-  Widget _buildConnectInput() {
-    return Column(
+    final connectorInput = Column(
       crossAxisAlignment: CrossAxisAlignment.stretch,
       children: [
         Row(
@@ -177,7 +157,7 @@ class _ConnectDialogState extends State<ConnectDialog>
             ),
             const SizedBox(width: defaultSpacing),
             ElevatedButton(
-              onPressed: actionInProgress ? null : _connect,
+              onPressed: actionInProgress ? null : () => unawaited(_connect()),
               child: const Text('Connect'),
             ),
           ],
@@ -191,6 +171,26 @@ class _ConnectDialogState extends State<ConnectDialog>
           ),
         ),
       ],
+    );
+
+    return LandingScreenSection(
+      title: 'Connect',
+      child: Column(
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          Text(
+            'Connect to a Running App',
+            style: textTheme.titleMedium,
+          ),
+          const SizedBox(height: denseRowSpacing),
+          Text(
+            'Enter a URL to a running Dart or Flutter application',
+            style: textTheme.bodySmall,
+          ),
+          const Padding(padding: EdgeInsets.only(top: 20.0)),
+          connectorInput,
+        ],
+      ),
     );
   }
 
@@ -251,6 +251,7 @@ class ImportFileInstructions extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final textTheme = Theme.of(context).textTheme;
     return LandingScreenSection(
       title: 'Load DevTools Data',
       child: Column(
@@ -258,13 +259,13 @@ class ImportFileInstructions extends StatelessWidget {
         children: [
           Text(
             'Import a data file to use DevTools without an app connection.',
-            style: Theme.of(context).textTheme.titleMedium,
+            style: textTheme.titleMedium,
           ),
           const SizedBox(height: denseRowSpacing),
           Text(
             'At this time, DevTools only supports importing files that were originally'
             ' exported from DevTools.',
-            style: Theme.of(context).textTheme.bodySmall,
+            style: textTheme.bodySmall,
           ),
           const SizedBox(height: defaultSpacing),
           ElevatedButton(
@@ -300,6 +301,7 @@ class AppSizeToolingInstructions extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final textTheme = Theme.of(context).textTheme;
     return LandingScreenSection(
       title: 'App Size Tooling',
       child: Column(
@@ -307,13 +309,13 @@ class AppSizeToolingInstructions extends StatelessWidget {
         children: [
           Text(
             'Analyze and view diffs for your app\'s size',
-            style: Theme.of(context).textTheme.titleMedium,
+            style: textTheme.titleMedium,
           ),
           const SizedBox(height: denseRowSpacing),
           Text(
             'Load Dart AOT snapshots or app size analysis files to '
             'track down size issues in your app.',
-            style: Theme.of(context).textTheme.bodySmall,
+            style: textTheme.bodySmall,
           ),
           const SizedBox(height: defaultSpacing),
           ElevatedButton(
--- a/packages/devtools_app/lib/src/framework/notifications_view.dart
+++ b/packages/devtools_app/lib/src/framework/notifications_view.dart
@@ -64,7 +64,7 @@ class _NotificationsState extends State<_Notifications> with AutoDisposeMixin {
     if (_overlayEntry == null) {
       _overlayEntry = OverlayEntry(
         maintainState: true,
-        builder: _buildOverlay,
+        builder: (_) => _NotificationOverlay(notifications: _notifications),
       );
       SchedulerBinding.instance.scheduleFrameCallback((_) {
         Overlay.of(context).insert(_overlayEntry!);
@@ -114,7 +114,7 @@ class _NotificationsState extends State<_Notifications> with AutoDisposeMixin {
     bool didDismiss = false;
     // Make a copy so we do not remove a notification from [_notifications]
     // while iterating over it.
-    final notifications = List<_Notification>.from(_notifications);
+    final notifications = List<_Notification>.of(_notifications);
     for (final notification in notifications) {
       if (notification.message.text == message) {
         _notifications.remove(notification);
@@ -139,7 +139,21 @@ class _NotificationsState extends State<_Notifications> with AutoDisposeMixin {
     });
   }
 
-  Widget _buildOverlay(BuildContext context) {
+  @override
+  Widget build(BuildContext context) {
+    return widget.child;
+  }
+}
+
+class _NotificationOverlay extends StatelessWidget {
+  const _NotificationOverlay({
+    required List<_Notification> notifications,
+  }) : _notifications = notifications;
+
+  final List<_Notification> _notifications;
+
+  @override
+  Widget build(BuildContext context) {
     return Align(
       alignment: Alignment.bottomRight,
       child: Padding(
@@ -161,11 +175,6 @@ class _NotificationsState extends State<_Notifications> with AutoDisposeMixin {
         ),
       ),
     );
-  }
-
-  @override
-  Widget build(BuildContext context) {
-    return widget.child;
   }
 }
 
@@ -245,9 +254,9 @@ class _NotificationState extends State<_Notification>
               child: Column(
                 crossAxisAlignment: CrossAxisAlignment.start,
                 children: [
-                  _buildMessage(),
+                  _NotificationMessage(widget: widget, context: context),
                   const SizedBox(height: defaultSpacing),
-                  _buildActions(),
+                  _NotificationActions(widget: widget),
                 ],
               ),
             ),
@@ -256,8 +265,19 @@ class _NotificationState extends State<_Notification>
       ),
     );
   }
+}
 
-  Widget _buildMessage() {
+class _NotificationMessage extends StatelessWidget {
+  const _NotificationMessage({
+    required this.widget,
+    required this.context,
+  });
+
+  final _Notification widget;
+  final BuildContext context;
+
+  @override
+  Widget build(BuildContext context) {
     return Text(
       widget.message.text,
       style: Theme.of(context).textTheme.bodyLarge,
@@ -265,13 +285,22 @@ class _NotificationState extends State<_Notification>
       maxLines: 6,
     );
   }
+}
 
-  Widget _buildActions() {
-    if (widget.message.actions.isEmpty) return const SizedBox();
+class _NotificationActions extends StatelessWidget {
+  const _NotificationActions({
+    required this.widget,
+  });
+
+  final _Notification widget;
+
+  @override
+  Widget build(BuildContext context) {
+    final actions = widget.message.actions;
+    if (actions.isEmpty) return const SizedBox();
     return Row(
       mainAxisAlignment: MainAxisAlignment.end,
-      children:
-          widget.message.actions.joinWith(const SizedBox(width: denseSpacing)),
+      children: actions.joinWith(const SizedBox(width: denseSpacing)),
     );
   }
 }
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
@@ -19,7 +19,7 @@ class AppSizeAnalysisTable extends StatelessWidget {
   }) {
     final treeColumn = _NameColumn(
       currentRootLevel: controller.isDeferredApp.value
-          ? rootNode.children[0].level
+          ? rootNode.children.first.level
           : rootNode.level,
     );
     final sizeColumn = _SizeColumn();
@@ -110,7 +110,7 @@ class _SizeColumn extends ColumnData<TreemapNode> {
         );
 
   @override
-  dynamic getValue(TreemapNode dataObject) => dataObject.byteSize;
+  Comparable getValue(TreemapNode dataObject) => dataObject.byteSize;
 
   @override
   String getDisplayValue(TreemapNode dataObject) {
@@ -220,7 +220,7 @@ class _DiffColumn extends ColumnData<TreemapNode> {
 
   // Ensure sort by absolute size.
   @override
-  dynamic getValue(TreemapNode dataObject) => dataObject.unsignedByteSize;
+  int getValue(TreemapNode dataObject) => dataObject.unsignedByteSize;
 
 // TODO(peterdjlee): Add up or down arrows indicating increase or decrease for display value.
   @override
--- a/packages/devtools_app/lib/src/screens/debugger/breakpoint_manager.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/breakpoint_manager.dart
@@ -22,15 +22,13 @@ class BreakpointManager extends Disposer {
 
   final _breakPositionsMap = <String, List<SourcePosition>>{};
 
-  final _breakpoints = ValueNotifier<List<Breakpoint>>([]);
-
   ValueListenable<List<Breakpoint>> get breakpoints => _breakpoints;
-
-  final _breakpointsWithLocation =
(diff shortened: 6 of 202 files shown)


packages/devtools_app/lib/src/app.dart
```diff
@@ -112,21 +112,21 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
     });
 
     _isDarkThemeEnabled = preferences.darkModeTheme.value;
-    preferences.darkModeTheme.addListener(() {
+    addAutoDisposeListener(preferences.darkModeTheme, () {
```
jacob314, line 115: Fyi @polina-c this caught some cases where were adding listeners but never removing.

packages/devtools_app/lib/src/app.dart
```diff
@@ -277,8 +278,8 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
         return DevToolsScaffold.withChild(
           key: UniqueKey(),
           ideTheme: ideTheme,
-          child: _providedControllers(
-            offline: true,
+          child: MultiProvider(
```
jacob314, line 281: @kenzieschmoll not sure if this is clearer.
Technically this doesn't solve the lint as the lint still gets mad about methods returning List<Widget> but I think it solves the spirit and in some ways is a little clearer than the opaque helper.

packages/devtools_app/lib/src/app.dart
```diff
@@ -456,7 +453,7 @@ class OpenSettingsAction extends StatelessWidget {
     return DevToolsTooltip(
       message: 'Settings',
       child: InkWell(
-        onTap: () async {
+        onTap: () {
```
jacob314, line 456: here and elsewhere the lints caught a ton of cases where we were pretending we were doing useful async things when we weren't. Fyi @pq

packages/devtools_app/lib/src/app.dart
```diff
@@ -576,12 +574,15 @@ class CheckboxSetting extends StatelessWidget {
 /// Conditional screens can be added to this list, and they will automatically
 /// be shown or hidden based on the [Screen.conditionalLibrary] provided.
 List<DevToolsScreen> get defaultScreens {
+  final inspectorTreeController = InspectorTreeController();
```
jacob314, line 577: @kenzieschmoll this is an attempt to placate the duplicate args lint which I gave up on. This one should probably be reverted.

packages/devtools_app/lib/src/framework/framework_core.dart
```diff
@@ -19,8 +19,9 @@ import '../shared/primitives/utils.dart';
 import '../shared/scripts/script_manager.dart';
 import '../shared/survey.dart';
 
-typedef ErrorReporter = void Function(String title, dynamic error);
+typedef ErrorReporter = void Function(String title, Object error);
```
jacob314, line 22: lints were good at catching a ton of uses of dynamic that were left in our code.

packages/devtools_app/lib/src/framework/landing_screen.dart
```diff
@@ -177,7 +157,7 @@ class _ConnectDialogState extends State<ConnectDialog>
             ),
             const SizedBox(width: defaultSpacing),
             ElevatedButton(
-              onPressed: actionInProgress ? null : _connect,
+              onPressed: actionInProgress ? null : () => unawaited(_connect()),
```
jacob314, line 160: this is one of the more unfortunate cases for the lints. They make you call out the cases where you are firing and forgetting async methods. On the other hand, that does make the code look a bit ugly.

packages/devtools_app/lib/src/framework/landing_screen.dart
```diff
@@ -251,20 +251,21 @@ class ImportFileInstructions extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final textTheme = Theme.of(context).textTheme;
```
jacob314, line 254: @kenzieschmoll I love this lint warning about duplicate code. It false positives a little bit with enum entries and methods with side effects but in general is tremendously useful in helping people optimize duplicate lookups in build methods.

packages/devtools_app/lib/src/framework/notifications_view.dart
```diff
@@ -139,7 +139,21 @@ class _NotificationsState extends State<_Notifications> with AutoDisposeMixin {
...
   }
 
-  Widget _buildOverlay(BuildContext context) {
+  @override
+  Widget build(BuildContext context) {
+    return widget.child;
+  }
+}
+
+class _NotificationOverlay extends StatelessWidget {
```
jacob314, line 148: @kenzieschmoll I moved some but not all cases where we had functions that should have been widgets over to using widgets.

polina-c (approved):

incendial:
@jacob314 thank you for mentioning!

> avoid-ignoring-return-values. <-- too many false positives. wish this one could be enabled.

Yeah, I like the idea behind it too, but unfortunately, it's really hard to distinguish if a value supposed to be used or was left unused by mistake. Right now this is more like a "run from time to time" type of a rule. Sorry for that.

> This one seems interesting but has too many false positives. Gave it a try including kitchen sink adding parameters to ignore. Generally it didn't seem to catch that many real issues but had tons of false positives. # - no-equal-arguments:

Do you mean that the amount of false positives was too much to even list in the rule config?

> avoid-returning-widgets  This one is nice but has a lot of false positives. Would be great to figure out how to make it more tolerant of reasonable patterns where a separate widget would just make things ugly.

That's interesting. If I remember correctly, we cleaned it up from all the known false positives we had. Could you share the cases you consider false positive for this rule so I can handle them properly?

packages/devtools_app/lib/src/framework/landing_screen.dart
incendial: > It false positives a little bit with enum entries 

Could you share an example? I'd like to fix that. Not sure that the problem with side effects is fixable, probably a candidate for // ignore.

packages/devtools_app/lib/src/app.dart
kenzieschmoll: The only downside of pulling the MultiProvider out of the helper is that this creates some duplicated code where we have to manually specify MultiProvider wherever _providedControllers is used, but I do see how _providedControllers already sounds like it should return a List as is. 

Why did `child: _providedControllers` trigger a lint to begin with?

packages/devtools_app/lib/src/app.dart
kenzieschmoll: what does the duplicate args lint prevent? two arguments aren't allowed to have the same value?

agreed this should be reverted as this is more verbose without an obvious benefit

packages/devtools_app/lib/src/framework/landing_screen.dart
```diff
@@ -130,29 +130,9 @@ class _ConnectDialogState extends State<ConnectDialog>
...
-          _buildConnectInput(),
-        ],
-      ),
-    );
-  }
+    final textTheme = Theme.of(context).textTheme;
 
-  Widget _buildConnectInput() {
-    return Column(
+    final connectorInput = Column(
```
kenzieschmoll, line 135: should we make this a private helper widget instead?

packages/devtools_app/lib/src/framework/notifications_view.dart
```diff
@@ -114,7 +114,7 @@ class _NotificationsState extends State<_Notifications> with AutoDisposeMixin {
     bool didDismiss = false;
     // Make a copy so we do not remove a notification from [_notifications]
     // while iterating over it.
-    final notifications = List<_Notification>.from(_notifications);
+    final notifications = List<_Notification>.of(_notifications);
```
kenzieschmoll, line 117: was this caught by a lint? what is the lint preventing?

packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
```diff
@@ -126,8 +126,8 @@ class _SizeColumn extends ColumnData<TreemapNode> {
 
   @override
   int compare(TreemapNode a, TreemapNode b) {
-    final Comparable valueA = getValue(a);
-    final Comparable valueB = getValue(b);
+    final Comparable valueA = getValue(a) as Comparable;
+    final Comparable valueB = getValue(b) as Comparable;
```
kenzieschmoll, line 130: are these casts still necessary after changing the return type of getValue?

packages/devtools_app/lib/src/screens/debugger/breakpoint_manager.dart
```diff
@@ -130,18 +129,18 @@ class BreakpointManager extends Disposer {
   }
 
   void _updateAfterIsolateReload(
-    Event reloadEvent,
+    Event _,
```
kenzieschmoll, line 132: probably don't need this trailing comma anymore since this should all fit on one line

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -1472,10 +1496,12 @@ final copyPackagePathOption = ScriptPopupMenuOption(
 final copyFilePathOption = ScriptPopupMenuOption(
   label: 'Copy file path',
   icon: Icons.content_copy,
-  onSelected: (_, controller) async {
-    return Clipboard.setData(
-      ClipboardData(text: await fetchScriptLocationFullFilePath(controller)),
-    );
+  onSelected: (_, controller) {
+    unawaited(() async {
+      Clipboard.setData(
```
kenzieschmoll, line 1501: this is confusing... why does Clipboard.setData need to be wrapped as an unawaited async callback?

packages/devtools_app/lib/src/screens/debugger/controls.dart
```diff
@@ -88,7 +90,7 @@ class _DebuggingControlsState extends State<DebuggingControls>
               // Enable while paused + not resuming and selected isolate is not
               // a system isolate.
               onPressed: ((isPaused && !resuming) && !isSystemIsolate)
-                  ? controller.resume
+                  ? () => unawaited(controller.resume())
```
kenzieschmoll, line 93: this is a bit uglier to have to do this everywhere

packages/devtools_app/lib/src/screens/debugger/variables.dart
```diff
@@ -79,14 +78,12 @@ class ExpandableVariable extends StatelessWidget {
       shrinkWrap: true,
       dataDisplayProvider: (variable, onPressed) =>
           displayProvider(context, variable, onPressed, debuggerController),
-      onItemSelected: (variable) async =>
-          onItemPressed(variable, debuggerController),
+      onItemSelected: (variable) async => onItemPressed(variable),
```
kenzieschmoll, line 81: does onItemPressed need to be awaited or unawaited?

packages/devtools_app/lib/src/screens/inspector/inspector_screen_details_tab.dart
```diff
@@ -122,14 +122,10 @@ class _InspectorExpandCollapseButtonsState
...
   void _onCollapseClick() {
-    unawaited(
-      blockWhileInProgress(() async {
-        ga.select(
-          gac.inspector,
-          gac.collapseAll,
-        );
-        await widget.controller.collapseDetailsToSelected();
-      }),
+    ga.select(
```
kenzieschmoll, line 125: why do we no longer want to blockWhileInProgress?

packages/devtools_app/lib/src/screens/memory/shared/primitives/painting.dart
```diff
@@ -25,13 +25,11 @@ Widget createDashWidget(Color color) {
 }
 
 Widget createSolidLine(Color color) {
-  return Padding(
+  return Container(
+    height: 6,
+    width: 20,
```
kenzieschmoll, line 30: not your fault but these should be named consts

packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
```diff
@@ -565,7 +553,7 @@ class NetworkRequestOverviewView extends StatelessWidget {
     );
   }
 
-  Widget _buildHttpTimeGraph(BuildContext context) {
+  Widget _buildHttpTimeGraph() {
```
kenzieschmoll, line 556: we should move these to private helper widgets instead of helper methods

packages/devtools_app/lib/src/screens/performance/panes/raster_stats/raster_stats_controller.dart
```diff
@@ -68,4 +68,14 @@ class RasterStatsController extends PerformanceFeatureController {
...
+
+  @override
+  Future<void> init() {
+    return Future.value();
+  }
+
+  @override
+  Future<void> onBecomingActive() {
+    return Future.value();
+  }
```
kenzieschmoll, line 80: what lint is forcing these overrides?

kenzieschmoll: if `PerformanceFeatureController` returns these values by default, do the subclasses need these empty overrides?

packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_controller_web.dart
```diff
@@ -106,6 +106,7 @@ class PerfettoControllerImpl extends PerfettoController {
       ..height = '100%'
       ..width = '100%';
 
+    // TODO(jacobr): why does this have to be ignored?
```
kenzieschmoll, line 109: I'm not sure on the details of why, but this is a weird flutter web thing. @ditman do you know the reasoning for this?

packages/devtools_app/lib/src/screens/profiler/cpu_profile_controller.dart
```diff
@@ -458,13 +454,15 @@ class CpuProfilerController
...
+    } catch (e, stackTrace) {
       // In the event of an error, reset the data to the original CPU profile.
       final filteredOriginalData = cpuProfileStore.lookupProfile(
         label: _wrapWithFilterSuffix(userTagNone),
       )!;
       _dataNotifier.value = filteredOriginalData.getActive(viewType.value);
-      throw Exception('Error loading data with tag "$tag": ${e.toString()}');
+      Error.throwWithStackTrace(
+          Exception('Error loading data with tag "$tag": ${e.toString()}'),
+          stackTrace);
```
kenzieschmoll, line 465: nit: tc

packages/devtools_app/lib/src/screens/profiler/cpu_profile_controller.dart
```diff
@@ -492,16 +490,14 @@ class CpuProfilerController
...
-        data = CpuProfilePair.fromUserTag(fullData, tag);
-      }
+      data = tag == groupByUserTag || tag == groupByVmTag
+          ? CpuProfilePair.withTagRoots(
+              fullData,
+              tag == groupByUserTag
+                  ? CpuProfilerTagType.user
+                  : CpuProfilerTagType.vm,
+            )
+          : CpuProfilePair.fromUserTag(fullData, tag);
```
kenzieschmoll, line 500: this double nest is harder to read imo

packages/devtools_app/lib/src/shared/ui/gtags.dart
```diff
@@ -22,7 +22,7 @@ import '../analytics/analytics.dart' as ga;
 bool _debugAnalytics = false;
 
 @JS('gtag')
-external void _gTagCommandName(String command, String name, [dynamic params]);
+external void _gTagCommandName(String command, String name, [Object params]);
```
kenzieschmoll, line 25: should this be `Object?` - here and below

packages/devtools_app/test/test_infra/utils/file_utils.dart
```diff
@@ -33,11 +33,9 @@ Future<String> resolvePackagePath(String package) async {
   // so the packages directory is the current directory's grandparent.
   // TODO(https://github.com/flutter/flutter/issues/20907): Remove the workaround here.
   String packagesPath;
-  if (io.Directory.current.path.endsWith('test')) {
-    packagesPath = io.Directory.current.parent.parent.path;
-  } else {
-    packagesPath = io.Directory.current.parent.path;
-  }
+  packagesPath = io.Directory.current.path.endsWith('test')
```
kenzieschmoll, line 36: I wonder why the duplicated code lint isn't catching that `io.Directory.current` and `io.Directory.current.parent` are duplicated

packages/devtools_app/lib/src/app.dart
```diff
@@ -191,10 +191,10 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
   }
 
   Widget _buildTabbedPage(
-    BuildContext context,
+    BuildContext _,
```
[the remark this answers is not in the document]
jacob314: _buildTabbedPage is private and the only use is within this file is where it torn off. Glad to hear any number of `_` should be allowed.

packages/devtools_app/lib/src/app.dart
jacob314: The lint is one to guide you towards creating widgets rather than creating functions that return widgets. It catches a lot of ugly code (particularly cases where the BuildContext was passed around) as well as some false positives like this.

jacob314: I wonder if the duplicate code is a little more idiomatic in that it is clear what we are doing is injecting providers.

packages/devtools_app/lib/src/app.dart
jacob314: @kenzi let me know if these cases where we are doing async stuff in tap handlers is useful or noise.

packages/devtools_app/lib/src/app.dart
jacob314: yeah I don't think it is useful. I do think code is generally a bit harder to read when there are multiple args passed in with the exact same expression particularly if the args aren't named but generally I think it has far too many false positives.
Reverted.

packages/devtools_app/lib/src/framework/landing_screen.dart
jacob314: The lint fires for no good reason if you import a library with a named prefix.
For example:
```
import 'library_with_enum' as libraryName
void main() {
  print(libraryName.EnumName.entry1);
  print(libraryName.EnumName.entry2);

}
```

packages/devtools_app/lib/src/app.dart
jacob314: @incendial this was the case where I couldn't keep the code idiomatic and avoid the lint about functions returning widgets. Seems odd the lint ever fires for functions returning Lists of widgets. Notice that the lint still fires with my fix here as it is upset about the method called that returns a List of all providers for the class.

packages/devtools_app/lib/src/app.dart
jacob314: > > This one seems interesting but has too many false positives. Gave it a try including kitchen sink adding parameters to ignore. Generally it didn't seem to catch that many real issues but had tons of false positives. # - no-equal-arguments:
> 
> Do you mean that the amount of false positives was too much to even list in the rule config?
Yeah I was tossing tons of names in the false positives list in the rules config but not seeing any cases where the lint was adding value so I gave up.

> > avoid-returning-widgets  This one is nice but has a lot of false positives. Would be great to figure out how to make it more tolerant of reasonable patterns where a separate widget would just make things ugly.
> 
> That's interesting. If I remember correctly, we cleaned it up from all the known false positives we had. Could you share the cases you consider false positive for this rule so I can handle them properly?
I tagged you in the comment where I tried to make this work idiomatically for a case where we needed a helper method that computed the list of all Providers to inject with a MultiProvider.

packages/devtools_app/lib/src/framework/landing_screen.dart
jacob314: I think this pattern is better for cases where you are just trying to avoid a very deep and hard to read build method. The # of params this widget would need relative to its impl would be silly and placing the code inline avoids the illusion that Flutter does anything smart optimizing it.

packages/devtools_app/lib/src/framework/notifications_view.dart
jacob314: This is a lint suggested at https://github.com/dart-lang/linter/issues/2555
that Dart Code Metrics implemented.

packages/devtools_app/lib/src/screens/app_size/app_size_table.dart
jacob314: Unfortunately yes. They were hidden by using `dynamic`.

packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_controller_web.dart
ditman: This needs to be ignored because the API surface of `dart:ui` for the web is bigger than for mobile (it has the `webOnlyXXX` methods that are not defined in the mobile version, for example, or the `platformViewRegistry` object).

`analyzer` flags this as you calling "things that don't exist" in the API because it only sees the proper `dart:ui` mobile package, not the web re-implementation with extra methods.

A workaround to this ignore is to apply a _shim_ for the web version of dart:ui (but I've only used it in web-only packages), like [this one](https://github.com/flutter/plugins/tree/fd2841fd065f07ad429d56cd54ec9723f4d1953b/packages/video_player/video_player_web/lib/src/shims), [used here](https://github.com/flutter/plugins/blob/fd2841fd065f07ad429d56cd54ec9723f4d1953b/packages/video_player/video_player_web/lib/video_player_web.dart#L12).

* This is the (venerable) tracking issue: https://github.com/flutter/flutter/issues/41563

packages/devtools_app/lib/src/app.dart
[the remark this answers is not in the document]
incendial: Do I understand correctly: you mean that for private methods used as tear off the rule should not check for unused parameters?

packages/devtools_app/lib/src/app.dart
incendial: I see, never seen this case before, tbh 😅. If I change the rule to ignore Providers, will this be enough? The more I look at this case, the more I don't understand how a possible "right way" without a method could look like.

> Seems odd the lint ever fires for functions returning Lists of widgets

Very common case is when people create a private method returning List<Widget> and then spread it in the widget's Column or Row.

packages/devtools_app/lib/src/app.dart
incendial: > Yeah I was tossing tons of names in the false positives list in the rules config but not seeing any cases where the lint was adding value so I gave up.

This rule initially was created for models mapping, when you can have like 2 String fields and then accidentally pass same value to both fields. Maybe it should be restricted somehow to not trigger on widgets, but I'm not sure. 

> I tagged you in the comment where I tried to make this work idiomatically for a case where we needed a helper method that computed the list of all Providers to inject with a MultiProvider.

Thank you, answered above.

packages/devtools_app/lib/src/framework/landing_screen.dart
incendial: Thank you, I've added to backlog!

packages/devtools_app/lib/src/app.dart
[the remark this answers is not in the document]
jacob314: correct as the unused parameters were only there to match the signature required to use it as a tearoff. Its similar to why you shouldn't complain about unused parameters for an @override

packages/devtools_app/lib/src/app.dart
jacob314: So far the only case that is really blocking me is the provider case. The Column and Row cases are interesting. I do buy that those cases are probably usually bugs.

packages/devtools_app/lib/src/app.dart
jacob314: Maybe it should only apply to certain types of parameters. Seems most relevant for string fields where there is more of a sign of duplicate logic than for num fields or arbitrary class fields.

packages/devtools_app/test/test_infra/utils/file_utils.dart
jacob314: I tuned it up to 3 dupes to reduce the # of false positives. Can revisit with 2 dupes after the initial cleanup lands.

packages/devtools_app/lib/src/shared/ui/gtags.dart
jacob314: Done. Here and elsewhere made a lot more cases of `dynamic` into `Object?` instead of `Object`

packages/devtools_app/lib/src/shared/charts/treemap.dart
```diff
@@ -609,11 +609,7 @@ class TreemapNode extends TreeNode<TreemapNode> {
     if (!showDiff) {
       return backgroundColor ?? mainUiColor;
     }
-    if (byteSize < 0) {
-      return treemapDecreaseColor;
-    } else {
-      return treemapIncreaseColor;
-    }
+    return byteSize < 0 ? treemapDecreaseColor : treemapIncreaseColor;
```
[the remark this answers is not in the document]
jacob314: Fixed.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_controller.dart
jacob314: reverted

packages/devtools_app/lib/src/app.dart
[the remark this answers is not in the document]
incendial: Added to backlog, will ship in the next release

packages/devtools_app/lib/src/app.dart
incendial: Good point, I'll take some time to think

packages/devtools_app/lib/src/app.dart
incendial: Thank you! Added to backlog, will ship in the next release

CoderDake (commented):
Took a look at part of this, only 1 comment so far

packages/devtools_app/lib/src/screens/profiler/cpu_profile_controller.dart
jacob314: Done. Interestingly this is a case the DCM trailing comma lint will flag. As a follow up I may try enabling it.

packages/devtools_app/lib/src/framework/landing_screen.dart
jacob314: It would be interesting if there was a version of the button class that played nicely with async callbacks. For example, the ideal version wouldn't `unawait` the future (which just drops it on the ground) and would instead make the button show as disabled until the onPressed callback completes. That would probably be generally useful in DevTools. We have hacked around it with custom logic on some pages but a general solution might be feasible.

packages/devtools_app/lib/src/screens/debugger/debugger_screen.dart
```diff
@@ -359,17 +361,19 @@ class _DebuggerStatusState extends State<DebuggerStatus> with AutoDisposeMixin {
   void initState() {
     super.initState();
 
-    addAutoDisposeListener(widget.controller.isPaused, _updateStatus);
+    addAutoDisposeListener(
+        widget.controller.isPaused, () => unawaited(_updateStatus()));
```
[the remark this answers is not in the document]
jacob314: done. All regular lints are now resolved here and elsewhere.

packages/devtools_app/lib/src/screens/debugger/debugger_screen.dart
```diff
@@ -359,17 +361,19 @@ class _DebuggerStatusState extends State<DebuggerStatus> with AutoDisposeMixin {
...
   }
 
   @override
   void didUpdateWidget(DebuggerStatus oldWidget) {
     super.didUpdateWidget(oldWidget);
 
     // todo: should we check that widget.controller != oldWidget.controller?
-    addAutoDisposeListener(widget.controller.isPaused, _updateStatus);
+    addAutoDisposeListener(
+        widget.controller.isPaused, () => unawaited(_updateStatus()));
```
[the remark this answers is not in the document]
jacob314: done here and elsewhere.

packages/devtools_app/lib/src/screens/debugger/variables.dart
jacob314: Fyi @incendial. Wonder why the lints for unneeded async didn't trigger here.
This code could be further simplified to just 
`onItemSelected: onItemPressed`

packages/devtools_app/lib/src/screens/inspector/inspector_screen_details_tab.dart
jacob314: @incendial this is an example of a good success story for the Dart Code Metrics lints. The lints caught that ga.select never needed to be async which triggered further lints to clarify that we never needed to be calling `blockWhileInProgress` as nothing async was actually happening in the whole block.

packages/devtools_app/lib/src/screens/memory/shared/primitives/painting.dart
jacob314: Agreed. FYI there is a Dart Code Metrics lint for exactly that problem we can enable next. Leaving moving these const off this CL as it is already too big and too error prone.

packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
jacob314: Leaving that for a follow-up enabling the DCM lint to use private helper widgets after enabling the associated lint.

packages/devtools_app/lib/src/screens/performance/panes/raster_stats/raster_stats_controller.dart
jacob314: This was a judgment call. The base class had 3 methods tagged as async that weren't really async. 1 of the 3 methods was actually overridden everywhere so could be made abstract. These other two were overridden almost everywhere except for this case where a no-op implementation had to be added. In general, if a member should almost always be overridden I think it is cleaner to make it abstract. In addition, overriding a method returning future is very ugly as you end up awaiting the super version of the method which is quite silly given the super version was. a no-op.

packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_controller_web.dart
jacob314: Thanks. Referenced the issue.

jacob314:
All comments addressed and build is now green. PTAL.

kenzieschmoll (approved):

packages/devtools_app/lib/src/screens/debugger/variables.dart
incendial: Agree, noted

PR #4932, opened 2022-12-10
Document method contract for the case study.

polina-c:
The contract should be explained to make it clear what fixes will not break the functionality.

### DIFF
--- a/case_study/memory_leaks/leaking_counter_1/lib/main.dart
+++ b/case_study/memory_leaks/leaking_counter_1/lib/main.dart
@@ -29,32 +29,39 @@ class MyHomePage extends StatefulWidget {
 }
 
 class MyIncrementer {
-  MyIncrementer(this.increment);
-
+  MyIncrementer(this.increment, this.screen);
+  final Scaffold? screen;
   final VoidCallback increment;
 }
 
 class _MyHomePageState extends State<MyHomePage> {
   int _counter = 0;
 
-  late MyIncrementer _incrementer = MyIncrementer(() => setState(() {
-        _counter++;
-      }));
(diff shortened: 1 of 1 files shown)


case_study/memory_leaks/leaking_counter_1/lib/main.dart
```diff
@@ -29,32 +29,39 @@ class MyHomePage extends StatefulWidget {
 }
 
 class MyIncrementer {
-  MyIncrementer(this.increment);
-
+  MyIncrementer(this.increment, this.screen);
+  final Scaffold? screen;
```
kenzieschmoll, line 33: why do we need to store the widget in this class? It is bad practice to store widget objects, so I'm wondering if we should put this into a case study app. (unless of course, this is to demonstrate that it is bad practice to store Widget objects?)

polina-c: yes, it is bad practice and source of leak

kenzieschmoll (approved):

case_study/memory_leaks/leaking_counter_1/lib/main.dart
jacob314: A slightly more natural looking thing you could do would be to store a `BuildContext` object. Our quick fix to turn a function into a widget often accidentally adds a BuildContext as a parameter to a widget.

polina-c: I moves screen calculation out of closure to make it leaking.
I added context, but it does not contribute to the leak as it is always the same. What is leaking here is every time built screen.

Update: https://github.com/flutter/devtools/pull/4942

Issue #2856, opened 2021-04-02
safari show white blank page


allanGoldZhiKang:
![image](https://user-images.githubusercontent.com/29418848/113389000-da937480-93c1-11eb-8b8a-b6a8855e8c21.png)

allanGoldZhiKang:
![image](https://user-images.githubusercontent.com/29418848/113389053-f565e900-93c1-11eb-8064-04ed1877fa56.png)

jacob314:
Fyi @DanTup 
Safari should work better than this.

DanTup:
Using the latest bleeding-edge build of Dart and running `dart devtools` then trying to open in Safari I also get a quite page. The error looks like this:

![Screenshot 2022-12-13 at 10 51 41](https://user-images.githubusercontent.com/1078012/207298857-cee6a3d9-abba-4586-963f-556a05646f4d.png)

![Screenshot 2022-12-13 at 10 52 37](https://user-images.githubusercontent.com/1078012/207298865-fe1d7680-07fa-4a97-9cf5-18d4ecfe2b4e.png)

I found that regex in the syntax highlighting grammar.. I don't know why it's ok in Chrome but broken in Safari, but I'll do some digging..

DanTup:
I have a fix up at https://github.com/flutter/devtools/pull/4938 that will just disable syntax highlighting if it fails to initialize (since making it work without negative-lookbehinds will be significant effort). With that change, DevTools loads and works in Safari:

![image](https://user-images.githubusercontent.com/1078012/207322555-5c57f700-7b08-4175-8ad8-084fa579ac1e.png)

PR #4938, opened 2022-12-13
Don't fail at startup if textmate grammar can't be parsed by browser

DanTup:
The textmate grammar uses negative lookbehinds which are not supported in Safari. I've had a go at changing the grammar, but I can't find a way to support what we need (without a significant change to how the textmate grammar works) so for now this change will just handle errors and leave syntax highlighting disabled for browsers that don't support this.

Before this change, DevTools fails to load in safari:

image

With the change, it loads and prints a warning about the failed syntax highlighting in the console:

Screenshot 2022-12-13 at 12 40 16

And using the app works as normal:

Screenshot 2022-12-13 at 12 43 35

Fixes https://github.com/flutter/devtools/issues/2856.

### DIFF
--- a/packages/devtools_app/lib/src/screens/debugger/syntax_highlighter.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/syntax_highlighter.dart
@@ -8,6 +8,7 @@ import 'dart:convert';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart' show rootBundle;
 
+import '../../shared/config_specific/logger/logger.dart';
 import '../../shared/primitives/utils.dart';
 import '../../shared/theme.dart';
 import 'span_parser.dart';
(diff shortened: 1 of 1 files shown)


elliette (approved):

DanTup:
@elliette should I hold off landing this until bots are green, or is it likely they'll be this way for a while? (I'm less familiar with the process and current status here)

elliette:
> @elliette should I hold off landing this until bots are green, or is it likely they'll be this way for a while? (I'm less familiar with the process and current status here)

@kenzieschmoll for input! I haven't done much devtools development recently

kenzieschmoll:
I'm trying to fix the bots now - https://github.com/flutter/devtools/pull/4941. The breakage is due to the move to Dart version > 3.0.0, since a lot of the packages we depend on have an upper bound sdk constraint of 3.0.0.

kenzieschmoll:
the temporary fix has landed (https://github.com/flutter/devtools/pull/4944). If you merge master, the bots should go green.

DanTup:
Yep, all green. Thanks!

DanTup:
@kenzieschmoll @CoderDake I realised I didn't add this to any changelog, and there doesn't seem to be a milestone for the next release. Assuming this should be noted (it makes DevTools load in Safari), is https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/framework/release_notes/release-notes-next.md the correct place to add it? (it looks like it, though there's nothing else currently listed there so I'm unsure if it's just a template).

kenzieschmoll:
Yes, any notable user facing issues should be documented in the release-notes-next.md file. We automatically generate the CHANGELOG when we do the release, but the release notes should be drafted as we go.

DanTup:
Thanks - I've opened https://github.com/flutter/devtools/pull/4966 to add this.

wilmarques:
I was waiting for this so much!!!

When will it land?

Issue #4943, opened 2022-12-13
DevTools is broken for Dart 3.0.0


kenzieschmoll:
The first failure we see is due to the upper bound constraint of package:cupertino_icons (attempted fix [here](https://github.com/flutter/packages/pull/2938)). We likely will see similar issues with other packages that have an upper bound sdk constraint of 3.0.0.

kenzieschmoll:
Pinning to `flutter-3.7-candidate.3` to temporarily unbreak the build:
```
Flutter 3.7.0-4.0.pre.1 • channel flutter-3.7-candidate.3 • git@github.com:flutter/flutter.git
Framework • revision e937e159e2 (6 days ago) • 2022-12-07 11:00:11 -0800
Engine • revision 0efb0ef405
Tools • Dart 2.19.0 (build 2.19.0-463.0.dev) • DevTools 2.20.0
```

kenzieschmoll:
This would have caught if we had registered our tests in the flutter/tests: https://github.com/flutter/devtools/issues/4433

PR #4945, opened 2022-12-13
Update cupertino_icons constraint to one supporting sound null safety

parlough:
Versions of cupertino_icons before 1.0.3 didn't have a lower bound of `>= 2.12`, breaking usage of Dart 3, which requires sound null safety. This caused package resolution to fail when using a Dart 3 SDK (https://github.com/flutter/devtools/actions/runs/3688647196/jobs/6243745534).

Contributes to #4943


parlough:
Looks like https://github.com/flutter/devtools/pull/4941 incorporates the necessary changes. Closing in favor of that PR..

PR #4955, opened 2022-12-16
Stop double counting self-referenced classes

polina-c:


packages/devtools_app/lib/src/screens/memory/shared/heap/heap.dart
```diff
@@ -194,19 +204,27 @@ class ObjectSet extends ObjectSetStats {
   static ObjectSet empty = ObjectSet()..seal();
 
   final objectsByCodes = <IdentityHashCode, AdaptedHeapObject>{};
+  final excludedFromRetained = <IdentityHashCode>{};
```
jacob314, line 207: nit: is this "excludedFromRetained" or "alreadyIncludedInRetained" or similar?

packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
```diff
@@ -108,21 +108,28 @@ class AdaptedHeapData {
...
   HeapPath(this.objects);
 
   final List<AdaptedHeapObject> objects;
 
+  late final bool isRetainedBySameClass = () {
+    if (objects.isEmpty) return false;
+
+    final theClass = objects.last.heapClass;
+
+    final firstWithSameClass =
```
jacob314, line 121: This is simpler and I think the intent is a bit clearer if rewritten as:

```
if (objects.length < 2) return false;
return objects.take(objects.length - 1).any((object) => object.heapClass == theClass);
````

File: packages/devtools_app/lib/src/screens/memory/shared/heap/heap.dart (hunk 1 above)
Comment (polina-c) [reply packages/devtools_app/lib/src/screens/memory/shared/heap/heap.dart]: renamed

File: packages/devtools_app/lib/src/screens/memory/panes/diff/controller/heap_diff.dart [hunk 3]
```diff
@@ -164,15 +164,18 @@ class ObjectSetDiff {
       if (inBefore && inAfter) continue;
 
       final object = before.objectsByCodes[code] ?? after.objectsByCodes[code]!;
+      final excludeFromRetained =
```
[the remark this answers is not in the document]
polina-c: refactored for readability

packages/devtools_app/test/memory/shared/heap/heap_test.dart
```diff
@@ -0,0 +1,117 @@
...
+      ],
+      rootIndex: 0,
+    ),
+    expectedClassARetainedSize: 4,
+  ),
+];
+
+void main() {
+  for (final t in _classSizeTests) {
+    test(
```
jacob314, line 88: In general avoid dynamically creating test cases like this. It makes it hard for IDE tooling. 
@natebosch would you be ok with a lint saying not to nest `test` in a for loop?
Instead have a single test case and in the test case make sure the expect clauses make it unambiguous which class name the failure is about. For example by augmenting the output returned for a failure or by having multiple expect statements one for each class name in the heap you are interested in.

packages/devtools_app/test/memory/shared/heap/model_test.dart
```diff
@@ -26,4 +76,21 @@ void main() {
 
     expect(json, AdaptedHeapData.fromJson(json).toJson());
   });
+
+  for (final t in _heapPathTests) {
```
jacob314, line 80: Same as other test.

packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
polina-c: `take` will create one more array

Knowing that the method will run for every object, I would avoid it. 

Updated early return.

packages/devtools_app/lib/src/screens/memory/shared/heap/heap.dart
```diff
@@ -133,15 +133,19 @@ class SingleClassStats extends ClassStats {
     assert(!isSealed);
     final object = data.objects[objectIndex];
     assert(object.heapClass.fullName == heapClass.fullName);
-    objects.countInstance(object);
 
     final path = data.retainingPath(objectIndex);
+    objects.countInstance(
+      object,
+      excludeFromRetained: path?.isRetainedBySameClass ?? false,
```
jacob314, line 140: clever!

jacob314 (commented):
Looks good with a couple minor comments.

packages/devtools_app/lib/src/screens/memory/shared/heap/model.dart
polina-c: actually, I see 'take' returns itarable

polina-c: done

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/heap_diff.dart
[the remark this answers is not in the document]
jacob314: Did you upload the refactor?
What I need is a comment explaining the invariants that make this logic correct.
Is the deal that there should really be a single retained excludedFromRetained set used for both before and after or it it something different?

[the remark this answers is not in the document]
polina-c: `if (inBefore && inAfter) continue;`

In this code any object is either in before or in after. Never together.
It is obvious for those who read the method from the beginning.

jacob314 (approved):
lgtm

Issue #4934, opened 2022-12-12
Create integration testing harness


kenzieschmoll:
Use a combination of flutter driver, webdriver, package:integration_test, etc. to write real integration tests for DevTools.

kenzieschmoll:
Issues encountered:
- https://github.com/flutter/flutter/issues/116936 (potential blocker)
- https://github.com/flutter/devtools/issues/4658 causes some noisy failures (@elliette we may want to prioritize this if it is low cost to clean up)
- Unable to run more than one test in a file (talking with @yjbanov about this)

kenzieschmoll:
Making some progress on hooking up integration tests for DevTools. First, we need to add support for screenshot tests for flutter web in package:integration_test - https://github.com/flutter/flutter/pull/117114/

kenzieschmoll:
> Unable to run more than one test in a file

Because of this bug (https://github.com/flutter/flutter/issues/116936), I commented out our call in main to `usePathUrlStrategy();`and then since that wasn't set on the first call to main, we return early and perform a redirect for the second test: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/main.dart#L26

So this means that solving this bug https://github.com/flutter/flutter/issues/116936 will be a blocker.

kenzieschmoll:
Another issue we have encountered along the way: https://github.com/flutter/flutter/issues/117415

kenzieschmoll:
tracking issue to move to SkiaGoldClient at some point here: https://github.com/flutter/devtools/issues/4978

kenzieschmoll:
This is completed.

PR #4960, opened 2022-12-16
Add initial framework for integration testing.

kenzieschmoll:
This PR adds 

- integration_tests/test/infra/test_app_driver.dart : this code was copied from the existing file flutter_test_driver.dart, but paired down to use only what we need right now. This code will be used to spin up a test app that we will connect to DevTools for integration testing.
- test_driver/integration_test.dart : this is the test driver we will use for integration tests

### DIFF
--- a/packages/devtools_app/integration_test/test_infra/io_utils.dart
+++ b/packages/devtools_app/integration_test/test_infra/io_utils.dart
@@ -1,0 +1,46 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+Stream<String> transformToLines(Stream<List<int>> byteStream) {
+  return byteStream
+      .transform<String>(utf8.decoder)
+      .transform<String>(const LineSplitter());
+}
+
+mixin IoMixin {
+  final stdoutController = StreamController<String>.broadcast();
+
+  final stderrController = StreamController<String>.broadcast();
+
+  final streamSubscriptions = <StreamSubscription<String>>[];
+
+  void listenToProcessOutput(
+    Process process, {
+    void Function(String) printCallback = _defaultPrintCallback,
+  }) {
+    streamSubscriptions.addAll([
+      transformToLines(process.stdout)
+          .listen((String line) => stdoutController.add(line)),
+      transformToLines(process.stderr)
+          .listen((String line) => stderrController.add(line)),
+
+      // This is just debug printing to aid running/debugging tests locally.
+      stdoutController.stream.listen(printCallback),
+      stderrController.stream.listen(printCallback),
+    ]);
+  }
+
+  Future<void> cancelAllStreamSubscriptions() async {
+    await Future.wait(streamSubscriptions.map((s) => s.cancel()));
+    streamSubscriptions.clear();
+  }
+
+  static void _defaultPrintCallback(String line) {
+    print(line);
+  }
+}
--- a/packages/devtools_app/integration_test/test_infra/io_utils.dart
+++ b/packages/devtools_app/integration_test/test_infra/io_utils.dart
@@ -12,7 +12,7 @@ Stream<String> transformToLines(Stream<List<int>> byteStream) {
       .transform<String>(const LineSplitter());
 }
 
-mixin IoMixin {
+mixin IOMixin {
   final stdoutController = StreamController<String>.broadcast();
 
   final stderrController = StreamController<String>.broadcast();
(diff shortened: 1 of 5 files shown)


polina-c:
Why `test_driver/integration_test.dart` is not in test_infra?
Are you sure we want to have separate test_infra in integration_test? Would it be more convenient to have it shared between different types of testing? 

Couple options:
1. take test_infra out of test and integration_test and make it shared
2. Declare that test is lower level than integration_test and use test/test-infra for both

packages/devtools_app/integration_test/test_infra/test_app_driver.dart
```diff
@@ -0,0 +1,244 @@
...
+    String? event,
+    int? id,
+    Duration? timeout,
+    bool ignoreAppStopEvent = false,
+  }) async {
+    final Completer<Map<String, dynamic>> response =
+        Completer<Map<String, dynamic>>();
+    late StreamSubscription<String> sub;
+    sub = stdoutController.stream.listen((String line) async {
+      final dynamic json = _parseFlutterResponse(line);
```
bkonyi, line 143: Consider pulling this closure into its own method.

bkonyi, line 143: Is this type necessary? `_parseFlutterResponse` has a known, non-`dynamic` return type.

packages/devtools_app/integration_test/test_infra/test_app_driver.dart
```diff
@@ -0,0 +1,244 @@
...
+      if (json == null) {
+        return;
+      } else if ((event != null && json['event'] == event) ||
```
bkonyi, line 146: Nit: use constants for keys, here and elsewhere.

packages/devtools_app/integration_test/test_infra/test_app_driver.dart
```diff
@@ -0,0 +1,244 @@
...
+          (id != null && json['id'] == id)) {
+        await sub.cancel();
+        response.complete(json);
+      } else if (!ignoreAppStopEvent && json['event'] == 'app.stop') {
```
bkonyi, line 150: Do we have constants defined in `devtools_shared` for these daemon event names?

packages/devtools_app/integration_test/test_infra/test_app_driver.dart
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  Future<T> _timeoutWithMessages<T>(
+    Future<T> Function() f, {
+    Duration? timeout,
+    String? message,
+  }) {
+    // Capture output to a buffer so if we don't get the response we want we can show
+    // the output that did arrive in the timeout error.
+    final StringBuffer messages = StringBuffer();
```
bkonyi, line 183: Nit: consider removing LHS types for declarations where it's clear what the type should be.

packages/devtools_app/integration_test/test_infra/io_utils.dart
```diff
@@ -0,0 +1,46 @@
...
+import 'dart:convert';
+import 'dart:io';
+
+Stream<String> transformToLines(Stream<List<int>> byteStream) {
+  return byteStream
+      .transform<String>(utf8.decoder)
+      .transform<String>(const LineSplitter());
+}
+
+mixin IoMixin {
```
bkonyi, line 15: Ubernit: `s/Io/IO/`

bkonyi (commented):
DBR

kenzieschmoll:
> Why `test_driver/integration_test.dart` is not in test_infra? Are you sure we want to have separate test_infra in integration_test? Would it be more convenient to have it shared between different types of testing?
> 
> Couple options:
> 
> 1. take test_infra out of test and integration_test and make it shared
> 2. Declare that test is lower level than integration_test and use test/test-infra for both

Official Flutter guidance recommends code structure:
```
lib/
  ...
integration_test/
test/
```

I did try to create a shared test folder, but libraries under test/ and integration_test/ could not import anything outside of their direct folders or lib/.

Unit tests and integration tests will have some different "test_infra" so it is okay to have two folders. Once we build out our integration tests more fully, it may make sense to put anything that could be shared between test/ and integration_test into packages/devtools_test so that both test/ and integration_test can share it. This can be done at a later time though when we have a better idea of what will need to be shared.

packages/devtools_app/integration_test/test_infra/test_app_driver.dart
kenzieschmoll: probably not. All this code was written in pre-null safe Dart land and copied over in this PR. updated.

packages/devtools_app/integration_test/test_infra/test_app_driver.dart
kenzieschmoll: created enum FlutterDaemonConstants for now with a TODO to consider placing these constants in devtools_shared or in flutter depending on where they would be used.

polina-c (approved):

kenzieschmoll:
Work towards https://github.com/flutter/devtools/issues/4934

PR #4984, opened 2022-12-22
Partially enable trailing commas lint. Fixed about 200 out of 600 violations

jacob314:
RELEASE_NOTE_EXCEPTION=no changes in functionally. Just lots of commas.
Fixed all violations in non test code and some of the test violations.
In general, some of the trailing commas don't impact formatting (arrays already use flutter style without a trailing comma) and the violations from test code seem less controversial as they seem aligned with @kenzieschmoll code review comments).

### DIFF
--- a/packages/devtools_app/lib/src/framework/app_error_handling.dart
+++ b/packages/devtools_app/lib/src/framework/app_error_handling.dart
@@ -22,24 +22,26 @@ import '../shared/config_specific/logger/logger.dart';
 void setupErrorHandling(Future Function() appStartCallback) {
   // First, run all our code in a new zone.
   return runZonedGuarded(
-      // ignore: avoid-passing-async-when-sync-expected this ignore should be fixed.
-      () async {
-    WidgetsFlutterBinding.ensureInitialized();
+    // ignore: avoid-passing-async-when-sync-expected this ignore should be fixed.
+    () async {
+      WidgetsFlutterBinding.ensureInitialized();
 
-    final FlutterExceptionHandler? oldHandler = FlutterError.onError;
+      final FlutterExceptionHandler? oldHandler = FlutterError.onError;
 
-    FlutterError.onError = (FlutterErrorDetails details) {
-      _reportError(details.exception, details.stack ?? StackTrace.empty);
+      FlutterError.onError = (FlutterErrorDetails details) {
+        _reportError(details.exception, details.stack ?? StackTrace.empty);
 
-      if (oldHandler != null) {
-        oldHandler(details);
-      }
-    };
+        if (oldHandler != null) {
+          oldHandler(details);
+        }
+      };
 
-    return appStartCallback();
-  }, (Object error, StackTrace stack) {
-    _reportError(error, stack);
-  });
+      return appStartCallback();
+    },
+    (Object error, StackTrace stack) {
+      _reportError(error, stack);
+    },
+  );
 }
 
 void _reportError(Object error, StackTrace stack) {
--- a/packages/devtools_app/lib/src/framework/status_line.dart
+++ b/packages/devtools_app/lib/src/framework/status_line.dart
@@ -88,7 +88,7 @@ class StatusLine extends StatelessWidget {
             OpenAboutAction(),
           ],
         ),
-      ]
+      ],
     ];
   }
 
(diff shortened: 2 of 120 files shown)


packages/devtools_app/lib/src/framework/app_error_handling.dart
```diff
@@ -22,24 +22,26 @@ import '../shared/config_specific/logger/logger.dart';
...
-
-      if (oldHandler != null) {
-        oldHandler(details);
-      }
-    };
-
-    return appStartCallback();
-  }, (Object error, StackTrace stack) {
-    _reportError(error, stack);
-  });
```
jacob314, line 42: Example of fixes that I think match what @kenzieschmoll would suggest in a code review.

packages/devtools_app/lib/src/framework/status_line.dart
```diff
@@ -186,7 +186,7 @@ class StatusLine extends StatelessWidget {
                           style: textTheme.bodyMedium,
                           overflow: TextOverflow.clip,
                         ),
-                      ]
```
jacob314, line 189: Example where a trailing comma doesn't change what dartfmt does but where it does seem fairly consistent with how we are using trailing commas.

packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -743,7 +743,7 @@ extension AppSizeJsonFileExtension on DevToolsJsonFile {
     'macos',
     'windows',
     'linux',
-    'web'
+    'web',
```
jacob314, line 746: this case isn't consistent with how we use trailing commas but seems harmless enough. The best argument in favor for this case is just that the trailing commas make the diffs smaller if you happen to edit the List later.

jacob314 (commented):
Fyi @pq for whether this gives you insights on how we might want to tweak the trailing comma support in the Dart linter. We wouldn't want to require trailing commas at the end of all List objects but some of the other cases do help reduce extra nesting by dartfmt and seem generally aligned with Flutter style.

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
```diff
@@ -44,12 +46,12 @@ class _TraceCheckBoxColumn extends ColumnData<TracedClass>
...
     return Checkbox(
       value: item.traceAllocations,
-      onChanged: (value) async {
+      onChanged: (value) {
         ga.select(
           gac.memory,
           '${gac.MemoryEvent.tracingTraceCheck}-$value',
         );
-        await controller.setAllocationTracingForClass(item.cls, value!);
+        unawaited(controller.setAllocationTracingForClass(item.cls, value!));
```
CoderDake, line 54: is this comma related?

CoderDake (approved):
LGTM.
You'll need to add a release note exception mentioning that this isn't user facing :)

pq:
> Fyi @pq for whether this gives you insights on how we might want to tweak the trailing comma support in the Dart linter. We wouldn't want to require trailing commas at the end of all List objects but some of the other cases do help reduce extra nesting by dartfmt and seem generally aligned with Flutter style.

Interesting.  Thanks!  It looks like we can probably settle on a pretty simple heuristic and it could really cut down on code review churn.

Hilariously, such a lint would have avoided this exchange just the other day!

https://dart-review.googlesource.com/c/sdk/+/276770/12..17/pkg/analysis_server/lib/src/services/correction/fix_internal.dart#b1444

packages/devtools_app/lib/src/screens/memory/panes/allocation_tracing/allocation_profile_class_table.dart
jacob314: oops. reverted. I really need to turn on auto-save in VSCode. That was supposed to be for a separate CL but I must have left the file unsaved and then hit save-all.

jacob314:
> https://dart-review.googlesource.com/c/sdk/+/276770/12..17/pkg/analysis_server/lib/src/services/correction/fix_internal.dart#b1444

Ah interesting. @bwilkerson for the analyzer code base would you be ok with requiring trailing commas at the end of more list literals matching what this lint is doing?

bwilkerson:
My comment wasn't related to wanting to use trailing commas everywhere, it was about wanting to be consistent within a single map/list literal. I don't think I'd find a lot of value to enabling trailing_commas, but I also wouldn't object too much if other team members really want it. My biggest objection (aside from a strong belief that the formatter should just do a better job of formatting without requiring trailing commas) is that we don't have a corresponding check to tell us when we should remove trailing commas.

PR #4990, opened 2022-12-24
Update case study.

polina-c:
Remove recursion and simplify case study to make it closer to real cases.

Screenshot 2022-12-24 at 8 56 38 AM

RELEASE_NOTE_EXCEPTION=[not part of DevTools]


jacob314 (approved):
lgtm

PR #4991, opened 2022-12-24
Move some console related code from debugger and inspector screens to shared/console.

polina-c:
RELEASE_NOTE_EXCEPTION=[no functional changes]


packages/devtools_app/lib/src/shared/console/eval/inspector_service.dart
```diff
@@ -15,13 +15,14 @@ import 'package:flutter/cupertino.dart';
 import 'package:flutter/foundation.dart';
```
jacob314, line 15: This has a lot of non-console stuff in it. I would suggest either leaving it in the inspector and accepting slightly tangled imports or move it to a shared directory. Seems odd to have this core inspector logic under the console rather than inspector or shared.

packages/devtools_app/lib/src/shared/console/widgets/description.dart
```diff
@@ -255,7 +250,7 @@ class DiagnosticsNodeDescription extends StatelessWidget {
     final defaultStyle = DefaultTextStyle.of(context).style;
     final baseStyle = style ?? defaultStyle;
     TextStyle textStyle = baseStyle.merge(
-      inspector_text_styles.textStyleForLevel(
```
jacob314, line 258: this seems a bit inverted. these are much more inspector specific styles than console styles. If you want a neutral name, these could be diagnostics_text_styles.

packages/devtools_app/lib/src/shared/console/eval/diagnostics_node.dart
```diff
@@ -2,19 +2,18 @@
 // Use of this source code is governed by a BSD-style license that can be
```
jacob314, line 2: This also isn't a console thing. If you want to move it, it should go in shared not console.

packages/devtools_app/lib/src/shared/console/primitives/text_styles.dart
```diff
@@ -0,0 +1,70 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/material.dart';
+
+import '../../theme.dart';
+
+// Enum-like static classes are ok.
```
jacob314, line 9: why not continue the previous pattern of top level members imported with a prefix? In general if we were to use one pattern I would think we would migrate to that pattern not to this pattern that violates the lints we have enabled.

jacob314 (approved):
lgtm
LGTM as long as there is a cleanup CL after the 3 chained CLs are landed moving things to a common dir.

packages/devtools_app/lib/src/shared/console/primitives/text_styles.dart
polina-c: Discussed offline and agreed enum-like static classes are more convenient than top level members.

packages/devtools_app/lib/src/shared/console/eval/inspector_service.dart
polina-c: Addressed in https://github.com/flutter/devtools/pull/5006.

packages/devtools_app/lib/src/shared/console/widgets/description.dart
polina-c: Renamed and moved to shared root in #5021

PR #4992, opened 2022-12-25
Extract evalService from debuggerController that serves console.

polina-c:
RELEASE_NOTE_EXCEPTION=[no functional changes]


packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -1154,7 +1154,7 @@ class _LineItemState extends State<LineItem>
 
     if (word != '') {
       try {
-        final response = await controller.evalAtCurrentFrame(word);
+        final response = await controller.evalService.evalAtCurrentFrame(word);
```
jacob314, line 1157: This case is tricky as you wouldn't want any injected variables from the console to show up when evaluating while hovering over source code. Maybe this case should take an extra named parameter like `includeConsoleVariables: false`

packages/devtools_app/lib/src/shared/console/eval/eval_service.dart
```diff
@@ -0,0 +1,139 @@
...
+class EvalService {
+  EvalService({
+    required this.isolateRef,
+    required this.variables,
+    required this.frameForEval,
+    required this.isPaused,
+    required this.service,
+  });
+
+  final IsolateRef? Function() isolateRef;
```
jacob314, line 47: It would seem more consistent to use ValueListenable for all of these rather than callbacks.

packages/devtools_app/lib/src/shared/console/eval/eval_service.dart
```diff
@@ -0,0 +1,139 @@
...
+
+  final VmServiceWrapper Function() service;
+
+  final EvalHistory evalHistory = EvalHistory();
+
+  final cache = AutocompleteCache();
+
+  String get _isolateRefId {
+    final id = isolateRef()?.id;
+    if (id == null) return '';
```
jacob314, line 63: why is returning '' for the isolateRefId ever ok?

packages/devtools_app/lib/src/shared/console/eval/eval_service.dart
polina-c: ValueListenable looks too powerful for me. I want it to be clear that EvalService does not need to listen for the changes. It just needs up-to-date value.

packages/devtools_app/lib/src/shared/console/eval/eval_service.dart
polina-c: I do not know. I copied it from previous code. Added 'TODO'.

packages/devtools_app/lib/src/shared/console/eval/eval_service.dart
jacob314: I don't quite buy this. The autocomplete cache is bogus if the isolate changed, for example. ValueListenable does not force you to always listen for changes. As appropriate you can always just use .value to get the current value. On the other end of the spectrum, VmServiceWrapper needs to just be a fixed instance because there is no reason anyone writing this code should worry it can changed.,

polina-c: My goal is clarity of contract for the method.
I want the constructor parameters to be smallest needed set of what the object needs to function. Of course, not at cost of convenience.  

About ValueListenable  vs function. I remember a story: when I passed ValueListenable, I got comments in code review like 'why this object need to listen to every change? it is performance overhead'. And then I needed to explain that the object just needs values.

So, it affects readability of code too.

Does it change something for you?

polina-c: About VmServiceWrapper: as I understand when reconnect happens, DebuggerController is not recreated, but it resets some variables. 
So, EvalService will not recreate, but just reset. 

Yes, it may be not the best implementation, but it is what we currently have. And it will take more refactoring to change.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -578,20 +471,7 @@ class DebuggerController extends DisposableController
...
-    libraryMemberAutocompleteCache.clear();
-    libraryMemberAndImportsAutocompleteCache.clear();
-  }
-
-  /// Get the populated [Obj] object, given an [ObjRef].
-  ///
-  /// The return value can be one of [Obj] or [Sentinel].
-  Future<Obj> getObject(ObjRef objRef) {
-    return _service.getObject(_isolateRefId, objRef.id!);
+    evalService.cache.clear();
```
jacob314, line 474: this is badly coupled. Instead EvalService should listen for isolate changes itself instead of requiring the debuggerControl to call this method at the appropriate time.

packages/devtools_app/lib/src/shared/console/eval/eval_service.dart
polina-c: And, I did not understand you point about consistency. In Flutter framework there is ton of methods and constructors that take functions, not listeners.

jacob314: Each time an app connects you get a new DebuggerController. How it happens flows back to how the provider for the DebuggerController is initialized.
The private getter for the service in debuggerController is really just calling
`serviceManager.service!`
Would be clearer to just call that global from EvalService rather than making it look like it is injected in.

jacob314: In general, I think ValueListenable is almost always the right choice if something can actually change. If it can't change then pass in the actual value when initializing.

polina-c: Converted service to variable, isPaused and variables to listenable.

packages/devtools_app/lib/src/screens/debugger/codeview.dart
polina-c: Discussed and agreed `evalAtCurrentFrame` will never be invoked for global scope.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
polina-c: thanks. fixed.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -34,21 +34,29 @@ class DebuggerController extends DisposableController
...
+      _initialize();
     }
   }
 
   final codeViewController = CodeViewController();
 
+  late final EvalService evalService = EvalService(
+    isolateRef: isolateRef,
+    variables: variables,
+    frameForEval: () => _frameForEval,
```
jacob314, line 55: _frameForEval should also be made a ValueListenable for consistency. That would probably cleanup some of the debugger UI that should be listening for changes to the frameForEval.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -34,21 +34,29 @@ class DebuggerController extends DisposableController
...
     }
     if (serviceManager.hasService) {
-      initialize();
+      _initialize();
     }
   }
 
   final codeViewController = CodeViewController();
 
+  late final EvalService evalService = EvalService(
```
jacob314, line 52: make this final instead of `late final`.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
polina-c: I don't believe it makes sense to recalculate and update `frameForEval` every time `_selectedStackFrame` or `_stackFramesWithLocation` change, from both code readability and performance perspective.

Deleted getter and moved the logic to the function, as the getter is not used in other places..

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
polina-c: Initializer cannot use getters:

Screenshot 2022-12-27 at 2 56 24 PM

jacob314: Yeah. Move the initializer to the constructor to avoid this issue. In general, I would avoid initializing fields inline except for very trivial cases. Otherwise the initialization order semantics can trip some people up when reviewing.

jacob314 (approved):
I'll be out tomorrow. Would be nice to remove the late final moving it to a constructor.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
polina-c: Even after the move it is still late, because it cannot be initialized in constructor initializers, but only in the body. Did i miss something? Let's follow up then.

PR #4993, opened 2022-12-26
Enable console for memory screen.

polina-c:
RELEASE_NOTE_EXCEPTION=[changes under the flag]

1. Rename DebuggerConsole to ConsolePane and move to shared/console
2. Enable console for memory screen under experimental flag
3. Move state that needs to be shared between inspector, debugger, and console, from DebuggerController to AppState owned by ServiceConnectionManager.


packages/.vscode/launch.json
```diff
@@ -0,0 +1,59 @@
+{
+    // Adjusted copy of devtools_app/.vscode/launch.json,
```
jacob314, line 2: @kenzieschmoll what do you think of making this the default config? Seems like everyone should be seeing errors in the devtools_test package.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -52,6 +52,8 @@ class DebuggerController extends DisposableController
       isPaused: isPaused,
     );
 
+    setGlobal(EvalService, evalService);
```
jacob314, line 55: Creating a controller object should not have side effects like setting a global. Either move this logic to other places globals are set or just hang the evalService off the DebuggerController object. All places the Console was used can just be given access to the DebuggerController. Generally I think that is idiomatic as the DebugController is the source of truth for things like whether you are paused at a breakpoint and what the selected stack frame is so there is no reason the console should not know those things.

packages/devtools_app/lib/src/screens/memory/memory_screen.dart
```diff
@@ -43,6 +43,9 @@ class MemoryScreen extends Screen {
 
   @override
   Widget build(BuildContext context) => const MemoryBody();
+
+  @override
+  bool showConsole(bool embed) => true;
```
jacob314, line 48: add a TODO noting that we would ideally not show the console when embedded and instead support the same console functionality through the native console in VSCode.

packages/devtools_app/lib/src/shared/console/widgets/consolePane.dart
```diff
@@ -0,0 +1,79 @@
...
+import '../../globals.dart';
+import '../../theme.dart';
+import '../console.dart';
+import '../console_service.dart';
+import 'evaluate.dart';
+
+// TODO(devoncarew): Show some small UI indicator when we receive stdout/stderr.
+
+@override
+PreferredSizeWidget buildConsolePaneHeader() {
```
jacob314, line 18: why does this top level method have an override? I would switch it to be a 
ConosolePaneHeader widget.

packages/devtools_app/lib/src/shared/console/widgets/consolePane.dart
```diff
@@ -0,0 +1,79 @@
...
+}
+
+class _ConsolePaneState extends State<ConsolePane> {
+  ValueListenable<List<ConsoleLine>> get stdio =>
+      serviceManager.consoleService.stdio;
+
+  @override
+  void initState() {
+    super.initState();
+    serviceManager.consoleService.ensureServiceInitialized();
```
jacob314, line 58: this is a code smell. Why do you have a Stateful widget with no state? Ensuring the service is initialized should be controlled elsewhere not through initState.

packages/devtools_app/lib/src/shared/globals.dart
```diff
@@ -53,6 +54,8 @@ NotificationService get notificationService =>
 BreakpointManager get breakpointManager =>
     globals[BreakpointManager] as BreakpointManager;
 
+EvalService get evalService => globals[EvalService] as EvalService;
```
jacob314, line 57: I would remove this service or change where it is initialized. Seems a bit tightly coupled to the DebuggerController to have as a top level global unless there is no other way to conveniently achieve the same effect.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
polina-c: Moved state that needs to be shared between inspector, debugger, and console, from DebuggerController to AppState owned by ConnectedApp.

packages/devtools_app/lib/src/screens/memory/memory_screen.dart
polina-c: Added for both debugger and  memory.

packages/.vscode/launch.json
polina-c: In this PR `settings.json` updates working directory to devtools_app, to make console experience the same as in devtools_app.

packages/devtools_app/lib/src/screens/memory/memory_screen.dart
jacob314: Not needed for debugger as IDEs never show the debugger screen as they have their own better more integrated debuggers.

packages/README.md
```diff
@@ -0,0 +1,10 @@
+# Coding agreements in DevTools
+
+## ValueListeneble and ValueNotifier
+
+When an object owns and exposes a listenable value,
+we declare the related fields always in the same order:
+
+1. Public getter for ValueListeneble
+2. Private ValueNotifier
+3. Public setter
```
jacob314, line 10: Do we really surface a public setter? I thought instances of this patter were a public getter with no setter.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -42,29 +43,32 @@ class DebuggerController extends DisposableController
...
 
-  final codeViewController = CodeViewController();
+  late final evalService = EvalService();
 
-  late final EvalService evalService;
+  final codeViewController = CodeViewController();
 
   bool _firstDebuggerScreenLoaded = false;
 
+  AppState get appState => serviceManager.appState;
```
jacob314, line 63: Why try to pretend that something from the serviceManager is part of the controller? I would just reference the serviceManager directly where this is used to make it clear that the DebuggerController is not needed for things it is not needed for. That will make it easier to detect the cases where we are now passing around the DebuggerController but don't need to.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -42,29 +43,32 @@ class DebuggerController extends DisposableController
...
-      isolateRef: isolateRef,
-      variables: variables,
-      frameForEval: () =>
-          _selectedStackFrame.value?.frame ??
-          _stackFramesWithLocation.value.safeFirst?.frame,
-      isPaused: isPaused,
+    addAutoDisposeListener(_selectedStackFrame, () => _updateCurrentFrame());
+    addAutoDisposeListener(
+      _stackFramesWithLocation,
+      () => _updateCurrentFrame(),
```
jacob314, line 49: nit: why is this
`() => _updateCurrentFrame()`
instead of
`_updateCurrentFrame`
?

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -167,21 +163,18 @@ class DebuggerController extends DisposableController
...
-    final id = isolateRef.value?.id;
+    final id = appState.isolateRef.value?.id;
     if (id == null) return '';
     return id;
   }
 
   void _switchToIsolate(IsolateRef? ref) async {
-    _isolateRef.value = ref;
-    _isPaused.value = false;
+    appState.setIsolateRef(ref);
```
jacob314, line 176: Dart style says to avoid methods that start with the words get and set.
I would make this an actual setter on the appState object instead.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -386,7 +379,7 @@ class DebuggerController extends DisposableController
     // serviceManager.isolateManager.selectedIsolateState.isPaused.value;
     // listening for changes there instead of having separate logic.
     await _getStackOperation?.cancel();
-    _isPaused.value = paused;
+    appState.setPaused(paused);
```
jacob314, line 382: appState would drive updating whether it is paused itself rather than depending on the debuggerController to drive it.

packages/devtools_app/lib/src/screens/debugger/debugger_screen.dart
```diff
@@ -48,6 +48,7 @@ class DebuggerScreen extends Screen {
 
   static final id = ScreenMetaData.debugger.id;
 
+  // TODO(polina-c): when embedded and console features should be in native console in VSCode
```
jacob314, line 51: not needed as the debugger should never be embedded in vscode.

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -47,14 +45,12 @@ class _InspectorTreeRowWidget extends StatefulWidget {
...
     required this.scrollControllerX,
     required this.viewportWidth,
-    required this.debuggerController,
   }) : super(key: key);
 
   final _InspectorTreeState inspectorTreeState;
 
   InspectorTreeNode get node => row.node;
   final InspectorTreeRow row;
-  final DebuggerController debuggerController;
```
jacob314, line 57: nice to see all this cleaned up.

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -93,6 +93,7 @@ class ServiceConnectionManager {
   late final ServiceExtensionManager _serviceExtensionManager;
 
   ConnectedApp? connectedApp;
+  AppState appState = AppState();
```
jacob314, line 96: Seems odd to have an appState object before there is actually a connected app. I would keep this consistent with other data in the class that is null unless there is actually an app.

packages/devtools_app/lib/src/shared/console/widgets/evaluate.dart
```diff
@@ -48,6 +47,8 @@ class ExpressionEvalFieldState extends State<ExpressionEvalField>
 
   final evalTextFieldKey = GlobalKey(debugLabel: 'evalTextFieldKey');
 
+  final _evalService = EvalService();
```
jacob314, line 50: Seems like there needs to be a global instances of the EvalService rather than creating one with each EvalField. Creating one with each eval field would cause problems later when you start adding extra evalService functionality for injecting `$1, $_, etc`

packages/devtools_app/lib/src/shared/console/widgets/evaluate.dart
```diff
@@ -310,14 +311,13 @@ class ExpressionEvalFieldState extends State<ExpressionEvalField>
     serviceManager.consoleService.appendStdio('> $expressionText\n');
     setState(() {
       historyPosition = -1;
-      widget.evalService.evalHistory.pushEvalHistory(expressionText);
+      _evalService.appState.evalHistory.pushEvalHistory(expressionText);
```
jacob314, line 314: I would avoid giving the illusion that appState is connected to _evalService when it isn't and just reference the global service directly.

packages/README.md
polina-c: hm. I cannot find example in the code, but it is what I remember Kenzie said to me some time ago.
@kenzieschmoll , did I recollect it right?

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
polina-c: Just `get': https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

Actual setter causes lint complain that wants the setter to be of the same type as getter, that is ValueListenable, while setter takes IsolateRef?

packages/devtools_app/lib/src/shared/connected_app.dart
```diff
@@ -210,3 +213,60 @@ class OfflineConnectedApp extends ConnectedApp {
...
+    classes.clear();
+    libraryMemberAndImportsAutocomplete.clear();
+    libraryMemberAutocomplete.clear();
+  }
+}
+
+class AppState {
+  ValueListenable<IsolateRef?> get isolateRef => _isolateRef;
+  final _isolateRef = ValueNotifier<IsolateRef?>(null);
+  void setIsolateRef(IsolateRef? value) {
```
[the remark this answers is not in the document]
polina-c: May be. It is not needed for functionality I need to build and it should be separate PR. Added TODO.

packages/devtools_app/lib/src/screens/debugger/console.dart
```diff
@@ -1,82 +0,0 @@
...
-}
-
-class _DebuggerConsoleState extends State<DebuggerConsole> {
-  ValueListenable<List<ConsoleLine>> get stdio =>
-      serviceManager.consoleService.stdio;
-
-  @override
-  void initState() {
-    super.initState();
-    serviceManager.consoleService.ensureServiceInitialized();
```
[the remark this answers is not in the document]
polina-c: it is now in [console_service.dart](https://github.com/flutter/devtools/pull/4993/files#diff-e8b87092138c1836ec4edb60a7bff230a98fd98525e697a6deca7c1426543faa), instead of assertion.

packages/devtools_app/lib/src/shared/connected_app.dart
```diff
@@ -210,3 +213,60 @@ class OfflineConnectedApp extends ConnectedApp {
...
+    if (value == isolateRef.value) return;
+    _isolateRef.value = value;
+    cache._clear();
+  }
+
+  ValueListenable<List<DartObjectNode>> get variables => _variables;
+  final _variables = ValueNotifier<List<DartObjectNode>>([]);
+  void setVariables(List<DartObjectNode> value) => _variables.value = value;
```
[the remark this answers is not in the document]
polina-c: I feel the same. Added todo to to explain it.

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
polina-c: how appState would know about it?

packages/devtools_app/lib/src/shared/connected_app.dart
```diff
@@ -248,13 +250,16 @@ class AppState {
...
   ValueListenable<List<DartObjectNode>> get variables => _variables;
   final _variables = ValueNotifier<List<DartObjectNode>>([]);
   void setVariables(List<DartObjectNode> value) => _variables.value = value;
 
   ValueListenable<bool> get isPaused => _isPaused;
   final _isPaused = ValueNotifier<bool>(false);
-  void setPaused(bool value) => _isPaused.value = value;
+
+  /// This setter should be invoked only by debugger.
+  void setPausedOnBreakpoint(bool value) => _isPaused.value = value;
```
jacob314, line 262: Fyi @kenzieschmoll what pattern should devtools standardize on for this case? Generally it is violating the Dart style guide to have members that start with the word set but I do see the benefit of this approach relative to having the getter be type ValueNotifier.

packages/devtools_app/lib/src/shared/console/console_service.dart
```diff
@@ -164,11 +164,7 @@ class ConsoleService extends Disposer {
   ///
   /// Note that this output might be truncated after significant output.
   ValueListenable<List<ConsoleLine>> get stdio {
-    assert(
-      _serviceInitialized,
-      '`ConsoleService.ensureServiceInitialized` must be called before '
-      'interacting with the ConsoleService.',
-    );
+    ensureServiceInitialized();
```
jacob314, line 167: This is probably ok but not ideal. It will mean the the console is guaranteed to show up empty the first frame it is displayed but does ensure we don't subscribe unless we need to. @kenzieschmoll any thoughts on this judgement call?

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
polina-c: From Jacob:

appState would know about it by moving the logic relevant to this into appState instead of debuggerController
there should be a very small number of events that are listened for to determine the paused state for isolates and appState should contain that logic rather than debugController intermediating it

For example, when you get an Isolate object, it tells you the current paused state
and when you get events that you paused at a breakpoint then you are paused

Fwiw, I wouldn't be surprised if debugger_controller has some bugs in how it is tracking the paused state currently as it seems to be doing some slightly unsafe things modifying the paused state based on UI actions rather than notifications from the VM Service

polina-c: Added TODO.

packages/devtools_app/lib/src/shared/connected_app.dart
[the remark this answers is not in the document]
CoderDake: I think for now this should be named ConsolePaneState. 
I think we should add a dart doc to the class mentioning that this is just used to store the state for the consolePane.

packages/devtools_app/lib/src/shared/connected_app.dart
kenzieschmoll: I cannot find the PR, but several months ago we had this same discussion. The Dart style guide only says to avoid starting methods with "get" but does not say to do so for "set". I remember this because I thought the same thing as you, to avoid "get" and "set", but couldn't find the style guide evidence to support that. Can you find a reference to where we are getting the idea to avoid "set" as well?

packages/README.md
```diff
@@ -0,0 +1,11 @@
+# Coding agreements in DevTools
```
kenzieschmoll, line 1: We should link to Effective Dart and the Flutter style guide here:
e.g. ordering guide described here: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#order-other-class-members-in-a-way-that-makes-sense

packages/devtools_app/lib/src/shared/connected_app.dart
polina-c: maybe from other languages: https://dev.to/scottshipp/avoid-getters-and-setters-whenever-possible-c8m

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -94,6 +94,9 @@ class ServiceConnectionManager {
 
   ConnectedApp? connectedApp;
 
+  AppState get appState => _appState!;
+  AppState? _appState;
```
kenzieschmoll, line 98: instead of making this it's own member on serviceManager, can we move it into ConnectedApp? That way it is clear that this state is for the connected app and not the DevTools app. This would read nicer and tightly couple the AppState instance with the ConnectedApp instance.

packages/devtools_app/lib/src/shared/connected_app.dart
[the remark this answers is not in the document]
polina-c: This class is intended to represent application state

isPaused and isolateRef is application state, not console pane state

The values are intended to be used by other features and we will do set of refactoring that will streamline the code towards this vision.

packages/devtools_app/lib/src/service/service_manager.dart
polina-c: Yes, I started with it. The problem is that debugger sometimes uses the values of AppState when the ConnectedApp is null.

Most likely it is bug of our features and/or test mocking, but it will be significant effort to fix.
Added TODO.

PR #5004, opened 2023-01-04
Make serviceManager.isolateManager source of truth for current isolate.

polina-c:

### DIFF
--- a/packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
@@ -82,7 +82,6 @@ class DebuggerController extends DisposableController
 
     final appState = serviceManager.appState;
 
-    appState.setIsolateRef(null);
     appState.setPausedOnBreakpoint(false);
     _resuming.value = false;
     _lastEvent = null;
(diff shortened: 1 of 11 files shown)


packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -170,9 +170,7 @@ class DebuggerController extends DisposableController
     // and modify to detect if app is paused from the isolate
     // https://github.com/flutter/devtools/pull/4993#discussion_r1060845351
 
-    serviceManager.appState
-      ..setIsolateRef(ref)
-      ..setPausedOnBreakpoint(false);
+    serviceManager.appState..setPausedOnBreakpoint(false);
```
kenzieschmoll, line 173: remove the .. now that there is only one method call

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -540,7 +538,7 @@ class DebuggerController extends DisposableController
         .map(
           (v) => DartObjectNode.create(
             v,
-            serviceManager.appState.isolateRef.value,
+            serviceManager.isolateManager.selectedIsolate.value,
```
kenzieschmoll, line 541: optional - we could add a convenience getter to service manager:
`IsolateRef? get selectedIsolate => isolateManager.selectedIsolate.value;` and use that everywhere for brevity

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
jacob314: Filed https://github.com/dart-lang/linter/issues/3960 to consider adding a lint to handle to warn about this case.

jacob314 (approved):
lgtm

PR #5010, opened 2023-01-06
Update RetainingPathWidget to use VmServiceObjectLinks for VM service objects

bkonyi:
These links can be used to jump to other VM service objects in the VM Tools Objects tab.

image

RELEASE_NOTE_EXCEPTION=VM Developer Tooling


packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -467,35 +472,89 @@ class RetainingPathWidget extends StatelessWidget {
 
   /// Describes the given RetainingObject [object] and its parentListIndex,
   /// parentMapKey, and parentField where applicable.
-  String _retainingObjectDescription(RetainingObject object) {
+  Widget _retainingObjectDescription(
```
kenzieschmoll, line 475: make this a private helper widget so that we don't have to pass context around

kenzieschmoll (approved):

PR #5020, opened 2023-01-09
Move next release notes.

polina-c:
Make next release notes accessible faster from dev env.


packages/devtools_app/NEXT_RELEASE_NOTES.md
```diff
@@ -6,9 +6,7 @@ This is draft for future release notes, that are going to land on
 Dart & Flutter DevTools - A Suite of Performance Tools for Dart and Flutter
 
 ## General updates
-
-- Fix several issues in syntax highlighting that would color variable names that contain reserved words incorrectly and leave `extends`/`implements` clauses uncolored for some classes [#4948](https://github.com/flutter/devtools/pull/4948)
```
CoderDake, line 10: We need to keep these until we do the next release

CoderDake (commented):
we will also need to update the location in `tool/update_version.dart` so that our version update code will reset md file when we do a minor or major version update :)

packages/devtools_app/NEXT_RELEASE_NOTES.md
polina-c: Did not mean to change this. It seems to be result of merge.

CoderDake (approved):

PR #5024, opened 2023-01-10
Add sample data dropdown to landing screen for use in integration tests

kenzieschmoll:
This is flagged to ensure this sample data picker never shows up in release mode. This allows for loading sample data into DevTools without the data existing in a local file system. 

RELEASE_NOTE_EXCEPTION=testing change


packages/devtools_app/test/shared/landing_screen_test.dart
```diff
@@ -20,8 +18,38 @@ void main() {
...
-    expect(find.text('Connect to a Running App'), findsOneWidget);
-    expect(find.text('App Size Tooling'), findsOneWidget);
+    await tester.pumpWidget(wrap(const LandingScreenBody()));
+    expect(find.byType(ConnectDialog), findsOneWidget);
+    expect(find.byType(ImportFileInstructions), findsOneWidget);
+    expect(find.byType(SampleDataDropDownButton), findsNothing);
+    expect(find.byType(AppSizeToolingInstructions), findsOneWidget);
+  });
+
+  testWidgetsWithWindowSize(
```
CoderDake, line 28: Could we add a test to make sure this doesn't show in release mode?

kenzieschmoll: There isn't an easy way that I'm aware of to mock the value of kReleaseMode. The only way to test this would be to add another job to our github actions that runs the test suite in release mode, which may be a bit overkill for a test this small.

CoderDake (approved):

PR #5028, opened 2023-01-11
Preliminary support for record types when vm_service <10.0.0

elliette:
From what I can tell, we don't yet have access to the field names for records when `vm_service` is <10.0.0. This adds preliminary support for records by simply displaying the values without names. 

Inspecting the following shows:

`var record = (1, a: "Hello there", 3, b: 4, true);`

Screenshot 2023-01-10 at 4 24 43 PM

RELEASE_NOTE_EXCEPTION='Record types not fully supported yet, will update release notes once they are'

### DIFF
--- a/packages/devtools_app/lib/src/shared/object_tree.dart
+++ b/packages/devtools_app/lib/src/shared/object_tree.dart
@@ -783,15 +783,24 @@ List<DartObjectNode> _createVariablesForFields(
 }) {
   final variables = <DartObjectNode>[];
   for (var field in instance.fields!) {
-    final name = field.decl!.name;
-    if (existingNames != null && existingNames.contains(name)) continue;
-    variables.add(
-      DartObjectNode.fromValue(
-        name: name,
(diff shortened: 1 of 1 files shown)


bkonyi:
Good timing! `vm_service` `10.0.0` was just published yesterday 🥳

bkonyi (approved):
LGTM for now, since I think it'll take Derek some time to roll `vm_service` through the DevTools deps.

kenzieschmoll:
Is the record unordered? 
`var record = (1, a: "Hello there", 3, b: 4, true);` doesn't appear to match the order in your screenshot

elliette:
Follow up on ordering (from @derekxu16):

> With vm_service < 10.0.0 clients can't get the field names, so there's no way to determine the order. With the field names you can tell which are positional, because the names will be ints instead of Strings. Under the hood, the service returns all the positional fields in order first, and then all the named fields.

PR #5033, opened 2023-01-12
Move root library detection to IsolateState.

polina-c:

### DIFF
--- a/packages/devtools_app/lib/src/service/isolate_state.dart
+++ b/packages/devtools_app/lib/src/service/isolate_state.dart
@@ -69,11 +69,11 @@ class RootInfo {
 
   final String? library;
   final String? package;
-}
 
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/service/isolate_state.dart
```diff
@@ -55,3 +63,17 @@ class IsolateState {
...
 }
+
+class RootInfo {
+  RootInfo(this.library) : package = _libraryToPackage(library);
+
+  final String? library;
+  final String? package;
+}
+
+String? _libraryToPackage(String? library) {
```
bkonyi, line 74: If this is only used in `RootInfo`, it should be a member of that class.

bkonyi (approved):

PR #5034, opened 2023-01-12
Update package:vm_service dependency to ^10.0.0

derekxu16:


derekxu16:
Do I have to add a changelog entry? If so, should the entry be for 2.20.1 or 2.21.0?

kenzieschmoll (approved):
no need to update the changelog - we generate the changelog entries before the release.

derekxu16:
I opened this because I thought that `package:devtools_shared` had to be published before `package:dds` because of a version solving problem. I found a way to publish `package:dds`, so I don't think this change is needed.

Issue #4207, opened 2022-06-21
Replace the DevTools trace viewer with the Perfetto UI


kenzieschmoll:
This will solve many issues that users experience with our trace viewer:
- https://github.com/flutter/devtools/issues/4205
- https://github.com/flutter/devtools/issues/3877
- https://github.com/flutter/devtools/issues/3800
- https://github.com/flutter/devtools/issues/3568
- https://github.com/flutter/devtools/issues/3454
- https://github.com/flutter/devtools/issues/2793
- https://github.com/flutter/devtools/issues/2609
- https://github.com/flutter/devtools/issues/2332
- https://github.com/flutter/devtools/issues/2222
- https://github.com/flutter/devtools/issues/2210
- https://github.com/flutter/devtools/issues/393
- https://github.com/flutter/devtools/issues/329

For more details, see go/embedded-perfetto (g3).

TODOs before shipping:

- [x] add test coverage for Perfetto trace processor and Perfetto iFrame view
- [x] perform team bug bash
- [x] test offline cases
- [ ] add analytics where necessary

Release blockers from bug bash:
- [x] https://github.com/flutter/devtools/issues/5122
- [x] https://github.com/flutter/devtools/issues/5120

kenzieschmoll:
This feature has been flipped to `beta`. On target to release to the Dart SDK within 1 week.

kenzieschmoll:
TODO after turning on and releasing: clean up feature flag.

PR #5051, opened 2023-01-13
Add Perfetto integration test with screenshots

kenzieschmoll:
This PR adds screenshot testing functionality to the integration testing framework in DevTools, and adds an integration test for the Perfetto trace viewer as its first use.

The goldens are tested with a diff tolerance to account for some differences in how things are rendered on the CI vs locally. Long term, this isn't a great solution, and we should look into a way to do server-side testing of these images or at least to generate and test the goldens in a consistent environment using something like Docker: https://github.com/flutter/devtools/issues/5068.

The integration_test/README.md file has been updated with instructions.

Work towards https://github.com/flutter/devtools/issues/4207

RELEASE_NOTE_EXCEPTION=[test changes only]


CoderDake (approved):

CoderDake (approved):
LGTM

I wonder if we should keep an eye on the large performance file. First thing I wonder is if we could make a smaller version of it if it starts causing performance issues for us in the future.

kenzieschmoll:
> First thing I wonder is if we could make a smaller version of it if it starts causing performance issues for us in the future.

reduced the size of the performance test data file.

CoderDake (approved):

CoderDake (approved):

PR #5053, opened 2023-01-14
Store an instance and its references to console

polina-c:
RELEASE_NOTE_EXCEPTION=[under the flag]


packages/devtools_app/lib/src/screens/memory/shared/primitives/instance_set_view.dart
```diff
@@ -2,17 +2,23 @@
...
+
 import 'package:flutter/material.dart';
-import 'package:vm_service/vm_service.dart';
 
 import '../../../../shared/analytics/constants.dart';
 import '../../../../shared/common_widgets.dart';
 import '../../../../shared/primitives/utils.dart';
 
-typedef SampleObtainer = InstanceRef Function();
+abstract class ClassSampler {
```
kenzieschmoll, line 13: why do we need an abstract class for this? will there be another class that extends ClassSampler at some point?

packages/devtools_app/lib/src/shared/console/console_service.dart
```diff
@@ -68,9 +78,35 @@ class VariableConsoleLine extends ConsoleLine {
   }
 }
 
+class GraphConsoleLine extends ConsoleLine {
+  GraphConsoleLine(this.graph, {bool forceScrollIntoView = false})
+      : super._(
+          forceScrollIntoView,
```
kenzieschmoll, line 84: no need for trailing comma if the values fit onto one line, which this would

packages/devtools_app/lib/src/screens/memory/shared/primitives/instance_set_view.dart
```diff
@@ -38,41 +44,47 @@ class InstanceSetView extends StatelessWidget {
...
-      const _MenuForSubset('Store first 100 instances as a console variable'),
-      const _MenuForSubset('Store all instances as a console variable'),
+List<Widget> _menu(ClassSampler sampleObtainer) => [
+      _StoreAsVariableMenu(
+        'Store as a console variable',
+        sampleObtainer,
+      ),
+      MenuItemButton(
+        onPressed: sampleObtainer.instanceGraphToConsole,
+        child: const Text('Browse an instance references in console'),
```
kenzieschmoll, line 88: This wording could use a little work. Maybe "Browser references for a single instance"

packages/devtools_app/lib/src/screens/memory/shared/primitives/instance_set_view.dart
polina-c: Yes, table that diffs snapshots, will need another implementation.

packages/devtools_app/lib/src/screens/memory/shared/primitives/instance_set_view.dart
polina-c: Made it 'Browse references for a single instance in console'

I want the word 'console', because without it user will not pay attention to console and will miss what is going on.

kenzieschmoll (approved):

Issue #2860, opened 2021-04-05
Not showing response tab under network 


mohitsharma3829:
Dev tool not showing response tab

kenzieschmoll:
What version of DevTools are you on?
@bkonyi will this be fixed with the 2.1.0 release?

bkonyi:
No, this the same issue we've been having. The full fix should be available
once the next Flutter beta goes out.

SeasonLeee:
@kenzieschmoll @bkonyi hi... I know the response tab is buggy since the advent of the dart Dev Tool... but right now it seems the whole network intercepting is not working since the publish of the 2.1.0 version...

![image](https://user-images.githubusercontent.com/13610283/113758498-89c7a700-9746-11eb-9d60-1d5b4a7d0147.png)

I swear to God I have made network request(otherwise my app won't work....)

And we have at least the requesting networking showing in the past version... what's wrong? should I update my Flutter?

<details>
<summary>Here is the flutter doctor -v</summary>

```
[✓] Flutter (Channel stable, 2.0.3, on macOS 11.2.3 20D91 darwin-x64, locale en)
    • Flutter version 2.0.3 at /Users/benjaminli/software_dev/sdks/flutter
    • Framework revision 4d7946a68d (3 weeks ago), 2021-03-18 17:24:33 -0700
    • Engine revision 3459eb2436
    • Dart version 2.12.2

[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.2)
    • Android SDK at /Users/benjaminli/Library/Android/Sdk
    • Platform android-30, build-tools 30.0.2
    • ANDROID_HOME = /Users/benjaminli/Library/Android/Sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6915495)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 12.4, Build version 12D4e
    • CocoaPods version 1.10.1

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 4.1)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
[...]

PR #5056, opened 2023-01-17
Improve Network Screen Reliability

CoderDake:
# Constraints
This PR depends on https://github.com/dart-lang/sdk/issues/51035 making it's way into sdk dart-lank/sdk main branch.

# Description

Related to https://github.com/flutter/devtools/issues/2860

This PR Makes the following improvements to `Network Screen` in Devtools.
- Fixes duplicate requests showing up in the request list
  - The `CurrentNetworkRequests` class is used to store requests in a map, and changes the updating procedures to be a bit less intense
- Improves performance
   - Instead of fetching  full response data for all requests, immediately, only request the data when the request is being inspected.
   - Responsive Loading UI has been added to the Response/Request tabs to let the user know that the body is loading.
- Fixes Response/Request tabs sometimes being empty
   - Fetching body response data is more reliable now, and  it's loading is communicated clearly to the user through spinners on the Response/Request tabs
- In the Response Tab, JSON responses now automatically expand the first level of their data automatically
- Websocket requests are now limited to those that have occurred since the recording started, not since the app started.


packages/devtools_app/test/test_infra/test_data/network.dart
```diff
@@ -88,7 +88,7 @@ final Map<String, dynamic> testSocket1Json = {
 
 final testSocket2 = WebSocket(SocketStatistic.parse(testSocket2Json)!, 0);
 final Map<String, dynamic> testSocket2Json = {
-  'id': 1,
+  'id': '11111',
```
CoderDake, line 91: To ensure the websocket Id's don't override httprequest ids, I've given then different ids :)

packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -352,3 +323,52 @@ class NetworkController extends DisposableController
     }
   }
 }
+
+class CurrentNetworkRequests {
```
bkonyi, line 327: Add some documentation here explaining what this class is for.

packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -352,3 +323,52 @@ class NetworkController extends DisposableController
...
+  CurrentNetworkRequests({required this.onDataUpdate});
+
+  Map<String, NetworkRequest> requestsMapping = <String, NetworkRequest>{};
```
bkonyi, line 330: This can be:

```dart
final requestsMapping = <String, NetworkRequest>{};
```

packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -352,3 +323,52 @@ class NetworkController extends DisposableController
...
+  VoidCallback onDataUpdate; // TODO: Callback naming rules
```
bkonyi, line 331: Can we remove this TODO?

packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
```diff
@@ -163,28 +174,39 @@ class HttpResponseView extends StatelessWidget {
...
-    }
-    return Padding(
-      padding: const EdgeInsets.all(denseSpacing),
-      child: SingleChildScrollView(child: child),
+    return ValueListenableBuilder(
+      valueListenable: data.updateCount,
+      builder: (_, __, ___) {
+        Widget child;
+        final theme = Theme.of(context);
+        // We shouldn't try and display an image response view when using the
```
bkonyi, line 182: This comment probably isn't relevant anymore. @kenzieschmoll, have we dropped support for scraping HTTP events from the timeline?

packages/devtools_app/lib/src/shared/http/http_request_data.dart
```diff
@@ -55,14 +56,33 @@ class DartIOHttpRequestData extends NetworkRequest {
...
-  Future<void> getFullRequestData() {
-    return serviceManager.service!
-        .getHttpProfileRequest(
-          _request.isolateId,
-          _request.id.toString(),
-        )
-        .then((updated) => _request = updated);
+  /// A notifier that changes when the request data, or it's response body
+  /// changes.
+  ValueNotifier<int> updateCount = ValueNotifier<int>(0);
```
bkonyi, line 63: Should this be modifiable outside of this class? If not, you'll want to make this private and expose a public `ValueListenable<int> get updateCount => _updateCount` instead.

packages/devtools_app/lib/src/shared/http/http_request_data.dart
```diff
@@ -55,14 +56,33 @@ class DartIOHttpRequestData extends NetworkRequest {
...
+        _request.id.toString(),
+      );
+      _request = updated;
+      updateCount.value++;
+      final fullRequest = _request as HttpProfileRequest;
+      _responseBody = utf8.decode(fullRequest.responseBody!);
+      _requestBody = utf8.decode(fullRequest.requestBody!);
+      return;
+    } catch (_) {
+      rethrow;
```
bkonyi, line 82: I don't think we need this catch here. This should do the same thing:

```dart
try {
  // Do something
} finally {
  // Do something else.
}
```

packages/devtools_app/pubspec.yaml
```diff
@@ -188,4 +188,4 @@ dependency_overrides:
   devtools_shared:
     path: ../devtools_shared
   devtools_test:
-    path: ../devtools_test
+    path: ../devtools_test
```
bkonyi, line 191: We still need a newline here.

packages/devtools_app/lib/src/screens/network/network_controller.dart
kenzieschmoll, line 330: nit: rename to `requestsById` so it is clear from the name what the map is between

packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -352,3 +323,52 @@ class NetworkController extends DisposableController
...
+      final webSocket = WebSocket(socket, timelineMicrosOffset);
+
+      if (requestsMapping.containsKey(webSocket.id)) {
+        // If we override an entry that is not a Websocket then that means
+        // the ids of the requestMapping entries may collide with other types
+        // of requests.
+        assert(requestsMapping[webSocket.id] is WebSocket);
+      }
+
+      requestsMapping[webSocket.id] = webSocket;
```
kenzieschmoll, line 367: can you comment to explain why there is no concept of "merge" for web socket requests like we have for http requests

packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -110,37 +90,28 @@ class NetworkController extends DisposableController
...
-        selectedRequest.value = webSocket;
-      }
-      currentValues.add(webSocket);
+    currentRequests.updateWebSocketRequests(sockets, timelineMicrosOffset);
+
+    // If we have updated data for the selected web socket, we need to update
+    // the value.
+    final currentSelectedRequestId = selectedRequest.value?.id;
+    if (currentSelectedRequestId != null) {
+      selectedRequest.value ??=
```
kenzieschmoll, line 102: does this ??= operation ever do anything? We wouldn't be in this if block if `selectedRequest.value` was null, so I think this assignment is dead code

packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -110,37 +90,28 @@ class NetworkController extends DisposableController
...
       timelineMicrosOffset: timelineMicrosOffset,
       httpRequests: httpRequests!,
-      currentValues: currentValues,
-      outstandingRequestsMap: outstandingRequestsMap,
+      currentRequests: currentRequests,
     );
 
     return NetworkRequests(
-      requests: currentValues,
+      requests: currentRequests.requestsMapping.values.toList(),
```
kenzieschmoll, line 113: consider making a getter on CurrentNetworkRequests, `requests` so that the requests map can be private, allowing only the CurrentNetworkRequests class to make changes

packages/devtools_app/lib/src/screens/network/network_controller.dart
CoderDake: oh shoot yes, this snuck back in on me.

packages/devtools_app/lib/src/screens/network/network_request_inspector.dart
```diff
@@ -78,6 +78,18 @@ class NetworkRequestInspector extends StatelessWidget {
               _buildTab(tabName: NetworkRequestInspector._cookiesTabTitle),
           ],
         ];
+        final tabViews = [
+          if (data != null) ...[
```
kenzieschmoll, line 82: this will throw an exception when data is null because the length of `tabs` will not match `tabViews`. If you put this back where it was, this code won't be reached when data is null. Another option is to make this a function

`List<Widget> generateTabViews() => [...];` so that way we don't pay the cost of creating all these objects when they aren't needed when `data` is null. Same would apply to the `tabs` list

packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
```diff
@@ -135,23 +135,34 @@ class HttpRequestView extends StatelessWidget {
...
-      );
-    }
-    return Padding(
-      padding: const EdgeInsets.all(denseSpacing),
-      child: SingleChildScrollView(
-        child: child,
-      ),
+    return ValueListenableBuilder(
+      valueListenable: data.updateCount,
+      builder: (_, __, ___) {
```
kenzieschmoll, line 140: let `context` be named here so that any `.of(context)` calls use the context from the ValueListenableBuilder and not the build method

packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
```diff
@@ -135,23 +135,34 @@ class HttpRequestView extends StatelessWidget {
...
+        }
+        Widget child;
+        if (requestContentType.contains('json')) {
+          child = JsonViewer(encodedJson: data.requestBody!);
+        } else {
+          child = Text(
+            data.requestBody!,
+            style: theme.fixedFontStyle,
+          );
+        }
```
kenzieschmoll, line 158: nit: this could be shortened using a ternary

packages/devtools_app/lib/src/shared/http/http_request_data.dart
kenzieschmoll, line 63: why is this notifier necessary? is this just a hack to make sure the response view rebuilds when the response comes available?

packages/devtools_app/lib/src/shared/http/http_request_data.dart
```diff
@@ -55,14 +56,33 @@ class DartIOHttpRequestData extends NetworkRequest {
...
-        )
-        .then((updated) => _request = updated);
+  /// A notifier that changes when the request data, or it's response body
+  /// changes.
+  ValueNotifier<int> updateCount = ValueNotifier<int>(0);
+  bool isFetchingFullData = false;
+
+  Future<void> getFullRequestData() async {
+    try {
+      // TODO: test theis already fetching shortcircuit
```
kenzieschmoll, line 68: delete todo or fix spelling

packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
```diff
@@ -163,28 +174,39 @@ class HttpResponseView extends StatelessWidget {
...
-        responseBody,
-        style: theme.fixedFontStyle,
-      );
-    }
-    return Padding(
-      padding: const EdgeInsets.all(denseSpacing),
-      child: SingleChildScrollView(child: child),
+    return ValueListenableBuilder(
+      valueListenable: data.updateCount,
+      builder: (_, __, ___) {
```
kenzieschmoll, line 179: same comment here about letting context be named

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1818,10 +1818,19 @@ class _JsonViewerState extends State<JsonViewer>
     with ProvidedControllerMixin<DebuggerController, JsonViewer> {
   late Future<void> _initializeTree;
   late DartObjectNode variable;
+  Future<void> _buildAndExpand(
```
kenzieschmoll, line 1821: new line above this method

packages/devtools_app/lib/src/shared/http/http_request_data.dart
kenzieschmoll: maybe instead of using a count notifier where we don't actually use the count, we can use a bool value notifier `isFetching` or `isComplete` or something that we can change to notify the request view to rebuild. Unless I'm misunderstanding something and we actually expect this notifier to be notified more than once?

kenzieschmoll: Oh I see now it is used from merge too. Maybe rename to `requestUpdatedNotifier` to make it clear that we aren't actually counting anything.

packages/devtools_app/lib/src/screens/network/network_request_inspector.dart
kenzieschmoll: ooo or a better idea. Make `tabs` and `tabViews` late final so they are only generated when the variables are used

packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
```diff
@@ -135,23 +135,34 @@ class HttpRequestView extends StatelessWidget {
...
-      ),
+    return ValueListenableBuilder(
+      valueListenable: data.updateCount,
+      builder: (context, __, ___) {
+        final theme = Theme.of(context);
+        final requestHeaders = data.requestHeaders;
+        final requestContentType = requestHeaders?['content-type'] ?? '';
+        final isLoading = data.isFetchingFullData;
+        if (isLoading) {
+          return CenteredCircularProgressIndicator(
```
bkonyi, line 146: Does this not have a const constructor?

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1818,10 +1818,19 @@ class _JsonViewerState extends State<JsonViewer>
...
+    DartObjectNode variable,
+  ) async {
+    // Build the root node
+    await buildVariablesTree(variable);
+    // Build the contents of all children
```
bkonyi, line 1826: Does this only build the second level of children (i.e., not recursively down the tree)?

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1841,7 +1850,19 @@ class _JsonViewerState extends State<JsonViewer>
...
+  }
+
+  @override
+  void initState() {
+    super.initState();
+    _updateVariablesTree();
+  }
+
+  @override
+  void didUpdateWidget(covariant JsonViewer oldWidget) {
```
bkonyi, line 1863: Why is this covariant?

packages/devtools_app/lib/src/shared/http/http_request_data.dart
```diff
@@ -55,14 +55,33 @@ class DartIOHttpRequestData extends NetworkRequest {
...
-        .getHttpProfileRequest(
-          _request.isolateId,
-          _request.id.toString(),
-        )
-        .then((updated) => _request = updated);
+  final ValueNotifier<int> _updateCount = ValueNotifier<int>(0);
+
+  /// A notifier that changes when the request data, or it's response body
+  /// changes.
+  ValueListenable<int> get updateCount => _updateCount;
```
bkonyi, line 64: Do we even use the value of this count? If not, consider renaming this and returning a `ValueListenable<void>` to indicate the value isn't useful.

packages/devtools_app/lib/src/shared/http/http_request_data.dart
```diff
@@ -55,14 +55,33 @@ class DartIOHttpRequestData extends NetworkRequest {
...
+      final updated = await serviceManager.service!.getHttpProfileRequest(
+        _request.isolateId,
+        _request.id.toString(),
+      );
+      _request = updated;
+      _updateCount.value++;
+      final fullRequest = _request as HttpProfileRequest;
+      _responseBody = utf8.decode(fullRequest.responseBody!);
+      _requestBody = utf8.decode(fullRequest.requestBody!);
+      return;
```
bkonyi, line 81: Unnecessary return.

packages/devtools_app/lib/src/shared/http/http_request_data.dart
```diff
@@ -229,6 +248,7 @@ class DartIOHttpRequestData extends NetworkRequest {
 
   /// Merges the information from another [HttpRequestData] into this instance.
   void merge(DartIOHttpRequestData data) {
+    _updateCount.value++;
```
bkonyi, line 251: You might want to do this update after setting `_request` to make sure we don't trigger a rebuild until the data is updated.

packages/devtools_app/pubspec.yaml
```diff
@@ -9,8 +9,8 @@ version: 2.22.0-dev.4
 repository: https://github.com/flutter/devtools/tree/master/packages/devtools_app
 
 environment:
-  sdk: '>=2.17.0 <4.0.0'
-  flutter: '>=3.0.0'
+  sdk: ">=2.17.0 <4.0.0"
```
bkonyi, line 12: Single quotes?

packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -352,3 +323,60 @@ class NetworkController extends DisposableController
     }
   }
 }
+
+/// Class for managing the set of all current websocket requests, and
+/// http profile requests.
+class CurrentNetworkRequests {
+  CurrentNetworkRequests({required this.onDataUpdate});
```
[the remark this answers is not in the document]
CoderDake: ty for catching this.
I changed the place where the data fetching happens, and left this behind :)

packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
CoderDake: it won't allow it because mediumProgressSize isn't const, it is a getter :/

packages/devtools_app/lib/src/shared/common_widgets.dart
CoderDake: Looks like maybe the autocomplete picked it up from inside state.

packages/devtools_app/lib/src/shared/common_widgets.dart
CoderDake: correct

packages/devtools_app/pubspec.yaml
CoderDake: looks like autoformatting automatically updated this when I saved the file :/

packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -76,32 +79,10 @@ class NetworkController extends DisposableController
   void _processHttpProfileRequests({
     required int timelineMicrosOffset,
     required List<HttpProfileRequest> httpRequests,
```
kenzieschmoll, line 81: are the httpRequests the new requests here? if so maybe we can name newHttpRequests to make that clear and to contrast with currentRequests

packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -110,37 +91,28 @@ class NetworkController extends DisposableController
...
-      currentValues.add(webSocket);
+    currentRequests.updateWebSocketRequests(sockets, timelineMicrosOffset);
+
+    // If we have updated data for the selected web socket, we need to update
+    // the value.
+    final currentSelectedRequestId = selectedRequest.value?.id;
+    if (currentSelectedRequestId != null) {
+      selectedRequest.value =
+          currentRequests.getRequest(currentSelectedRequestId);
     }
```
kenzieschmoll, line 105: should this be done after we process the http profile requests? seems odd to be updating the selected value only after updating web socket requests

packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -247,12 +218,13 @@ class NetworkController extends DisposableController
...
     _requests.value = NetworkRequests();
+    _currentNetworkRequests.clear();
     resetFilter();
-    _updateData();
+    _handleRequestListChanges();
     _updateSelection();
   }
 
-  void _updateData() {
+  void _handleRequestListChanges() {
```
kenzieschmoll, line 227: maybe rename to _filterAndRefreshSearchMatches() to make it clear what this does?

packages/devtools_app/lib/src/screens/network/network_controller.dart
CoderDake: they are new or updated requests so I will go for `newOrUpdatedHttpRequests`

packages/devtools_app/lib/src/screens/network/network_controller.dart
CoderDake: we could, but since we only merge when http profile requests get updated, then the instance remains the same.
So it will still be pointing to the correct instance

CoderDake:
NOTE: this is blocked until we can get a flutter candidate with vm_service:11 on it

kenzieschmoll (approved):

bkonyi (approved):

bkonyi (approved):

PR #5059, opened 2023-01-18
Handle edge cases for eval and pass needed parameters to HeapObjectGraph.

polina-c:
RELEASE_NOTE_EXCEPTION=[changes under the flag]


packages/devtools_app/lib/src/screens/memory/panes/diff/controller/sampler.dart
```diff
@@ -11,42 +11,51 @@ import '../../../shared/primitives/class_name.dart';
...
     final instances = await serviceManager.service!.getInstances(
       isolateId,
       theClass.id!,
       1,
     );
 
-    return instances.instances!.first as InstanceRef;
+    final result = instances.instances!.first;
+
+    if (result is InstanceRef) return result;
```
bkonyi, line 38: In what scenario would this not be an `InstanceRef`? If it's only an error case, consider throwing an exception rather than having a nullable return type.

bkonyi (approved):

Issue #5013, opened 2023-01-06
Memory > Tracing loses context.


polina-c:
After clicking code link in order to jump to debugger loses selections and collected traces.

Kenzie:
those items will just need to be stored in the tracing controller and fed to the UI via a value listenable

PR #5062, opened 2023-01-18
Fix tracing state management.

polina-c:
Fixes https://github.com/flutter/devtools/issues/5013

Store tracing controller in memory controller, so that it is preserved after switching to other screen.


bkonyi:
What's being fixed here?

polina-c:
> What's being fixed here?

Added description

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
```diff
@@ -68,8 +69,9 @@ class MemoryController extends DisposableController
   /// The controller is late to enable test injection.
   late final DiffPaneController diffPaneController;
 
-  /// Controller for [AllocationProfileTableView].
-  final allocationProfileController = ProfilePaneController();
+  final profilePaneController = ProfilePaneController();
+
+  final tracingPaneController = TracingPaneController();
```
kenzieschmoll, line 74: now that we are storing this in the controller, does this need to be cleared or disposed at any point by the memory controller?

kenzieschmoll (approved):
one question to double check, then lgtm

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
polina-c: will follow up in separate PR

polina-c: https://github.com/flutter/devtools/pull/5071

Issue #5046, opened 2023-01-13
light mode, link on selected row is not visible


polina-c:
<img width="906" alt="Screenshot 2023-01-13 at 11 01 06 AM" src="https://user-images.githubusercontent.com/12115586/212405694-5cc97c2e-faa2-4494-b8e3-a721364d909d.png">

Suggestion for selected row:

<img width="782" alt="Screenshot 2023-01-13 at 11 01 45 AM" src="https://user-images.githubusercontent.com/12115586/212401092-22a5653c-32f8-4bc7-815a-7beb61c213ce.png">

polina-c:
@malloc-error

malloc-error:
I recommend using the bright blue background, but then check the white text for legibility. If that white text is too hard to read, then change it to black.

PR #5064, opened 2023-01-18
Update row selection to resolve conflict with link color.

polina-c:
Fixes https://github.com/flutter/devtools/issues/5046

Screenshot 2023-01-18 at 8 08 02 AM


polina-c:
@malloc-error

kenzieschmoll:
let's make the link color white for the selected row. There is not enough contrast, as is.

malloc-error:
I agree with Kenzie on the contrast issue. I think the most straightforward solution is as Kenzie said to make the link text white. The underline probably provides enough context that it's a link, and therefore a different category of text. 

If there's a very bright gray in the existing palette, you could try that for the link text color, to create a bit of visual distinction with the text on the left (the way the un-selected rows are with blue and black text). 

In any case, we could look into cases like this in the color audit later this quarter.

polina-c:
Implemented in other PR: https://github.com/flutter/devtools/pull/5067

PR #5071, opened 2023-01-19
Recreate tracing controller on reconnect.

polina-c:
RELEASE_NOTE_EXCEPTION=[already covered in https://github.com/flutter/devtools/pull/4614] 

Follow up for https://github.com/flutter/devtools/pull/5062#discussion_r1081324667


bkonyi (approved):

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
```diff
@@ -304,6 +305,9 @@ class MemoryController extends DisposableController
     _memoryTracker?.stop();
     _memoryTrackerController.add(_memoryTracker);
 
+    tracingPaneController.dispose();
+    tracingPaneController = TracingPaneController();
```
kenzieschmoll, line 309: why does this controller need to be recreated and the other two (profilePaneController and diffPaneController) do not?

polina-c: good question that triggered nice refactoring. Thank you!

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
```diff
@@ -40,6 +40,34 @@ class OfflineFileException implements Exception {
   String toString() => message;
 }
 
+class MemoryChildControllers {
```
kenzieschmoll, line 43: rename to MemoryFeatureControllers to match naming in other screens.

It would be worth investigating state management between the performance page and the memory page to see if we should use a common approach there. The Performance page and the Memory page have a similar setup where each tab contains a somewhat self contained feature. In the performance screen, we have a concept of "activeFeature" and each feature controller extends `PerformanceFeatureController`, which is an abstract class with lifecycle methods that each feature controller should implement (some examples `init`, `onBecomingActive`, `setOfflineData`, `clearData`, etc.). This setup is nice from an organization standpoint but also prevents us from doing business logic work in a controller that we don't need to be doing if a feature is not active. Not pressing to do in this PR, but just food for thought.

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
```diff
@@ -66,12 +94,8 @@ class MemoryController extends DisposableController
...
   }
 
-  /// The controller is late to enable test injection.
-  late final DiffPaneController diffPaneController;
-
-  final profilePaneController = ProfilePaneController();
-
-  final tracingPaneController = TracingPaneController();
+  /// Sub-controllers of memory controller.
+  late final MemoryChildControllers children;
```
kenzieschmoll, line 98: nit: rename to `featureControllers`

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
polina-c: thanks

Noted this for the time I will work to support offline mode for memory screen.

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
polina-c: featureControllers is too long. how about just `controllers' ?

PR #5074, opened 2023-01-19
Add GC statistics to the allocation profile in VM developer mode

bkonyi:
This change surfaces some details about GC performance, including:
  - Heap space usage
  - Heap space capacity
  - \# of collections
  - Average collection latency (i.e., time taken to perform a GC)

**Screenshot:**
image

RELEASE_NOTE_EXCEPTION=VM developer functionality


packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
```diff
@@ -204,6 +205,209 @@ class _FieldSizeColumn extends ColumnData<ProfileRecord> {
   bool get numeric => true;
 }
 
+abstract class _HeapStatsColumn extends ColumnData<AdaptedProfile> {
```
kenzieschmoll, line 208: should this be _GcStatsColumn?

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
```diff
@@ -204,6 +205,209 @@ class _FieldSizeColumn extends ColumnData<ProfileRecord> {
...
+        return profile.newSpaceGCStats;
+      case _HeapGeneration.oldSpace:
+        return profile.oldSpaceGCStats;
+      case _HeapGeneration.total:
+        return profile.totalGCStats;
+    }
+  }
+}
+
+class _HeapNameColumn extends ColumnData<AdaptedProfile> {
```
kenzieschmoll, line 231: _GCStatsNameColumn?

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
```diff
@@ -204,6 +205,209 @@ class _FieldSizeColumn extends ColumnData<ProfileRecord> {
...
+          fixedWidthPx: scaleByFontFactor(200),
+        );
+
+  @override
+  String? getValue(AdaptedProfile dataObject) {
+    return 'GC Statistics';
+  }
+}
+
+class _HeapUsageColumn extends _HeapStatsColumn {
```
kenzieschmoll, line 244: same with all of these _GCUsageColumn, _GCCapacityColumn, _GCLatencyColumn to name them more appropriately with the column titles

kenzieschmoll: hmm actually maybe I'm wrong on _HeapUsageColumn and _HeapCapacityColumn I don't think I understood the context. However, I still think _GCLatencyColumn is a more fitting name

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
```diff
@@ -204,6 +205,209 @@ class _FieldSizeColumn extends ColumnData<ProfileRecord> {
...
+          fixedWidthPx: scaleByFontFactor(_defaultNumberFieldWidth),
+        );
+
+  @override
+  double? getValue(AdaptedProfile dataObject) {
+    final stats = getGCStats(dataObject);
+    if (stats.averageCollectionTime.isNaN) {
+      return 0;
+    }
+    return getGCStats(dataObject).averageCollectionTime * 1000;
```
kenzieschmoll, line 325: we should use the msText helper here

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
```diff
@@ -204,6 +205,209 @@ class _FieldSizeColumn extends ColumnData<ProfileRecord> {
...
+    ),
+  ];
+
+  static final _columns = [
+    _HeapNameColumn(),
+    for (final generation in [
+      _HeapGeneration.total,
+      _HeapGeneration.newSpace,
+      _HeapGeneration.oldSpace,
+    ]) ...[
```
kenzieschmoll, line 373: loop through `_HeapGeneration.values` here instead

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
```diff
@@ -235,6 +439,27 @@ class AllocationProfileTableViewState
...
         ),
+        ValueListenableBuilder<bool>(
+          valueListenable: preferences.vmDeveloperModeEnabled,
+          builder: (context, vmDeveloperModeEnabled, _) {
+            if (!vmDeveloperModeEnabled) return const SizedBox.shrink();
+            return Column(
+              children: [
+                SizedBox(
+                  // _GCStatsTable is a table with two header rows (column groups
+                  // and columns) and one data row. We add a slight padding to
```
kenzieschmoll, line 450: can we use a Padding widget to add this padding so that it is more clear what the padding value is?

packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -83,6 +84,12 @@ class FlatTable<T> extends StatefulWidget {
 
   final List<ColumnGroup>? columnGroups;
 
+  /// Determines if the headers for column groups should be rendered.
+  ///
+  /// If set to false and `columnGroups` is non-null and non-empty, only
+  /// dividing lines will be drawn for each column group boundary.
```
kenzieschmoll, line 90: super nit: can you change to "only vertical dividing lines" to make it clear which divider you are referring to. When I first read this I thought we were talking about the horizontal divider lines between column group headers and column headers.

packages/devtools_app/test/memory/profile/allocation_profile_table_view_test.dart
```diff
@@ -90,6 +95,10 @@ void main() {
         expect(find.text('Total'), findsOneWidget);
         expect(find.text('Old Space'), findsOneWidget);
         expect(find.text('New Space'), findsOneWidget);
+        expect(find.text('Usage'), findsNWidgets(3));
+        expect(find.text('Capacity'), findsNWidgets(3));
+        expect(find.text('Collections'), findsNWidgets(3));
+        expect(find.text('Latency'), findsNWidgets(3));
```
kenzieschmoll, line 101: should you check that the values for these columns from your sample data are here as well?

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
bkonyi: Ordering is important here, which is why I decided to iterate through this list instead.

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
bkonyi: This isn't returning text and `averageCollectionTime` is in seconds.

bkonyi: We use `msText` below.

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
bkonyi: Maybe padding is the wrong word here, but that doesn't accomplish what this extra 1 pixel does. The child of `SizedBox` needs an extra pixel of space to disable the scrolling behavior of the underlying table. Padding will just add the extra pixel around the child widget without disabling the scroll behavior.

packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
```diff
@@ -5,6 +5,7 @@
 import 'package:vm_service/vm_service.dart';
 
 import '../../shared/globals.dart';
+import '../memory/panes/profile/profile_view.dart';
```
kenzieschmoll, line 8: is this import intentional?

kenzieschmoll (approved):

PR #5076, opened 2023-01-20
Fix the broken build

kenzieschmoll:

### DIFF
--- a/packages/devtools_test/lib/src/wrappers.dart
+++ b/packages/devtools_test/lib/src/wrappers.dart
@@ -136,7 +136,7 @@ void testWidgetsWithContext(
       for (Type type in oldValues.keys) {
         final oldGlobal = oldValues[type];
         if (oldGlobal != null) {
-          setGlobal(type, oldValues[type]);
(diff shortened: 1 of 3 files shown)


packages/devtools_test/lib/src/wrappers.dart
```diff
@@ -134,7 +134,12 @@ void testWidgetsWithContext(
     } finally {
       // restore previous global values
       for (Type type in oldValues.keys) {
-        setGlobal(type, oldValues[type]);
+        final oldGlobal = oldValues[type];
+        if (oldGlobal != null) {
```
jacob314, line 138: nit: using `oldValues.containsKey` would be safer in case it ever matters whether the oldValue was null rather than ommitted.

jacob314 (approved):
lgtm

packages/devtools_test/lib/src/wrappers.dart
kenzieschmoll: oldValues[type] will be explicitly set to null above on line 128, so oldValues will actually contain the key here

PR #5077, opened 2023-01-20
Modify build script to build DevTools from release mode.

kenzieschmoll:
This PR backfills the changes that were added to tag 2.20.1 for the hotfix release. Diff here: https://github.com/flutter/devtools/compare/v2.20.0...v2.20.1

We do not update the version here because the current version of DevTools is already beyond 2.20.1.

RELEASE_NOTE_EXCEPTION=[tool changes only]

### DIFF
--- a/tool/build_release.sh
+++ b/tool/build_release.sh
@@ -15,17 +15,27 @@ DEVTOOLS_DIR="${TOOL_DIR}/.."
 
 pushd $TOOL_DIR
 
-# Use the Flutter SDK from flutter-sdk/.
-FLUTTER_DIR="`pwd`/flutter-sdk"
-PATH="$FLUTTER_DIR/bin":$PATH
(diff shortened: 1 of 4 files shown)


tool/build_release.sh
```diff
@@ -15,18 +15,28 @@ DEVTOOLS_DIR="${TOOL_DIR}/.."
...
+  ./update_flutter_sdk.sh
+fi
 
 popd
 
 # echo on
 set -ex
 
+which flutter
+flutter --version
```
CoderDake, line 38: If these are meant to be here, then adding a bit of context to the messages could be helpful:
```suggestion
echo "Flutter Path: $(which flutter)"
echo "Flutter Version: $(flutter --version)"
```

CoderDake (approved):

PR #5084, opened 2023-01-21
Support record inspection

elliette:
Adds support for inspecting records:

Screenshot 2023-01-23 at 1 23 31 PM

Screenshot 2023-01-23 at 1 23 54 PM

Note: Displaying the length of a record before it is expanded requires https://github.com/dart-lang/sdk/issues/51104, I've added a TODO once it is resolved. I also noticed that type signatures are missing for `Map` and `List` types (eg, the `Map` in the inspected `Record`. I've opened up https://github.com/flutter/devtools/issues/5092 for that.


packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -275,11 +275,18 @@ class VmServiceWrapper implements VmService {
     String isolateId,
     String objectId,
     int limit, {
-    String? classId,
+    bool? includeSubclasses,
```
elliette, line 278: This is a breaking change in `vm_service` v.10.0.0

packages/devtools_app/pubspec.yaml
```diff
@@ -188,3 +188,6 @@ dependency_overrides:
     path: ../devtools_shared
   devtools_test:
     path: ../devtools_test
+  vm_service: ^10.0.0
```
elliette, line 191: Overrides should be removed before submitting (requires `vm_service` v.10.0.0 in flutter)

packages/devtools_test/pubspec.yaml
```diff
@@ -39,6 +39,7 @@ dependency_overrides:
     path: ../devtools_app #OVERRIDE_FOR_DEVELOPMENT
   devtools_shared: #OVERRIDE_FOR_DEVELOPMENT
     path: ../devtools_shared #OVERRIDE_FOR_DEVELOPMENT
+  vm_service: ^10.0.0
```
elliette, line 42: Overrides should be removed before submitting

packages/devtools_app/lib/src/shared/diagnostics/variable_factory.dart
```diff
@@ -478,6 +478,57 @@ List<DartObjectNode> createVariablesForElements(
...
+  }
+  for (final field in namedFields) {
+    variables.add(
+      DartObjectNode.fromValue(
+        name: field.name,
+        value: field.value,
+        isolateRef: isolateRef,
+      ),
+    );
+  }
```
kenzieschmoll, line 517: this can all be inlined in the list:
```
final variables = [
  // Always show positional fields before named fields:
  for (final field in positionalFields)
    DartObjectNode.fromValue(
        // Positional fields are designated by their getter syntax, eg $0, $1,
        // $2, etc:
        name: '\$${field.name}',
        value: field.value,
        isolateRef: isolateRef,
      ),
  ...
];
```

kenzieschmoll, line 517: this can all be inlined in the list:
```
final variables = [
  // Always show positional fields before named fields:
  for (final field in positionalFields)
    DartObjectNode.fromValue(
        // Positional fields are designated by their getter syntax, eg $0, $1,
        // $2, etc:
        name: '\$${field.name}',
        value: field.value,
        isolateRef: isolateRef,
      ),
  ...
];
```

elliette: Good point, done!

kenzieschmoll (approved):
LGTM - please update PR description before submitting so that changelog generation script will no show "do not submit"

packages/devtools_shared/pubspec.yaml
```diff
@@ -12,7 +12,7 @@ dependencies:
   path: ^1.8.0
   shelf: ^1.1.0
   usage: ^4.0.0
-  vm_service: ">=10.1.0 <12.0.0"
+  vm_service: ">=10.1.2 <12.0.0"
```
kenzieschmoll, line 15: can we write ^10.1.2 or do we need the upper bound constraint?

elliette: @bkonyi - the upperbound constraint is because 11.0.0 is going to be released soon, and it won't actually be a breaking change (if I understand correctly) so we want to allow both

Issue #4887, opened 2022-12-02
Double clicking refresh causes perfetto bug


CoderDake:
![Screen Shot 2022-12-02 at 2 48 45 PM](https://user-images.githubusercontent.com/1386322/205374412-e5611d3b-4dcb-421c-bfd5-a961af4c0c82.png)

```
To assist with debugging please attach or link to the trace you were viewing.

Viewed on: http://localhost:52684

Error: Value doesn't exist
Error: Value doesn't exist
    at assertExists (http://localhost:52684/assets/packages/perfetto_compiled/dist/v30.0-8d760d3b2/frontend_bundle.js:2178:15)
    at TraceController.<anonymous> (http://localhost:52684/assets/packages/perfetto_compiled/dist/v30.0-8d760d3b2/frontend_bundle.js:343391:56)
    at Generator.next (<anonymous>)
    at fulfilled (http://localhost:52684/assets/packages/perfetto_compiled/dist/v30.0-8d760d3b2/frontend_bundle.js:4534:62)

v30.0-8d760d3b2 8d760d3b2a25abefae6fd63be3a095ae5bfb8d59
UA: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36
```

# Details

When double clicking the timeline refresh button quickly, I noticed that an error comes up if you give the page a chance to load.
I tried reproducing this with single clicks but was not able to
I tried reproducing with slower double clicks and was not able to

PR #5093, opened 2023-01-24
Block refresh timeline button while load is already in progress

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/4887

RELEASE_NOTE_EXCEPTION=[fixes a bug that is behind a flag]

### DIFF
--- a/packages/devtools_app/lib/src/shared/primitives/utils.dart
+++ b/packages/devtools_app/lib/src/shared/primitives/utils.dart
@@ -217,7 +217,7 @@ void executeWithDelay(
   }
 }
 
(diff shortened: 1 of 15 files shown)


packages/devtools_app/lib/src/shared/primitives/utils.dart
```diff
@@ -217,7 +217,7 @@ void executeWithDelay(
   }
 }
 
-Future<void> delayForBatchProcessing({int micros = 0}) async {
+Future<void> delayToReleaseUiThread({int micros = 0}) async {
```
polina-c, line 220: good renaming

polina-c (approved):

PR #5098, opened 2023-01-24
Create initial structure for references.

polina-c:

https://user-images.githubusercontent.com/12115586/214422938-9fd47d47-4287-4f26-8d9a-7b88e604c241.mov


kenzieschmoll:
Do we want to disable these for Inspector entries in the console? I can't see a use case for looking at references of widgets from the inspector.

packages/devtools_app/lib/src/shared/diagnostics/generic_instance_reference.dart
```diff
@@ -36,3 +39,16 @@ class GenericInstanceRef {
...
+  members,
+  refRoot,
+
+  staticRefRoot,
+  staticInRefs,
+  staticOutRefs,
+
+  liveRefRoot,
+  liveInRefs,
+  liveOutRefs,
```
kenzieschmoll, line 53: can you add dart doc for each of these? ///

packages/devtools_app/lib/src/shared/diagnostics/tree_builder.dart
```diff
@@ -116,11 +116,82 @@ void _setupGrouping(DartObjectNode variable) {
...
+      break;
+    case ExpandType.staticOutRefs:
+      variable.addChild(
+        DartObjectNode.references(
+          ExpandType.staticOutRefs,
+          '<static outbound refs>',
+          variable.ref!,
+        ),
+      );
+      break;
```
kenzieschmoll, line 156: are these placeholders currently? if so can you add a comment?

packages/devtools_app/lib/src/shared/diagnostics/generic_instance_reference.dart
```diff
@@ -19,11 +19,14 @@ import 'diagnostics_node.dart';
 /// Sentinel, or primitive type.
 class GenericInstanceRef {
   GenericInstanceRef({
+    this.expandType = ExpandType.members,
```
kenzieschmoll, line 22: instead of having adding expandType to GenericInstanceRef, why not create a new class `ObjectReferences extends GenericInstanceRef`? then below, instead of checking `expandType != ExpandType.members` (which feels a bit brittle), you can check the type: `variable.ref == ObjectReferences`.

packages/devtools_app/lib/src/shared/diagnostics/generic_instance_reference.dart
```diff
@@ -36,3 +39,16 @@ class GenericInstanceRef {
 
   final IsolateRef? isolateRef;
 }
+
+enum ExpandType {
```
kenzieschmoll, line 43: I think we can name this a bit better. Perhaps, ObjectReferencesChildType or similar?

polina-c: RefNodeType ?

kenzieschmoll:
> It is just one gray line, so it will not impact UX.
> 
> But:
> 
> it will make users aware that this feature exist, i.e. will increase discoverability of the feature
> Will avoid confusion about why sometimes references are here and sometimes are not.
> How does it sound?

Will it add extra overhead to fetch the references for each widget inspector selection? It is not uncommon for a user to select many inspector widgets before finding the one they are looking for. Fetching references for each one seems like too much cost to justify the benefits (when in this case, the benefits aren't for the user journey of UI debugging, but rather for UI consistency and discoverability). 

> Will avoid confusion about why sometimes references are here and sometimes are not.

Don't the Inspector objects in the console look different than Memory objects that are dumped in the console? If so, I don't think it would be confusing that these objects come with different data.

packages/devtools_app/lib/src/shared/diagnostics/generic_instance_reference.dart
```diff
@@ -40,15 +37,35 @@ class GenericInstanceRef {
...
+  /// Subitem of [refRoot] for static references.
   staticRefRoot,
+
+  /// Subitem of [staticRefRoot] for inbound static references.
   staticInRefs,
+
+  /// Subitem of [staticRefRoot] for outbound static references.
   staticOutRefs,
 
+  /// Subitem of [refRoot] for kive references.
```
kenzieschmoll, line 63: sp: live

packages/devtools_app/lib/src/shared/diagnostics/tree_builder.dart
```diff
@@ -118,61 +118,63 @@ void _setupGrouping(DartObjectNode variable) {
 
 void addChildReferences(
   DartObjectNode variable,
-  ExpandType expandType,
+  RefNodeType expandType,
```
kenzieschmoll, line 121: rename variable refNodeType

polina-c:
> > It is just one gray line, so it will not impact UX.
> > But:
> > it will make users aware that this feature exist, i.e. will increase discoverability of the feature
> > Will avoid confusion about why sometimes references are here and sometimes are not.
> > How does it sound?
> 
> Will it add extra overhead to fetch the references for each widget inspector selection? It is not uncommon for a user to select many inspector widgets before finding the one they are looking for. Fetching references for each one seems like too much cost to justify the benefits (when in this case, the benefits aren't for the user journey of UI debugging, but rather for UI consistency and discoverability).
> 
> > Will avoid confusion about why sometimes references are here and sometimes are not.
> 
> Don't the Inspector objects in the console look different than Memory objects that are dumped in the console? If so, I don't think it would be confusing that these objects come with different data.

The same object can be dumped from inspector and memory and will look the same.
It takes number of clicks to take heap snapshot. For some applications heap snapshot can turn out to be expensive. So, the option to start exploration from inspector may be viable.

polina-c:
Reminder: remove temporary debug print before merge.

kenzieschmoll:
CC @jacob314 to weigh in on inspector node discussion if you have opinions here.

polina-c:
> Will it add extra overhead to fetch the references for each widget inspector selection? It is not uncommon for a user to select many inspector widgets before finding the one they are looking for. Fetching references for each one seems like too much cost to justify the benefits (when in this case, the benefits aren't for the user journey of UI debugging, but rather for UI consistency and discoverability).

The references are requested only when user expands the top item.

packages/devtools_app/lib/src/shared/diagnostics/tree_builder.dart
```diff
@@ -116,11 +117,84 @@ void _setupGrouping(DartObjectNode variable) {
   }
 }
 
+void addChildReferences(
+  DartObjectNode variable,
+  RefNodeType refNodeType,
+) {
+  final ref = variable.ref!;
```
kenzieschmoll, line 124: nit: just to be double careful here, we should `assert(FeatureFlags.evalAndBrowse)`

jacob314:
> 
We should not surface in-bound and out-bound references for objects unless they were added to the console by the memory page or a user has right clicked and opted into the advanced functionality for an object.
Motivations:
1. Inspector users and general debugger users will be confused by seeing a references item in the list of properties as they are not typically included in debugger views of an object.
2. in-bound and out-bound references will not be available for Flutter web so we need to hide them for that case.
3. It will be a little confusing if users happening to be debugging classes with field names that overlap with the name "references". This is fine if it is in the context of an action triggered from the memory page but confusing overall.
4. Concerned collecting the references might be slow so not something you want a user who has logic debugging performance expectations to accidentally trigger. 

On the other hand I could be convinced if there are cases where seeing the references is truly useful for widget debugging case.

kenzieschmoll:
I agree will all 4 of Jacob's points.

> On the other hand I could be convinced if there are cases where seeing the references is truly useful for widget debugging case.

I agree here as well, but it seems that this should be an opt-in use case if this use case exists, and likely a lower priority use case still. To my knowledge, users have not requested this feature.

polina-c:
I will hide the item for non-memory-originated items in following PRs, as this one does not contain needed flags.

packages/devtools_app/test/debugger/typed_data_variable_test.dart
```diff
@@ -8,6 +8,7 @@ import 'dart:typed_data';
 import 'package:devtools_app/src/service/service_manager.dart';
 import 'package:devtools_app/src/shared/diagnostics/dart_object_node.dart';
 import 'package:devtools_app/src/shared/diagnostics/tree_builder.dart';
+import 'package:devtools_app/src/shared/feature_flags.dart';
```
kenzieschmoll, line 11: why are we adding an import but not using it?

kenzieschmoll (approved):
one comment but lgtm

Issue #5023, opened 2023-01-10
Perfetto does not load correctly in the Dart SDK


bkonyi:
When opening DevTools for a Dart CLI application, the iframe for Perfetto displays various errors depending on what flags were provided to the VM.

**`dart --observe foo.dart`**: `missing or invalid authentication code`

**`dart --observe --disable-service-auth-codes foo.dart`**: `{"jsonrpc":"2.0", "error":{"code":-32601,"message":"Method not found","data":{"request":{"method":"assets","params":{"mode":"embedded","hideSidebar":"true"}}}},"id":""}`

There's currently no way to work around these errors, making the Performance page effectively useless when debugging Dart CLI applications.

kenzieschmoll:
![image](https://user-images.githubusercontent.com/43759233/213552759-1a3cdd6c-198d-40de-9d1e-864d40fc0faa.png)

PR #5104, opened 2023-01-24
Fix state management and offline bugs with the Perfetto trace viewer

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/5023. We were not properly setting "isActiveFeature" for non-flutter apps. This PR fixes the state management issues around that for both online and offline mode.

RELEASE_NOTE_EXCEPTION=[fixes a bug that is behind a flag]

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_web.dart
+++ b/packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_web.dart
@@ -39,7 +39,11 @@ class _PerfettoState extends State<Perfetto> with AutoDisposeMixin {
     _perfettoController = widget.perfettoController as PerfettoControllerImpl;
     _viewController = _PerfettoViewController(_perfettoController)..init();
 
-    _loadActiveTrace();
+    // If [_perfettoController.activeTraceEvents] has a null value, the trace
+    // data has not yet been initialized.
+    if (_perfettoController.activeTraceEvents.value != null) {
+      _loadActiveTrace();
+    }
     addAutoDisposeListener(
       _perfettoController.activeTraceEvents,
       _loadActiveTrace,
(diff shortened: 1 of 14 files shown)


packages/devtools_app/lib/src/screens/performance/performance_controller.dart
```diff
@@ -131,6 +137,28 @@ class PerformanceController extends DisposableController
           }
         }),
       );
+    } else {
+      final shouldLoadOfflineData = offlineController
```
kenzieschmoll, line 141: this was copied from performance_screen.dart, where it used to live in didChangeDependencies

packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_web.dart
```diff
@@ -53,7 +57,10 @@ class _PerfettoState extends State<Perfetto> with AutoDisposeMixin {
   }
 
   void _loadActiveTrace() {
-    _viewController._loadTrace(_perfettoController.activeTraceEvents.value);
+    assert(_perfettoController.activeTraceEvents.value != null);
+    _viewController._loadTrace(
+      _perfettoController.activeTraceEvents.value ?? [],
+    );
```
polina-c, line 63: Why don't we bang instead of assert here?

With assert, if null squeeze somehow, feature will silently stop working properly and we will never find this out.

Another option is to function gracefully and issue analytic event.

polina-c (approved):
One comment then lgtm.

packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_web.dart
kenzieschmoll: used band instead.

PR #5105, opened 2023-01-24
Update vm_service to 11.0.0 and finish profiling migration

CoderDake:
Blocked waiting for https://dart-review.googlesource.com/c/sdk/+/279122 to be merged and be available on the latest flutter candidate.

Addresses some comments in: https://github.com/flutter/devtools/issues/2860

# Details
This PR migrates vm_service to 11.0.0 in order to benefit from  certain `id`s being `String` instead of `int`.
This finalizes the migration that will ensure that all recorded network requests show up in the Network tab. Previously integer overflows were causing those requests to be lost.

RELEASE_NOTE_EXCEPTION=Internal migration only


packages/devtools_app/test/network/network_model_test.dart
```diff
@@ -62,8 +62,8 @@ void main() {
     });
 
     test('equals and hash return the correct value', () {
-      expect(testSocket1.hashCode, equals(0));
-      expect(testSocket2.hashCode, equals(1));
+      expect(testSocket1.hashCode, equals(0.hashCode));
+      expect(testSocket2.hashCode, equals(1.hashCode));
```
kenzieschmoll, line 66: shouldn't this be '0'.hashcode since we are dealing with strings now?

kenzieschmoll (approved):
2 questions but generally lgtm

CoderDake:
The tests should pass once the new test_core package is published

bkonyi (approved):

PR #5127, opened 2023-01-26
Make httpProfile* requests flexible for both int and String

CoderDake:
The changes in  https://dart-review.googlesource.com/c/sdk/+/279122 will immediately apply in g3, and so devtools needs to be able to handle the vm_service:11.0.0 changes, before they can be published.

This change makes sure that devtools will still behave well, regardless of whether `SocketStatistic.id` or `HttpProfileRequestRef.id` are `int` or `String`.

# Testing
Ensure all failing regressions from https://dart-in-g3-qa-prod.corp.google.com/dg3/Home#/cbuild/find/e7d8261f9ab2b6d472c49d8364e8510a4c0fc827
Pass for both vm_service:10.0.0 and vm_service:11.0.0

## Results
- The contents of this PR were tested with the following
  - Test against vm_service:11.0.0
    - see https://dart-review.googlesource.com/c/sdk/+/279122?tab=comments 
  - Tested against vm_service:10.0.0
- To ensure that my tests actually tested the changes properly I also ran those regressions on a clean client, and saw that they failed. So I know that this PR will resolve the issues created in them.
RELEASE_NOTE_EXCEPTION=Not user facing, in preparation for migration.


packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -843,12 +843,18 @@ class VmServiceWrapper implements VmService {
...
   Future<HttpProfileRequest> getHttpProfileRequest(
     String isolateId,
-    int id,
+    String id,
   ) async {
     assert(await isHttpProfilingAvailable(isolateId));
     return trackFuture(
       'getHttpProfileRequest',
-      _vmService.getHttpProfileRequest(isolateId, id),
+      _vmService.callMethod(
```
bkonyi, line 851: Leave a TODO here to switch to the actual implementation once we roll forward to v11.

bkonyi (approved):

kenzieschmoll (approved):

PR #5131, opened 2023-01-27
Add Class Hierarchy explorer to program structure explorer

bkonyi:
Also fixed an issue where non-core libraries could not be jumped to in the program structure explorer due to an assumption that all libraries would be at the root of the program explorer.

**Demo:**

https://user-images.githubusercontent.com/24210656/215142294-6b4ce065-cf4a-4e19-93c4-66359159b441.mov

RELEASE_NOTE_EXCEPTION=VM developer mode functionality


packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
```diff
@@ -352,12 +352,16 @@ class ProgramExplorerController extends DisposableController
     final LibraryRef targetLib = scriptObj.library!;
 
     // Search targetLib only on the root level nodes.
-    final libNode = _searchRootObjectNodes(targetLib)!;
+    var libNode = _searchRootObjectNodes(targetLib);
+    libNode ??= breadthFirstSearchObject(
```
kenzieschmoll, line 356: can you add a comment here explaining in which case the libNode would be null and then we'd have to run the bfs?

packages/devtools_app/lib/src/screens/vm_developer/object_inspector/class_hierarchy_explorer.dart
```diff
@@ -0,0 +1,34 @@
...
+  final ObjectInspectorViewController controller;
+
+  @override
+  Widget build(BuildContext context) {
+    return TreeView<ClassHierarchyNode>(
+      dataRootsListenable:
+          controller.classHierarchyController.selectedIsolateClassHierarchy,
+      dataDisplayProvider: (node, onPressed) => VmServiceObjectLink<Class>(
+        object: node.cls,
+        onTap: (cls) async => await controller.findAndSelectNodeForObject(cls),
```
kenzieschmoll, line 28: having async code in a build method feels fishy. The UI shouldn't be blocked on awaiting this operation

packages/devtools_app/lib/src/screens/vm_developer/object_inspector/class_hierarchy_explorer.dart
```diff
@@ -0,0 +1,34 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: 23

packages/devtools_app/lib/src/shared/ui/tab.dart
```diff
@@ -193,15 +193,15 @@ class _AnalyticsTabbedViewState extends State<AnalyticsTabbedView>
...
         Expanded(
-          child: widget.trailingWidgets[_currentTabControllerIndex],
+          child: TabBar(
+            labelColor: theme.textTheme.bodyLarge?.color,
+            controller: _tabController,
+            tabs: widget.tabs,
+            isScrollable: true,
+          ),
         ),
+        widget.trailingWidgets[_currentTabControllerIndex],
```
kenzieschmoll, line 204: are we sure this change isn't affecting UI on other pages? all the inspector golden updates seem suspicious.

packages/devtools_app/lib/src/screens/vm_developer/object_inspector/class_hierarchy_explorer_controller.dart
```diff
@@ -0,0 +1,78 @@
...
+// found in the LICENSE file.
+
+import 'package:collection/collection.dart';
+import 'package:flutter/foundation.dart';
+import 'package:vm_service/vm_service.dart';
+
+import '../../../shared/globals.dart';
+import '../../../shared/primitives/trees.dart';
+
+class ClassHierarchyExplorerController {
```
kenzieschmoll, line 12: can you add dart doc that defines what "class hierarchy" is? what makes one class the parent or child of another?

packages/devtools_app/lib/src/screens/vm_developer/object_inspector/class_hierarchy_explorer.dart
bkonyi: The underlying `TapGestureRecognizer` that invokes this callback never awaits, so it shouldn't block the UI thread. Do you still want me to remove async/await here?

packages/devtools_app/lib/src/screens/vm_developer/object_inspector/class_hierarchy_explorer_controller.dart
bkonyi: Added documentation to the widget.

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
```diff
@@ -351,13 +351,22 @@ class ProgramExplorerController extends DisposableController
         await service.getObject(isolateId, targetScript.id!) as Script;
     final LibraryRef targetLib = scriptObj.library!;
 
-    // Search targetLib only on the root level nodes.
-    final libNode = _searchRootObjectNodes(targetLib)!;
+    // Search targetLib only on the root level nodes (this is the most common
+    // scenario).
+    var libNode = _searchRootObjectNodes(targetLib);
+
+    // If we couldn't the target library as a root node, it's possible we have
```
kenzieschmoll, line 358: nit: couldn't find the target?

packages/devtools_app/lib/src/screens/vm_developer/object_inspector/class_hierarchy_explorer.dart
kenzieschmoll: we should either leave as is and replace `await` with `unawaited(...)`, or can you write this as a tearoff instead `onTap: controller.findAndSelectNodeForObject`?

packages/devtools_app/lib/src/screens/vm_developer/object_inspector/class_hierarchy_explorer_controller.dart
```diff
@@ -0,0 +1,78 @@
...
+  ValueListenable<List<ClassHierarchyNode>> get selectedIsolateClassHierarchy =>
+      _selectedIsolateClassHierarchy;
+  final _selectedIsolateClassHierarchy =
+      ValueNotifier<List<ClassHierarchyNode>>([]);
```
kenzieschmoll, line 16: does this need to be a list? looks like there is only ever 1 or 0 elements in it

packages/devtools_test/lib/src/mocks/fake_class_hierarchy_explorer_controller.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// ignore: implementation_imports
+import 'package:devtools_app/src/screens/vm_developer/object_inspector/class_hierarchy_explorer_controller.dart';
```
kenzieschmoll, line 6: you could just expose this class in devtools_app.dart instead

packages/devtools_test/lib/src/mocks/fake_object_inspector_view_controller.dart
```diff
@@ -0,0 +1,18 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// ignore_for_file: implementation_imports
+import 'package:devtools_app/src/screens/vm_developer/object_inspector/class_hierarchy_explorer_controller.dart';
+import 'package:devtools_app/src/screens/vm_developer/object_inspector/object_inspector_view_controller.dart';
```
kenzieschmoll, line 7: same

kenzieschmoll (approved):
a few comments, but overall lgtm

PR #5134, opened 2023-01-28
Improve performance of heap snapshot.

### LINKED ISSUE #4822: slow performance when taking a heap snapshot
we should batch work so we don't freeze the ui while a snapshot is being taken and the data is being processed.

polina-c:
closes #4822

### DIFF
--- a/packages/devtools_app/lib/src/shared/memory/adapted_heap_data.dart
+++ b/packages/devtools_app/lib/src/shared/memory/adapted_heap_data.dart
@@ -81,7 +81,7 @@ class AdaptedHeapData {
 
     final objects = <AdaptedHeapObject>[];
     for (final i in Iterable.generate(graph.objects.length)) {
-      await _uiReleaser.step();
+      if (_uiReleaser.step()) await _uiReleaser.releaseUi();
       final object =
           AdaptedHeapObject.fromHeapSnapshotObject(graph.objects[i], i);
       objects.add(object);
--- a/packages/devtools_app/lib/src/shared/memory/adapted_heap_data.dart
+++ b/packages/devtools_app/lib/src/shared/memory/adapted_heap_data.dart
@@ -5,6 +5,7 @@
 import 'package:collection/collection.dart';
 import 'package:vm_service/vm_service.dart';
 
+import '../../../devtools_app.dart';
 import 'class_name.dart';
 import 'simple_items.dart';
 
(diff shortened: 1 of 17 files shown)


polina-c:
Initial benchmark for snapshot: 30 seconds for gallery homepage
Conversion of methods to late final and excluding toLongString from equality operator reduced time to 10 seconds.
Async processing increased time to 20 seconds.

packages/devtools_app/lib/src/shared/memory/adapted_heap_data.dart
```diff
@@ -5,6 +5,7 @@
 import 'package:collection/collection.dart';
 import 'package:vm_service/vm_service.dart';
 
+import '../../../devtools_app.dart';
```
bkonyi, line 8: Avoid importing `devtools_app.dart`.

packages/devtools_app/lib/src/shared/memory/class_name.dart
```diff
@@ -102,15 +101,16 @@ class HeapClassName {
...
-  bool get isSentinel => className == 'Sentinel' && library.isEmpty;
+  late final isSentinel = className == 'Sentinel' && library.isEmpty;
 
-  bool get isRoot => className == 'Root' && library.isEmpty;
+  late final isRoot = className == 'Root' && library.isEmpty;
 
   /// Whether a class can hold a reference to an object
   /// without preventing garbage collection.
-  bool get isWeakEntry {
+  late final bool isWeakEntry = () {
```
bkonyi, line 113: See comments on https://github.com/flutter/devtools/pull/5135

bkonyi (commented):
I think most of my comments on https://github.com/flutter/devtools/pull/5135 apply here as well.

packages/devtools_app/lib/src/shared/memory/adapted_heap_data.dart
polina-c: addressed in https://github.com/flutter/devtools/pull/5135

packages/devtools_app/lib/src/shared/memory/class_name.dart
polina-c: Separated method from initializer.

kenzieschmoll:
20 seconds still seems like a long time. Are there ways we can optimize further?

polina-c:
> 20 seconds still seems like a long time. Are there ways we can optimize further?

Will try in separate PR.

Issue #5045, opened 2023-01-13
Remove the Source selector on the memory screen


kenzieschmoll:
![Screenshot 2023-01-13 at 10 47 56 AM](https://user-images.githubusercontent.com/43759233/212396311-ecd39d2b-c926-4cfe-8ce4-7b52aa496844.png)

This doesn't appear to be hooked up to anything currently. If this also makes the Export button useless in the current state, we should remove that button too until we have import / export support fully hooked up for the memory screen

PR #5135, opened 2023-01-29
Remove export and source from memory screen.

polina-c:
Fixes https://github.com/flutter/devtools/issues/5045

I postponed code clean up, because some code will be helpful for future implementation of offline mode.


packages/devtools_app/lib/src/screens/memory/panes/leaks/diagnostics/leak_analyzer.dart
```diff
@@ -10,8 +10,8 @@ import '../../../shared/heap/spanning_tree.dart';
 import 'model.dart';
 
 /// Analyzes notGCed leaks and returns result of the analysis.
-NotGCedAnalyzed analyseNotGCed(NotGCedAnalyzerTask task) {
-  analyzeHeapAndSetRetainingPaths(task.heap, task.reports);
+Future<NotGCedAnalyzed> analyseNotGCed(NotGCedAnalyzerTask task) async {
```
bkonyi, line 13: Ubernit: use analyze instead of analyse for consistency.

packages/devtools_app/lib/src/shared/memory/adapted_heap_data.dart
```diff
@@ -70,14 +71,21 @@ class AdaptedHeapData {
...
+    final objects = <AdaptedHeapObject>[];
+    for (final i in Iterable.generate(graph.objects.length)) {
+      if (_uiReleaser.step()) await _uiReleaser.releaseUi();
+      final object =
+          AdaptedHeapObject.fromHeapSnapshotObject(graph.objects[i], i);
+      objects.add(object);
+    }
 
-    return AdaptedHeapData(objects);
+    final result = AdaptedHeapData(objects);
```
bkonyi, line 87: Just return `AdaptedHeapData` instead of assigning it to a local.

packages/devtools_app/lib/src/shared/memory/class_name.dart
```diff
@@ -102,15 +101,16 @@ class HeapClassName {
...
-  bool get isSentinel => className == 'Sentinel' && library.isEmpty;
+  late final isSentinel = className == 'Sentinel' && library.isEmpty;
 
-  bool get isRoot => className == 'Root' && library.isEmpty;
+  late final isRoot = className == 'Root' && library.isEmpty;
 
   /// Whether a class can hold a reference to an object
   /// without preventing garbage collection.
-  bool get isWeakEntry {
+  late final bool isWeakEntry = () {
```
bkonyi, line 113: Is this just to lazily initialize `isWeakEntry`? This behavior might not be obvious to everyone, so I'd recommend leaving `isWeakEntry` as a getter and creating a nullable private property that is set the first time `isWeakEntry` is invoked.

We need to do extra runtime checks with late fields anyway, so there's no benefit doing it this way over using a nullable private field when it comes to performance.

packages/devtools_app/lib/src/shared/memory/class_name.dart
```diff
@@ -130,14 +130,22 @@ class HeapClassName {
     // class names.
     assert(false, 'Unexpected library for $className: $library.');
     return false;
-  }
+  }();
 
-  String get shortName {
+  late final shortName = () {
```
bkonyi, line 135: Ditto

packages/devtools_app/lib/src/shared/memory/class_name.dart
```diff
@@ -130,14 +130,22 @@ class HeapClassName {
...
     return className;
-  }
+  }();
+
+  ClassType? _cachedClassType;
 
   ClassType classType(String? rootPackage) {
+    if (_cachedClassType != null) return _cachedClassType!;
+    _cachedClassType = _classType(rootPackage);
+    return _cachedClassType!;
```
bkonyi, line 145: This can be `return _cachedClassType ??= _classType(rootPackage)!`

packages/devtools_app/lib/src/shared/memory/class_name.dart
```diff
@@ -149,22 +157,21 @@ class HeapClassName {
...
-  bool get isPackageless =>
-      library.isEmpty ||
+  late final bool isPackageless = library.isEmpty ||
       (!library.startsWith(PackagePrefixes.dart) &&
           !library.startsWith(PackagePrefixes.genericDartPackage));
 
   /// True, if the package has prefix `dart:` or has perfix `package:` and is
   /// published by Dart or Flutter org.
-  bool get isDartOrFlutter {
+  late final isDartOrFlutter = () {
```
bkonyi, line 174: Ditto

packages/devtools_app/lib/src/shared/memory/class_name.dart
polina-c: I started doing `late final` initializers after @jacob314 suggested it as a recommended way to do lazy initializations in one of code reviews. 

For people who know this pattern, `late final` seems to be more concise and readable lazy initialization than private field.

Jacob, could you comment?

polina-c:
I managed to mess up this PRs with another one (https://github.com/flutter/devtools/pull/5134). So, converting this to draft for now to land performance first.

packages/devtools_app/lib/src/shared/memory/class_name.dart
jacob314: In general, I think it is reasonable to expect everyone to understand what `late final` means in Dart as it is a feature that was thoughtfully added to Dart as part of supporting null safety. I do agree it is a bit of surprising feature at first.
There is a tiny performance benefit as compilers can know that a late final field is only initialized once. This is also why we can't have `late` objects that have a `dispose` lifecycle because that would violate the compilers assumptions.

I wonder if makes sense to get some guidance on this added to https://dart.dev/guides/language/effective-dart

bkonyi: AFAIK, each access to a `late final` field still requires a sentinel check to ensure the field has been initialized, which is basically the equivalent to a null check on each access, so there's no benefit over using one or the other performance wise.

I do personally think this is less readable (declaring a closure only to invoke it immediately looks smelly to me) than the alternative.

bkonyi: At the very least, I think these closures should be made into named methods to avoid having large blocks of initialization logic inline.

packages/devtools_app/lib/src/shared/memory/class_name.dart
polina-c: resolved in https://github.com/flutter/devtools/pull/5134

packages/devtools_app/lib/src/shared/memory/class_name.dart
polina-c: resolved in https://github.com/flutter/devtools/pull/5134

packages/devtools_app/lib/src/shared/memory/class_name.dart
polina-c: Separated calculations into method.

packages/devtools_app/lib/src/screens/memory/panes/leaks/diagnostics/leak_analyzer.dart
polina-c: resolved in https://github.com/flutter/devtools/pull/5134

kenzieschmoll (commented):
we should also delete controller.memorySource logic

polina-c:
> we should also delete controller.memorySource logic. This should not be needed for future implementation of offline mode since the offline snapshot will use a different instance of MemoryController

https://github.com/flutter/devtools/issues/5145

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
```diff
@@ -313,9 +260,12 @@ class MemoryController extends DisposableController
     );
   }
 
+  /// This flag will be needed for offline mode implementation.
+  bool offline = false;
+
```
kenzieschmoll, line 265: this actually will not be needed. We will use `offlineController.offlineMode.value` to check whether we are in offline mode or not.

packages/devtools_app/lib/src/screens/memory/panes/chart/chart_control_pane.dart
```diff
@@ -83,9 +83,7 @@ class _ChartControlPaneState extends State<ChartControlPane>
             ),
             const SizedBox(width: defaultSpacing),
             ClearButton(
-              onPressed: controller.memorySource == MemoryController.liveFeed
-                  ? _clearTimeline
-                  : null,
+              onPressed: controller.offline ? null : _clearTimeline,
```
kenzieschmoll, line 86: we'll actually just hide this control in offline mode. Why not just delete controller.offline and all its uses until offline mode is supported for the memory screen? Otherwise we'll get into a state where some things have a half baked "offline" implementation. I'd suggest setting this onPressed value to _clearTimeline unconditionally

packages/devtools_app/test/memory/memory_screen_test.dart
```diff
@@ -48,7 +48,7 @@ void main() {
   }
 
   void initControllerState() {
-    controller.offline.value = true;
+    controller.offline = true;
```
kenzieschmoll, line 51: seems like we shouldn't be relying on tests that test a code path that is now dead.

packages/devtools_app/lib/src/screens/memory/memory_controller.dart
polina-c: I want to preserve all references to this variable, so that we do not forget to review them and decide what will be their behavior in offline mode.

packages/devtools_app/lib/src/screens/memory/shared/primitives/memory_timeline.dart
```diff
@@ -19,7 +19,7 @@ enum ContinuesState {
 
 /// All Raw data received from the VM and offline data loaded from a memory log file.
 class MemoryTimeline {
-  MemoryTimeline(this.offline);
+  MemoryTimeline();
```
kenzieschmoll, line 22: nit: this constructor is now redundant and can be removed.

packages/devtools_app/test/memory/memory_screen_test.dart
```diff
@@ -104,150 +93,18 @@ void main() {
...
-        ),
-      ) as Text;
-
-      expect(
-        memorySources.data,
-        '${controller.memorySourcePrefix}${MemoryController.liveFeed}',
-      );
-    });
-
-    testWidgetsWithWindowSize('Chart Select Hover Test', windowSize,
```
kenzieschmoll, line 139: why are we removing this test? seems we just need to switch the canned data to be loaded as live data instead of offline data, and then we can still test the chart hover card

polina-c: The test is broken. It does not test hover card. There is no card on the image.

polina-c: And there is Terry's TODO in the comments.

kenzieschmoll (approved):

PR #5163, opened 2023-02-01
Fix analyzer regression introduced in latest flutter candidate

CoderDake:
new flutter candidate breaks this test. Fixing it here
RELEASE_NOTE_EXCEPTION=internal only


kenzieschmoll (approved):

bkonyi (approved):

kenzieschmoll:
nit: can you update your PR description so that it explains what the fix is? (changelog generator script will use the PR description)

PR #5181, opened 2023-02-02
Merge Source column and method name columns

kenzieschmoll:
This ensures that the source URI is always visible for a deeply nested tree. This also allows for more flexible width on the source uris so that we do not have to ellipsize them.

Screenshot 2023-02-02 at 2 10 18 PM


bkonyi (approved):
LGTM once checks are fixed.

packages/devtools_app/test/cpu_profiler/profiler_screen_test.dart
```diff
@@ -5,57 +5,33 @@
...
+  setUp(() async {
+    // Modify the character width that will be used to calculate column sizes
+    // in the profiler table. The flutter_tester device uses a redacted font.
+    setAssumedMonospaceCharacterWidth(14.0);
+    scene = CpuProfilerDefaultScene();
+    await scene.setUp();
+  });
 
-  const windowSize = Size(2000.0, 1000.0);
+  const windowSize = Size(15000.0, 15000.0);
```
bkonyi, line 26: Do these need to be this big?

packages/devtools_app/test/memory/tracing/tracing_view_test.dart
```diff
@@ -88,50 +50,66 @@ void main() {
...
-    setUp(() {
-      setGlobal(DevToolsExtensionPoints, ExternalDevToolsExtensionPoints());
-      setGlobal(NotificationService, NotificationService());
-      setGlobal(OfflineModeController, OfflineModeController());
-      setGlobal(IdeTheme, IdeTheme());
-      setGlobal(PreferencesController, PreferencesController());
+    setUp(() async {
+      // Modify the character width that will be used to calculate column sizes
+      // in the profiler table. The flutter_tester device uses a redacted font.
+      setAssumedMonospaceCharacterWidth(14.0);
```
bkonyi, line 78: Can we make a test util for this that we can call without specifying the actual character width each time? This should be the same for all the tests, right?

bkonyi (approved):

PR #5193, opened 2023-02-06
Live outbound references for map.

polina-c:

Screenshot 2023-02-06 at 10 50 04 AM


packages/devtools_app/lib/src/shared/diagnostics/variable_factory.dart
```diff
@@ -344,8 +365,25 @@ List<DartObjectNode> createVariablesForAssociations(
     false,
     (p, e) => p || isPrimativeInstanceKind(e.key.kind),
   );
-  for (var i = 0; i < associations.length; i++) {
+  for (final i in Iterable.generate(associations.length)) {
```
kenzieschmoll, line 368: what's the reason for this change? This requires us to create another iterable instead of just looping through the one we already have `associations`

packages/devtools_app/lib/src/shared/diagnostics/variable_factory.dart
```diff
@@ -344,8 +365,25 @@ List<DartObjectNode> createVariablesForAssociations(
...
     final association = associations[i];
+
+    if (asReferences) {
```
kenzieschmoll, line 371: will we eventually have to differentiate between whether this should be added as a live reference or a static reference?

packages/devtools_app/lib/src/shared/diagnostics/variable_factory.dart
polina-c: reverted

packages/devtools_app/lib/src/shared/diagnostics/variable_factory.dart
polina-c: Static references never show up here. I plan to reorganize code to make it more supportable.

kenzieschmoll (approved):

PR #5196, opened 2023-02-06
Consolidate memory browsing code in one library.  

polina-c:

No functional changes.

### DIFF
--- a/packages/devtools_app/lib/src/shared/diagnostics/tree_builder.dart
+++ b/packages/devtools_app/lib/src/shared/diagnostics/tree_builder.dart
@@ -15,7 +15,9 @@ import '../primitives/utils.dart';
 import 'dart_object_node.dart';
 import 'diagnostics_node.dart';
 import 'generic_instance_reference.dart';
+import 'helpers.dart';
 import 'inspector_service.dart';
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/shared/diagnostics/tree_builder.dart
```diff
@@ -249,44 +140,32 @@ Future<void> _addInstanceRefItems(
...
-            value: ref.value,
-            isolateRef: ref.isolateRef,
-            heapSelection: ref.heapSelection,
-          ),
-        ),
-        index: 0,
-      );
+    if (FeatureFlags.evalAndBrowse &&
+        ref != null &&
+        ref.heapSelection != null) {
```
kenzieschmoll, line 152: this can be simplified to ref?.heapSelection != null

kenzieschmoll (approved):

packages/devtools_app/lib/src/shared/diagnostics/tree_builder.dart
polina-c: I need it to promote ref to non-nullable

PR #5202, opened 2023-02-07
Make limit for number of references configurable.

polina-c:
RELEASE_NOTE_EXCEPTION=[does not affect released features]

We need this because vm service asks to set limit in request.

Screenshot 2023-02-07 at 1 29 03 PM


packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -312,13 +312,21 @@ class InspectorPreferencesController extends DisposableController
...
   static const _androidCollectionEnabledStorageId =
       'memory.androidCollectionEnabled';
 
+  /// If false, mamory chart will be collapsed.
   final showChart = ValueNotifier<bool>(true);
   static const _showChartStorageId = 'memory.showChart';
 
+  /// Number of references to request from vm service,
+  /// when browsing references in console.
+  final refLimit = ValueNotifier<int>(100);
```
kenzieschmoll, line 327: put 100 into a const `_defaultBrowseReferencesLimit`

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -355,6 +363,23 @@ class MemoryPreferencesController extends DisposableController
...
+        );
+
+        ga.select(
+          gac.memory,
+          gac.MemoryEvent.browseRefLimit,
+        );
+      },
+    );
+    refLimit.value =
+        int.tryParse(await storage.getValue(_refLimitStorageId) ?? '') ?? 100;
```
kenzieschmoll, line 382: use const that you create above here instead of 100

kenzieschmoll (approved):
lgtm after comments

Issue #2524, opened 2020-11-30
Support search and filter in tree tables


kenzieschmoll:

jakemac53:
+1 I would like this :)

kenzieschmoll:
Filtering is complete. This issue now tracks search support for tree tables.

PR #5204, opened 2023-02-07
Add text filtering to the CPU profiler

kenzieschmoll:
Before this PR, the CPU profiler only supported a small set of "toggle" filters. This adds a query filter to the profiler's filter dialog. This PR also increases the test coverage for FilterControllerMixin. 

This required some refactoring / cleanup to the existing Filter functionality. Work towards https://github.com/flutter/devtools/issues/2524.


packages/devtools_app/lib/src/screens/profiler/cpu_profile_controller.dart
```diff
@@ -551,11 +531,13 @@ class CpuProfilerController
     resetSearch();
   }
 
+  @override
   void dispose() {
     _dataNotifier.dispose();
     _selectedCpuStackFrameNotifier.dispose();
     _processingNotifier.dispose();
     transformer.dispose();
+    super.dispose();
```
bkonyi, line 540: Nit: `super.dispose()` should probably be the first call in this method.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_controller.dart
```diff
@@ -593,24 +610,43 @@ class CpuProfilerController
...
+    filter ??= activeFilter.value;
     final filterCallback = (CpuStackFrame stackFrame) {
-      var shouldInclude = true;
-      for (final toggleFilter in filter.toggleFilters!) {
+      for (final toggleFilter in filter!.toggleFilters) {
         if (toggleFilter.enabled.value) {
-          shouldInclude =
-              shouldInclude && toggleFilter.includeCallback(stackFrame);
-          if (!shouldInclude) return false;
+          if (!toggleFilter.includeCallback(stackFrame)) return false;
```
bkonyi, line 620: Nit: consider collapsing these two conditionals into a single check.

packages/devtools_app/lib/src/screens/profiler/cpu_profiler.dart
```diff
@@ -39,11 +39,9 @@ class CpuProfiler extends StatefulWidget {
         tabs = [
           if (summaryView != null)
             _buildTab(key: summaryTab, tabName: 'Summary'),
-          if (!data.isEmpty) ...[
-            _buildTab(key: bottomUpTab, tabName: 'Bottom Up'),
-            _buildTab(key: callTreeTab, tabName: 'Call Tree'),
-            _buildTab(key: flameChartTab, tabName: 'CPU Flame Chart'),
-          ],
+          _buildTab(key: bottomUpTab, tabName: 'Bottom Up'),
```
bkonyi, line 42: Can the tabs handle no data now?

packages/devtools_app/lib/src/shared/ui/filter.dart
```diff
@@ -10,68 +11,158 @@ import '../primitives/auto_dispose.dart';
...
+/// storing the state of those filters.
+///
+/// To use this mixin, you must implement [filterData] as well as either or both
+/// [createToggleFilters] and [createQueryFilterArgs].
+///
+/// Classes mixing in [FilterControllerMixin] must also extend
+/// [DisposableController] and mixin [AutoDisposeControllerMixin], and a class
+/// can subscribe to updates to the active filter by calling
+/// [subscribeToFilterChanges].
+mixin FilterControllerMixin<T> implements AutoDisposeControllerMixin {
```
bkonyi, line 27: If this can only be applied to classes extending `DisposableController`, we should enforce that:

```dart
mixin FilterControllerMixin<T> on DisposableController implements AutoDisposeControllerMixin {...}
```

packages/devtools_app/lib/src/shared/ui/filter.dart
```diff
@@ -10,68 +11,158 @@ import '../primitives/auto_dispose.dart';
...
+  static const filterTagSeparator = '-#-';
+
   final filteredData = ListValueNotifier<T>([]);
 
-  final activeFilter = ValueNotifier<Filter<T>?>(null);
+  // TODO(kenz): replace [Filter] class with a pattern when available. All we
```
bkonyi, line 32: Ubernit: "replace [Filter] class with a record when available"

packages/devtools_app/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -27,7 +27,6 @@
 		33CC10F32044A3C60003C045 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 33CC10F22044A3C60003C045 /* Assets.xcassets */; };
 		33CC10F62044A3C60003C045 /* MainMenu.xib in Resources */ = {isa = PBXBuildFile; fileRef = 33CC10F42044A3C60003C045 /* MainMenu.xib */; };
 		33CC11132044BFA00003C045 /* MainFlutterWindow.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33CC11122044BFA00003C045 /* MainFlutterWindow.swift */; };
-		C5F94F99E5361C4E4D541DAE /* Pods_RunnerTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 58CC043C129572D76DC706CE /* Pods_RunnerTests.framework */; };
```
bkonyi, line 30: Was this intentionally committed?

packages/devtools_app/test/cpu_profiler/cpu_profiler_controller_test.dart
```diff
@@ -25,6 +26,14 @@ void main() {
   group('CpuProfileController', () {
     late CpuProfilerController controller;
 
+    Future<void> disableAllFiltering() async {
+      for (final filter in controller.activeFilter.value.toggleFilters) {
+        filter.enabled.value = false;
+      }
+      controller.setActiveFilter();
+      await shortDelay();
```
bkonyi, line 34: What's the reason for the short delay? Can you add a comment?

packages/devtools_app/test/cpu_profiler/cpu_profiler_controller_test.dart
```diff
@@ -104,37 +114,54 @@ void main() {
...
-      controller.toggleFilters[2].enabled.value = true;
-      expect(
-        controller.generateToggleFilterSuffix(),
-        equals(
-          'Hide Native code,Hide core Dart libraries,Hide core Flutter libraries',
-        ),
-      );
+      // At this point, data is filtered by the default toggle filter values.
+      var filteredData = controller.dataNotifier.value!;
+      expect(filteredData.stackFrames.values.length, equals(12));
```
bkonyi, line 134: Ubernit: the `equals(...)` matcher is the default, so it's not necessary. Feel free to remove it if you'd like.

packages/devtools_app/test/cpu_profiler/cpu_profiler_controller_test.dart
```diff
@@ -104,37 +114,54 @@ void main() {
...
 
-      controller.toggleFilters[1].enabled.value = false;
-      expect(
-        controller.generateToggleFilterSuffix(),
-        equals('Hide Native code,Hide core Flutter libraries'),
-      );
+      controller.setActiveFilter(query: 'uri:dart:vm');
+      await shortDelay();
```
bkonyi, line 137: What are the delays for?

packages/devtools_app/test/shared/filter_test.dart
```diff
@@ -0,0 +1,295 @@
...
+        ),
+      );
+
+      // Query and toggle filter.
+      controller.toggleFilters[0].enabled.value = true;
+      controller.toggleFilters[1].enabled.value = true;
+      controller.setActiveFilter(query: 'Ba cat:foobar');
+      expect(
+        controller.filteredData.value.toString(),
+        equals('[1-FooBar-foobar]'),
```
bkonyi, line 93: I think we need one more test case here... ;-)

```dart
controller.toggleFilters[0].enabled.value = true;
controller.toggleFilters[1].enabled.value = true;
controller.setActiveFilter(query: 'Basset');
// There ain't nothin' but a hound dog
expect(
  controller.filteredData.value.toString(),
  equals('[5-Basset Hound-dog]'),
);
```

packages/devtools_app/test/shared/filter_test.dart
```diff
@@ -0,0 +1,295 @@
...
+        controller.filteredData.value.toString(),
+        equals('[]'),
+      );
+    });
+
+    test('isFilterActive', () {
+      controller.toggleFilters[0].enabled.value = true;
+      controller.toggleFilters[1].enabled.value = false;
+      controller.setActiveFilter();
+      expect(controller.isFilterActive, equals(true));
```
bkonyi, line 110: Ubernit: just use `true` or `isTrue`

packages/devtools_app/test/shared/filter_test.dart
```diff
@@ -0,0 +1,295 @@
...
+        categoryFilterId: QueryFilterArgument(keys: ['cat', 'c']),
+      };
+
+  @override
+  void filterData(Filter<_TestDataClass> filter) {
+    super.filterData(filter);
+    if (filter.isEmpty) {
+      filteredData
+        ..clear()
+        ..addAll(data);
```
bkonyi, line 228: We can remove the `else` if we just return here, which will make this a bit easier to read.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_controller.dart
kenzieschmoll: dispose is the one exception where we want to dispose last so that a subclass's disposable items are disposed before the superclass. Most of our codebase follows this pattern, but we probably need to clean up a few cases for consistency.

packages/devtools_app/lib/src/screens/profiler/cpu_profiler.dart
kenzieschmoll: Yeah. This is necessary because if a filter is applied that results in no stack frames, the user needs to still have a way to widen the filter. When we got rid of the profiler for an "empty" profile, there was no way for the user to get back to the data they were looking at and filtering.

packages/devtools_app/test/cpu_profiler/cpu_profiler_controller_test.dart
kenzieschmoll: added comment

packages/devtools_app/test/cpu_profiler/cpu_profiler_controller_test.dart
kenzieschmoll: ack. this is a big clean up, so I filed https://github.com/flutter/devtools/issues/5210 to track.

packages/devtools_app/test/cpu_profiler/cpu_profiler_controller_test.dart
kenzieschmoll: `CpuProfilerController.filterData`, which is triggered by the call to `setActiveFilter` via a listener callback, calls an unawaited future. We await a short delay here to ensure that that future completes.

bkonyi (approved):

PR #5206, opened 2023-02-08
Add navigation support to the object inspector

bkonyi:
Allows for navigating back/forward using the browser's back/forward buttons.

This change also updates DevToolsRouterDelegate to keep track of its own navigation history, allowing for history navigation for non-browser targets in the future.

RELEASE_NOTE_EXCEPTION=VM developer mode feature

### DIFF
--- a/packages/devtools_app/lib/src/screens/vm_developer/object_inspector/class_hierarchy_explorer.dart
+++ b/packages/devtools_app/lib/src/screens/vm_developer/object_inspector/class_hierarchy_explorer.dart
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
+
 import 'package:flutter/material.dart';
 import 'package:vm_service/vm_service.dart';
 
@@ -35,7 +37,9 @@ class ClassHierarchyExplorer extends StatelessWidget {
           controller.classHierarchyController.selectedIsolateClassHierarchy,
       dataDisplayProvider: (node, onPressed) => VmServiceObjectLink<Class>(
         object: node.cls,
-        onTap: controller.findAndSelectNodeForObject,
+        onTap: (cls) => unawaited(
+          controller.findAndSelectNodeForObject(context, cls),
+        ),
       ),
       onItemSelected: (_) => null,
       emptyTreeViewBuilder: () => const CenteredCircularProgressIndicator(),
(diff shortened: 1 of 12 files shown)


packages/devtools_app/lib/src/screens/vm_developer/object_inspector/class_hierarchy_explorer.dart
```diff
@@ -35,7 +37,9 @@ class ClassHierarchyExplorer extends StatelessWidget {
           controller.classHierarchyController.selectedIsolateClassHierarchy,
       dataDisplayProvider: (node, onPressed) => VmServiceObjectLink<Class>(
         object: node.cls,
-        onTap: controller.findAndSelectNodeForObject,
+        onTap: (cls) => unawaited(
+          controller.findAndSelectNodeForObject(context, cls),
```
kenzieschmoll, line 41: passing BuildContext around is dangerous. As this context could become disposed while the controller is still trying to use it. Can we access the router without BuildContext? Or if that is not possible, can we lookup the router in widget code so that we don't have a potentially stale build context floating around?

bkonyi: Unfortunately, `Router.navigate` is a static method that requires a `BuildContext`, so there's no way to even look up the router and cache it for later :(

kenzieschmoll: Then if possible, we should have the widget code handle the calls to Router.navigate. We don't want to be passing around BuildContext outside of widget code. The widgets could respond to an event from the controller, and then call the `Router.navigate(context ...)` from somewhere where the build context is accessible (e.g. a StatefulWidget)

Hixie:
I checked with @bkonyi and confirmed that this is still on their radar.

bkonyi:
Picking this back up, but there's been some regressions in the router behavior that need to be resolved before this change can be verified and made ready to land (see https://github.com/flutter/flutter/issues/130433).

bkonyi:
Strange router behavior reported in https://github.com/flutter/flutter/issues/130433 has been fixed and was able to verify this change works correctly in Chrome. @kenzieschmoll, this is ready for another review.

Hixie:
I checked with @bkonyi and confirmed that this is still on their radar as of September 20th.

goderbauer:
(triage) I spoke to @bkonyi and he says this is still on his radar ... eventually.

goderbauer:
(triage): Spoke to @bkonyi last week and he promised to get back to this before he retires.

goderbauer:
(triage): @bkonyi said last week the previous statement is still true.

goderbauer:
(triage): @bkonyi assured me that he still hasn't retired.

bkonyi:
Realistically, I'm not going to get back to this. And no, I'm not retiring... yet.

PR #5215, opened 2023-02-08
Save set of instances as variable.

polina-c:

https://user-images.githubusercontent.com/12115586/217890144-c8ee2605-5485-49a3-b9ff-5e360c926eaa.mov


packages/devtools_app/lib/src/shared/diagnostics/tree_builder.dart
```diff
@@ -119,6 +119,23 @@ void _setupGrouping(DartObjectNode variable) {
   }
 }
 
+Future<void> _addInstanceSetItems(
+  DartObjectNode variable,
+  IsolateRef? isolateRef,
+  InstanceSet instanceSet,
+) async {
+  final instances = instanceSet.instances;
+  if (instances == null) return;
```
bkonyi, line 128: `instances` should never be null here, but could be an empty list.

packages/devtools_app/lib/src/shared/diagnostics/variable_factory.dart
```diff
@@ -482,6 +482,25 @@ List<DartObjectNode> createVariablesForList(
...
 }
 
+List<DartObjectNode> createVariablesForInstanceSet(
+  int offset,
+  int childCount,
+  List<ObjRef> instances,
+  IsolateRef? isolateRef,
+) {
+  final variables = <DartObjectNode>[];
+  for (int i = offset; i < offset + childCount; i++) {
```
bkonyi, line 492: Are we sure that `offset + childCount < instances.length`?

packages/devtools_app/lib/src/shared/diagnostics/tree_builder.dart
```diff
@@ -124,8 +124,7 @@ Future<void> _addInstanceSetItems(
   IsolateRef? isolateRef,
   InstanceSet instanceSet,
 ) async {
-  final instances = instanceSet.instances;
-  if (instances == null) return;
+  final instances = instanceSet.instances ?? [];
```
bkonyi, line 127: `instanceSet.instances` will never be null, so this isn't necessary.

bkonyi (approved):

PR #5224, opened 2023-02-09
Stop referencing inspector_service from diagnostics_node to simplify mental model.

polina-c:

before:

Screenshot 2023-02-09 at 2 00 04 PM

after:

Screenshot 2023-02-09 at 2 06 19 PM

### DIFF
--- a/packages/devtools_app/lib/src/screens/inspector/layout_explorer/ui/layout_explorer_widget.dart
+++ b/packages/devtools_app/lib/src/screens/inspector/layout_explorer/ui/layout_explorer_widget.dart
@@ -183,7 +183,7 @@ abstract class LayoutExplorerWidgetState<W extends LayoutExplorerWidget,
   // this is required so that we don't change focus
   //   when tapping on a child is also Flex-based widget.
   Future<void> setSelectionInspector(RemoteDiagnosticsNode node) async {
-    final service = node.inspectorService;
+    final service = node.objectGroupApi;
     if (service is ObjectGroup) {
       await service.setSelectionInspector(node.valueRef, false);
     }
--- a/packages/devtools_app/lib/src/screens/inspector/layout_explorer/ui/layout_explorer_widget.dart
+++ b/packages/devtools_app/lib/src/screens/inspector/layout_explorer/ui/layout_explorer_widget.dart
@@ -184,7 +184,7 @@ abstract class LayoutExplorerWidgetState<W extends LayoutExplorerWidget,
   //   when tapping on a child is also Flex-based widget.
   Future<void> setSelectionInspector(RemoteDiagnosticsNode node) async {
     final service = node.objectGroupApi;
-    if (service is ObjectGroup) {
+    if (service != null && service is ObjectGroup) {
       await service.setSelectionInspector(node.valueRef, false);
     }
   }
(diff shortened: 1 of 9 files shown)


kenzieschmoll (commented):
This may have g3 implications, as there is another implementation of the inspector service in google3. We should avoid heavy churn to the inspector service and diagnostics node code if it is not critical. I suggest we either assess how important this change is to bring through, or that you have a g3 fix ready to apply before landing this PR

packages/devtools_app/lib/src/screens/inspector/layout_explorer/ui/layout_explorer_widget.dart
```diff
@@ -183,8 +183,8 @@ abstract class LayoutExplorerWidgetState<W extends LayoutExplorerWidget,
   // this is required so that we don't change focus
   //   when tapping on a child is also Flex-based widget.
   Future<void> setSelectionInspector(RemoteDiagnosticsNode node) async {
-    final service = node.inspectorService;
-    if (service is ObjectGroup) {
+    final service = node.objectGroupApi;
+    if (service != null && service is ObjectGroup) {
```
kenzieschmoll, line 187: this null check should be redundant. `service is ObjectGroup` should return false if service is null

polina-c: For some reason `service is ObjectGroup` does not give non-nullness that I need in this code block.

polina-c:
> This may have g3 implications, as there is another implementation of the inspector service in google3. We should avoid heavy churn to the inspector service and diagnostics node code if it is not critical. I suggest we either assess how important this change is to bring through, or that you have a g3 fix ready to apply before landing this PR

I checked [go/roll-devtools](https://goto.google.com/roll-devtools). There is no simple way to provide commit number in the roll to g3 process. How about rolling to g3 immediately after merge of this PR?

kenzieschmoll:
> There is no simple way to provide commit number in the roll to g3 process.

You can change the devtools copybara script to point to your git repo and branch instead of flutter/devtools master. We should prepare a g3 fix prior to landing this PR as this will very likely have g3 implications that could break an internal customer.

kenzieschmoll:
@matt-ragonese will landing this PR cause any issues for your custom screen in g3?

matt-ragonese:
> @matt-ragonese will landing this PR cause any issues for your custom screen in g3?

@polina-c - I think this PR will affect our custom screen in G3, which is currently broken (for other reasons). Would it be possible to temporarily hold off on this change until our internal use case is functional again?

polina-c:
> > @matt-ragonese will landing this PR cause any issues for your custom screen in g3?
> 
> @polina-c - I think this PR will affect our custom screen in G3, which is currently broken (for other reasons). Would it be possible to temporarily hold off on this change until our internal use case is functional again?

sure

Let me know when it is good time.

polina-c:
@matt-ragonese , how the custom screen in G3 is doing?
Is it ok to push the PR forward?

polina-c:
> > @matt-ragonese , how the custom screen in G3 is doing? Is it ok to push the PR forward?
> 
> Apologies for the delay, @polina-c. Yes, this PR should now be unblocked. As @kenzieschmoll mentioned, please prepare a G3 fix prior to merging this PR. Thank you for your patience here!

cl/539191820

Can you test it?
(ignore lints for now)

Issue #5229, opened 2023-02-10
build_runner builds could be faster


jakemac53:
Using generate_for and disabling unused builders will speed up builds significantly

CoderDake:
Thank you for these tips, improvements in this area will speed up our CI builds a lot!

PR #5230, opened 2023-02-10
speed up build_runner builds

jakemac53:
- Use generate_for to target specific input files
- Disabling the mockito generator where it isn't used

Closes https://github.com/flutter/devtools/issues/5229

RELEASE_NOTE_EXCEPTION=Dev exp changes only

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/devtools_app/build.yaml
+++ b/packages/devtools_app/build.yaml
@@ -5,3 +5,9 @@ targets:
         options:
           dart2js_args:
           - -O1
+      # We don't generate any mocks in this package.
+      mockito|mockBuilder:
+        enabled: false
+      stager|stagerAppBuilder:
+        generate_for:
+          - test/test_infra/scenes/**.dart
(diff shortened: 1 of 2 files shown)


jakemac53:
For me this makes the devtools_app build go from a bit over a minute to <20 seconds, and takes devtools_app from >30 seconds down to under 20.

CoderDake:
I'm excited to see how much this speeds up the Github builds 🎉

jakemac53:
Fwiw the largest signal that you could be optimizing here is comparing the number of actions to the number of outputs, I noticed this could be optimized because I saw `[INFO] Succeeded after 1m 15s with 5 outputs (1268 actions)`, and now you get `[INFO] Succeeded after 19.1s with 5 outputs (5 actions)` (this is in devtools_app).

Some builders will have lots of optional actions so it isn't always a good signal, but for simple builds it is :)

CoderDake (approved):

jakemac53:
Looks like it reduces the github action times a bit over a minute :)

jakemac53:
@CoderDake I will let you hit the button when you are ready :)

bkonyi:
Thanks Jake!

Issue #5231, opened 2023-02-10
Clicking `pause` button can crash DevTools


elliette:
Ran into this debugging an internal app. Because we are `null`-asserting `mainIsolate` when checking whether the app is paused, if the `mainIsolate` is `null` DevTools will crash: https://github.com/flutter/devtools/blob/a235825fcb774fb445e6c2456f59a76b503799b4/packages/devtools_app/lib/src/screens/debugger/controls.dart#L37

Stack trace from Chrome DevTools:
```
main.dart.js:12053 Uncaught 
main.dart.js:75177 NoSuchMethodError: method not found: '_isPaused' on null
log @ main.dart.js:75177
main.dart.js:74215 NoSuchMethodError: method not found: '_isPaused' on null
main.dart.js:74215     at _DebuggingControlsState.didChangeDependencies$0 (https://b2607f8b04800100000144d90c0a823628fa9000000000000000001.proxy.googlers.com/main.dart.js:119632:168)
main.dart.js:74215     at StatefulElement._firstBuild$0 (https://b2607f8b04800100000144d90c0a823628fa9000000000000000001.proxy.googlers.com/main.dart.js:252207:30)
main.dart.js:74215     at StatefulElement.mount$2 (https://b2607f8b04800100000144d90c0a823628fa9000000000000000001.proxy.googlers.com/main.dart.js:252139:12)
main.dart.js:74215     at MultiChildRenderObjectElement.inflateWidget$2 (https://b2607f8b04800100000144d90c0a823628fa9000000000000000001.proxy.googlers.com/main.dart.js:251885:19)
```

srawlins:
Closing as resolved: `isMainIsolatePaused` and `isolateManager.mainIsolateState` are now safely handled with null-safe access throughout debugging controls and service management, avoiding null assertions when `mainIsolateState` is null.

If there is more work to do here, please let us know by filing a new issue with up to date information. Thanks!

PR #5232, opened 2023-02-10
Handle a null value for `mainIsolateState` 

elliette:
Work towards https://github.com/flutter/devtools/issues/5231 (doesn't fix underlying issue of `mainIsolateState` being `null`, but prevents a crash).

### DIFF
--- a/packages/devtools_app/lib/src/screens/debugger/controls.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/controls.dart
@@ -33,9 +33,10 @@ class _DebuggingControlsState extends State<DebuggingControls>
   void didChangeDependencies() {
     super.didChangeDependencies();
     if (!initController()) return;
-    addAutoDisposeListener(
-      serviceManager.isolateManager.mainIsolateState!.isPaused,
-    );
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/screens/debugger/controls.dart
```diff
@@ -33,9 +33,10 @@ class _DebuggingControlsState extends State<DebuggingControls>
   void didChangeDependencies() {
     super.didChangeDependencies();
     if (!initController()) return;
-    addAutoDisposeListener(
-      serviceManager.isolateManager.mainIsolateState!.isPaused,
-    );
+    final mainIsolateState = serviceManager.isolateManager.mainIsolateState;
```
kenzieschmoll, line 36: instead of doing a null check on all of these, just use `?` instead of `!`. addAutoDisposeListener will return early if the listenable is null.

elliette:
> general question, why would the mainIsolateState be null if we are actively debugging?

I'm not sure, I haven't been able to reproduce this (only encountered it once). Need to do more investigation

kenzieschmoll (approved):

PR #5246, opened 2023-02-13
Reroot references.

polina-c:

https://user-images.githubusercontent.com/12115586/218583847-f3ae3ec9-939b-487a-916e-6e14dcff7700.mov


packages/devtools_app/lib/src/shared/console/console_service.dart
```diff
@@ -74,6 +74,34 @@ class VariableConsoleLine extends ConsoleLine {
 /// Source of truth for the state of the Console including both events from the
 /// VM and events emitted from other UI.
 class ConsoleService extends Disposer {
+  void appendBrowsableInstance({
+    required InstanceRef? instanceRef,
+    required IsolateRef? isolateRef,
```
bkonyi, line 79: When will we have a null isolate?

packages/devtools_app/lib/src/shared/console/eval/eval_service.dart
```diff
@@ -108,4 +110,25 @@ class EvalService extends DisposableController with AutoDisposeControllerMixin {
...
+        .classes!
+        .firstWhere((ref) => object.heapClass.matches(ref));
+
+    final instances = await serviceManager.service!.getInstances(
+      isolateId,
+      theClass.id!,
+      preferences.memory.refLimit.value,
+    );
+
+    return (instances.instances ?? []).firstWhereOrNull(
```
bkonyi, line 130: Nit: can the empty list be `const`?

packages/devtools_app/lib/src/shared/console/eval/eval_service.dart
```diff
@@ -108,4 +110,25 @@ class EvalService extends DisposableController with AutoDisposeControllerMixin {
...
+      (i) => i is InstanceRef && i.identityHashCode == object.code,
+    ) as InstanceRef?;
```
bkonyi, line 132: I think you can remove this cast if you explicitly define the type parameter for the empty list as `InstanceRef`.

packages/devtools_app/lib/src/shared/console/widgets/display_provider.dart
```diff
@@ -15,6 +15,25 @@ import '../../diagnostics/dart_object_node.dart';
 import '../../primitives/simple_items.dart';
 import 'description.dart';
 
+VariableSelectionControls _selectionControls({
+  required DartObjectNode variable,
+  required Function(TextSelectionDelegate delegate)? onInspect,
```
bkonyi, line 20: Nit: consider making this optional.

packages/devtools_app/lib/src/shared/diagnostics/references.dart
```diff
@@ -141,18 +183,22 @@ Future<void> addChildReferences(
...
-      final isolateRef = variable.ref!.isolateRef;
+      final isolateRef = ref.isolateRef;
+
       final instance = await getObject(
         isolateRef: isolateRef,
         value: ref.instanceRef!,
         variable: variable,
       );
 
+      // ?????
```
bkonyi, line 194: ????? indeed 😉

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -326,7 +326,7 @@ class MemoryPreferencesController extends DisposableController
   /// when browsing references in console.
   final refLimitTitle = 'Limit for number of requested live instances.';
   final refLimit = ValueNotifier<int>(_defaultRefLimit);
-  static const _defaultRefLimit = 100;
+  static const _defaultRefLimit = 100000;
```
bkonyi, line 329: Does this correspond to a maximum number of VM service objects that can be requested / tracked? If so, there's a good chance service IDs will expire before you hit this limit.

polina-c: It is parameter, passed to getInstances at the moment. It seems we will not need it as soon as we will start using getInstancesAsArray.

packages/devtools_app/lib/src/shared/console/widgets/display_provider.dart
polina-c: I want it to be explicitly specified even when it is null.

packages/devtools_app/lib/src/shared/console/eval/eval_service.dart
polina-c: nope, because instances returns list of ObjRef

packages/devtools_app/lib/src/shared/console/console_service.dart
polina-c: It is nullable for consistency with other methods in this class. May be we should refactor/review the code to deal better with nullability of IsolateRef.

bkonyi (approved):

kenzieschmoll:
DBC: we should make sure that the tree view can be traversed with arrow keys to give the best UX here. Not sure if we already support this or not, but we do support this for the tree table so we might be able to steal some of that logic to implement if the tree view does not support this already.

Issue #4069, opened 2022-05-05
Expression evaluation is only supported in the scope of the current frame


bkonyi:
As far as I can tell, it's only possible to evaluate an expression in the context of the current frame. While this is the most common use case, it should be possible to evaluate in the context of specific libraries, classes, and instances.

@rmacnak-google, feel free to add more details here if I missed anything.

kenzieschmoll:
To meet the requirements here, we may be able to add a file selector to the eval console that defaults to the root library, but allows the user to select a different library if they want to. @rmacnak-google would that support the use case described here? 

After https://github.com/flutter/devtools/pull/5248 lands, we'll have the ability to evaluate when the app is not paused, so that should get us part way there.

rmacnak-google:
No. A file selector would allow evaluating in a chosen library*, but not in a chosen class, instance or frame.

(*Almost. The relationship from files to libraries is not quite a function because of shared parts.)

Since this issue has been labeled as about the debugger, I will focus on the case of frames. More fundamentally, the console is a [broken debuggering UI](https://gbracha.blogspot.com/2008/07/debugging-visual-metaphors.html) because it makes all the per-frame operations like stepping or eval modal.

Observatory fixes this slightly making the display of source and local variables non-modal; eval is still modal but it is at least *possible* to select a frame in its console.

![image](https://user-images.githubusercontent.com/8495071/218856549-44b13c57-4329-4e7d-8cdc-68eaa3921ade.png)

Hopscotch fixes this all the way: notice that `a + b` evaluated in the second frame (and the affordence to step with respect to any frame).

![image](https://user-images.githubusercontent.com/8495071/218838871-8884c116-3c39-4fe4-b8f5-b6a4cc41190c.png)

PR #5248, opened 2023-02-14
Support evaluation for running app.

polina-c:
https://user-images.githubusercontent.com/12115586/218803466-3a7e3bf4-b58f-4bd5-9313-734fb1b7b443.mov


packages/devtools_app/lib/src/shared/console/eval/eval_service.dart
```diff
@@ -60,6 +60,19 @@ class EvalService extends DisposableController with AutoDisposeControllerMixin {
     return await _service.getObject(ref, objRef.id!);
   }
 
+  Future<Response> evalInRunningApp(
```
kenzieschmoll, line 63: Can you add some dart doc to this describing that this evaluation occurs in the isolate's root library?

packages/devtools_app/lib/src/shared/console/widgets/evaluate.dart
```diff
@@ -315,7 +308,13 @@ class ExpressionEvalFieldState extends State<ExpressionEvalField>
     try {
       // Response is either a ErrorRef, InstanceRef, or Sentinel.
       final isolateRef = serviceManager.isolateManager.selectedIsolate.value;
-      final response = await evalService.evalAtCurrentFrame(expressionText);
+      final Response response;
+      if (serviceManager.isMainIsolatePaused) {
+        response = await evalService.evalAtCurrentFrame(expressionText);
+      } else {
+        response =
+            await evalService.evalInRunningApp(isolateRef!, expressionText);
```
kenzieschmoll, line 316: should we be null asserting isolateRef? asking because _emitRefToConsole allows a nullable isolate below. Maybe we should add a null check if we cannot call the eval on a null isolate

kenzieschmoll:
Work towards https://github.com/flutter/devtools/issues/4069

packages/devtools_app/lib/src/shared/console/widgets/evaluate.dart
```diff
@@ -312,8 +312,12 @@ class ExpressionEvalFieldState extends State<ExpressionEvalField>
       if (serviceManager.isMainIsolatePaused) {
         response = await evalService.evalAtCurrentFrame(expressionText);
       } else {
+        if (isolateRef == null) {
+          _emitToConsole('Isolate should be defined for evaluation.');
```
kenzieschmoll, line 316: nit: "Cannot evaluate expression because the selected isolate is null."

kenzieschmoll (approved):
one comment on wording, but lgtm

packages/devtools_app/lib/src/shared/console/widgets/evaluate.dart
```diff
@@ -315,7 +308,19 @@ class ExpressionEvalFieldState extends State<ExpressionEvalField>
     try {
       // Response is either a ErrorRef, InstanceRef, or Sentinel.
       final isolateRef = serviceManager.isolateManager.selectedIsolate.value;
-      final response = await evalService.evalAtCurrentFrame(expressionText);
+      final Response response;
+      if (serviceManager.isMainIsolatePaused) {
```
kenzieschmoll, line 312: Is it possible for the main isolate to be paused but the selectedIsolate to be unpaused? Shouldn't we be evaluating in the selected isolate, not the main (in the event that these are different)?

PR #5249, opened 2023-02-14
Add github workflow to close stale issues

CoderDake:
This workflow will mark issues as stale if a "waiting for response" issue has not had any activity in 14 days.
14 days after an issue becomes stale, it will then be closed if there is no activity on it.

NOTE: this is being pushed up in debug mode so it can first be tested. It will then be fully rolled out in a follow up PR.

RELEASE_NOTE_EXCEPTION=Internal process only

### DIFF
--- a/.github/workflows/close-inactive-issues.yaml
+++ b/.github/workflows/close-inactive-issues.yaml
@@ -15,7 +15,7 @@ jobs:
         with:
           debug-only: true # TODO: Delete this line when we are ready to implement this for our process.
           only-issue-labels: "waiting for customer response"
-          days-before-issue-stale: 30
+          days-before-issue-stale: 14
           days-before-issue-close: 14
           stale-issue-label: "stale"
           stale-issue-message: "This issue is stale because it has been open for 30 days with no activity."
--- a/.github/workflows/close-inactive-issues.yaml
+++ b/.github/workflows/close-inactive-issues.yaml
@@ -16,11 +16,11 @@ jobs:
         with:
           debug-only: true # TODO: Delete this line when we are ready to implement this for our process.
           only-issue-labels: "waiting for customer response"
-          days-before-issue-stale: 14
+          days-before-issue-stale: -1
           days-before-issue-close: 14
           stale-issue-label: "stale"
           stale-issue-message: "This issue is stale because it has been open for 30 days with no activity."
-          close-issue-message: "This issue was closed because it has been inactive for 14 days since being marked as stale."
+          close-issue-message: "Without additional information we're unfortunately not sure how to resolve this issue. We're going to close this issue for now, but please don't hesitate to comment on the issue if you have any more information for us; we're happy to reopen. Thanks for your contribution!"
           days-before-pr-stale: -1
           days-before-pr-close: -1
           repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/close-inactive-issues.yaml
```diff
@@ -0,0 +1,26 @@
+name: Close inactive issues
+on:
+  schedule:
+    - cron: "30 1 * * *"
```
kenzieschmoll, line 4: can you add a comment describing what time this is?

.github/workflows/close-inactive-issues.yaml
```diff
@@ -0,0 +1,26 @@
...
+      # See https://github.com/actions/stale for
+      # details on the parameters for this action
+      - uses: actions/stale@v7
+        with:
+          debug-only: true # TODO: Delete this line when we are ready to implement this for our process.
+          only-issue-labels: "waiting for customer response"
+          days-before-issue-stale: 14
+          days-before-issue-close: 14
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue is stale because it has been open for 30 days with no activity."
```
kenzieschmoll, line 22: when does an issue get marked as stale?

.github/workflows/close-inactive-issues.yaml
```diff
@@ -0,0 +1,26 @@
...
+          close-issue-message: "This issue was closed because it has been inactive for 14 days since being marked as stale."
```
kenzieschmoll, line 23: should this say "since being marked as 'waiting for customer response'"?

.github/workflows/close-inactive-issues.yaml
jacob314: Can we align this with what flutter-intellij does today? I don't think it currently has a concept of "stale" but that does look like a nice touch and could lead to a few more comments with repro steps over just closing.

CoderDake: Right now it is set up to mark as stale after 14 days of inactivity, then if there are 14 more days of inactivity it will close the ticket

CoderDake: @jacob314 it looks like they wait 7 days and then just close it with a message. 
Is that what you would like?
https://github.com/flutter/flutter-intellij/issues/4762#issuecomment-688400423

kenzieschmoll: should stale-issue-message say 14 instead of 30 now?

CoderDake:
@CoderDake make sure to also remove label for "waiting for response" when someone responds

kenzieschmoll:
> @CoderDake make sure to also remove label for "waiting for response" when someone responds

We'd also want to remove "stale" when someone responds

.github/workflows/close-inactive-issues.yaml
```diff
@@ -0,0 +1,26 @@
...
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      # See https://github.com/actions/stale for
+      # details on the parameters for this action
+      - uses: actions/stale@v7
+        with:
+          debug-only: true # TODO: Delete this line when we are ready to implement this for our process.
+          stale-issue-label: "waiting for customer response"
```
CoderDake, line 18: @jacob314 @kenzieschmoll 
This still needs to be tested, but hypothetically these settings will result in the following behaviour:
- user submits issue
- we ask for more info
- we mark it with "waiting for response"
- the workflow picks up the issue as something to track, since it has the label.
- if the user responds within 14 days the the label would be removed
- if they don't then the issue will be closed with the message

kenzieschmoll: This sgtm

kenzieschmoll (approved):

PR #5258, opened 2023-02-16
Add support for viewing CPU profiler ticks per instruction

bkonyi:
This change adds a `Profiler Ticks` column to the object inspector's code viewer, allowing for compiler developers to determine how "hot" specific instructions are in generated code.

image

RELEASE_NOTE_EXCEPTION=VM developer mode functionality


packages/devtools_app/lib/src/screens/vm_developer/object_inspector/vm_object_model.dart
```diff
@@ -265,6 +265,46 @@ class CodeObject extends VmObject {
...
+  @override
+  Future<void> initialize() async {
+    await super.initialize();
+
+    final service = serviceManager.service!;
+    final isolateId = serviceManager.isolateManager.selectedIsolate.value!.id!;
+
+    // Attempt to retrieve the CPU profile data for this code object.
+    try {
+      final samples = await service.getCpuSamples(isolateId, 0, ~0);
```
kenzieschmoll, line 284: what is ~0 for?

packages/devtools_app/lib/src/screens/vm_developer/object_inspector/vm_object_model.dart
```diff
@@ -265,6 +265,46 @@ class CodeObject extends VmObject {
...
+      final codes = samples.codes;
+
+      ProfileCode? match;
```
kenzieschmoll, line 287: this can be
`ProfileCode? match = codes.firstWhereOrNull((profileCode) => profileCode.code == ref);`

packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -226,14 +226,9 @@ class FlatTableState<T> extends State<FlatTable<T>> with AutoDisposeMixin {
...
-          oldWidget.columns.map((c) => c.title),
-          newWidget.columns.map((c) => c.title),
-        ) ||
-        !collectionEquals(
-          oldWidget.columnGroups?.map((c) => c.title),
-          newWidget.columnGroups?.map((c) => c.title),
-        );
+    final columnsChanged =
+        !collectionEquals(oldWidget.columns, newWidget.columns) ||
+            !collectionEquals(oldWidget.columnGroups, newWidget.columnGroups);
```
kenzieschmoll, line 231: this may have unanticipated implications if columns and column group objects are accidentally being re-created. Then the instances would not match and this method would return true when we'd expect it to return false

packages/devtools_app/test/vm_developer/object_inspector/vm_code_display_test.dart
```diff
@@ -39,15 +39,27 @@ void main() {
...
+        for (int i = 0; i < addressCount; ++i) ...[
           (i * 4 + offset).toRadixString(16),
           'unknown',
           'noop',
           null,
         ]
       ]);
 
+      final ticksTable = CpuProfilerTicksTable.parse(
+        sampleCount: 1000,
```
kenzieschmoll, line 53: addressCount instead of 1000?

packages/devtools_app/lib/src/screens/vm_developer/object_inspector/vm_object_model.dart
bkonyi: I _thought_ it'd give me `INT_MAX`, but it gives `INT32_MAX` because JS weirdness. I should probably just use `maxJsInt` instead.

packages/devtools_app/test/vm_developer/object_inspector/vm_code_display_test.dart
bkonyi: The sample count doesn't need to match the address count as the sample count is the number of samples collected throughout the entire process, not just those collected in the context of this piece of code.

packages/devtools_app/lib/src/shared/table/table.dart
bkonyi: Hm, I can't remember exactly why I had needed this change and things seem to work properly without it, so I'll just revert it.

packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
```diff
@@ -66,7 +66,7 @@ class ProgramExplorerController extends DisposableController
   }
 
   /// Initializes the program structure.
-  void initialize() {
+  void initialize() async {
```
kenzieschmoll, line 69: this should return `Future<void>`

kenzieschmoll (approved):

PR #5270, opened 2023-02-17
Handle expiration of variables.

polina-c:

https://user-images.githubusercontent.com/12115586/219900575-b93c1fdf-1c42-47b6-bfc7-3795c1e78bc8.mov


packages/devtools_app/lib/src/shared/vm_utils.dart
```diff
@@ -2,12 +2,49 @@
...
+    preferences.memory.refLimit.value,
+  ))
+      .instances
+      ?.firstWhereOrNull(
+        (instance) =>
+            (instance is InstanceRef) &&
+            (instance.identityHashCode == hashCode),
+      );
+
+  if (result is InstanceRef) return result;
```
bkonyi, line 48: I think we can just `return result` here since we do the `is InstanceRef` check in the `firstWhereOrNull(...)` invocation above.

bkonyi (approved):

packages/devtools_app/lib/src/shared/console/primitives/scope.dart
```diff
@@ -0,0 +1,74 @@
...
+  /// Sets [refreshScopeChangeMessage] if scope changed.
+  Future<bool> refreshRefs(String isolateId) async {
+    removedVariables.clear();
+    final isolateItems = _refs[isolateId] ?? {};
+    var result = false;
+
+    final variableNames = [...isolateItems.keys];
+    for (final name in variableNames) {
+      final oldItem = isolateItems[name]!;
+      final refreshedItem = await _refreshRef(oldItem, isolateId);
```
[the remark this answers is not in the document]
polina-c: We will refresh just references to lists, not entire lists. 
I cannot test it yet, because https://github.com/dart-lang/sdk/issues/51393 is not lended yet.

In addition, yes, it is not a goal to perfectly cover every possible use case: https://docs.google.com/document/d/1M65VISBzYwiySgjKU3iM_gvcDKUzOjxGCUlQWtHOtVM/edit#bookmark=id.tmfpyez29qgr

We will do it on demand.

packages/devtools_app/lib/src/shared/vm_utils.dart
polina-c: true, but compiler does not know this and type promotion does not happen

PR #5289, opened 2023-02-21
UI polish for memory page. 

polina-c:

Explain static and live references:
Screenshot 2023-02-21 at 2 12 54 PM

Add sign for delta values:
Screenshot 2023-02-21 at 1 58 52 PM


packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
```diff
@@ -117,6 +117,14 @@ class _InstanceColumn extends ColumnData<DiffClassStats> {
     }
   }
 
+  @override
+  String getDisplayValue(DiffClassStats classStats) {
+    // Add leading sign for delta values.
+    final value = getValue(classStats);
+    if (dataPart != _DataPart.delta || value <= 0) return value.toString();
+    return '+$value';
```
kenzieschmoll, line 125: seems like we should also be adding the negative sign when the value is negative

polina-c: when value is negative, the sign is already here

packages/devtools_app/lib/src/shared/diagnostics/references.dart
```diff
@@ -69,7 +69,7 @@ Future<void> addChildReferences(
 
       variable.addAllChildren([
         DartObjectNode.references(
-          'live',
+          'live (application in its current state)',
```
kenzieschmoll, line 72: wording nit. maybe:

'live (objects currently alive in the application)"
"static (objects alive at the time of snapshot ${selection.heap.snapshotName})"

packages/devtools_app/test/test_infra/fixtures/memory_app/lib/main.dart
```diff
@@ -35,11 +35,16 @@ class MyHomePage extends StatefulWidget {
...
   final _garbage = <_MyGarbage>[];
   int _counter = 0;
+  _MyGarbage _gcableItem = _MyGarbage(0);
 
   void _incrementCounter() {
     setState(() {
       _counter++;
       _garbage.add(_MyGarbage(0));
+      if (identityHashCode(_gcableItem) < 0) {
+        print('we need this line to use the field');
```
kenzieschmoll, line 45: remove print

kenzieschmoll (approved):
could nits but lgtm

packages/devtools_app/lib/src/shared/memory/adapted_heap_data.dart
```diff
@@ -117,6 +117,8 @@ class AdaptedHeapData {
 
   late DateTime created;
 
+  String snapshotName = '';
```
kenzieschmoll, line 120: should this be late final?

polina-c: It is initialized after creation and i do not want anything to fail if it is accessed before initialization.

PR #5290, opened 2023-02-21
Add column `persisted` and saving item to console from diff.

polina-c:
Show persisted objects for diff, to enable saving them to console as variable.
 

Screenshot 2023-02-21 at 3 00 57 PM

https://user-images.githubusercontent.com/12115586/220505581-af1137cd-e66e-4e9f-be26-aee5e876b29d.mov


kenzieschmoll (commented):
Can you add test coverage for this new set we are tracking?

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
```diff
@@ -127,6 +143,41 @@ class _InstanceColumn extends ColumnData<DiffClassStats> {
...
+    DiffClassStats data, {
+    bool isRowSelected = false,
+    VoidCallback? onPressed,
+  }) {
+    if (!FeatureFlags.evalAndBrowse) return null;
+    final objects = _instances(data);
+    final theHeap = heap;
+
+    // There is no detailed information, if [dataPart] equals to [_DataPart.delta].
+    if (objects is! ObjectSet || theHeap == null) return null;
```
kenzieschmoll, line 159: why not check if `dataPart == _DataPart.delta` then if that is the case we are looking for?

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
```diff
@@ -127,6 +143,41 @@ class _InstanceColumn extends ColumnData<DiffClassStats> {
...
+        InstanceSetButton(
+          textStyle:
+              isRowSelected ? theme.selectedTextStyle : theme.regularTextStyle,
+          count: getValue(data),
+          gaContext: gac.MemoryAreas.snapshotSingle,
+          sampleObtainer: sampleObtainer,
+          showMenu: showMenu,
+        ),
+      ],
+    );
```
kenzieschmoll, line 179: this is duplicated in classes_table_single. Let's put into a helper widget so we don't have to define all this twice

packages/devtools_app/lib/src/screens/memory/shared/heap/heap.dart
```diff
@@ -215,6 +215,9 @@ class ObjectSet extends ObjectSetStats {
   static ObjectSet empty = ObjectSet()..seal();
 
   final objectsByCodes = <IdentityHashCode, AdaptedHeapObject>{};
+
+  /// Subset of objects which size is not included bu the method [countInstance]
+  /// into retained size of this set.
```
kenzieschmoll, line 220: docs nit: 
```
/// Subset of objects that are excluded from the retained size calculation for this set.
///
/// See [countInstance]
```

also, maybe rename to `objectsExcludedFromRetainedSize` to make this more self documenting

packages/devtools_app/test/test_infra/fixtures/memory_app/lib/main.dart
```diff
@@ -35,16 +35,16 @@ class MyHomePage extends StatefulWidget {
 class _MyHomePageState extends State<MyHomePage> {
   final _garbage = <_MyGarbage>[];
   int _counter = 0;
-  _MyGarbage _gcableItem = _MyGarbage(0);
+  _MyGarbage _gcableItem = _MyGarbage(0, 'Should be gced, initial.');
```
kenzieschmoll, line 38: is this memory_app used in any tests? I did a search in our codebase and didn't see any tests that use this app

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
polina-c: Because I need type promotion. Added asserts.

packages/devtools_app/test/test_infra/fixtures/memory_app/lib/main.dart
polina-c: For now I use it for manual testing. I plan to use it for integration testing in future.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
kenzieschmoll: what about using `dataPart == _DataPart.delta` for the if statement since that is what we we are truly checking for? and then using local variables to get the type promotion you need. If `objects is! ObjectSet || theHeap == null` eventually becomes true for another `_DataPart` type, then this could lead to bugs where the if passes when we wouldn't expect it to

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/instances.dart
```diff
@@ -0,0 +1,59 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: what about creating a file common.dart where we can place widgets that are used in different places of the memory screen? Not every class needs its own file like other languages

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_single.dart
```diff
@@ -105,23 +104,12 @@ class _InstanceColumn extends ColumnData<SingleClassStats>
...
-              : null,
-          showMenu: showMenu,
-        ),
-      ],
+    return InstanceTableCell(
+      data.objects,
+      heap,
+      data.heapClass,
+      isSelected: isRowSelected,
+      gaContext: gac.MemoryAreas.snapshotDiff,
```
kenzieschmoll, line 112: should this be snapshotSingle?

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/instances.dart
polina-c: 'Not every class needs its own file' does not mean 'put all small classes in one huge library'.
For me it means 'Put classes into one library if they serve shared goal and thus can be seen as one unit of functionality, and file is of reasonable size and this is observable'.

Example of such library with many classes is instance_set_button.dart

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
```diff
@@ -154,12 +154,17 @@ class _InstanceColumn extends ColumnData<DiffClassStats>
...
+    if (dataPart == _DataPart.delta) {
+      assert(theHeap == null);
+      assert(objects is! ObjectSet);
       return null;
     }
-    assert(dataPart != _DataPart.delta);
+
+    if (objects is! ObjectSet || theHeap == null) {
+      throw StateError(
+        'Other than ${_DataPart.delta} columns should have details provided.',
```
kenzieschmoll, line 165: nit: 'All columns except ${_DataPart.delta} should have objects and heap data available.'

kenzieschmoll (approved):
one wording comment then lgtm

PR #5291, opened 2023-02-21
Fix drop-box colors.

polina-c:

Before:

Screenshot 2023-02-21 at 3 07 46 PM

After:

Screenshot 2023-02-21 at 3 09 27 PM

Screenshot 2023-02-21 at 3 25 39 PM

### DIFF
--- a/packages/devtools_app/lib/src/shared/common_widgets.dart
+++ b/packages/devtools_app/lib/src/shared/common_widgets.dart
@@ -1215,6 +1215,8 @@ class RoundedDropDownButton<T> extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final bgColor = Theme.of(context).colorScheme.defaultBackgroundColor;
+
     return RoundedOutlinedBorder(
       child: Center(
         child: Container(
@@ -1231,6 +1233,7 @@ class RoundedDropDownButton<T> extends StatelessWidget {
               style: style,
               selectedItemBuilder: selectedItemBuilder,
               items: items,
+              focusColor: bgColor,
             ),
           ),
         ),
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1231,6 +1233,7 @@ class RoundedDropDownButton<T> extends StatelessWidget {
               style: style,
               selectedItemBuilder: selectedItemBuilder,
               items: items,
+              focusColor: bgColor,
```
kenzieschmoll, line 1236: Does this remove the focus effect on hover? If so, we shouldn't remove that. I originally filed a bug for this here but didn't have a minimal repro to attach: https://github.com/flutter/flutter/issues/116293

polina-c: It removes color effect, but mouse pointer changes, that looks to be enough.

Web:



https://user-images.githubusercontent.com/12115586/220485605-eb90485a-db55-4ca3-8333-376d0361a957.mov

Macos:



https://user-images.githubusercontent.com/12115586/220485628-33cc884b-07bb-4816-8d1f-86859453a7da.mov

kenzieschmoll:
thanks for confirming the hover effect still works.

I did notice that for light theme, it appears that `defaultBackgroundColor` is not the same color as the button. The selector still has two colors:
image

it is subtle, so won't block on it. But worth taking a quick look to see if we can get the right color.

kenzieschmoll (approved):
one polish request, but lgtm

polina-c:
> thanks for confirming the hover effect still works.
> 
> I did notice that for light theme, it appears that `defaultBackgroundColor` is not the same color as the button. The selector still has two colors: image
> 
> it is subtle, so won't block on it. But worth taking a quick look to see if we can get the right color.

Thanks. Fixed

PR #5297, opened 2023-02-22
Minor improvements.

polina-c:
RELEASE_NOTE_EXCEPTION=[too minor to mention]

1. Enable passing widget to column group
2. Remove tooltip for retaining path as it is too large and confusing.

### DIFF
--- a/packages/devtools_app/lib/src/shared/table/table_data.dart
+++ b/packages/devtools_app/lib/src/shared/table/table_data.dart
@@ -131,12 +131,13 @@ mixin PinnableListEntry {
 class ColumnGroup {
   ColumnGroup({required this.title, required this.range, this.tooltip});
 
-  final String title;
+  ColumnGroup.fromText(
+      {required Sting title, required String? tooltip, required this.range});
+
+  final Widget title;
 
   /// The range of column indices for columns that make up this group.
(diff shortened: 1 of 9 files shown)


packages/devtools_app/lib/src/shared/table/table_data.dart
```diff
@@ -129,14 +130,21 @@ mixin PinnableListEntry {
...
-  ColumnGroup({required this.title, required this.range, this.tooltip});
+  ColumnGroup({required this.title, required this.range});
 
-  final String title;
+  ColumnGroup.fromText({
+    required String title,
+    required Range range,
+    String? tooltip,
+  }) : this(
+          title: maybeWrapWithTooltip(child: Text(title), tooltip: tooltip),
```
kenzieschmoll, line 140: instead of adding an additional constructor just add `maybeWrapWithTooltip` to the one line change in table.dart.

polina-c: I want to be able to pass widget to the constructor to create drop down for size type (shallow or retained).
So, we need two constructors: one that takes text and another that takes widget.

kenzieschmoll (approved):

Issue #841, opened 2019-07-24
Add method table to CPU profiler


kenzieschmoll:
Observatory has a method table view for cpu profiles. The table shows the list of methods with their consumption percentages. It also shows a table of callees and callers.

![Screen Shot 2019-07-24 at 11 21 05 AM](https://user-images.githubusercontent.com/43759233/61818514-d863cf00-ae05-11e9-9ccf-7741a2e29863.png)

escamoteur:
This would be extremely helpful to detect bottlenecks. Especially such a view should also show the number a function is called and its total time that it is used.

Here it would be helpful if we got functions in dart to start stop the sampling to analyse just a segment of out App.

kenzieschmoll:
Thanks for the feedback! Tracking issue for programatic start / stop here: https://github.com/flutter/devtools/issues/947.

kenzieschmoll:
Here is the source from observatory: https://github.com/dart-lang/sdk/blob/main/runtime/observatory/lib/src/elements/cpu_profile_table.dart

PR #5303, opened 2023-02-24
Add UI and data classes for CPU method table

kenzieschmoll:
Work towards https://github.com/flutter/devtools/issues/841.

Screenshot 2023-02-23 at 4 07 13 PM

This PR creates the method table interface and prepares the controller and data classes for this feature. TODO in follow up PRs: generate the method table graph from the active CPU profile data and properly calculate the inclusive and exclusive counts.

RELEASE_NOTE_EXCEPTION=feature behind flag


packages/devtools_app/lib/src/shared/table/table_data.dart
```diff
@@ -160,3 +160,58 @@ extension ColumnDataExtension<T> on ColumnData<T> {
     }
   }
 }
+
+abstract class TimeAndPercentageColumn<T> extends ColumnData<T> {
```
bkonyi, line 164: Can you add some documentation, here and elsewhere?

packages/devtools_app/lib/src/shared/primitives/graph.dart
```diff
@@ -0,0 +1,70 @@
...
+  final successors = <GraphNode>{};
+
+  /// Maps predecessor [GraphNode]s from [predecessors] to the number of
+  /// outgoing edges going to [this] node.
+  ///
+  /// For example:
+  ///        A (predecessor node)
+  ///      /  \
+  ///     |    |
+  ///     \    /
```
bkonyi, line 21: Supernit: the bottom of these edges doesn't have the same width as the top

bkonyi (approved):
Generally LGTM, but I think we could use some more documentation.

PR #5317, opened 2023-02-27
Add support for viewing CPU profile statistics

bkonyi:
image


bkonyi:
> nit: I still think this info may be better implemented as a single row (maybe half of the table row height) at the bottom of the profiler. this info would be more discoverable that way and wouldn't add another button to the already busy toolbar.

Are we sure we want to take up that additional screen real estate?

kenzieschmoll:
> Are we sure we want to take up that additional screen real estate?

We aren't that constrained vertically on this page, and we could make this one small row (maybe 20px high?) below the profiler, so we wouldn't have to worry about hacking this into the table widget itself.

kenzieschmoll:
closing this in favor of https://github.com/flutter/devtools/pull/5340/. Thanks for the initial work here, Ben!

PR #5322, opened 2023-02-27
Add widget tree build analytics

CoderDake:
The widget tree intermittently has issues where the now rows are displayed.

When rolling out changes we want to be able to show that we are making an impact. So this PR adds analytics that:
- track when the InspectorTreeController is initialized
- track how many rows are in the tree, for each time we update the tree being shown.

Having each of these data points will allow us to see understand:
- how long each user has to wait until they see widgets show up
- If users are having issues getting any widgets to show at all, especially after the first few attempts

RELEASE_NOTE_EXCEPTION=Analytics only


packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -973,6 +976,18 @@ class _InspectorTreeState extends State<InspectorTree>
...
+      gac.inspector,
+      'InspectorTreeBuild',
+      nonInteraction: true,
+      screenMetricsProvider: () => InspectorTreeBuildMetrics(
+        isSummaryTree: widget.isSummaryTree,
+        buildCount: ++_buildCount,
+        rowCount: treeControllerLocal?.numRows,
+      ),
+    );
+
```
kenzieschmoll, line 990: I wouldn't put this in build. We aren't concerned with counting builds from the widget lifecycle as much as we are concerned with when the inspector tree data is updated and if the data is showing 0 rows then we know we have a problem.

Another thing to think about. Analytics should be tracking something that is stable over time - if code in the inspector changes in such a way that we are intentionally triggering more rebuilds, when the analytics data becomes bogus since the baseline is always changing.

packages/devtools_app/lib/src/shared/ui/gtags.dart
```diff
@@ -30,22 +30,24 @@ class GTag {
   static const String _event = 'event';
   static const String _exception = 'exception';
 
-  static bool get shouldSendAnalytics => kReleaseMode || _debugAnalytics;
+  static bool get _shouldSendAnalytics => kReleaseMode;
```
kenzieschmoll, line 33: get rid of `_shouldSendAnalytics` and use `kReleaseMode` directly now that it is a getter for one variable

packages/devtools_app/lib/src/app.dart
```diff
@@ -581,8 +581,8 @@ List<DevToolsScreen> get defaultScreens {
     DevToolsScreen<InspectorController>(
       InspectorScreen(),
       createController: (_) => InspectorController(
-        inspectorTree: InspectorTreeController(),
-        detailsTree: InspectorTreeController(),
+        inspectorTree: InspectorTreeController(gaId: 'inspectorTree'),
+        detailsTree: InspectorTreeController(gaId: 'detailsTree'),
```
kenzieschmoll, line 585: make these strings constants on the InspectorTreeControllerMetrics class since that is where they are used. Or make them an enum if they will be used in more places than just here

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -147,6 +166,17 @@ class InspectorTreeController extends Object
...
       _root = node;
       _populateSearchableCachedRows();
+
+      ga.select(
+        gac.inspector,
+        'InspectorTreeControllerRootChange',
+        nonInteraction: true,
+        screenMetricsProvider: () => InspectorTreeControllerMetrics(
+          inspectorTreeControllerId: gaId,
+          rootSetCount: ++_rootSetCount,
```
kenzieschmoll, line 176: why do we need to count how many times the root is set? If we are sending a GA event each time the root is sent, including the count seems redundant

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -147,6 +166,17 @@ class InspectorTreeController extends Object
     setState(() {
       _root = node;
       _populateSearchableCachedRows();
+
+      ga.select(
```
kenzieschmoll, line 170: how often is this `set root` method called?

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
CoderDake: this will help us distinguish how many widgets they are seeing throughout the lifecycle of the tab.
The `InspectorTreeControllerInitialized` Item will catch when the controller is first loaded.
Then each subsequent `InspectorTreeControllerRootChange` will determine if they ever get nodes, and if so, if they always continue to get nodes.

For example if we are seeing a lot of `InspectorTreeControllerInitialized` with no follow up root sets, then we know users are seeing blank screens (or else leaving the page before anythign could be loaded)

Another case could be  if rootSetCount>=1 continue to have only 1 row being shown then this could indicate pubRoot directory issues.
We could also see if they are hitting refresh a lot and continually only seeing 1 row.

Continuing to have more events after rootSetCount>=1 would let us get a better peek later in the lifecycle. Perhaps there are blips where they are shown no widgets etc.

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
CoderDake: - first time navigating to tab
- each time refresh is hit
- any time a widget is navigated to, the details tree may need to set it's root again. (if the selection is outside the current display.
- (I may have missed some others, but the above should catch the main ones :D)

CoderDake: So it should essentially cover every time we are trying to show the user a new tree :)

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
kenzieschmoll: shouldn't we be able to infer what 'root set' we are on though by the session id and the timestamp of the analytics event? (just a hypothesis, I haven't confirmed this). 

another question: are the inspector tree controllers recreated each time the user switches from the inspector tab to a different DevTools tab and then back?

packages/devtools_app/test/inspector/utils/inspector_tree.dart
```diff
@@ -12,13 +12,14 @@ import 'package:flutter_test/flutter_test.dart';
...
 ) {
-  final controller = InspectorTreeController()
-    ..config = InspectorTreeConfig(
-      summaryTree: false,
-      treeType: FlutterTreeType.widget,
-      onNodeAdded: (_, __) {},
-      onClientActiveChange: (_) {},
-    );
+  final controller =
+      InspectorTreeController(gaId: 'inspectorTreeControllerFromNode')
```
kenzieschmoll, line 16: what about making gaId optional and only making the calls into GA when gaId is not null?

CoderDake: I initially avoided it since I would be changing to code to accomodate tests. 
Since there is a second voice that wondered it I am ok with doing it then :D

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
CoderDake: I will have to check this again in the morning but I believe that this resets each time we come back to the tab.

packages/devtools_app/lib/src/shared/analytics/metrics.dart
```diff
@@ -39,3 +39,18 @@ class ProfilerScreenMetrics extends ScreenAnalyticsMetrics {
   final int cpuSampleCount;
   final int cpuStackDepth;
 }
+
+class InspectorTreeControllerMetrics extends ScreenAnalyticsMetrics {
```
kenzieschmoll, line 43: nit: rename to InspectorScreenMetrics so that this can be a general class for all inspector custom metrics.

make parameters optional, and add dart doc for each field.

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
CoderDake: I did a double check and the controller continues to exist even when we navigate away. So the count is representative of how many times we have set the root over the lifetime of the controller, not necessarily the lifetime of the tab.


For some added detail, i had a chat with Kenzie to describe what the `rootSetCount` will help with. The count allows us to understand how many rows we are showing to the user at different times of the controller lifecycle. 
We discussed how it would still be possible to get this information without the count, but having the count will make our queries much simpler. So we decided that it is ok to keep.

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -147,6 +166,17 @@ class InspectorTreeController extends Object
...
+        gac.inspector,
+        'InspectorTreeControllerRootChange',
```
kenzieschmoll, line 172: make this a constant in the analytics constants file

packages/devtools_app/lib/src/shared/analytics/metrics.dart
```diff
@@ -39,3 +39,26 @@ class ProfilerScreenMetrics extends ScreenAnalyticsMetrics {
...
 }
+
+class InspectorScreenMetrics extends ScreenAnalyticsMetrics {
+  InspectorScreenMetrics({
+    required this.rootSetCount,
+    required this.rowCount,
+    required this.inspectorTreeControllerId,
+  });
+
+  static const String inspectorTreeGaId = 'inspectorTree';
```
kenzieschmoll, line 50: nit: rename to summaryTreeGaId and 'summaryTree'

kenzieschmoll (approved):
LGTM after a few nits are addressed

packages/devtools_app/lib/src/shared/analytics/_analytics_web.dart
```diff
@@ -104,6 +104,9 @@ class GtagEventDevTools extends GtagEvent {
     int? heap_diff_objects_before, // metric7
     int? heap_diff_objects_after, // metric8
     int? heap_objects_total, // metric9
+    int? root_set_count, // metric10
+    int? row_count, // metric11
+    String? inspector_tree_controller_id, // metric12
```
kenzieschmoll, line 109: Screenshot 2023-03-08 at 11 05 46 AM

metrics can only be Integers, doubles, or time. We have two options here. We can make this an int: 0 for summary and 1 for details tree (this may be the easiest option)? Or we can make custom screen dimensions like we have made custom screen metrics, and make the inspector tree id a custom dimension

PR #5330, opened 2023-02-28
Clean up next release notes.

polina-c:


packages/devtools_app/release_notes/NEXT_RELEASE_NOTES.md
```diff
@@ -1,12 +1,12 @@
 This is draft for future release notes, that are going to land on
 [the Flutter website](https://docs.flutter.dev/development/tools/devtools/release-notes).
 
-# DevTools 2.22.0 release notes
+# DevTools 2.23.0 release notes
```
CoderDake, line 4: when you run `dart update_version.dart auto -t minor`
The release notes will automatically be cleaned up for you.
I believe the next release notes helper also does this automatically

polina-c: Do you suggest to remove this header or cancel this chage?

CoderDake: I think this should be fine actually, but we may not want to push this until our version is set to 2.23.0 as well :)

CoderDake (approved):

packages/devtools_app/release_notes/NEXT_RELEASE_NOTES.md
polina-c: May be we should delete the header at all. In some next PR.

Issue #4658, opened 2022-10-28
Use Flutter Web's default service worker and don't download Canvaskit ourselves


elliette:
The main reason we wrote our own service worker was due to Flutter Web downloading `canvaskit.wasm` from a CDN:

> The problem is that the Flutter service worker does not cache `canvaskit.wasm`. This is not ideal, because canvaskit.wasm is the second largest asset that we load. Flutter does not include canvaskit in its service worker because by default canvaskit is loaded from a remote https://unpkg.com URL, and is not bundled as part of the app’s binary. 

In Flutter Web, Canvaskit is now bundled with the app, and not downloaded from a CDN (see https://github.com/flutter/flutter/issues/60069#issuecomment-994103190). This means that it should be cached by the [service worker generated by Flutter Web](https://github.com/flutter/flutter/blob/a076fee040389040a852ae0db84be546938dc264/packages/flutter_tools/lib/src/build_system/targets/web.dart#L499).

I believe we should now be able to remove the step to [download CanvasKit](https://github.com/flutter/devtools/blob/9f56d7fcee71f3078ce77accc48f2f7eeceb66f3/tool/build_release.sh#L29-L54) from our release build, as well as delete our service worker and use the one generated by Flutter Web instead.

elliette:
This is fixed now that the changes to support this in g3 has also landed.

Issue #5148, opened 2023-01-31
Remove the FLUTTER_WEB_CANVASKIT_URL flag from our build script


kenzieschmoll:
From @mdebbar:

> The --web-renderer=canvaskit should replace the FLUTTER_WEB_USE_SKIA dart define. 
> 
> And then FLUTTER_WEB_CANVASKIT_URL should be replaced with a runtime config: 
> 
> * https://docs.flutter.dev/development/platform-integration/web/renderers#runtime-configuration
> * https://docs.flutter.dev/development/platform-integration/web/initialization (look for canvasKitBaseUrl)

PR #5331, opened 2023-03-01
Use the default Flutter service worker instead of our custom service worker

elliette:
Work towards https://github.com/flutter/devtools/issues/3896, https://github.com/flutter/devtools/issues/4658

Use the default Flutter service worker generated by Flutter Web instead of our own. 

* Changes to `index.html` were copied from the default `index.html` generated by Flutter when you run `flutter create`
* `pwa-strategy` changed from `none` to `offline-first` so that Flutter generates a service worker

Confirmed that the new service worker works by running `dart run tool/build_e2e.dart`. Need to also update the copybara script before submitting.


packages/devtools_test/lib/src/mocks/fake_vm_service_wrapper.dart
```diff
@@ -403,8 +403,7 @@ class FakeVmServiceWrapper extends Fake implements VmServiceWrapper {
   ) async {
     final httpProfile = await getHttpProfile(isolateId);
     return Future.value(
-      httpProfile.requests
-          .firstWhereOrNull((request) => request.id == id),
+      httpProfile.requests.firstWhereOrNull((request) => request.id == id),
```
elliette, line 406: This change was due to running `dart format` from the root

polina-c (approved):

packages/devtools_app/web/index.html
```diff
@@ -45,88 +45,32 @@
...
+    window.addEventListener('load', function(ev) {
+      // Download main.dart.js
+      _flutter.loader.loadEntrypoint({
+        serviceWorker: {
+          serviceWorkerVersion: serviceWorkerVersion,
+        },
+        onEntrypointLoaded: function(engineInitializer) {
+          engineInitializer.initializeEngine().then(function(appRunner) {
+            appRunner.runApp();
+          });
```
kenzieschmoll, line 70: can you set the config to use canvas kit here now that we are using the default service worker?
https://docs.flutter.dev/development/platform-integration/web/renderers#runtime-configuration.

This would also allow us to remove `--dart-define=FLUTTER_WEB_CANVASKIT_URL=canvaskit/ \` from `build_release.sh` below.

Here is the devtools issue for that: https://github.com/flutter/devtools/issues/5148

packages/devtools_app/web/index.html
```diff
@@ -45,88 +45,36 @@
...
+    window.addEventListener('load', function(ev) {
+      // Download main.dart.js
+      _flutter.loader.loadEntrypoint({
+        serviceWorker: {
+          serviceWorkerVersion: serviceWorkerVersion,
+        },
+        onEntrypointLoaded: function(engineInitializer) {
+          engineInitializer.initializeEngine({
+            renderer: 'canvaskit', 
+            canvasKitBaseUrl: 'canvaskit'
```
kenzieschmoll, line 70: curious, does the call to initializeEngine work without setting the base url? In the example on the flutter website, I only saw "renderer" specified, so I'm wondering why we need this too

kenzieschmoll (approved):
LGTM

Just making sure - have we verified that canvas kit is properly loaded with the new changes?

packages/devtools_app/web/index.html
elliette: Done!

packages/devtools_app/web/index.html
elliette: Yep, looks like this is required otherwise it tries to download canvaskit from the CDN, which is surprising. Opened https://github.com/flutter/flutter/issues/121743 in Flutter

elliette:
> LGTM
> 
> Just making sure - have we verified that canvas kit is properly loaded with the new changes?

Yep, all is working (even with no internet connection)

Issue #4962, opened 2022-12-16
Memory: enable eval and browse


polina-c:
The features are needed to retire observatory.

Related issues:
eval: https://github.com/flutter/devtools/issues/4191
browse: https://github.com/flutter/devtools/issues/4757

polina-c:
- [ ] http://go/devtools-memory-eval-and-browse-design

annagrin:
Related: conversations about full repl support (in the debugger and outside) https://github.com/dart-lang/webdev/issues/1480

We don't have a solution yet but do have ideas. Let us know if this becomes a priority (changes would need to be done in CFE, ddc, vm?, dwds, devtools, and also probably setup tools like flutter tools and wevdev)

PR #5333, opened 2023-03-01
Support closures and records in outbound live references.

polina-c:

https://github.com/flutter/devtools/issues/4962

https://user-images.githubusercontent.com/12115586/222518390-fdaa5123-e113-4d43-8a59-4877cc16f652.mov


packages/devtools_app/lib/src/shared/diagnostics/dart_object_node.dart
```diff
@@ -273,23 +274,31 @@ class DartObjectNode extends TreeNode<DartObjectNode> {
 
   @override
   bool get isExpandable {
+    final theRef = ref;
```
elliette, line 277: I don't think we need to assign this to a new variable

packages/devtools_app/lib/src/shared/diagnostics/dart_object_node.dart
```diff
@@ -273,23 +274,31 @@ class DartObjectNode extends TreeNode<DartObjectNode> {
...
+    final instanceRef = theRef?.instanceRef;
+    if (theRef is ObjectReferences) {
+      if (theRef.instanceRef?.length == 0) return false;
```
elliette, line 280: Could we use `instanceRef` here (instead of `theRef.instanceRef`)?

elliette (approved):
LGTM with a couple comments

packages/devtools_app/lib/src/shared/diagnostics/dart_object_node.dart
polina-c: I need type promotion to access theRef.refNodeType little later

elliette: Ah got it. Optional suggestion:

```
if (ref is ObjectReferences) {
  objectRef = ref as ObjectReferences;
  ... 
  return objectRef.refNodeType...
}

PR #5338, opened 2023-03-01
Isolate runDevTools so that it can accessed internally and externally.

CoderDake:
While trying to update main.dart I realized that the google3 entry point did not receive my changes.

This refactor isolates [runDevTools] so that each entry point can run it. This will avoid future confusion around keeping both main.dart files in sync.

RELEASE_NOTE_EXCEPTION=Internal refactor only

### DIFF
--- a/packages/devtools_app/lib/initialization.dart
+++ b/packages/devtools_app/lib/initialization.dart
@@ -21,7 +21,7 @@ import 'src/shared/primitives/utils.dart';
 /// Any initialization that needs to happen before running DevTools, regardless
 /// of context, should happen here.
 ///
-/// If the intialization is specific to running Devtools internally or
+/// If the intialization is specific to running Devtools in google3 or
 /// externally, then it should be added to that respective main.dart file.
 Future<void> runDevTools({
   bool shouldEnableExperiments = false,
--- a/packages/devtools_app/lib/initialization.dart
+++ b/packages/devtools_app/lib/initialization.dart
@@ -21,9 +21,8 @@ import 'src/shared/primitives/utils.dart';
 /// Any initialization that needs to happen before running DevTools, regardless
 /// of context, should happen here.
 ///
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/initialization.dart
```diff
@@ -0,0 +1,91 @@
+import 'package:flutter/material.dart';
```
kenzieschmoll, line 1: copyright 2023

packages/devtools_app/lib/main.dart
```diff
@@ -2,93 +2,33 @@
...
 void main() async {
-  await runDevTools();
+  await externalRunDevTools();
 }
 
-Future<void> runDevTools({
+Future<void> externalRunDevTools({
   bool shouldEnableExperiments = false,
   List<DevToolsJsonFile> sampleData = const [],
+  List<DevToolsScreen>? screens,
```
kenzieschmoll, line 24: remove the screens param here since we will never use it externally

packages/devtools_app/lib/main.dart
```diff
@@ -2,93 +2,33 @@
...
-import 'src/shared/config_specific/ide_theme/ide_theme.dart';
-import 'src/shared/config_specific/url/url.dart';
-import 'src/shared/config_specific/url_strategy/url_strategy.dart';
-import 'src/shared/feature_flags.dart';
 import 'src/shared/globals.dart';
-import 'src/shared/preferences.dart';
-import 'src/shared/primitives/url_utils.dart';
 import 'src/shared/primitives/utils.dart';
 
+/// This is the entrypoint for running DevTools normally.
```
kenzieschmoll, line 12: r/normally/externally. or "running DevTools in a non-google3 context"

kenzieschmoll (approved):
couple comments but lgtm

PR #5342, opened 2023-03-02
Fix heap for columns.

polina-c:

Fix issue of picking wrong heap snapshot, by moving state from non-static columns to controller.


packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
```diff
@@ -128,6 +126,9 @@ class _InstanceColumn extends ColumnData<DiffClassStats>
     }
   }
 
+  static late HeapDataObtainer before;
+  static late HeapDataObtainer after;
```
kenzieschmoll, line 130: instead of storing data objects on the ColumnData classes, store them in a controller that the column methods can look the data up from. We should avoid storing data UI components to avoid state management issues. Same for all columns below

packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -80,6 +80,10 @@ class FlatTable<T> extends StatefulWidget {
   })  : selectionNotifier = selectionNotifier ?? ValueNotifier<T?>(null),
         super(key: key);
 
+  /// List of columns to render.
+  ///
+  /// If different column instances are provided for two instances of the table
+  /// with the same [dataKey], the table will render first instances.
```
kenzieschmoll, line 86: This isn't true. `dataKey` is used to lookup the `_TableUiState` for the table, which stores sort column index, sort direction, and scroll offset.

packages/devtools_app/lib/src/shared/diagnostics/references.dart
```diff
@@ -68,14 +68,15 @@ Future<void> addChildReferences(
           _refreshStaticSelection(ref.heapSelection, ref.instanceRef);
 
       variable.addAllChildren([
-        DartObjectNode.references(
-          'live (objects currently alive in the application)',
-          ObjectReferences.copyWith(
-            ref,
-            refNodeType: RefNodeType.liveRefRoot,
-            heapSelection: selection,
+        if (ref.instanceRef != null)
```
kenzieschmoll, line 71: why do we check that `ref.instanceRef != null` here but not below for static references? similar question is why are the if checks difference for static and live?

packages/devtools_app/lib/src/shared/table/table.dart
polina-c: It is what I observed. The invoked method 'build' belongs to first column of the type, not to column, provided to the table. That is not a problem if the column does not contain state.

packages/devtools_app/lib/src/shared/diagnostics/dart_object_node.dart
```diff
@@ -276,13 +276,9 @@ class DartObjectNode extends TreeNode<DartObjectNode> {
...
-    if (theRef is ObjectReferences) {
-      if (instanceRef?.length == 0) return false;
-      return theRef.refNodeType.isRoot ||
-          children.isNotEmpty ||
-          childCount > 0 ||
-          !isPrimitiveInstanceKind(instanceRef?.kind);
-    }
+
+    if (isRootForReferences(ref)) return true;
+
```
kenzieschmoll, line 281: there is a logic change here - is it intentional? We used to return early from `isExpandable` for the ObjectReferences case, and now we do not. Now we only return if true, and if false, we continue through the rest of the isExpandable getter. 

Perhaps instead, this should be 
```
if (theRef is ObjectReferences) {
  return isRootForReferences(ref);
}
```

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -146,6 +144,28 @@ class DiffPaneController extends DisposableController {
       ),
     );
   }
+
+  ClassFilterButton _classFilterButton() => ClassFilterButton(
+        filter: core.classFilter,
+        onChanged: applyFilter,
+        rootPackage: serviceManager.rootInfoNow().package,
+      );
```
kenzieschmoll, line 152: this is a bit of a strange pattern. We shouldn't have controller classes be responsible for creating widgets. Instead, pass a `diffPaneController` parameter to ClassFilterButton

packages/devtools_app/lib/src/shared/diagnostics/references.dart
polina-c: Because we use different mechanisms to show static and live references.
It can be that we can show static, but not live and vise versa.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/simple_controllers.dart
```diff
@@ -4,7 +4,48 @@
...
+    required this.totalHeapSize,
+    required this.filterButton,
+  });
+
+  // We use functions, not [ValueListener], where we do not want widgets
+  // to subscribe for the changes, for performance reasons.
+
+  final HeapDataObtainer heap;
+  final int Function() totalHeapSize;
+  final Widget filterButton;
```
kenzieschmoll, line 30: we shouldn't be passing widgets to controller code. Widgets should only be created as part of the widget lifecycle.

packages/devtools_app/lib/src/shared/diagnostics/dart_object_node.dart
polina-c: Yes, I found bug: we do not expand item, if there is no live reference, while we want to show static references from snapshot in this case.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
```diff
@@ -79,15 +79,15 @@ class DiffClassNameColumn extends ColumnData<DiffClassStats>
       mainAxisAlignment: MainAxisAlignment.spaceBetween,
       children: [
         Expanded(child: defaultHeaderRenderer()),
-        classFilterButton,
+        controller.classFilterButton,
```
kenzieschmoll, line 82: create the ClassFilterButton here directly so that the widget is created during the widget lifecycle (since buildHeader is called during the build of the Table)

packages/devtools_app/lib/src/shared/table/table.dart
kenzieschmoll: IIRC what we discussed last week, what you were seeing was a state management bug with when the columns were being defined and updated, but this doesn't have anything to do with the `dataKey` since this `dataKey` is only used to lookup the values I mentioned in my comment above. So we should remove this comment.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
polina-c: replaced with controller, thanks

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/simple_controllers.dart
```diff
@@ -14,20 +15,32 @@ class RetainingPathController {
   final invert = ValueNotifier<bool>(true);
 }
 
+class ClassFilterController {
```
kenzieschmoll, line 18: This isn't really a controller, but rather is a simple data structure. I will push back here and suggest that this class isn't necessary at all. IMO this is overcomplicating something that could easily be declarative when the ClassFilterButton widget is created:

```
ClassFilterButton(
  filter: controller.core.classFilter,
  onChanged: controller.applyFilter,
  rootPackage: serviceManager.rootInfoNow().package,
);
```

This is much more terse and is easier to follow.

packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -82,8 +82,11 @@ class FlatTable<T> extends StatefulWidget {
 
   /// List of columns to render.
   ///
-  /// If different column instances are provided for two instances of the table
-  /// with the same [dataKey], the table will render first instances.
+  /// If different column instances are provided to different instances of the table
+  /// all table instances will render first instance of the provided column type.
+  ///
+  /// So, if you have many instances of a concrete table, it is recommended to
+  /// make columns static and share between table instances.
```
kenzieschmoll, line 89: Please update this comment to the following. 

These [ColumnData] elements should be defined as static OR if they cannot be defined as static, they should not manage stateful data. [FlatTableState.didUpdateWidget] checks if the columns have changed before re-initializing the table controller, and the columns are compared by title only.

See also [FlatTableState. _tableConfigurationChanged].

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
```diff
@@ -336,27 +323,26 @@ class ClassesTableDiff extends StatelessWidget {
...
         // no matter what the data passed to it is.
         const dataKey = 'ClassesTableDiff';
-        final columns = _ClassesTableDiffColumns(
-          classFilterButton,
+
+        final columns = _columns.putIfAbsent(
           sizeType,
-          before: before,
-          after: after,
+          () => ClassesTableDiffColumns(sizeType, controller),
```
kenzieschmoll, line 345: define these outside of build as a top level field on the class.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_single.dart
```diff
@@ -138,90 +139,78 @@ class _ShallowSizeColumn extends ColumnData<SingleClassStats> {
...
-    _ClassNameColumn(classFilterButton),
-    _InstanceColumn(heap),
+    _ClassNameColumn(controller),
+    _InstanceColumn(controller),
     _ShallowSizeColumn(),
     retainedSizeColumn,
   ];
 }
 
+typedef HeapSizeObtainer = int Function();
```
kenzieschmoll, line 185: this appears to be unused in this PR

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_single.dart
```diff
@@ -138,90 +139,78 @@ class _ShallowSizeColumn extends ColumnData<SingleClassStats> {
...
     // We want to preserve the sorting and sort directions for ClassesTableDiff
     // no matter what the data passed to it is.
     const dataKey = 'ClassesTableSingle';
-    final columns = _ClassesTableSingleColumns(
-      totalSize,
-      classFilterButton,
-      heap,
-    );
+
+    final columns = _columns ??= _ClassesTableSingleColumns(controller);
```
kenzieschmoll, line 206: don't define in build. Same comment as above.

packages/devtools_app/lib/src/shared/memory/adapted_heap_data.dart
```diff
@@ -56,6 +56,8 @@ class HeapObjectSelection {
   }
 }
 
+typedef HeapDataObtainer = AdaptedHeapData Function();
```
kenzieschmoll, line 59: nit: rename to HeapDataCallback? "Obtainer" makes it sound like this is a Dart Class and not a function typedef

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -146,6 +143,28 @@ class DiffPaneController extends DisposableController {
...
+
+  late final _classFilterController = ClassFilterController(
+    filter: core.classFilter,
+    onChanged: applyFilter,
+    rootPackage: serviceManager.rootInfoNow().package,
+  );
+
+  late final ClassesTableSingleController classesTableSingleController =
+      ClassesTableSingleController(
+    selection: derived.selectedSingleClassStats,
```
kenzieschmoll, line 155: instead of passing in the selection notifier, define this notifier directly in the `ClassesTableSingleController` class. That way we aren't duplicating the notifiers and it can only be modified through the `ClassesTableSingleController` class.

 Same comment for `ClassesTableDiffController`'s selection notifier as well

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/simple_controllers.dart
polina-c: These three data items are carried together through places, where we do not want widgets to be created (DiffPaneController). So, we need to bundle them together. But, I agree, they are not controllers. 

How about renaming the class to ClassFilterData?

polina-c: Discussed offline.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/class_data.dart
```diff
@@ -0,0 +1,62 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: 2023

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/class_data.dart
```diff
@@ -0,0 +1,62 @@
...
+
+import 'package:flutter/foundation.dart';
+
+import '../../../../../shared/memory/adapted_heap_data.dart';
+import '../../../shared/heap/class_filter.dart';
+import '../../../shared/heap/heap.dart';
+import '../../../shared/primitives/simple_elements.dart';
+import 'heap_diff.dart';
+
+class RetainingPathController {
```
kenzieschmoll, line 13: nit: rename to RetainingPathData

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/class_data.dart
```diff
@@ -0,0 +1,62 @@
...
+    required this.totalHeapSize,
+    required this.filterController,
+  });
+
+  // We use functions, not [ValueListener], where we do not want widgets
+  // to subscribe for the changes, for performance reasons.
+
+  final HeapDataCallback heap;
+  final int Function() totalHeapSize;
+  final ClassFilterData filterController;
```
kenzieschmoll, line 42: r/filterController/filterData

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/class_data.dart
```diff
@@ -0,0 +1,62 @@
...
+  });
+
+  final selectedSizeType = ValueNotifier<SizeType>(SizeType.retained);
+
+  // We use functions, not [ValueListener], where we do not want widgets
+  // to subscribe for the changes, for performance reasons.
+
+  final HeapDataCallback before;
+  final HeapDataCallback after;
+  final ClassFilterData filterController;
```
kenzieschmoll, line 60: same here. filterData

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/class_data.dart
```diff
@@ -0,0 +1,62 @@
...
+class ClassesTableDiffData {
+  ClassesTableDiffData({
+    required this.before,
+    required this.after,
+    required this.filterController,
+  });
+
+  final selectedSizeType = ValueNotifier<SizeType>(SizeType.retained);
+
+  // We use functions, not [ValueListener], where we do not want widgets
```
kenzieschmoll, line 55: nit: here and above r/ValueListener/ValueListenable

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -184,13 +175,32 @@ class DerivedData extends DisposableController with AutoDisposeControllerMixin {
   DerivedData(this._core) {
     _selectedItem = ValueNotifier<SnapshotItem>(_core.selectedItem);
 
+    final classFilterData = ClassFilterData(
+      filter: _core.classFilter,
+      onChanged: applyFilter,
+      rootPackage: serviceManager.rootInfoNow().package,
```
kenzieschmoll, line 181: will this value ever change throughout the lifetime of an app connection?

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/class_data.dart
```diff
@@ -0,0 +1,62 @@
...
+    required this.rootPackage,
+  });
+
+  final ValueListenable<ClassFilter> filter;
+  final ApplyFilterCallback onChanged;
+  final String? rootPackage;
+}
+
+class ClassesTableSingleData {
+  ClassesTableSingleData({
```
kenzieschmoll, line 31: can you add dart doc to members on this class as well as ClassesTableDiffData

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/class_filter.dart
```diff
@@ -15,27 +14,24 @@ import '../../../../../shared/memory/class_name.dart';
...
-    required this.filter,
-    required this.onChanged,
-    required String? rootPackage,
-  }) : rootPackage = _adaptRootPackageForFilter(rootPackage);
+  ClassFilterButton(this.controller)
+      : rootPackage = _adaptRootPackageForFilter(controller.rootPackage);
 
-  final ValueListenable<ClassFilter> filter;
-  final Function(ClassFilter) onChanged;
+  final ClassFilterData controller;
```
kenzieschmoll, line 28: rename to filterData

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
```diff
@@ -28,29 +29,29 @@ enum _DataPart {
...
+  DiffClassNameColumn(this.controller)
       : super(
           'Class',
           titleTooltip: 'Class name',
           fixedWidthPx: scaleByFontFactor(200.0),
           alignment: ColumnAlignment.left,
         );
 
-  final Widget classFilterButton;
+  final ClassesTableDiffData controller;
```
kenzieschmoll, line 44: rename to diffData

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
```diff
@@ -28,29 +29,29 @@ enum _DataPart {
   persisted,
 }
 
-class _ClassNameColumn extends ColumnData<DiffClassStats>
+class DiffClassNameColumn extends ColumnData<DiffClassStats>
```
kenzieschmoll, line 32: can this be private?

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
```diff
@@ -96,7 +97,7 @@ class _InstanceColumn extends ColumnData<DiffClassStats>
 
   final _DataPart dataPart;
 
-  final AdaptedHeapData? heap;
+  final ClassesTableDiffData controller;
```
kenzieschmoll, line 100: rename to diffData here and elsewhere

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
```diff
@@ -148,23 +149,24 @@ class _InstanceColumn extends ColumnData<DiffClassStats>
...
-    final theHeap = heap;
 
     if (dataPart == _DataPart.delta) {
-      assert(theHeap == null);
       assert(objects is! ObjectSet);
       return null;
     }
 
-    if (objects is! ObjectSet || theHeap == null) {
+    final heapObtainer =
```
kenzieschmoll, line 158: nit: rename to heapDataCallback to match the typedef

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_diff.dart
```diff
@@ -238,44 +240,36 @@ class _SizeColumn extends ColumnData<DiffClassStats> {
...
+  ClassesTableDiffColumns(this.sizeType, this.controller);
 
-  final Widget classFilterButton;
-
-  late final sizeDeltaColumn = _SizeColumn(_DataPart.delta, sizeTypeToShow);
-
-  final AdaptedHeapData before;
-  final AdaptedHeapData after;
+  final SizeType sizeType;
+  final ClassesTableDiffData controller;
```
kenzieschmoll, line 247: rename to diffData

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_single.dart
```diff
@@ -17,25 +16,27 @@ import '../../../../../shared/utils.dart';
...
+  _ClassNameColumn(this.controller)
       : super(
           'Class',
           titleTooltip: 'Class name',
           fixedWidthPx: scaleByFontFactor(200.0),
           alignment: ColumnAlignment.left,
         );
 
-  final Widget classFilterButton;
+  final ClassesTableSingleData controller;
```
kenzieschmoll, line 35: rename to `data` or similar. here and elsewhere

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/class_data.dart
polina-c: I think this one is controller, because it really controls how we show retaining path. User choose what details they want and we rebuild the widget.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
polina-c: nope. removed from constructor

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_single.dart
```diff
@@ -72,23 +72,23 @@ class _ClassNameColumn extends ColumnData<SingleClassStats>
...
+  _InstanceColumn(this.classedData)
       : super(
           'Instances',
           titleTooltip: nonGcableInstancesColumnTooltip,
           fixedWidthPx: scaleByFontFactor(180.0),
           alignment: ColumnAlignment.right,
         );
 
-  final ClassesTableSingleData controller;
+  final ClassesTableSingleData classedData;
```
kenzieschmoll, line 91: typo? is this supposed to be "classesData" instead of "classedData"?

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_single.dart
```diff
@@ -138,95 +139,79 @@ class _ShallowSizeColumn extends ColumnData<SingleClassStats> {
...
   @override
-  String getDisplayValue(SingleClassStats classStats) {
-    final value = getValue(classStats);
+  String getDisplayValue(SingleClassStats dataObject) {
+    final value = getValue(dataObject);
 
     final bytes = prettyPrintRetainedSize(value)!;
 
-    final percents = '${(value * 100 / totalSize).round()}%';
+    final percents = '${(value * 100 / classedData.totalHeapSize()).round()}%';
```
kenzieschmoll, line 164: use the `percent2` helper from utils.dart. You can change change this helper method to take a `fractionDigits` parameter that defaults to 2 so that it can be modified if you want 0 digits here.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_single.dart
```diff
@@ -138,95 +139,79 @@ class _ShallowSizeColumn extends ColumnData<SingleClassStats> {
...
-  /// Is needed to calculate percentage.
-  final int totalSize;
+  late final retainedSizeColumn = _RetainedSizeColumn(classedData);
 
-  final Widget classFilterButton;
-
-  late final retainedSizeColumn = _RetainedSizeColumn(totalSize);
-
-  final AdaptedHeapData heap;
+  final ClassesTableSingleData classedData;
```
kenzieschmoll, line 175: same typo question here: classesData?

packages/devtools_app/test/memory/diff/widgets/class_filter_test.dart
```diff
@@ -39,63 +33,81 @@ final _tests = [
...
+  except: '',
+  only: '',
+);
+
 void main() {
-  late DiffSnapshotScene scene;
+  final scene = DiffSnapshotScene();
+  // The setup should happen one time, so that there is only one instance of
+  // DiffPaneController accross tests.
+  // Otherwize tests may exchange controllers.
```
kenzieschmoll, line 46: can you add a TODO to debug this further? This is a bug with how we are managing state that we should fix

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/classes_table_single.dart
```diff
@@ -161,22 +161,23 @@ class _RetainedSizeColumn extends ColumnData<SingleClassStats> {
 
     final bytes = prettyPrintRetainedSize(value)!;
 
-    final percents = '${(value * 100 / classedData.totalHeapSize()).round()}%';
+    final percents =
+        percent2(value / classData.totalHeapSize(), fractionDigits: 0);
```
kenzieschmoll, line 165: rename from percent2 to percent now that the 2 part is configurable

kenzieschmoll (approved):

Issue #5078, opened 2023-01-20
Bottom Up calculation for CPU profiler is wrong


kenzieschmoll:
Recreate the example here and verify that our bottom up and call tree representations match what the android studio profilers show: https://youtu.be/v4kCRZ_O4Lc?t=1085

kenzieschmoll:
I recreated the example from the linked Android Studio profiler video:
![Screenshot 2023-02-10 at 11 06 51 AM](https://user-images.githubusercontent.com/43759233/218176849-d4567834-c574-4044-9dff-873c3615a533.png)
![Screenshot 2023-02-10 at 11 08 08 AM](https://user-images.githubusercontent.com/43759233/218176963-23740aca-f1ee-4815-a5c8-cb50c7691f54.png)

Here is DevTools with same profile:

**Top Down (matches Android Studio)**
![Screenshot 2023-02-10 at 11 04 06 AM](https://user-images.githubusercontent.com/43759233/218176390-5804f1d8-27ae-46a4-bce3-a70d0614a3b0.png)
![Screenshot 2023-02-10 at 11 04 16 AM](https://user-images.githubusercontent.com/43759233/218176409-531c53a8-c475-4cbd-8bc3-748f9979ecc3.png)

**Bottom Up (does not match Android Studio)**
![Screenshot 2023-02-10 at 11 03 57 AM](https://user-images.githubusercontent.com/43759233/218176475-096ea8d4-aa8e-41b0-a9d0-8d20210283bb.png)

The difference in Android Studio's algorithm is described at this time stamp in the video ([link](https://youtu.be/v4kCRZ_O4Lc?t=1059)), when the speaker describes the calculation for B. 

CC @scheglov @bkonyi

kenzieschmoll:
The Chrome DevTools CPU profiler also has slightly different results for the same data:
![Screenshot 2023-02-10 at 4 07 17 PM](https://user-images.githubusercontent.com/43759233/218224654-c7390a55-8ec1-4b77-b33f-2d5c223281b6.png)
![Screenshot 2023-02-10 at 4 07 38 PM](https://user-images.githubusercontent.com/43759233/218224660-938d71cc-e416-4e63-94b1-4cf6352e53a0.png)

The Chrome DevTools bottom up table does not create a bottom up root for node D (same as the Dart DevTools CPU profiler but different than the Android Studio profiler). I find it confusing that AS creates a bottom up root for node D, when no samples actually end with D.

[...]

PR #5348, opened 2023-03-03
Fix error in bottom up timing calculation

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/5078.

Before (does not match Android Studio or Chrome DevTools):
image

After (does match Android Studio and Chrome DevTools calculation - see roots `B` and `A`):
Screenshot 2023-03-03 at 10 33 11 AM

This PR also polishes the tooltips to explain this calculation.

Bottom up root **self** time
Screenshot 2023-03-03 at 10 17 56 AM
Bottom up root **total** time
Screenshot 2023-03-03 at 10 18 04 AM

Bottom up child **self** time (total is the same thing with "Total" instead of "Self"):
Screenshot 2023-03-03 at 10 17 39 AM


packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -211,7 +212,7 @@ class CpuProfileData {
     // Initialize all CPU samples.
     final stackTraceEvents =
         (json[traceEventsKey] ?? []).cast<Map<String, dynamic>>();
-    final samples = stackTraceEvents
+    final List<CpuSampleEvent> samples = stackTraceEvents
```
bkonyi, line 215: Nit: any reason we added a type here?

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -1012,8 +1012,7 @@ class CpuStackFrame extends TreeNode<CpuStackFrame>
     if (copySampleCounts) {
       copy
         ..exclusiveSampleCount = exclusiveSampleCount
-        ..inclusiveSampleCount =
-            resetInclusiveSampleCount ? null : inclusiveSampleCount;
+        ..inclusiveSampleCount = inclusiveSampleCount;
```
bkonyi, line 1015: I swear I wanted to make this change before and we decided it wasn't correct... 😂 😭

bkonyi (approved):

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
kenzieschmoll: I think so too 😆  except this change needed to be made in combination with the change in `generateBottomUpRoots` to set `currentBottomUpRoot.inclusiveSampleCount = null` when `node.exclusiveSampleCount == 0`. After this PR, we should finally have this calculation right 😅

Issue #5329, opened 2023-02-28
The default `count` parameter for `getObject` should be `null` not 0


elliette:
According to the `vm_service` [docs](https://api.flutter.dev/flutter/vm_service/VmService/getObject.html): 

> The offset and count parameters are used to request subranges of Instance objects with the kinds: String, List, Map, Uint8ClampedList, Uint8List, Uint16List, Uint32List, Uint64List, Int8List, Int16List, Int32List, Int64List, Flooat32List, Float64List, Inst32x3List, Float32x4List, and Float64x2List. These parameters are otherwise ignored.

DevTools currently checks for those types, and sends a request with `childCount` as the length of the instance (if it's not already set) or 0 if it's not one of those types. However, we keep missing types that should have `childCount` set to the length of the instance (for example, `Record` and `Set`). 

Setting the `count` to the `length` of the instance is equivalent to setting it to `null`. In both cases, the `vm_service` will return the entire object. Therefore, I think it makes more sense to have the default value be `null`, that way we aren't accidentally requesting empty objects.

PR #5323, opened 2023-02-28
Support `Set` instance inspection in debugger page

elliette:
Work towards https://github.com/flutter/devtools/issues/5263

Can inspect `Sets` for VM service connected apps:

Screenshot 2023-02-27 at 3 55 27 PM

Requires work on DWDS-side to work for web apps, see https://github.com/dart-lang/webdev/issues/2001

PR #5356, opened 2023-03-04
Do not send count and offset params when requesting the entire object

elliette:
Fixes https://github.com/flutter/devtools/issues/5329

Previously, for objects where there was no explicitly set offset and childCount, we were sending `getObject` requests with `offset` set to 0 and `count` set to the length of the object (if it was a `List`, `Map`, `Record` or `Set` instance) or 0 if it wasn't. 

This was causing a few problems. If we missed any of the instance kinds that had a `length` (for example, we were missing [`Set` instances](https://github.com/flutter/devtools/pull/5323)), then we would send a `getObject` request with `count` equal to 0.  The VM service would then correctly respond with an empty object. 

This PR ignores sending `offset` and `count` params in `getObject` requests, unless we are explicitly requesting a partial object.

### DIFF
--- a/packages/devtools_app/test/shared/dart_object_node_test.dart
+++ b/packages/devtools_app/test/shared/dart_object_node_test.dart
@@ -1,0 +1,130 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter_test/flutter_test.dart';
+
+import '../test_infra/utils/variable_utils.dart';
+
+void main() {
+  setUp(() {
+    resetRef();
+    resetRoot();
+  });
+
+  group('childCount', () {
+    group('for map instances', () {
+      test('entire map', () {
+        final map = buildMapVariable(length: 3);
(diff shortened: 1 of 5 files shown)


packages/devtools_app/test/shared/dart_object_node_test.dart
```diff
@@ -0,0 +1,130 @@
...
+        final map = buildMapVariable(length: 3);
+        expect(map.isPartialObject, isFalse);
+      });
+      test('map grouping with offset', () {
+        final mapGrouping = buildMapGroupingVariable(
+          length: 10,
+          offset: 2,
+          count: 4,
+        );
+        expect(mapGrouping.isPartialObject, isTrue);
```
kenzieschmoll, line 85: instead of duplicating all the groups for childCount and isPartialObject, can we just make 'map instances', 'list instances' etc. top level groups for which we check the childCount and isPartialObject values?

also, should we check Set objects too?

elliette: Good idea, refactored the test. Also added test cases for sets

kenzieschmoll (approved):

PR #5365, opened 2023-03-07
Make FlatTable column sizing configurable

kenzieschmoll:

This adds a `sizeToFit` parameter for `FlatTable` that defaults to true (current implementation), but when false, will size the table according to its content.

### DIFF
--- a/packages/devtools_app/lib/src/shared/table/column_widths.dart
+++ b/packages/devtools_app/lib/src/shared/table/column_widths.dart
@@ -20,7 +20,7 @@ const columnGroupSpacingWithPadding = columnGroupSpacing + 2 * defaultSpacing;
 const columnSpacing = defaultSpacing;
 
 extension FlatColumnWidthExtension<T> on FlatTableController<T> {
-  List<double> computeColumnWidthsForSingleWideColumn(List<T> data) {
+  List<double> computeColumnWidthsSizeToContent(List<T> data) {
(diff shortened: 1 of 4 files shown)


packages/devtools_app/lib/src/shared/table/column_widths.dart
```diff
@@ -20,7 +20,40 @@ const columnGroupSpacingWithPadding = columnGroupSpacing + 2 * defaultSpacing;
 const columnSpacing = defaultSpacing;
 
 extension FlatColumnWidthExtension<T> on FlatTableController<T> {
-  List<double> computeColumnWidths(double maxWidth) {
+  List<double> computeColumnWidthsForSingleWideColumn(List<T> data) {
```
polina-c, line 23: Could you add comment on behavior?

packages/devtools_app/lib/src/shared/table/column_widths.dart
```diff
@@ -20,7 +20,40 @@ const columnGroupSpacingWithPadding = columnGroupSpacing + 2 * defaultSpacing;
...
+    final widths = <double>[];
+    for (var column in columns) {
+      double width;
+      if (column.fixedWidthPx != null) {
```
polina-c, line 27: Swap if and else:    `column.fixedWidthPx == null...`

polina-c (approved):

PR #5366, opened 2023-03-08
Add Method Table to CPU profiler

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/841

Screenshot 2023-03-08 at 3 28 25 PM


packages/devtools_app/lib/src/screens/profiler/panes/method_table/method_table_model.dart
```diff
@@ -27,48 +37,102 @@ class MethodTableGraphNode extends GraphNode {
...
+      required double Function(MethodTableGraphNode) percentCallback,
+    }) {
+      return nodes
+          .cast<MethodTableGraphNode>()
+          // Sort in descending order.
+          .sorted((a, b) => percentCallback(b).compareTo(percentCallback(a)))
+          .map(
+            (node) => '${node.display} - ${percent2(percentCallback(node))}',
+          )
+          .join('\n    ');
```
bkonyi, line 102: What's with the spaces after the newlines?

kenzieschmoll: It is for an indent in the display

kenzieschmoll: will make a named variable to make that more clear

packages/devtools_app/lib/src/screens/profiler/panes/method_table/method_table.dart
```diff
@@ -38,6 +39,8 @@ class CpuMethodTable extends StatelessWidget {
   }
 }
 
+// TODO(kenz): ensure that this table automatically scrolls to the selected
```
bkonyi, line 42: When would this happen?

packages/devtools_app/lib/src/screens/profiler/panes/method_table/method_table_model.dart
```diff
@@ -12,63 +14,136 @@ class MethodTableGraphNode extends GraphNode {
...
+  String get display =>
+      '$name${_sourceUri.isNotEmpty ? ' - ($_sourceUri)' : ''}';
 
-  late Duration totalTime = Duration(
-    microseconds:
-        (totalTimeRatio * profileMetaData.time!.duration.inMicroseconds)
-            .round(),
-  );
+  /// The number of cpu samples where this method is on top of the stack.
+  int get selfCount => _selfCount;
```
bkonyi, line 58: Can this just be `final int selfCount`? Ditto for `totalCount`.

packages/devtools_app/lib/src/screens/profiler/panes/method_table/method_table.dart
kenzieschmoll: If you select a method from anywhere in the method graph (callees and callers), the graph reroots to that method, which means the method table should scroll to the new selection. We may eventually want to maintain some sort of history here for this reason.

packages/devtools_app/lib/src/screens/profiler/panes/method_table/method_table_model.dart
kenzieschmoll: _selfCount and _totalCount are non-final because they are modified in the `merge` method.

PR #5371, opened 2023-03-08
Minor reorg for integration tests.

polina-c:


packages/devtools_app/integration_test/test_infra/run/run_test.dart
```diff
@@ -8,8 +8,8 @@ import 'dart:io';
```
kenzieschmoll, line 8: this file is intentionally private so that it is clear this is an internal file intended to be used by the run_tests.dart script only.

packages/devtools_app/integration_test/test/live_connection/app_test.dart
```diff
@@ -5,14 +5,14 @@
...
 import 'package:devtools_app/devtools_app.dart';
-import 'package:devtools_app/src/framework/landing_screen.dart';
-import 'package:devtools_app/src/framework/release_notes/release_notes.dart';
 import 'package:devtools_app/src/shared/primitives/simple_items.dart';
 import 'package:devtools_test/devtools_integration_test.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
 
+import '../../test_infra/utils/connection.dart';
```
kenzieschmoll, line 14: we already have an `integration_test_utils.dart` filed in `packages/devtools_test`

we should consolidate. I can't remember why the integration_test_utils.dart file originally needed to be in `package:devtools_test`, but if it can be moved into `devtools_app/integration_test`, let's do that.

packages/devtools_app/integration_test/test_infra/run/run_test.dart
polina-c: Per my understanding leading '_' in file name means 'this file is not referenced outside of this folder, it can be references only inside this folder'.

So, after moving, the file became facade of the folder and to others became internals.

How does it sound?

packages/devtools_app/integration_test/test/live_connection/app_test.dart
polina-c: Tests fail because they do not see '../../test_infra/utils/connection.dart'. Moved to devtools_test.

packages/devtools_app/integration_test/test_infra/run/run_test.dart
polina-c: We would benefit from shared meaning of leading '_' in file name, written somewhere.

kenzieschmoll: I tried to look up official style guide guidance around this and didn't find anything. We do not follow the pattern you described all across DevTools. Search for `import '.*/_` as a regexp search in VS code and you'll see several places where we are importing private files. Some of them don't need to be private (_log_details.dart, _logs_table.dart, _message_column.dart), and some of them we are importing a platform specific configuration.

If we want to use this pattern, then we should make the rest of the codebase consistent so that it is self documenting. If so, can you add a comment to the top of this run_test.dart file then that states tests should not be run through this script, but should instead use the run_tests.dart script which performs essential set up steps?

> We would benefit from shared meaning of leading '_' in file name, written somewhere.

We shouldn't invest time into maintaining our own style guide IMO. If the codebase is consistent it is self documenting.

kenzieschmoll (approved):

PR #5381, opened 2023-03-09
Add documentation to ColumnData and clean up some tables

kenzieschmoll:
RELEASE_NOTE_EXCEPTION=[no user facing changes]

refactor only

### DIFF
--- a/packages/devtools_app/lib/src/screens/app_size/code_size_attribution.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/code_size_attribution.dart
@@ -83,9 +83,6 @@ class _CallGraphWithDominatorsState extends State<CallGraphWithDominators> {
 class CallGraphView extends StatefulWidget {
   const CallGraphView({required this.node});
 
-  static const Key fromTableKey = Key('CallGraphView - From table');
-  static const Key toTableKey = Key('CallGraphView - To table');
-
   final CallGraphNode node;
 
   @override
@@ -93,10 +90,6 @@ class CallGraphView extends StatefulWidget {
 }
 
 class _CallGraphViewState extends State<CallGraphView> {
-  final _fromColumn = FromColumn();
-
-  final _toColumn = ToColumn();
-
   late CallGraphNode selectedNode;
 
   @override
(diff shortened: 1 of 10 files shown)


packages/devtools_app/lib/src/screens/app_size/code_size_attribution.dart
```diff
@@ -193,13 +172,59 @@ class _CallGraphViewState extends State<CallGraphView> {
...
+}
+
+class _CallGraphTable extends StatelessWidget {
+  const _CallGraphTable({
+    required this.tableType,
+    required this.selectedNode,
+    required this.onNodeSelected,
+  });
+
+  static const _fromDataKey = 'call-graph-from';
```
bkonyi, line 197: Can you make these properties of `_CallGraphTableType` instead? That way we go from:

```dart
dataKey: tableType == _CallGraphTableType.from ? _fromDataKey : _toDataKey
```

To something like:

```dart
dataKey: tableType.dataKey
```

bkonyi (approved):

packages/devtools_app/lib/src/shared/table/table_data.dart
```diff
@@ -13,7 +13,17 @@ import '../theme.dart';
...
+///
+/// [ColumnData] objects should be defined as static where possible, and should
+/// not manage any stateful data. The table controllers expect columns to be
+/// solely responsible for declaring how to layout table data.
+///
+/// Any data that can't be stored on the [dataObject] may be accessed by passing
+/// a long-living controller to the constructor of the [ColumnData] subclass.
+///
+/// The controller is expected to be alive for the duration of the app
+/// connection.
```
polina-c, line 26: What happens after reconnection? Does table reset and start using new ColumnData instances?

kenzieschmoll: good question. I added dispose method to the table state classes to ensure we null out the table controller, which would mean it is guaranteed to be recreated for a new widget if the state stays alive. This should never happen, but it doesn't hurt to be extra cautious. 

If a column is not static (because a controller has to be passed to it), then it will be recreated for a new instance of the widget that creates the column with whatever controller is passed. 

When DevTools disconnects from an app, the existing controllers are destroyed and new ones will be created for the next app connection.

Issue #5091, opened 2023-01-23
Audit performance of the various DevTools screens and components


bkonyi:
A common concern raised by some Dash engineers is that DevTools often feels slower than existing tooling in Observatory.

Some examples of scenarios with poor performance:

- Opening DevTools while connecting to a CLI application automatically fetches and loads a timeline, locking up the UI thread until the timeline is fetched and processed
- Poor performance when creating filtered CPU profiles

Performance of the various tools should be audited to ensure that current power users of Observatory do not experience a degradation of their development workflows.

This is a blocker for deprecating and removing Observatory from the Dart SDK.

cc @a-siva @jacob314

kenzieschmoll:
Another possible performance optimization: stop copying data for flat tables https://github.com/flutter/devtools/pull/5384/files#r1132405605

kenzieschmoll:
> Opening DevTools while connecting to a CLI application automatically fetches and loads a timeline, locking up the UI thread until the timeline is fetched and processed

@bkonyi is this still an issue with the new perfetto integration? If it's not a problem now, there may not be a reason to invest in creating an "index" screen for cli apps.

bkonyi:
I'm personally content with the landing page performance as Perfetto doesn't seem to cause the UI thread to jank when loading the initial timeline like it did before, so users can still switch to another tab while the timeline is being processed. However, I do feel having an overview landing page could be useful, particularly in VM developer mode (cc @a-siva for his thoughts), but that can be pulled out into its own issue.

I think we can close this issue once https://github.com/flutter/devtools/issues/5703 is resolved.

kenzieschmoll:
Landing screen is tracked here: https://github.com/flutter/devtools/issues/4603. Since #5703 has been broken out into smaller tracking issues and is a meta tracking issue for all the smaller items, I'm going to let that work be tracked there. I will close this out, and I will add the appropriate labels to the other issues.

PR #5384, opened 2023-03-10
Add search to method table

kenzieschmoll:
RELEASE_NOTE_EXCEPTION=[no need for release note since this feature hasn't been released yet. The release notes has a note for adding the method table in general]
 
mt-search


packages/devtools_app/lib/src/shared/table/table_controller.dart
```diff
@@ -72,7 +90,16 @@ class FlatTableController<T> extends TableControllerBase<T> {
     ColumnData<T>? secondarySortColumn,
     String? dataKey,
   }) {
-    var data = List<T>.of(_originalData);
+    late List<T> data;
+    if (sortOriginalData) {
+      data = _modifiableOriginalData;
+    } else {
+      // TODO(kenz): copying the list for every sort could cause performance
```
bkonyi, line 97: Did you add this option because of performance issues in this method? If so, I don't see an issue with setting `sortOriginalData` as the default since the data is typically only being fed to the table controller and is replaced / added to rather than being used for additional processing, so order shouldn't really matter.

bkonyi (approved):

packages/devtools_app/lib/src/shared/table/table_controller.dart
kenzieschmoll: In some places, the original list could be the same list being used by the controller, so I decided to make this opt-in for now to ensure we don't break anything in other views. I agree it would be beneficial to make this behavior the default instead of the copy method that we have been using, but I wanted to do an audit of this behavior outside of the scope of this PR.

kenzieschmoll: Added a comment to https://github.com/flutter/devtools/issues/5091

CoderDake (approved):

bkonyi (approved):

PR #5383, opened 2023-03-09
Save all instances of class to console.

polina-c:
RELEASE_NOTE_EXCEPTION=[behind flag]

https://user-images.githubusercontent.com/12115586/224187482-08ea8474-e31b-443a-bf0e-c7d524c12e67.mov

PR #5388, opened 2023-03-10
Fix golden test at master.

polina-c:


polina-c:
Submitted the fix as part of https://github.com/flutter/devtools/pull/5383

Issue #5392, opened 2023-03-10
Can't inspect Debugger panel due to uncaught error


elliette:
I'm unable to inspect any widgets in the debugger panel due to the following error whenever I click on it to inspect:

<img width="864" alt="Screenshot 2023-03-10 at 3 07 31 PM" src="https://user-images.githubusercontent.com/21270878/224444766-a13768f3-0114-4098-a9b8-8a00c86a7906.png">

PR #5394, opened 2023-03-10
Can inspect widgets in Debugger panel

elliette:
Fixes https://github.com/flutter/devtools/issues/5392

To be honest, I don't entirely understand what was happening. But if you use the widget inspector to inspect the debugger panel of your running DevTools (while doing local development work) your running DevTools will crash with the 
error seen in https://github.com/flutter/devtools/issues/5392

This fixes that by returning `null` if `RouteInformation.state` (in this case `CodeViewSourceLocationNavigationState`) can't be cast into a `Map`

RELEASE_NOTE_EXCEPTION="No user facing changes, just affects our ability to debug DevTools"

### DIFF
--- a/packages/devtools_app/lib/src/shared/routing.dart
+++ b/packages/devtools_app/lib/src/shared/routing.dart
@@ -72,11 +72,7 @@ class DevToolsRouteInformationParser
     final configuration = DevToolsRouteConfiguration(
       path,
       uri.queryParameters,
-      routeInformation.state == null
-          ? null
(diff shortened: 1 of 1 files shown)


kenzieschmoll (approved):
odd, but lgtm

CC @bkonyi in case you have any ideas

kenzieschmoll:
I believe enabling the widget inspector triggers a hot reload, so this could be something broken with our state management

PR #5405, opened 2023-03-13
Clean up and refactor search code.

kenzieschmoll:
This PR reduces the formerly large `SearchFieldMixin` to include only what needs to be in a mixin (the state management for initializing and disposing search components). I moved controller functionality to the controller mixins `SearchControllerMixin` and `AutoCompleteSearchControllerMixin` and moved widget helper methods into formal Widgets `SearchField` and `AutoCompleteSearchField`.

### DIFF
--- a/packages/devtools_app/lib/src/shared/ui/search.dart
+++ b/packages/devtools_app/lib/src/shared/ui/search.dart
@@ -828,12 +828,12 @@ mixin SearchFieldMixin<T extends StatefulWidget>
   SearchControllerMixin get searchController;
 
   @override
-  void didChangeDependencies() {
-    super.didChangeDependencies();
-    // We call this in [didChangeDependencies] in case the [searchController] is
-    // only available when [BuildContext] is available. This happens whenever a
-    // controller is provided through package:provider.
+  void initState() {
+    super.initState();
     if (this is ProvidedControllerMixin) {
+      // Controllers provided through package:provider will not be ready until
+      // [didChangeDependencies] is called, so ensure [searchController] is
+      // ready before calling [searchController.initSearch].
       (this as ProvidedControllerMixin).callWhenControllerReady((_) {
         searchController.initSearch();
       });
--- a/packages/devtools_app/lib/src/shared/ui/search.dart
+++ b/packages/devtools_app/lib/src/shared/ui/search.dart
@@ -19,7 +19,8 @@ import '../theme.dart';
 import '../ui/utils.dart';
 import '../utils.dart';
 
-// TODO(kenz): break this file up into managable pieces.
+// TODO(https://github.com/flutter/devtools/issues/5416): break this file up
+// into managable pieces.
 
 /// Top 10 matches to display in auto-complete overlay.
 const defaultTopMatchesLimit = 10;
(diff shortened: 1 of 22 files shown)


packages/devtools_app/lib/src/shared/ui/search.dart
```diff
@@ -62,6 +61,14 @@ mixin SearchControllerMixin<T extends DataSearchStateMixin> {
...
   Duration? get debounceDelay => null;
 
+  /// Text field controller for the [SearchField] that this instance of
+  /// [SearchControllerMixin] controls.
+  SearchTextEditingController? searchTextFieldController;
+
+  /// Focus node for the [SearchField] that this instance of
+  /// [SearchControllerMixin] controls.
+  FocusNode? searchFieldFocusNode;
+
```
elliette, line 71: When would these be `null` (why are they nullable?)

kenzieschmoll: These are nulled out in dispose and re-initialized in initSearch. The enclosing controller mixin will live longer than the search field for which the text field and the focus node are used. However, there are several places in the controller that need to access the text field controller and the focus node, so we store them in the controller, and then init and dispose them with the widget lifecycle (this is what SearchFieldMixin does). I don't love this solution, but it seemed cleaner than what we had.

elliette: Got it, that makes sense. I'm mostly trying to figure out if there is a way having all the null-assert `!s` for these fields. Would it be possible to change these to be `late` fields (initialized in `initSearch`) and disposed of but not set to `null` during dispose? Or do they need to be reset back to `null`?

kenzieschmoll: Setting to null ensures that we don't try to dispose an already disposed object, whereas using late wouldn't protect against that. For example, we dispose these objects in initSearch before re-initializing them, and we wouldn't be able to do that if we used late.

packages/devtools_app/lib/src/shared/ui/search.dart
```diff
@@ -19,6 +19,8 @@ import '../theme.dart';
 import '../ui/utils.dart';
 import '../utils.dart';
 
+// TODO(kenz): break this file up into managable pieces.
```
bkonyi, line 22: Should this have an issue associated with it?

packages/devtools_app/lib/src/shared/ui/search.dart
elliette: Got it, thanks! Optional - switch these (and `rawKeyboardFocusNode` below) to private nullable fields, and add corresponding non-nullable getter for them. Besides that, LGTM!

elliette (approved):
LGTM with one suggestion

packages/devtools_app/lib/src/shared/ui/search.dart
```diff
@@ -673,6 +720,30 @@ mixin AutoCompleteSearchControllerMixin on SearchControllerMixin {
...
+
+  void selectFromSearchField(String selection) {
+    searchTextFieldController!.clear();
+    search = selection;
+    clearSearchField(force: true);
+    selectTheSearch = true;
+    closeAutoCompleteOverlay();
+  }
+
+  void clearSearchField({force = false}) {
```
bkonyi, line 732: This parameter should have a type.

bkonyi (approved):

PR #5411, opened 2023-03-14
Draft integration test for eval and browse.

polina-c:


packages/devtools_app/integration_test/test_infra/run/_in_file_args.dart
```diff
@@ -0,0 +1,62 @@
...
+            'test/test_infra/fixtures/flutter_app';
+
+  /// If true, experiments will be enabled in the test.
+  final bool experimentsOn;
+
+  /// Path to the application to connect to.
+  final String appPath;
+}
+
+const _argPrefix = '// test-argument:';
```
bkonyi, line 42: I think this should be more permissive even if we're going to enforce this format in code review. Changing this to a regex will allow for more flexibility for whitespace (which means less head scratching when someone's writing a test for the first time and wonders why `//test-argument:foo=bar` doesn't work) and will also simplify parsing by using groups.

Here's an example pattern that could work where group(1) is the matched key and group(2) is the matched value:

`\/\/\s*test-arguments:\s*(\w*)=(\w*)`

polina-c: Yes, I started with regex, but then found that code readability became much worse. This means people will scratch their heads maintaining this code. 
So, this is choice between two downsides. So I prefer to keep current option.

bkonyi: Nothing a little documentation wouldn't fix... :-)

I still prefer the regex since it's less brittle and more permissive, but maybe @kenzieschmoll or @jacob314 will have an opinion.

polina-c: Let's save their time. I will switch to regex :)

packages/devtools_app/integration_test/run_tests.dart
```diff
@@ -56,25 +53,29 @@ void main(List<String> args) async {
...
+      await _runTest(modifiableArgs, testTarget);
     }
   }
 }
 
-void _maybeAddOfflineArgument(List<String> args, String testTarget) {
-  if (testTarget.startsWith(_offlineIndicator)) {
-    args.add(TestArgs.offlineArg);
+Future<void> _runTest(List<String> modifiableArgs, String testFilePath) async {
+  _maybeAddOfflineArgument(modifiableArgs, testFilePath);
```
kenzieschmoll, line 63: is this still necessary? It looks like we are no longer using `TestArgs.offline`

packages/devtools_app/integration_test/run_tests.dart
```diff
@@ -56,25 +53,29 @@ void main(List<String> args) async {
...
+
+  final inFileArgs = InFileArgs(testFilePath);
+  final cliArgs = TestArgs(modifiableArgs);
```
kenzieschmoll, line 66: nit: maybe these classes  (and variables) are more aptly named TestFileArgs and TestRunnerArgs

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
```diff
@@ -0,0 +1,48 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// Do not delete these arguments. They are parsed parsed by test runner.
+// test-argument:experimentsOn=true
```
kenzieschmoll, line 6: nit: use `experiments-on` or `experiments_on` to better match standard argument formatting. We could also just use `experiments` to simplify

packages/devtools_app/integration_test/test_infra/run/_in_file_args.dart
```diff
@@ -0,0 +1,61 @@
...
+
+/// Test arguments, defined inside the test file as a comment.
+class InFileArgs {
+  factory InFileArgs(String testFilePath) {
+    final content = File(testFilePath).readAsStringSync();
+    return InFileArgs.fromFileContent(content);
+  }
+
+  factory InFileArgs.fromFileContent(String fileContent) {
+    final values = _parseFileContent(fileContent);
```
kenzieschmoll, line 21: nit: name something other than values, as `values` is already a field name for the Map type. Reading `values.keys` below is a bit confusing at first

packages/devtools_app/integration_test/test_infra/run/_in_file_args.dart
```diff
@@ -0,0 +1,61 @@
...
+);
+
+Map<InFileArgItems, dynamic> _parseFileContent(String fileContent) {
+  final matches = _argRegex.allMatches(fileContent);
+
+  final entries = matches.map<MapEntry<InFileArgItems, dynamic>>(
+    (RegExpMatch m) {
+      final name = m.group(1) ?? '';
+      if (name.isEmpty)
+        throw 'Name of test argument name should be provided: [${m.group(0)}].';
```
kenzieschmoll, line 54: don't throw plain text. Throw an 'Exception' or 'ArgumentError'

also, remove second "name"

packages/devtools_app/test/integration_test/in_file_args_test.dart
```diff
@@ -0,0 +1,63 @@
...
+''',
+    output: InFileArgs.fromValues({
+      InFileArgItems.experimentsOn: true,
+      InFileArgItems.appPath: _testAppPath,
+    }),
+  ),
+];
+
+void main() {
+  for (final t in tests) {
```
kenzieschmoll, line 44: I don't think this test will actually get run on the bots. I would be surprised since it is in the integration_test folder. We can move the file args parsing code and its test to devtools_test, and then run the dart test on the bots from there.

This also isn't an "integration test" so it should probably be moved under a "test" folder anyway (e.g. devtools_test/test). After moving it to devtools_test, you'll need to tweak the 'main' job in bots.sh to test this directory like we do for devtools_shared

polina-c: Why not? 
All tests under `packages/devtools_app/test` run, and this one will too.

packages/devtools_app/integration_test/run_tests.dart
polina-c: We stopped using experiments, but still using offline. I think it is convenient to have them in separate folders and not having to provide additional signal that they are offline.

packages/devtools_app/test/integration_test/in_file_args_test.dart
polina-c: Renamed folder to `integration_test_infra` to avoid confusing.

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
polina-c: This mean I will need to introduce field for enum of the arguments instead of just using their name.

May be it is ok to have different standard for different type of arguments.

packages/devtools_app/integration_test/run_tests.dart
kenzieschmoll: It looks like we don't use the `TestArgs.offline` field anymore though. We are now using the absence of a test app uri to signal "offline" right?

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
kenzieschmoll: > This mean I will need to introduce an additional field for the enum of the arguments instead of just using their name.

Can you explain what you mean by this? I'm not following. I'm just suggesting a trivial change from the key "experimentsOn" to "experiments" or "experiments-on"

packages/devtools_app/test/integration_test/in_file_args_test.dart
kenzieschmoll: This file isn't under devtools_app/test, so it is not covered by that use case

packages/devtools_app/integration_test/test_infra/run/_in_file_args.dart
```diff
@@ -5,31 +5,31 @@
...
     }
 
-    return InFileArgs.fromValues(values);
+    return TestFileArgs.fromItems(testFileArgItems);
   }
 
-  InFileArgs.fromValues(Map<InFileArgItems, dynamic> values)
-      : experimentsOn = values[InFileArgItems.experimentsOn] ?? false,
-        appPath = values[InFileArgItems.appPath] ??
+  TestFileArgs.fromItems(Map<TestFileArgItems, dynamic> values)
```
kenzieschmoll, line 30: nit: rename "fromItems" to "fromMap" or "parse", and rename 'values' to 'map'

packages/devtools_app/integration_test/test_infra/run/_in_file_args.dart
```diff
@@ -44,16 +44,17 @@ final _argRegex = RegExp(
...
   final matches = _argRegex.allMatches(fileContent);
 
-  final entries = matches.map<MapEntry<InFileArgItems, dynamic>>(
+  final entries = matches.map<MapEntry<TestFileArgItems, dynamic>>(
     (RegExpMatch m) {
       final name = m.group(1) ?? '';
       if (name.isEmpty)
-        throw 'Name of test argument name should be provided: [${m.group(0)}].';
+        throw ArgumentError(
+            'Name of test argument should be provided: [${m.group(0)}].');
```
kenzieschmoll, line 55: trailing comma

packages/devtools_app/test/integration_test/in_file_args_test.dart
kenzieschmoll: New location sgtm, disregard my last comment.

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
kenzieschmoll: after looking at enum, I see what you mean. leaving as is sgtm.

packages/devtools_app/integration_test/run_tests.dart
polina-c: True. Cleaned up both offline and experiments TestRunnerArgs.

packages/devtools_app/integration_test/test_infra/run/run_test.dart
```diff
@@ -297,22 +295,6 @@ class TestRunnerArgs {
   /// This value will only be used when [offline] has not been set to true.
```
kenzieschmoll, line 295: please remove all the references to [offline] in the comments

kenzieschmoll (approved):
one dart doc comment then lgtm

PR #5419, opened 2023-03-14
Fine tune references in console.

polina-c:

https://user-images.githubusercontent.com/12115586/225167586-49b6faab-e3f5-4d78-8d0e-ec5298a612c2.mov

Screenshot 2023-03-14 at 4 57 47 PM

### DIFF
--- a/packages/devtools_app/lib/src/shared/diagnostics/generic_instance_reference.dart
+++ b/packages/devtools_app/lib/src/shared/diagnostics/generic_instance_reference.dart
@@ -124,7 +124,8 @@ enum RefNodeType {
 
   final RefDirection? direction;
 
-  bool get isRoot => {refRoot, staticRefRoot, liveRefRoot}.contains(this);
+  bool get isRoot => const {refRoot, staticRefRoot, liveRefRoot}.contains(this);
 
-  bool get isLive => {liveOutRefs, liveInRefs, liveRefRoot}.contains(this);
+  bool get isLive =>
+      const {liveOutRefs, liveInRefs, liveRefRoot}.contains(this);
 }
--- a/packages/devtools_app/lib/src/shared/diagnostics/generic_instance_reference.dart
+++ b/packages/devtools_app/lib/src/shared/diagnostics/generic_instance_reference.dart
@@ -90,10 +90,10 @@ class ObjectReferences extends GenericInstanceRef {
     final result = heapSelection.countOfReferences(refNodeType.direction);
     if (result != null) return result;
 
-    final inatance = value;
-    if (inatance is! InstanceRef) return null;
+    final instance = value;
+    if (instance is! InstanceRef) return null;
 
-    return inatance.length;
+    return instance.length;
   }
 }
 
--- a/packages/devtools_app/lib/src/shared/diagnostics/generic_instance_reference.dart
+++ b/packages/devtools_app/lib/src/shared/diagnostics/generic_instance_reference.dart
@@ -63,7 +63,9 @@ class ObjectReferences extends GenericInstanceRef {
     required IsolateRef super.isolateRef,
     required super.value,
     required HeapObjectSelection super.heapSelection,
-  });
+  }) {
+    if (refNodeType.isLive) assert(value != null);
+  }
 
   ObjectReferences.copyWith(
     ObjectReferences ref, {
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/shared/diagnostics/generic_instance_reference.dart
```diff
@@ -63,7 +63,9 @@ class ObjectReferences extends GenericInstanceRef {
     required IsolateRef super.isolateRef,
     required super.value,
     required HeapObjectSelection super.heapSelection,
-  });
+  }) {
+    if (refNodeType.isLive) assert(value != null);
```
bkonyi, line 67: Nit: `assert(!refNodeType.isLive || value != null)`

packages/devtools_app/lib/src/shared/diagnostics/generic_instance_reference.dart
```diff
@@ -113,4 +125,6 @@ enum RefNodeType {
   final RefDirection? direction;
 
   bool get isRoot => {refRoot, staticRefRoot, liveRefRoot}.contains(this);
+
+  bool get isLive => {liveOutRefs, liveInRefs, liveRefRoot}.contains(this);
```
bkonyi, line 129: Can these sets be `const`?

packages/devtools_app/lib/src/shared/diagnostics/references.dart
```diff
@@ -87,9 +86,9 @@ Future<void> addChildReferences(
             ),
           ),
       ]);
-      break;
+      return;
```
bkonyi, line 89: Nit: Are these returns necessary? We won't require `break`s in `case` statements starting with 3.0, so it'd probably be cleaner to leave them as is and then remove them when we can.

packages/devtools_app/lib/src/shared/diagnostics/references.dart
```diff
@@ -279,18 +282,26 @@ void _addLiveReferenceToNode(
   HeapObjectSelection heapSelection, {
   String namePrefix = '',
 }) {
-  if (instance is! InstanceRef && instance is! ContextRef) return;
+  if (instance is! ObjRef) throw 'Unexpected type: ${instance.runtimeType}.';
```
bkonyi, line 285: Don't throw bare strings. Consider throwing `TypeError` or `ArgumentError` instead.

bkonyi: Also, if this is always expected to be an `ObjRef`, the `instance` parameter should have type `ObjRef` or `ObjRef?`. We should avoid using `Object` when we have a known expected type as it's always better to have a static analysis failure than a runtime failure.

packages/devtools_app/lib/src/shared/diagnostics/generic_instance_reference.dart
polina-c: Current option looks more readable for me.

packages/devtools_app/lib/src/shared/diagnostics/generic_instance_reference.dart
polina-c: yes!

packages/devtools_app/lib/src/shared/diagnostics/references.dart
polina-c: cool!

In this exact case the method early returns sometimes in the middle of the case. So, I want it to be consistent  and make sure all options do not count on any 'after switch' operations in this method.

packages/devtools_app/lib/src/shared/diagnostics/references.dart
polina-c: I agree. I am not confident about expectations for this method. I just never saw non ObjRef so far. 
So, If I ever get non ObjRef, I (or user) will get exception and will add processing for it.

polina-c: I dream to refactor this code eventually to make contracts clearer.

packages/devtools_app/lib/src/shared/diagnostics/references.dart
polina-c: Where can I read more about non-required break? Right now absence of break means  that the case is combined with next one. So, I am wondering how backward compatibility is handled.

packages/devtools_app/lib/src/shared/diagnostics/generic_instance_reference.dart
```diff
@@ -83,6 +85,16 @@ class ObjectReferences extends GenericInstanceRef {
...
   IsolateRef get isolateRef => super.isolateRef!;
+
+  int? get childCount {
+    final result = heapSelection.countOfReferences(refNodeType.direction);
+    if (result != null) return result;
+
+    final inatance = value;
+    if (inatance is! InstanceRef) return null;
+
+    return inatance.length;
```
CoderDake, line 96:
```suggestion
    final instance = value;
    if (instance is! InstanceRef) return null;

    return instance.length;
```

kenzieschmoll (approved):

Issue #5287, opened 2023-02-21
"No frames available" error when debugger is paused inside async javascript code.


annagrin:
Repro steps:

Flutter's `example/hello_world` directory:
- run `flutter run -d chrome`
- open the Dart DevTools uri
- hit pause and run until you DevTools does not show where it stopped (takes about 5-6 attempts for me)
- evaluate "true" in the evaluation box
- observe "No frames available" error

<img width="1162" alt="image" src="https://user-images.githubusercontent.com/12737689/220459489-4b4115a8-00e3-4c33-92e4-fbe9520e0aa0.png">

Chrome does show a list of frames available, but all of them async and pretty sure they are all from js code:

<img width="1051" alt="image" src="https://user-images.githubusercontent.com/12737689/220460400-35386ad6-7e00-4146-8292-396eabddc79b.png">

**Some background**

In this example I paused the app, and hit javascript code with no frames on the stack related to dart.

@elliette  looks like the error comes from DevTools, and it happens before the call to `chromeProxyService.evaluateInFrame`. 
https://github.com/flutter/devtools/blob/79099e863bc45795be63d85f5479c65e1953a7d6/packages/devtools_app/lib/src/shared/console/eval/eval_service.dart#L137

Potential solutions: 

- Make dwds continue executing until dart code is hit. Not sure what happens if it never does - in that case the pause will appear not working, so maybe we need a timeout with a warning? This change, if we prefer it, would be most easily done in dwds: https://github.com/dart-lang/webdev/issues/1999

- Or simply assume we only have javascript frames in that case and report an error from DevTools describing that evaluating inside javascript code is not currently supported.

annagrin:
Update: tried to repro the issue in VSCode but could not. Here is the output for the same scenario described in a repro, but using debugger from VSCode:

<img width="1019" alt="image" src="https://user-images.githubusercontent.com/12737689/220462704-619c8dad-a5e8-4156-8ff3-b743df440bed.png">

annagrin:
Update: dwds does have frames when the app is paused:

```[[Frame index: 3433], [Frame index: 3434], [Frame index: 3435], [Frame
index: 3446], [Frame index: 3456], [Frame index: 3457], [Frame index: 3460], [Frame index: 3461],
[Frame index: 3470], [Frame index: 3476], [Frame index: 3477], [Frame index: 3480], [Frame index:
3481], [Frame index: 3484]]```

[...]

PR #5427, opened 2023-03-15
If hitting "pause" pauses in native code, show stack trace and allow expression evaluation 

elliette:
Work towards https://github.com/flutter/devtools/issues/5287

Requires DWDS-side change as well: https://github.com/dart-lang/webdev/pull/2038

When a user hits "pause" and Chrome pauses within JS code, we show the JS stack trace (below) and allow expression evaluation in the main isolate's root library (this was added to DevTools in https://github.com/flutter/devtools/pull/5248)

Screenshot 2023-03-15 at 3 23 30 PM

Without the DWDS changes, DevTools instead shows a warning (and allows expression evaluation in the main isolate's root library while paused):

Screenshot 2023-03-16 at 10 54 26 AM


CoderDake (approved):

packages/devtools_app/lib/src/screens/debugger/call_stack.dart
```diff
@@ -89,10 +89,12 @@ class _CallStackState extends State<CallStack>
...
           children: [
-            TextSpan(
-              text: ' $locationDescription',
-              style: selected ? theme.selectedTextStyle : theme.subtleTextStyle,
-            ),
+            if (locationDescription != null)
+              TextSpan(
+                text: ' $locationDescription',
+                style:
+                    selected ? theme.selectedTextStyle : theme.subtleTextStyle,
```
kenzieschmoll, line 96: at master this text style is:

```
              style: selected
                  ? theme.selectedSubtleTextStyle
                  : theme.subtleTextStyle,
```

maybe a merge is needed

packages/devtools_app/lib/src/screens/debugger/debugger_controller.dart
```diff
@@ -388,9 +389,31 @@ class DebuggerController extends DisposableController
     // we fetch all variable objects twice (once in _getFullStack and once in
     // in_createStackFrameWithLocation).
     if (await serviceManager.connectedApp!.isDartWebApp) {
+      final topFrame = pauseEvent?.topFrame;
+      if (topFrame == null) {
```
kenzieschmoll, line 393: maybe we should add an analytics event here too so that we can catch this sort of thing in the future and measure the occurrence of this issue over time

elliette: Done! Added a `pausedWithNoFrames` event

packages/devtools_app/lib/src/screens/debugger/call_stack.dart
elliette: Done!

PR #5430, opened 2023-03-16
Potential fix for DevTools going a bit blurry when window is expanded 

elliette:
Note: Need to confirm this fixes the issue when I'm at home (I can only reproduce this on my monitor at home, not in the office)

Details here: https://github.com/flutter/flutter/issues/122541

When expanded:

Screenshot 2023-03-13 at 2 56 19 PM

When not expanded:
Screenshot 2023-03-13 at 2 56 03 PM

`RELEASE_NOTE_EXCEPTION="No new feature / not a fix for a feature"`

### DIFF
--- a/packages/devtools_app/web/index.html
+++ b/packages/devtools_app/web/index.html
@@ -53,6 +53,9 @@
   <!-- This script adds the flutter initialization JS code -->
   <script src="flutter.js" defer></script> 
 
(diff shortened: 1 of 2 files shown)


bkonyi (approved):

elliette:
Confirmed that this fixes the blurriness issue on my monitor.

PR #5431, opened 2023-03-16
Integration tests for eval.

polina-c:


packages/devtools_app/integration_test/test/live_connection/performance_screen_event_recording_test.dart
```diff
@@ -27,10 +26,10 @@ void main() {
     logStatus(
       'Open the Performance screen and switch to the Timeline Events tab',
     );
-    await tester.tap(
-      find.widgetWithText(Tab, ScreenMetaData.performance.title),
-    );
-    await tester.pump(longPumpDuration);
+
+    await switchToScreen(tester, ScreenMetaData.performance);
+    await tester.pump(safePumpDuration);
```
kenzieschmoll, line 31: please switch this back to the longPumpDuration. IIRC, this test was flaky without it.

kenzieschmoll (approved):
lgtm with one comment

packages/devtools_app/integration_test/test/live_connection/performance_screen_event_recording_test.dart
polina-c: switchToScreen already contains one pump. So I replaced one  longPumpDuration with two safePumpDuration, that should be enough.

PR #5438, opened 2023-03-17
Remove Catapult trace viewer from bundled Perfetto

bkonyi:
Saves ~2.6MiB or 5.6% of the DevTools artifacts size


tool/update_perfetto.sh
```diff
@@ -45,6 +45,9 @@ find ./ -name '*.js.map' -exec rm {} \;
 find ./ -name 'traceconv.wasm' -exec rm {} \;
 find ./ -name 'traceconv_bundle.js' -exec rm {} \;
```
kenzieschmoll, line 46: nit can you move these down to the deleting catapult files? these are for the legacy UI as well

kenzieschmoll:
Have you manually tested the Perfetto trace viewer to ensure core functionality was not broken by the perfetto asset update? This comes with risk that our code could be broken by upstream changes to Perfetto, so if we update the assets we should do some stress testing beyond what our tests cover.

You can also consider only deleting the catapult files and updating the script if you want.

tool/update_perfetto.sh
bkonyi: Done.

bkonyi:
> Have you manually tested the Perfetto trace viewer to ensure core functionality was not broken by the perfetto asset update? This comes with risk that our code could be broken by upstream changes to Perfetto, so if we update the assets we should do some stress testing beyond what our tests cover.
> 
> You can also consider only deleting the catapult files and updating the script if you want.

Ah good point, I just figured things would work. Running the script seems to break some fonts / backgrounds, so I've just reverted the roll portion of this PR.

kenzieschmoll (approved):

PR #5449, opened 2023-03-20
Test that references are listed.

polina-c:
RELEASE_NOTE_EXCEPTION=[testing]


packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
```diff
@@ -116,5 +124,77 @@ class _EvalTester {
...
+    Finder finder, {
+    Duration? duration,
+    Finder? next,
+  }) async {
+    int tryNumber = 0;
+
+    Future<void> action() async {
+      logStatus('tapping #$tryNumber to find \n'
+          '[$finder]\n');
+      tryNumber++;
```
bkonyi, line 178: Nit: let's update `action` to take `action(int attempt)` instead of incrementing the try counter here. That way we can declare `int tryNumber` just before the loop and keep the increment logic in the loop.

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
```diff
@@ -116,5 +124,77 @@ class _EvalTester {
...
+
+    if (next == null) return null;
+
+    // Tthese tries are needed because tap in console is flaky.
+    while (tryNumber < 10) {
+      try {
+        final items = tester.widgetList(next);
+        if (items.isNotEmpty) return items.first;
+        await action();
+      } on StateError {
```
bkonyi, line 194: What throws `StateError`? `tester.widgetList`?

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
```diff
@@ -116,5 +124,77 @@ class _EvalTester {
...
+        await action();
+      }
+    }
+    throw 'Cound not find $next';
```
bkonyi, line 198: Avoid throwing bare `String`s.

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
polina-c: yes. added comment

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
```diff
@@ -2,17 +2,22 @@
...
+// Do not delete these arguments. They are parsed by test runner.
 // test-argument:appPath="test/test_infra/fixtures/memory_app"
+// test-argument:experimentsOn=true
 
 // ignore_for_file: avoid_print
 
+import 'package:devtools_app/src/screens/memory/panes/control/primary_controls.dart';
+import 'package:devtools_app/src/screens/memory/panes/diff/widgets/snapshot_list.dart';
+import 'package:devtools_app/src/shared/banner_messages.dart';
+import 'package:devtools_app/src/shared/common_widgets.dart';
```
kenzieschmoll, line 14: consider importing devtools_app.dart in tests instead of importing everything individually

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
```diff
@@ -116,5 +124,82 @@ class _EvalTester {
     await tester.pumpAndSettle();
     await simulateKeyDownEvent(LogicalKeyboardKey.enter);
     await simulateKeyUpEvent(LogicalKeyboardKey.enter);
+    await tester.pump(longPumpDuration);
+  }
+
+  Future<void> switchToSnapshotsAndTakeOne() async {
```
kenzieschmoll, line 130: can we make this method a private helper instead of a method on EvalTester? This method doesn't really have anything to do with eval.

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
```diff
@@ -116,5 +124,82 @@ class _EvalTester {
...
+  }) async {
+    int tryNumber = 0;
+
+    Future<void> action() async {
+      logStatus('tapping #$tryNumber to find \n'
+          '[$finder]\n');
+      tryNumber++;
+      await tester.tap(finder);
+      await tester.pump(duration);
+      await tester.pumpAndSettle();
```
kenzieschmoll, line 184: we shouldn't need both pump and pumpAndSettle

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
```diff
@@ -116,5 +124,82 @@ class _EvalTester {
...
+    }
+
+    await action();
+
+    if (next == null) return null;
+
+    // Tthese tries are needed because tap in console is flaky.
```
kenzieschmoll, line 191: sp: these

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
```diff
@@ -116,5 +124,82 @@ class _EvalTester {
...
+  Future<Widget?> tapAndPump(
+    Finder finder, {
+    Duration? duration,
+    Finder? next,
+  }) async {
+    int tryNumber = 0;
+
+    Future<void> action() async {
+      logStatus('tapping #$tryNumber to find \n'
+          '[$finder]\n');
```
kenzieschmoll, line 180: this can fit on one line

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
```diff
@@ -116,5 +124,82 @@ class _EvalTester {
...
+    // Sort by class.
+    await tapAndPump(find.text('Class'));
+
+    // Select class.
+    await tapAndPump(find.text('MyApp'));
+  }
+
+  /// Taps and settles.
+  ///
+  /// If [next] is provided, will repeat the tap till [next] combined with [at] returns results.
```
kenzieschmoll, line 169: r/till/ until

kenzieschmoll, line 169: there is no parameter [at]. Please update the uses of at in this dart doc

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
```diff
@@ -116,5 +124,82 @@ class _EvalTester {
...
+    // Close warning and chart to get screen.
+    await tapAndPump(find.byKey(DebugModeMemoryMessage.closeKey));
+    await tapAndPump(find.text(PrimaryControls.memoryChartText));
+
+    // Make console wider.
+    // The distance is big enough to see more items in console,
+    // but not too big to make classes in snapshot hidden.
+    const dragDistance = -320.0;
+    await tester.drag(
+      find.byKey(ConsolePaneHeader.theKey),
```
kenzieschmoll, line 143: find by type instead of creating a key just for testing

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
```diff
@@ -116,5 +124,82 @@ class _EvalTester {
...
     await simulateKeyUpEvent(LogicalKeyboardKey.enter);
+    await tester.pump(longPumpDuration);
+  }
+
+  Future<void> switchToSnapshotsAndTakeOne() async {
+    // Open memory screen.
+    await switchToScreen(tester, ScreenMetaData.memory);
+
+    // Close warning and chart to get screen.
+    await tapAndPump(find.byKey(DebugModeMemoryMessage.closeKey));
```
kenzieschmoll, line 135: instead of creating a key, Use flutter test finders to find the button. You can find by icon Icons.close that is a descendent of BannerMessage

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
polina-c: Not everything is available via devtools_app.dart and we want to be consistent. So, I suggest to keep individual imports the same way as in application.

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
polina-c: It needs state and helpers of this class. So, renamed the class to make it covering wider scope.

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
polina-c: pumpAndSettle does not take nullable duration, so code is simpler this way

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
kenzieschmoll: Using devtools_app.dart is fine in tests, but we do not need to enforce it. In tests it is fine to use either or 👍

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
kenzieschmoll: pumpAndSettle does take a duration parameter. if you pass the duration there, then you can remove the call to pump.

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
```diff
@@ -36,74 +41,77 @@ void main() {
...
-        expect(
-          find.textContaining(expectedResponse),
-          findsOneWidget,
-        );
-      }
+      expect(expectedResponse, findsOneWidget);
     } catch (e) {
       // In case of unexpected response take golden for troubleshooting.
-      print('Unexpected response: $e');
+      print(e.toString());
```
kenzieschmoll, line 111: use logStatus instead

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
polina-c: There are other close icons. For example, to delete snapshot.

packages/devtools_app/lib/src/shared/banner_messages.dart
```diff
@@ -127,7 +127,9 @@ class BannerMessage extends StatelessWidget {
     required this.textSpans,
     required this.screenId,
     required this.messageType,
+    this.closeKey,
```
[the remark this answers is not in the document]
polina-c: Tried. For some reasons the finder you suggested did not find the button.

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
polina-c: duration here is nullable
pumpAndSettle takes non-nullable duration

Yes, it can be handled in the code at cost of readability. I prefer to stay with readability.

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
kenzieschmoll: You can use the descendant finder to ensure that the close icon is a child of BannerMessage. We should avoid creating keys just for testing when a widget can be looked up easily with test finders.

packages/devtools_app/lib/src/shared/banner_messages.dart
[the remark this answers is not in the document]
kenzieschmoll: The finder I suggested was just an example from somewhere else in the code. You'll need to change the types to the widget types you are looking for.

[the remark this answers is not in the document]
polina-c: Got this:
```
[FlutterError]: The finder "zero widgets with icon "IconData(U+0E16A)" that has ancestor(s) with type "DebugModeMemoryMessage" (ignoring offstage widgets)" (used in a call to "tap()") could not find any matching widgets.
```

for the code:
```
find.descendant(
        of: find.byType(DebugModeMemoryMessage),
        matching: find.byIcon(Icons.close),
      )
```

kenzieschmoll (approved):

PR #5453, opened 2023-03-20
Clean up search code

kenzieschmoll:
This further cleans up the search code to 
1. use the default `SearchController.matchesForSearch` behavior where possible
2. limit where the `SearchFieldMixin`(which is responsible for managing search field state) is used directly. The `SearchField` and `StatelessSearchField` widgets are now documented to describe where each should be used.
3. search field UI polish

Further cleanup is possible but not the highest priority.

### DIFF
--- a/packages/devtools_app/lib/src/screens/logging/logging_controller.dart
+++ b/packages/devtools_app/lib/src/screens/logging/logging_controller.dart
@@ -828,9 +828,8 @@ class LogData with SearchableDataMixin {
 
   @override
   bool matchesSearchToken(RegExp regExpSearch) {
-    return (summary != null &&
-            summary!.caseInsensitiveContains(regExpSearch)) ||
-        (details != null && details!.caseInsensitiveContains(regExpSearch));
+    return (summary?.caseInsensitiveContains(regExpSearch) == true) ||
+        (details?.caseInsensitiveContains(regExpSearch) == true);
   }
 
   @override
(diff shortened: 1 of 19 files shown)


packages/devtools_app/lib/src/screens/logging/logging_controller.dart
```diff
@@ -851,6 +826,13 @@ class LogData with SearchableDataMixin {
     return false;
   }
 
+  @override
+  bool matchesSearchToken(RegExp regExpSearch) {
+    return (summary != null &&
```
bkonyi, line 831: Nit: maybe do something like this instead:

```dart
return (summary?.caseInsensitiveContains(regExpSearch) == true) || (details?.caseInsensitiveContains(regExpSearch) == true);
```

packages/devtools_app/lib/src/screens/profiler/cpu_profiler.dart
```diff
@@ -197,13 +197,15 @@ class _CpuProfilerState extends State<CpuProfiler>
             // well. This will require implementing search for tree tables.
             if (CpuProfiler.searchableTabKeys.contains(currentTab.key)) ...[
               if (currentTab.key == ProfilerTab.methodTable.key)
-                CpuProfilerSearchField<MethodTableController>(
+                SearchField<MethodTableController>(
                   searchController: widget.controller.methodTableController,
+                  containerPadding: const EdgeInsets.all(0.0),
```
bkonyi, line 202: Nit: use `EdgeInsets.zero` here and below.

bkonyi (approved):

PR #5455, opened 2023-03-20
Point users to Impeller from shader compilation warnings on iOS

kenzieschmoll:
Screenshot 2023-03-20 at 1 29 23 PM

The note suggesting that users try Impeller will only show up when the connected app is running on iOS. The Impeller link takes users to the [wiki](https://github.com/flutter/flutter/wiki/Impeller).

Fixes https://github.com/flutter/devtools/issues/5349

CC @chinmaygarde @dnfield @jonahwilliams

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/performance_utils.dart
+++ b/packages/devtools_app/lib/src/screens/performance/performance_utils.dart
@@ -87,6 +87,8 @@ void debugTraceEventCallback(VoidCallback callback) {
 
 const preCompileShadersDocsUrl = 'https://docs.flutter.dev/perf/shader';
 
+const impellerWikiUrl = 'https://github.com/flutter/flutter/wiki/Impeller';
+
 extension TraceEventExtension on TraceEvent {
   bool get isThreadNameEvent =>
       phase == TraceEvent.metadataEventPhase &&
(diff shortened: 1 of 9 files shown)


packages/devtools_app/lib/src/screens/performance/performance_utils.dart
```diff
@@ -87,6 +87,8 @@ void debugTraceEventCallback(VoidCallback callback) {
 
 const preCompileShadersDocsUrl = 'https://docs.flutter.dev/perf/shader';
 
+const impellerWikiUrl = 'https://github.com/flutter/flutter/wiki/Impeller';
```
jonahwilliams, line 90: @zanderso / @chinmaygarde are we keeping the wiki as the canonical source of documentation for Impeller?

jonahwilliams (approved):
LGTM with question

packages/devtools_app/lib/src/screens/performance/performance_utils.dart
kenzieschmoll: If there is a better link to point users to, let me know and I will change in a follow up PR. Thanks.

zanderso: Yeah the wiki link is fine for now. If/when we shift to flutter.dev, we'll file an issue to update and also put a forwarding link in the wiki.

Issue #5425, opened 2023-03-15
Investigate logging of errors in futures


CoderDake:
While developing the logging code, I discovered that an error happening in `getFullRequestData` was not being reported to the error logging.

After some research I found out that using `unawaited` is only ever meant to be used on code that is always excepted to return a value. It's documentation states:
```
If the future completes with an error,
/// it was likely a mistake to not await it.
/// That error will still occur and will be considered unhandled
/// unless the same future is awaited (or otherwise handled) elsewhere too.
/// Because of that, `unawaited` should only be used for futures that
/// are *expected* to complete with a value.
```

As a result of this research we should investigate our use of unawaited, and whether errors thrown in futures will be logged

CoderDake:
It may also be worth logging any javascript errors. I'm not sure if this is possible though.
example:

CoderDake:
After doing some research it appears that this may be a web specific bug.
I haven't been able to create a minimum reproduction of it yet though.

CoderDake:
Reminder: Tag @elliette if I end up making a dart-sdk bug for this.

elliette:
Note: We "solved" this in DWDS by creating a helper function to catch any exceptions thrown in `unawaited` futures. See https://github.com/dart-lang/webdev/pull/1938

```

void safeUnawaited(
  Future<void> future, {
  void Function(dynamic, StackTrace)? onError,
}) {
  onError ??= (error, stackTrace) =>
      _logger.warning('Error in unawaited Future:', error, stackTrace);
  unawaited(future.catchError(onError));
}
```

CoderDake:
It looks like #5456 addressed this issue. I've tested with other unawaited sections and the errors seem to now get appropriately reported.
While investigating this I found that we should be adding a newline to the last log entry when copying though.
That has been addressed here: https://github.com/flutter/devtools/pull/5538

PR #5456, opened 2023-03-20
Fix async callback errors escaping guarded zone

CoderDake:
Relates to https://github.com/flutter/devtools/issues/5425#issue-1625723424
Fixes https://github.com/flutter/devtools/issues/5458

This PR fixes an issue where errors thrown in futures, or async functions might not be caught by the `runGuardedZoned` zone.

As it turns out [SyntaxHighlighter.initialize()](https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/debugger/syntax_highlighter.dart#L41) makes a call to `WidgetsFlutterBinding.ensureInitialized()`, which means that `WidgetsBinding.instance` used to get initialized outside of the `runZonedGuarded`.  This resulted in the guarded zone not catch async errors from that WidgetBinding.

To overcome this I've wrapped  all of the initialization code in the app error handling (to make sure that the error handling code always gets the first call to `WidgetsFlutterBinding.ensureInitialized()`).
I have also added an assert to ensure that any main.dart file that tries to `runDevTools`, hasn't already initialized the `WidgetsBinding` either.

### DIFF
--- a/packages/devtools_app/lib/src/framework/app_error_handling.dart
+++ b/packages/devtools_app/lib/src/framework/app_error_handling.dart
@@ -62,14 +62,14 @@ void setupErrorHandling(Future Function() appStartCallback) {
 /// Checks to see if [WidgetsBinding.instance] has already been set.
 ///
 /// If calling [WidgetsBinding.instance] throws a 'Binding has not yet been
-/// initialized' error, then we know that it has not bee initialized yet.
+/// initialized' error, then we know that it has not been initialized yet.
 /// Otherwise someone may have already called
 /// [WidgetsFlutterBinding.ensureInitialized]. This is not allowed since
 /// it will prevent us from logging async errors inside [runZonedGuarded]
 bool _widgetsBindingInstanceIsUninitialized() {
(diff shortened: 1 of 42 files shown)


packages/devtools_app/lib/src/framework/app_error_handling.dart
```diff
@@ -57,6 +59,27 @@ void setupErrorHandling(Future Function() appStartCallback) {
...
+///
+/// If calling [WidgetsBinding.instance] throws a 'Binding has not yet been
+/// initialized' error, then we know that it has not bee initialized yet.
+/// Otherwise someone may have already called
+/// [WidgetsFlutterBinding.ensureInitialized]. This is not allowed since
+/// it will prevent us from logging async errors inside [runZonedGuarded]
+bool _widgetsBindingInstanceIsUninitialized() {
+  try {
+    // WidgetsBinding.instance should throw an error here since we expect
+    // it to uninitialized.
```
kenzieschmoll, line 72: nit: to be

PR #5460, opened 2023-03-21
Fix a performance regression in timeline event processing

kenzieschmoll:
We were previously doing a O(n) search to see if the thread name event is a duplicate. After this change, the duplicate check takes constant time O(1).

This was bringing the performance page to a crawl when N got large.

Work towards https://github.com/flutter/devtools/issues/5091

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/panes/timeline_events/timeline_events_controller.dart
+++ b/packages/devtools_app/lib/src/screens/performance/panes/timeline_events/timeline_events_controller.dart
@@ -64,9 +64,9 @@ class TimelineEventsController extends PerformanceFeatureController
 
   /// Set of thread_name trace events.
   ///
-  /// These events are returned with each [VMService.getVMTimeline] response,
-  /// and we do not want to store duplicates in [allTraceEvents].
-  final threadNameTraceEvents = <TraceEvent>{};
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/screens/performance/panes/timeline_events/timeline_events_controller.dart
```diff
@@ -235,9 +231,8 @@ class TimelineEventsController extends PerformanceFeatureController
     // available in the engine.
     int? uiThreadId;
     int? rasterThreadId;
-    for (TraceEvent event in threadNameEvents) {
-      final name = event.args!['name'] as String;
-
+    for (ThreadNameEvent event in threadNameEvents) {
+      final name = event.name!;
```
CoderDake, line 235: Is name guaranteed to be set? If so maybe it shouldn't be optional

CoderDake (approved):

packages/devtools_app/lib/src/screens/performance/panes/timeline_events/timeline_events_controller.dart
bkonyi: As far as I know, the name should always be present.

PR #5471, opened 2023-03-22
Fix references for grouping.

polina-c:

before:

Screenshot 2023-03-21 at 9 12 38 PM

after:

Screenshot 2023-03-21 at 9 11 44 PM


packages/devtools_app/lib/src/shared/diagnostics/tree_builder.dart
```diff
@@ -178,6 +180,7 @@ Future<void> _addChildrenToInstanceVariable({
   required Instance value,
   required IsolateRef? isolateRef,
   Set<String>? existingNames,
+  required HeapObjectSelection? heapSelection,
```
bkonyi, line 183: Nit: move up with the other required parameters.

bkonyi (approved):

Issue #5176, opened 2023-02-02
Table flickers during refresh in Profile Memory view


InMatrix:
Since the setting “Refresh on GC” was on by default, the table often re-rendered itself when I was in the middle of reading some numbers. The content of the whole table would all disappear for a moment and then come back. This introduced enough disruption to make me lose track of what I was trying to see.

![Image](https://user-images.githubusercontent.com/348942/216459754-880f5734-7555-42e7-aba2-0daf3e6bc5fc.gif)

Recommendations:
* [Ideal solution] Updating table cells in place continuously, matching the UX of macOS Activity Monitor and Chrome Task Manager
* [Workaround] Make "Refresh on GC" off by default

PR #5472, opened 2023-03-22
Stop auto-refreshing on GC to avoid screen blinking.

polina-c:

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/panes/profile/profile_pane_controller.dart
+++ b/packages/devtools_app/lib/src/screens/memory/panes/profile/profile_pane_controller.dart
@@ -24,7 +24,7 @@ class ProfilePaneController extends DisposableController
   /// Specifies if the allocation profile should be refreshed when a GC event
   /// is received.
   ValueListenable<bool> get refreshOnGc => _refreshOnGc;
-  final _refreshOnGc = ValueNotifier<bool>(true);
+  final _refreshOnGc = ValueNotifier<bool>(false);
 
   bool _initialized = false;
 


bkonyi:
Do we have an issue to track this? Disabling refresh on GC by default will make it less visible, but the issue is still there.

We should show a spinner on top of the old data when we're reloading the allocation profile, but I think we end up clearing the data backing the table during the refresh and replacing it which is what's causing the flicker.

polina-c:
Yes: https://github.com/flutter/devtools/issues/5176

The UX is weird, especially knowing 'Profile' is default tab on the memory screen.
Yes, it is possible to design better fix, but we do not have capacity for this before Flutter 3 cut.

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_pane_controller.dart
```diff
@@ -24,7 +24,7 @@ class ProfilePaneController extends DisposableController
   /// Specifies if the allocation profile should be refreshed when a GC event
   /// is received.
   ValueListenable<bool> get refreshOnGc => _refreshOnGc;
-  final _refreshOnGc = ValueNotifier<bool>(true);
+  final _refreshOnGc = ValueNotifier<bool>(false);
```
bkonyi, line 27: Leave a comment referencing https://github.com/flutter/devtools/issues/5176 and stating that this should be turned back on by default once it's resolved.

bkonyi (approved):
Okay, just wanted to make sure. LGTM once we add a comment referencing the issue.

Issue #5192, opened 2023-02-05
Network Response "Copy to Clipboard" doesn't work


hansmboron:
<-- Please describe your problem here. Be sure to include repro steps. -->
___
**DevTools version**: 2.20.1
**Connected Device**:
CPU / OS: arm64 (64 bit) android
Dart Version: 2.19.1
Flutter Version: 3.7.1 / stable
Framework / Engine: 7048ed95a5 / 800594f1f4
Connected app type: Flutter native (debug build)

CoderDake:
@hansmboron could you please provide more information about which copy button wasn't working for you?

hansmboron:
![image](https://user-images.githubusercontent.com/47091709/217058659-a24da7dc-dd35-4f06-867c-f3578f8d1d7d.png)
The icon from the Response in a network request. I was using Linux and now using Windows don't copy nothing too.

CoderDake:
Thank you i'll take a look into this once we fix some other issues in the network tab

PR #5476, opened 2023-03-22
Fix network copy bug

CoderDake:
Fixes https://github.com/flutter/devtools/issues/5192

I noticed while testing the network response and requests tabs that one could click copy while the data was still loading.
this could result in empty copies. 

Another issue I seemed to intermittently see was that the data would still be empty even after data was loaded. This only happened intermittently thought.

This PR changes the copy buttons so that they won't show until there is data for them to copy, and also makes sure that they are more tightly coupled with the state of data being fetched.

### DIFF
--- a/packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
+++ b/packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
@@ -178,6 +178,7 @@ class HttpViewTrailingCopyButton extends StatelessWidget {
   const HttpViewTrailingCopyButton(this.data, this.dataSelector);
   final DartIOHttpRequestData data;
   final String? Function(DartIOHttpRequestData) dataSelector;
+
   @override
   Widget build(BuildContext context) {
     return ValueListenableBuilder(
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
```diff
@@ -170,6 +170,35 @@ class HttpRequestView extends StatelessWidget {
...
+  final String? Function(DartIOHttpRequestData) dataSelector;
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder(
+      valueListenable: data.requestUpdatedNotifier,
+      builder: (context, __, ___) {
+        final dataToCopy = dataSelector(data);
+        final isLoading = data.isFetchingFullData;
+        if (dataToCopy == null || dataToCopy == '' || isLoading)
+          return Container();
```
bkonyi, line 189: Nit: add braces around this return.

bkonyi (approved):

packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
```diff
@@ -170,6 +170,37 @@ class HttpRequestView extends StatelessWidget {
   }
 }
 
+/// A button for copying [DartIOHttpRequestData] contents.
+///
+/// If there is no content to copy, the button will not show. The copy contents
```
kenzieschmoll, line 175: instead of hiding the button, we should just disable the onPressed handler. This will keep the UI consistent and not confuse users why sometimes the button is there and sometimes it is not. This is pretty standard across the rest of DevTools. For example, we don't remove a pause button when we are already paused, we just disable the onPressed handler so it shows up as disabled.

CoderDake: SGTM
Issue created: https://github.com/flutter/devtools/issues/5519

PR #5481, opened 2023-03-22
Reject variable assignment for web applications.

polina-c:


packages/devtools_app/lib/src/shared/console/eval/eval_service.dart
```diff
@@ -177,12 +177,23 @@ class EvalService extends DisposableController with AutoDisposeControllerMixin {
...
   }
 
+  /// If scope is supported, returns null.
+  ///
+  /// Otherwise, returns a message explaining why scope is not supported.
+  String? messageIfScopeNotSupported() {
+    // Web does not support scopes yet.
+    final isWeb = serviceManager.connectedApp?.isDartWebAppNow ?? true;
+    if (isWeb) {
+      return 'Scope variables are not supported for web applications.';
```
kenzieschmoll, line 192: instead of returning a string, wdyt about emitting to the console from this method, and then returning a bool indicating whether or not scope is supported. Then we could just check the bool value returned `isScopeSupported` instead of checking the nullness of a string.

kenzieschmoll: we could add a parameter `emitWarningToConsole` if there are cases where we want to perform the check but not emit a message to console.

packages/devtools_app/lib/src/shared/console/widgets/evaluate.dart
```diff
@@ -414,12 +414,19 @@ class ExpressionEvalFieldState extends State<ExpressionEvalField>
...
   /// If [expressionText] is assignment like `var x=$1`, processes it.
   ///
-  /// Returns true is the text was parsed as assignment.
+  /// Returns true if the text was parsed as assignment.
   bool _tryProcessAssignment(String expressionText) {
     if (!FeatureFlags.evalAndBrowse) return false;
 
     final assignment = ConsoleVariableAssignment.tryParse(expressionText);
     if (assignment == null) return false;
+    const result = true;
```
bkonyi, line 423: Nit: if we're going to use a constant for the success case, let's just name this constant to represent that. Maybe something like `const kSuccess = true`?

packages/devtools_app/lib/src/shared/console/eval/eval_service.dart
bkonyi: +1, if we're only using this message in one spot at the moment, it'd be cleaner to have this method return a boolean and then print this message when that check returns false.

bkonyi (approved):

bkonyi (commented):
Undoing accidental approval... 🙃

bkonyi (approved):

PR #5491, opened 2023-03-23
Fix update_flutter_sdk.sh tagging for latest flutter candidate

CoderDake:
This PR fixes update_flutter_sdk.sh to use the new tags.
It also adds some validation to make sure the user has the right upstream ste.

RELEASE_NOTE_EXCEPTION=Fix for internal helper


tool/update_flutter_sdk.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -ex
```
CoderDake, line 1: -e only

kenzieschmoll (approved):

Issue #4108, opened 2022-05-17
Add test coverage for "Review History" functionality after app disconnect.


bkonyi:
It's often useful to be able to continue to inspect specific data in Observatory after the target application has closed (e.g., inspecting a CPU profile, timeline, etc). DevTools should continue to show the most recent data after an application has disconnected instead of blocking the entire screen with a "Disconnected" message.

jacob314:
This is supported by the memory page in DevTools but not other pages.

kenzieschmoll:
I'm going to land this in two parts:
1. support offline viewing after we have lost connection for pages that support viewing offline data
2. add test coverage after upcoming deadline (this feature is currently completely broken, so it will not be any more broken after my changes, and our timeline is running short)

kenzieschmoll:
This is resolved wrt to implementing the feature (at least to the extent that it can be right now with limited offline viewing support in general in DevTools). Re-purposing this issue to track adding test coverage, so removing the observatory labels and adjusting priority.

PR #5509, opened 2023-03-23
Add support for viewing data history after app disconnect

kenzieschmoll:
This adds support to "review history" after an app disconnect for screens that support offline data viewing. Right now, that is only the Performance page and the Cpu Profiler page, but when other pages support offline viewing, we will get this "review history" behavior for free.

The workflow is as follows:
1. User is looking at data on a screen that supports offline snapshots
2. App disconnects
3. Upon disconnect, the data from the current screen is saved as the current offline data in DevTools
4. If a user clicks "Review History", we navigate to the offline snapshot screen and load the data.

One other change is that the "Review History" button will only show up if there is some pending offline data to show.

Work towards https://github.com/flutter/devtools/issues/4108. I have a todo to add testing. As I explained in the issue, this feature is currently completely broken, so landing my change will not make this feature any more broken if something were to go wrong. With the upcoming deadline early next week, there are more important issues to address than adding this test, but I will come back to this and clean it up.


packages/devtools_app/lib/src/framework/initializer.dart
```diff
@@ -78,7 +82,11 @@ class _InitializerState extends State<Initializer>
           !connectionState.userInitiatedConnectionState) {
         // Try to reconnect (otherwise, will fall back to showing the
         // disconnected overlay).
-        unawaited(_attemptUrlConnection());
+        try {
+          unawaited(_attemptUrlConnection());
```
bkonyi, line 86: I don't think this try-catch does anything if you don't await `_attemptUrlConnection` since it'll throw an asynchronous error if it throws. You'll most likely want to use `Future.catchError` here instead.

packages/devtools_app/lib/src/framework/initializer.dart
```diff
@@ -155,8 +163,29 @@ class _InitializerState extends State<Initializer>
...
+  }
+
+  void _reviewHistory() {
+    assert(offlineController.offlineDataJson.isNotEmpty);
+
+    offlineController.enterOfflineMode(
+      offlineApp: offlineController.previousConnectedApp!,
+    );
+    hideDisconnectedOverlay();
+    final args = {
```
bkonyi, line 179: Nit: add type parameters here?

bkonyi (approved):
Generally LGTM

Issue #5426, opened 2023-03-15
Migrate `log` to `_log.info`


CoderDake:
In https://github.com/flutter/devtools/pull/5404 we are making the switch to creating `Logger`s for each file. The logging package will then decide what gets logged where. For example anything INFO and above will be logged to console.
For this reason we will need to migrate our usage of `log` to use the new `Loggers`.

CoderDake:
Talked KenZ about this, may need some more discussion around how we name and expose our [Logger]s

CoderDake:
- [ ] make sure to make the logger file private when we migrate everything.

PR #5516, opened 2023-03-24
Migrate to new logging system.

CoderDake:
Finish up the change to the new logging system. This PR changes all references of `log` to a call to a `Logger`.
Also an `avoid-print` lint has been added so that we will be incentivized to use the new logging system.

The [documentation for the logging package](https://pub.dev/packages/logging) mentions that each Logger should have ` a unique name to easily identify the source of the log messages.`

So for our purposes, each`Logger('<name>')` has  name a based on the name of the file where the Logger is created.

This should help give us some important insight to the location of the log message, especially when receiving logs from release versions of Devtools(where the stacktraces won't be giving us much useful information)

RELEASE_NOTE_EXCEPTION=Internal logging migration


packages/devtools_app/analysis_options.yaml
```diff
@@ -11,6 +11,10 @@ analyzer:
     # expected.
     - test/test_infra/test_data/syntax_highlighting/**
 
+linter:
```
kenzieschmoll, line 14: should this be added to the main analysis options file instead? or is it intentional we only want this in devtools_app
https://github.com/flutter/devtools/blob/master/analysis_options.yaml

kenzieschmoll: is there a way to exclude the test directory for this print?

packages/devtools_app/lib/src/framework/release_notes/release_notes.dart
```diff
@@ -20,6 +21,8 @@ import '../../shared/config_specific/server/server.dart' as server;
 import '../../shared/primitives/auto_dispose.dart';
 import '../../shared/theme.dart';
 
+final _log = Logger('lib/src/framework/release_notes/release_notes');
```
kenzieschmoll, line 24: using the full path will be painful when we end up refactoring things. What about using a logical name rather than a path name? e.g. "release notes", "framework core", "performance screen", etc

packages/devtools_app/lib/src/framework/release_notes/release_notes.dart
```diff
@@ -224,10 +227,9 @@ class ReleaseNotesController {
         // version of DevTools.
         _releaseNotesMarkdown.value = null;
         toggleReleaseNotesVisible(false);
-        logger.log(
+        _log.warning(
           'Warning: could not find release notes for DevTools version '
```
kenzieschmoll, line 231: can the "Warning:" prefix go away now? I believe `_log.warning` puts "[WARNING]" at the font of the log

packages/devtools_app/lib/src/screens/debugger/codeview_controller.dart
```diff
@@ -272,9 +274,9 @@ class CodeViewController extends DisposableController
           (last, e) => last..addAll(e.entries),
         ),
       );
-    } catch (e) {
+    } catch (e, st) {
       // Ignore - not supported for all vm service implementations.
-      log('$e');
+      _log.shout(e, e, st);
```
kenzieschmoll, line 279: should these be shouts or warnings if our intent is to ignore the exception?

packages/devtools_app/analysis_options.yaml
CoderDake: When adding it at the top level it required a lot more changes, so I wanted to keep it as relevant to our logging changes as possible.

I tried really hard to find a way to exclude directories here, but unfortunately it isn't supported :( So i opted to do the per-file ignores which I know is pretty lame :(
https://github.com/dart-lang/linter/issues/320

packages/devtools_app/lib/src/screens/performance/panes/raster_stats/raster_stats_controller.dart
```diff
@@ -38,7 +42,7 @@ class RasterStatsController extends PerformanceFeatureController {
       final rasterStats = RasterStats.parse(json);
       setData(rasterStats);
     } catch (e, st) {
-      logger.log('Error collecting raster stats: $e\n\n$st');
+      _log.info('Error collecting raster stats: $e', e, st);
```
kenzieschmoll, line 45: let's do .shout here

packages/devtools_app/lib/src/framework/release_notes/release_notes.dart
CoderDake: I had opted not to do that, to keep the number of changes in this PR down.
Would you be alright with me doing that as a follow up?

packages/devtools_app/lib/src/screens/debugger/codeview_controller.dart
CoderDake: oops yes thanks for catching these.

CoderDake:
Fixes https://github.com/flutter/devtools/issues/5426

Issue #5510, opened 2023-03-23
Exception thrown when taking heap snapshot


kenzieschmoll:
using dart_developer_timeline.dart sample app

```
html_dart2js.dart:39929 [zoneGuarded]: Bad state: No element
error @ html_dart2js.dart:39929
log @ logger_html.dart:18
(anonymous) @ logger_helpers.dart:34
_rootRunUnary @ zone.dart:1414
runUnary @ zone.dart:1307
runUnaryGuarded @ zone.dart:1216
[_sendData] @ stream_impl.dart:339
[_add] @ stream_impl.dart:271
[_sendData] @ broadcast_stream_controller.dart:377
add @ broadcast_stream_controller.dart:244
[_publish] @ logger.dart:287
log @ logger.dart:211
severe @ logger.dart:269
_reportError @ app_error_handling.dart:63
(anonymous) @ app_error_handling.dart:54
runBinary @ zone.dart:1665
(anonymous) @ zone.dart:1775
[_processUncaughtError] @ zone.dart:1080
handleUncaughtError @ zone.dart:1284
_propagateToListeners @ future_impl.dart:697
[_complete] @ future_impl.dart:575
(anonymous) @ future.dart:423
_rootRun @ zone.dart:1390
run @ zone.dart:1300
runGuarded @ zone.dart:1208
(anonymous) @ zone.dart:1248
_rootRun @ zone.dart:1398
run @ zone.dart:1300
(anonymous) @ zone.dart:1232
internalCallback @ isolate_helper.dart:48
setTimeout (async)
_isolate_helper.TimerImpl.new @ isolate_helper.dart:55
_createTimer @ async_patch.dart:188
_rootCreateTimer @ zone.dart:1473
createTimer @ zone.dart:1360
new @ timer.dart:54
delayed @ future.dart:421
delayToReleaseUiThread @ utils.dart:250
runBody @ async_patch.dart:84
_async @ async_patch.dart:123
delayToReleaseUiThread @ utils.dart:247
releaseUi @ utils.dart:273
_heapStatistics @ heap.dart:37
(anonymous) @ async_patch.dart:45
_rootRunUnary @ zone.dart:1406
runUnary @ zone.dart:1307
handleValue @ future_impl.dart:147
handleValueCallback @ future_impl.dart:784
_propagateToListeners @ future_impl.dart:813
[_complete] @ future_impl.dart:575
(anonymous) @ future.dart:423
_rootRun @ zone.dart:1390
[...]

PR #5520, opened 2023-03-24
Fix error and create error handling for snapshot taking.

polina-c:
Fixes https://github.com/flutter/devtools/issues/5510

https://user-images.githubusercontent.com/12115586/227653370-aab71e18-cd60-4ae0-afd8-98975e4e350f.mov


packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -46,36 +46,36 @@ class DiffPaneController extends DisposableController {
...
-    derived._updateValues();
+    try {
+      final heapData = await snapshotTaker.take();
+      await item.initializeHeapData(heapData);
+
+      final newElementIndex = snapshots.value.length - 1;
+      core._selectedSnapshotIndex.value = newElementIndex;
+    } catch (e) {
+      snapshots.remove(item);
+      core._selectedSnapshotIndex.value = 0;
```
kenzieschmoll, line 73: instead of 0 should this be the last valid snapshot in the list? For example, if i was looking at snapshot 2, and then took snapshot 3, I'd expect the index after error to be snapshot 2

kenzieschmoll: in fact, if we move line 69 and 70 into the finally block, that would allow us to remove this line entirely.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_list.dart
```diff
@@ -47,6 +48,19 @@ class _ListControlPane extends StatelessWidget {
...
   final DiffPaneController controller;
 
+  Future<void> _takeSnapshot(BuildContext context) async {
+    try {
+      await controller.takeSnapshot();
+    } catch (e, trace) {
+      await showDialog(
+        context: context,
+        builder: (context) => UnexpectedErrorDialog(
+          errorDetails: 'Error while taking snapshot.\n$e\n$trace',
```
bkonyi, line 58: Nit: "Encountered an error while taking a heap snapshot: ..."

packages/devtools_app/lib/src/shared/dialogs.dart
```diff
@@ -34,6 +39,44 @@ final dialogTextFieldDecoration = InputDecoration(
...
+    required this.errorDetails,
+  });
+
+  final String errorDetails;
+
+  @override
+  Widget build(BuildContext context) {
+    return DevToolsDialog(
+      title: const Text('Unexpected Error'),
+      content: Text(errorDetails),
```
bkonyi, line 55: The stack trace associated with the error should ideally be using our monospace styling.

bkonyi: Actually, don't we have a widget that can format and display stack traces? We should use that if possible.

packages/devtools_app/lib/src/shared/dialogs.dart
```diff
@@ -50,12 +50,18 @@ class UnexpectedErrorDialog extends StatelessWidget {
...
   Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+
     return DevToolsDialog(
       title: const Text('Unexpected Error'),
-      content: Text(errorDetails),
+      content: Text(
+        errorDetails,
+        style: theme.regularTextStyle
+            .apply(fontFamily: theme.fixedFontStyle.fontFamily),
```
kenzieschmoll, line 60: use `theme.fixedFontStyle`

packages/devtools_app/lib/src/shared/dialogs.dart
```diff
@@ -64,10 +70,10 @@ class UnexpectedErrorDialog extends StatelessWidget {
           ),
         ),
         DialogTextButton(
-          child: const Text('Create Issue'),
+          child: const Text('Create issue'),
           onPressed: () => unawaited(
             launchUrl(
-              devToolsExtensionPoints.issueTrackerLink().url,
+              newDevToolsIssueUri(issueDetails: errorDetails).toString(),
```
kenzieschmoll, line 76: we should use `devToolsExtensionPoints.issueTrackerLink()`, otherwise this will take g3 users to github, which we don't necessarily want.

packages/devtools_app/lib/src/shared/new_issue.dart
```diff
@@ -0,0 +1,45 @@
+// Copyright 2018 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: 2023

packages/devtools_app/lib/src/shared/new_issue.dart
```diff
@@ -0,0 +1,45 @@
...
+  final issueBody = issueBodyItems.join('\n');
+
+  return Uri.parse('https://$newDevToolsIssueUriDisplay').replace(
+    queryParameters: {
+      'body': issueBody,
+    },
+  );
+}
+
+List<String> _issueLinkDetails() {
```
kenzieschmoll, line 23: this needs to be public - it is used by the g3 implementation

packages/devtools_app/lib/src/shared/new_issue.dart
```diff
@@ -0,0 +1,45 @@
...
+
+import '../../devtools.dart' as devtools;
+import 'globals.dart';
+import 'utils.dart';
+
+const newDevToolsIssueUriDisplay = 'github.com/flutter/devtools/issues/new';
+
+Uri newDevToolsIssueUri({String? issueDetails}) {
+  final issueBodyItems = _issueLinkDetails();
+  if (issueDetails != null) issueBodyItems.add(issueDetails);
```
kenzieschmoll, line 13: nit: I would pre-pend the other details instead of appending at the end. That way in the github issue, the exception is first and the less pertinent devtools version information is at the bottom

packages/devtools_app/lib/src/shared/new_issue.dart
```diff
@@ -9,8 +9,8 @@ import 'utils.dart';
 const newDevToolsIssueUriDisplay = 'github.com/flutter/devtools/issues/new';
 
 Uri newDevToolsIssueUri({String? issueDetails}) {
```
kenzieschmoll, line 11: now that there is only one use of this `newDevToolsIssueUri` method, it might make make the code easier to follow to put it back in the `issueTrackerLink` method in `extensions_external.dart`. 

Perhaps a greater reason to do this: the code in `newDevToolsIssueUri` is specific to the external extension points, so it would be safer to have **only** the `extensions_external.dart` file have access to this helper. Otherwise, the internal extensions points file will import `new_issue.dart` and will have access to code that was only intended to be used in the external extension points.

 This will also make the code more parallel with how the g3 implementation of extension points is written.

packages/devtools_app/lib/src/shared/new_issue.dart
```diff
@@ -9,8 +9,8 @@ import 'utils.dart';
...
-  final issueBodyItems = _issueLinkDetails();
-  if (issueDetails != null) issueBodyItems.add(issueDetails);
+  final issueBodyItems = issueLinkDetails();
```
kenzieschmoll, line 12: nit: having the if check inside the list is a little cleaner. This can be written as follows :
```dart
final issueBodyItems = [
  if (issueDetails != null) issueDetails,
  ...issueLinkDetails(),
];
```

packages/devtools_app/lib/src/shared/new_issue.dart
```diff
@@ -0,0 +1,45 @@
...
+import 'utils.dart';
+
+const newDevToolsIssueUriDisplay = 'github.com/flutter/devtools/issues/new';
+
+Uri newDevToolsIssueUri({String? issueDetails}) {
+  final issueBodyItems = issueLinkDetails();
+  if (issueDetails != null) issueBodyItems.insert(0, issueDetails);
+  final issueBody = issueBodyItems.join('\n');
+
+  return Uri.parse('https://$newDevToolsIssueUriDisplay').replace(
```
CoderDake, line 16: It is possible for the url param to have too many characters. Should we try to limit the length of the body details in order to avoid hitting that?
https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-an-issue

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_list.dart
```diff
@@ -47,6 +48,20 @@ class _ListControlPane extends StatelessWidget {
 
   final DiffPaneController controller;
 
+  Future<void> _takeSnapshot(BuildContext context) async {
+    try {
+      await controller.takeSnapshot();
+    } catch (e, trace) {
```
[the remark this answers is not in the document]
polina-c: The error dialog box will give all the details, copyable to clipboard.

packages/devtools_app/lib/src/shared/dialogs.dart
```diff
@@ -34,6 +39,51 @@ final dialogTextFieldDecoration = InputDecoration(
   ),
 );
 
+/// A dialog, that reports unexpected error and allows to copy details and create issue.
+class UnexpectedErrorDialog extends StatelessWidget {
```
bkonyi, line 43: Can you add a screenshot of the updated dialog to the PR comments? :-)

packages/devtools_app/lib/src/shared/new_issue.dart
polina-c: Added limit. Thanks.

packages/devtools_app/lib/src/shared/dialogs.dart
polina-c: Screenshot 2023-03-28 at 10 32 47 AM

polina-c: It is scrollable when text is too large.

packages/devtools_app/lib/src/extension_points/extensions_external.dart
```diff
@@ -53,3 +46,47 @@ class ExternalDevToolsExtensionPoints implements DevToolsExtensionPoints {
...
+
+const _newDevToolsIssueUriDisplay = 'github.com/flutter/devtools/issues/new';
+
+Uri _newDevToolsGitHubIssueUri({String? issueDetails}) {
+  const _maxGitHubUriLength = 8190;
+
+  final issueBody = [
+    if (issueDetails != null) issueDetails,
+    ..._issueLinkDetails(),
+  ].join('\n');
```
kenzieschmoll, line 58: since there is a max URI length, we should make sure that the `issueLinkDetails` items are always included. So instead of trimming at the end, we should trim the `issueDetails` to fit with the complete content of `issueLinkDetails()`.

Better to cut off the end of a stack trace than the version / connected app information.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_list.dart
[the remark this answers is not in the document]
kenzieschmoll: we also want to log this to our analytics though. @CoderDake will _log.shout propagate the error to our analytics?

packages/devtools_app/lib/src/extension_points/extensions_external.dart
```diff
@@ -49,44 +49,43 @@ class ExternalDevToolsExtensionPoints implements DevToolsExtensionPoints {
 
 const _newDevToolsIssueUriDisplay = 'github.com/flutter/devtools/issues/new';
 
-Uri _newDevToolsGitHubIssueUri({String? issueDetails}) {
+Uri _newDevToolsGitHubIssueUriLengthSafe({String? issueDetails}) {
   const _maxGitHubUriLength = 8190;
+  final context = issueLinkDetails();
```
kenzieschmoll, line 54: avoid naming things 'context'. This is an loaded term in Flutter.

packages/devtools_app/lib/src/extension_points/extensions_external.dart
```diff
@@ -49,44 +49,43 @@ class ExternalDevToolsExtensionPoints implements DevToolsExtensionPoints {
...
+  final fullUri =
+      _newDevToolsGitHubIssueUri(issueDetails: issueDetails, context: context);
+
+  final lengthToCut = fullUri.toString().length - _maxGitHubUriLength;
+  if (lengthToCut <= 0) return fullUri;
+
+  if (issueDetails == null)
+    throw StateError(
+      'Issue details cannot be null, because length limit is reached.',
+    );
```
kenzieschmoll, line 65: this error is a bit confusing 🤔  instead of throwing an error, lets add the check for `issueDetails == null` to the if check on line 60.

`if (issueDetails == null || lengthToCut <=0) return fullUri;`

packages/devtools_app/lib/src/extension_points/extensions_external.dart
```diff
@@ -49,44 +49,43 @@ class ExternalDevToolsExtensionPoints implements DevToolsExtensionPoints {
 
 const _newDevToolsIssueUriDisplay = 'github.com/flutter/devtools/issues/new';
 
-Uri _newDevToolsGitHubIssueUri({String? issueDetails}) {
+Uri _newDevToolsGitHubIssueUriLengthSafe({String? issueDetails}) {
```
kenzieschmoll, line 52: nit: maybe rename issueDetails parameter to `additionalInfo` or similar. The names `issueDetails` and `issueLinkDetails` are similar enough that it makes this tough to follow.

packages/devtools_app/lib/src/extension_points/extensions_external.dart
```diff
@@ -49,44 +49,43 @@ class ExternalDevToolsExtensionPoints implements DevToolsExtensionPoints {
...
+    issueDetails: truncatedDetails,
+    context: context,
+  );
+  assert(truncatedUri.toString().length <= _maxGitHubUriLength);
+  return truncatedUri;
+}
+
+Uri _newDevToolsGitHubIssueUri({
+  String? issueDetails,
+  required List<String> context,
```
kenzieschmoll, line 79: same comment regarding 'context'. please rename. Also, style nit: place required parameters first.

packages/devtools_app/lib/src/extension_points/extensions_external.dart
```diff
@@ -19,8 +21,10 @@ class ExternalDevToolsExtensionPoints implements DevToolsExtensionPoints {
   Link issueTrackerLink({String? issueDetails}) {
     return Link(
       display: _newDevToolsIssueUriDisplay,
-      url: _newDevToolsGitHubIssueUriLengthSafe(issueDetails: issueDetails)
-          .toString(),
+      url: newDevToolsGitHubIssueUriLengthSafe(
+        additionalInfo: issueDetails,
```
kenzieschmoll, line 25: rename issueDetails to additonalInfo to match param

packages/devtools_app/lib/src/extension_points/extensions_external.dart
```diff
@@ -19,8 +21,10 @@ class ExternalDevToolsExtensionPoints implements DevToolsExtensionPoints {
...
+        environment: issueLinkDetails(),
```
kenzieschmoll, line 26: we don't need to pass this in as a parameter. The issueLinkDetails can be looked up directly from inside the `newDevToolsGitHubIssueUriLengthSafe` method

packages/devtools_app/test/extentsion_points/extension_external_test.dart
```diff
@@ -0,0 +1,20 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: nit name test so that it matches file name. Need an 's' after extension to match the file it is testing.

packages/devtools_app/lib/src/extension_points/extensions_external.dart
```diff
@@ -53,3 +50,52 @@ class ExternalDevToolsExtensionPoints implements DevToolsExtensionPoints {
...
+    additionalInfo: truncatedInfo,
+    environment: environment,
+  );
+  assert(truncatedUri.toString().length <= maxGitHubUriLength);
+  return truncatedUri;
+}
+
+Uri _newDevToolsGitHubIssueUri({
+  String? additionalInfo,
+  required List<String> environment,
```
kenzieschmoll, line 89: order required parameters first

packages/devtools_app/lib/src/extension_points/extensions_external.dart
polina-c: It is not very important which method makes this call.
I moved the call for testability: I do not want the test to worry about mocking environment.

kenzieschmoll (approved):

PR #5521, opened 2023-03-24
Make _MemoryChartPaneState.didChangeDependencies null safer.

polina-c:
RELEASE_NOTE_EXCEPTION=[REASON GOES HERE]

Fixes: https://github.com/flutter/devtools/issues/5497

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/panes/chart/chart_pane.dart
+++ b/packages/devtools_app/lib/src/screens/memory/panes/chart/chart_pane.dart
@@ -77,6 +77,8 @@ class _MemoryChartPaneState extends State<MemoryChartPane>
               : _hoverEventsHeight)
           : 0);
 
+  static int get _timestamp => DateTime.now().millisecondsSinceEpoch;
+
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
@@ -89,22 +91,25 @@ class _MemoryChartPaneState extends State<MemoryChartPane>
           _hideHover();
         }
         final tapLocation = widget.chartController.event.tapLocation.value;
-        if (tapLocation?.tapDownDetails != null) {
-          final tapData = tapLocation!;
-          final index = tapData.index;
-          final timestamp = tapData.timestamp!;
+        if (tapLocation == null) return;
+        final tapDownDetails = tapLocation.tapDownDetails;
+        if (tapDownDetails == null) return;
 
-          final copied = TapLocation.copy(tapLocation);
-          widget.chartController.vm.tapLocation.value = copied;
-          widget.chartController.android.tapLocation.value = copied;
+        final copied = TapLocation.copy(tapLocation);
+        widget.chartController.vm.tapLocation.value = copied;
+        widget.chartController.android.tapLocation.value = copied;
 
-          final allValues = ChartsValues(controller, index, timestamp);
-          _showHover(
-            context,
-            allValues,
-            tapData.tapDownDetails!.globalPosition,
-          );
-        }
+        final allValues = ChartsValues(
+          controller,
+          tapLocation.index,
+          tapLocation.timestamp ?? _timestamp,
+        );
+
+        _showHover(
+          context,
+          allValues,
+          tapDownDetails.globalPosition,
+        );
       }
     });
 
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/screens/memory/panes/chart/chart_pane.dart
```diff
@@ -89,22 +91,25 @@ class _MemoryChartPaneState extends State<MemoryChartPane>
...
-          final allValues = ChartsValues(controller, index, timestamp);
-          _showHover(
-            context,
-            allValues,
-            tapData.tapDownDetails!.globalPosition,
-          );
-        }
+        if (tapLocation == null) return;
+        final tapDownDetails = tapLocation.tapDownDetails;
+        if (tapDownDetails == null) return;
```
kenzieschmoll, line 96: shorten to: `if (tapLocation?.tapDownDetails == null) return;`

packages/devtools_app/lib/src/screens/memory/panes/chart/chart_pane.dart
```diff
@@ -114,50 +119,53 @@ class _MemoryChartPaneState extends State<MemoryChartPane>
...
+        controller,
+        tapLocation.index,
+        tapLocation.timestamp ?? _timestamp,
+      );
+
+      _showHover(
+        context,
+        allValues,
+        tapDownDetails.globalPosition,
+      );
```
kenzieschmoll, line 168: use helper here too

packages/devtools_app/lib/src/screens/memory/panes/chart/chart_pane.dart
bkonyi: FWIW, I think I'd rather stick with `if (tapLocation?.tapDownDetails == null) return;` and then change line 98 to `final copied = TapLocation.copy(tapLocation!);` since that will only introduce one extra null check. Then line 111 could just be `tapLocation.tapDownDetails!.globalPosition`.

packages/devtools_app/lib/src/screens/memory/panes/chart/chart_pane.dart
```diff
@@ -114,50 +119,53 @@ class _MemoryChartPaneState extends State<MemoryChartPane>
...
-
-          _showHover(
-            context,
-            allValues,
-            tapData.tapDownDetails!.globalPosition,
-          );
-        }
+        if (tapLocation == null) return;
+        final tapDownDetails = tapLocation.tapDownDetails;
+        if (tapDownDetails == null) return;
```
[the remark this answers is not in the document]
polina-c: We have TODO for this on line 87

packages/devtools_app/lib/src/screens/memory/panes/chart/chart_pane.dart
polina-c: We have TODO for this on line 87

packages/devtools_app/lib/src/screens/memory/panes/chart/chart_pane.dart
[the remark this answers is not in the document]
kenzieschmoll: In the spirit of yak-shaving, it wouldn't hurt to complete the TODO in this PR. But not going to block on it.

packages/devtools_app/lib/src/screens/memory/panes/chart/chart_pane.dart
polina-c: This PR is to fix null reference error. It is not clear where is it originated. So, I prefer to avoid any bungs in this code.

packages/devtools_app/lib/src/screens/memory/panes/chart/chart_pane.dart
```diff
@@ -79,94 +79,55 @@ class _MemoryChartPaneState extends State<MemoryChartPane>
...
+    List<ValueNotifier<TapLocation?>> allLocations,
+  ) {
+    addAutoDisposeListener(tapLocation, () {
+      final value = tapLocation.value;
+      if (value == null) return;
 
-    addAutoDisposeListener(widget.chartController.android.tapLocation, () {
       if (_hoverOverlayEntry != null) {
         _hideHover();
+        return;
```
kenzieschmoll, line 92: this added return is a logic change - do we want to return early after we hide the hover card? we didn't before adding this helper

polina-c: It seems we had bug. You needed to press 'Esc' to hide the popup window. Now it hides on click at the chart.

Issue #5335, opened 2023-03-01
Embedded DevTools Debugger in chrome is not dense


kenzieschmoll:
The inspector appears to be properly in dense mode,  but the debugger does not. We either need to ensure dense mode is enabled, or tweak the debugger to have more "dense-aware" UI.

Separate but related issue - we probably want to expose the DevTools settings button in the footer like we do for the report feedback and about devtools buttons in embedded mode. 

![Screenshot 2023-03-01 at 10 15 42 AM](https://user-images.githubusercontent.com/43759233/222227852-34574d0c-264c-4a85-b68c-29730fb12f78.png)

elliette:
Note: Seems like the logging page also doesn't have "dense mode"

PR #5524, opened 2023-03-24
Use `CheckboxSettings` from `shared/common_widgets` in the settings dialog

elliette:
Work towards https://github.com/flutter/devtools/issues/5335 (specifically working towards "Separate but related issue - we probably want to expose the DevTools settings button in the footer like we do for the report feedback and about devtools buttons in embedded mode.")

 `RELEASE_NOTE_EXCEPTION="No user facing changes"`


packages/devtools_app/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -215,7 +205,6 @@
 			isa = PBXNativeTarget;
 			buildConfigurationList = 331C80DE294CF71000263BE5 /* Build configuration list for PBXNativeTarget "RunnerTests" */;
 			buildPhases = (
-				2B0491DF3BA97DDCC97A774C /* [CP] Check Pods Manifest.lock */,
```
CoderDake, line 218: Are these pbxproj changes intentional?

CoderDake (approved):

packages/devtools_app/macos/Runner.xcodeproj/project.pbxproj
elliette: Kind of - this file changes rather frequently as a result of running on `macos` (see https://github.com/flutter/devtools/commits/b398f1585348a921699dde42387d4459353ec3f3/packages/devtools_app/macos/Runner.xcodeproj/project.pbxproj) . I believe there is a reason we don't gitignore it but I forget why - @kenzieschmoll ?

PR #5533, opened 2023-03-28
Improve the performance of generating a bottom up profile

kenzieschmoll:
This improves the performance of generating a bottom up profile by ~60%.

We were using `List.remove`, which triggered a call to `List.copy`. This was quite expensive. Now we build up a new list of `mergedRoots` instead of  calling `remove` on the original list, and call `clear` + `addAll` after we are done merging roots. 

We also limit the list traversal using a `traverseIndex` and limit the calls to `CpuStackFrame.matches`, which can also be expensive for a large N.

Work towards https://github.com/flutter/devtools/issues/5091


bkonyi (approved):
Awesome improvement! LGTM!

CoderDake (approved):

PR #5538, opened 2023-03-28
Each log entry should be on its own line

CoderDake:
The last line of log entries doesn't have a new line at the end so `lnav` can get confused.
Adding this small fix to address this

RELEASE_NOTE_EXCEPTION=Small fix for logging entries


packages/devtools_app/lib/src/shared/log_storage.dart
```diff
@@ -45,7 +45,8 @@ class LogStorage {
             if (e.stackTrace != null) 'stackTrace': e.stackTrace.toString(),
           }),
         )
-        .join('\n');
+        .map((line) => line + '\n') // Each entry needs to be on it's own line
+        .join();
```
kenzieschmoll, line 49: what about adding an extension method `joinWithTrailing(Pattern pattern)` that calls join and then appends one additional pattern at the end? We have `StringExtension` in utils.dart you can add this to.

packages/devtools_app/lib/src/shared/log_storage.dart
```diff
@@ -6,6 +6,8 @@ import 'dart:collection';
 import 'dart:convert';
 import 'package:logging/logging.dart';
 
+import '../../devtools_app.dart';
```
kenzieschmoll, line 9: don't import the devtools_app library - import used libraries directly

packages/devtools_app/lib/src/shared/primitives/utils.dart
```diff
@@ -1180,6 +1180,13 @@ extension StringExtension on String {
   }
 }
 
+extension IterableExtension<T> on Iterable<T> {
+  /// Joins the iterable with [separator], and also adds a trailing [separator].
+  String joinWithTrailing([String separator = '']) {
+    return '${join(separator)}\n';
```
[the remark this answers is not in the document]
CoderDake: oops yep! I was moving a bit too fast on this one 😜

packages/devtools_app/lib/src/shared/primitives/utils.dart
```diff
@@ -1180,6 +1180,13 @@ extension StringExtension on String {
   }
 }
 
+extension IterableExtension<T> on Iterable<T> {
+  /// Joins the iterable with [separator], and also adds a trailing [separator].
+  String joinWithTrailing([String separator = '']) {
+    return '${join(separator)}$separator';
```
kenzieschmoll, line 1186: on second thought, should this return the empty string if the list is empty?

kenzieschmoll (approved):

PR #5541, opened 2023-03-28
Add analytics to eval and browse.

polina-c:

### DIFF
--- a/packages/devtools_app/lib/src/shared/analytics/constants.dart
+++ b/packages/devtools_app/lib/src/shared/analytics/constants.dart
@@ -175,13 +175,17 @@ const exitOfflineMode = 'exitOfflineMode';
 // the page data has loaded and is ready to interact with.
 const pageReady = 'pageReady';
 
-/// Console area.
-const consoleHelp = 'consoleHelp';
-
 /// Documentation actions shared across screens.
 const documentationLink = 'documentationLink';
 String topicDocumentationButton(String topic) => '${topic}DocumentationButton';
 String topicDocumentationLink(String topic) => '${topic}DocumentationLink';
+
+/// Analytic event constants specific for console.
+class ConsoleEvent {
+  static const helpInline = 'consoleHelpInline';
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/shared/analytics/constants.dart
```diff
@@ -245,6 +250,14 @@ class MemoryEvent {
...
 
   static const browseRefLimit = 'browseRefLimit';
+
+  static const dropOneLiveVariable = 'dropOneLiveVariable';
+  static const dropOneStaticVariable = 'dropOneStaticVariable';
+  static String dropAllLiveToConsole({
+    required bool includeSubclasses,
+    required bool includeImplementers,
+  }) =>
+      'dropAllVariables${includeSubclasses ? 'Subclasses' : ''}${includeImplementers ? 'Imlementers' : ''}';
```
kenzieschmoll, line 260: nit: add an underscore before Subclasses and Implementers so there is some separation

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -877,13 +877,17 @@ class ToolbarAction extends StatelessWidget {
     Key? key,
     this.size,
     this.style,
+    this.gaScreen,
+    this.gaSelection,
   }) : super(key: key);
```
kenzieschmoll, line 882: add an assert: `assert((gaScreen == null) == (gaSelection == null))`

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -893,7 +897,12 @@ class ToolbarAction extends StatelessWidget {
         tapTargetSize: MaterialTapTargetSize.shrinkWrap,
         textStyle: style,
       ),
-      onPressed: onPressed,
+      onPressed: () {
+        if (gaScreen != null || gaSelection != null) {
```
kenzieschmoll, line 901: change from || to &&

assert above will verify we pass both or neither

kenzieschmoll (approved):

Issue #2763, opened 2021-03-05
Dvorak WASD (,AOE) support in timeline view


Hixie:
Chrome's chrome://tracing, and Observatory, support ",aoe" navigation in the timeline view, which is useful for Dvorak users. (It's the same as WASD, `,`=W, `a`=A, `o`=S, `e`=D). It would be great if we could do the same.

devoncarew:
Contributions welcome :)

https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/charts/flame_chart.dart#L336

Hixie:
Is there a test I should update as well?

kenzieschmoll:
You can add a test similar to https://github.com/flutter/devtools/blob/b4d0613ab5f80a66f8e68bd9da0c434486e091ce/packages/devtools_app/test/flame_chart_test.dart#L50

Hixie:
thanks

Hixie:
It would be relevant for anywhere where the WASD keys navigate.

PR #5545, opened 2023-03-28
Add support for ,AOE (Dvorak) key navigation in the flame chart

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/2763. @Hixie

### DIFF
--- a/packages/devtools_app/lib/src/shared/charts/flame_chart.dart
+++ b/packages/devtools_app/lib/src/shared/charts/flame_chart.dart
@@ -455,10 +455,11 @@ abstract class FlameChartState<T extends FlameChart,
   KeyEventResult _handleKeyEvent(RawKeyEvent event) {
     // Only handle down events so logic is not duplicated on key up.
     if (event is RawKeyDownEvent) {
-      // Handle zooming / navigation from W-A-S-D keys.
+      // Handle zooming / navigation from WASD keys or ,AOE keys (Dvorak users)
       // TODO(kenz): zoom in/out faster if key is held. It actually zooms slower
       // if the key is held currently.
-      if (event.logicalKey == LogicalKeyboardKey.keyW) {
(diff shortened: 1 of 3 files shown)


polina-c:
Can you add video to demonstrate UX?
Should we link documentation/explanation for those who do not know what is AOE?

Hixie:
woo!

Hixie:
@polina-c `,AOE` are the four keys in the same position in a Dvorak key layout as `WASD` on a QWERTY key layout.
https://en.wikipedia.org/wiki/Dvorak_keyboard_layout

CoderDake (approved):

packages/devtools_app/lib/src/shared/charts/flame_chart.dart
```diff
@@ -455,10 +455,11 @@ abstract class FlameChartState<T extends FlameChart,
   KeyEventResult _handleKeyEvent(RawKeyEvent event) {
     // Only handle down events so logic is not duplicated on key up.
     if (event is RawKeyDownEvent) {
-      // Handle zooming / navigation from W-A-S-D keys.
+      // Handle zooming / navigation from WASD keys or ,AOE keys (Dvorak users)
       // TODO(kenz): zoom in/out faster if key is held. It actually zooms slower
       // if the key is held currently.
-      if (event.logicalKey == LogicalKeyboardKey.keyW) {
+      if (event.logicalKey == LogicalKeyboardKey.keyW ||
+          event.logicalKey == LogicalKeyboardKey.comma) {
```
bkonyi, line 462: Does this mean that ,OE will also allow for navigation by QWERTY users? If so, we should add a setting that allows for this to be toggled if we're unable to detect keyboard layout.

bkonyi (changes requested):

PR #5631, opened 2023-04-10
Polish perfetto CSS for theme switches

kenzieschmoll:
Reloads the CSS on theme switches so CSS constants in Perfetto or reinitialized for the new stylesheet. This PR also polishes some of the CSS for better dark & light mode styling.

RELEASE_NOTE_EXCEPTION=[too small to mention in release notes]


polina-c:
What if Perfetto updates their original packages?
It would make sense to document which changes to apply after loading their updated code. For example, we can list PR's that represent the changes.

PR #5637, opened 2023-04-11
Rename perfetto_compiled to perfetto_ui_compiled to match g3 config

kenzieschmoll:
No functional changes - just a rename.


.gitignore
```diff
@@ -29,3 +29,6 @@ packages/devtools_test/lib/src/mocks/generated.mocks.dart
 **/test/**/failures/
 *.dill
 DEPENDENCIES.md
+
+# Ignore logs generated by VS Code
+**/logs/**
```
elliette, line 34: Is this change intentional?

elliette (approved):

.gitignore
kenzieschmoll: Yeah I can break it out to a separate PR if we want, but I accidentally committed some logs that VS code created in my first commit, so just figured I'd add it while I was here.

PR #5639, opened 2023-04-11
Releasing from 2.24.0-dev.1 to 2.24.0

polina-c:
RELEASE_NOTE_EXCEPTION=Version Bump


kenzieschmoll:
What is this PR for? We shouldn't be doing the 2.24.0 release for another month or so.

PR #5646, opened 2023-04-12
Remove Inspector use of `getRootRenderObject` and `getSelectedRenderObject`

kenzieschmoll:
this will simplify multi-window support.
CC @goderbauer


jacob314 (approved):
lgtm

Issue #5479, opened 2023-03-22
Setup DCM for bots


CoderDake:

polina-c:
CC @incendial

polina-c:
- [ ] configure keys (@CoderDake )
- [ ] configure bot (draft: https://github.com/flutter/devtools/pull/5582)
- [ ] create README on how to maintain keys (@CoderDake )
- [ ] add steps on how to turn on DCM locally (@incendial )
- [ ] enable/fix first set of lints (@polina-c https://github.com/flutter/devtools/pull/5623)
- [ ] enable/fix other valuable lints

polina-c:
https://github.com/flutter/devtools/pull/5655

Issue #5636, opened 2023-04-11
Run DCM as a Github workflow in a PR from a forked repository 


elliette:
All DevTools development work is done in a fork of DevTools, instead of branching off the `flutter/devtools` repo. This means that PRs to DevTools are opened from the contributor's fork. 

Unfortunately, this is causing problems when trying to add DCM analyze as a Github workflow step when a PR is opened:

-  I believe the only way to share the email / key with the workflow (and not leak it) is to use encrypted secrets. This is what the `${{ secrets.DCM_CI_KEY }}` refers to here: ttps://github.com/flutter/devtools/pull/5582
- But Github does not allow secrets to be run on workflows triggered from a PR coming from a fork of the repo:

> "With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository." [source](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow)

@incendial We'll need a way to support running `dcm analyze` for Github workflows coming from a forked repository.

incendial:
@elliette any chance you'd be comfortable moving DCM checks to [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) event instead of `pull_request`? Since DCM does not build or run the code looks like using this event is safe.

elliette:
Hi @incendial! That should work for us, thanks! However, I'm now running into a different issue while setting up DCM (on DevTools and on https://github.com/dart-lang/webdev/tree/master/dwds). I've filed https://github.com/dart-code-checker/dart-code-metrics/issues/1220, thanks!

incendial:
> That should work for us, thanks! 

Happy to hear that! 🎉

> However, I'm now running into a different issue while setting up DCM

Sorry for inconvenience, in the latest release the underlying caching was refactored a bit too much, deleting `.dartServer/dart-code-metrics-teams/` should help. Won't happen again.

PR #5582, opened 2023-03-31
Adding DCM to workflows

CoderDake:
This PR sets up a workflow for Dart Code Metrics

Addresses https://github.com/flutter/devtools/issues/5479

RELEASE_NOTE_EXCEPTION=Internal process only

# To Do
- [ ] Code Analysis needs to be fixed before we can merge this

PR #5629, opened 2023-04-10
docs: add dcm activation guide to contributing guidelines

incendial:
RELEASE_NOTE_EXCEPTION=Internal process only

Addresses https://github.com/flutter/devtools/issues/5479

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [X] I signed the [CLA].
- [X] I listed at least one issue that this PR fixes in the description above.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [-] I added new tests to check the change I am making, or there is a reason for not adding tests.

PR #5650, opened 2023-04-13
Add Github workflow for Dart Code Metrics

elliette:
- Taking over from https://github.com/flutter/devtools/pull/5582 while @CoderDake is on leave. 
- Resolves https://github.com/flutter/devtools/issues/5479, https://github.com/flutter/devtools/issues/5636

Before submitting, we need to:

- [x] Add new DCM_CI_KEY and DCM_EMAIL secrets to repo settings (see b/278604924)
- [x] Switch `pull_request` to `pull_request_target` (this is less secure, but will allow the workflow to be run from PRs opened from forks)
- [x] Verify that we have the setting "Approval necessary for first time contributors to run workflows" enabled (see https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks#about-workflow-runs-from-public-forks, this addresses the security vulnerabilities of using `pull_request_target). Note: This is the default setting, but would be good to confirm (see b/278604924)

See https://github.com/flutter/devtools/actions/runs/4748777461/jobs/8435356187 for previous workflow run (using `pull_request` trigger)


.github/workflows/dcm.yaml
```diff
@@ -0,0 +1,38 @@
...
+# license of DCM, and will need to be changed if we decide to purchase an
+# actual license. The trial license will expire on April 11, 2024. 
+
+# If you need to install and run DCM locally, please refer to CONTRIBUTING.md
+# for instructions.
+
+name: Dart Code Metrics
+
+on:
+  pull_request:
```
elliette, line 16: This will need to be changed to `pull_request_target` before submitting to run on PRs opened from forks (see https://github.com/flutter/devtools/issues/5636). However, to trigger workflows before a PR has been submitted, we need to use `pull_request` due to laxer security for `pull_request_target`: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

.github/workflows/dcm.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+# NOTE: DCM usage in DevTools is currently experimental! 
+
+# The DCM_CI_KEY and DCM_EMAIL secrets are set in the admin settings page
+# of https://github.com/flutter/devtools. These are associated with a trial
+# license of DCM, and will need to be changed if we decide to purchase an
+# actual license. The trial license will expire on April 11, 2024. 
+
+# If you need to install and run DCM locally, please refer to CONTRIBUTING.md
+# for instructions.
```
elliette, line 11: Once https://github.com/flutter/devtools/pull/5629 is submitted

.github/workflows/dcm.yaml
```diff
@@ -0,0 +1,38 @@
+# A CI workflow to run the Dart Code Metrics analyzer (https://dcm.dev).
+
+# NOTE: DCM usage in DevTools is currently experimental! 
+
+# The DCM_CI_KEY and DCM_EMAIL secrets are set in the admin settings page
+# of https://github.com/flutter/devtools. These are associated with a trial
+# license of DCM, and will need to be changed if we decide to purchase an
+# actual license. The trial license will expire on April 11, 2024.
```
Hixie, line 8: how would someone who needs to set things up again get a copy of these secrets?

elliette: @jacob314 For this, should we say reach out to you or file a bug to get access to the secrets?

elliette: We chatted about how to handle this at DevTools standup today. Added a link to the secrets to the comment here, and added suggestions to https://github.com/flutter/devtools/pull/5629

tool/bots.sh
```diff
@@ -152,11 +152,6 @@ $ dart run integration_test/run_tests.dart --headless"
     # TODO(https://github.com/flutter/devtools/issues/1987): rewrite integration tests.
     dart run integration_test/run_tests.dart --headless
 
-else
```
elliette, line 155: Treat `BOT` as an optional input, if no bot configuration is provided, only run the shared part of `bots.sh`.

polina-c (approved):

PR #5655, opened 2023-04-14
Fix some dcm lints.

polina-c:


packages/.vscode/settings.json
```diff
@@ -2,4 +2,8 @@
     // Set current working directory to devtools_app.
     "terminal.integrated.cwd": "devtools_app",
     "dart.showTodos": false,
+    "editor.codeActionsOnSave": {
+        "source.fixAll": true,
+        "source.dcm.fixAll": true
```
kenzieschmoll, line 7: Do we really trust these enough to automatically apply fixes on save? I'd worry that fixes for false positives would slip in and potentially cause issues.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/sampler.dart
```diff
@@ -54,13 +54,15 @@ class HeapClassSampler extends ClassSampler {
...
-      _outputError(error, trace);
+    } catch (error) {
+      _outputError(error);
       return null;
     }
   }
 
-  void _outputError(Object error, StackTrace trace) {
+  void _outputError(
+    Object error,
```
bkonyi, line 64: I think this change was a false positive. Line 66 should be `$trace` instead of `trace`.

packages/devtools_app/lib/src/framework/status_line.dart
```diff
@@ -159,7 +159,7 @@ class StatusLine extends StatelessWidget {
               DevToolsTooltip(
                 message: 'Device Info',
                 child: InkWell(
-                  onTap: () async {
+                  onTap: () {
                     unawaited(
```
kenzieschmoll, line 163: if the callback is no longer async, do we still need to mark this unawaited?

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/sampler.dart
kenzieschmoll, line 64: remove trailing comma since this will fit on one line

packages/devtools_app/lib/src/shared/console/console_service.dart
```diff
@@ -163,7 +163,7 @@ class ConsoleService extends Disposer {
     return item.variable;
   }
 
-  void appendMessage(String text) {}
+  void appendMessage() {}
```
kenzieschmoll, line 166: this appears to be unused. Can we remove?

bkonyi (approved):
Did a fairly high-level lookover of the changes, but LGTM overall.

packages/.vscode/settings.json
polina-c: Removed.

packages/devtools_app/lib/src/framework/status_line.dart
polina-c: yes, otherwise lint is complaining that we drop the future

packages/devtools_app/lib/src/shared/primitives/simple_items.dart
```diff
@@ -50,3 +50,26 @@ const String traceEventsFieldName = 'traceEvents';
 const closureName = '<closure>';
 
 const anonymousClosureName = '<anonymous closure>';
+
+const _memoryDocUrl =
+    'https://docs.flutter.dev/development/tools/devtools/memory';
+const _consoleDocUrl =
+    'https://docs.flutter.dev/development/tools/devtools/console';
+
+enum DocLinks {
```
kenzieschmoll, line 59: nit: rename to MemoryDocsLinks since these are all for the memory screen

packages/devtools_app/lib/src/shared/primitives/simple_items.dart
```diff
@@ -50,3 +50,26 @@ const String traceEventsFieldName = 'traceEvents';
...
+  chart(_memoryDocUrl, 'expandable-chart'),
+  profile(_memoryDocUrl, 'profile-memory-tab'),
+  diff(_memoryDocUrl, 'diff-snapshots-tab'),
+  trace(_memoryDocUrl, 'trace-instances-tab'),
+  console(_consoleDocUrl, null),
```
kenzieschmoll, line 64: remove the last comma and replace with semicolon so we don't have a single line with a semicolon below

kenzieschmoll (approved):
a couple more comments but lgtm

packages/devtools_app/lib/src/shared/primitives/simple_items.dart
polina-c: It is easier to reorganize items, when they all have the same sign at the end. 
We have single line with semicolon in other places of the app (search by `  ;`). It looks as good separator between items and fields/methods.

polina-c: I am open to fix all single semicolons in next PR if you still think it is right thing to do.

kenzieschmoll: In general having a comma and a semi-colon seems messy. I'm in favor of fixing all uses of this in a follow-up PR.  Example from flutter: https://dart.dev/language/enum#declaring-enhanced-enums.

packages/devtools_app/lib/src/shared/primitives/simple_items.dart
kenzieschmoll: then we should rename each enum value to start with 'memory' so that these are more accurately named.

PR #5665, opened 2023-04-17
Move flutter-prep job to reusable workflow file

elliette:
- Work towards https://github.com/flutter/devtools/issues/5479
- Preparatory work for https://github.com/flutter/devtools/pull/5650

We will need to re-use the `flutter-prep` job in the DCM workflow file. This PR moves it into a reusable workflow file so that it can be called outside of `build.yaml`. 

See https://docs.github.com/en/actions/using-workflows/reusing-workflows for Github documentation on reusing workflows. 

`RELEASE_NOTE_EXCEPTION="No user facing changes"`


kenzieschmoll (approved):
refactor lgtm. But generally, why do we need to call the flutter-prep job more than once? It should only need to be done once per PR commit and every job can use the cached flutter sdk.

kenzieschmoll:
Ah I see. Okay what you have sgtm them. Only question is do we need to do the flutter prep job for both mac and linux for DCM or can we limit that to only one of the platforms?

elliette:
> Ah I see. Okay what you have sgtm them. Only question is do we need to do the flutter prep job for both mac and linux for DCM or can we limit that to only one of the platforms?

Ah good point, we only need to run on one OS. Let me try to update this to have a list of OS provided as a param

elliette:
`flutter-prep` now takes an input of either `macos` or `ubuntu`. If neither is provided, it will run against both.

PR #5671, opened 2023-04-18
Use switch expressions and pattern syntax in a few places

kenzieschmoll:
Just trying out some of the new Dart 3 features. We could do more cleanups like this in the future and should try to start using this syntax for new code where it is helpful.

CC @pq @bwilkerson

### DIFF
--- a/packages/devtools_app/lib/src/shared/primitives/utils.dart
+++ b/packages/devtools_app/lib/src/shared/primitives/utils.dart
@@ -1094,6 +1094,8 @@ class DevToolsFile<T> {
 
 final _lowercaseLookup = <String, String>{};
 
+bool nullOrEmpty(String? str) => str == null || str.isEmpty;
+
 // TODO(kenz): consider moving other String helpers into this extension.
 // TODO(kenz): replace other uses of toLowerCase() for string matching with
 // this extension method.
--- a/packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/program_explorer_controller.dart
@@ -343,19 +343,16 @@ class ProgramExplorerController extends DisposableController
 
     // Otherwise, we need to find the target script to determine the library
     // the target node is listed under.
-    ScriptRef? targetScript;
-    if (object is ClassRef) {
-      targetScript = object.location?.script;
-    } else if (object is FieldRef) {
-      targetScript = object.location?.script;
-    } else if (object is FuncRef) {
-      targetScript = object.location?.script;
-    } else if (object is Code) {
-      final ownerFunction = object.function;
-      targetScript = ownerFunction?.location?.script;
-    } else if (object is ScriptRef) {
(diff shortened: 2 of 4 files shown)


kenzieschmoll:
Even though the analyzer does not report any errors, this code does not build:
```dart
    final SourceLocation? sourceLocation = switch (object) {
      FieldRef(:final location) ||
      FuncRef(:final location) ||
      ClassRef(:final location) =>
        location,
      _ => null,
    };
```
`FieldRef.location`, `FuncRef.location`, and `ClassRef.location` all have a type of `SourceLocation?`, but the following build error occurs:
```dart
lib/src/screens/debugger/program_explorer_model.dart:349:23: Error: Variable pattern 'location' doesn't have the same type or finality in all cases.
      FieldRef(:final location) ||
                      ^^^^^^^^
lib/src/screens/debugger/program_explorer_controller.dart:347:23: Error: Variable pattern 'location' doesn't have the same type or finality in all cases.
      ClassRef(:final location?) ||
                      ^^^^^^^^
```
@bwilkerson

scheglov:
This might be a bug in CFE.
As we can see, you specified `final` in all branches.
And as I checked, the type of the property is `SourceLocation?`.
So, both the type and the finality are consistent.

bkonyi (approved):
That looks much better! 🎉🎉🎉

kenzieschmoll:
> This might be a bug in CFE.
> As we can see, you specified final in all branches.
> And as I checked, the type of the property is SourceLocation?.
> So, both the type and the finality are consistent.

Why is `final SouceLocation?` invalid? I get the same error when changing `final` to `SourceLocation?`:
```dart
    final SourceLocation? sourceLocation = switch (object) {
      FieldRef(:SourceLocation? location) ||
      FuncRef(:SourceLocation? location) ||
      ClassRef(:SourceLocation? location) =>
        location,
      _ => null,
    };
```

```
lib/src/screens/debugger/program_explorer_model.dart:349:33: Error: Variable pattern 'location' doesn't have the same type or finality in all cases.
      FieldRef(:SourceLocation? location) ||
```

scheglov:
I don't see any error in the analyzer, with one or another variant.
As I said, this might be a bug in CFE.

kenzieschmoll:
Filed https://github.com/dart-lang/sdk/issues/52099 to track.

packages/devtools_app/lib/src/shared/primitives/utils.dart
```diff
@@ -1094,6 +1094,8 @@ class DevToolsFile<T> {
 
 final _lowercaseLookup = <String, String>{};
 
+bool nullOrEmpty(String? str) => str == null || str.isEmpty;
```
jacob314, line 1097: nit: maybe make this an extension method on `String?`

kenzieschmoll: done. Addressed in https://github.com/flutter/devtools/pull/5677

PR #5677, opened 2023-04-19
Create NullableStringExtension and isNullOrEmpty helper

kenzieschmoll:

### DIFF
--- a/packages/devtools_app/lib/src/shared/primitives/utils.dart
+++ b/packages/devtools_app/lib/src/shared/primitives/utils.dart
@@ -1095,7 +1095,10 @@ class DevToolsFile<T> {
 final _lowercaseLookup = <String, String>{};
 
 extension NullableStringExtension on String? {
-  bool get isNullOrEmpty => this == null || this!.isEmpty;
+  bool get isNullOrEmpty {
+    final self = this;
+    return self == null || self.isEmpty;
+  }
 }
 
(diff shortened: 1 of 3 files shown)


bkonyi (approved):

packages/devtools_app/lib/src/shared/primitives/utils.dart
```diff
@@ -1094,7 +1094,9 @@ class DevToolsFile<T> {
 
 final _lowercaseLookup = <String, String>{};
 
-bool nullOrEmpty(String? str) => str == null || str.isEmpty;
+extension NullableStringExtension on String? {
+  bool get isNullOrEmpty => this == null || this!.isEmpty;
```
scheglov, line 1098: In the analyzer we use `self` to allow type promotions.
In general using null asserts `!` is bad.
```dart
bool get isNullOrEmpty {
    final self = this;
    return self == null || self.isEmpty;
}
```

Or, if you really like patterns
```dart
  bool get isNullOrEmpty {
    return switch (this) {
      null || String(isEmpty: true) => true,
      _ => false,
    };
  }
```
Unfortunately, we [have](https://github.com/dart-lang/language/issues/3008) to write `String` type in the object pattern.

kenzieschmoll: Done.

PR #5683, opened 2023-04-19
Fix all remaining DCM warnings in `devtools_shared`

elliette:
Work towards https://github.com/flutter/devtools/issues/5479

### DIFF
--- a/packages/devtools_shared/pubspec.yaml
+++ b/packages/devtools_shared/pubspec.yaml
@@ -16,4 +16,5 @@ dependencies:
   webkit_inspection_protocol: '>=0.5.0 <2.0.0'
 
 dev_dependencies:
+  dart_code_metrics: ^5.7.2
   test: ^1.21.2
--- a/packages/devtools_shared/lib/src/memory/memory_json.dart
+++ b/packages/devtools_shared/lib/src/memory/memory_json.dart
@@ -308,7 +308,7 @@ class AllocationMemoryJson extends MemoryJson<ClassHeapStats> {
           ],
           // We'll just fudge the old space numbers.
           '_old': const <int>[0, 0, 0],
-        }
+        },
     ];
     return updatedPayload;
   }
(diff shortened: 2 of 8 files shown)


packages/devtools_shared/pubspec.yaml
```diff
@@ -16,4 +16,5 @@ dependencies:
   webkit_inspection_protocol: '>=0.5.0 <2.0.0'
 
 dev_dependencies:
+  dart_code_metrics: ^5.7.2
   test: ^1.21.2
```
polina-c, line 20: Can you comment why the dependency is needed?

elliette: Hmm having it as a dev dependency is part of the installation instructions, but I don't why it's necessary https://dartcodemetrics.dev/docs/getting-started/installation

I  believe @kenzieschmoll was running into an issue due to not having this as a dev dependency. 

I already added it as a dev dependency to devtools_app and devtools_test

kenzieschmoll: From the [instructions](https://dartcodemetrics.dev/docs/cli#available-commands), it was necessary to run commands like: `dart run dart_code_metrics:metrics check-unused-code lib`, `check-unused-files`, etc.

kenzieschmoll: but maybe the same can be accomplished by using the `dcm` executable directly.. so maybe it isn't needed after all.

kenzieschmoll: @incendial this is something that may need docs updates. That you need _either_ the dev dependency _or_ the dcm executable, but not necessarily both.

kenzieschmoll (approved):

packages/devtools_shared/pubspec.yaml
incendial: You're referring to the old website, please use the new one instead (https://dcm.dev/docs/teams/getting-started/ and https://dcm.dev/docs/teams/cli/).

There are two versions available: one is a dart package and requires the dependency to be listed and another one is a standalone executable (former is the "Individuals" version and the latter is "Teams). You have a Teams version, so no dependency is needed.

In our docs they are also separated:
Screenshot 2023-04-20 at 11 08 13
Screenshot 2023-04-20 at 11 08 39

And the url is different.

Hope this helps resolve the misunderstanding.

PR #5690, opened 2023-04-20
Clean up some inconsistent styling on the app size screen

kenzieschmoll:
Use helper widget `_AppSizeView` to maintain consistent styling for overall framework of page.

### DIFF
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_screen.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_screen.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 
 import 'package:devtools_shared/devtools_shared.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/screens/app_size/app_size_screen.dart
```diff
@@ -619,18 +531,90 @@ class _DiffViewState extends State<DiffView>
...
+                ],
+              ),
+            ),
+          ],
+        ),
+      ),
+    );
+  }
 
   Widget _buildTreemap() {
```
polina-c, line 607: Should it be widget?

polina-c (approved):
lgtm with one comment

packages/devtools_app/lib/src/screens/app_size/app_size_screen.dart
kenzieschmoll: declared the treemap directly in the build method instead to remove this helper

PR #5695, opened 2023-04-21
Remove unused dependencies

kenzieschmoll:
@incendial it could be interesting to create a lint that flags unused dependencies.

### DIFF
--- a/packages/devtools_app/pubspec.yaml
+++ b/packages/devtools_app/pubspec.yaml
@@ -23,7 +23,6 @@ dependencies:
   async: ^2.0.0
   codicon: ^1.0.0
   collection: ^1.15.0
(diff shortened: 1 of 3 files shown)


incendial:
Unfortunately, it requires a full project analysis, so a a lint it can be too expensive. But how about a cli command (check)?

bkonyi (approved):

kenzieschmoll:
> But how about a cli command (check)?

A cli command sounds great.

incendial:
@kenzieschmoll the command is now available in dcm 1.4.0 https://dcm.dev/docs/teams/cli/check-dependencies/. If you want, I can create an a separate issue for its integration.

Issue #3799, opened 2022-03-08
CPU profiler is slow, freezes, dies


scheglov:
The attached JPG is actually BZIP2 file with `Export` result.

Expand all, click on the scroll thumb and try to scroll.
It freezes, does not actually scroll, then Chrome complains that the tab is not responding, try to scroll a few more times, the tab just dies.

scheglov:
[dart_devtools_2022_3_7-1646698474397000.json.gz](https://github.com/flutter/devtools/files/8201671/dart_devtools_2022_3_7-1646698474397000.json.gz)

scheglov:
I did not actually tried to open this file, I used whatever was initially produced, but this is the export.

kenzieschmoll:
Even when replacing each table row with a very simple colored container, the janky scrolling continues to occur. You can see in the widget rebuild stats chart in intelliJ that when we scroll very quickly, it seems that we are still building every item from the previous scroll position to the new one. Sometimes 1000s of widgets. When we have the more complicated table row for a cpu stack frame (with lots of text and listeners), this doesn't just jank, but completely freezes the list view.
![janky-scrolling](https://user-images.githubusercontent.com/43759233/233726993-bc9985ea-0cee-4930-b57c-1ad694bfebca.gif)

Sounds like this is an instance of this issue: https://github.com/flutter/flutter/issues/75399

kenzieschmoll:
Edit: we were setting `itemExtent: widget.rowItemExtent`, and after digging a bit deeper, it turns out `widget.rowItemExtent` was in fact null 😅

PR #5696, opened 2023-04-21
Fix slow scrolling performance in tree tables

kenzieschmoll:
It looked like `ListView.itemExtent` was being set on the table's vertical list widget, but we were actually setting the `ListView.itemExtent` to a null value. This PR fixes that. Behold, snappy scrolling in the CPU profiler 🎉 

performant-scrolling

Fixes https://github.com/flutter/devtools/issues/3799, part of https://github.com/flutter/devtools/issues/5091


polina-c (approved):

jacob314 (commented):
lgtm

Issue #5710, opened 2023-04-24
Expression evaluation autocomplete replaces entire input


elliette:
Broken out from https://github.com/flutter/devtools/issues/5703, find repro steps there

 Typing tab or enter during expression evaluation replaces replaces everything in input, instead of everything up to the last dot (eg. `myClass.someMetho` -> enter to accept `someMethod` replaces all of `myClass.someMetho` with `someMethod`)

Issue #5703, opened 2023-04-24
DevTools slow and unhelpful


jensjoha:
So I tried out DevTools.

Acknowledgement up front: I realize that what I try to navigate in below are big files and stuff might be more fun with smaller files.

This is on top-of-tree on dart (92239e6b273e3a1223887a33cd55533839aeaf29).

This is my experience:

```
out/ReleaseX64/dart --enable-vm-service --pause_isolates_on_start pkg/front_end/tool/_fasta/compile.dart pkg/front_end/tool/_fasta/compile.dart
```

- Click the link to open devtools.
- Click "debugger" on the top tab.
- Now I want to add a breakpoint on `finishFunction` in `body_builder.dart`. I'd like to be able to just write that somewhere (I could in observatory) --- but I at least can't find that.
- So I'll navigate to it via ctrl+p (which, granted, is much nicer than if I wanted to navigate to stuff in observatory).
- It's slow to open the file though, and there's no indication that it's doing anything. It took more than 6.5 seconds for it to load.
- Now I want to go to the `finishFunction` method as I try to hit "ctrl+f" to search. Seemingly nothing happens, but then ~6 seconds later a search thing shows up. Clicking the "x" to close the search thing again takes another ~6 seconds. Bringing it back again takes yet another ~6 seconds. (For the fun of it I recorded in chromes devtools, and indeed there's a call taking 6.65 seconds, going down a long chain of stuff seemingly ending in a function `b2A` taking 6.47 seconds before breaking up into smaller calls).
- Anyways, I paste `finishFunction` into the search box and it quickly goes to the first instance. It's a comment and not what I'm looking for so I hit <enter> thinking it would take me to the next instance. It does not. I don't know if there's any shortcut I can press, but I haven't found it --- I have to use my mouse to hit the (small) "down" arrow thing. I find the function and set a breakpoint.
- I click the "resume" button.
- Once it gets there I'd like to start stepping. I have no indication of keyboard shortcuts for this so I use my mouse.
- Stepping in the same file seems to be somewhat responsive, but stepping into another file is slow. Again (although I didn't time it this time) ~6 seconds.
- Stepping out of the function (and back into `body_builder.dart`) first took maybe a second, but it was at the top of the file, then after 6+ seconds it jumped down to where it actually was supposed to be, then spend some more seconds where scrolling with the mouse did nothing. At this point some of the line it's in also looks weird: `final SourceFunctionBuilder builder` looks fuzzy --- see screenshot below.
[...]

PR #5717, opened 2023-04-25
Fix expression evaluation autocomplete bug 

elliette:
Fixes https://github.com/flutter/devtools/issues/5710

Fixes the bug where selecting a field during expression evaluation would replace everything in the input, instead of everything up to the last dot (eg. `myClass.someMetho` -> enter to accept `someMethod` would replace all of `myClass.someMetho` with just `someMethod`).

### DIFF
--- a/packages/devtools_app/lib/src/shared/console/widgets/evaluate.dart
+++ b/packages/devtools_app/lib/src/shared/console/widgets/evaluate.dart
@@ -76,7 +76,7 @@ class ExpressionEvalFieldState extends State<ExpressionEvalField>
     });
     addAutoDisposeListener(
       _autoCompleteController.selectTheSearchNotifier,
-      _handleSearch,
+      _handleSearchTermSelected,
     );
     addAutoDisposeListener(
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/shared/console/widgets/evaluate.dart
```diff
@@ -134,16 +134,17 @@ class ExpressionEvalFieldState extends State<ExpressionEvalField>
...
+  }
+
+  Future<void> _handleSearch() async {
     final searchingValue = _autoCompleteController.search;
 
     _autoCompleteController.clearCurrentSuggestion();
 
     if (searchingValue.isNotEmpty) {
-      if (_autoCompleteController.selectTheSearch) {
-        _autoCompleteController.resetSearch();
```
elliette, line 144: Resetting the search term here was the crux of the bug. This would clear whatever the user had already inputted.

kenzieschmoll (approved):

PR #5721, opened 2023-04-26
Use records for AnalyticsTabbedView

kenzieschmoll:
Use the new records language feature to pair a tab and its respective tab view. Benefits:

1. Merges conditional checks that should be the same within the tabs list and tabViews list, preventing tab controller errors where the tabs and tabViews list lengths get out of sync.

Before:
```dart
final tabs = [
  if (foo) myFooTab,
  if (bar) myBarTab,
  if (baz) myBazTab,
];
final tabViews = [
  if (foo) myFooTabView,
  if (bar) myBarTabView,
  if (baz) myBazTabView,
];
```
After: 
```dart
final tabs = [
  if (foo) (tab: myFooTab, tabView: myFooTabView),
  if (bar) (tab: myBarTab, tabView: myBarTabView),
  if (baz) (tab: myBazTab, tabView: myBazTabView),
];
```

2. Eliminates record-like dart classes we were using to group tabs and tab views.

### DIFF
--- a/packages/devtools_app/lib/src/screens/inspector/inspector_screen_details_tab.dart
+++ b/packages/devtools_app/lib/src/screens/inspector/inspector_screen_details_tab.dart
@@ -29,20 +29,21 @@ class InspectorDetails extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final tabs = [
-      _buildTab(tabName: 'Layout Explorer'),
-      _buildTab(
-        tabName: 'Widget Details Tree',
-        trailing: InspectorExpandCollapseButtons(controller: controller),
+      (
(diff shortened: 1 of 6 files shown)


kenzieschmoll:
A concern was brought up about code size with replacing a class with a record in some cases. To verify this change does not have a code size impact, I performed the following comparison.

Comparing the size of the build output with and without this change:
- with this change: **56.8 MB**
- master: **56.8**

Compared by running the following and then looking at the size of the `devtools_app/build/web` folder
```
flutter clean
rm -rf build/web
flutter pub get
flutter build web \
  --web-renderer canvaskit \
  --pwa-strategy=offline-first \
  --release \
  --no-tree-shake-icons
```

Issue #5708, opened 2023-04-24
Highlighted line is blurry when paused at breakpoint


elliette:
Broken out from https://github.com/flutter/devtools/issues/5703, find repro steps there

When paused at a breakpoint, the line that is highlighted is blurry (looks like we might be rendering the text on top of itself)

PR #5732, opened 2023-04-27
Fix blurriness on focused line in code view

elliette:
Fixes https://github.com/flutter/devtools/issues/5708

### Before:

Screenshot 2023-04-27 at 2 34 16 PM

### After:

Screenshot 2023-04-27 at 2 34 29 PM

### DIFF
--- a/packages/devtools_app/lib/src/screens/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/codeview.dart
@@ -1207,7 +1207,7 @@ class _LineItemState extends State<LineItem>
               // to allow us to render this as a proper overlay as similar
               // functionality exists to render the selection handles properly.
               Opacity(
-                opacity: 0.5,
+                opacity: 0.0,
                 child: RichText(
                   text: truncateTextSpan(widget.lineContents, column - 1),
                 ),
--- a/packages/devtools_app/lib/src/screens/debugger/codeview.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/codeview.dart
@@ -1184,11 +1184,13 @@ class _LineItemState extends State<LineItem>
     final column = widget.pausedFrame?.column;
     if (column != null) {
       final breakpointColor = theme.colorScheme.breakpointColor;
-
+      final widthToCurrentColumn = calculateTextSpanWidth(
+        truncateTextSpan(widget.lineContents, column - 1),
+      );
       // The following constants are tweaked for using the
       // 'Icons.label_important' icon.
       const colIconSize = 13.0;
-      const colLeftOffset = -3.0;
+      final colLeftOffset = -3.0 + widthToCurrentColumn;
       const colBottomOffset = 13.0;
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -1207,7 +1207,7 @@ class _LineItemState extends State<LineItem>
               // to allow us to render this as a proper overlay as similar
               // functionality exists to render the selection handles properly.
               Opacity(
```
kenzieschmoll, line 1209: what is the point of using a completely transparent opacity widget now?

kenzieschmoll: is the comment above this widget still relevant?

elliette: Hmm the comment (and this code) is 2 years old: https://github.com/flutter/devtools/pull/2665. I can investigate the feasibility if you think that's better!

I think I accidentally made it semi-transparent when debugging https://github.com/flutter/devtools/pull/4651, and forgot to switch it back

kenzieschmoll:
If everything is WAI with the Opacity widget removed, then maybe that is stale code that we don't need?

kenzieschmoll:
Oh wait a minute, with opacity 0 the text doesn't show up at all. I was getting this backwards for some reason thinking that we were laying a transparent widget on top of the text instead of hiding the text itself. The comment makes sense now. LGTM, then. Whether we can remove the workaround in the comment or not, doing so may be out of scope.

kenzieschmoll:
unless you've already applied a fix and verified things are WAI with your last commit :)

elliette:
> Oh wait a minute, with opacity 0 the text doesn't show up at all. I was getting this backwards for some reason thinking that we were laying a transparent widget on top of the text instead of hiding the text itself. The comment makes sense now. LGTM, then. Whether we can remove the workaround in the comment or not, doing so may be out of scope.

Just removed it by using our `TextSpan` measurement utility. Let me know what you think! Happy to change it back as well

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -1184,11 +1184,13 @@ class _LineItemState extends State<LineItem>
...
       final breakpointColor = theme.colorScheme.breakpointColor;
-
+      final widthToCurrentColumn = calculateTextSpanWidth(
+        truncateTextSpan(widget.lineContents, column - 1),
+      );
       // The following constants are tweaked for using the
       // 'Icons.label_important' icon.
       const colIconSize = 13.0;
-      const colLeftOffset = -3.0;
+      final colLeftOffset = -3.0 + widthToCurrentColumn;
```
kenzieschmoll, line 1193: nit: change to `widthToCurrentColumn - 3.0` (if the 3 is still needed here? a comment could help)

kenzieschmoll (approved):

PR #5747, opened 2023-05-03
Add experimental VS Code extension prototype

DanTup:
@jacob314 @kenzieschmoll this needs more work before it's worth properly reviewing or merging, but having a PR might make it easier to talk about(/comment on) the code.

## Projects

- VS Code extension (`vs_code_extension` folder)
- Flutter app for UI (`flutter_ui` folder)
- Dart package to allow interacting with the Dart extension (`dart_code_api`)
- NPM/TypeScript package for the VS Code extension parts that might be reused across multiple extension projects (`dart_code_api_npm`)

## General Design

The Dart extension exports an API (currently [on a branch](https://github.com/Dart-Code/Dart-Code/tree/flutter-tooling-extension-sample)) that can be used by other extensions, which includes things like:

- events for debug sessions starting/ending and VM Service becoming available
- APIs to send requests to the LSP server

Any VS Code extension can consume these APIs by writing a pure TypeScript extension (and optionally using the type definitions in `dart_code_api_npm`), but the Dart (`dart_code_api`) package also simplifies using `postMessage` to send messages between an embedded webview and a wrapper over that API (essentially making the same API available via TypeScript to extension code, or Dart to embedded Flutter apps).

Additionally the Dart part of the API exposes some additional helpers for calling the VS Code APIs (for example `executeCommand` is implemented) which don't need to go via the Dart extension. This is make it easier to do some basic VS Code interaction from the Flutter UI without needing to write TypeScript.

## Open Questions

- Should the iframe/communication code in `dart_code_api_npm` just be inlined into the extension as a starting point (to allow extensions to customise it if needed?)
- Should the TypeScript type definitions in `dart_code_api_npm` live closed to Dart-Code (since they are a description of the API it exports and should be kept in-sync with no breaking changes)
- how useful is having (approximately) the same API available to both TypeScript and Dart parts of the code? (originally the TS was just proxying directly over to Dart-Code, but that made it impossible to use the APIs from extension TypeScript code without building weird JSON objects)

## Other TODOs

- code-gen to avoid having to manually keep a bunch of things in-sync (the TS type definitions, the TS proxy code, the Dart API classes)
- improving LSP typing (eg. publishing the LSP types from analysis_server and being able to reuse them here)
- better handling of iframe not always being visible (eg. if a debug session starts before you show the side bar, it misses the event and has no way to know about the session)
- running a non-dev version (this currently relies on `flutter run`, but shipping would require building a release app and hosting it with a web server inside the extension)
- could we make debugging (and hot-reload etc.) work? currently we can't run the Dart debug extension nor launch Chrome here
- lots of TODOs in the code
- hover example has hard-coded paths


kenzieschmoll:
One clarifying question: is the `dart_code_api` a web only package, or will there be an implementation that allows use from a non-web platform?

jacob314:
It would be nice if you could iterate on a VSCode sidebar using Flutter Desktop so that you could use hot reload. The "sidebar" wouldn't need to actually render in VSCode but would need to be able to use the same VSCode APIs. I expect that would significantly improve productivity iterating on something like a property editor.

DanTup:
Yeah, that sounds really appealing (as does being able to use the debugger!).. This would require some other communication channel than `postMessage` though. I'll think about this a little more 🙂

DanTup:
I've been a bit side-tracked and not looked back at this lately, but I had started trying to make a list of APIs that I think could be useful to expose to a plugin (whether it's hosted inside VS Code or DevTools). This could be something like JSONRPC over `postMessage` (either to VS Code, or DevTools), but with a Dart class to wrap it (for ex. to present events as a `Stream<>` and make requests a `Future`-returning method).

For easier debugging, we could swap out `postMessage` with a `WebSocket` or similar, allowing "VS Code embedded" plugins to run outside of VS Code (for debugging and hot-reload), but hopefully without any change in functionality/behaviour besides the not being rendered inside the sidebar.

My initial thoughts on an API (not all necessarily supported immediately, but as an idea of how sections of functionality could be optional depending on where the plugin is hosted):

LSP:
  - Is there an LSP server available?
  - What are the Client + Server capabilities of the LSP server?
  - Send this LSP request / get response
    - Is this a whitelist? Allowing requests that modify state like
      textDocument/didOpen could be problematic
  - Subscribe to LSP events?

Debug Sessions:
  - What debug sessions are available
    - This may always be 0/1 in DevTools, but could be >1 in VS Code
  - Subscribe to debug session events
    - Start/Stop/VM Service available

DAP:
  - Is a debug adapter available (for a given debug session)?
  - What are the Client + Server capabilities of the DAP server?
  - Send this DAP request / get response
  - Subscribe to DAP requests?

VS Code:
  - Is VS Code available?
  - Execute this VS Code command/args
  - Get some basic workspace info
    - Do we have a Dart/Fluter SDK?
    - Are there any Dart/Flutter projects open?
    - (other things that might support a basic sidebar for helping discoverability)
  - Subscribe to some editor events
    - Changes to document contents and/or location of cursor in document (eg. for property editor)  

DevTools:
  - Is DevTools available?
  - ?

State?:
  - Should there be some way for simple state storage (eg. if a user toggles some setting)?
    - Unsure if things like localStorage are available inside VS Code
  - Is state storage available?
  - Store this value
  - Get this value

DanTup:
See https://github.com/flutter/devtools/pull/5921 for a sort-of updated version of this. It's currently not focused on external VS Code extensions, but could be expanded to support that (using the same `DartApi` class) and I think is a lot cleaner than this one (which has a very custom protocol and some dodgy hacks to proxy the messages). It also supports connecting a WebSocket so you can run outside of VS Code (for full Flutter debugging) while still having it connected to the VS Code instance and using the API.

DanTup:
Closing in favour of #6104.

PR #5756, opened 2023-05-05
Update DCM workflow

elliette:
- Requires label `run-dcm-workflow` to run
- Only runs on PRs targeting the master branch

### DIFF
--- a/.github/workflows/dcm-label.yaml
+++ b/.github/workflows/dcm-label.yaml
@@ -1,0 +1,20 @@
+# A CI workflow to remind the contributor to add the run-dcm-workflow label.
+
+name: DCM Label Reminder
+
+on:
+  pull_request:
+    types: [synchronize, opened, reopened, labeled, unlabeled]
+    paths:
+      - '**/*.dart'
+
+jobs:
+  changelog-reminder:
+    if: ${{ !contains(github.event.*.labels.*.name, 'run-dcm-workflow') }}
+    name: Prevent submission
+    runs-on: ubuntu-latest
+    steps:
+      - name: Label reminder
+        run: |
+          echo "Please add the 'run-dcm-workflow' label to run DCM checks."
+          exit 1
--- a/.github/workflows/dcm-label.yaml
+++ b/.github/workflows/dcm-label.yaml
@@ -16,5 +16,5 @@ jobs:
     steps:
       - name: Label reminder
         run: |
-          echo "Please add the 'run-dcm-workflow' label to run DCM checks."
+          echo "::error Please add the 'run-dcm-workflow' label to run DCM checks."
           exit 1
--- a/.github/workflows/dcm-label.yaml
+++ b/.github/workflows/dcm-label.yaml
@@ -1,4 +1,4 @@
-# A CI workflow to remind the contributor to add the run-dcm-workflow label.
+# A CI workflow to remind the contributor to add the 'run-dcm-workflow' label.
 
 name: DCM Label Reminder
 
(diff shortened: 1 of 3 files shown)


.github/workflows/dcm.yaml
```diff
@@ -15,15 +15,19 @@ name: Dart Code Metrics
...
 jobs:
   flutter-prep:
     uses: ./.github/workflows/flutter-prep.yaml
     with:
       os-name: ubuntu
+      requires-label: 'run-dcm-workflow'
       needs-checkout-merge: true
 
   dcm:
+    if: ${{ contains(github.event.*.labels.*.name, 'run-dcm-workflow') }}
```
CoderDake, line 30: Is there a reason why we want this on all types of events instead of just for pull_requests?

elliette: Good point, switched to just pull_requests!

.github/workflows/flutter-prep.yaml
```diff
@@ -13,13 +13,18 @@ on:
       needs-checkout-merge:
         description: 'Whether the PR should be merged during the checkout step. Necessary for pull_request_target workflows.'
         type: boolean
+      requires-label:
+        description: 'Specifies the label required for the workflow to run. If not provided, the workflow will always run.'
```
kenzieschmoll, line 17: should this say "If not provided, the workflow will not run." ?

CoderDake (approved):

.github/workflows/dcm.yaml
```diff
@@ -15,15 +15,20 @@ name: Dart Code Metrics
...
+    types: [synchronize, opened, reopened, labeled, unlabeled]
+    branches:
+      - master
 
 jobs:
   flutter-prep:
     uses: ./.github/workflows/flutter-prep.yaml
     with:
       os-name: ubuntu
+      requires-label: 'run-dcm-workflow'
```
CoderDake, line 27: Should this be on the dcm job?

CoderDake (changes requested):

.github/workflows/flutter-prep.yaml
elliette: Nope, it will always run unless a label is required to run.

.github/workflows/dcm.yaml
```diff
@@ -15,15 +15,20 @@ name: Dart Code Metrics
 
 on:
```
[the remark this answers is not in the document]
elliette: Done! Added it to push on master

.github/workflows/dcm-label.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+jobs:
+  changelog-reminder:
+    if: ${{ !contains(github.event.*.labels.*.name, 'run-dcm-workflow') }}
+    name: Prevent submission
+    runs-on: ubuntu-latest
+    steps:
+      - name: Label reminder
+        run: |
+          echo "Please add the 'run-dcm-workflow' label to run DCM checks."
```
CoderDake, line 19: You can have this pop up as an error in the UI using ::error
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#using-workflow-commands-to-access-toolkit-functions

.github/workflows/dcm.yaml
```diff
@@ -14,16 +14,24 @@
 name: Dart Code Metrics
 
 on:
-  pull_request_target
+  push:
+    branches:
+      - master
+  pull_request_target:
+    types: [synchronize, opened, reopened, labeled, unlabeled]
```
CoderDake, line 21: Could you leave a comment about why these types are important, or perhaps about why others aren't desired?

CoderDake (approved):

elliette:
Merging because the same test failures are seen on master.

PR #5783, opened 2023-05-12
Make device info text flexible so it doesn't overflow

CoderDake:
New flutter candidate seems to have broken these tests. So I've made the text on the device info dialog wrap to avoid overflow.

 
Screenshot 2023-05-12 at 11 54 57 AM

### DIFF
--- a/packages/devtools_app/lib/src/shared/device_dialog.dart
+++ b/packages/devtools_app/lib/src/shared/device_dialog.dart
@@ -50,9 +50,11 @@ class DeviceDialog extends StatelessWidget {
               child: Row(
                 children: [
                   Text('${entry.title}: ', style: boldText),
-                  SelectableText(
-                    entry.description,
-                    style: theme.subtleTextStyle,
+                  Flexible(
+                    child: SelectableText(
+                      entry.description,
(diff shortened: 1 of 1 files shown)


bkonyi (approved):

kenzieschmoll:
We shouldn't need the text to break to two lines. It doesn't overflow locally, only when running `flutter test`. 

This test passes with `flutter run test/shared/device_dialog_test.dart` but fails with `flutter test test/shared/device_dialog_test.dart` with a RenderFlex overflow error. 

This started happening when our flutter version bumped to
```
Flutter 3.11.0-4.0.pre • channel unknown • unknown source
Framework • revision 65dfb555c0 (4 days ago) • 2023-05-08 09:51:28 +0200
Engine • revision 23f730efbf
Tools • Dart 3.1.0 (build 3.1.0-83.0.dev) • DevTools 2.23.1
```

I confirmed the test passes normally on flutter master. Can we instead, temporarily, use the flutter master channel for our testing?

kenzieschmoll:
Closing this as this will be resolved by switching to the master flutter branch

PR #5809, opened 2023-05-15
Fix lint warnings on master

kenzieschmoll:


packages/devtools_app/test/test_infra/fixtures/riverpod_app/pubspec.yaml
```diff
@@ -1,10 +1,11 @@
 name: riverpod_app
-description: A new Flutter project.
+description: A test app that uses package:riverpod.
 publish_to: "none" # Remove this line if you wish to publish to pub.dev
 version: 1.0.0+1
 
 environment:
-  sdk: '>=2.17.0 <4.0.0'
+  sdk: '>=3.0.0-417.0.dev <4.0.0'
```
bkonyi, line 7: Why not just `>=3.0.0` here and above?

bkonyi (approved):

packages/devtools_app/test/test_infra/fixtures/riverpod_app/pubspec.yaml
kenzieschmoll: I was copying what we use in devtools_app/pubspec.yaml. We can probably upgrade everything to >=3.0.0 now.

Issue #5709, opened 2023-04-24
Out-of-sync file names and file contents


elliette:
Broken out from https://github.com/flutter/devtools/issues/5703, find repro steps there

File name and file contents are not updated at the same time. This is especially noticeable when it takes a long time to load a file. If contents of a file fail to load, the two are out of sync. 

![switching files](https://user-images.githubusercontent.com/21270878/234134778-e28aed7e-8616-49c2-98e1-9347f8aa1afe.gif)

PR #5814, opened 2023-05-16
Add a default error notification

elliette:
Adds a default error notification that can be triggered with `NotificationService.pushError(errorMessage)`.

The notification will display the error, and include a button to report it to our issue tracker. The opened bug will be pre-filled with "Reporting error: {{error message}}" as the title. 
 
Screenshot 2023-05-16 at 9 48 14 AM

Work towards https://github.com/flutter/devtools/issues/5709 (if we are unable to parse a file, we would like to notify the user of what has happened).

### DIFF
--- a/packages/devtools_app/lib/src/shared/notifications.dart
+++ b/packages/devtools_app/lib/src/shared/notifications.dart
@@ -55,7 +55,9 @@ class NotificationService {
         unawaited(
           launchUrl(
             devToolsExtensionPoints
-                .issueTrackerLink(additionalInfo: errorMessage)
+                .issueTrackerLink(
+                  issueTitle: 'Reporting error: $errorMessage',
+                )
                 .url,
           ),
         );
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/shared/notifications.dart
```diff
@@ -39,6 +44,37 @@ class NotificationService {
...
+        );
+      },
+    );
+    return pushNotification(
+      NotificationMessage(
+        errorMessage,
+        isError: true,
+        actions: [reportErrorAction],
+        // Double the duration so that the user has time to report the error:
+        duration: NotificationMessage.defaultDuration * 2,
```
kenzieschmoll, line 72: likely out of scope for this PR (since I don't think we have this for regular notifications?) but we should probably give the users a way to dismiss notifications. Especially long lasting ones like this one.

kenzieschmoll (approved):

packages/devtools_app/lib/src/shared/notifications.dart
elliette: SG! Will follow up with a PR to make notifications dismissable

PR #5820, opened 2023-05-17
usePathUrlStrategy from flutter_web_plugins

mdebbar:
The `usePathUrlStrategy` function can be imported from` flutter_web_plugins/url_strategy.dart`. No need to do a conditional import either.


DanTup (approved):
Ah, nice :-)

kenzieschmoll (approved):

PR #5822, opened 2023-05-17
Notifications can be dismissed

elliette:
Follow up to https://github.com/flutter/devtools/pull/5814
Work towards https://github.com/flutter/devtools/issues/5703, https://github.com/flutter/devtools/issues/5709 (if we are unable to parse a file, we would like to notify the user of what has happened).

1. Allow notifications to be dismissed by the user by closing the "X" at the top right:

Screenshot 2023-05-18 at 4 11 42 PM

2. Fixes the positioning of short notifications:

After:
Screenshot 2023-05-18 at 4 22 16 PM

Before:
Screenshot 2023-05-18 at 4 04 23 PM

3. Makes the VM service connection failure notification an error notification.

### DIFF
--- a/packages/devtools_app/lib/src/framework/notifications_view.dart
+++ b/packages/devtools_app/lib/src/framework/notifications_view.dart
@@ -255,7 +255,7 @@ class _NotificationState extends State<_Notification>
                 crossAxisAlignment: CrossAxisAlignment.start,
                 children: [
                   if (widget.message.isDismissable)
-                    _DismissAction(widget: widget, context: context),
+                    _DismissAction(widget: widget),
                   _NotificationMessage(widget: widget, context: context),
                   const SizedBox(height: defaultSpacing),
                   _NotificationActions(widget: widget),
@@ -272,11 +272,9 @@ class _NotificationState extends State<_Notification>
 class _DismissAction extends StatelessWidget {
   const _DismissAction({
     required this.widget,
-    required this.context,
   });
 
   final _Notification widget;
-  final BuildContext context;
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/framework/notifications_view.dart
```diff
@@ -254,6 +254,8 @@ class _NotificationState extends State<_Notification>
               child: Column(
                 crossAxisAlignment: CrossAxisAlignment.start,
                 children: [
+                  if (widget.message.isDismissible)
+                    _DismissAction(widget: widget),
                   _NotificationMessage(widget: widget, context: context),
```
kenzieschmoll, line 259: can these two items be in a Row so that the text and the dismiss button are in alignment?

packages/devtools_app/lib/src/framework/notifications_view.dart
```diff
@@ -267,6 +269,29 @@ class _NotificationState extends State<_Notification>
   }
 }
 
+class _DismissAction extends StatelessWidget {
+  const _DismissAction({
+    required this.widget,
```
kenzieschmoll, line 274: instead of passing in a widget, can we just pass in an `onPressed` callback?

packages/devtools_app/lib/src/framework/notifications_view.dart
elliette: Done! In working on that, noticed some other weird alignment issues (see PR description for images)

packages/devtools_app/lib/src/framework/notifications_view.dart
elliette: Done!

kenzieschmoll (approved):

PR #5827, opened 2023-05-19
File names should only be updated if parsing the file succeeds

elliette:
Fixes https://github.com/flutter/devtools/issues/5709

Before we were wrapping our script parsing step in an `unawaited`, therefore we weren't waiting for the parsing to succeed to update the file name. 

This meant that:
- if a script took a long time to parse (ie it was a big script), we would immediately show the new name but still show the contents of the old script
- if the script parsing failed, we would still show the old file but with the new file's name

Now we wait for parsing to succeed to update the file name. If parsing fails, we show an error to the user.


packages/devtools_app/lib/src/screens/debugger/codeview_controller.dart
```diff
@@ -211,33 +215,45 @@ class CodeViewController extends DisposableController
...
     _currentScriptRef.value = null;
     parsedScript.value = null;
-    showScriptLocation(scriptLocation);
+    await showScriptLocation(scriptLocation);
   }
 
   /// Show the given script location (without updating the script navigation
-  /// history).
-  void _showScriptLocation(
+  /// history). Returns a boolean value representing success or failure.
```
kenzieschmoll, line 226: dart doc nit: put empty line after first sentence

packages/devtools_app/test/debugger/debugger_scripts_test.dart
```diff
@@ -44,14 +44,16 @@ void main() {
   ) async {
     await tester.pumpWidget(
       wrapWithControllers(
-        const DebuggerScreenBody(),
+        const NotificationsView(
```
kenzieschmoll, line 47: wrapWithControllers already includes a NotificationsView in the widget tree, so we shouldn't need to add this here

kenzieschmoll (approved):
a couple nits then lgtm

packages/devtools_app/lib/src/screens/debugger/codeview_controller.dart
elliette: done!

packages/devtools_app/test/debugger/debugger_scripts_test.dart
elliette: done!

Issue #5797, opened 2023-05-15
DCM not working on master branch


CoderDake:
https://github.com/flutter/devtools/actions/runs/4983109255/jobs/8919718036
![Screenshot 2023-05-15 at 1 09 15 PM](https://github.com/flutter/devtools/assets/1386322/d31bd319-6b5e-453d-9a74-d68a72e2d8db)

PR #5829, opened 2023-05-19
Don't run DCM workflow when merging to master

elliette:
Fixes https://github.com/flutter/devtools/issues/5797

Changes introduced:
- only runs DCM workflow on PRs, not when merging to master
- adds the run-dcm-workflow label for daily bump PRs

### DIFF
--- a/.github/workflows/daily-dev-bump.yaml
+++ b/.github/workflows/daily-dev-bump.yaml
@@ -107,8 +107,7 @@ jobs:
           # Change github credentials back to the actions bot.
           GH_TOKEN="$ORIGINAL_GH_TOKEN"
 
-          gh pr edit $PR_URL $FLAGS --add-label "run-dcm-workflow"
-          gh pr edit $PR_URL $FLAGS --add-label "autosubmit"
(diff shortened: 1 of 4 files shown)


kenzieschmoll:
Does the DCM workflow need to be run for this PR?

elliette:
> Does the DCM workflow need to be run for this PR?

Ah, we were skipping the label reminder if there were no changes to Dart files (which is why that never ran). Now we are also skipping the DCM workflow itself if there are no changes to Dart files.

.github/workflows/daily-dev-bump.yaml
```diff
@@ -107,6 +107,7 @@ jobs:
           # Change github credentials back to the actions bot.
           GH_TOKEN="$ORIGINAL_GH_TOKEN"
 
+          gh pr edit $PR_URL $FLAGS --add-label "run-dcm-workflow"
           gh pr edit $PR_URL $FLAGS --add-label "autosubmit"
```
kenzieschmoll, line 111: nit: I think this can be a comma separated list:
`"autosubmit,run-dcm-workflow"`

kenzieschmoll (approved):
lgtm

PR #5842, opened 2023-05-24
Show an error message if sourcemaps are disabled

elliette:
See b/284033857

Will require updates to the internal `DevToolsExtensionPoints` once the change is rolled into g3.


kenzieschmoll (approved):
one nit but lgtm

CoderDake (approved):

PR #5853, opened 2023-05-26
Dismissible notifications do not expire

elliette:
Follow-up to https://github.com/flutter/devtools/pull/5822

I think it makes sense for dismissible notifications to not expire, since the fact that they are dismissible means we expect the user to take action on them.


kenzieschmoll:
Is this how VS code behaves with notifications that are dismissible? @DanTup

elliette:
> Is this how VS code behaves with notifications that are dismissible? @DanTup

It is! (Saying this as someone who currently has VS Code notifications lol)

Screenshot 2023-05-26 at 11 16 53 AM

kenzieschmoll (approved):

DanTup:
In VS Code, all notifications (besides things like progress notifications) are dismissable and they never expire, however:

1. They will "hide" themselves (into the notification bell - where you can get back to them) after 10-15s(?)
2. Another notification with exactly the same text will dismiss/replace any previous one

Eg., if I open a project and it says "Some things can be fixed with 'dart fix'" and I just ignore it, then after 15s it will disappear from the screen, but if I click the notification bell it's still there. The API for notifications (which returns which button was clicked) will sit waiting indefinitely if the user never responds to the message (even if the notification has collapsed into the bell icon).

Issue #5707, opened 2023-04-24
Clicking on a stackframe takes you to the start of the file, not the line location


elliette:
Broken out from https://github.com/flutter/devtools/issues/5703, find repro steps there

![switching files](https://user-images.githubusercontent.com/21270878/234134565-5634f382-b14e-44f0-a02b-399e00b96086.gif)

PR #5854, opened 2023-05-26
Add a debugger panel integration test

elliette:

Adds an integration test for basic debugger panel functionality.


packages/devtools_app/integration_test/test/live_connection/debugger_panel_test.dart
```diff
@@ -0,0 +1,155 @@
...
+    expect(testApp.vmServiceUri, isNotNull);
+  });
+
+  testWidgets('Debugger panel', (tester) async {
+    await pumpAndConnectDevTools(tester, testApp);
+    await switchToScreen(tester, ScreenMetaData.debugger);
+    await tester.pump(safePumpDuration);
+
+    logStatus(
+      'looking for the main.dart file',
```
kenzieschmoll, line 31: super nit: no need for trailing comma if this will fit on one line. Here and for logs below

packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -1066,6 +1067,7 @@ class _LinesState extends State<Lines> with AutoDisposeMixin {
           builder: (context, focusLine, _) {
             final isFocusedLine = focusLine == lineNum;
             return LineItem(
+              key: Key('Line Item $lineNum'),
```
kenzieschmoll, line 1070: do we need to add keys just for testing? I think we can accomplish finding these widgets with test finders instead. For example you could look for all the LineItem widgets and grab the first instead of looking for Line Item 1

kenzieschmoll: Adding keys can have more side effects than just adding a debug tag to your widget, so in general, we should only use keys when we are using them for a reason rooted in the flutter lifecycle, not just for testing purposes. I always go back to this video, which gives a good explanation: https://www.youtube.com/watch?v=kn0EOS-ZiIc

packages/devtools_app/integration_test/test/live_connection/debugger_panel_test.dart
```diff
@@ -0,0 +1,155 @@
...
+
+    final countVariableFinder = find.textContaining('count:');
+    expect(countVariableFinder, findsOneWidget);
+
+    logStatus(
+      'switching stackframes',
+    );
+
+    // Tap on the stackframe:
+    await tester.tap(frameFinder);
```
kenzieschmoll, line 111: Are we supposed to be switching to a different stack frame than the one we are already on?

packages/devtools_app/integration_test/test/live_connection/debugger_panel_test.dart
```diff
@@ -0,0 +1,155 @@
...
+  final lineWidget = getWidgetFromFinder<LineItem>(lineItemFinder);
+  return lineWidget.lineContents.toPlainText();
+}
+
+bool isLineFocused(Finder lineItemFinder) {
+  final lineWidget = getWidgetFromFinder<LineItem>(lineItemFinder);
+  return lineWidget.focused;
+}
+
+Finder findStackFrameWithText(String text) => find.byWidgetPredicate(
```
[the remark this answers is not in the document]
kenzieschmoll: example of using `find.descendant` https://github.com/flutter/devtools/blob/master/packages/devtools_test/lib/src/integration_test/integration_test_utils.dart#L42

packages/devtools_app/integration_test/test/live_connection/debugger_panel_test.dart
elliette: Ah good point! Removed the commas

packages/devtools_app/integration_test/test/live_connection/debugger_panel_test.dart
elliette: We are! Goal is to catch https://github.com/flutter/devtools/issues/5707

packages/devtools_app/lib/src/screens/debugger/codeview.dart
elliette: That was a helpful video! I've removed the keys. We want to make sure the gutter/lines get correctly scrolled into view (for go-to-line and for switching to a new stack frame), so when that happens we first visible line item won't be line item 1. 

I've switched to use finders to get the gutter item and line item based on their text content (eg '"40"` and `"count++"'`) and then verifying that they are horizontally aligned.

packages/devtools_app/integration_test/test/live_connection/debugger_panel_test.dart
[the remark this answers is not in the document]
elliette: Oh nice! Switched to using `find.descendant` (and `find.ancestor` for the gutter and line items)

kenzieschmoll (approved):

PR #5858, opened 2023-05-29
Add DartDevtoolWorkflowBot back to dcm checks

CoderDake:
The way that labels resolve for our daily dev bump PRs make it difficult to use the `autosubmit` label. So they are failing to autosubmit now.

This PR adds back the DartDevtoolWorkflowBot exceptions so that the dcm workflows will run, and have a chance to pass on the Daily Dev Bumps.

### DIFF
--- a/.github/workflows/dcm.yaml
+++ b/.github/workflows/dcm.yaml
@@ -33,7 +33,7 @@ jobs:
       needs-checkout-merge: true
 
   dcm:
-    if: contains(github.event.pull_request.labels.*.name, 'run-dcm-workflow') 
+    if: contains(github.event.pull_request.labels.*.name, 'run-dcm-workflow') || github.event.pull_request.user.login == 'DartDevtoolWorkflowBot'
     name: Dart Code Metrics
     needs: flutter-prep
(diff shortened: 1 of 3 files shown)


kenzieschmoll:
I will defer to @elliette on this PR, since she has been working with the dcm workflow.

polina-c (approved):

.github/workflows/dcm.yaml
```diff
@@ -29,11 +29,11 @@ jobs:
     uses: ./.github/workflows/flutter-prep.yaml
     with:
       os-name: ubuntu
-      requires-label: "run-dcm-workflow"
+      requires-label: ${{ github.event.pull_request.user.login == 'DartDevtoolWorkflowBot' && "" || "run-dcm-workflow" }}
```
elliette, line 32: I might be mistaken, but I think this means that the `flutter_prep` workflow will always run for the daily dev bumps (see:https://github.com/flutter/devtools/blob/master/.github/workflows/flutter-prep.yaml#LL28C15-L28C15) Not a big deal, but probably not what's intended

PR #5865, opened 2023-05-31
Add support for routing to a standalone screen

kenzieschmoll:
All this work is hidden behind the new feature flag `vsCodeSidebarTooling` (https://github.com/flutter/devtools/issues/5868).

This PR adds a route for each 'standalone' screen, which will be used to power embedded tooling UIs in VS Code. Each 'standalone' screen has its own route for accessing tool GUIs from outside of DevTools main navigation structure. So, for example, VS code could load a url like this in an iFrame to show a flutter panel or an interactive property editor:

- http://localhost:62874/vsCodeFlutterPanel?embed=true

For these standalone tools, we will not include the devtools scaffolding that provides the header and the footer like we do for embedded DevTools pages. The fact that these GUIs are hosted by devtools is more of an implementation detail and likely isn't something the user needs to be aware of.

Screenshot 2023-06-02 at 1 56 16 PM

In the future, we might be able to improve this routing using deferred loading or even a separate entry point that we compile separately from the main devtools entry point. However, for the MVP, I opted to hook into our existing routing system. 

I'm open to suggestions for better naming than "standalone". I thought about using the term "embedded" but didn't want to overloaded that term even more. I thought about "tabless" too, but that seems to tie this feature too close to the implementation details of our navigation structure in a way that will not be meaningful if that ever changes.

@DanTup @jacob314


DanTup:
> I'm open to suggestions for better naming than "standalone".

Does the name need to be declared statically, or could we just give each "id" its own route:

- `/sidebar`
- `/propertyEditor`

If not, some other options (though I'm not sure they're better than standalone):

- `/extension?id=...`
- `/tool?id=...`
- `/external?id=...`

kenzieschmoll:
> For these standalone tools, we might also want to hide the DevTools footer like we hide the header for embedded DevTools pages. The fact that these GUIs are hosted by devtools is more of an implementation detail and likely isn't something the user needs to be aware of.

@DanTup what is your opinion on this?

DanTup:
> we might also want to hide the DevTools footer like we hide the header for embedded DevTools pages. The fact that these GUIs are hosted by devtools is more of an implementation detail and likely isn't something the user needs to be aware of.

I agree. I think things like header/footer make it feel more like "tool A embedded inside tool B" - and although it is, I think the experience is better if that's somewhat invisible and it looks and feels just like a single tool.

kenzieschmoll:
> I agree. I think things like header/footer make it feel more like "tool A embedded inside tool B" - and although it is, I think the experience is better if that's somewhat invisible and it looks and feels just like a single tool.

I have removed the DevTools scaffolding and updated the screenshot

kenzieschmoll:
@DanTup this is ready for review and can land now if you're happy with it. This is behind an experiment flag so the support will not be shipped until that flag is flipped on and we have a real use case to use it for - e.g. the first vs code sidebar tool we implement).

DanTup (approved):
LGTM!

PR #5866, opened 2023-06-01
Fix scrolling to selected stack frame

elliette:
The fix for https://github.com/flutter/devtools/issues/5707 was regressed by https://github.com/flutter/devtools/pull/5827

The good news is the debugger panel test I'm working on did catch the regression: https://github.com/flutter/devtools/pull/5854

The bad news is https://github.com/flutter/devtools/pull/5827 was already submitted!

In any case, this fixes the regression and now when you click on a stack frame we scroll to the correct location:

scrolling to stack frame

### DIFF
--- a/packages/devtools_app/lib/src/screens/debugger/call_stack.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/call_stack.dart
@@ -44,7 +44,9 @@ class _CallStackState extends State<CallStack>
             final frame = stackFrames[index];
             return _buildStackFrame(
               frame,
-              frame == selectedFrame || frame == _clickedOnFrame,
+              _clickedOnFrame != null
+                  ? frame == _clickedOnFrame
+                  : frame == selectedFrame,
             );
           },
         );
--- a/packages/devtools_app/lib/src/screens/debugger/call_stack.dart
+++ b/packages/devtools_app/lib/src/screens/debugger/call_stack.dart
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
+
 import 'package:flutter/material.dart' hide Stack;
 import 'package:vm_service/vm_service.dart';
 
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/screens/debugger/call_stack.dart
```diff
@@ -20,6 +20,8 @@ class CallStack extends StatefulWidget {
 
 class _CallStackState extends State<CallStack>
     with ProvidedControllerMixin<DebuggerController, CallStack> {
+  StackFrameAndSourcePosition? selectedFrame;
```
kenzieschmoll, line 23: it looks like this can get out of sync with `controller.selectedStackFrame` now that this widget is not responding to changes from this value listenable. Can we still pull the selected stack from from `controller.selectedStackFrame` so that the UI is always in sync with the controller?

elliette: We can, though if we just do that the UI will seem unresponsive (see gif - the row is pink when it is tapped, and then turns grey when the `controller.selectedStackFrame` is updated). 

delay in selecting stack frame

Explanation:
- we currently have a `postFrameCallback` in `codeview.dart` that will scroll to the selected line when the `selectedStackFrame` is updated
- we are waiting until after the script is successfully parsed and loaded to update the `selectedStackFrame` (not waiting was what caused the regression)
- therefore there is a lag after a user has clicked and before the `selectedStackFrame` is updated

I've updated it so that we are highlighting if either the row has just been clicked, or if the frame is selected. It means that for a short time we have two frames highlighted, but at least it seems responsive. WDYT?


delay 2

packages/devtools_app/lib/src/screens/debugger/call_stack.dart
```diff
@@ -38,7 +42,10 @@ class _CallStackState extends State<CallStack>
           itemExtent: defaultListItemHeight,
           itemBuilder: (_, index) {
             final frame = stackFrames[index];
-            return _buildStackFrame(frame, frame == selectedFrame);
+            return _buildStackFrame(
+              frame,
+              frame == selectedFrame || frame == _clickedOnFrame,
```
kenzieschmoll, line 47: so that we don't have two frames highlighted, we could change this to:
`_clickedOnFrame != null ? frame == _clickedOnFrame : frame == selectedFrame`

elliette: Good idea, done!

kenzieschmoll (approved):

Issue #5882, opened 2023-06-06
Consider disabling  flutter/devtools-reviewers for automatic version bump PRs


kenzieschmoll:
For automatic version bump PRs, we are adding `fluttergithubbot` as the reviewer. Example: https://github.com/flutter/devtools/pull/5881. When `fluttergithubbot` is the reviewer, we don't also need a reviewer from `flutter/devtools-reviewers`, but since this is added for every PR someone on the team gets an unnecessary review request. Can we make a rule that `flutter/devtools-reviewers` is only added when `fluttergithubbot` is not the reviewer? CC @CoderDake

PR #5892, opened 2023-06-08
Reduce version bump codeowner spam

CoderDake:
To avoid top level code-owners from being spammed, the CODEOWNERS are being removed for files involved in the version bumps. For example see[this auto bump PR](https://github.com/flutter/devtools/pull/5890/files)
After doing some research, it looks like it is possible to set files/directories so they have no owners. 

See[about-code-owners](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) which describes the situation as:
```
# In this example, @octocat owns any file in the `/apps`
# directory in the root of your repository except for the `/apps/github`
# subdirectory, as its owners are left empty.
/apps/ @octocat
/apps/github
```

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -27,9 +27,9 @@ vm_developer/       @bkonyi
 # Tooling files
 /tool/              @CoderDake
 
-# DartDevtoolWorkflowBot added here to prevent other CODEOWNERS
-# from being added to auto generated dev bumps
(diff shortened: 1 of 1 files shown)


kenzieschmoll (approved):

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/5882

PR #5917, opened 2023-06-13
Run integration tests connected to a `chrome` device

elliette:

Follow up to https://github.com/flutter/devtools/pull/5854

- Adds support for running integration tests connected to a `chrome` device
- Runs all supported tests connected to a `chrome` device during CI

The goal is to help catch regressions that are specific to web apps.


packages/devtools_app/integration_test/run_tests.dart
```diff
@@ -26,23 +26,41 @@ const _testDirectory = 'integration_test/test';
 const _testSuffix = '_test.dart';
 const _offlineIndicator = 'integration_test/test/offline';
 
+// A mapping of test app device to the supported tests for that device.
+final _deviceToTestPaths = <String, List<String>>{
+  'chrome': [
+    'debugger_panel_test.dart',
+    'app_test.dart',
+    // TODO(https://github.com/flutter/devtools/issues/5874): Enable once supported on web.
+    // 'eval_and_browse_test.dart',
```
elliette, line 35: FYI @annagrin @polina-c

packages/devtools_app/integration_test/run_tests.dart
```diff
@@ -26,23 +26,41 @@ const _testDirectory = 'integration_test/test';
 const _testSuffix = '_test.dart';
 const _offlineIndicator = 'integration_test/test/offline';
 
+// A mapping of test app device to the unsupported tests for that device.
```
kenzieschmoll, line 29: nit: three / for dart doc

packages/devtools_app/integration_test/run_tests.dart
```diff
@@ -57,6 +75,14 @@ void main(List<String> testRunnerArgs) async {
   }
 }
 
+bool _isUnsupportedForDevice(String testPath, {required String testAppDevice}) {
```
kenzieschmoll, line 78: can we rename this method to be worded in positive tense `isSupportedForDevice`? the double negatives make this a bit hard to follow

packages/devtools_app/integration_test/run_tests.dart
```diff
@@ -57,6 +75,14 @@ void main(List<String> testRunnerArgs) async {
...
+  if (testAppDevice == 'flutter-tester') return false;
```
kenzieschmoll, line 79: can we create an enum for the testAppDevice types? `flutter-tester` and `chrome` for now, but likely `cli` in the future as well.

packages/devtools_app/integration_test/test_infra/run/_test_app_driver.dart
```diff
@@ -51,6 +53,9 @@ abstract class _TestApp with IOMixin {
   /// of app.
   final String testAppPath;
 
+  /// The device the test app should run on, e.g. flutter-tester, chrome.
+  final String testAppDevice;
```
kenzieschmoll, line 57: let's use the enum value here (see previous comment)

packages/devtools_app/integration_test/test_infra/run/run_test.dart
```diff
@@ -273,32 +276,37 @@ class TestRunnerArgs {
     final argWithTestTarget =
         args.firstWhereOrNull((arg) => arg.startsWith(testTargetArg));
     final target = argWithTestTarget?.substring(testTargetArg.length);
-    assert(
-      target != null,
-      'Please specify a test target (e.g. ${testTargetArg}path/to/test.dart',
-    );
```
kenzieschmoll, line 279: why are we allowing target to be null now?

elliette: We need to parse the args to see if the `testAppDevice` arg is provided, but the parsing itself was asserting that the `testTarget` wasn't null. Therefore if we were trying to run all web tests (and therefore not specifying a test target), it would fail with the assertion error.

packages/devtools_app/integration_test/run_tests.dart
elliette: Done, also moved to be a private static field of the `TestAppDevice` enhanced enum

packages/devtools_app/integration_test/run_tests.dart
elliette: Done, also changed to a method on the `TestAppDevice` enhanced enum

packages/devtools_app/integration_test/run_tests.dart
elliette: Done!

packages/devtools_app/integration_test/test_infra/run/_test_app_driver.dart
elliette: Done!

packages/devtools_app/integration_test/test_infra/run/run_test.dart
```diff
@@ -315,3 +316,45 @@ class TestRunnerArgs {
...
+  }
+
+  String get argName {
+    switch (this) {
+      case flutterTester:
+        return 'flutter-tester';
+      case chrome:
+        return 'chrome';
+    }
+  }
```
kenzieschmoll, line 353: instead of making this a getter, we can have a field `argName` that is set from a constructor 

```dart
enum TestAppDevice {
  flutterTester('flutter-tester'),
  chrome('chrome');

  const TestAppDevice(this.argName);

  final String argName;
```

packages/devtools_app/integration_test/test_infra/run/run_test.dart
elliette: Not 100% sure the change I've made is what you asked for, but I added the assertion back and moved the check for whether the test supports the test app device from `main` to `_runTest` (so that we only do the check once we have a test target). LMK what you think!

packages/devtools_app/integration_test/test_infra/run/run_test.dart
elliette: Oooh I didn't know that was possible with enums! Switched to setting it from a constructor

packages/devtools_app/integration_test/test_infra/run/run_test.dart
```diff
@@ -300,10 +310,50 @@ class TestRunnerArgs {
   /// This value will only be used for tests with live connection.
   late final String? testAppUri;
 
+  /// The type of device for the test app to run on.
+  late final String testAppDevice;
```
kenzieschmoll, line 314: can this be type `TestAppDevice`?

packages/devtools_app/integration_test/run_tests.dart
```diff
@@ -61,6 +61,13 @@ Future<void> _runTest(
   List<String> testRunnerArgs,
   String testFilePath,
 ) async {
+  final parsedArgs = TestRunnerArgs(testRunnerArgs);
+  final testAppDevice = TestAppDevice.fromArgName(parsedArgs.testAppDevice)!;
+  if (!testAppDevice.supportsTest(testFilePath)) {
+    // Skip test, since it is not supported for device.
+    return;
+  }
+
```
kenzieschmoll, line 70:
```suggestion
  TestRunnerArgs testRunnerArgs
) async {
  if (!testRunnerArgs.testAppDevice.supportsTest(testRunnerArgs.testTarget)) {
    // Skip test, since it is not supported for device.
    return;
  }

```

packages/devtools_app/integration_test/test_infra/run/run_test.dart
kenzieschmoll: Added a couple suggestions below that would need to be paired with this:
```dart
  if (testTargetProvided) {
    final args = TestRunnerArgs(testRunnerArgs);
    // TODO(kenz): add support for specifying a directory as the target instead
    // of a single file.
    await _runTest(args);
  } else {
    // Run all tests since a target test was not provided.
    final testDirectory = Directory(_testDirectory);
    final testFiles = testDirectory
        .listSync(recursive: true)
        .where((testFile) => testFile.path.endsWith(_testSuffix));
    for (final testFile in testFiles) {
      final testTarget = testFile.path;
      final args = TestRunnerArgs([
        ...testRunnerArgs,
        '${TestRunnerArgs.testTargetArg}$testTarget',
      ]);
      await _runTest(args);
    }
  }
```

elliette: Oh nice! I think I've applied all the suggestions

packages/devtools_app/integration_test/test_infra/run/run_test.dart
elliette: Yes, switched it over

packages/devtools_app/integration_test/test_infra/run/run_test.dart
```diff
@@ -290,8 +289,10 @@ class TestRunnerArgs {
 
     final argWithTestAppDevice =
         args.firstWhereOrNull((arg) => arg.startsWith(testAppDeviceArg));
-    testAppDevice = argWithTestAppDevice?.substring(testAppDeviceArg.length) ??
-        'flutter-tester';
+    testAppDevice = TestAppDevice.fromArgName(
+      argWithTestAppDevice?.substring(testAppDeviceArg.length) ??
+          'flutter-tester',
```
kenzieschmoll, line 294: nit: use TestAppDevice.flutterTester.argName instead of the 'flutter-tester' string

kenzieschmoll (approved):
one nit and lgtm

PR #5921, opened 2023-06-15
Connect to VS Code using a JSON-RPC API over postMessage or WebSocket

DanTup:
@jacob314 @kenzieschmoll interested in your thoughts on this (I don't expect to land it like this, it's just for discussion).

It's a bit of an evolution of #5747 (although only focusing on the DevTools-hosted panel for now). Using something like JSON-RPC between the Dart panel and the VS Code extension seems better than the custom code in #5747 (and would make it easier to make available to TypeScript extensions without having to build all API classes also in TypeScript).

It expands the panel Kenzie added to show information about the selected device from VS Code, and also a button that changes the device. It listens to events from VS Code for whenever the selected device changes.

When the panel is run with `?embed=true` on the querystring, it uses `postMessage` to communicate to VS Code. If you run it without, it'll show a form that lets you paste in a WebSocket URL and will connect over that instead (the API behaves identically, allowing you to use hot reload and debugging while still interacting with VS Code).

Here's a video showing it working with both setups:

https://drive.google.com/file/d/14kvLNJHJXq2P8F8YKeJxEuTZruUleoK9/view?usp=sharing

At 0:22 I run a "Dart: Connect External DevTools Sidebar" command which provides a WS URL (it's offscreen just in the debug console atm, but it would be shown at the top somewhere easily copyable), which is then pasted into the DevTools sidebar page.

Currently the API only includes some VS Code stuff (`executeComand`, `getSelectedDevice`, `onSelectedDeviceChanged`) but could be expanded to have `lsp` and other APIs that the Dart extension can provide access to.

The VS Code changes (which needs some tidying up) are on the [Dart-Code `new-sidebar` branch](https://github.com/Dart-Code/Dart-Code/tree/new-sidebar). The APIs specifically [here](https://github.com/Dart-Code/Dart-Code/blob/new-sidebar/src/extension/api/dart_tooling_api.ts).

### DIFF
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -297,7 +297,7 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
   Map<String, UrlParametersBuilder> get _standaloneScreens {
     return {
       for (final type in StandaloneScreenType.values)
-        type.name: (_, __, args, ___) => type.screen,
+        type.name: (_, __, args, ___) => type.build(embedded: isEmbedded(args)),
     };
   }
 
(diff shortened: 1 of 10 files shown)


DanTup:
Closing in favour of #6104.

PR #5937, opened 2023-06-21
Add class type agenda to help for each memory tab.

polina-c:


kenzieschmoll:
Can you add screenshots to the PR description? thanks

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -62,13 +63,20 @@ class _SnapshotItemContent extends StatelessWidget {
               crossAxisAlignment: CrossAxisAlignment.start,
               children: [
                 Expanded(
-                  child: Markdown(
-                    data: _snapshotDocumentation,
-                    onTapLink: (text, url, title) async =>
-                        await launchUrl(url!),
+                  child: ListView(
+                    children: [
+                      SizedBox(
```
kenzieschmoll, line 68: instead of wrapping in a ListView and a SizedBox, why not just put the `ClassTypeLegend` in the parent column before the `MoreInfoLink`?

polina-c: Because we want it to be scrolled together with the main content. So, it should be in ListView, that provides scrolling.

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
```diff
@@ -671,6 +671,8 @@ class _ProfileHelpLink extends StatelessWidget {
           const Text('The allocation profile tab displays information about\n'
               'allocated objects in the Dart heap of the selected\n'
               'isolate.'),
+          const SizedBox(height: denseSpacing),
+          const ClassTypeLegend(),
           MoreInfoLink(
             url: DocLinks.profile.value,
             gaScreenName: '',
```
kenzieschmoll, line 678: This _ProfileHelpLink widget is duplicated below. We can create a shared widget MemoryTabHelp to share this code across tabs.

packages/devtools_app/lib/src/screens/memory/shared/widgets/shared_memory_widgets.dart
```diff
@@ -57,3 +57,25 @@ class HeapClassView extends StatelessWidget {
...
+
+class ClassTypeLegend extends StatelessWidget {
+  const ClassTypeLegend({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.start,
+      children: [
+        const Text('Class types legend:'),
```
kenzieschmoll, line 69: nit: r/types/type

packages/devtools_app/lib/src/screens/memory/shared/widgets/shared_memory_widgets.dart
```diff
@@ -57,3 +57,25 @@ class HeapClassView extends StatelessWidget {
     );
   }
 }
+
+class ClassTypeLegend extends StatelessWidget {
```
bkonyi, line 61: Ubernit: can you add some documentation describing this widget?

packages/devtools_app/lib/src/screens/memory/panes/diff/diff_pane.dart
```diff
@@ -62,13 +63,20 @@ class _SnapshotItemContent extends StatelessWidget {
...
+                        height: 450,
+                        child: Markdown(
+                          data: _snapshotDocumentation,
+                          onTapLink: (text, url, title) async =>
```
bkonyi, line 72: Nit: does this need to be `async`? I'm not sure if the call to `onTapLink` is `await`ed.

bkonyi (approved):

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
polina-c: The structure is similar, but the is different. It can become more different in future. And it is just for two tabs. I prefer not to generalize for now.

kenzieschmoll (approved):

PR #5961, opened 2023-06-27
Initial structure for snapshot analysis.

polina-c:


packages/devtools_app/lib/src/app.dart
```diff
@@ -290,6 +291,17 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
           ),
         );
       },
+      snapshotAnalysisScreenId: (_, __, args, ____) {
```
kenzieschmoll, line 294: why do we need a new screen for this tool? I thought we were going to reuse the diff / snapshot view on the memory screen to support loading an offline snapshot.

packages/devtools_app/lib/src/framework/landing_screen.dart
```diff
@@ -352,6 +359,46 @@ class AppSizeToolingInstructions extends StatelessWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    final textTheme = Theme.of(context).textTheme;
+    return LandingScreenSection(
+      title: 'Snapshot Analysis',
+      child: Column(
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          Text(
+            'Analyze and diff the saved memory snapshots',
```
kenzieschmoll, line 375: wording nit: "Analyze and diff memory heap snapshots"

packages/devtools_app/lib/src/framework/landing_screen.dart
```diff
@@ -352,6 +359,46 @@ class AppSizeToolingInstructions extends StatelessWidget {
...
+            style: textTheme.titleMedium,
+          ),
+          const SizedBox(height: denseRowSpacing),
+          Text(
+            'Analyse memory heap snapshots, previously saved with DevTools or Dart leak_tracker.',
```
kenzieschmoll, line 380: sp: Analyze

"heap snapshots that were previously saved from DevTools or package:leak_tracker"

and making 'package:leak_tracker' a hyperlink here would be good.

packages/devtools_app/lib/src/app.dart
polina-c: Yes, we will reuse memory diff widgets, but the entire screen will be different.

kenzieschmoll: This is probably fine for now as you are working on implementing the UI, but we should revisit this design with flutter.dev/go/static-tooling-in-devtools in mind. I think we should provide an experience integrated with the Memory screen directly that supports a seamless "upgrade" and "downgrade" experience for when your app connection state changes. For example, you can imagine it would also be useful to compare a snapshot from a running application to one that you have saved on disk as a baseline. 

For now though, only add this screen when the experimental flag is enabled.

polina-c: Yes, i will make the widgets supporting this story. The screens will share the widgets.

polina-c: I made the instructions conditional on landing page. I think it is ok to make routing non-conditional, because users will not discover the route without instructions.

Makes sense?

polina-c: Non conditional routing will make it possible to analyze snapshots from analyzer, without rebuilding devtools.

kenzieschmoll: Not hiding this behind a flag would mean that a user could access this screen, even when the experiment is disabled, by manually changing the url. While this is unlikely, we shouldn't give users access to unshipped features for ease of development.

It is simple to enable experiments for local development - we have the VS code run configuration for this, or you can replace the experiment level with `true` when developing.

polina-c:
Closing as we agreed on different design.

Issue #5974, opened 2023-06-29
Notification invisible background prevents clicks.


CoderDake:
When exploring https://github.com/flutter/devtools/issues/4972

I found that the notifications block more of the screen than they appear to. There is an invisible background that prevents elements from being clicked behind the notifications

PR #5975, opened 2023-06-29
Expand notifications to fill the space they block

CoderDake:
Fixes https://github.com/flutter/devtools/issues/5974

This PR expands notifications to completely cover the area that they block.

Before Fix:

https://github.com/flutter/devtools/assets/1386322/b28e1051-6d37-4927-b259-52645733824d

After Fix:

https://github.com/flutter/devtools/assets/1386322/87923d0e-6e43-48ba-97a7-1d6c56e3f182


kenzieschmoll:
Instead of making the notifications wider than they need to be, why can't we just change the amount of space that they are "blocking" to match the width of the notificaiton?

kenzieschmoll:
Looks like `_notificationWidth` is calculated based off of a fixed value `_notificationHeight` instead of being calculated based off of the content of the notification itself. The width really should be more of a max constraint that a fixed width constraint

CoderDake:
@kenzieschmoll I attempted multiple methods of accomplishing that, unfortunately none of them were fruitful
- IgnorePointer would prevent dismissible notifications from being manually dismissed
- I tried making a container that would occupy the space beside the notification, so hopefully i could IgnorePointer in that area. Unfortunately even though I could create a minimal app showing how to share the space with a container, trying to do it with the notifications key messing with the layout.
- I researched other methods and it looks like there is a third party package we could use to accomplish a click through background.

While researching those I realized that other notification systems have static width notifications, so I figured it could be appropriate for us to do the same.

CoderDake:
> _notificationWidth

@kenzieschmoll  unfortunately that width represents a column on the right side that the individual notifications are contained within.

You mentioning this kind of makes me wonder if I could get that top level notification wrap to shrink onto the content. So I'll take a look to see if I can get the overlay entry to shrink onto the notifications itself, given the contstraints.

CoderDake:
@kenzieschmoll I just double checked my knowledge and it appears that my intuition was right that the overlay entry wraps all of the notifications, so even if I could get it to shrink to the largest one there would still be an invisible area that we couldn't click through.
I get the feeling that if we wanted to allow dynamic width notifications we would need to rewrite the notifications setup so that each notification is it's own separate OverlayEntry

kenzieschmoll (approved):

Issue #5955, opened 2023-06-26
Document how to get field information for retaining path.


polina-c:
See https://github.com/dart-lang/sdk/issues/50546

Lack of information about field name makes it harder to troubleshoot memory leaks.

polina-c:
It is costly and noisy to have field information for retaining path statistics.
There is API to get this information during debugging for concrete instances: leak_tracker.formattedRetainingPath(WeakReference ref)

We need to add documentation to use this API while troubleshooting memory.

PR #5984, opened 2023-07-04
Add location to retaining path.

polina-c:
Fixes https://github.com/flutter/devtools/issues/5955


polina-c:
To triage, yes it is still planned, but slipped to to be done later :)

PR #5989, opened 2023-07-05
Remove methods toJson and fromJson.

polina-c:
Use serialization for HeapSnapshotGraph, to serialize AdaptedHeapData.
It makes the code conciser and improve maintainability.


bkonyi:
Can you update the description? I don't have the context as to what this change is supposed to accomplish.

bkonyi (approved):
Looks like the coffee just needed to kick in. LGTM, but still consider updating the description :)

Issue #5954, opened 2023-06-26
Enable rename for snapshot


polina-c:
One of UX studies showed that lack of ability to rename snapshot is painpoint.

PR #5997, opened 2023-07-06
Add a context menu to rename or delete a heap snapshot from the list

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/5954. 

rename-or-delete


polina-c:
Thanks for putting it together. This way the feature does not look to be discoverable. Why do not we add icon with pen to the selected snapshot? Or, we can add context menu (vertical `...`) with two items: Delete, Rename.

kenzieschmoll:
A user may want to edit a snapshot name even if it is not selected, so for this reason, having an icon that only appears for the selected snapshot may also have usability issues. Adding an overflow menu (vertical `...`) for each snapshot item could work, though the UI may look busy with that icon on each row.

I would argue that right click is discoverable - double click was admittedly a bonus item that users may or may not find.
CC @InMatrix

polina-c:
Renaming is not needed very often. So, I think it is ok to show the icon just for selected row, the same way as we do it for table:

Screenshot 2023-07-06 at 2 54 18 PM

Right click is discoverable in general, but not in DevTools, where we do not utilize it a lot. Even if user discovers it, they will not expect second row to become editable after selecting 'Rename' for first row.

Also, implementation with  vertical `...` will be conciser in code, and thus cheaper to support.

kenzieschmoll:
> they will not expect second row to become editable after selecting 'Rename' for first row.

This is not the case. Only the row that was right clicked will become editable. 

I will look into adding the overflow menu icon for the selected row, but I don't see an issue with also leaving the right click functionality. We haven't used right click much in the past because flutter's context menu support for web wasn't as good as it is now.

polina-c:
> I will look into adding the overflow menu icon for the selected row, but I don't see an issue with also leaving the right click functionality. We haven't used right click much in the past because flutter's context menu support for web wasn't as good as it is now. 

Would it be possible to incapsulate right click handling into mixin somehow? So that it does not crowd the code of the widget, where it is used? 

BTW, in modern cross-plat UI, context actions are more and more behind vertical ellipsis.
So, it is becoming user habit to assume there is no actions, if there is no ellipsis:

Screenshot 2023-07-07 at 8 31 44 AM

polina-c:
> This is not the case. Only the row that was right clicked will become editable.

On the attached video I see click on 'Rename' just one time, and then I see two snapshots are renamed. How it is?

kenzieschmoll:
> On the attached video I see click on 'Rename' just one time, and then I see two snapshots are renamed. How it is?

This was from double clicking in the second row, which I've since removed.

kenzieschmoll:
Okay this is ready for another review. I removed the right click and double click functionality in favor of the context menu. The benefit of maintaining both affordances is not worth the cost. Updated the PR description.

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_list.dart
```diff
@@ -106,52 +107,195 @@ class _ListControlPane extends StatelessWidget {
...
+    required this.onDelete,
+  }) : super(key: key);
 
   final SnapshotItem item;
 
+  final int index;
+
+  final bool selected;
+
+  final ValueListenable<int?> editIndex;
```
polina-c, line 128: Can you add doc comment for the new members. Some are not obvious (editIndex, onEdit vs onEditComplete)

kenzieschmoll: done

polina-c (approved):
LGTM after comments

Issue #4353, opened 2022-08-05
Only log network traffic if a user is on the Network page


kenzieschmoll:
This is causing memory bloat for user application. See for more details and some suggested solutions (making this opt in, showing a warning on memory page when this is enabled, etc.) - https://github.com/dart-lang/sdk/issues/49506

mraleph:
I suggest this issue to be reprioritised to something closer to ~P1.

kenzieschmoll:
@mraleph fix for this is up for review and will make it into the next beta / stable release.

mraleph:
@kenzieschmoll thank you!

PR #5998, opened 2023-07-06
Stop logging http to timeline by default and warn users about memory implications

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/4353. (please review with Hide Whitespace enabled)

This stops enabling http timeline logging by default on the performance page (this can still be enabled from the timeline stream settings dialog on the Timeline Events view). This setting will continue to be enabled by default when a user switches to the network page.

Additionally, this PR adds a warning on the memory screen when this setting is enabled. Users have made several reports being confused by the memory dart:io is holding onto for debugging. This warning allows the user to disable the http logging and automatically dismisses.

http-logging-warning

### DIFF
--- a/packages/devtools_app/lib/src/shared/banner_messages.dart
+++ b/packages/devtools_app/lib/src/shared/banner_messages.dart
@@ -391,7 +391,7 @@ class HttpLoggingEnabledMessage {
       textSpans: [
         const TextSpan(
           text: '''
-Http traffic is being logged for debugging purposes. This may result in increased memory usage for your app. If this is not intentional, consider ''',
+HTTP traffic is being logged for debugging purposes. This may result in increased memory usage for your app. If this is not intentional, consider ''',
         ),
         TextSpan(
           text: 'disabling http logging',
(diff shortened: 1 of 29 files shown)


packages/devtools_app/lib/src/shared/banner_messages.dart
```diff
@@ -377,6 +375,47 @@ You are opting in to a high CPU sampling rate. This may affect the performance o
...
+
+  BannerMessage build(BuildContext context) {
+    final theme = Theme.of(context);
+    late final BannerWarning message;
+    message = BannerWarning(
+      key: key,
+      textSpans: [
+        const TextSpan(
+          text: '''
+Http traffic is being logged for debugging purposes. This may result in increased memory usage for your app. If this is not intentional, consider ''',
```
bkonyi, line 394: Nit: HTTP should always be capitalized in strings since it's an acronym.

packages/devtools_app/release_notes/NEXT_RELEASE_NOTES.md
```diff
@@ -18,7 +18,7 @@ TODO: Remove this section if there are not any general updates.
 TODO: Remove this section if there are not any general updates.
 
 ## Memory updates
-TODO: Remove this section if there are not any general updates.
+* Warn users when Http logging may be affecting their app's memory consumption - [#5998](https://github.com/flutter/devtools/pull/5998)
```
bkonyi, line 21: Ditto

bkonyi (approved):

PR #5999, opened 2023-07-07
Create a new shared `AnalyticsDropDownButton`

bkonyi:
Provides a wrapper around `RoundedDropDownButton` that provides support for Google Analytics.


packages/devtools_app/lib/src/shared/ui/vm_flag_widgets.dart
```diff
@@ -64,33 +63,34 @@ class CpuSamplingRateDropdown extends StatelessWidget {
...
-      child: Text(samplingRate.display),
+  ({DropdownMenuItem<String> item, String gaId}) _buildMenuItem(
+    CpuSamplingRate samplingRate,
+  ) {
+    return (
+      item: DropdownMenuItem<String>(
+        value: samplingRate.value,
+        child: Text(samplingRate.display),
+      ),
+      gaId: CpuSamplingRateExtension.fromValue(samplingRate.value).displayShort
```
kenzieschmoll, line 93: I think this can just be `samplingRate.displayShort`

packages/devtools_app/lib/src/shared/ui/vm_flag_widgets.dart
```diff
@@ -64,33 +63,34 @@ class CpuSamplingRateDropdown extends StatelessWidget {
...
+          gaDropDownId: gac.CpuProfilerEvents.profileGranularity.name,
+          message:
+              'The frequency at which the CPU profiler will sample the call stack',
+          value: safeValue,
+          items: [
+            _buildMenuItem(CpuSamplingRate.low),
+            _buildMenuItem(CpuSamplingRate.medium),
+            _buildMenuItem(CpuSamplingRate.high),
+          ],
+          onChanged: _onSamplingFrequencyChanged,
```
kenzieschmoll, line 79: we need to remove the analytics call from `_onSamplingFrequencyChanged` or else we will send analytics twice.

packages/devtools_app/lib/src/shared/ui/drop_down_button.dart
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  void _onChanged(T? newValue) {
+    if (sendAnalytics) {
+      final gaId = items
+              ?.firstWhereOrNull(
+                (element) => element.item == newValue,
+              )
+              ?.gaId ??
+          '<empty>';
```
kenzieschmoll, line 78: do we want to assert that gaId is not null? I don't think we should ever hit this case, and then if we did for some reason, would we still want to send analytics for `<empty>`?

bkonyi: `gaId` will never be null, but I think `newValue` could be if the selection is cleared. Is it possible for a drop down button selection to be cleared? If not, I think we can do a null assertion here and drop the `<empty>` case.

bkonyi: I'm going to assume that we can't clear the selection from the standard UI unless we explicitly set `value` to `null`. In that case, we'll just throw a `StateError` when `list.firstWhere(...)` fails to find a match.

kenzieschmoll (approved):

Issue #5990, opened 2023-07-05
Support static tooling in DevTools


kenzieschmoll:
Design doc: https://docs.google.com/document/d/1i6ISzzSKKDiiVftTnkCCvdgXEjK2eqYlPjCOKTjnvx8/edit?resourcekey=0-uyAiZEjFplyibsAkDAqKJw#

This will unblock things like a deep link validation tool and VS code sidebar tooling (https://github.com/flutter/devtools/issues/5868).

PR #6010, opened 2023-07-10
Add support for static tooling in DevTools and add a "Home" screen

kenzieschmoll:
This PR refactors the DevTools landing or "connect" screen to be a screen that is available from the tab bar, with and without a connected app.
- When there is no connected app, the home screen will show the connect dialog. 
- When there is a connected app, the home screen will show a summary of the app connection (everything that the device dialog in the lower right used to show). We can add more items to this home screen in the future as it makes sense.

The tab bar shows only tools that are available, so runtime tabs will be hidden when there is no connected app, but will appear once a connection is established.

static-tooling-with-footer

Note: The blue footer in this gif that indicates connection state will be implemented as a follow up PR.

Fixes https://github.com/flutter/devtools/issues/5990


kenzieschmoll:
FYI @InMatrix @nancyhu2023 @zoeyfan @chunhtai @hangyujin

packages/devtools_app/test/shared/connection_info_test.dart
```diff
@@ -275,48 +236,4 @@ void main() {
       },
     );
   });
-
-  group('VMFlagsDialog', () {
```
kenzieschmoll, line 279: moved to `vm_flag_widgets_test.dart`

kenzieschmoll:
From @polina-c 
> Minor thing: it looks as if App Size disappears, but it is still here on the very right.
Are you sure we want to reorder tabs?
Users remember tab order and reordering will create confusion.

Right now this is implemented so that only tabs that are available without an app connection will show up when there is no connected app (static mode). Then when an app connects, we will show other supported tabs. The screens are defined in a static order right now (see https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/app.dart#L478), so to keep the ordering consistent, we'd have to order all tabs that support static tooling first, and IMO that ordering won't always make sense. The most commonly used tools should still be ordered at the front of the list, even if that means the static tools are now inserted back to their rightful position amongst the entire set of screens.

polina-c (approved):

PR #6015, opened 2023-07-11
Color DevTools footer to indicate app connection state

kenzieschmoll:
Part of https://github.com/flutter/devtools/issues/5990, and a follow up change to https://github.com/flutter/devtools/pull/6010. See  flutter.dev/go/static-tooling-in-devtools for original proposal.

When an app is connected, the footer will be colored with the DevTools primary color. When no app is connected, the footer will be the same color as the DevTools background.

static-tooling-with-footer

CC @nancyhu2023 @InMatrix


polina-c:
Cool!

Minor thing: it looks as if App Size disappears, but it is still here on the very right.
Are you sure we want to reorder tabs?
Users remember tab order and reordering will create confusion.

packages/devtools_app/lib/src/framework/report_feedback_button.dart
```diff
@@ -14,7 +14,9 @@ import '../shared/globals.dart';
 import '../shared/theme.dart';
 
 class ReportFeedbackButton extends StatelessWidget {
```
polina-c, line 16: Can you add doc comment for this widget?

polina-c:
Would it make sense to check color before and after connection in devtools_app/integration_test/test/live_connection/app_test.dart?

kenzieschmoll:
> Would it make sense to check color before and after connection in devtools_app/integration_test/test/live_connection/app_test.dart?

Done.

> Minor thing: it looks as if App Size disappears, but it is still here on the very right.
Are you sure we want to reorder tabs?
Users remember tab order and reordering will create confusion.

This discussion should likely happen on https://github.com/flutter/devtools/pull/6010, since that PR is where this change is implemented. Moving the discussion there.

elliette (approved):

Issue #6019, opened 2023-07-11
Text selection in debugger panel is poor user experience


elliette:
![selecting codeview before](https://github.com/flutter/devtools/assets/21270878/e6699fcb-339e-4c07-89d0-ae8c136e83f4)

![selecting inspect before](https://github.com/flutter/devtools/assets/21270878/70533f46-b68f-4ace-9a90-f8392c5986b0)

![selecting before console](https://github.com/flutter/devtools/assets/21270878/291ec997-82af-45bc-9756-856262523131)

Issue #6018, opened 2023-07-11
Cannot "inspect" variables when running Devtools as a web app


elliette:
Inspect on macos:

![INSPECT-macos](https://github.com/flutter/devtools/assets/21270878/f48f57f3-770f-44f5-bf74-9197c84258b3)

Inspecting on chrome:

![INSPECT chrome](https://github.com/flutter/devtools/assets/21270878/ffe5b44d-afb6-43d1-b873-88894c46f6d7)

This means users are unable to use this feature, since we release DevTools as a web app.

elliette:
This is happening because we are using the Chrome's native context menus, so any of the Flutter context menus don't appear. 

Flutter recently added the ability to disable the Chrome's native context menu, but that currently prevents a user from pasting into a text field. See https://github.com/flutter/engine/pull/43360 (which should fix that bug) for details.

justinmc:
The paste button bug will be fixed with this framework PR: https://github.com/flutter/flutter/pull/132093.  Sorry for blocking for you guys! 😊

Issue #5963, opened 2023-06-27
Console UX feedback 


elliette:
Feedback from a customer team, one of their biggest pain points was interacting with the console:

- all the text is the same color
- horizontal line dividing text lines is a pain
- cannot select multiple lines of text

PR #6001, opened 2023-07-07
Remove tooltip during variable inspection 

elliette:
Fixes https://github.com/flutter/devtools/issues/6000

The tooltips were simply restating visible text, and therefore were not conforming to Material specs.

PR #6020, opened 2023-07-11
Debugger panel text selection improvements

elliette:
Fixes https://github.com/flutter/devtools/issues/6019
Work towards https://github.com/flutter/devtools/issues/5963

Note: Includes TODOS for https://github.com/flutter/devtools/issues/6018, which requires https://github.com/flutter/flutter/issues/129692 to be fixed first

**Demos (after):**

NATIVE - codeview 2

NATIVE - vars

NATIVE - console

**Demos (before):**

selecting codeview before

selecting inspect before

selecting before console


packages/devtools_app/lib/src/app.dart
```diff
@@ -105,6 +106,14 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
   void initState() {
     super.initState();
 
+    // TODO(https://github.com/flutter/devtools/issues/6018): Once
+    // https://github.com/flutter/flutter/issues/129692 is fixed, disable the
+    // browser's native context menu on secondary-click, and instead use the
+    // menu provided by Flutter:
+    if (kIsWeb) {
+      unawaited(BrowserContextMenu.enableContextMenu());
```
kenzieschmoll, line 115: do we really want to do this at the app-wide level? Will this break any flutter context menu actions we use elsewhere? Perhaps we could enable this menu more local to the console / debugger.

elliette: I'm pretty sure this is a no-op (the context menu is already globally enabled) - what we really want to do is disable it, but currently are blocked by https://github.com/flutter/flutter/issues/129692). I can switch this to disabling and just comment it out.

packages/devtools_app/lib/src/shared/console/widgets/display_provider.dart
```diff
@@ -75,10 +53,10 @@ class DisplayProvider extends StatelessWidget {
     }
 
     final hasName = variable.name?.isNotEmpty ?? false;
-    return DevToolsTooltip(
```
kenzieschmoll, line 78: is there a reason we removed the tooltip here?

packages/devtools_app/lib/src/shared/console/widgets/display_provider.dart
```diff
@@ -100,22 +78,47 @@ class DisplayProvider extends StatelessWidget {
...
+        ContextMenuButtonItem(
+          onPressed: () {
+            ContextMenuController.removeAny();
+            _handleInspect(context);
+          },
+          label: 'Inspect',
+        ),
+      );
+    }
+    return menuButtons;
```
kenzieschmoll, line 116: this can be condensed:
```dart
return [
  if (variable.isRerootable)
    // reroot button
  if (serviceManager.inspectorService != null && variable.isRoot)
    // inspect button
];
```

packages/devtools_app/lib/src/shared/console/widgets/display_provider.dart
elliette: Because this was initially branched off of https://github.com/flutter/devtools/pull/6001 (which hasn't been submitted yet)

packages/devtools_app/lib/src/app.dart
kenzieschmoll: What about disabling the context menu closer to the source of where we have actually implemented our own context menus? Like in this example how the browser context menu is disabled and reenabled within the lifecycle of this widget: https://github.com/flutter/flutter/blob/e12167d98c44ad1977934399757a4351821f4f95/examples/api/lib/material/context_menu/context_menu_controller.0.dart#L34-L50

elliette: I considered doing that, though I'm not sure where the best place would be. Would we want it on the entire debugger panel (and if so, would it be weird that that one panel has different context menus then the rest of DevTools)? Or on each selectable text region (which could get us into a weird state, eg context menu is disabled when codeview inits its state, then when the variable tree inits its state, then the variable tree is removed and the context menu is enabled, even though we still want it for codeview)?

I think we probably want the Flutter context menus everywhere. This matches the behavior for macos, and so ideally means the experience during development matches the experience in release-mode.

kenzieschmoll (approved):

Issue #6021, opened 2023-07-11
provider_integration_test flaking after latest widget inspector changes in flutter/flutter


kenzieschmoll:
Example failure: https://github.com/flutter/devtools/actions/runs/5525147736/jobs/10078437415?pr=6015
```
package:devtools_app/src/shared/eval_on_dart_library.dart 458:7  EvalOnDartLibrary.asyncEval
  Evaluation `() {  final result = widgetInspectorService.toObject("inspector-0", "asyncEval-2") as List;  widgetInspectorService.disposeGroup("asyncEval-2");  return result;}()` failed with [ErrorRef id: objects/67, kind: UnhandledException, message: Unhandled exception:
  type 'Null' is not a subtype of type 'List<dynamic>' in type cast
  #0      Eval.<anonymous closure> ()
  #1      Eval ()
  ]
  package:devtools_app/src/shared/eval_on_dart_library.dart 500:11  EvalOnDartLibrary.safeEval
```

My hypothesis is that this is caused by this commit to flutter/flutter: https://github.com/flutter/flutter/commit/bff6b93683de8be01d53a39b6183f230518541ac

kenzieschmoll:
Upgrading this to P0 as this issue is breaking our build at master.

polina-c:
Test that flakes:

```
      test(
        'listens to updates from the application side',
        () async {
          final container = ProviderContainer();
          addTearDown(container.dispose);

          final countSub = container.listen(
            instanceProvider(
              const InstancePath.fromProviderId(
                '0',
                pathToProperty: [
                  PathToProperty.objectProperty(
                    name: '_count',
                    ownerUri: 'package:provider_app/main.dart',
                    ownerName: 'Counter',
                  ),
                ],
              ),
            ).future,
            (prev, next) {},
          );

          await expectLater(
            countSub.read(),
            completion(
              isA<NumInstance>()
                  .having((e) => e.displayString, 'displayString', '0'),
            ),
          );

          await evalOnDartLibrary.asyncEval(
            'await tester.tap(find.byKey(Key("increment"))).then((_) => tester.pump())',
            isAlive: isAlive,
          );

          await expectLater(
            countSub.read(),
            completion(
              isA<NumInstance>()
                  .having((e) => e.displayString, 'displayString', '1'),
            ),
          );
        },
        timeout: const Timeout.factor(12),
      );
```

polina-c:
PR to remove the test: https://github.com/flutter/devtools/pull/6027

[...]

PR #6027, opened 2023-07-12
Remove one test from provider_screen_integration_test.dart

polina-c:
The test became flaky and needs to be recreated, may be as real integration test.

Issue: https://github.com/flutter/devtools/issues/6021

### DIFF
--- a/packages/devtools_app/test/provider/provider_screen_integration_test.dart
+++ b/packages/devtools_app/test/provider/provider_screen_integration_test.dart
@@ -782,6 +782,52 @@ void main() {
         },
         timeout: const Timeout.factor(12),
       );
+
+      test(
+        'listens to updates from the application side',
+        () async {
+          final container = ProviderContainer();
+          addTearDown(container.dispose);
+
+          final countSub = container.listen(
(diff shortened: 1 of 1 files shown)


polina-c:
cc @rrousselGit

kenzieschmoll (commented):
instead of deleting it, just skip it and link to the issue so that we do not lose context on the test that needs to be recreated.

packages/devtools_app/test/provider/provider_screen_integration_test.dart
```diff
@@ -122,714 +122,721 @@ void main() {
...
+                    .having((e) => e.displayString, 'displayString', '1'),
+              ),
+            );
+          },
+          timeout: const Timeout.factor(12),
+        );
+      });
+    },
+    skip:
+        true, //TODO(rrousselGit): https://github.com/flutter/devtools/issues/6021
```
kenzieschmoll, line 838: please add only to the test that is flaking - not to the whole group

kenzieschmoll (approved):

Issue #6024, opened 2023-07-12
Store a text file to track the flutter candidate that a build was tested against


kenzieschmoll:
Sometimes when debugging an issue, we have a need to bisect to find the commit that first introduced the regression. An issue that comes up with this is that an older version of DevTools may not be compatible with the latest version of Flutter that master is currently built from. Right now there is not an easy way to figure out which flutter version a DevTools commit was tied to without looking at the CI job logs (which may or may not have expired by this time).

One idea is to have the DartDevToolWorkflowBot automatically update some file `flutter_candidate_version.txt` with the latest candidate version number when it changes. The bot can run daily, and only create a PR if the latest flutter candidate version has changed. This PR should be able to be auto merged just like the daily dev bump PRs.

CC @CoderDake @bkonyi

PR #6034, opened 2023-07-14
workflow for updating flutter candidate

CoderDake:
This PR creates a workflow that will run daily to update the flutter-candidate.txt file. The objective of this is to have a daily resolution for flutter candidate versions, so in the future we know what version to use with legacy devtools versions.

Example of what the PR will look like.
https://github.com/flutter/devtools/pull/6037

Fixes https://github.com/flutter/devtools/issues/6024


.github/workflows/flutter-candidate-update.yaml
```diff
@@ -0,0 +1,68 @@
...
+permissions:
+  contents: write
+  pull-requests: write
+
+env:
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  update-candidate:
+    name: Bump Version
```
kenzieschmoll, line 17: Can you name this 'Update Flutter Candidate Version' or similar?

tool/README.md
```diff
@@ -31,3 +31,245 @@ Debug logs found in `Settings > Copy Logs` are saved such that they can be read
 
 For a quick tutorial on how to navigate logs using `lnav`
 you can give [ their tutorial ](https://lnav.org/tutorials) a try.
+
+## Flutter versions that work with this instance of DevTools
```
kenzieschmoll, line 35: I would change this to
## Finding the Flutter version that a DevTools commit is compatible with

Because this readme update won't even be present if a user is looking at an old version of DevTools, so they should be reading these instructions on the master branch.

kenzieschmoll (approved):
a couple nits but lgtm.

.github/workflows/flutter-candidate-update.yaml
CoderDake: oops ty!

PR #6039, opened 2023-07-14
Add support for new timeline event names `LAYOUT (root)` and `PAINT (root)`

kenzieschmoll:
These event names are changing with multi-view support and we need to support frame phase detection for the new names (and old ones for a while).

@goderbauer


goderbauer (approved):
LGTM

Issue #1632, opened 2020-02-14
Provide an API to create extensions for the devtools


simc:
It would be very helpful if there was a way to create custom extensions for the devtool. For example, I'd like to write an extension to inspect the contents of the [Hive DB](/hivedb/hive). But I think there are many interesting extensions, the community could create.

The Stetho tool from Facebook allows [something similar](https://github.com/facebook/stetho#custom-dumpapp-plugins) for the Chrome DevTools.

And the flipper tool also has a very powerful way to [create extensions](https://fbflipper.com/docs/tutorial/intro.html). It even allows creating [custom UI](https://fbflipper.com/docs/tutorial/js-custom.html).

britannio:
Apologies for the useless comment but this is my no. 1 request.
Few use cases
- Monitoring network activity for a network package e.g. dio or graphql
- Tracking state changes in Redux
- Creating custom dashboards with key metrics for your app
- Inspecting databases as @leisim suggested
- ...

cedvdb:
The use cases are legion tbh, any kind of inspection of state is useful there. Be it db state, application state, ui state, cache state, whatever state. Those kind of things are impractical to print inside the console because:

 - It's usually pretty big and would be unreadable
 - It usually involve a tree structure 

I would also like the possibility to have extension, but if that's not planned if there was a message bus mechanism where one could start another app that connects to that bus and display the required information with a desktop app it would be nice.

Let me rephrase this, if extensions is not planned, it would be nice to have the ability to:
  - start a flutter app that emits messages.
  - start another app that displays those messages.

Maybe this is doable today but my guess is that the process will be too involved (development wise) to make it practical.

leisim:
It would even be helpful to allow third parties to access an active debugging session without requiring the user to find the websocket link in the logs.

Use cases:
- [Isar Inspector](https://github.com/isar/isar#isar-inspector)
- Custom integration test VSCode extension

leisim:
@kenzieschmoll Is there something we could help with to make this happen? Or are there alternatives? Riverpod and Provider are integrated into devtools. Can anyone submit an extension?

jacob314:
Here's a link to our in progress design on supporting Flutter DevTools Plugins.  Comments appreciated! https://docs.google.com/document/d/1BWX8YQ962Vsx-EUDuDHRG7RX94fJlZSDRu299YRwasE/edit

leisim:
@jacob314 Wow that is incredible 😍 can't wait to build some plugins

[...]

PR #6040, opened 2023-07-14
Refactor extensions code to prepare for devtools extensions

kenzieschmoll:
This refactors the code to prepare a location for DevTools extensions work.

https://github.com/flutter/devtools/issues/1632


packages/devtools_app/lib/src/extensions/static/README.md
```diff
@@ -0,0 +1,5 @@
+## What are static extensions?
```
polina-c, line 1: What is the point of static extension? Why it is not just part of application?

kenzieschmoll: Good idea. Done.

PR #6047, opened 2023-07-17
Load snapshot files.

polina-c:


polina-c:
Still want to create it. Busy on other tasks.

PR #6055, opened 2023-07-18
Initial commit for `package:devtools_extensions`

kenzieschmoll:
This PR adds the `devtools_extensions` package to `devtools/packages/`. The initial commit for this package includes some shared data models that will be used by both DevTools and extension authors.

work towards https://github.com/flutter/devtools/issues/1632.


packages/devtools_extensions/lib/src/api.dart
```diff
@@ -0,0 +1,18 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+enum DevToolsExtensionEventType {
```
polina-c, line 5: Can you add doc comments to the enum and each member?

packages/devtools_extensions/lib/src/model.dart
```diff
@@ -0,0 +1,50 @@
...
+  static const _typeKey = 'type';
+  static const _dataKey = 'data';
+
+  static DevToolsExtensionEvent ping =
+      DevToolsExtensionEvent(DevToolsExtensionEventType.ping);
+
+  static DevToolsExtensionEvent pong =
+      DevToolsExtensionEvent(DevToolsExtensionEventType.pong);
+
+  final DevToolsExtensionEventType type;
```
polina-c, line 37: What's the point to differentiate types of extensions if they behave equally?

packages/devtools_extensions/lib/src/model.dart
```diff
@@ -0,0 +1,50 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'api.dart';
+
+/// Data model for a devtools extension event that will be sent and received
+/// over 'postMessage' between DevTools and the embedded extension iFrame.
+class DevToolsExtensionEvent {
```
polina-c, line 9: Can you explain in doc comment why the events are needed, and how they are supposed to be used?

packages/devtools_extensions/lib/src/api.dart
```diff
@@ -0,0 +1,18 @@
...
+  ping,
```
polina-c, line 6: Are ping and pong the same as request and response?
Request and response seem to be more traditional.

kenzieschmoll: Added documentation. Ping and Pong are to verify the initial connection. This is what is done for posting messages to perfetto as well.

packages/devtools_extensions/lib/src/model.dart
kenzieschmoll: Each extension will share some underlying framework for integration with DevTools (connection to DevTools, connection to a VM service, etc.), but the functionality that each extension provides will be entirely dependent on the use case a package or extension author addresses.

packages/devtools_extensions/lib/src/model.dart
kenzieschmoll: Added a reference to [DevToolsExtensionEventType]. This code is incomplete right now since we are starting with the `ping`, `pong` support, but in the future we will also support events like `connectToVmService`, maybe `showNotification` etc. When we support more events, the `DevToolsExtensionEventType` enum will be self documenting.

polina-c (approved):

packages/devtools_extensions/devtools_extensions.iml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bkonyi, line 1: Does this file need to be included?

packages/devtools_extensions/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
...
+version: 0.0.1
+homepage:
+
+environment:
+  sdk: '>=3.0.0'
+  flutter: '>=3.0.0'
+
+dependencies:
+  flutter:
+    sdk: flutter
```
bkonyi, line 12: I'm assuming we plan on exposing widgets / `ValueNotifier`s through this package?

bkonyi (approved):

packages/devtools_extensions/pubspec.yaml
kenzieschmoll: Yeah, this package will also include a template for extension authors to place at the root of their devtools extension app

packages/devtools_extensions/devtools_extensions.iml
kenzieschmoll: we include the .iml file for the other packages `devtools_app`, `devtools_shared`, `devtools_test`, so following that pattern and including this one as well.

PR #6060, opened 2023-07-19
Add embedded iFrame logic for DevTools extensions

kenzieschmoll:
https://github.com/flutter/devtools/assets/43759233/9d455aa0-4d76-4c09-bc59-5b9f00f449a3

Work towards https://github.com/flutter/devtools/issues/1632.

### DIFF
--- a/packages/devtools_app/lib/src/extensions/embedded/_controller_web.dart
+++ b/packages/devtools_app/lib/src/extensions/embedded/_controller_web.dart
@@ -2,8 +2,92 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+// ignore_for_file: avoid_web_libraries_in_flutter, as designed
+import 'dart:async';
+import 'dart:html' as html;
+import 'dart:ui' as ui;
+
+import 'package:devtools_extensions/devtools_extensions.dart';
+
 import 'controller.dart';
 
+/// Incrementer for the extension iFrame view that will live for the entire
+/// DevTools lifecycle.
+///
+/// Each time [EmbeddedExtensionControllerImpl.init] is called, we create a new
+/// [html.IFrameElement] and register it to
+/// [EmbeddedExtensionControllerImpl.viewId] via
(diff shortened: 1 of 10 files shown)


packages/devtools_app/lib/src/extensions/embedded/_controller_web.dart
```diff
@@ -2,8 +2,92 @@
...
 class EmbeddedExtensionControllerImpl extends EmbeddedExtensionController {
-  // TODO(kenz): implement web controller for embedded extension
+  EmbeddedExtensionControllerImpl(super.extensionConfig);
+
+  /// The view id for the extension iFrame.
+  ///
+  /// See [_viewIdIncrementer] for an explanation of why we use an incrementer
+  /// in the id.
+  late final viewId =
+      'devtools-extension-${extensionConfig.name}-${_viewIdIncrementer++}';
```
polina-c, line 33: Are we going to send this view id in messages? If yes, how about decreasing size of the packages by shortening to something like

'ext-${extensionConfig.name}-${_viewIdIncrementer++}';

packages/devtools_app/lib/src/extensions/embedded/_view_web.dart
```diff
@@ -2,19 +2,150 @@
...
+          default:
+            notificationService.push(
+              'Unknown event received from extension: ${e.data}',
+            );
+        }
+      }
+    }
+  }
+
+  Future<void> _pingExtensionUntilReady() async {
```
polina-c, line 127: Do we want time out and give user friendly message if extension is never ready?

kenzieschmoll: added a notification to communicate to the user that something went wrong because the extension is not ready. I also added a TODO to consider giving the user a way to retry or file an issue against the extension package.

packages/devtools_app/lib/src/extensions/embedded/_controller_web.dart
kenzieschmoll: we will not see this view id in messages, but shortened nonetheless.

polina-c (approved):

PR #6066, opened 2023-07-20
Add `DevToolsExtension` template to `devtools_extension` package

kenzieschmoll:
Work towards https://github.com/flutter/devtools/issues/1632.


packages/devtools_extensions/lib/src/template/README.md
polina-c, line 1: How about moving template out of `lib` to `example` and make it separate package with its own pubspec.yaml?
You will have to create `example` anyway, because it will be requested by pub.get publishing code.

kenzieschmoll: the template isn't an example - it is actually the file an extension author would import to write a devtools extension.

```
import 'package:devtools_extensions/api.dart';

import 'package:devtools_extensions/template.dart';

void main() {
  runApp(const FooPackageDevToolsPlugin());
}

class FooPackageDevToolsPlugin extends StatelessWidget {
  const FooPackageDevToolsPlugin({super.key});

  @override
    return MaterialApp(
      title: 'Foo DevTools Extension',
      home: const DevToolsExtension(
        child: FooExtensionHomePage(),
      ),
    );
  }
}
```

kenzieschmoll: open to renaming the file from `template.dart` to `devtools_extension.dart` or `extension.dart` if that would make things clearer

kenzieschmoll: added an `example` dir in https://github.com/flutter/devtools/pull/6069

packages/devtools_extensions/lib/src/template/devtools_extension.dart
```diff
@@ -0,0 +1,64 @@
...
+import 'package:logging/logging.dart';
+
+import '../../api.dart';
+
+part 'extension_manager.dart';
+
+/// A manager that allows extensions to interact with DevTools or the DevTools
+/// extensions framework.
+///
+/// A couple use case examples include posting messages to DevTools or
```
CoderDake, line 18:
```suggestion
/// A couple of use case examples include posting messages to DevTools or
```

packages/devtools_extensions/lib/src/template/extension_manager.dart
```diff
@@ -0,0 +1,59 @@
...
+              DevToolsExtensionEvent.pong.toJson(),
+              e.origin,
+            );
+            break;
+          case DevToolsExtensionEventType.pong:
+            // Ignore. DevTools extensions should not receive or handle pong
+            // events.
+            break;
+          case DevToolsExtensionEventType.unknown:
+            _log.info('Unrecognized event received by extension: ${e.data}');
```
CoderDake, line 47: might be worth making this more severe as we would only expect this to happen during dev when callers would want to know asap that they've sent an incompatible event.

CoderDake (approved):

packages/devtools_extensions/lib/src/template/extension_manager.dart
kenzieschmoll: changed to warning and logged message for the default case as well

polina-c (approved):

Issue #6056, opened 2023-07-18
Migrate debugger panel and eval to use DAP instead of the VM Service


elliette:
We would like to reduce the work required to support new language features by migrating part of Dart DevTools over to using [DAP.](https://microsoft.github.io/debug-adapter-protocol/) 

DAP is the abstract protocol designed by Microsoft to be used between a debugging tool (e.g., VS Code, Dart DevTools) and a debugger (e.g., VM Service). 

With Dart DevTools using the DAP interface to represent variables, we will only need to implement features in the Dart DAP package, instead of in both the Dart DAP package and in Dart DevTools.

![stop-continue-terminate](https://github.com/flutter/devtools/assets/21270878/6f30cb31-9da9-4fac-b50a-4591b4534b26)

elliette:
### General:

- [ ] Create feature flag for DAP
- [ ] Disable feature flag for DAP 

### Variables support:

- [ ] https://github.com/dart-lang/sdk/issues/52991
- [ ] Support variable types and variable paging (requires DevTools sending `Initialization` request)
- [ ] Support inspect and re-root functionality for variables 

### Evaluation support:

- [ ] Move eval service logic from DevTools to DAP, including implementing the DAP Completions [request](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Completions)
- [ ] TODO: Add more subtasks during development

### Frames support:

- [ ] TODO: Add more subtasks during development

PR #6070, opened 2023-07-21
Migrate `Variables` to use DAP

elliette:
Note: hidden behind feature flag 

Work towards https://github.com/flutter/devtools/issues/6056

dap_variables


packages/devtools_app/test/debugger/debugger_screen_dap_variables_test.dart
```diff
@@ -0,0 +1,147 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
```
jacob314, line 1: nit: year is 2023.

packages/devtools_app/lib/src/shared/tree.dart
```diff
@@ -35,7 +35,7 @@ class TreeView<T extends TreeNode<T>> extends StatefulWidget {
   /// Invoked when a tree node is selected. If [onItemExpanded] is not
   /// provided, this method will also be called when the expand button is
   /// tapped.
-  final FutureOr<void> Function(T) onItemSelected;
+  final FutureOr<void> Function(T)? onItemSelected;
```
jacob314, line 38: nit: instead of making this nullable why not just provide a NOOP implementation of this for now for your use case. I imagine you will want to provide a onItemSelected callback at some point.

packages/devtools_app/test/debugger/debugger_screen_dap_variables_test.dart
```diff
@@ -0,0 +1,147 @@
...
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mockito/mockito.dart';
+
+void main() {
+  late FakeServiceManager fakeServiceManager;
+  late MockDebuggerController debuggerController;
+  late MockScriptManager scriptManager;
+  late MockVmServiceWrapper vmService;
+
+  const windowSize = Size(4000, 4000);
```
jacob314, line 20: nit: prefer a typical rather than enormous window size where possible.

packages/devtools_app/lib/src/shared/tree.dart
elliette: I made this `nullable` because I am using `onItemExpanded` instead. The doc for this method says: "If [onItemExpanded] is not provided, this method will also be called when the expand button is tapped" which makes it sound like it should be nullable. I also see we are passing it as a no-op in other parts of DevTools ( I can remove that if we want to keep it nullable): 
 https://github.com/flutter/devtools/blob/ec967fc55ff06a66095117214d2980166bd44bd4/packages/devtools_app/lib/src/screens/vm_developer/object_inspector/class_hierarchy_explorer.dart#L39

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -1092,6 +1096,77 @@ class VmServiceWrapper implements VmService {
         parser: ObjectStore.parse,
       );
 
+  Future<dap.VariablesResponseBody?> dapVariablesRequest(
+    dap.VariablesArguments args,
+  ) async {
+    final response = await _sendDapRequest(
+      dap.Request(
+        command: 'variables',
+        seq: 0,
```
DanTup, line 1105: We don't currently rely on it in the adapter, but `seq` is specified as an increasing/unique number so I'd make a field for this and use `_seq++` in these calls.

(maybe `_sendDapRequest` could just take the `command` and `args` and automatically unwrap the response to avoid some boilerplate in each of these?)

packages/devtools_app/lib/src/shared/tree.dart
```diff
@@ -117,7 +117,9 @@ class _TreeViewState<T extends TreeNode<T>> extends State<TreeView<T>>
     if (widget.onItemExpanded == null && item.isExpandable) {
       item.toggleExpansion();
     }
-    await widget.onItemSelected(item);
+    if (widget.onItemSelected != null) {
+      await widget.onItemSelected!(item);
+    }
```
DanTup, line 122: I think you could avoid the `if`s and `!`s by doing something like `await widget.onItemSelected?.call(item);`

packages/devtools_app/pubspec.yaml
```diff
@@ -23,10 +23,11 @@ dependencies:
   async: ^2.0.0
   codicon: ^1.0.0
   collection: ^1.15.0
-  dds_service_extensions: ^1.3.2
-  devtools_shared: 2.26.0-dev.6
+  dap: ^1.0.0
```
DanTup, line 26: I'd make this `dap: 1.1.0`. There's a new `lazy` field on `VariablePresentationHint ` that may be useful :-)

DanTup (commented):
Just a few drive-by comments :-)

packages/devtools_app/test/debugger/debugger_screen_dap_variables_test.dart
elliette: Oops, updated!

packages/devtools_app/lib/src/shared/tree.dart
elliette: Switched to removing the places where we are passing this as a no-op

packages/devtools_app/test/debugger/debugger_screen_dap_variables_test.dart
elliette: Done!

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
elliette: Ah good to know! Switched to incrementing it, and moving some of the boilerplate (plus adding error handling) to `_sendDapRequest` method

packages/devtools_app/lib/src/shared/tree.dart
elliette: Oh nice, done!

packages/devtools_app/pubspec.yaml
elliette: Thanks! Done

packages/devtools_app/lib/src/shared/console/widgets/display_provider.dart
```diff
@@ -175,6 +176,50 @@ class DisplayProvider extends StatelessWidget {
   }
 }
 
+class DapDisplayProvider extends StatelessWidget {
```
bkonyi, line 179: This could use some documentation.

bkonyi (approved):
LGTM!

PR #6094, opened 2023-07-25
Add a server endpoint for serving devtools extensions

kenzieschmoll:
This PR
- moves `extension_model.dart` to `packages/devtools_shared` so that we can reuse this model from the devtools server
- creates `ExtensionManager` that manages moving devtools extension assets to `build/devtools_extensions` so that the server can serve them (implemented in https://dart-review.googlesource.com/c/sdk/+/305043)
- adds a server endpoint to serve devtools extensions

work towards https://github.com/flutter/devtools/issues/1632


packages/devtools_app/lib/src/extensions/extension_screen.dart
```diff
@@ -155,3 +155,10 @@ class EmbeddedExtensionHeader extends StatelessWidget {
     );
   }
 }
+
+extension ExtensionConfigExtension on DevToolsExtensionConfig {
+  IconData get icon => IconData(
```
bkonyi, line 160: Why does this need to be an extension instead of being a property of `DevToolsExtensionConfig`?

packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -0,0 +1,168 @@
...
+
+/// The location for the DevTools extension, relative to the parent package's
+/// root.
+const extensionLocation = 'extension/devtools';
+
+/// The default location for the DevTools extension, relative to
+/// `<parent_package_root>/extension/devtools/`.
+const extensionBuildDefault = 'build';
+
+class ExtensionsManager {
```
bkonyi, line 24: What does this class do? We should have some documentation here.

packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -0,0 +1,168 @@
...
+  ExtensionsManager({required this.buildDir});
+
+  final String buildDir;
+
+  String get servedExtensionsPath => path.join(buildDir, extensionRequestPath);
+
+  final devtoolsExtensions = <DevToolsExtensionConfig>[];
+
+  void serveAvailableExtensions(String? rootPath) {
```
bkonyi, line 33: This needs documentation

packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -0,0 +1,168 @@
...
+    String extensionPackageName,
+    String extensionPath,
+  ) {
+    final newExtensionPath =
+        path.join(servedExtensionsPath, extensionPackageName);
+    copyPathSync(extensionPath, newExtensionPath);
+  }
+}
+
+/// Mark copied from package:extension_discovery. Remove once it is published.
```
bkonyi, line 113: Is this comment related to `Extension`? Maybe this should be a TODO?

packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -0,0 +1,168 @@
...
+
+/// Information about an extension for target package.
+class Extension {
+  Extension._({
```
bkonyi, line 117: Does anything create an instance of `Extension` yet? I don't see any references in this library and it doesn't have a public constructor.

packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -0,0 +1,168 @@
...
+    required this.config,
+  });
+
+  /// Name of the package providing an extension.
+  final String package;
+
+  /// Absolute path to the package root.
+  ///
+  /// This folder usually contains a `lib/` folder and a `pubspec.yaml`
+  /// (assuming dependencies are fetched using the pub package manager).
```
bkonyi, line 130: We make references to fetching using `dart pub` below, but use "fetched using the pub package manager" verbiage here. We should probably pick one or the other (FWIW, I prefer just saying "dart pub").

packages/devtools_shared/lib/src/extensions/extension_model.dart
```diff
@@ -0,0 +1,74 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
```
bkonyi, line 1: Just confirming, this is a straight move from `package:devtools_app/src/extensions/embedded/extension_model.dart` right?

bkonyi (commented):
First pass.

packages/devtools_app/lib/src/shared/config_specific/server/_server_web.dart
```diff
@@ -337,6 +338,31 @@ DevToolsJsonFile _devToolsJsonFileFromResponse(
...
+    final uri = Uri(
+      path: apiServeAvailableExtensions,
+      queryParameters: {extensionRootPathPropertyName: rootPath},
+    );
+    final resp = await request(uri.toString());
+    if (resp?.status == HttpStatus.ok) {
+      final parsedResult = json.decode(resp!.responseText!);
+      final extensionsAsJson =
+          (parsedResult[extensionsResultPropertyName]! as List<Object?>)
+              .cast<Map<String, Object?>>();
```
bkonyi, line 354: There's a possibility this cast could fail. If the result is `List<Object?>`, casting to a non-nullable `List<Map<String, Object?>>` would throw if there is a null entry. It probably doesn't make much of a difference since we'd throw during the `as List<Object>` cast if there's a null entry in the list, but can we instead do `(result as List<Object>).cast<Map<String, Object?>>()` to make it clear that we don't expect a null entry?

packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -0,0 +1,168 @@
...
+const extensionLocation = 'extension/devtools';
+
+/// The default location for the DevTools extension, relative to
+/// `<parent_package_root>/extension/devtools/`.
+const extensionBuildDefault = 'build';
+
+class ExtensionsManager {
+  ExtensionsManager({required this.buildDir});
+
+  final String buildDir;
```
bkonyi, line 27: All public properties and methods on this class need documentation.

packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -0,0 +1,168 @@
...
+      // TODO(kenz): use 'findExtensions' from package:extension_discovery once it
+      // is published.
+      // final extensions = findExtensions(
+      //   'devtools',
+      //   packageConfig: '$rootPath/.dart_tool/package_config.json',
+      // );
+      final extensions = <Extension>[];
+      for (final extension in extensions) {
+        final config = extension.config;
+        if (config is! Map) {
```
bkonyi, line 46: If `config` is always expected to be `Map`, why don't we restrict it to `Map<Object, Object?>` in `Extension`?

packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -0,0 +1,168 @@
...
+    _resetServedPluginsDir();
+    for (final extension in devtoolsExtensions) {
+      _moveToServedExtensionsDir(extension.name, extension.path);
+    }
+  }
+
+  void _resetServedPluginsDir() {
+    final buildDirectory = Directory(buildDir);
+    if (!buildDirectory.existsSync()) {
+      throw Exception('The build directory does not exist.');
```
bkonyi, line 91: Can we use a more specific exception type here?

packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -0,0 +1,168 @@
...
+    servedExtensionsDir.createSync();
+  }
+
+  void _moveToServedExtensionsDir(
+    String extensionPackageName,
+    String extensionPath,
+  ) {
+    final newExtensionPath =
+        path.join(servedExtensionsPath, extensionPackageName);
+    copyPathSync(extensionPath, newExtensionPath);
```
bkonyi, line 109: In general, we should avoid using `fooSync()` APIs in production as they block the event loop for the duration of the call.

packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -0,0 +1,168 @@
...
+    }
+    servedExtensionsDir.createSync();
+  }
+
+  void _moveToServedExtensionsDir(
+    String extensionPackageName,
+    String extensionPath,
+  ) {
+    final newExtensionPath =
+        path.join(servedExtensionsPath, extensionPackageName);
```
bkonyi, line 108: Style nit: consider adding a trailing comma

bkonyi (commented):
Another pass.

packages/devtools_shared/lib/src/extensions/extension_manager.dart
kenzieschmoll: See discussion at https://github.com/dart-lang/tools/pull/129#discussion_r1269800754

packages/devtools_shared/lib/src/extensions/extension_manager.dart
kenzieschmoll: Yes everything below here will be published from package:extension_discovery, but I  copied the type here so I can implement `serveAvailableExtensions` using this type.

Initial PR for package:extension_discovery is here: https://github.com/dart-lang/tools/pull/129

packages/devtools_shared/lib/src/extensions/extension_manager.dart
kenzieschmoll: We use the type on line 43.

packages/devtools_shared/lib/src/extensions/extension_manager.dart
kenzieschmoll: Ack. But this code is copied from https://github.com/dart-lang/tools/pull/129, and it will eventually be removed from devtools_shared. It is only in this package temporarily. I will prune the comments and make it a private class to make that more clear.

packages/devtools_app/lib/src/extensions/extension_screen.dart
bkonyi: Makes sense, thanks for explaining!

packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -0,0 +1,129 @@
...
+    devtoolsExtensions.clear();
+
+    if (rootPath != null) {
+      // TODO(kenz): use 'findExtensions' from package:extension_discovery once it
+      // is published.
+      // final extensions = findExtensions(
+      //   'devtools',
+      //   packageConfig: '$rootPath/.dart_tool/package_config.json',
+      // );
+      final extensions = <Extension>[];
```
bkonyi, line 43: I think this is an error now that you've renamed `Extension` to `_Extension`.

packages/devtools_shared/lib/src/devtools_api.dart
```diff
@@ -64,3 +64,12 @@ const String apiGetTestAppSizeFile = '${apiPrefix}getTestAppSizeFile';
 const String baseAppSizeFilePropertyName = 'appSizeBase';
 
 const String testAppSizeFilePropertyName = 'appSizeTest';
+
+/// Serves any available extensions and returns a list of their configurations
+/// to DevTools.
+const String apiServeAvailableExtensions =
+    '${apiPrefix}serveAvailableExtensions';
+
+const extensionRootPathPropertyName = 'rootPath';
```
bkonyi, line 73: Nit: documentation for these two property strings. Maybe add types to be consistent with the rest of the constants here?

packages/devtools_shared/lib/src/extensions/extension_model.dart
kenzieschmoll: I added dart doc to the members and some validation in the factory constructor so it should be re-reviewed. I also made the json parsing for the icon code point able to handle both strings and ints

packages/devtools_app/lib/src/shared/config_specific/server/_server_web.dart
kenzieschmoll: Replacing `List<Object?>` with `List<Object>` resulted in this error:
```
TypeError: Instance of 'JSArray<dynamic>': type 'JSArray<dynamic>' is not a subtype of type 'List<Object>'
```

Instead, I added a `.whereNotNull()` call in the chain to ensure we aren't casting a potentially null element to a non null map.

packages/devtools_shared/lib/src/extensions/extension_manager.dart
kenzieschmoll: used FileSystemException

packages/devtools_app/lib/src/shared/config_specific/server/_server_web.dart
bkonyi: Huh, that's interesting... Using `.whereNotNull()` is a good call.

packages/devtools_shared/lib/src/extensions/extension_model.dart
```diff
@@ -0,0 +1,99 @@
...
+    final codePointFromJson = json[materialIconCodePointKey];
+    const defaultCodePoint = 0xf03f;
+    if (codePointFromJson is String?) {
+      codePoint =
+          int.tryParse(codePointFromJson ?? '0xf03f') ?? defaultCodePoint;
+    } else {
+      codePoint = codePointFromJson as int? ?? defaultCodePoint;
+    }
+
+    final name = json[nameKey] as String?;
```
bkonyi, line 28: We should use pattern matching here for JSON validation:

```dart
  if (json
      case {
        nameKey: String name,
        pathKey: String path,
        issueTrackerKey: String issueTracker,
        versionKey: String version,
      }) {
    return DevToolsExtensionConfig._(
      name: name,
      path: path,
      issueTrackerLink: issueTrackerLink,
      version: version,
      materialIconCodePoint: codePoint,
    );
  }
  throw StateError(
    'missing required fields ${nullFields.toString()} in the extension '
    'config.json',
  );
```

bkonyi (approved):
LGTM with one last suggestion.

packages/devtools_shared/lib/src/extensions/extension_model.dart
kenzieschmoll: I will address this comment as part of https://github.com/flutter/devtools/pull/6097.

PR #6097, opened 2023-07-25
Load extensions from DevTools server instead of using stubs.

kenzieschmoll:
https://github.com/flutter/devtools/pull/6094 needs to land first before this can land.


packages/devtools_app/lib/src/extensions/extension_service.dart
```diff
@@ -14,17 +15,36 @@ class ExtensionService extends DisposableController
       _availableExtensions;
   final _availableExtensions = ValueNotifier<List<DevToolsExtensionConfig>>([]);
 
-  void initialize() {
-    addAutoDisposeListener(serviceManager.connectedState, () {
-      _refreshAvailableExtensions();
+  Future<void> initialize() async {
+    await maybeRefreshExtensions();
+    addAutoDisposeListener(serviceManager.connectedState, () async {
```
bkonyi, line 20: Can this just be `autoDisposeListener(serviceManager.connectedState, maybeRefreshExtensions)`?

packages/devtools_app/lib/src/extensions/extension_service.dart
```diff
@@ -14,17 +15,36 @@ class ExtensionService extends DisposableController
...
   }
 
-  // TODO(kenz): actually look up the available extensions from devtools server,
-  // based on the root path(s) from the available isolate(s).
-  int _count = 0;
-  void _refreshAvailableExtensions() {
-    _availableExtensions.value =
-        debugExtensions.sublist(0, _count++ % (debugExtensions.length + 1));
+  Future<void> _refreshAvailableExtensions(String? rootPath) async {
+    final extensions = await server.refreshAvailableExtensions(rootPath);
```
bkonyi, line 37: Nit: Consider `_availableExtensions.value = await server.refreshAvailableExtensions(rootPath)` instead

packages/devtools_app/lib/src/extensions/extension_service.dart
```diff
@@ -14,17 +15,36 @@ class ExtensionService extends DisposableController
...
+    _availableExtensions.value = extensions;
+  }
+}
+
+Future<String?> _connectedAppRootPath() async {
+  var fileUri = await serviceManager.rootLibraryForSelectedIsolate();
+  if (fileUri == null) return null;
+
+  if (fileUri.endsWith('/lib/main.dart')) {
```
bkonyi, line 46: What happens if the root library isn't `main.dart`? There's no requirement for the entry point of a Dart or Flutter program to be in `main.dart`. I'm not even sure there's a requirement for it to be in a library of the form `lib/<script>` for Flutter, but there definitely isn't for Dart programs.

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -539,6 +539,23 @@ class ServiceConnectionManager {
     await whenValueNonNull(isolateManager.mainIsolate);
     return libraryUriAvailableNow(uri);
   }
+
+  Future<String?> rootLibraryForSelectedIsolate() async {
+    if (!connectedState.value.connected) return null;
+
+    final selectedIsolateRef = isolateManager.mainIsolate.value?.id;
```
bkonyi, line 546: I don't think this can be null if `connectedState.value.connected == true`.

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -539,6 +539,23 @@ class ServiceConnectionManager {
...
+    if (selectedIsolateRef == null) return null;
+
+    final selectedIsolate = await service!.getIsolate(selectedIsolateRef);
```
bkonyi, line 549: If the isolate is already set to be the main isolate, it should either be in the process of becoming runnable or already runnable, meaning `IsolateManager` should have the full `Isolate` object. Consider doing something along these lines instead:

```dart
final isolateState = isolateManager.isolateState(selectedIsolateRef);
await isolateState.waitForIsolateLoad();
final rootLib = isolateState.rootInfo!.library;
```

packages/devtools_app/lib/src/extensions/extension_service.dart
kenzieschmoll: Is there a convention for finding the root directory that contains the `.dart_tool` folder? Is this guaranteed to always be the parent of `lib/`

bkonyi: I'm not sure, especially if you don't have access to the FS. If you did, you'd just walk the directory chain until you found the directory containing `.dart_tool`.

I think you're probably safe to assume that the entrypoint is under `lib/`, so if you update this to `fileUri.endsWith(RegExp(r'\/lib\/[^\/.]*.dart'))` this should work the vast majority of the time.

kenzieschmoll: Optionally, I could just pass the whole root library to the server and have the server look for the .dart_tool folder?

bkonyi: That'd be the most accurate way of handling it.

packages/devtools_app/lib/src/service/service_manager.dart
kenzieschmoll: Ack. I traced the code back and technically `_computeMainIsolate` could return null in the isolate managers initialization. I'd rather fail gracefully here than blow up for some corner case.

packages/devtools_app/lib/src/extensions/extension_service.dart
kenzieschmoll: Used the regexp for now but added a todo to move this logic to the server.

packages/devtools_shared/lib/src/extensions/extension_model.dart
```diff
@@ -25,31 +25,27 @@ class DevToolsExtensionConfig {
...
+        }) {
+      return DevToolsExtensionConfig._(
+        name: name,
+        path: path,
+        issueTrackerLink: issueTracker,
+        version: version,
+        materialIconCodePoint: codePoint,
+      );
+    } else {
+      final requiredKeys = {nameKey, pathKey, issueTrackerKey, versionKey};
```
bkonyi, line 43: I think `requiredKeys` can be `const`.

packages/devtools_shared/lib/src/extensions/extension_model.dart
```diff
@@ -25,31 +25,27 @@ class DevToolsExtensionConfig {
...
       throw StateError(
-        'missing required fields ${nullFields.toString()} in the extension '
-        'config.json',
+        'Missing required fields ${requiredKeys.difference(json.keys.toSet())} '
```
bkonyi, line 45: More of an FYI than anything, this won't catch the case where a key is expected but of the wrong type. Maybe we should throw a different exception if `requiredKeys.differences(json.keys.toSet()) == const {}`?

bkonyi (approved):

packages/devtools_shared/lib/src/extensions/extension_model.dart
kenzieschmoll: addressed.

PR #6098, opened 2023-07-25
Add `--no-build-app` flag to the `build_e2e.dart` script.

kenzieschmoll:
When developing with the server, sometimes only the server needs to be reloaded and we do not need to pay the cost of rebuilding the DevTools flutter web app. Adding `--no-build-app` supports this workflow.


tool/build_e2e.dart
```diff
@@ -8,13 +8,18 @@ const argDevToolsBuild = 'devtools-build';
 const argUpdatePerfetto = '--update-perfetto';
 const argNoUpdateFlutter = '--no-update-flutter';
 
+// Add this flag if devtools_app does not need to be rebuilt.
+const argNoBuildApp = '--no-build-app';
+
 void main(List<String> args) async {
```
polina-c, line 14: Can you comment on `main` what items are built to make it clear what is remaining after argNoBuildApp is applied.

polina-c (approved):
Lgtm after addressing one comment

PR #6104, opened 2023-07-26
Add some simple side bar functionality with a mock editor environment

DanTup:
This is (another) evolution of #5747 / #5921.

I wanted to make it easy to work on this panel without needing to embed inside VS Code (not just because messing around with connecting WebSockets is clunky, but because it also means having Dart-Code set up for development). I tried hosting the panel inside an iframe (where the outer page could simulate the editor), but hosting Flutter apps inside an iframe seems to break the debugging flow (the inner app fails to start up with some odd javascript errors, but I think the debugger would fail to work with it even with that fixes).

So this version hosts everything in one page by just wrapping the `VsCodeFlutterPanel` with another widget (currently called `VsCodeFlutterPanelMock` but it's not a great name) which plays the part of the IDE. It sets up streams (to emulate postMessage, WebSockets, whatever) and logs the data going over it. It provides a mock implementation of the API that VS Code will provide and some buttons to trigger events that would originate in VS Code.

This should allow working on the panel and defining a concrete API (but adding it to the `DartApi` classes and a mock implementation), as well as making it possible to write tests, before implementing the same API in VS Code.

https://github.com/flutter/devtools/assets/1078012/2e94a9ae-1733-4c77-9ce2-417a1e69bb8b

@kenzieschmoll interested in your thoughts. Right now the bits that are there are very basic (for example the devices passed over only have IDs), but I think it'll be quicker to iterate like this without having to simultaneously provide the VS Code implementation.

### DIFF
--- a/packages/devtools_app/lib/src/standalone_ui/api/dart_tooling_api.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/api/dart_tooling_api.dart
@@ -4,6 +4,14 @@
 
 import 'vs_code_api.dart';
 
+/// An API exposed to Dart tooling extensions.
+///
+/// APIs are grouped into child APIs that are exposed as fields. Each field is a
+/// `Future` that will return null if the requested API is unavailable (for
+/// example the VS Code APIs if not running inside VS Code, or the LSP APIs if
+/// no LSP server is available).
 abstract interface class DartToolingApi {
+  /// Access to APIs provided by VS Code and/or the Dart/Flutter VS Code
+  /// extensions.
   Future<VsCodeApi?> get vsCode;
 }
--- a/packages/devtools_app/lib/src/standalone_ui/api/dart_tooling_api.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/api/dart_tooling_api.dart
@@ -4,7 +4,7 @@
 
 import 'vs_code_api.dart';
 
-/// An API exposed to Dart tooling extensions.
+/// An API exposed to Dart tooling surfaces.
 ///
 /// APIs are grouped into child APIs that are exposed as fields. Each field is a
 /// `Future` that will return null if the requested API is unavailable (for
--- a/packages/devtools_app/lib/src/standalone_ui/api/dart_tooling_api.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/api/dart_tooling_api.dart
@@ -1,0 +1,9 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'vs_code_api.dart';
+
+abstract interface class DartToolingApi {
+  Future<VsCodeApi?> get vsCode;
+}
--- a/packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
@@ -9,7 +9,6 @@ import 'package:json_rpc_2/json_rpc_2.dart' as json_rpc_2;
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 import 'package:stream_channel/stream_channel.dart';
-import 'package:web_socket_channel/web_socket_channel.dart';
 
 import '../../../shared/config_specific/logger/logger_helpers.dart';
 import '../../../shared/config_specific/post_message/post_message.dart';
@@ -29,7 +28,8 @@ const _enablePostMessageVerboseLogging = false;
 
 final _log = Logger('tooling_api');
 
-/// An API for interacting with Dart tooling.
+/// An API used by Dart tooling surfaces to interact with Dart tools that expose
+/// APIs such as Dart-Code and the LSP server.
 class DartToolingApiImpl implements DartToolingApi {
   DartToolingApiImpl.rpc(this._rpc) {
     unawaited(_rpc.listen());
@@ -54,11 +54,6 @@ class DartToolingApiImpl implements DartToolingApi {
       postMessageController,
     );
     return DartToolingApiImpl.rpc(json_rpc_2.Peer.withoutJson(channel));
-  }
-
-  /// Connects the API over the provided WebSocket.
-  factory DartToolingApiImpl.webSocket(WebSocketChannel socket) {
-    return DartToolingApiImpl.rpc(json_rpc_2.Peer(socket.cast<String>()));
   }
 
   final json_rpc_2.Peer _rpc;
@@ -108,6 +103,7 @@ abstract base class ToolApiImpl {
   @protected
   Stream<Map<String, Object?>> events(String name) {
     final streamController = StreamController<Map<String, Object?>>.broadcast();
+    unawaited(rpc.done.then((_) => streamController.close()));
     rpc.registerMethod('$apiName.$name', (json_rpc_2.Parameters parameters) {
       streamController.add(parameters.asMap.cast<String, Object?>());
     });
--- a/packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
@@ -1,0 +1,116 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:json_rpc_2/json_rpc_2.dart' as json_rpc_2;
+import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
+import 'package:stream_channel/stream_channel.dart';
+import 'package:web_socket_channel/web_socket_channel.dart';
+
+import '../../../shared/config_specific/logger/logger_helpers.dart';
+import '../../../shared/config_specific/post_message/post_message.dart';
+import '../../../shared/constants.dart';
+import '../dart_tooling_api.dart';
+import '../vs_code_api.dart';
+import 'vs_code_api.dart';
+
+final _log = Logger('tooling_api');
+
+/// Whether to enable verbose logging for postMessage communication.
+///
+/// This is useful for debugging when running inside VS Code.
+///
+/// TODO(dantup): Make a way for this to be enabled by users at runtime for
+///  troubleshooting. This could be via a message from VS Code, or something
+///  that passes a query param.
+const _enablePostMessageVerboseLogging = false;
+
+/// An API for interacting with Dart tooling.
+class DartToolingApiImpl implements DartToolingApi {
+  DartToolingApiImpl.rpc(this._rpc) {
+    unawaited(_rpc.listen());
+  }
+
+  /// Connects the API using 'postMessage'. This is only available when running
+  /// on web and embedded inside VS Code.
+  factory DartToolingApiImpl.postMessage() {
+    if (_enablePostMessageVerboseLogging) {
+      setDevToolsLoggingLevel(verboseLoggingLevel);
+    }
+    final postMessageController = StreamController();
+    postMessageController.stream.listen((message) {
+      _log.info('==> $message');
+      postMessage(message, '*');
+    });
+    final channel = StreamChannel(
+      onPostMessage.map((event) {
+        _log.info('<== ${jsonEncode(event.data)}');
+        return event.data;
+      }),
+      postMessageController,
+    );
+    return DartToolingApiImpl.rpc(json_rpc_2.Peer.withoutJson(channel));
+  }
+
+  /// Connects the API over the provided WebSocket.
+  factory DartToolingApiImpl.webSocket(WebSocketChannel socket) {
+    return DartToolingApiImpl.rpc(json_rpc_2.Peer(socket.cast<String>()));
+  }
+
+  final json_rpc_2.Peer _rpc;
+
+  /// An API that provides Access to APIs related to VS Code, such as executing
+  /// VS Code commands or interacting with the Dart/Flutter extensions.
+  ///
+  /// Lazy-initialized and completes with `null` if VS Code is not available.
+  @override
+  late final Future<VsCodeApi?> vsCode = VsCodeApiImpl.tryConnect(_rpc);
+
+  void dispose() {
+    unawaited(_rpc.close());
+  }
+}
+
+/// Base class for the different APIs that may be available.
+abstract base class ToolApiImpl {
+  ToolApiImpl(this.rpc);
+
+  static Future<Map<String, Object?>?> tryGetCapabilities(
+    json_rpc_2.Peer rpc,
+    String apiName,
+  ) async {
+    try {
+      final response = await rpc.sendRequest('$apiName.getCapabilities')
+          as Map<Object?, Object?>;
+      return response.cast<String, Object?>();
+    } catch (_) {
+      // Any error initializing should disable this functionality.
+      return null;
+    }
+  }
+
+  @protected
+  final json_rpc_2.Peer rpc;
+
+  @protected
+  String get apiName;
+
+  @protected
+  Future<T> sendRequest<T>(String method, [Object? parameters]) async {
+    return (await rpc.sendRequest('$apiName.$method', parameters)) as T;
+  }
+
+  /// Listens for an event '[apiName].[name]' that has a Map for parameters.
+  @protected
+  Stream<Map<String, Object?>> events(String name) {
+    final streamController = StreamController<Map<String, Object?>>.broadcast();
+    rpc.registerMethod('$apiName.$name', (json_rpc_2.Parameters parameters) {
+      streamController.add(parameters.asMap.cast<String, Object?>());
+    });
+    return streamController.stream;
+  }
+}
--- a/packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
@@ -18,8 +18,6 @@ import '../dart_tooling_api.dart';
 import '../vs_code_api.dart';
 import 'vs_code_api.dart';
 
-final _log = Logger('tooling_api');
-
 /// Whether to enable verbose logging for postMessage communication.
 ///
 /// This is useful for debugging when running inside VS Code.
@@ -29,6 +27,8 @@ final _log = Logger('tooling_api');
 ///  that passes a query param.
 const _enablePostMessageVerboseLogging = false;
 
+final _log = Logger('tooling_api');
+
 /// An API for interacting with Dart tooling.
 class DartToolingApiImpl implements DartToolingApi {
   DartToolingApiImpl.rpc(this._rpc) {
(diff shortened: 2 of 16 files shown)


DanTup:
@kenzieschmoll I've pushed some changes here. It's not currently visually appealing, but the basic framework is there. I have a branch of Dart-Code that costs the sidebar and provides the same functionality that the mock environment does here.

This currently uses postMessage inside VS Code and just simple streams in the mock environment - both are just JSON-RPC so I believe should be an easy migration to support DTD in the future.

APIs are implemented in a way that is conditional (so the panel can check which APIs are available) and supports capabilities (so if a whole API is available, we can still check its capabilities to help with the API changing over time). I separated the APIs into an interface and an implementation so that breaking changes should be more obvious and it's clear exactly what needs to be mirrored between here and the other side.

It'll need tests + making look nicer, but I'd like to get some feedback on the general approach first. You can run this using the new launch configuration here (which will load the mock environment, showing the embedded panel and some buttons/log stream):

image

kenzieschmoll (commented):
can you mark which parts of this prototype would become obsolete once we have a Dart Tooling Daemon we can use to pass messages back and forth?

DanTup:
> can you mark which parts of this prototype would become obsolete once we have a Dart Tooling Daemon we can use to pass messages back and forth?

I don't actually think much of this will go away, but rather just be changed slightly. For example the `DartToolingApiImpl.postMessage()` constructor would likely be replaced with one that connects to the DTD instead, and the helpers for sending requests/listening to streams (`DartToolingApiImpl.sendRequest` and `DartToolingApiImpl.events`) might change slightly but I think everything else will probably remain (we still want the API classes and JSON handling for type safety - although these are something we might be able to code-gen - but for now they're hand-written).

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
```diff
@@ -2,18 +2,119 @@
...
+          ElevatedButton(
+            onPressed: () =>
+                unawaited(widget.api.executeCommand('flutter.createProject')),
+            child: const Text('New Flutter Project'),
+          ),
+        StreamBuilder(
+          stream: widget.api.devicesChanged,
+          builder: (context, snapshot) {
+            if (!snapshot.hasData) {
+              return const Text('');
```
kenzieschmoll, line 86: return `SizedBox.shrink()` for a placeholder instead

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
```diff
@@ -2,18 +2,119 @@
...
+                      TextButton(
+                        child: Text(device.name),
+                        onPressed: () =>
+                            unawaited(widget.api.selectDevice(device.id)),
+                      ),
+                      Text(
+                        device.id == deviceEvent.selectedDeviceId
+                            ? '(selected)'
+                            : '',
+                      ),
```
kenzieschmoll, line 104: this text can be part of the TextButton child Text widget above.

```
Text(
  '${device.name}'
  '${device.id == deviceEvent.selectedDeviceId ? ' (selected)' : ''}',
)
```

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
```diff
@@ -2,18 +2,119 @@
...
+            child: const Text('New Flutter Project'),
+          ),
+        StreamBuilder(
+          stream: widget.api.devicesChanged,
+          builder: (context, snapshot) {
+            if (!snapshot.hasData) {
+              return const Text('');
+            }
+            final deviceEvent = snapshot.data!;
+            return Table(
```
kenzieschmoll, line 89: can we use a ListView or a column instead?

packages/devtools_app/lib/src/standalone_ui/vs_code/mock_environment/dart_tooling_mock_api.dart
```diff
@@ -0,0 +1,164 @@
...
+  /// A set of mock devices that can be presented for testing.
+  final _mockDevices = <VsCodeDevice>[
+    VsCodeDeviceImpl(
+      id: 'myMac',
+      name: 'Mac',
+      category: 'desktop',
+      emulator: false,
+      emulatorId: null,
+      ephemeral: false,
+      platform: 'dartwin-x64',
```
kenzieschmoll, line 87: darwin or dartwin?

packages/devtools_app/lib/src/standalone_ui/vs_code/mock_environment/dart_tooling_mock_api.dart
```diff
@@ -0,0 +1,164 @@
...
+      emulator: false,
+      emulatorId: null,
+      ephemeral: true,
+      platform: 'android-x64',
+      platformType: 'android',
+    ),
+  ];
+
+  /// The current set of devices being presented to the embedded panel.
+  final List<VsCodeDevice> _devices = [];
```
kenzieschmoll, line 103: nit: put type on right hand side:
`final _devices = <VsCodeDevice>[]`

packages/devtools_app/lib/src/standalone_ui/vs_code/mock_environment/flutter_panel_mock_editor.dart
```diff
@@ -0,0 +1,143 @@
...
+import '../../../shared/theme.dart';
+import '../flutter_panel.dart';
+import 'dart_tooling_mock_api.dart';
+
+/// A simple UI that acts as a stand-in host IDE to simplify the development
+/// workflow when working on embedded tooling.
+///
+/// This UI interacts with [MockDartToolingApi] to allow triggering events that
+/// would normally be fired by the IDE and also shows a log of recent requests.
+class VsCodeFlutterPanelMockEditor extends StatefulWidget {
```
kenzieschmoll, line 20: I wonder if this should be implemented as a stager app that lives in our test/ directory, that way it doesn't have to get shipped with our production code. We have stager apps for several DevTools screens that can be used for simplified development and testing.

packages/devtools_app/lib/src/standalone_ui/api/impl/vs_code_api.dart
```diff
@@ -0,0 +1,149 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:json_rpc_2/json_rpc_2.dart' as json_rpc_2;
+import 'package:meta/meta.dart';
```
kenzieschmoll, line 6: do we need to add a dep on meta? I remember an effort to remove this dependency from devtools a while back, though I can't recall why. Probably because we were using it for `@required` and some other signatures that are now provided by flutter out of the box.

packages/devtools_app/lib/src/standalone_ui/api/dart_tooling_api.dart
```diff
@@ -0,0 +1,17 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'vs_code_api.dart';
+
+/// An API exposed to Dart tooling extensions.
```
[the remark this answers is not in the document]
DanTup: Done! Although I'm not sure these terms are completely clear to me (surfaces vs extensions).

I was thinking "Dart Tooling Extensions" would mean any extension that could consume the new APIs (via DTD), and that "DevTools Extensions" would be a subset of "Dart Tooling Extensions".

packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
```diff
@@ -0,0 +1,116 @@
...
+/// This is useful for debugging when running inside VS Code.
+///
+/// TODO(dantup): Make a way for this to be enabled by users at runtime for
+///  troubleshooting. This could be via a message from VS Code, or something
+///  that passes a query param.
+const _enablePostMessageVerboseLogging = false;
+
+final _log = Logger('tooling_api');
+
+/// An API for interacting with Dart tooling.
```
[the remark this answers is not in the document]
DanTup: I changed this to:

```
/// An API used by Dart tooling surfaces to interact with Dart tools that expose
/// APIs such as Dart-Code and the LSP server.
```

Is that clearer? I couldn't come up with a (non-vague) term to describe the components that will generally provide APIs.

packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
```diff
@@ -0,0 +1,116 @@
...
+      onPostMessage.map((event) {
+        _log.info('<== ${jsonEncode(event.data)}');
+        return event.data;
+      }),
+      postMessageController,
+    );
+    return DartToolingApiImpl.rpc(json_rpc_2.Peer.withoutJson(channel));
+  }
+
+  /// Connects the API over the provided WebSocket.
```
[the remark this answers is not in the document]
DanTup: I've removed this as it was unused. It was originally used to let you connect to VS Code without being embedded inside it (as a way to simplify debugging when interacting with VS Code), however I think that's somewhat superseded by the mock environment - and if we do want to support it in future we'll just use the DTD.

packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
```diff
@@ -0,0 +1,116 @@
...
+
+  @protected
+  Future<T> sendRequest<T>(String method, [Object? parameters]) async {
+    return (await rpc.sendRequest('$apiName.$method', parameters)) as T;
+  }
+
+  /// Listens for an event '[apiName].[name]' that has a Map for parameters.
+  @protected
+  Stream<Map<String, Object?>> events(String name) {
+    final streamController = StreamController<Map<String, Object?>>.broadcast();
```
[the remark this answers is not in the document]
DanTup: I've added a line here to automatically close it when `rpc` closes which I think does what you wanted (but also handles the connection going away without `dispose` being called explicitly).

packages/devtools_app/lib/src/standalone_ui/api/impl/vs_code_api.dart
```diff
@@ -0,0 +1,149 @@
...
+  @override
+  Future<bool> selectDevice(String id) {
+    return sendRequest(
+      'selectDevice',
+      {'id': id},
+    );
+  }
+}
+
+class VsCodeDeviceImpl implements VsCodeDevice {
```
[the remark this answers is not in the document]
DanTup: The goal was to make the API much clearer so it can more easily be mirrored/compared to VS Code. Without interfaces it might also be quite easy to make an accidental breaking change to the interface (and the mock environment) without VS Code.

I added some more comments to the interface classes to make this clearer, although I'm not married to the idea - if you'd prefer not to have them (and aren't concerned about accidental breakage), I'm happy to remove :-)

packages/devtools_app/lib/src/standalone_ui/api/impl/vs_code_api.dart
```diff
@@ -0,0 +1,149 @@
...
+  VsCodeDeviceImpl.fromJson(Map<String, Object?> json)
+      : this(
+          id: json['id'] as String,
+          name: json['name'] as String,
+          category: json['category'] as String?,
+          emulator: json['emulator'] as bool,
+          emulatorId: json['emulatorId'] as String?,
+          ephemeral: json['ephemeral'] as bool,
+          platform: json['platform'] as String,
+          platformType: json['platformType'] as String?,
```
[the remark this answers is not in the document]
DanTup: Done! I put them in the API interface class for now (since they're really part of the API) but if we combine them they'll come back here. FWIW I'd really prefer to code-gen this because there's a lot of boilerplate (that would be copy/pasted and easy to have mistakes), but probably something for later.

packages/devtools_app/lib/src/standalone_ui/api/impl/vs_code_api.dart
```diff
@@ -0,0 +1,149 @@
...
+class VsCodeDevicesEventImpl implements VsCodeDevicesEvent {
+  VsCodeDevicesEventImpl({
+    required this.selectedDeviceId,
+    required this.devices,
+  });
+
+  VsCodeDevicesEventImpl.fromJson(Map<String, Object?> json)
+      : this(
+          selectedDeviceId: json['selectedDeviceId'] as String?,
+          devices: (json['devices'] as List)
```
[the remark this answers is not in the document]
DanTup: Done!

packages/devtools_app/lib/src/standalone_ui/api/impl/vs_code_api.dart
```diff
@@ -0,0 +1,149 @@
...
+  @override
+  late final Stream<VsCodeDevicesEvent> devicesChanged;
+
+  @override
+  late final VsCodeCapabilities capabilities;
+
+  @override
+  Future<Object?> executeCommand(String command, [List<Object?>? arguments]) {
+    return sendRequest(
+      'executeCommand',
```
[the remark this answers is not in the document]
DanTup: Done!

packages/devtools_app/lib/src/standalone_ui/api/impl/vs_code_api.dart
```diff
@@ -0,0 +1,149 @@
...
+}
+
+class VsCodeCapabilitiesImpl implements VsCodeCapabilities {
+  VsCodeCapabilitiesImpl(this._raw);
+
+  final Map<String, Object?>? _raw;
+
+  @override
+  bool get executeCommand => _raw?['executeCommand'] == true;
+}
```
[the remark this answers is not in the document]
DanTup: Good spot, fixed!

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
```diff
@@ -2,18 +2,119 @@
...
+
+  final DartToolingApi api;
 
-class VsCodeFlutterPanel extends StatelessWidget {
-  const VsCodeFlutterPanel({super.key});
+  @override
+  State<VsCodeFlutterPanel> createState() => _VsCodeFlutterPanelState();
+}
 
+class _VsCodeFlutterPanelState extends State<VsCodeFlutterPanel> {
```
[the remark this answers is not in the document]
DanTup: You're right, fixed!

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
```diff
@@ -2,18 +2,119 @@
...
   Widget build(BuildContext context) {
     assert(FeatureFlags.vsCodeSidebarTooling);
-    return const Center(
-      child: Text('TODO: a panel for flutter actions in VS Code'),
+
+    final api = widget.api;
+
+    return Column(
+      children: [
+        const Text(''),
```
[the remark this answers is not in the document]
DanTup: I replaced this with `const SizedBox(height: 20),` (it was to add some padding to prevent the first item in the column touching the top of the window), but I'm not sure if that's the best way to do this?

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
DanTup: That would include it in the button so the styling would change - this was deliberately separate:

image

Although, this "design" is just a placeholder and we definitely don't want to ship it looking like this (we probably want to show more than just the name of a device for ex.).

packages/devtools_app/lib/src/standalone_ui/vs_code/mock_environment/flutter_panel_mock_editor.dart
DanTup: Oh that might be better. I'm not very familiar with stager though - I'll take a look at changing it this week. It would definitely be nicer if we could keep development screen out of the main app. Thanks!

packages/devtools_app/lib/src/standalone_ui/api/impl/vs_code_api.dart
DanTup: I added it to use @protected to try and avoid some of the base methods in the API classes showing up to the widget code. Unfortunately code completion still shows them (https://github.com/dart-lang/sdk/issues/35449), though it does generate a warning if you use them.

It was mostly just to enforce everything going through methods (eg. don't call `sendRequest('someString')` directly from a widget), but it's a fairly small benefit. If there's a reason to avoid the dependency (and it wasn't just removed because it was _unnecessary_), we could remove it. I tracked the change back to https://github.com/flutter/devtools/pull/3484 but I don't understand the reason - wdyt?

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
```diff
@@ -73,41 +65,43 @@ class _VsCodeConnectedPanelState extends State<_VsCodeConnectedPanel> {
   Widget build(BuildContext context) {
     return Column(
       children: [
+        const SizedBox(height: 20),
```
kenzieschmoll, line 68: nit use `defaultSpacing` so that we are consistent with other devtools spacing

packages/devtools_app/test/test_infra/scenes/standalone_ui/vs_code_mock_editor.dart
```diff
@@ -74,20 +91,35 @@ class _VsCodeFlutterPanelMockEditorState
...
+          splitters: [
+            PreferredSize(
+              preferredSize: const Size.fromHeight(1),
+              child: SizedBox(
+                height: 1,
+                child:
+                    ColoredBox(color: editorTheme.editorTerminalSplitterColor),
+              ),
+            ),
+          ],
```
kenzieschmoll, line 119: nit: do we need to manually specify a splitter here? is the `DefaultSplitter` that is used when `splitters` is null not sufficient?

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
kenzieschmoll: Fine to land as is and iterate with some ux input later

kenzieschmoll (commented):
This is looking good. I'm happy we were able to move the mock environment into a stager app so that it is not part of our production code. We'll likely need some clean up and iteration on the UI, but okay to land and iterate while we build these embedded features out.

packages/devtools_app/test/test_infra/scenes/standalone_ui/vs_code_mock_editor.dart
DanTup: Not really. I was trying to make it look a little more like VS Code (so it gave a better idea of how it'd look) so this was to set the colour and make it narrower. It didn't work particularly well though, so I've removed it for now.

kenzieschmoll (approved):

DanTup:
The changes _above_ this comment were just a rebase with no changes (because it's been a while since I opened this). Seems like a few things have changed in master and there are some errors, so the changes _after_ this comment will be real changes to fix them.

DanTup:
Turned out to just be needing imports for `package:devtools_app_shared/ui.dart` and no other changes, so merging.

Issue #6095, opened 2023-07-25
Add instructions for how to use Chrome DevTools to view timeline events for Dart web apps


kenzieschmoll:
This will require enabling the performance page for web and conditionally switching out the body to show these instructions when we have a web app.

PR #6107, opened 2023-07-26
Performance panel instructions for web apps

elliette:
Adds instructions to the performance panel for web apps on how to use Chrome DevTools for performance profiling. 

For Flutter web users:

Screenshot 2023-07-26 at 2 59 02 PM

For Dart web users:

Screenshot 2023-07-26 at 2 59 35 PM

Fixes https://github.com/flutter/devtools/issues/6095

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/performance_screen.dart
+++ b/packages/devtools_app/lib/src/screens/performance/performance_screen.dart
@@ -156,6 +156,12 @@ const timelineTaskLink =
     'https://api.flutter.dev/flutter/dart-developer/TimelineTask-class.html';
 const debugBuildsLink =
     'https://api.flutter.dev/flutter/widgets/debugProfileBuildsEnabled.html';
+const debugUserBuildsLink =
+    'https://api.flutter.dev/flutter/widgets/debugProfileBuildsEnabledUserWidgets.html';
+const debugLayoutsLink =
+    'https://api.flutter.dev/flutter/rendering/debugProfileLayoutsEnabled.html';
+const debugPaintsLink =
+    'https://api.flutter.dev/flutter/rendering/debugProfilePaintsEnabled.html';
 const profileModeLink = 'https://docs.flutter.dev/testing/build-modes#profile';
 const performancePanelLink =
     'https://developer.chrome.com/docs/devtools/performance';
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/screens/performance/performance_screen.dart
```diff
@@ -131,8 +133,60 @@ class WebPerformanceScreenBody extends StatelessWidget {
...
+const profileModeLink = 'https://docs.flutter.dev/testing/build-modes#profile';
+const performancePanelLink =
+    'https://developer.chrome.com/docs/devtools/performance';
+
+const flutterWebInstructionsMd = '''
+# How to use Chrome DevTools for performance profiling
+
+The timeline events emitted by the Flutter framework as it works to build
+frames, draw scenes, and track other activity such as HTTP request timings and
+garbage collection are exposed in the Chrome DevTools performance panel.
```
kenzieschmoll, line 168: wording nit to make this a bit easier to read.

"The Flutter framework emits timeline events as it works to build frames, draw scenes, and track other activity such as HTTP request timings and garbage collections. These events are exposed in the Chrome DevTools performance panel for debugging."

packages/devtools_app/lib/src/screens/performance/performance_screen.dart
```diff
@@ -131,8 +133,60 @@ class WebPerformanceScreenBody extends StatelessWidget {
...
+
+You can also emit your own timeline events using the `dart:developer` 
+[Timeline]($timelineLink) and [TimelineTask]($timelineTaskLink) APIs for further
+performance analysis.
+
```
kenzieschmoll, line 173: Can we add a more detailed section here?
`## Optional flags to enhance tracing`

We should these flags and a small description:
- debugProfileBuildsEnabled
- debugProfileBuildsEnabledUserWidgets
- debugProfileLayoutsEnabled
- debugProfilePaintsEnabled
-   HttpClient.enableTimelineLogging (we may need to check if this can be enabled for web since this is from dart:io...if it can't be enabled for web, then we should remove the reference to HTTP request timings in the above description)

packages/devtools_app/lib/src/screens/performance/performance_screen.dart
```diff
@@ -131,8 +133,60 @@ class WebPerformanceScreenBody extends StatelessWidget {
...
+## Instructions:
+
+1. *[Optional]* Set [debugProfileBuildsEnabled]($debugBuildsLink) to true to 
+track your app's widget builds.
```
kenzieschmoll, line 177: then if we have the "Optional flags to enhance tracing" section above, here we can say:
"*[Optional]* set any desired tracing flags to true from your app's main method."

packages/devtools_app/lib/src/screens/performance/performance_screen.dart
```diff
@@ -131,8 +133,60 @@ class WebPerformanceScreenBody extends StatelessWidget {
...
+2. Run your Flutter web app in [profile mode]($profileModeLink).
+3. Open up the [Chrome DevTools' Performance panel]($performancePanelLink) for
+your application, and start recording to capture timeline events.
+''';
+
+const dartWebInstructionsMd = '''
```
kenzieschmoll, line 183: for Dart web app users, should we also recommend that they build their app in AOT mode?

packages/devtools_app/lib/src/screens/performance/performance_screen.dart
elliette: Done!

packages/devtools_app/lib/src/screens/performance/performance_screen.dart
elliette: Added those! HttpClient.enableTimelineLogging can't be used for web apps, so removed from list (and from the above description)

packages/devtools_app/lib/src/screens/performance/performance_screen.dart
elliette: Done

packages/devtools_app/lib/src/screens/performance/performance_screen.dart
elliette: There's different ways to do so (depending on if they are using webdev, acx, etc) so leaving as is for now

kenzieschmoll (approved):

PR #6111, opened 2023-07-27
Add extensionActivationState endpoint to the devtools server

kenzieschmoll:
This endpoint manages reading and writing to/from the `devtools_options.yaml` file. This file will store DevTools extension activation states and will live in a project's root directory (similar to `analysis_options.yaml`).

Example `devtools_options.yaml` file:
```
extensions:
  - foo: true
  - bar: false
```
In this scenario:
- the devtools extension for `package:foo` has been enabled by the user 
- the devtools extension for `package:bar` has been disabled by the user
- the devtools extension for some other `package:baz` has been neither enabled nor disabled (meaning that DevTools should prompt the user to activate).

This API will be consumed by DevTools app in a follow up PR.

work towards https://github.com/flutter/devtools/issues/1632


packages/devtools_shared/lib/src/server/server_api.dart
```diff
@@ -207,23 +208,62 @@ class ServerApi {
 
       // ----- Extensions api. -----
 
-      case apiServeAvailableExtensions:
+      case ExtensionsApi.apiServeAvailableExtensions:
```
bkonyi, line 211: This `switch` statement is ~220 lines now and should probably be refactored (this doesn't have to block this PR, but consider applying these changes to the new APIs for now). A few things that could be changed:

- Almost every request creates a variable for `queryParams`, so it's probably just worth declaring before the switch statement
- Each API endpoint should have its own method. Ideally, the switch statement will only contain a single function call for each case
- I'd create a helper function for creating bad request responses for missing query parameters since they're all basically the same format
- Going even further, if there's a check we perform multiple times, we could also use a helper to perform the check (e.g., `bool checkRootPath(Map<String, Object?> queryParams)` or `bool checkRootPath(String? rootPath)`)
- It might also be worth creating a `FutureOr<shelf.Response> encodeResponse(Object? response) => api.getCompleted(jsonEncode(response))`

packages/devtools_shared/test/extensions/extension_activation_test.dart
```diff
@@ -0,0 +1,110 @@
...
+  group('$DevToolsOptions', () {
+    late DevToolsOptions options;
+    late Directory tmpDir;
+
+    const tmpPath = '_tmp';
+    final tmpUri = Uri.parse(tmpPath);
+
+    setUp(() {
+      options = DevToolsOptions();
+      tmpDir = Directory(tmpPath)..createSync();
```
bkonyi, line 21: Why not just [`Directory.createTempSync()`](https://api.dart.dev/stable/3.0.7/dart-io/Directory/createTempSync.html)?

packages/devtools_shared/lib/src/server/server_api.dart
kenzieschmoll: good ideas. Implemented for the two extension related apis and added a TODO for the rest.

packages/devtools_shared/lib/src/server/server_api.dart
```diff
@@ -315,3 +294,64 @@ class ServerApi {
...
       shelf.Response(HttpStatus.noContent);
 }
+
+abstract class _ExtensionsApiHandler {
+  static FutureOr<shelf.Response> handleServeAvailableExtensions(
+    ServerApi api,
+    Map<String, String> queryParams,
+    ExtensionsManager extensionsManager,
+  ) async {
+    final missingRequiredParams = await ServerApi._checkRequiredParameters(
```
bkonyi, line 304: Something doesn't feel right about this. Do we need to return `FutureOr<T>` for any of the methods in this class? AFAICT, it's just causing async pollution throughout this file. We definitely shouldn't need to do any asynchronous work to check an entry in a map.

`FutureOr<T>` should really only be used when declaring a type for a callback can either be synchronous or asynchronous.

Consider updating the functions in this file to only return `shelf.Response` if they're not doing any asynchronous work. Otherwise, they can return `Future<shelf.Response>` where appropriate.

kenzieschmoll: good catch. The `ServerApi.handle` method still needs to return `FutureOr` since some of the api handlers do async activity, but I cleaned up what I could.

bkonyi (approved):

PR #6112, opened 2023-07-27
Use `onItemExpanded` instead of `onItemSelected` for `ExpandableVariable`

elliette:
Just a small thing I noticed during DAP migration - we are using `onItemSelected` when really what we care about is whether the variable has been expanded. Therefore `onItemExpanded` is more appropriate (and allows us to remove the check for `isExpanded`)


DanTup (approved):
LGTM

Issue #2989, opened 2021-05-06
Create ValueListenablesBuilder class


jacob314:
We have cases where we need to listen on multiple value listenables but creating a stateful widget wouldn't make sense. Multiple nested value listenable builders are inefficient and hard to read.
Create a `ValueListenablesBuilder` that lets you listen on 2 or 3 value listenables with a builder callback that matches the # of value listenables passed to the constructor.
Alternately, create a util that builds a ValueListenable from multiple value listenables returning a tuple with the current values.

kenzieschmoll:
Reopening this issue. We have DualValueListenableBuilder, which is helpful, but there are still cases where we have more than two ValueListenables we want to listen to and build on changes.

PR #6129, opened 2023-07-31
Replace `DualValueListenableBuilder` with `MultiValueListenableBuilder`

kenzieschmoll:
This allows us to listen and rebuild on an arbitrary number of listeners instead of being limited to 2.

Fixes https://github.com/flutter/devtools/issues/2989

### DIFF
--- a/packages/devtools_app/lib/src/shared/primitives/utils.dart
+++ b/packages/devtools_app/lib/src/shared/primitives/utils.dart
@@ -1214,6 +1214,10 @@ extension ListExtension<T> on List<T> {
     }
     return false;
   }
+
+  T get second => this[1];
+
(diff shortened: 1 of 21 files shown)


packages/devtools_app/lib/src/shared/primitives/utils.dart
```diff
@@ -1214,6 +1214,10 @@ extension ListExtension<T> on List<T> {
     }
     return false;
   }
+
+  T get second => this[1];
+
+  T get third => this[2];
```
bkonyi, line 1220: We should really just add extensions for `first` thru `ninth` or `ninetyninth` if we want to be future proof... 😜

bkonyi (approved):

PR #6132, opened 2023-07-31
Remove `devtools_shared` dependency on `package:io`

kenzieschmoll:
Copying this code prevents us from having to introduce an SDK dependency on `package:io`, which is likely overkill for use of this one method.


CoderDake (commented):
Are there any specs we could copy along with this?

bkonyi (approved):

PR #6139, opened 2023-08-01
Use `package:extension_discovery` to detect DevTools extensions

kenzieschmoll:
Depends on https://github.com/dart-lang/tools/pull/129 landing and `package:extension_discovery` being published.

I have verified this package works as expected for the DevTools use case of detecting available extensions for a running application.
CC @jonasfj


packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -57,13 +58,16 @@ class ExtensionsManager {
...
+          path.join(
+            rootPath,
+            '.dart_tool',
+            'package_config.json',
+          ),
+        ),
+      );
       for (final extension in extensions) {
         final config = extension.config;
         if (config is! Map) {
```
jonasfj, line 73: nit: I'd suggest printing a warning to stderr, if for some reason the extension doesn't work.

Or at-least do it, if the extension is in the root package (`rootPath == extension.rootUri`. If I'm working on a package that contains an extension, I should probably be told if the extension doesn't work. Less important, if the extension is from a dependency.

kenzieschmoll: I can't think of a scenario where the root path would be equal to the `extension.rootUir`. For example, if I'm a package author for `package:foo`, my extension would live at `path/to/foo/extension/devtools`, but root uri would be for some app that uses foo (`path/to/some_app_that_depends_on_foo`), not `path/too/foo`

kenzieschmoll:
Before landing this, we need to make sure that package:extension_discovery has been added as a DEP in the Dart SDK and that this package is rolled into google3.

packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -57,17 +58,31 @@ class ExtensionsManager {
...
+            path.join(
+              rootPath,
+              '.dart_tool',
+              'package_config.json',
+            ),
+          ),
+        );
+      } catch (e) {
+        print('[ERROR] `findExtensions` failed: $e');
+        extensions = <Extension>[];
```
polina-c, line 75: Do we want to report thid to analytics?

packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -57,17 +58,31 @@ class ExtensionsManager {
...
+      }
+
       for (final extension in extensions) {
         final config = extension.config;
         if (config is! Map) {
           // Fail gracefully. Invalid content in the extension's config.json.
+          print(
+            '[WARNING] invalid config.json content for ${extension.package}:\n'
+            '$config',
```
polina-c, line 84: And this?

polina-c (approved):
Couple comments and LGTM.

PR #6140, opened 2023-08-02
Add ability to enable and disable an extension from the extension screen

kenzieschmoll:
When an extension is loaded for the first time and there is no entry in `devtools_options.yaml` for this extension, an enable prompt will be shown in place of the extension iFrame:
Screenshot 2023-08-01 at 5 11 18 PM

The user's answer will be recorded to the `devtools_options.yaml` file by proxy of the devtools server.

Once an extension has been enabled, there will be an option to disable the extension from the extension tool header:
Screenshot 2023-08-01 at 4 59 59 PM

When clicked, the user will be asked if they really want to disable the extension. This is preferred because an accidental click here could destroy any state in the extension tool. This also makes the experience less jarring when the screen is removed and the user is navigated back to the DevTools home screen.
Screenshot 2023-08-01 at 5 00 08 PM

This PR also ensures that tabs for extensions are automatically added / removed based on the enabled state for each extension.


packages/devtools_app/lib/src/extensions/extension_screen.dart
```diff
@@ -112,50 +124,6 @@ class ExtensionView extends StatelessWidget {
   }
 }
 
-// TODO(kenz): add button to deactivate extension once activate / deactivate
```
kenzieschmoll, line 115: moved to extension_screen_controls.dart

polina-c:
Do we want to be transparent to the user that we will record the answer in the file, so that they know they could save this file in source control?
Where the file is located?

kenzieschmoll:
> Do we want to be transparent to the user that we will record the answer in the file, so that they know they could save this file in source control?
Where the file is located?

If the user does not already have a `devtools_options.yaml` file, one will be created, and if they do already have one, it will be modified. So they should see this come up in their git diff automatically, and from there they can choose to `.gitignore` it or check it in. I don't know if we should add this sort of information into DevTools itself, but we can certainly add this in documentation. For example, we don't notify users that we persist settings in a `~/.flutter-devtools/.devtools`, we just save them there.

polina-c (approved):

PR #6147, opened 2023-08-03
Add Deep link validation tool  home page with fake data

hangyujin:
Work towards https://github.com/flutter/devtools/issues/6013.

 All work for the deep link validation tool will be hidden behind FeatureFlags.deepLinkValidation until the tool is ready to ship.

D785360C-BBB8-4AB1-AC6A-8E0D8F967CD2

![build.yaml badge]


packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -3,9 +3,115 @@
...
+    this.domainError = false,
+    this.pathError = false,
+  });
+
+  String os;
+  String path;
+  String domain;
+  String scheme;
+  bool domainError;
+  bool pathError;
```
kenzieschmoll, line 57: can these be final?

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -3,9 +3,115 @@
...
+const List<String> paths = [
+  '/shoes/..*',
+  '/Clothes/..*',
+  '/Toys/..*',
+  '/Jewelry/..*',
+  '/Watches/..* ',
+  '/Glasses/..*',
+];
+
+List<LinkData> allLinkDatas = [
```
kenzieschmoll, line 17: make these lists private and add dart doc to describe what they are for

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -3,9 +3,115 @@
 // found in the LICENSE file.
 
 import 'package:flutter/material.dart';
-
 import '../../shared/screen.dart';
 
+const List<String> paths = [
```
kenzieschmoll, line 8: nit: either remove type or put on the RHS:
`const _paths = <String>[ ... ];`

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -3,9 +3,115 @@
...
+  for (var path in paths)
+    LinkData(
+      os: 'Android',
+      domain: 'm.chinese.shopping.com',
+      path: path,
+      pathError: path.contains('shoe'),
+    ),
+];
+
+class LinkData {
```
kenzieschmoll, line 42: please add dartdoc

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -3,9 +3,115 @@
...
+
+  String os;
+  String path;
+  String domain;
+  String scheme;
+  bool domainError;
+  bool pathError;
+  String get searchLabel => (os + path + domain + scheme).toLowerCase();
+
+  DataRow buildRow(
```
kenzieschmoll, line 60: instead of having a method that returns a widget, create a stateless widget `_LinkDataRow`. In general, always having a method that returns a widget is code smell that we should just create a stateless widget instead. This is cleaner stylistically and it makes it easier to debug widget trees with the inspector.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -3,9 +3,115 @@
...
+    MaterialStateProperty<Color?>? color,
+  }) {
+    return DataRow(
+      color: color,
+      cells: [
+        DataCell(Text(os)),
+        DataCell(Text(scheme)),
+        DataCell(
+          domainError
+              ? Row(
```
kenzieschmoll, line 71: I would play with the screen size a bit and make sure this does not overflow. If it does, you'll likely need to use Flexible or Expanded around some of the children

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -3,9 +3,115 @@
...
+                  children: [
+                    Icon(
+                      Icons.error,
+                      color: Theme.of(context).colorScheme.error,
+                    ),
+                    const SizedBox(width: 10),
```
kenzieschmoll, line 77: we have lots of consts you can reuse in `theme.dart` to ensure consistent styling across our tools. Use `denseSpacing` (8.0) here and elsewhere instead. We also have `defaultSpacing` with is 16.0. Those are likely the two most popular spacing consts you'll be using.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
kenzieschmoll, line 71: instead of using a ternary, you can conditionally add children to the list in Row:

```
Row(
  children: [
    if (domainError)
      Padding(
        padding: const EdgeInsets.only(right: denseSpacing),
        child: Icon (...),
      ),
    Text(domain),
  ],
)
```

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -3,9 +3,115 @@
...
+                    Text(domain),
+                  ],
+                )
+              : Text(domain),
+        ),
+        DataCell(
+          pathError
+              ? Row(
```
kenzieschmoll, line 85: same comment here about conditionally adding items to the Row's children based on the value of `pathError` rather than using a ternary

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
kenzieschmoll, line 42: consider putting classes related to deep link data models in a different file `deep_links_model.dart` so that `deep_links_screen.dart` can consist of widget classes and UI code only.

Across devtools we generally keep UI classes (flutter widgets), model classes, and controller classes all in their separate categories. Not one class per file by any means, but grouping classes by UI, model, and controller helps with organization

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -24,8 +130,112 @@ class DeepLinksScreen extends Screen {
...
+            ),
+          ],
+        ),
+        const SizedBox(height: 10),
+        buildDataTable(context),
+      ],
     );
   }
+
+  Widget buildDataTable(BuildContext context) {
```
kenzieschmoll, line 193: make this a stateless widget class instead of having a method that returns a widget

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -24,8 +130,112 @@ class DeepLinksScreen extends Screen {
...
+class _DeepLinkPageState extends State<DeepLinkPage> {
+  String responseBody = 'empty response';
+  int? selectedRowIndex;
+  String searchContent = '';
+  bool bundleByDomain = false;
+
+  void setBundleByDomain(bool shouldBundleByDomain) {
+    setState(() {
+      bundleByDomain = shouldBundleByDomain;
+    });
```
kenzieschmoll, line 153: instead of using setState directly, I recommend creating a controller class with ValueNotifiers that the UI listens to. If you keep state in the DeepLinkPage state object directly, the state will be lost if the user switches to a separate screen or does something to cause a complete tear down and rebuild of the state object.

We do this for almost every screen in DevTools. Look where `DeepLinkPage` is created to see how other screens pass in a controller that will live for the entire lifecycle of DevTools. Then you can keep anything stateful in the controller object and never have to call set state directly or store data in the widget state.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -24,8 +130,112 @@ class DeepLinksScreen extends Screen {
...
+
+  void setResponse(String response) {
+    setState(() {
+      responseBody = response;
+    });
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return ListView(
```
kenzieschmoll, line 164: use a column instead and wrap the data table in an Expanded widget

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -24,8 +130,112 @@ class DeepLinksScreen extends Screen {
...
+              onChanged: (value) {
+                setState(() {
+                  searchContent = value.toLowerCase();
+                });
+              },
+              constraints: BoxConstraints.tight(const Size(200, 40)),
+            ),
+          ],
+        ),
+        const SizedBox(height: 10),
```
kenzieschmoll, line 187: use denseSpacing or defaultSpacing

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -24,8 +130,112 @@ class DeepLinksScreen extends Screen {
...
+        Row(
+          children: [
+            Expanded(child: buildTitle('All deep links', context)),
+            InkWell(
+              onTap: () {
+                setBundleByDomain(!bundleByDomain);
+              },
+              child: const Text('Bundle by domain'),
+            ),
+            const SizedBox(width: 10),
```
kenzieschmoll, line 175: use denseSpacing or defaultSpacing

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -24,8 +130,112 @@ class DeepLinksScreen extends Screen {
...
+      dataRowMinHeight: bundleByDomain ? 150 : null,
+      dataRowMaxHeight: bundleByDomain ? 150 : null,
+      rows: rows,
+    );
+  }
+}
+
+Widget buildTitle(String text, BuildContext context) {
+  final textTheme = Theme.of(context).textTheme;
+  return Text(
```
kenzieschmoll, line 237: declare this Text widget directly in the build method it is used in above instead of having a method that returns a widget. This is too simple to warrant creating its own stateless widget class

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -24,8 +130,112 @@ class DeepLinksScreen extends Screen {
...
+    ];
+
+    return DataTable(
+      columns: const [
+        DataColumn(label: Text('OS')),
+        DataColumn(label: Text('Scheme')),
+        DataColumn(label: Text('Domain')),
+        DataColumn(label: Text('Path')),
+      ],
+      dataRowMinHeight: bundleByDomain ? 150 : null,
```
kenzieschmoll, line 228: please put 150 in a named const and use that here and below

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -24,8 +130,112 @@ class DeepLinksScreen extends Screen {
...
+
+    final List<DataRow> rows = [
+      for (var i = 0; i < linkDatas.length; i++)
+        linkDatas[i].buildRow(
+          context,
+          color: i % 2 == 0
+              ? MaterialStateProperty.all<Color>(
+                  Theme.of(context).colorScheme.onSurface.withOpacity(0.2),
+                )
+              : null,
```
kenzieschmoll, line 217: use `alternatingColorForIndex` from `common_widgets.dart`

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -24,8 +130,112 @@ class DeepLinksScreen extends Screen {
...
+          children: [
+            Expanded(child: buildTitle('All deep links', context)),
+            InkWell(
+              onTap: () {
+                setBundleByDomain(!bundleByDomain);
+              },
+              child: const Text('Bundle by domain'),
+            ),
+            const SizedBox(width: 10),
+            SearchBar(
```
kenzieschmoll, line 176: may be too advanced for the current use case, but consider using `SearchField` from `search.dart` in DevTools, which has advanced functionality to hook search into your screen controller and provides some advanced features like navigation through results.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -24,8 +130,112 @@ class DeepLinksScreen extends Screen {
...
+              leading: const Icon(Icons.search),
+              onChanged: (value) {
+                setState(() {
+                  searchContent = value.toLowerCase();
+                });
+              },
+              constraints: BoxConstraints.tight(const Size(200, 40)),
```
kenzieschmoll, line 183: please use `defaultSearchFieldWidth` and `defaultTextFieldHeight` from `theme.dart`. Many consts in theme.dart are also wrapped with `scaleByFontFactor` which ensures your UI will scale nicely as the font factor changes

kenzieschmoll (commented):
done with first pass

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
chunhtai: agree, this seems like a data class, it would make more sense to move this method out of this class.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -3,9 +3,115 @@
...
+
+  LinkData mergeByDomain(LinkData? linkData) {
+    if (linkData == null) {
+      return this;
+    }
+
+    return LinkData(
+      os: os,
+      domain: domain,
+      path: '${linkData.path}\n$path',
```
chunhtai, line 109: seems weird to use an arbitrary delimiter, why not keep a list instead?

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -3,9 +3,115 @@
...
+  }
+
+  LinkData mergeByDomain(LinkData? linkData) {
+    if (linkData == null) {
+      return this;
+    }
+
+    return LinkData(
+      os: os,
+      domain: domain,
```
chunhtai, line 108: does the domain need to be the same?

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
hangyujin: tried and it looks like DataTable widget already handles narrow screen well.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
hangyujin: Updated it to use 

ValueListenableBuilder<List<LinkData>>(
      valueListenable: controller.linkDatasNotifier,

so  it is not a  a stateless widget  now.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
hangyujin: The data table can be longer than the screen height.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
hangyujin: yes

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
kenzieschmoll: While `ValueListenableBuilder` is not a stateless widget, a helper widget would still be stateless. In general, we never want to return a widget from a method. Always prefer a helper widget instead. 

```
class _DataTable extends StatelessWidget {
  ...
  
    return ...;
  }
}
```

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -24,8 +28,138 @@ class DeepLinksScreen extends Screen {
...
+          dataRowMinHeight: bundleByDomain ? bundledDataRowHeight : null,
+          dataRowMaxHeight: bundleByDomain ? bundledDataRowHeight : null,
+          rows: rows,
+        );
+      },
     );
   }
 }
+
+DataRow buildRow(
```
kenzieschmoll, line 122: same here. Make this a helper widget instead of having a method that returns a widget

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_model.dart
```diff
@@ -0,0 +1,37 @@
...
+  final bool domainError;
+  final bool pathError;
+
+  String get searchLabel => (os + paths.join() + domain + scheme).toLowerCase();
+
+  LinkData mergeByDomain(LinkData? linkData) {
+    if (linkData == null) {
+      return this;
+    }
+
```
chunhtai, line 29: can you add an assert that the domain is the same?

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
hangyujin: done

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_model.dart
hangyujin: done

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
hangyujin: the `DataRow` is not a widget, it's a data structure used in `DataTable`

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -4,7 +4,11 @@
 
 import 'package:flutter/material.dart';
 
-import '../../shared/screen.dart';
+import '../../../devtools_app.dart';
+import 'deep_links_controller.dart';
+import 'deep_links_model.dart';
+
+const bundledDataRowHeight = 150.0;
```
chunhtai, line 11:
```suggestion
const kBundledDataRowHeight = 150.0;
```

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_controller.dart
```diff
@@ -2,4 +2,51 @@
...
+    List<LinkData> linkDatas = searchContent.isNotEmpty
+        ? allLinkDatas
+            .where(
+              (linkData) => linkData.searchLabel.contains(searchContent),
+            )
+            .toList()
+        : allLinkDatas;
+
+    if (bundleByDomain) {
+      final Map<String, LinkData> bundleByDomainMap = {};
```
chunhtai, line 43:
```suggestion
      final bundleByDomainMap = <String, LinkData>{};
```
same for other map initialization

chunhtai: @kenzieschmoll is there a lint for this?

chunhtai (approved):
LGTM if also LGT @kenzieschmoll , left some nit picks

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_controller.dart
kenzieschmoll: Not that I'm aware of. CC @pq.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_controller.dart
```diff
@@ -2,4 +2,51 @@
...
+import 'package:flutter/foundation.dart';
+
+import 'deep_links_model.dart';
+import 'fake_data.dart';
+
+
+class DeepLinksController {
+
+  final _linkDatasNotifier = ValueNotifier<List<LinkData>>(allLinkDatas);
+  ValueListenable<List<LinkData>> get linkDatasNotifier => _linkDatasNotifier;
```
kenzieschmoll, line 14: style nit: order getters before private variables: https://github.com/flutter/devtools/blob/master/packages/README.md#order-of-getters-and-setters. Here and elsewhere

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -4,7 +4,11 @@
 
 import 'package:flutter/material.dart';
 
-import '../../shared/screen.dart';
+import '../../../devtools_app.dart';
```
kenzieschmoll, line 7: do not import devtools_app.dart directly inside of `lib/`. This is fine to import in tests, but inside lib we should import the libraries we need directly instead of importing this large file that exports many files.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
kenzieschmoll: Do we really want the controls to scroll out of view if the data table is large though? I suggest using a column and then making the data table itself scrollable so that we do not lose the controls to scroll through the table. You can see an example of this on other DevTools pages like the logging page, for example, where the controls at the top of the screen are fixed and only the table scrolls.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -24,8 +28,144 @@ class DeepLinksScreen extends Screen {
...
+
+class _DataTable extends StatelessWidget {
+  const _DataTable({required this.bundleByDomain, required this.linkDatas});
+  final bool bundleByDomain;
+  final List<LinkData> linkDatas;
+
+  @override
+  Widget build(BuildContext context) {
+    final List<DataRow> rows = [
+      for (var i = 0; i < linkDatas.length; i++)
```
kenzieschmoll, line 104: will `linkDatas` be large? We should be careful about building the entire table if there will be a lot of data here, as this could cause performance issues. If the data set is large, we should build this table in a lazy way.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_controller.dart
pq: [`omit_local_variable_types`](https://dart.dev/tools/linter-rules/omit_local_variable_types) handles this case but it may be a bridge too far... Check it out?

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
hangyujin: SearchableFlatTable seems to highlight the result but I want to filter the data with the search content.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
hangyujin: Thank you! I've updated it to use the FlatTable

chunhtai: some large company may have ~1000 deeplink.

hangyujin:
Updated UI to match the new mock 
20D8B5A1-B956-4DC5-89FC-2D582CD5C14F

2C8D31DA-4278-42C1-A7A1-FBCA73E5D8D8

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_controller.dart
```diff
@@ -7,46 +7,63 @@ import 'package:flutter/foundation.dart';
...
-        bundleByDomainMap[linkData.domain] =
-            linkData.mergeByDomain(bundleByDomainMap[linkData.domain]);
-      }
-      linkDatas = bundleByDomainMap.values.toList();
-    }
     _linkDatasNotifier.value = linkDatas;
   }
+
+  List<LinkData> get getLinkDatasByPath {
+    final Map<String, LinkData> linkDatasByPath = {};
```
kenzieschmoll, line 53: nit: here and for the method below, put types on the RHS: `final linkDatasByPath = <String, LinkData>{};`

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_model.dart
```diff
@@ -3,37 +3,180 @@
...
+  @override
+  Widget getDisplayWidget(LinkData dataObject, BuildContext context) {
+    return Row(
+      children: [
+        if (dataObject.domainError)
+          Padding(
+            padding: const EdgeInsets.only(right: denseSpacing),
+            child: Icon(
+              Icons.error,
```
kenzieschmoll, line 65: add `size: defaultIconSize`

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_model.dart
```diff
@@ -3,37 +3,180 @@
...
+              color: Theme.of(context).colorScheme.error,
+            ),
+          ),
```
kenzieschmoll, line 68: consider adding an `else` statement that adds a `SizedBox()` of the same width as the total Padding + Icon width, so that the table rows do not have staggered indents depending on whether there is an error or not

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_model.dart
```diff
@@ -3,37 +3,180 @@
...
+        const SizedBox(width: denseSpacing),
+        Text(
+          dataObject.path,
+          overflow: TextOverflow.ellipsis,
+        ),
+      ],
```
kenzieschmoll, line 109: This is repeated above. Please create a helper widget `_ErrorAwareText` or similar that you can use in both DomainColumn and PathColumn

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_model.dart
```diff
@@ -3,37 +3,180 @@
...
+    if (dataObject.domainError || dataObject.pathError) {
+      return Text(
+        getValue(dataObject),
+        overflow: TextOverflow.ellipsis,
+        style: TextStyle(color: Theme.of(context).colorScheme.error),
+      );
+    } else {
+      return const Text(
+        'No issues found',
+        style: TextStyle(color: Color.fromARGB(255, 156, 233, 195)),
```
kenzieschmoll, line 165: can we use a color from `theme.dart` instead of creating a new one here?

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -52,119 +62,251 @@ class _DeepLinkPageState extends State<DeepLinkPage>
...
-    return ListView(
-      children: [
-        Row(
-          children: [
-            Expanded(
-              child: Text(
-                'All deep links',
-                style: Theme.of(context).textTheme.bodyLarge,
+    return DefaultTabController(
+      length: 3,
```
kenzieschmoll, line 66: Use `TableView.values.length`

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_controller.dart
```diff
@@ -7,46 +7,63 @@ import 'package:flutter/foundation.dart';
...
   final _searchContentNotifier = ValueNotifier<String>('');
-  ValueListenable<String> get searchContentNotifier => _searchContentNotifier;
-  set searchContent(String content) {
-    _searchContentNotifier.value = content;
-    _updateLinks();
+  final _showSpitScreenNotifier = ValueNotifier<bool>(false);
+
+  set showSpitScreen(bool value) {
+    _showSpitScreenNotifier.value = value;
   }
```
kenzieschmoll, line 27: If we are going to have a public getter and setter for a notifier, then there is no need to have a private notifier and public ValueListenable. So for this one, I reccomend we just make the `_showSplitScreenNotifier` public

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_controller.dart
```diff
@@ -7,46 +7,63 @@ import 'package:flutter/foundation.dart';
 import 'deep_links_model.dart';
 import 'fake_data.dart';
 
-
 class DeepLinksController {
-
-  final _linkDatasNotifier = ValueNotifier<List<LinkData>>(allLinkDatas);
   ValueListenable<List<LinkData>> get linkDatasNotifier => _linkDatasNotifier;
+  ValueListenable<String> get searchContentNotifier => _searchContentNotifier;
+  ValueListenable<bool> get showSpitScreenNotifier => _showSpitScreenNotifier;
```
kenzieschmoll, line 13: style nit: order getters, private members, and setters per the guidelines here: https://github.com/flutter/devtools/blob/master/packages/README.md#order-of-getters-and-setters

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -52,119 +62,251 @@ class _DeepLinkPageState extends State<DeepLinkPage>
...
+      child: Column(
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          Row(
```
kenzieschmoll, line 70: Can you try to use `AreaPaneHeader` for consistent styling with the rest of DevTools?

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -52,119 +62,251 @@ class _DeepLinkPageState extends State<DeepLinkPage>
...
-            const SizedBox(width: denseSpacing),
-            SearchBar(
-              leading: const Icon(Icons.search),
-              onChanged: (value) {
-                controller.searchContent = value;
-              },
-              constraints: BoxConstraints.tight(
-                Size(defaultSearchFieldWidth, defaultTextFieldHeight),
+              const SizedBox(width: denseSpacing),
+              SearchBar(
```
kenzieschmoll, line 79: Are you able to use `SearchField` from `search.dart`? There are a few examples of this throughout DevTools, but this would give you a search field styled consistently with the rest of DevTools that supports navigation through search results.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -52,119 +62,251 @@ class _DeepLinkPageState extends State<DeepLinkPage>
...
+                  controller.searchContent = value;
+                },
+                constraints: BoxConstraints.tight(
+                  Size(wideSearchFieldWidth, defaultTextFieldHeight),
+                ),
+              ),
+            ],
+          ),
+          const SizedBox(height: denseSpacing),
+          const TabBar(
```
kenzieschmoll, line 92: consider using `AnalyticsTabbedView` so that you will automatically get analytics reported for tab switches. This would be good information to know which tab users are using the most.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -52,119 +62,251 @@ class _DeepLinkPageState extends State<DeepLinkPage>
...
+  final TableView tableView;
+  final DeepLinksController controller;
+  final bool showSpitScreen;
+
+  @override
+  Widget build(BuildContext context) {
+    return Row(
+      children: [
+        Expanded(
+          child: _DataTable(
```
kenzieschmoll, line 154: wrap with `RoundedOutlinedBorder`. To make this look right with the header, you may need to play with the `show*` parameters.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -52,119 +62,251 @@ class _DeepLinkPageState extends State<DeepLinkPage>
...
             linkDatas: linkDatas,
-            bundleByDomain: controller.bundleByDomain,
+            controller: controller,
           ),
         ),
+        if (showSpitScreen)
+          Expanded(
+            child: ValueListenableBuilder<LinkData?>(
+              valueListenable: controller.selectedLink,
+              builder: (context, selectedLink, _) => _ValidationDetailScreen(
```
kenzieschmoll, line 164: wrap this widget with a `RoundedOutlinedBorder` also

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -4,11 +4,21 @@
...
+import '../../shared/screen.dart';
+import '../../shared/table/table.dart';
+import '../../shared/table/table_data.dart';
+import '../../shared/theme.dart';
+import '../../shared/utils.dart';
 import 'deep_links_controller.dart';
 import 'deep_links_model.dart';
 
-const bundledDataRowHeight = 150.0;
+enum TableView {
```
kenzieschmoll, line 17: nit: maybe rename to `TableViewType` to make it more clear that this is a type enum and not a widget? When reading the code below,`TableView` sounded like a widget type

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -52,119 +62,251 @@ class _DeepLinkPageState extends State<DeepLinkPage>
...
+    final ColumnData<LinkData> domain = DomainColumn();
+    final ColumnData<LinkData> path = PathColumn();
 
-    return DataTable(
-      columns: const [
-        DataColumn(label: Text('OS')),
-        DataColumn(label: Text('Scheme')),
-        DataColumn(label: Text('Domain')),
-        DataColumn(label: Text('Path')),
+    return FlatTable(
```
kenzieschmoll, line 191: If you use `SearchableFlatTable` instead of `FlatTable` you should be able to get the search functionality you added above for free. You'll just have to make your `controller` mix in `SearchControllerMixin`

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -52,119 +62,251 @@ class _DeepLinkPageState extends State<DeepLinkPage>
...
-                    Icons.error,
-                    color: Theme.of(context).colorScheme.error,
-                  ),
+          Text(
+            'This tool assistants helps you diagnose Universal Links, App Links,'
+            ' and Custom Schemes in your app. Web check are done for the web association'
+            ' file on your website. App checks are done for the intent filters in'
+            ' the manifest and info.plist file, routing issues, URL format, etc.',
+            style: Theme.of(context).textTheme.bodySmall,
+          ),
```
kenzieschmoll, line 248: Consider adding a `PaddedDivider` widget here before 'Domain check'

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
```diff
@@ -52,119 +62,251 @@ class _DeepLinkPageState extends State<DeepLinkPage>
...
+          const Text('Domain check'),
+          _DomainCheckTable(linkData: linkData),
```
kenzieschmoll, line 250: wrap in Expanded

packages/devtools_app/lib/src/shared/table/table_data.dart
```diff
@@ -79,6 +79,8 @@ abstract class ColumnData<T> {
   /// Get the cell's display value from the given [dataObject].
   String getDisplayValue(T dataObject) =>
       getValue(dataObject)?.toString() ?? '';
+  
+  Widget? getDisplayWidget(T dataObject, BuildContext context) => null ;
```
kenzieschmoll, line 83: Delete this and have your columns implement `ColumnRenderer` instead. Example: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/logging/_kind_column.dart#L13

packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -1486,28 +1486,29 @@ class _TableRowState<T> extends State<TableRow<T>>
...
-          text: TextSpan(
-            text: column.getDisplayValue(node),
-            children: [
-              if (column.getCaption(node) != null)
-                TextSpan(
-                  text: ' ${column.getCaption(node)}',
-                  style: const TextStyle(
-                    fontStyle: FontStyle.italic,
-                  ),
+        content ??= column.getDisplayWidget(node, context) ??
```
kenzieschmoll, line 1489: once you have your columns implement `ColumnRenderer`, please revert changes to this file

kenzieschmoll (commented):
another round of review. Looking good!

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
hangyujin: I tried that and SearchableFlatTable highlights the search result, and I need to filter the result.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_model.dart
hangyujin: In our design mock, icons are aligned together with text like 
5D8DC136-CBC2-4EAC-A327-7C366B0A7817

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
hangyujin: Will update border later according to figma

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
hangyujin: Will update border later according to figma

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
hangyujin: There is no divider in figma so I will follow that.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
hangyujin: is  SearchField paired with  SearchableFlatTable?

hangyujin:
Hi @kenzieschmoll, I addressed most of comments but I am not sure about whether to use SearchableFlatTable and SearchField.

Also can I merge this PR as a version 0 and  consider updating it to use SearchableFlatTable later to unblock Chun-Heng to test his backend changes on top of this simple UI?

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
kenzieschmoll: this comment still stands as well

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
kenzieschmoll: Okay. Generally speaking though @nancyhu2023, does the design file in Figma use the same color scheme / style practices as the remainder of DevTools? Since the Deep Link tool is part of DevTools, I think it is important that we have some consistency with the design patterns used here and those used throughout the remainder of the Devtools tool suite.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
kenzieschmoll: They are not tightly coupled no. They can be used together (like for the CPU profiler Method Table, for example), but you can also use SearchField independently. It just requires a search controller, so you'l make your deep links controller mixin SearchControllerMixin for that. If you look at usages of `SearchField` in DevTools, you'll see several examples of using the `SearchField` widget with a controller that mixes in `SearchControllerMixin`. The logging screen and logging controller are an example of this.

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
hangyujin: I can use DevToolsClearableTextField, but I am still using `FlatTable` not `SearchableFlatTable` as this example is also using `FlatTable`

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_screen.dart
hangyujin: I think I will add analytics in a separated PR

kenzieschmoll (approved):
Looks like you'll need to merge with master and resolve a merge conflict, but lgtm!

Issue #6137, opened 2023-08-01
Dev bump workflow should not run when we are not already on a dev version


kenzieschmoll:
This PR is an example of the dev bump workflow bot behaving poorly right after a release: https://github.com/flutter/devtools/commit/8ec8a22dadaf2d39cf1dabf50aceed3aec26d5a1.

The correct move would have been to go from 2.26.0 to 2.27.0-dev.0, but the workflow mistakenly changed from 2.26.0 to 2.26.0-dev.0.

kenzieschmoll:
another example of an incorrect bump: https://github.com/flutter/devtools/pull/6142

CoderDake:
@kenzieschmoll The assumption is that this step is done manually as part of the release process.
https://github.com/flutter/devtools/blob/master/tool/RELEASE_INSTRUCTIONS.md#update-to-the-next-version

 This can be quickly done by running the command in the above step. 
 
 I can definitely add a check to make sure that nothing happens when we find ourselves in this case though.

kenzieschmoll:
short circuiting the workflow if we find ourself in that state sgtm

PR #6157, opened 2023-08-04
Check for dev bumps to protect release versions

CoderDake:
Fixes https://github.com/flutter/devtools/issues/6137

### DIFF
--- a/.github/workflows/daily-dev-bump.yaml
+++ b/.github/workflows/daily-dev-bump.yaml
@@ -58,6 +58,8 @@ jobs:
 
           ORIGINAL_VERSION=$(dart tool/update_version.dart current-version)
 
+          if [ -z ]
+
           if [ -z "$UPDATE_TYPE" ]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/daily-dev-bump.yaml
```diff
@@ -75,6 +75,15 @@ jobs:
             dart tool/update_version.dart auto --type major
             dart tool/update_version.dart auto --type dev
           elif [ "$UPDATE_TYPE" == "dev" ]; then
+            if ! echo "$ORIGINAL_VERSION" |grep -Eq "\-dev\.[0-9]+" ; then
```
bkonyi, line 78: Nit: can this whitespace be cleaned up a tiny bit to something like this?

`if ! echo "$ORIGINAL_VERSION" | grep -Eq "\-dev\.[0-9]+"; then`

bkonyi (approved):

PR #6179, opened 2023-08-10
Remove warning banner about breakpoints in Chrome 115

elliette:
Now that Chrome 116 is out, remove the warning about the breakpoints regression in Chrome 115.


DanTup:
> Now that Chrome 116 is out

FWIW, I'm still on 115 and Chrome doesn't acknowledge an update. Reading [this page](https://chromereleases.googleblog.com/2023/08/early-stable-update-for-desktop.html) maybe it's only a small percentage of users.

I don't know if this change will make it out before everyone else gets 116 though, so it might be a non-issue :-)

DanTup (approved):

PR #6184, opened 2023-08-11
Fixing broken master workflows

CoderDake:
Fixes https://github.com/flutter/devtools/actions/runs/5825881593/job/15798612424


polina-c (approved):

packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -77,7 +77,7 @@ class ExtensionsManager {
 
       for (final extension in extensions) {
         final config = extension.config;
-        if (config is! Map) {
+        if (config is Map) {
```
elliette, line 80: I don't think we want this  - this means we are now failing when the config is a map (instead of the original intention to fail when the config is a map).

kenzieschmoll: +1 this introduces a bug. Please change back to `is!`.

PR #6185, opened 2023-08-11
Ignore DCM error for type assertion

elliette:
Follow up to https://github.com/flutter/devtools/pull/6184

Address https://github.com/flutter/devtools/pull/6184#discussion_r1291568933


incendial:
@elliette thanks for the mention, looks like a false-positive, will check.

incendial:
@elliette which dcm version you have (dcm --version)? I can't reproduce it with the latest one.

incendial:
Hm, I see it only under one condition: if I comment out the 'package:extension_discovery/extension_discovery.dart' (so, basically, if the `Extension` is not resolved). Do you run `flutter pub get` before dcm checks? Also I've noticed that for me `sh tool/pull_and_refresh.sh` did not install `extension_discovery`, only running `flutter pub get` manually helped.

elliette:
> Hm, I see it only under one condition: if I comment out the 'package:extension_discovery/extension_discovery.dart' (so, basically, if the `Extension` is not resolved). Do you run `flutter pub get` before dcm checks? Also I've noticed that for me `sh tool/pull_and_refresh.sh` did not install `extension_discovery`, only running `flutter pub get` manually helped.

Thanks! Looks like we weren't - I've opened https://github.com/flutter/devtools/pull/6204 and will try re-running the workflow after that has been submitted.

CoderDake (approved):

Issue #6208, opened 2023-08-17
Create and publish `package:devtools_app_shared`


kenzieschmoll:
This package will include shared components (service management, theme, ui, utilities, etc.) for use in `package:devtools_app` and DevTools extensions.

- [x] Create the package - https://github.com/flutter/devtools/pull/6210
- [x] Clean up the README, CHANGELOG.md, and `example` directory.
  - https://github.com/flutter/devtools/pull/6218
  - https://github.com/flutter/devtools/pull/6230
- [x] Audit the exposed libraries (particularly the service management libraries and the theme)
  - https://github.com/flutter/devtools/pull/6235
  - https://github.com/flutter/devtools/pull/6262
- [x] include the font assets from devtools_app
  - https://github.com/flutter/devtools/pull/6265
- [x] Publish on pub.dev
  - https://pub.dev/packages/devtools_app_shared

PR #6210, opened 2023-08-17
Create `package:devtools_app_shared`

kenzieschmoll:
Work towards https://github.com/flutter/devtools/issues/6208.

This is a refactor to pull code out from `package:devtools_app` that will be shared and used by both DevTools and DevTools extensions.

`package:devtools_app_shared` exports:
- `service_extensions.dart` (exported separately so that the `as extensions` prefix can be used easily where this library is imported)
- `service.dart` (service management libraries `service_manager.dart`, which the DevTools ServiceConnectionManager extends with any devtools-specific functionality that we shouldn't expose, `connected_app.dart`, `isolate_manager.dart`, `service_extension_manager.dart`, `eval_on_dart_library.dart`, etc.)
- `ui.dart` (common UI elements like standard DevTools buttons, splitters, headers, dialogs, and theme)
- `utils.dart` (globals framework, auto dispose, list utilities and other utils)

This package is under construction and the remaining work before publishing is tracked on  https://github.com/flutter/devtools/issues/6208.


polina-c:
PACKAGES.md in the root that shows how packages depend on each other, will help to understand things faster.
Example: https://github.com/dart-lang/leak_tracker/blob/main/pkgs/leak_tracker/lib/DEPENDENCIES.md

polina-c:
lgtm

polina-c (approved):

PR #6216, opened 2023-08-18
Extensions cleanup and TODOs

kenzieschmoll:
This uses the service manager from `devtools_app_shared` in the extensions template and cleans up / adds some todos. Also reverts an accidental flag flip of the extensions feature flag.


packages/devtools_extensions/lib/src/template/extension_manager.dart
```diff
@@ -58,7 +56,7 @@ class ExtensionManager {
             break;
           case DevToolsExtensionEventType.vmServiceConnection:
             final vmServiceUri = extensionEvent.data?['uri'] as String?;
-            unawaited(appManager.connectToVmService(vmServiceUri));
+            unawaited(connectToVmService(vmServiceUri));
```
elliette, line 59: I know this `unawaited` was already here, but would be good to catch any errors in case connecting to the vm service throws one:

`unawaited(connectToVmService(vmServiceUri).catchError(...))`

elliette (approved):
One suggestion otherwise LGTM!

PR #6230, opened 2023-08-21
Add an `example` directory to `devtools_app_shared`

kenzieschmoll:
work towards https://github.com/flutter/devtools/issues/6208


packages/devtools_app_shared/example/ui_example.dart
```diff
@@ -0,0 +1,33 @@
...
+
+import 'package:devtools_app_shared/ui.dart';
+import 'package:flutter/material.dart';
+
+class ExampleWidget extends StatelessWidget {
+  const ExampleWidget({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    // Shared component
```
bkonyi, line 13: Maybe it's worth doing `import 'package:devtools_app_shared/ui.dart' as devtools;` to make it clear what widgets are from DevTools?

packages/devtools_app_shared/example/utils_example.dart
```diff
@@ -0,0 +1,72 @@
...
+  /// Example: setting and accessing globals.
+  setAndAccessGlobal();
+
+  /// Example: [ListValueNotifier]
+  useListValueNotifier();
+}
+
+MyCoolClass get coolClass => globals[MyCoolClass] as MyCoolClass;
+
+void setAndAccessGlobal() {
```
bkonyi, line 18: Consider adding some documentation to each of the use cases in this file describing:

- What functionality is being exercised
- Why it's available
- When a developer might want to use it

I'd also consider keeping each of these examples in their own self-contained libraries to keep things as simple as possible. Ideally each example could be executed standalone and clearly demonstrate the expected behaviors of the APIs.

kenzieschmoll: I split up the utils examples into individual files. Left the ui and service as they are.

bkonyi (approved):
Looks great!

PR #6235, opened 2023-08-22
Aggressively seal classes in `package:devtools_app_shared`

kenzieschmoll:
`devtools_app_shared` will be published on pub and will be used by DevTools extension authors to build extensions. In order to limit breaking changes to DevTools extensions, we would like to "seal" critical classes related to interacting with a connected app's vm service.

This PR aggressively marks service-related classes in `devtools_app_shared` as `final`, where possible. In cases where marking classes `final` would prohibit us from being able to effectively write tests, we use the `@sealed` annotation from `package:meta`.

Looking for:
- early feedback on these changes before I add a large commit full of refactors to support this change.
- feedback on whether we should seal other classes such as utilities (e.g. auto dispose functionality) and common ui components?

@bkonyi @jacob314

### DIFF
--- a/packages/devtools_app/test/shared/instance_viewer_test.dart
+++ b/packages/devtools_app/test/shared/instance_viewer_test.dart
@@ -1539,4 +1539,5 @@ void main() {
   });
 }
 
+// ignore: subtype_of_sealed_class, fake for testing.
 class FakeEvalOnDartLibrary extends Fake implements EvalOnDartLibrary {}
--- a/packages/devtools_app_shared/lib/src/service/connected_app.dart
+++ b/packages/devtools_app_shared/lib/src/service/connected_app.dart
@@ -5,6 +5,7 @@
 import 'dart:async';
 
 import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
 
 import 'eval_on_dart_library.dart';
 import 'flutter_version.dart';
@@ -15,6 +16,7 @@ final _log = Logger('connected_app');
 const flutterLibraryUri = 'package:flutter/src/widgets/binding.dart';
 const dartHtmlLibraryUri = 'dart:html';
 
+@sealed
 class ConnectedApp {
   ConnectedApp(this.serviceManager);
 
--- a/packages/devtools_app_shared/lib/src/service/eval_on_dart_library.dart
+++ b/packages/devtools_app_shared/lib/src/service/eval_on_dart_library.dart
@@ -11,6 +11,7 @@ import 'dart:math';
 
 import 'package:flutter/foundation.dart';
 import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
 import 'package:vm_service/vm_service.dart' hide Error;
 import 'package:vm_service/vm_service.dart' as vm_service;
 
@@ -29,6 +30,7 @@ class Disposable {
   }
 }
 
+@sealed
 class EvalOnDartLibrary extends DisposableController
     with AutoDisposeControllerMixin {
   EvalOnDartLibrary(
@@ -636,7 +638,7 @@ class EvalOnDartLibrary extends DisposableController
   }
 }
 
-class LibraryNotFound implements Exception {
+final class LibraryNotFound implements Exception {
   LibraryNotFound(this.name);
 
   final String name;
@@ -644,7 +646,7 @@ class LibraryNotFound implements Exception {
   String get message => 'Library matchining $name not found';
 }
 
-class FutureFailedException implements Exception {
+final class FutureFailedException implements Exception {
   FutureFailedException(this.expression, this.errorRef, this.stacktraceRef);
 
   final String expression;
@@ -657,9 +659,9 @@ class FutureFailedException implements Exception {
   }
 }
 
-class CancelledException implements Exception {}
+final class CancelledException implements Exception {}
 
-class UnknownEvalException implements Exception {
+final class UnknownEvalException implements Exception {
   UnknownEvalException({
     required this.expression,
     required this.scope,
@@ -676,7 +678,7 @@ class UnknownEvalException implements Exception {
   }
 }
 
-class SentinelException implements Exception {
+final class SentinelException implements Exception {
   SentinelException(this.sentinel);
 
   final Sentinel sentinel;
@@ -687,7 +689,7 @@ class SentinelException implements Exception {
   }
 }
 
-class EvalSentinelException extends SentinelException {
+final class EvalSentinelException extends SentinelException {
   EvalSentinelException({
     required this.expression,
     required this.scope,
@@ -703,7 +705,7 @@ class EvalSentinelException extends SentinelException {
   }
 }
 
-class EvalErrorException implements Exception {
+final class EvalErrorException implements Exception {
   EvalErrorException({
     required this.expression,
     required this.scope,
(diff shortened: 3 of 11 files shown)


bkonyi:
It feels weird using `@sealed` and `final` in the same file, but I think this is the right thing to do for testability.

I think we should be very aggressive with marking anything we expose through this package as `final` or some variation of `final`. We can always relax restrictions if we have a reason to, but once we release we can't tighten them.

packages/devtools_app/test/shared/instance_viewer_test.dart
```diff
@@ -1539,4 +1539,5 @@ void main() {
   });
 }
 
+// ignore: subtype_of_sealed_class, fake for testing.
```
jacob314, line 1542: I would avoid making changes that require ignoring analyzer warnings.

packages/devtools_app_shared/lib/src/service/eval_on_dart_library.dart
```diff
@@ -29,6 +30,7 @@ class Disposable {
   }
 }
 
+@sealed
```
jacob314, line 33: Avoid the legacy annotations that are from before Dart supported proper class modifiers. If this was to be sealed, you would use the `sealed` keyword. In practice I would just avoid making this `sealed` although you might consider making it `final`.
https://dart.dev/language/class-modifiers

kenzieschmoll: Discussed offline. `@sealed` seems like the cleanest solution to protect users from breaking changes (by warning then when they extend or implement a class they shouldn't), while not doing something hacky like including the test fake in the library with the final class it implements.

packages/devtools_test/lib/src/mocks/generated_mocks_factories.dart
```diff
@@ -161,6 +161,13 @@ MockVmServiceWrapper createMockVmServiceWrapperWithDefaults() {
   return service;
 }
 
+MockServiceConnectionManager createMockServiceConnectionWithDefaults() {
```
kenzieschmoll, line 164: new code

packages/devtools_test/lib/src/mocks/fake_service_manager.dart
```diff
@@ -20,35 +20,103 @@ import 'fake_vm_service_wrapper.dart';
 import 'generated.mocks.dart';
 import 'mocks.dart';
 
-class FakeServiceManager extends Fake implements ServiceConnectionManager {
-  FakeServiceManager({
+class FakeServiceConnectionManager extends Fake
```
kenzieschmoll, line 23: this file should be reviewed

packages/devtools_app_shared/lib/src/service/service_manager.dart
```diff
@@ -7,6 +7,7 @@ import 'dart:core';
 
 import 'package:flutter/foundation.dart';
 import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
```
kenzieschmoll, line 10: this file should be reviewed

packages/devtools_app_shared/lib/src/service/connected_app.dart
```diff
@@ -188,3 +193,40 @@ class ConnectedApp {
           flutterVersionKey: flutterVersionNow!.version,
       };
 }
+
+final class OfflineConnectedApp extends ConnectedApp {
```
kenzieschmoll, line 197: moved from package:devtools_app and marked final

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -150,7 +196,8 @@ class ServiceConnectionManager extends ServiceManager<VmServiceWrapper> {
   ///
   /// Throws an Exception if no 'FlutterView' is present in this isolate.
   Future<String> get flutterViewId async {
```
jacob314, line 198: superficially extensions like this seem equally useful for devtools extensions as for devtools proper.

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -131,17 +152,42 @@ class ServiceConnectionManager extends ServiceManager<VmServiceWrapper> {
...
+      serviceManager.isolateManager
+          .selectIsolate(serviceManager.isolateManager.isolates.value.first);
+    }
+    await serviceManager.isolateManager.init(isolates);
   }
 
-  @override
-  List<IsolateRef> isolatesFromVm(VM? vm) {
-    return vm?.isolatesForDevToolsMode() ?? <IsolateRef>[];
+  Future<String?> rootLibraryForMainIsolate() async {
```
jacob314, line 169: this seems like something an extension might want to know. how do you draw the line on whether to include it in the shared code or not?

jacob314:
> It feels weird using `@sealed` and `final` in the same file, but I think this is the right thing to do for testability.
> 
> I think we should be very aggressive with marking anything we expose through this package as `final` or some variation of `final`. We can always relax restrictions if we have a reason to, but once we release we can't tighten them.

Turns out there is an alternative to using `@sealed` that still allows reopening classes for testing purposes only with only a couple lines of boilerplate. 
https://github.com/dart-lang/language/issues/3106#issuecomment-1593145359
Given that, I believe we should deprecate `@sealed`  to avoid confusion and point users to  this technique.

packages/devtools_app_shared/lib/src/service/eval_on_dart_library.dart
kenzieschmoll: https://github.com/flutter/devtools/issues/6239

packages/devtools_app/lib/src/service/service_manager.dart
kenzieschmoll: Agreed we may want to expose this. For now, I'd like to keep the API small and add more to `devtools_app_shared` as it makes sense. At this point I am exposing everything needed for the package:provider extension as a customer zero, and will expose more as we get a better idea of what more initial customers need.

packages/devtools_app/lib/src/service/service_manager.dart
kenzieschmoll: Agreed we may want to expose this. For now, I'd like to keep the API small and add more to `devtools_app_shared` as it makes sense.

bkonyi (approved):
LGTM, but we should probably get another +1 as I had a hard time reviewing this thoroughly (Chrome + GitHub are thrashing hard when I open the "Files changed" tab).

bkonyi:
Also, do we plan on using the solution from https://github.com/dart-lang/language/issues/3106#issuecomment-1593145359 or landing with `@sealed` for now?

kenzieschmoll:
> Also, do we plan on using the solution from [dart-lang/language#3106 (comment)](https://github.com/dart-lang/language/issues/3106#issuecomment-1593145359) or landing with `@sealed` for now?

In this comment here, I explain why we need to use `@sealed` right now for classes we need to mock and not fake. I also filed https://github.com/flutter/devtools/issues/6239 to track removing `@sealed` if we come up with a better solution or decide to rewrite our tests to not use mocks for these "sealed" classes.

CoderDake (changes requested):
Seems alright to me, but it was definitely a lot to review :P

PR #6237, opened 2023-08-23
Add events to receive debug sessions into sidebar

DanTup:
This adds some events for listing debug sessions in the sidebar and buttons to open DevTools features from them.

As before, this is functional but the UI needs much work. It also needs tests but since this work is largely UI, I think it'll be better to add them once the UI is a little more complete.


packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
```diff
@@ -78,7 +139,9 @@ abstract interface class VsCodeDevicesEvent {
 abstract interface class VsCodeCapabilities {
   bool get executeCommand;
   bool get selectDevice;
+  bool get openDevToolsFeature;
```
kenzieschmoll, line 142: can we change this to "openDevToolsPage"? A "feature" makes me think of some item within a page (which has a set of features), but not the page itself.

DanTup: Yeah, I'll change this. I can't recall why I avoided page, but we use page elsewhere to mean the same thing, so we should make this match.

If we add support for jumping to something more specific in future, we'll probably need changes to the API anyway (and probably a new method would be the most compatible way to do that).

DanTup: Done!

packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
```diff
@@ -55,6 +63,47 @@ abstract interface class VsCodeDevice {
...
+  String get name;
+  String? get vmServiceUri;
+
+  /// The mode the app is running in.
+  ///
+  /// These values are defined by Flutter and at the time of writing can include
+  /// 'debug', 'profile', 'release' and 'jit_release'.
+  ///
+  /// This value may be unavailable (`null`) for older SDKs or for Dart/Test
+  /// sessions.
```
kenzieschmoll, line 83: do we need to worry about this since the version of DevTools that has the vs code sidebar support will never be an "older" SDK?

packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
```diff
@@ -78,7 +139,9 @@ abstract interface class VsCodeDevicesEvent {
 abstract interface class VsCodeCapabilities {
   bool get executeCommand;
   bool get selectDevice;
+  bool get openDevToolsPage;
```
kenzieschmoll, line 142: can we add dart doc for each of these?

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -0,0 +1,112 @@
...
+    // final isRelease = mode == 'release' || mode == 'jit_release';
+    final isFlutter = session.debuggerType?.contains('Flutter') ?? false;
+
+    return TableRow(
+      children: [
+        Text(
+          '$name ($mode)',
+          style: Theme.of(context).textTheme.titleSmall,
+        ),
+        if (api.capabilities.openDevToolsPage) ...[
```
kenzieschmoll, line 60: also, when we change this to a dropdown we should match the screen order in the DevTools tab bar

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -0,0 +1,112 @@
...
+          ),
+          _devToolsButton(
+            session,
+            ScreenMetaData.network,
+            enabled: isDebug,
+          ),
+          _devToolsButton(
+            session,
+            ScreenMetaData.logging,
+          ),
```
kenzieschmoll, line 90: missing the app size screen

packages/devtools_app/lib/src/standalone_ui/vs_code/devices.dart
```diff
@@ -0,0 +1,72 @@
...
+    super.key,
+  });
+
+  final VsCodeApi api;
+  final List<VsCodeDevice> devices;
+  final String? selectedDeviceId;
+
+  @override
+  Widget build(BuildContext context) {
+    return Column(
```
kenzieschmoll, line 25: we likely want to change this to use a table / list with selectable rows for reference in the next ui iteration

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
```diff
@@ -64,52 +66,59 @@ class _VsCodeConnectedPanelState extends State<_VsCodeConnectedPanel> {
...
-        const SizedBox(height: defaultSpacing),
-        if (widget.api.capabilities.executeCommand)
-          ElevatedButton(
-            onPressed: () =>
-                unawaited(widget.api.executeCommand('flutter.createProject')),
-            child: const Text('New Flutter Project'),
-          ),
-        if (widget.api.capabilities.selectDevice)
+    return Padding(
+      padding: const EdgeInsets.all(8.0),
```
kenzieschmoll, line 70: use the `denseSpacing` constant from devtools_app_shared/ui.dart

packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
```diff
@@ -12,8 +12,10 @@ abstract interface class VsCodeApi {
   VsCodeCapabilities get capabilities;
   Future<void> initialize();
   Stream<VsCodeDevicesEvent> get devicesChanged;
+  Stream<VsCodeDebugSessionsEvent> get debugSessionsChanged;
   Future<Object?> executeCommand(String command, [List<Object?>? arguments]);
   Future<bool> selectDevice(String id);
+  Future<void> openDevToolsPage(String debugSessionId, String page);
```
kenzieschmoll, line 18: can you add dart doc for these?

packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
```diff
@@ -69,6 +118,18 @@ abstract interface class VsCodeDevicesEvent {
...
 }
 
+/// This class defines a debug session event sent by the Dart/Flutter extensions
+/// in VS Code (and must match the implementation there).
+///
+/// All changes to this file should be backwards-compatible and use
+/// [VsCodeCapabilities] to advertise which capabilities are available and
+/// handle any changes in behaviour.
+abstract interface class VsCodeDebugSessionsEvent {
+  List<VsCodeDebugSession> get sessions;
```
kenzieschmoll, line 128: dart doc

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
```diff
@@ -64,52 +66,59 @@ class _VsCodeConnectedPanelState extends State<_VsCodeConnectedPanel> {
...
-      ],
+          const SizedBox(height: defaultSpacing),
+          if (widget.api.capabilities.selectDevice)
+            StreamBuilder(
+              stream: widget.api.devicesChanged,
+              builder: (context, snapshot) {
+                final devices = snapshot.data?.devices ?? const [];
+                final selectedDeviceId = snapshot.data?.selectedDeviceId;
+                return SizedBox(
+                  width: double.infinity,
```
kenzieschmoll, line 111: what happens without the sized box with infinite width, here and above?

packages/devtools_app/test/test_infra/test_data/dart_tooling_api/mock_api.dart
```diff
@@ -136,6 +144,9 @@ class MockDartToolingApi extends DartToolingApiImpl {
     return true;
   }
 
+  /// Simulates opening a DevTools feature.
+  Future<void> openDevToolsPage(json_rpc_2.Parameters parameters) async {}
```
kenzieschmoll, line 148: should we at least log something here?

packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
DanTup: Probably not, I've updated the comment to remove this (but include that it's not available before a session has started since the mode comes in the `flutter.appStarted` event).

packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
DanTup: Done!

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
DanTup: Added a TODO

packages/devtools_app/lib/src/standalone_ui/vs_code/devices.dart
DanTup: We're already using a table (this column is to put the heading above it). I've added a TODO (further down in `_createDeviceRow`) about using a highlighted row instead of label.

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
DanTup: Done!

packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
DanTup: Done!

packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
DanTup: Done!

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
DanTup: The widgets inside don't fill the width and are center aligned without it. I don't know if this is the right/preferred fix though :-)

Screenshot 2023-09-06 122629


Screenshot 2023-09-06 122700

packages/devtools_app/test/test_infra/test_data/dart_tooling_api/mock_api.dart
DanTup: We can, although it feels redundant - the request/response will already appear in the log so you can see the request was successful. We could log additional events like "Opened DevTools page foo" but for consistency we'd probably then log the other commands too and it could just result in the log having extra lines for the same things (and it's probably the JSON that is most useful).

packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
```diff
@@ -9,12 +9,46 @@
...
+  /// should take care when calling APIs that might evolve over time.
   Future<Object?> executeCommand(String command, [List<Object?>? arguments]);
+
+  /// Changes the current Flutter device.
+  ///
+  /// The selected device is the same one shown in the status bar in VS Code.
+  /// Calling this API will update the device for the whole VS Code extension.
   Future<bool> selectDevice(String id);
+
+  /// Opens a specific DevTools [page] for the debug session with ID
```
kenzieschmoll, line 50: Does this open the DevTools page in the embedded vs code window or in the browser? could be good to specify here.

packages/devtools_app/test/test_infra/test_data/dart_tooling_api/mock_api.dart
kenzieschmoll: if the request response are already logged, then that is sufficient 👍

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
kenzieschmoll: Ah, looks like we just need to set `crossAxisAlignment: CrossAxisAlignment.start` on the parent column, and then we can remove the infinitely sized boxes

kenzieschmoll (approved):
a couple more comments but lgtm

packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
DanTup: It could be either depending on user settings. I've updated the comment to note this.

PR #6238, opened 2023-08-23
Include new packages into upgrade process.

polina-c:


tool/upgrade.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+# Copyright 2023 The Chromium Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
```
CoderDake, line 5: To make this agnostic to the CWD you can use something like:
```
SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
ROOT_DIR="$SCRIPT_DIR/.."
```
Then use $ROOT_DIR for your pushds

CoderDake (approved):

PR #6245, opened 2023-08-24
Move `OutlineDecoration` to devtools_app_shared

kenzieschmoll:
This will allow this widget to be used by the devtools extension template and by extensions themselves


bkonyi (approved):
LGTM with some documentation added.

PR #6250, opened 2023-08-25
Add `DevToolsExtensionHostInterface` and clean up some service management for extensions

kenzieschmoll:
I will follow up with a PR to create a `_SimulatedDevToolsController` that implements the host interface (which will allow for a simulated environment for extension development)

### DIFF
--- a/packages/devtools_app/lib/src/extensions/embedded/_view_web.dart
+++ b/packages/devtools_app/lib/src/extensions/embedded/_view_web.dart
@@ -48,7 +48,8 @@ class _EmbeddedExtensionState extends State<EmbeddedExtension> {
 }
 
 class _ExtensionIFrameController extends DisposableController
-    with AutoDisposeControllerMixin {
+    with AutoDisposeControllerMixin
+    implements DevToolsExtensionHostInterface {
   _ExtensionIFrameController(this.embeddedExtensionController);
 
   final EmbeddedExtensionControllerImpl embeddedExtensionController;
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/extensions/embedded/_view_web.dart
```diff
@@ -181,4 +165,42 @@ class _ExtensionIFrameController extends DisposableController
...
     super.dispose();
   }
+
+  @override
+  void ping() {
+    _postMessage(DevToolsExtensionEvent(DevToolsExtensionEventType.ping));
+  }
+
+  @override
+  void vmServiceConnectionChanged({String? uri}) {
```
bkonyi, line 175: Should `uri` be non-nullable and required?

packages/devtools_extensions/lib/src/api/api.dart
```diff
@@ -29,3 +31,27 @@ enum DevToolsExtensionEventType {
...
 }
+
+/// Interface that a DevTools extension host should implement.
+/// 
+/// This interface is implemented by DevTools itself as well as by a simulated
+/// DevTools environment for simplifying extension development.
+abstract interface class DevToolsExtensionHostInterface {
+  /// This method should send a [DevToolsExtensionEventType.ping] event to the
+  /// DevTools extension to check that it is ready.
+  void ping();
```
bkonyi, line 42: It seems strange to have a `ping()` method here that the developer has to implement. Should it be part of this interface, or should we also provide a `postMessage(...)` interface and implement `ping()` using it?

kenzieschmoll: The extension developer shouldn't be implementing this interface. DevTools and the Simulated DevTools environment will implement this interface. DevTools extensions send a 'pong' event in response to a 'ping' event from the host. This is necessary to ensure the extension (normally in an embedded iFrame) is ready for interaction with the host. 

> should we also provide a postMessage(...) interface and implement ping() using it?

I'm not sure I understand this. What value would it add to do this? Also, the two implementation of this interface (DevTools and the Simulated DevTools environment https://github.com/flutter/devtools/pull/6251) need a different implementation of ping, so they wouldn't be able to share this from the interface.

bkonyi: If `ping()` could have a common implementation, we'd want to have `postMessage(...)` added to the interface so `ping()` could be implemented in terms of that. If this isn't going to be implemented by extension developers, I don't think it matters.

bkonyi (approved):

PR #6251, opened 2023-08-25
Add a simulated DevTools environment for developing extensions

kenzieschmoll:
The simulated environment allows the extension to be developed outside of an embedded iFrame environment. It implements the same `DevToolsExtensionHostInterface` that DevTools does so that:
- the extension can connect to a vm service uri
- simulated events can be sent from DevTools to the extension
- the user can see a log of post messages that go back and forth between the extension and devtools

Screenshot 2023-08-25 at 1 37 09 PM

Depends on https://github.com/flutter/devtools/pull/6250 landing first.

### DIFF
--- a/packages/devtools_app_shared/lib/src/ui/common.dart
+++ b/packages/devtools_app_shared/lib/src/ui/common.dart
@@ -688,7 +688,7 @@ Widget maybeWrapWithTooltip({
 }
 
 /// Displays a [json] map as selectable, formatted text.
-class FormattedJson extends StatelessWidget {
+final class FormattedJson extends StatelessWidget {
   const FormattedJson({
     super.key,
     this.json,
--- a/packages/devtools_app_shared/lib/src/ui/common.dart
+++ b/packages/devtools_app_shared/lib/src/ui/common.dart
@@ -713,3 +713,34 @@ final class FormattedJson extends StatelessWidget {
     );
   }
 }
+
+/// An extension on [ScrollController] to facilitate having the scrolling widget
+/// auto scroll to the bottom on new content.
+extension ScrollControllerAutoScroll on ScrollController {
+// TODO(devoncarew): We lose dock-to-bottom when we receive content when we're
+// off screen.
+
+  /// Return whether the view is currently scrolled to the bottom.
+  bool get atScrollBottom {
+    final pos = position;
+    return pos.pixels == pos.maxScrollExtent;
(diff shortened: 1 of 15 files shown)


polina-c:
Do we have a doc page under construction that explains how to author extensions?
It would be helpful to create a draft and fine tune it with every such PR, as it will help reviewers to understand the change better and, as side effect, we will have ready to publish guidance.

packages/devtools_app_shared/lib/src/ui/common.dart
```diff
@@ -713,3 +713,34 @@ final class FormattedJson extends StatelessWidget {
     );
   }
 }
+
+/// An extension on [ScrollController] to facilitate having the scrolling widget
+/// auto scroll to the bottom on new content.
+extension ScrollControllerAutoScroll on ScrollController {
```
polina-c, line 719: cool!

packages/devtools_extensions/lib/src/template/_simulated_devtools_environment/_simulated_devtools_environment.dart
```diff
@@ -29,25 +31,40 @@ class SimulatedDevToolsWrapper extends StatefulWidget {
...
   State<SimulatedDevToolsWrapper> createState() =>
       _SimulatedDevToolsWrapperState();
 }
 
-class _SimulatedDevToolsWrapperState extends State<SimulatedDevToolsWrapper> {
+class _SimulatedDevToolsWrapperState extends State<SimulatedDevToolsWrapper>
+    with AutoDisposeMixin {
   late final _SimulatedDevToolsController simController;
 
+  late ConnectedState connectionState;
```
polina-c, line 47: It seems connectionState should be final. Or, we should initialize it in something like 'disWidgetChange'

polina-c (approved):

kenzieschmoll:
> Do we have a doc page under construction that explains how to author extensions?

Working on that now. Follow up PR coming soon :)

packages/devtools_extensions/lib/src/template/_simulated_devtools_environment/_simulated_devtools_environment.dart
kenzieschmoll: this is late because it is initialized in `initState` and it cannot be final because we update it in a listener.

Issue #5973, opened 2023-06-29
Network tab freezes VSCode when displaying base64 image encoded strings


TiffApps:
Hi!

I was checking out the network tab to inspect some request/response objects (just opening the fields here and there to read inside), when at some point, VSCode froze, prompting me to Re-open, close, or wait.
I clicked wait the first time, then re-open the 2nd time.
It closed the window, opened the small "Updating VSCode window" for 10sec, before not re-opening VSCode after 2min.

I clicked on the VSCode icon to launch it, it worked right away, I waited a minute so everything can get initialized, but there was a notification about the terminal not closing correctly and something like "pty needs to restart". I clicked on restart, it wrote "restarting terminal etc" in blue in my terminal, and then the notification popped up again, so I repeated the operation 3 times before giving up and going into Terminal > Kill Terminal > New Terminal, but it immediately closes again as soon as it starts.

On the devtools side, I don't see the "Dart: Open the Network DevTools tab" anymore at all: it just disappeared from my IDE. Here is what I see:
![image](https://github.com/flutter/devtools/assets/33839996/ef61a054-edf0-4c10-88fe-bd7816b580ec)
("Dart: Open DevTools" being the one in the web browser, not in VSCode)

Further VSCode restarts didn't help, and I haven't tried restarting my PC yet in case it erases some internal VSCode logs that could be useful to know more about this issue.

As of right now, it means:
- I can't use VSCode's terminal anymore
- I can't use built-in DevTools anymore
Of course I expect a PC restart to fix both issues, but I'll just wait for your answer before trying.

Let me know if you need more details on any step, I'll gladly provide everything you need!

## Versions
OS: **Windows 10 22H2**
Flutter: **3.10.5**
Dart: **3.0.5**
DevTools: **2.23.1**
VSCode: **1.79.2**

DanTup:
> On the devtools side, I don't see the "Dart: Open the Network DevTools tab" anymore at all: it just disappeared from my IDE. Here is what I see:

The commands to open specific pages only appear when you have an active debug session (because they can only be loaded with a VM Service connection). You'll only see the generic "Open DevTools" command if you don't have a session. It's not clear if you had one open or not?


> I can't use VSCode's terminal anymore

[...]

PR #6254, opened 2023-08-25
Limit Network text view to 2^16 characters to prevent it from crashing

CoderDake:
For images, the request and response bodies can get large which cause the Text view to hang DevTools.

I've hardcoded the 2^16 character limit for now, but I could be persuaded to allow an override in case other textviews have different needs.

Fixes https://github.com/flutter/devtools/issues/5973 

### Truncated
Screenshot 2023-08-25 at 3 47 54 PM
### Under Character Limit

Screenshot 2023-08-25 at 3 47 58 PM

### DIFF
--- a/packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
+++ b/packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
@@ -156,13 +156,10 @@ class HttpRequestView extends StatelessWidget {
             ? requestContentType.any((element) => element.contains('json'))
             : requestContentType.contains('json');
 
-        Widget child;
+        Widget child; 
         child = isJson
             ? JsonViewer(encodedJson: data.requestBody!)
-            : Text(
-                data.requestBody!,
-                style: theme.fixedFontStyle,
-              );
+            : TextViewer(text: data.requestBody!);
         return Padding(
           padding: const EdgeInsets.all(denseSpacing),
           child: SingleChildScrollView(
@@ -356,10 +353,7 @@ class HttpTextResponseViewer extends StatelessWidget {
 
         return switch (currentLocalResponseType) {
           NetworkResponseViewType.json => JsonViewer(encodedJson: responseBody),
-          NetworkResponseViewType.text => Text(
-              responseBody,
-              style: textStyle,
-            ),
+          NetworkResponseViewType.text => TextViewer(text: responseBody),
           _ => const SizedBox(),
         };
       },
--- a/packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
+++ b/packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
@@ -142,6 +142,7 @@ class HttpRequestView extends StatelessWidget {
     return ValueListenableBuilder(
       valueListenable: data.requestUpdatedNotifier,
       builder: (context, __, ___) {
+        final theme = Theme.of(context);
         final requestHeaders = data.requestHeaders;
         final requestContentType = requestHeaders?['content-type'] ?? '';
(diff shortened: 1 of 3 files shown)


bkonyi:
I'm fine with this for now, but I think this warrants further investigation as to why it causes a hang. We shouldn't have an issue rendering 2^16 characters. Can you file an issue to follow up on this?

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1264,6 +1264,30 @@ class JsonViewer extends StatefulWidget {
   State<JsonViewer> createState() => _JsonViewerState();
 }
 
+class TextViewer extends StatelessWidget {
+  const TextViewer({
+    super.key,
+    required this.text,
+  });
+  final String text;
```
bkonyi, line 1272: This needs some whitespace.

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1264,6 +1264,30 @@ class JsonViewer extends StatefulWidget {
   State<JsonViewer> createState() => _JsonViewerState();
 }
 
+class TextViewer extends StatelessWidget {
```
bkonyi, line 1267: This needs some documentation. `TextViewer` is very generic without context.

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1264,6 +1264,30 @@ class JsonViewer extends StatefulWidget {
...
+    super.key,
+    required this.text,
+  });
+  final String text;
+  static const _maxLength = 65536; //2^16
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+    final String displayText;
+    // Limit the length of the displayed text to _MAX_LENGTH
```
bkonyi, line 1278: We should reference the issue here.

CoderDake: I'll reference it in the network code where we reference the TextViews.

packages/devtools_app/lib/src/shared/common_widgets.dart
```diff
@@ -1264,6 +1264,36 @@ class JsonViewer extends StatefulWidget {
...
+  });
+
+  final String text;
+  final int maxLength;
+  final TextStyle? style;
+
+  @override
+  Widget build(BuildContext context) {
+    final String displayText;
+    // Limit the length of the displayed text to _MAX_LENGTH
```
bkonyi, line 1284: s/`_MAX_LENGTH`/`maxLength`/

packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
```diff
@@ -356,8 +358,10 @@ class HttpTextResponseViewer extends StatelessWidget {
 
         return switch (currentLocalResponseType) {
           NetworkResponseViewType.json => JsonViewer(encodedJson: responseBody),
-          NetworkResponseViewType.text => Text(
-              responseBody,
+          // Using TextViewer fixes
+          // https://github.com/flutter/devtools/issues/5973
```
bkonyi, line 362: The above issue should be closed when this PR lands, but I'd file a new issue for this as it's a bandaid fix if we don't know why we're seeing the hang in the first place.

CoderDake: Created the following issues in response to this discussion:
- https://github.com/flutter/devtools/issues/6261
- https://github.com/flutter/devtools/issues/6263

bkonyi (approved):

PR #6255, opened 2023-08-25
Add initial README.md for `package:devtools_extensions`

kenzieschmoll:
There are a few TODOs in here that I will address as we continue to iterate on this package.


polina-c (approved):
How about making headers more consistent.  
Right now some are orders, some are infinitives and some are questions.

I've read somewhere that verbs are received better in instructions. Like this:

Setup your package...
Build a DevTools extension
Create initial stricture for your source code
Utilize helper packages
Create the extension web app

and so on

PR #6275, opened 2023-08-28
Add hot reload and hot restart buttons to the simulated devtools environment

kenzieschmoll:
This will allow extension developers to easily hot reload or hot restart the connected app:
Screenshot 2023-08-28 at 4 10 30 PM

### DIFF
--- a/packages/devtools_extensions/CHANGELOG.md
+++ b/packages/devtools_extensions/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.0.3-wip
+* Add hot reload and hot restart actions to the simulated DevTools environment.
+
 ## 0.0.2
 * Add a simulated DevTools environment that for easier development.
 * Add a `build_and_copy` command to build a devtools extension and copy the output to the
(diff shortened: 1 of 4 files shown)


packages/devtools_extensions/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 0.0.3-wip
```
polina-c, line 1: Why not dev? What is the difference?

polina-c (approved):

packages/devtools_extensions/CHANGELOG.md
kenzieschmoll: I believe wip indicates a package that is not yet published, but I don't think it is enforced. 
@jonasfj are there official pub guidelines around how you should version an unpublished package?

jonasfj: Not from pub, there are no guidelines.

For external package maintenance the Dart team generally follows:
https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#publishing-a-package

I didn't write it, and I'll admit I sometimes fail to follow it, but I like the idea of using a `-wip` to denote the fact that the version haven't been published yet.

CoderDake (approved):

PR #6276, opened 2023-08-29
Add `showNotification` and `showBannerMessage` to extension API

kenzieschmoll:
https://github.com/flutter/devtools/assets/43759233/fe86eef3-f98b-41bb-aa55-569fdf91eeac


packages/devtools_extensions/example/foo_devtools_extension/lib/main.dart
```diff
@@ -74,6 +74,29 @@ class _FooExtensionHomePageState extends State<FooExtensionHomePage> {
             ),
             const SizedBox(height: 48.0),
             Text('Received message from DevTools: $_message'),
+            const SizedBox(height: 48.0),
```
bkonyi, line 77: Nit: `const SizedBox(height: 64.0)` instead of two `SizedBox`es?

packages/devtools_extensions/lib/src/api/model.dart
```diff
@@ -53,3 +53,79 @@ class DevToolsExtensionEvent {
 
 /// A void callback that handles a [DevToolsExtensionEvent].
 typedef ExtensionEventHandler = void Function(DevToolsExtensionEvent event);
+
+class ShowNotificationExtensionEvent extends DevToolsExtensionEvent {
```
bkonyi, line 57: Are these classes accessible to developers? If so, we need documentation for them.

packages/devtools_extensions/example/foo_devtools_extension/lib/main.dart
kenzieschmoll: removed

packages/devtools_extensions/lib/src/api/model.dart
kenzieschmoll: moved to a file that is not exported to the user. Users should send events through the ExtensionManager. We can always expose later if we need to.

bkonyi (approved):

packages/devtools_extensions/lib/src/api/model.dart
kenzieschmoll: Update: I missed that we use these classes from devtools_app to parse, so made them public and documented in model.dart

Issue #4921, opened 2022-12-09
RenderFlex overflow on the logging screen


kenzieschmoll:
![Screenshot 2022-12-09 at 10 50 03 AM](https://user-images.githubusercontent.com/43759233/206773054-c7a25b25-b82b-4a35-97ca-11a469fc85e3.png)

kenzieschmoll:
We can remove the text on the Clear button and reduce the size of the search field when the screen is narrow. 

In general, it could be worth moving the show structured errors toggle into a settings menu.

PR #6281, opened 2023-08-29
Clean up Logging page top bar

CoderDake:
Fixes https://github.com/flutter/devtools/issues/4921

https://github.com/flutter/devtools/assets/1386322/ebe4f959-72b3-4d02-a755-a1bbef7fb577

# Open question
 The structured errors are saved in the service, not in the local preferences. Should we have a way to somehow let the user know that this setting is saved in their app?


kenzieschmoll:
> The structured errors are saved in the service, not in the local preferences. Should we have a way to somehow let the user know that this setting is saved in their app?

Have we had users complain about this? If not, leaving as is is probably fine because I don't think we are really even using this functionality in the logging page. IIRC, this was to support an Inspector-like view of errors, which we have an issue somewhere to add to DevTools

packages/devtools_app/lib/src/screens/logging/logging_screen.dart
```diff
@@ -122,13 +123,15 @@ class _LoggingScreenState extends State<LoggingScreenBody>
...
           gaSelection: gac.clear,
+          minScreenWidthForTextBeforeScaling: loggingMinVerboseWidth,
         ),
         const Spacer(),
-        const StructuredErrorsToggle(),
         const SizedBox(width: denseSpacing),
         // TODO(kenz): fix focus issue when state is refreshed
         SearchField<LoggingController>(
-          searchFieldWidth: wideSearchFieldWidth,
+          searchFieldWidth: includeText(context, loggingMinVerboseWidth)
```
kenzieschmoll, line 132: maybe rename `includeText` to `screenWiderThanMin` or `useNarrowDisplay` or similar since this applies to more cases than just including text

packages/devtools_app/lib/src/screens/logging/logging_screen.dart
```diff
@@ -181,3 +198,34 @@ class _LoggingScreenState extends State<LoggingScreenBody>
...
   }
 }
+
+class LoggingSettingsDialog extends StatelessWidget {
+  const LoggingSettingsDialog({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+    final dialogHeight = scaleByFontFactor(400.0);
```
kenzieschmoll, line 208: do we need the dialog height? if we don't have it, will this dialog just shrink to the size of its content?

kenzieschmoll (approved):

Issue #6289, opened 2023-08-30
Flutter web integration test times out and flakes


kenzieschmoll:
Sometimes the test fails with a timeout message, and sometimes it fails with this failure: https://gist.github.com/kenzieschmoll/69e7f58fa392e90219b02625ca0e259e

kenzieschmoll:
CC @elliette this is the debugger panel test for web

CoderDake:
Able to reproduce locally by changing https://github.com/CoderDake/devtools/blob/0be05d9e7b1a60c8a84969382f6e8b81e6bf48dd/packages/devtools_app/integration_test/test/live_connection/debugger_panel_test.dart#L110 to  1 millisecond:
```dart
  await tester.pumpAndSettle(const Duration(milliseconds: 1));
  ````
  
  I will change this to the long duration. Hopefully we will see less flake by waiting 6 seconds for the call stacks to resolve, rather than just waiting 3 seconds

CoderDake:
Found this issue in the integration_tests for the above PR
https://github.com/flutter/devtools/issues/6294

PR #6291, opened 2023-08-30
Fixing flakey debug panel test

CoderDake:
Fixes https://github.com/flutter/devtools/issues/6289

I was able to reproduce locally by reducing the pump duration, after setting a breakpoint, to 1 millisecond.

Therefore I think it would be reasonable to use the 6 second pump, rather than 3s, to give the callstack a bit more time to load.

If we are still having a consistent amount of flake, then we can rediscuss.


kenzieschmoll (approved):

kenzieschmoll:
can we also increase the timeout for the current failure so we don't hit this one as much?

elliette:
Thank you!!

Issue #6294, opened 2023-08-30
Test flake: integration_test, Did not receive expected app.started event


CoderDake:
https://github.com/flutter/devtools/actions/runs/6029174664/job/16358147433?pr=6291#step:4:826

It appears that sometimes the sample application times out after 120s looking for the app.started event

PR #6295, opened 2023-08-30
Fix App.started timeout in integration tests

CoderDake:
Fixes https://github.com/flutter/devtools/issues/6294

The app is likely timing out too early in some instances. I've doubled the timeout values for app start to `4 minutes`.

If we still have timeouts then we may need to do more investigation into why the app start is increasing so much.

### DIFF
--- a/packages/devtools_app/test/test_infra/flutter_test_driver.dart
+++ b/packages/devtools_app/test/test_infra/flutter_test_driver.dart
@@ -315,10 +315,9 @@ class FlutterRunTestDriver extends FlutterTestDriver {
     // to stdout. This avoids race conditions and deadlocks.
     final Future<Map<String, dynamic>> startedFuture =
         waitFor(event: 'app.started', timeout: appStartTimeout);
-    final connectedFuture =
-        waitFor(event: 'daemon.connected');
+    final connectedFuture = waitFor(event: 'daemon.connected');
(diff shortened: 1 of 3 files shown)


packages/devtools_app/test/test_infra/flutter_test_driver.dart
```diff
@@ -311,28 +311,32 @@ class FlutterRunTestDriver extends FlutterTestDriver {
...
+
     await super.setupProcess(
       args,
       flutterExecutable: flutterExecutable,
       runConfig: runConfig,
       pidFile: pidFile,
     );
 
+    // Await the startedFuture, which signals that the app has started.
+    _currentRunningAppId = (await startedFuture)['params']['appId'];
```
kenzieschmoll, line 330: do we expect this to happen before the `connected` future? Does it matter what order these complete in? If not, it may be safer to await both at once using `Future.wait`

CoderDake: It shouldn't matter what order they are awaited. 
In the  logs we would expect `daemon.connected` to happen earlier than `app.started`.
So waiting on the `app.started` future first just means that we would expect the `daemon.connected` future to be ready to complete immediately upon await.

kenzieschmoll (approved):

Issue #6144, opened 2023-08-02
Heap snapshot uses order of magnitude more memory than actual heap size


bkonyi:
While attempting to debug a memory issue in a relatively simple application using heap snapshot diffing, I encountered some DevTools crashes while attempting to take the second snapshot.

I opened Chrome DevTools and saw the following stats in the memory page:

Before snapshot:
![image](https://github.com/flutter/devtools/assets/24210656/09a77c1b-8f7f-47c3-806e-0699990d7b9a)

After snapshot (147.4 MiB heap):
![image](https://github.com/flutter/devtools/assets/24210656/4e1bc78a-79ad-4c55-8758-e9f401f4dc90)

After second snapshot attempt (tab crashes at 4GiB after hitting 64-bit Chrome's memory limit):
![image](https://github.com/flutter/devtools/assets/24210656/4685a537-aea7-4588-abe9-25f51e7810df)

This order of magnitude overhead after processing a heap snapshot makes the snapshot diffing tool unusable in the web version of DevTools, even when examining relatively small heaps.

FYI @kenzieschmoll

polina-c:
I tested it with gallery (that is relatively complicated) and I could take 10 snapshots before crashing.

Yes, we can add optimization at cost of code simplicity and user experience, but there still will be applications that will cause crash, because other platforms allow much bigger memory consumption than Chrome tab.

I suggest to start showing warning that DevTools can reach chrome memory limit and instruct to either delete existing snapshots or to switch to macos version of devtools.

bkonyi:
I just tested with gallery in debug mode as a MacOS desktop application and saw the following.

Before snapshot:
![image](https://github.com/flutter/devtools/assets/24210656/e0e40a1e-41cf-402b-8410-ad19782df2a6)

After first snapshot:
![image](https://github.com/flutter/devtools/assets/24210656/5930ea75-c78b-4be0-b3e3-965522f8bcb6)

Crash after starting 5th snapshot (was at ~4000 MiB after 4 snapshots):
![image](https://github.com/flutter/devtools/assets/24210656/b5d7b875-679f-4987-95cb-0e8b90ee61f5)

[...]

PR #6297, opened 2023-08-31
Decrease memory footprint of heap snapshot.

polina-c:
Contributes to https://github.com/flutter/devtools/issues/6144

This PR decreases memory footprint for snapshot for Counter application by 18%, by reusing the class HeapClassName. The gain is bigger for applications with bigger number of objects per class.


packages/devtools_app/lib/src/shared/memory/adapted_heap_object.dart
```diff
@@ -7,24 +7,36 @@ import 'package:vm_service/vm_service.dart';
...
 import 'simple_items.dart';
 
-/// Names for json fields.
-class _JsonFields {
-  static const String code = 'code';
-  static const String references = 'references';
-  static const String klass = 'klass';
-  static const String library = 'library';
-  static const String shallowSize = 'shallowSize';
+// TODO: remove wrappers.
```
bkonyi, line 10: Is this to be done in a future change? If so, can we leave a comment explaining why they're here?

bkonyi (approved):
Very nice! Thanks for starting to dig into this. Did you happen to try this with the gallery or https://github.com/bkonyi/dev_experience_exploration/tree/main/dart/life?

packages/devtools_app/lib/src/shared/memory/adapted_heap_object.dart
polina-c: Oh, forgot to remove. Thanks. It improved memory even more.

polina-c:
> Very nice! Thanks for starting to dig into this. Did you happen to try this with the gallery or https://github.com/bkonyi/dev_experience_exploration/tree/main/dart/life?

Not yet

packages/devtools_app/test/memory/shared/heap/heap_analyzer_golden_test.dart
```diff
@@ -2,13 +2,38 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:io';
```
kenzieschmoll, line 5: if we are adding a `dart:io` import, we should also add `@TestOn('vm')` since this test would no longer be able to be run on web, (as long as we are okay with this test no longer running on web)

kenzieschmoll: on second thought, maybe we should put this memory foot print test in another test library? having designated memory tests like this for DevTools is something we will likely want to do for other places where we suspect memory bloat.

Wdyt about creating
`test/memory/memory_footprint_test.dart` where we can put all tests for the memory page that are related to monitoring memory consumption of Devtools itself?

polina-c: good ideas!

packages/devtools_app/lib/src/shared/memory/class_name.dart
```diff
@@ -72,26 +72,40 @@ enum ClassType {
...
-
-  HeapClassName.fromHeapSnapshotClass(HeapSnapshotClass? theClass)
-      : this(
-          library: _library(
-            theClass?.libraryName,
-            theClass?.libraryUri.toString(),
-          ),
-          className: theClass?.name ?? '',
-        );
+  static final _instances = <HeapClassName>{};
```
kenzieschmoll, line 78: Does this set ever need to be cleared?

kenzieschmoll (approved):
one question, but LGTM

packages/devtools_app/lib/src/shared/memory/class_name.dart
polina-c: Good point. It should be cleared on reconnect. I can initiate the clearing in MemoryController. What is the best way to catch reconnect event?

kenzieschmoll: If you can verify that `MemoryController.dispose()` is called on disconnect, that would be a good place to put it.

kenzieschmoll: And if our controllers are not being disposed on disconnect, then that is a bug

kenzieschmoll: Looks like they are not being disposed - this is a leak https://github.com/flutter/devtools/issues/6301. 

For this PR, I would make sure that this set of instances is cleared out by proxy of calling `MemoryController.dispose` and then the leak in the issue I referenced should be fixed, which would then clear this set as expected.

polina-c: Thanks. Done.

PR #6299, opened 2023-08-31
Add an end to end example of for building a DevTools extension

kenzieschmoll:
The `devtools_extensions/example/README.md` explains how this example fits together.


packages/devtools_extensions/example/README.md
```diff
@@ -0,0 +1,43 @@
+# Example DevTools Extension
+
+This is an end-to-end example of a DevTools extension, including
+- the extension Flutter web app that will be embedded in DevTools
+(`foo/packages/foo_devtools_extension`)
+- the parent package that provides the DevTools extension (`foo/packages/foo`)
+- the end-user application that depends on `package:foo` (`app_that_uses_foo`),
+which will trigger a load of the `pacakge:foo` DevTools extension when debugging
```
polina-c, line 8: How about ordering it differently:

This is an end-to-end example of a DevTools extension, including:
- the package that enables the DevTools extension for end-user applications (`foo/packages/foo`)
- the end-user application that depends on `package:foo` (`app_that_uses_foo`),
which will trigger a load of the `pacakge:foo` DevTools extension when debugging
this app with DevTools.
- the extension Flutter web app that will be embedded in DevTools
(`foo/packages/foo_devtools_extension`)

packages/devtools_extensions/example/README.md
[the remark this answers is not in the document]
kenzieschmoll: Yes, I will use this example app when I add the end to end integration test in devtools_app.

packages/devtools_extensions/example/README.md
kenzieschmoll: re-ordered.

PR #6303, opened 2023-09-01
Enable and fix test, fix error handling and add new packages to upgrade.sh.

polina-c:


packages/devtools_app/integration_test/run_tests.dart
```diff
@@ -28,10 +28,10 @@ void main(List<String> args) async {
   Exception? exception;
   final chromedriver = ChromeDriver();
 
-  try {
-    // Start chrome driver before running the flutter integration test.
-    await chromedriver.start();
+  // Start chrome driver before running the flutter integration test.
+  await chromedriver.start();
```
kenzieschmoll, line 32: why is this moved out of the try block? if an exception is thrown we need to catch it.

polina-c: We rethrow the exception anyway.
But if start fails, finally also fails because late _process in ChromeDriver is not initialized.

kenzieschmoll: Let's just make `_process` nullable in the `ChromeDriver` class, and put this back into the try block. We only rethrow on line 62 for exceptions caught in the try block.

polina-c: What is the point to save exception in order to rethrow it?
I would just remove `catch` from here.

packages/devtools_app/integration_test/run_tests.dart
```diff
@@ -52,15 +51,9 @@ void main(List<String> args) async {
         await _runTest(newArgsWithTarget);
       }
     }
-  } on Exception catch (e) {
-    exception = e;
```
kenzieschmoll, line 56: please don't remove all this exception handling.

kenzieschmoll (approved):

packages/devtools_app/integration_test/run_tests.dart
polina-c: We discussed that this exception handling does not change behavior.

PR #6332, opened 2023-09-07
Fix minor updates so that the template properly resets

CoderDake:
While prepping a branch to update us to a dev version on the master branch I noticed that the NEXT_RELEASE template wasn't properly updating. To fix this I've added linting rules that will make sure we always await our futures.


tool/update_version.dart
```diff
@@ -0,0 +1,388 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: why was this file moved out of commands?

CoderDake: oh shoot this is an artifact I forgot to push up my remove

kenzieschmoll (approved):

PR #6335, opened 2023-09-07
Pull vm service uri from extension iFrame query parameters

kenzieschmoll:


packages/devtools_extensions/lib/src/template/extension_manager.dart
```diff
@@ -30,23 +30,28 @@ class ExtensionManager {
...
-    // DevTools app.
+    final vmServiceUri = queryParams['uri'];
     if (connectToVmService) {
-      // Request the vm service uri for the connected app. DevTools will
-      // respond with a [DevToolsPluginEventType.connectedVmService] event with
-      // containing the currently connected app's vm service URI.
-      postMessageToDevTools(
-        DevToolsExtensionEvent(DevToolsExtensionEventType.vmServiceConnection),
-      );
+      if (vmServiceUri != null) {
```
polina-c, line 43: Can you invert it to make it easier to read?
```
if (vmServiceUri == null) {
   ...
```

polina-c (approved):
One comment then lgtm

Issue #1941, opened 2020-05-07
Hide hot restart and hot reload buttons based on the availability of the service


kenzieschmoll:
In https://github.com/flutter/devtools/pull/1930, we hid these buttons for non-flutter apps, but this is problematic for dart server apps that support hot reload. 

We need to hide the buttons and update the available actions space on the app bar when the hot reload and hot restart services are not available

kenzieschmoll:
@julemand101 do you have a small sample app (non-flutter) that has the hot reload service available? I'm unable to repro this behavior with the Dart CLI apps I am using.

kenzieschmoll:
After further investigation, it doesn't appear that the `reloadServices` service is registered for Dart CLI apps by default. The Flutter framework calls `vmService.registerService('reloadSources')`, which is how DevTools is aware of the service. Calling that service ultimately bubbles down to calling the method `VmService.reloadSources`. The `VmService.reloadSources` RPC is what other packages like `hotreloader` also use: https://github.com/vegardit/dart-hotreloader/blob/main/lib/hotreloader.dart#L255. 

@bkonyi what is the best way to know if the `reloadSources` functionality is available for a Dart CLI app?

julemand101:
@kenzieschmoll Well, one case I know of is just running the program with JIT and the Dart VM Service enabled since that opens up for any program to ask the Dart program to reload its sources.

A simple program like this can be used (where we later will remove the comment):
```dart
import 'dart:async';

void main() {
  Timer.periodic(Duration(seconds: 3), myFunction);
}

void myFunction(Timer _) {
  // print('Hello world');
}
```

1. Run Dart code with `dart --observe --serve-observatory program.dart`
2. Visit the old Dart Observatory where a link to it is printed as the first line of output. E.g.:
```
The Dart VM service is listening on http://127.0.0.1:8181/Fdd5tkhH8Ds=/
```
3. Click on the main isolate and you can then find this panel:
![image](https://github.com/flutter/devtools/assets/116227/0a1a15b2-9f5d-450d-abcc-df76b7cee9fa)

4. Remove the `//` in the source code and click "Reload Source".
5. Notice the program now starts printing the message.

A more advanced example can be made where we can use `dev.Service.controlWebServer(enable: true)` from `dart:developer` to open the Dart VM Service and then connect to that and trigger the reloads.

kenzieschmoll:
[...]

PR #6341, opened 2023-09-08
Show hot restart and hot reload button based on service availability

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/1941


CoderDake (approved):

packages/devtools_app/lib/src/app.dart
```diff
@@ -284,21 +284,23 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
...
-                          .serviceManager.connectedApp?.isFlutterAppNow ??
-                      false) ...[
-                    const HotReloadButton(),
-                    const HotRestartButton(),
-                  ],
+                if (connectedToVmService) ...[
+                  // TODO(kenz): do we need to hide this for Dart Web?
+                  HotReloadButton(
+                    callOnVmServiceDirectly: !connectedToFlutterApp,
+                  ),
```
kenzieschmoll, line 301: @elliette does Dart Web have hot reload functionality like dart server apps?

elliette: It depends - for Flutter, hot reload simply triggers hot restart: https://github.com/flutter/flutter/issues/39848 For acx, hot reload is supported.

kenzieschmoll: Okay, so how can we make sure this Hot Reload button calls the right path for ACX? For ACX is it implemented through a service extension like for Flutter? Or for ACX do we need to call `service?.reloadSources` directly like we are now doing in this PR for Dart CLI apps?

packages/devtools_app/lib/src/app.dart
```diff
@@ -284,21 +284,23 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
...
+                  const HotRestartButton(),
```
bkonyi, line 302: Hot Restart isn't a concept in the Dart CLI apps, so this shouldn't be shown for non-Flutter applications.

packages/devtools_app_shared/lib/src/service/service_manager.dart
```diff
@@ -413,7 +415,13 @@ class ServiceManager<T extends VmService> {
 
   /// This can throw an [RPCError].
   Future<void> performHotReload() async {
-    await callServiceOnMainIsolate(hotReloadServiceName);
+    if (connectedApp?.isFlutterAppNow ?? false) {
+      await callServiceOnMainIsolate(hotReloadServiceName);
+    } else {
+      await service?.forEachIsolate((isolate) async {
```
bkonyi, line 421: Ubernit: create a local for `service` to avoid the repeated null checks.

bkonyi (approved):

packages/devtools_app/lib/src/app.dart
kenzieschmoll: With the latest commit, I hid this button for dart web apps. This is not a regression since we were only showing this button for flutter apps before this PR. I've added a TODO to show this button for Dart web apps when supported, but we need to figure out

packages/devtools_app/lib/src/app.dart
kenzieschmoll: Yes, this button hides itself based on whether the service is available.

kenzieschmoll: added a comment to make that more clear

PR #6363, opened 2023-09-12
Add Hot Reload + Hot Restart functionality to sidebar

DanTup:
This adds capabilities and buttons for hot reload/hot restart, and moves the DevTools buttons into a menu.

The sidebar overall still requires work (both visual - this is just default menu styling - and for the set of DevTools pages + enable conditions), this is just trying to get the functionality done since the visuals don't require coordination with Dart-Code.

https://github.com/flutter/devtools/assets/1078012/af33a25b-74a2-4d7c-86b3-ce165f03c59f

### DIFF
--- a/packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
@@ -76,7 +76,7 @@ abstract interface class VsCodeApi {
   static const jsonCommandParameter = 'command';
   static const jsonArgumentsParameter = 'arguments';
   static const jsonIdParameter = 'id';
-  static const openPageParameter = 'page';
+  static const jsonOpenPageParameter = 'page';
   static const jsonDebugSessionIdParameter = 'debugSessionId';
 }
 
--- a/packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
@@ -63,27 +63,21 @@ abstract interface class VsCodeApi {
   static const jsonApiName = 'vsCode';
 
   static const jsonInitializeMethod = 'initialize';
-
   static const jsonExecuteCommandMethod = 'executeCommand';
-  static const jsonExecuteCommandCommandParameter = 'command';
-  static const jsonExecuteCommandArgumentsParameter = 'arguments';
-
-  static const jsonDevicesChangedEvent = 'devicesChanged';
 
   static const jsonSelectDeviceMethod = 'selectDevice';
-  static const jsonSelectDeviceIdParameter = 'id';
+  static const jsonOpenDevToolsPageMethod = 'openDevToolsPage';
+  static const jsonHotReloadMethod = 'hotReload';
+  static const jsonHotRestartMethod = 'hotRestart';
 
-  static const openDevToolsPageMethod = 'openDevToolsPage';
-  static const openDevToolsPageDebugSessionIdParameter = 'debugSessionId';
-  static const openDevToolsPagePageParameter = 'page';
-
-  static const hotReloadMethod = 'hotReload';
-  static const hotReloadDebugSessionIdParameter = 'debugSessionId';
-
-  static const hotRestartMethod = 'hotRestart';
-  static const hotRestartDebugSessionIdParameter = 'debugSessionId';
-
+  static const jsonDevicesChangedEvent = 'devicesChanged';
   static const jsonDebugSessionsChangedEvent = 'debugSessionsChanged';
+
+  static const jsonCommandParameter = 'command';
+  static const jsonArgumentsParameter = 'arguments';
+  static const jsonIdParameter = 'id';
+  static const openPageParameter = 'page';
+  static const jsonDebugSessionIdParameter = 'debugSessionId';
 }
 
 /// This class defines a device exposed by the Dart/Flutter extensions in VS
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
```diff
@@ -71,6 +77,12 @@ abstract interface class VsCodeApi {
   static const openDevToolsPageDebugSessionIdParameter = 'debugSessionId';
   static const openDevToolsPagePageParameter = 'page';
 
+  static const hotReloadMethod = 'hotReload';
+  static const hotReloadDebugSessionIdParameter = 'debugSessionId';
+
+  static const hotRestartMethod = 'hotRestart';
+  static const hotRestartDebugSessionIdParameter = 'debugSessionId';
```
kenzieschmoll, line 84: maybe we should have a single parameter `debugSessionIdParameter` that can be reused instead of duplicating this three times.

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -29,88 +30,109 @@ class DebugSessions extends StatelessWidget {
...
     final isFlutter = session.debuggerType?.contains('Flutter') ?? false;
 
-    return TableRow(
+    return MenuBar(
       children: [
-        Text(
-          '$name ($mode)',
-          style: Theme.of(context).textTheme.titleSmall,
+        IconButton(
+          onPressed: api.capabilities.hotReload && (isDebug || !isFlutter)
```
kenzieschmoll, line 76: is `(isDebug || !isFlutter)` the same condition used in VS code? I'm wondering if this would be incorrect for dart web apps, or Dart CLI apps?

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -29,88 +30,109 @@ class DebugSessions extends StatelessWidget {
...
     // TODO(dantup): What to show if mode is unknown (null)?
-    final name = session.name;
     final mode = session.flutterMode;
     final isDebug = mode == 'debug';
     final isProfile = mode == 'profile';
     // final isRelease = mode == 'release' || mode == 'jit_release';
     final isFlutter = session.debuggerType?.contains('Flutter') ?? false;
 
-    return TableRow(
+    return MenuBar(
```
kenzieschmoll, line 73: let's just use a Row instead of a Menu Bar. Visually having the buttons all just be an icon (or text) against a transparent background is probably what we want.

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
DanTup: It is currently what Dart-Code does, yep:

https://github.com/Dart-Code/Dart-Code/blob/3efd2a89bc85a006f7771d92297ebbac93380af2/package.json#L681

I'm not sure about Web, but for Dart CLI there's not profile/release mode so it'll be enabled because of isDebug (my understanding is that `reloadSources` is always available).

If the rules are more complex than this, we should probably find a way for the editors to find out via the VM Service/DDS to avoid putting complex rules into editors (because that gets complicated when they change, since they have to support multiple SDK versions).

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
DanTup: The MenuBar seemed like an easy way to get the dropdown for the DevTools buttons. My plan was to make it transparent (I'm assuming that's possible) but I haven't looked too much at the visuals, I wanted to get all the functionality both here and matching in VS Code first since there's no dependency between them for the visuals.

packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
DanTup: Done! I also renamed the others so they're just the strings (with `json` prefix - something I'd been inconsistent with - and a suffix for method/event/parameter). I'd like to code-gen this code really (because copy/pasting it and adding new methods seems error-prone) but it hasn't yet felt worth the effort.

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
kenzieschmoll: We have a shared widget `AnalyticsDropDownButton` that you could use. We likely want to be tracking analytics for all this sidebar stuff so we can measure the usage anyway. Fine to leave as is for now and re-work the UI in a follow up

packages/devtools_app/lib/src/standalone_ui/api/vs_code_api.dart
```diff
@@ -63,27 +63,21 @@ abstract interface class VsCodeApi {
...
+  static const jsonHotReloadMethod = 'hotReload';
+  static const jsonHotRestartMethod = 'hotRestart';
 
+  static const jsonDevicesChangedEvent = 'devicesChanged';
   static const jsonDebugSessionsChangedEvent = 'debugSessionsChanged';
+
+  static const jsonCommandParameter = 'command';
+  static const jsonArgumentsParameter = 'arguments';
+  static const jsonIdParameter = 'id';
+  static const openPageParameter = 'page';
```
kenzieschmoll, line 79: is there a reason this is the only one that isn't prefixed with "json"?

DanTup: Oops, missed that one - thanks!

kenzieschmoll (approved):

PR #6364, opened 2023-09-12
Convert Fix Goldens to dart script

CoderDake:


tool/lib/commands/fix_goldens.dart
```diff
@@ -0,0 +1,101 @@
+// #!/bin/bash -e
```
kenzieschmoll, line 1: can we get rid of these? also add copyright please

tool/lib/commands/fix_goldens.dart
```diff
@@ -0,0 +1,101 @@
...
+import 'dart:async';
+import 'dart:io';
+
+import 'package:args/command_runner.dart';
+import 'package:io/io.dart';
+import 'package:path/path.dart' as path;
+
+import '../utils.dart';
+
+class FixGoldensCommand extends Command {
```
kenzieschmoll, line 12: dart doc

tool/lib/commands/fix_goldens.dart
```diff
@@ -0,0 +1,101 @@
...
+
+  @override
+  FutureOr? run() async {
+    // Change the CWD to the repo root
+    Directory.current = pathFromRepoRoot("");
+
+    final runId = argResults!['run-id']!;
+    final tmpDownloadDir = Directory(
+      path.join('.tmp${DateTime.now().millisecondsSinceEpoch}'),
+    );
```
kenzieschmoll, line 39: can we use `Directory.systemTemp.createTemp` instead?

tool/lib/commands/fix_goldens.dart
```diff
@@ -0,0 +1,101 @@
...
+      await processManager.runProcess(
+        CliCommand.from('gh', [
+          'run',
+          'download',
+          runId,
+          '-p',
+          '*golden_image_failures*',
+          '-D',
+          tmpDownloadDir.path,
+        ]),
```
kenzieschmoll, line 52: trailing comma. We should make sure the trailing commas lint is applied to the tool/ directory

tool/lib/commands/fix_goldens.dart
CoderDake: yep this was accidentally left behind

tool/lib/commands/fix_goldens.dart
```diff
@@ -0,0 +1,101 @@
...
+        }
+        await downloadedGolden.rename(destinationPath);
+
+        print("Fixed: $destinationPath");
+      }
+
+      print('Done updating ${downloadedGoldens.length} goldens');
+    } finally {
+      tmpDownloadDir.deleteSync(recursive: true);
+    }
```
kenzieschmoll, line 99: I think you need to call

```
    // Closes stdin for the entire program.
    await sharedStdIn.terminate();
```
at the end or this script will not terminate due to an open port

tool/lib/commands/fix_goldens.dart
CoderDake: @kenzieschmoll which lint is this? Currently tool has `require_trailing_commas`

tool/lib/commands/fix_goldens.dart
CoderDake: I haven't had any issues with this yet. What makes you think the stdin would hang?

kenzieschmoll (approved):

Issue #6272, opened 2023-08-28
Polish DevTools Extensions before releasing


kenzieschmoll:
- [ ] Add theme update support
- [ ] Ensure screen loads if refreshing browser while on an extension screen
- [ ] Load extensions on hot reload or hot restart event - if list of screens has changed, refresh.

PR #6368, opened 2023-09-12
Add an action to force a reload of a DevTools extension

kenzieschmoll:
This PR adds the "Force reload extension" action and moves the existing "Disable extension" action into the shared context menu.

Screenshot 2023-09-12 at 1 02 00 PM

force-reload

This forces a reload of the iFrame, which is a safety feature to protect against edge cases and the user getting into a bad state with an Extension iFrame. We will monitor usage of this action as we roll DevTools extensions out to users to see how often users are having to use this escape hatch.

Work towards https://github.com/flutter/devtools/issues/6272

### DIFF
--- a/packages/devtools_app/lib/src/extensions/embedded/_controller_web.dart
+++ b/packages/devtools_app/lib/src/extensions/embedded/_controller_web.dart
@@ -98,8 +98,7 @@ class EmbeddedExtensionControllerImpl extends EmbeddedExtensionController
   @override
   void forceReload() {
     if (_extensionIFrame.contentWindow != null) {
-      // ignore: unsafe_html, forcing reload by resetting the pre-existing
-      // IFrame src.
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/extensions/embedded/_controller_web.dart
```diff
@@ -95,6 +95,14 @@ class EmbeddedExtensionControllerImpl extends EmbeddedExtensionController
     extensionPostEventStream.add(DevToolsExtensionEvent(type, data: data));
   }
 
+  @override
+  void forceReload() {
+    if (_extensionIFrame.contentWindow != null) {
+      // ignore: unsafe_html, forcing reload.
```
elliette, line 101: Consider changing comment to `// ignore: unsafe_html, forcing reload by resetting the pre-existing IFRAME src`(or something like that) to make it clear why this is safe

elliette (approved):
One suggestion but otherwise LGTM

PR #6374, opened 2023-09-13
Clean up the extension example app

kenzieschmoll:
This cleans up a few things and adds the extension build to the example extension package foo (the extension will not load otherwise when the example app is connected to devtools)


packages/devtools_extensions/example/foo/packages/foo/extension/devtools/build/assets/AssetManifest.bin
polina-c, line 1: Are you sure we need this file in git?

kenzieschmoll: The `foo/extension/devtools/build/` directory needs the built assets of the extension in order for this example to be functional

polina-c (approved):
One comment then lgtm.

PR #6380, opened 2023-09-14
Improve the VS Code sidebar UI

DanTup:
Some tweaks to the sidebar UI. Currently looks like this:

https://github.com/flutter/devtools/assets/1078012/267c9006-8683-4433-950a-38178abc6544

Things that still need doing:

- [ ] background color is from the editor, not the sidebar
- [ ] test light theme
- [ ] need an icon for DevTools dropdown
- [ ] debug session names show literal `(null)` for mode while starting
- [ ] some items missing from DevTools menu
- [ ] DevTools page conditions need verifying/sharing
- [ ] ...

@kenzieschmoll @anderdobo

### DIFF
--- a/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
@@ -53,11 +53,15 @@ class DebugSessions extends StatelessWidget {
     // final isRelease = mode == 'release' || mode == 'jit_release';
     final isFlutter = session.debuggerType?.contains('Flutter') ?? false;
 
+    final label = session.flutterMode != null
+        ? '${session.name} (${session.flutterMode})'
+        : session.name;
+
     return TableRow(
       children: [
         Text(
-          '${session.name} (${session.flutterMode})',
-          style: Theme.of(context).textTheme.titleSmall,
+          label,
+          style: Theme.of(context).regularTextStyle,
         ),
         IconButton(
           onPressed: api.capabilities.hotReload && (isDebug || !isFlutter)
@@ -132,14 +136,13 @@ class _DevToolsMenu extends StatelessWidget {
     }
 
     return Directionality(
-      textDirection: reversedDirection,
+      // Reverse the direction so the menu is anchored on the far side and
+      // expands in the opposite direction with the icons on the right.
+      textDirection: normalDirection,
       child: MenuAnchor(
-        // TODO(dantup): How to flip the menu to be anchored from the right
-        //  and expand to the left?
         style: const MenuStyle(
           alignment: AlignmentDirectional.bottomStart,
         ),
-        alignmentOffset: const Offset(2, 0),
         menuChildren: [
           // TODO(dantup): Ensure the order matches the DevTools tab bar (if
           //  possible, share this order).
--- a/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
@@ -138,7 +138,7 @@ class _DevToolsMenu extends StatelessWidget {
     return Directionality(
       // Reverse the direction so the menu is anchored on the far side and
       // expands in the opposite direction with the icons on the right.
-      textDirection: normalDirection,
+      textDirection: reversedDirection,
       child: MenuAnchor(
         style: const MenuStyle(
           alignment: AlignmentDirectional.bottomStart,
--- a/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
@@ -184,7 +184,7 @@ class _DevToolsMenu extends StatelessWidget {
           tooltip: 'DevTools',
           // TODO(dantup): Icon for DevTools menu?
           icon: Icon(
-            Icons.developer_mode,
+            Icons.construction,
             size: actionsIconSize,
           ),
         ),
--- a/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
@@ -24,45 +24,28 @@ class DebugSessions extends StatelessWidget {
       children: [
         Text(
           'Debug Sessions',
-          style: Theme.of(context).textTheme.titleSmall,
+          style: Theme.of(context).textTheme.titleMedium,
         ),
         if (sessions.isEmpty)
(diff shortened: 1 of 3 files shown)


kenzieschmoll:
> need an icon for DevTools dropdown

Can we just use the Dart Icon? @InMatrix @nancyhu2023 should we create an icon specific to DevTools (like the Dart icon with a wrench or something?)

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -86,53 +73,119 @@ class _DevToolsMenu extends StatelessWidget {
...
+      child: MenuAnchor(
+        // TODO(dantup): How to flip the menu to be anchored from the right
+        //  and expand to the left?
+        style: const MenuStyle(
+          alignment: AlignmentDirectional.bottomStart,
+        ),
+        alignmentOffset: const Offset(2, 0),
+        menuChildren: [
+          // TODO(dantup): Ensure the order matches the DevTools tab bar (if
+          //  possible, share this order).
```
kenzieschmoll, line 145: Checkout the ScreenMetaData enum in screen.dart. We can try to match the order there. If we do this, we should also add a test to `visible_screens_test.dart` that ensures that the order of DevTools screens matches the order specified in the enum.

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -86,53 +73,119 @@ class _DevToolsMenu extends StatelessWidget {
...
+            textDirection: normalDirection,
+            child: Text(screen.title ?? screen.id),
+          ),
+          icon: Icon(screen.icon, size: actionsIconSize),
+        ),
+      );
+    }
+
+    return Directionality(
+      textDirection: reversedDirection,
```
kenzieschmoll, line 135: why do we need this directionality widget?

packages/devtools_app/lib/src/standalone_ui/vs_code/devices.dart
```diff
@@ -50,18 +48,29 @@ class Devices extends StatelessWidget {
...
           child: TextButton(
-            child: Text(device.name),
+            style: TextButton.styleFrom(
+              alignment: Alignment.centerLeft,
+              shape: const ContinuousRectangleBorder(),
+            ),
+            child: Text(
+              device.name,
+              style:
+                  TextStyle(color: isSelected ? colorScheme.onPrimary : null),
```
kenzieschmoll, line 69: can we do `Theme.of(context).regularTextStyle.copyWith(color: ...)`?

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -24,54 +24,41 @@ class DebugSessions extends StatelessWidget {
...
     final isDebug = mode == 'debug';
     final isProfile = mode == 'profile';
     // final isRelease = mode == 'release' || mode == 'jit_release';
     final isFlutter = session.debuggerType?.contains('Flutter') ?? false;
 
-    return MenuBar(
+    return TableRow(
       children: [
+        Text(
+          '${session.name} (${session.flutterMode})',
```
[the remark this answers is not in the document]
DanTup: I dropped the mode if it's null. I think if we want to show explicitly that it's starting, we should tweak the API to make a more explicit status rather than using this field to infer it (I think they're equivalent but it feels like a slightly weird assumption).

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -24,54 +24,41 @@ class DebugSessions extends StatelessWidget {
...
+          style: Theme.of(context).textTheme.titleSmall,
```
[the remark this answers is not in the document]
DanTup: Done! Although I'm not actually sure what the difference is between that and just not having it. When I add that line and/or comment it out, the spacing of some of the letters changes _very slightly_ when I hot reload, but only for one session and not the other. Seems almost like some rounding issue:


https://github.com/flutter/devtools/assets/1078012/2e79c2d1-adcc-4648-998f-b8612f010262

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
DanTup: It's to have the menu anchor on the right and expand left (and also to put the icons on the right). I've added a comment to make this clearer.

Screenshot 2023-09-15 105539

Screenshot 2023-09-15 105549

packages/devtools_app/lib/src/standalone_ui/vs_code/devices.dart
DanTup: Done! I also applied regularTextStyle to the `TextButton` because it was making the text bold by default which didn't match the debug sessions.

image

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -132,14 +136,13 @@ class _DevToolsMenu extends StatelessWidget {
     }
 
     return Directionality(
-      textDirection: reversedDirection,
+      // Reverse the direction so the menu is anchored on the far side and
+      // expands in the opposite direction with the icons on the right.
+      textDirection: normalDirection,
```
kenzieschmoll, line 141: if we want this revered, shouldn't this be reversedDirection?

packages/devtools_app/lib/src/standalone_ui/vs_code/devices.dart
```diff
@@ -38,7 +38,7 @@ class Devices extends StatelessWidget {
             children: [
               for (final device in devices)
                 _createDeviceRow(
-                  colorScheme,
+                  Theme.of(context),
```
kenzieschmoll, line 41: store theme in a var at the top of the build method so we don't have to call Theme.of(context) more than once

kenzieschmoll (approved):
This is looking good! one comment then lgtm

InMatrix:
> Can we just use the Dart Icon? @InMatrix @Nancyhu2023 should we create an icon specific to DevTools (like the Dart icon with a wrench or something?)

How about the construction icon from Google Symbols and Icons? https://fonts.gstatic.com/s/i/short-term/release/googlesymbols/construction/default/24px.svg. 

image

packages/devtools_app/lib/src/standalone_ui/vs_code/devices.dart
DanTup: Done!

DanTup:
I've switched to the construction icon for now, but can easily change if we find something better (or DevTools gets a specific icon).

Issue #6357, opened 2023-09-11
Call to `addPubRootDirectories` throws on hot restart


kenzieschmoll:
```
flutter: [ERROR]: [zoneGuarded]: ext.flutter.inspector.addPubRootDirectories: (-32601) Unknown method "ext.flutter.inspector.addPubRootDirectories".
flutter: [ERROR]: [PlatformDispatcher]: ext.flutter.inspector.addPubRootDirectories: (-32601) Unknown method "ext.flutter.inspector.addPubRootDirectories".
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: ext.flutter.inspector.addPubRootDirectories: (-32601) Unknown method "ext.flutter.inspector.addPubRootDirectories".
#0      new _OutstandingRequest (package:vm_service/src/vm_service.dart:1930:45)
#1      VmService._call (package:vm_service/src/vm_service.dart:2499:21)
#2      VmService.callServiceExtension (package:vm_service/src/vm_service.dart:2470:14)
#3      VmServiceWrapper.callServiceExtension (package:devtools_app/src/service/vm_service_wrapper.dart:162:18)
#4      InspectorServiceBase.invokeServiceMethodDaemonNoGroup (package:devtools_app/src/shared/diagnostics/inspector_service.dart:176:57)
<asynchronous suspension>
#5      InspectorService.addPubRootDirectories (package:devtools_app/src/shared/diagnostics/inspector_service.dart:422:5)
<asynchronous suspension>
#6      InspectorPreferencesController.addPubRootDirectories.<anonymous closure> (package:devtools_app/src/shared/preferences.dart:287:7)
<asynchronous suspension>
#7      InspectorPreferencesController._customPubRootDirectoryBusyTracker (package:devtools_app/src/shared/preferences.dart:355:7)
<asynchronous suspension>
#8      InspectorPreferencesController.addPubRootDirectories (package:devtools_app/src/shared/preferences.dart:283:5)
<asynchronous suspension>
#9      InspectorPreferencesController.loadCustomPubRootDirectories.<anonymous closure> (package:devtools_app/src/shared/preferences.dart:340:9)
<asynchronous suspension>
#10     InspectorPreferencesController._customPubRootDirectoryBusyTracker (package:devtools_app/src/shared/preferences.dart:355:7)
<asynchronous suspension>
[...]

PR #6382, opened 2023-09-14
Fix pubRootDirectory errors on hot restart

CoderDake:
On hot restart it appears that the preferences are calling out before the services get registered.
I've added a helper that waits until the services are reported before allowing them to be called.

Fixes https://github.com/flutter/devtools/issues/6357

### DIFF
--- a/packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
+++ b/packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
@@ -454,7 +454,7 @@ class InspectorService extends InspectorServiceBase {
       checkIfComplete();
     });
 
-    return futureListener.future;
+    return futureListener.future.timeout(const Duration(seconds: 15));
   }
 
   Future<void> addPubRootDirectories(List<String> rootDirectories) async {
--- a/packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
+++ b/packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
@@ -418,12 +418,53 @@ class InspectorService extends InspectorServiceBase {
     return false;
   }
 
+  /// Ensures that addPubRootDirectories, getPubRootDirectories, and
+  /// removePubRootDirectories have all registered as services
+  /// before completing.
+  Future<void> waitUntilPubRootDirectoriesInitialized() {
+    final futureListener = Completer();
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
```diff
@@ -418,12 +418,53 @@ class InspectorService extends InspectorServiceBase {
...
+    void checkIfComplete() {
+      if (addAvailable && getAvailable && removeAvailable) {
+        futureListener.complete();
+      }
+    }
+
+    serviceConnection.serviceManager
+        .registeredServiceListenable(
+            WidgetInspectorServiceExtensions.addPubRootDirectories.name)
+        .addListener(() {
```
kenzieschmoll, line 438: these listeners need to be removed once we are finished with them. We also need to make sure the value of the `registeredServiceListenable(WidgetInspectorServiceExtensions.addPubRootDirectories.name)` listenable is true before we say that it is available. The value could come back as false in which case the service would not be available.

packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
```diff
@@ -418,12 +418,53 @@ class InspectorService extends InspectorServiceBase {
...
+      removeAvailable = true;
+      checkIfComplete();
+    });
+
+    return futureListener.future.timeout(const Duration(seconds: 15));
+  }
+
   Future<void> addPubRootDirectories(List<String> rootDirectories) async {
+    await waitUntilPubRootDirectoriesServicesAvailable();
```
kenzieschmoll, line 461: what if we hit the timeout, we would not want to call the service if it is not available

packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
kenzieschmoll: example: https://github.com/flutter/devtools/blob/master/packages/devtools_app_shared/lib/src/service/connected_app.dart#L157-L171

packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
CoderDake: It looks like timeout throws a TimeoutException so I don't think it would continue running

Issue #6387, opened 2023-09-14
Showing banner message from extension has latency 


kenzieschmoll:
I have seen this in the performance page in DevTools too. My suspicion is that this has something to do with adding the banner message from a post frame callback and then having to wait for a frame to be triggered for it to show up

PR #6393, opened 2023-09-14
Fix latency issue with showing banner message from DevTools extension

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/6387

### DIFF
--- a/packages/devtools_app/lib/src/shared/banner_messages.dart
+++ b/packages/devtools_app/lib/src/shared/banner_messages.dart
@@ -30,15 +30,26 @@ class BannerMessagesController {
   final _messages = <String, ListValueNotifier<BannerMessage>>{};
   final _dismissedMessageKeys = <Key?>{};
 
-  void addMessage(BannerMessage message) {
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/shared/banner_messages.dart
```diff
@@ -30,15 +30,26 @@ class BannerMessagesController {
...
   final _dismissedMessageKeys = <Key?>{};
 
-  void addMessage(BannerMessage message) {
-    // We push the banner message in a post frame callback because otherwise,
-    // we'd be trying to call setState while the parent widget `BannerMessages`
-    // is in the middle of `build`.
-    WidgetsBinding.instance.addPostFrameCallback((_) {
+  void addMessage(
+    BannerMessage message, {
+    bool callInPostFrameCallback = true,
```
polina-c, line 35: Can you add doc comment that explain what it means and why it may be needed?

polina-c (approved):
One comment then LGTM

PR #6405, opened 2023-09-15
Adds more deeplink apis

chunhtai:
Adds 3 more api

androidAppLinkSettings
iosBuildOptions
iosUniversalLinkSettings

## Pre-launch Checklist

- [ ] I signed the [CLA].

![build.yaml badge]

### DIFF
--- a/packages/devtools_app/lib/src/shared/config_specific/server/_server_stub.dart
+++ b/packages/devtools_app/lib/src/shared/config_specific/server/_server_stub.dart
@@ -6,6 +6,7 @@
 // ignore_for_file: avoid-unused-parameters
 import 'dart:async';
 
+import 'package:devtools_shared/devtools_deeplink.dart';
 import 'package:devtools_shared/devtools_extensions.dart';
 
 import '../../development_helpers.dart';
@@ -98,6 +99,22 @@ Future<ExtensionEnabledState> extensionEnabledState({
 
 Future<List<String>> requestAndroidBuildVariants(String path) async =>
     const <String>[];
+
+Future<AppLinkSettings> requestAndroidAppLinkSettings(
+  String path, {
+  required String buildVariant,
+}) async =>
+    AppLinkSettings.empty;
+
+Future<XcodeBuildOptions> requestIosBuildOptions(String path) async =>
+    XcodeBuildOptions.empty;
+
+Future<UniversalLinkSettings> requestIosUniversalLinkSettings(
+  String path, {
+  required String configuration,
+  required String target,
+}) async =>
+    UniversalLinkSettings.empty;
 
 void logWarning() {
   throw Exception(unsupportedMessage);
--- a/packages/devtools_app/lib/src/shared/config_specific/server/_server_web.dart
+++ b/packages/devtools_app/lib/src/shared/config_specific/server/_server_web.dart
@@ -9,6 +9,7 @@ import 'dart:convert';
 import 'dart:html';
 
 import 'package:collection/collection.dart';
+import 'package:devtools_shared/devtools_deeplink.dart';
 import 'package:devtools_shared/devtools_extensions.dart';
 import 'package:devtools_shared/devtools_shared.dart';
 import 'package:logging/logging.dart';
@@ -423,10 +424,74 @@ Future<List<String>> requestAndroidBuildVariants(String path) async {
     if (resp?.status == HttpStatus.ok) {
       return json.decode(resp!.responseText!);
     } else {
-      logWarning(resp, apiResetDevTools);
+      logWarning(resp, DeeplinkApi.androidBuildVariants);
     }
   }
   return const <String>[];
+}
+
+Future<AppLinkSettings> requestAndroidAppLinkSettings(
+  String path, {
+  required String buildVariant,
+}) async {
+  if (isDevToolsServerAvailable) {
+    final uri = Uri(
+      path: DeeplinkApi.androidAppLinkSettings,
+      queryParameters: {
+        DeeplinkApi.deeplinkRootPathPropertyName: path,
+        DeeplinkApi.androidBuildVariantPropertyName: buildVariant,
+      },
+    );
+    final resp = await request(uri.toString());
+    if (resp?.status == HttpStatus.ok) {
+      return AppLinkSettings.fromJson(resp!.responseText!);
+    } else {
+      logWarning(resp, DeeplinkApi.androidAppLinkSettings);
+    }
+  }
+  return AppLinkSettings.empty;
+}
+
+Future<XcodeBuildOptions> requestIosBuildOptions(String path) async {
+  if (isDevToolsServerAvailable) {
+    final uri = Uri(
+      path: DeeplinkApi.iosBuildOptions,
+      queryParameters: {
+        DeeplinkApi.deeplinkRootPathPropertyName: path,
+      },
+    );
+    final resp = await request(uri.toString());
+    if (resp?.status == HttpStatus.ok) {
+      return XcodeBuildOptions.fromJson(resp!.responseText!);
+    } else {
+      logWarning(resp, DeeplinkApi.iosBuildOptions);
+    }
+  }
+  return XcodeBuildOptions.empty;
+}
+
+Future<UniversalLinkSettings> requestIosUniversalLinkSettings(
+  String path, {
+  required String configuration,
+  required String target,
+}) async {
+  if (isDevToolsServerAvailable) {
+    final uri = Uri(
+      path: DeeplinkApi.iosUniversalLinkSettings,
+      queryParameters: {
+        DeeplinkApi.deeplinkRootPathPropertyName: path,
+        DeeplinkApi.xcodeConfigurationPropertyName: configuration,
+        DeeplinkApi.xcodeTargetPropertyName: target,
+      },
+    );
+    final resp = await request(uri.toString());
+    if (resp?.status == HttpStatus.ok) {
+      return UniversalLinkSettings.fromJson(resp!.responseText!);
+    } else {
+      logWarning(resp, DeeplinkApi.iosUniversalLinkSettings);
+    }
+  }
+  return UniversalLinkSettings.empty;
 }
 
 void logWarning(HttpRequest? response, String apiType, [String? respText]) {
--- a/packages/devtools_shared/CHANGELOG.md
+++ b/packages/devtools_shared/CHANGELOG.md
@@ -1,4 +1,10 @@
+# 4.2.0
+
+- Adds `DeeplinkApi.androidAppLinkSettings`, `DeeplinkApi.iosBuildOptions`, and
+  `DeeplinkApi.iosUniversalLinkSettings` endpoints to ServerApi.
+
 # 4.1.0
+
 - Add shared integration test utilities to `package:devtools_shared`. These test
 utilities are exported as part of the existing `devtools_test_utils.dart` library.
 
--- a/packages/devtools_shared/lib/src/deeplink/app_link_settings.dart
+++ b/packages/devtools_shared/lib/src/deeplink/app_link_settings.dart
@@ -1,0 +1,62 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:convert';
+
+/// The app link related settings of a Android build of a Flutter project.
+class AppLinkSettings {
+  const AppLinkSettings._(this.applicationId, this.deeplinks);
+
+  factory AppLinkSettings.fromJson(String json) {
+    final jsonObject = jsonDecode(json);
+    return AppLinkSettings._(
+      jsonObject[_kApplicationIdKey] as String,
+      jsonObject[_kDeeplinksKey]
+          .map<AndroidDeeplink>(AndroidDeeplink._fromJsonObject),
+    );
+  }
+
+  /// Used when the the server can't retrieve app link settings.
+  static const empty = AppLinkSettings._('', <AndroidDeeplink>[]);
+
+  static const _kApplicationIdKey = 'applicationId';
+  static const _kDeeplinksKey = 'deeplinks';
+
+  /// The application id of the Android build of this Flutter project.
+  final String applicationId;
+
+  /// The supported deep link of the Android build of this Flutter project.
+  ///
+  /// This list also include deeplinks with custom scheme.
+  final List<AndroidDeeplink> deeplinks;
+}
+
+/// A deep link in a Android build of a Flutter project.
+///
+/// The deeplink is defined in intent filters of AndroidManifest.xml in the
+/// Android sub-project.
+class AndroidDeeplink {
+  AndroidDeeplink._(this.scheme, this.host, this.path);
+
+  factory AndroidDeeplink._fromJsonObject(Map<String, Object?> object) {
+    return AndroidDeeplink._(
+      object[_kSchemeKey] as String,
+      object[_kHostKey] as String,
+      object[_kPathKey] as String,
+    );
+  }
+
+  static const _kSchemeKey = 'scheme';
+  static const _kHostKey = 'host';
+  static const _kPathKey = 'path';
+
+  /// The scheme section of the deeplink.
+  final String scheme;
+
+  /// The host section of the deeplink.
+  final String host;
+
+  /// The path pattern section of the deeplink.
+  final String path;
+}
--- a/packages/devtools_shared/lib/src/deeplink/deeplink_manager.dart
+++ b/packages/devtools_shared/lib/src/deeplink/deeplink_manager.dart
@@ -6,52 +6,196 @@ import 'dart:convert';
 import 'dart:io';
 
 import 'package:meta/meta.dart';
+import 'package:path/path.dart' as path;
 
 class DeeplinkManager {
-  /// A regex to retrieve the json part of the stdout.
+  /// A regex to retrieve the json part from the stdout of Android analyzer.
   ///
   /// Example stdout:
   ///
   /// Running Gradle task 'printBuildVariants'...                        10.4s
   /// ["debug","release","profile"]
-  static final _buildVariantJsonRegex = RegExp(r'(\[.*\])');
+  static final _androidBuildVariantJsonRegex = RegExp(r'(\[.*\])');
+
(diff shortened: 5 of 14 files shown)


packages/devtools_app/lib/src/shared/config_specific/server/_server_web.dart
```diff
@@ -420,6 +421,70 @@ Future<List<String>> requestAndroidBuildVariants(String path) async {
...
+      queryParameters: {
+        DeeplinkApi.deeplinkRootPathPropertyName: path,
+        DeeplinkApi.androidBuildVariantPropertyName: buildVariant,
+      },
+    );
+    final resp = await request(uri.toString());
+    if (resp?.status == HttpStatus.ok) {
+      return AppLinkSettings.fromJson(resp!.responseText!);
+    } else {
+      logWarning(resp, apiResetDevTools);
```
kenzieschmoll, line 440: this should be DeeplinkApi.androidAppLinkSettings instead of apiResetDevTools. Similar for updating the logWarning calls below

packages/devtools_shared/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# NEXT
```
kenzieschmoll, line 1: 4.1.0? This isn't a breaking change so a minor bump should be okay. Can you update the pubspec.yaml file too?

packages/devtools_shared/lib/src/deeplink/app_link_settings.dart
```diff
@@ -0,0 +1,50 @@
...
+  /// The application id of the Android build of this Flutter project.
+  final String applicationId;
+
+  /// The supported deep link of the Android build of this Flutter project.
+  ///
+  /// This list also include deeplinks with custom scheme.
+  final List<AndroidDeeplink> deeplinks;
+}
+
+class AndroidDeeplink {
```
kenzieschmoll, line 31: dart doc

packages/devtools_app/lib/src/shared/config_specific/server/_server_stub.dart
```diff
@@ -102,6 +103,22 @@ Future<ExtensionEnabledState> extensionEnabledState({
 Future<List<String>> requestAndroidBuildVariants(String path) async =>
     const <String>[];
 
+Future<AppLinkSettings> requestAndroidAppLinkSettings(
+  String path, {
+  required String buildVariant,
+}) async =>
+    AppLinkSettings.empty;
+
+Future<XcodeBuildOptions> requestIosBuildOptions(String path) async =>
```
kenzieschmoll, line 112: should this be "variants" instead of "options" to match the android endpoint?

chunhtai: In iOS, it is call build options. `variant` is a keyword in Android gradle build system.

packages/devtools_shared/lib/src/deeplink/deeplink_manager.dart
```diff
@@ -54,4 +64,122 @@ class DeeplinkManager {
...
+    try {
+      jsonEncode(outputJson);
+    } on Error catch (e) {
+      return <String, String?>{
+        kErrorField: e.toString(),
+      };
+    }
+    return <String, String?>{
+      kOutputJsonField: outputJson,
+    };
```
kenzieschmoll, line 106: Most of this is duplicated above and in the methods below. Can we use a helper method to de-dupe?

packages/devtools_shared/lib/src/server/server_api.dart
```diff
@@ -391,7 +412,95 @@ abstract class _DeeplinkApiHandler {
...
+
+    final rootPath = queryParams[DeeplinkApi.deeplinkRootPathPropertyName]!;
+    final result = await deeplinkManager.getIosBuildOptions(rootPath: rootPath);
+    final error = result[DeeplinkManager.kErrorField] as String?;
+    if (error != null) {
+      api.serverError(error);
+    }
+    return api.getCompleted(
+      result[DeeplinkManager.kOutputJsonField]! as String,
+    );
```
kenzieschmoll, line 479: this block is duplicated a few times. Can we use a helper `resultOutputOrError(result)`?

packages/devtools_shared/lib/src/deeplink/deeplink_manager.dart
```diff
@@ -54,4 +64,122 @@ class DeeplinkManager {
       kOutputJsonField: outputJson,
     };
   }
+
+  Future<Map<String, Object?>> getAndroidAppLinkSettings({
+    required String rootPath,
+    required String buildVariant,
+  }) async {
+    final ProcessResult result = await runProcess(
+      'flutter',
```
chunhtai, line 73: @DanTup I remembered you mentioned we should have a way to locate flutter in dart app. The issue https://github.com/Dart-Code/Dart-Code/issues/4722 is closed. Is there now a way to get the path flutter binary?

DanTup: Yep, see https://github.com/Dart-Code/Dart-Code/issues/4722#issuecomment-1722276722 - that issue has been implemented and is available in the current pre-release version of the extension. If you're using that. the `FLUTTER_ROOT` env variable should be available.

Let me know if this doesn't seem to be the case.

packages/devtools_shared/lib/src/deeplink/deeplink_manager.dart
```diff
@@ -8,26 +8,84 @@ import 'dart:io';
...
   }
 
-  Future<Map<String, Object?>> getBuildVariants({
+  Future<ProcessResult> _runFlutterCommand(List<String> arguments) {
+    final String flutterRoot = Platform.environment['FLUTTER_ROOT']!;
+    return runProcess(
+      'flutter',
+      arguments: arguments,
+      workingDirectory: flutterRoot,
+    );
```
DanTup, line 50: `FLUTTER_ROOT` will be the SDK folder (without `bin`), so you probably need to combine these.

I'm also not sure that `workingDirectory` is always searched for the executable, so I think it should be prepended to the `executable`. So you might need something like:

```dart
    return runProcess(
      path.join(flutterRoot, 'bin', Platform.isWindows ? 'flutter.bat' : 'flutter'),
      arguments: arguments,
    );
```

I would also suggest not using `!` to assume this will always be set, in case DevTools is running in a context other than VS Code, or a version of the Dart plugin that doesn't include https://github.com/Dart-Code/Dart-Code/issues/4722 (such as the current stable version).

You should be able to test this using the pre-release extension, but let me know if that doesn't seem to be the case.

Dart pre-release versions

packages/devtools_app/release_notes/NEXT_RELEASE_NOTES.md
```diff
@@ -20,6 +20,9 @@ DevTools extension for your pub package, see the getting started guide for
 
 * Fixed theming bug in isolate selector - [#6403](https://github.com/flutter/devtools/pull/6403)
 
+* Adds `requestAndroidAppLinkSettings`, `requestIosBuildOptions`, and
+  `requestIosUniversalLinkSettings` endpoints to ServerApi.
+
```
kenzieschmoll, line 25: The release notes are intended for user-facing changes, so let's hold off on adding this note here until we ship the deep link feature in general. The entry you added to `devtools_shared/CHANGELOG.md` is sufficient. To get the release notes check on the bots, add the label "release notes not required"

packages/devtools_shared/lib/src/deeplink/app_link_settings.dart
```diff
@@ -0,0 +1,55 @@
...
+
+/// The app link related settings of a Android build of a Flutter project.
+class AppLinkSettings {
+  const AppLinkSettings._(this.applicationId, this.deeplinks);
+
+  factory AppLinkSettings.fromJson(String json) {
+    final jsonObject = jsonDecode(json);
+    return AppLinkSettings._(
+      jsonObject['applicationId'] as String,
+      jsonObject['deeplinks']
```
kenzieschmoll, line 15: please put 'applicationId' and 'deeplinks' in static consts on this class.

packages/devtools_shared/lib/src/deeplink/app_link_settings.dart
```diff
@@ -0,0 +1,55 @@
...
+
+  /// The supported deep link of the Android build of this Flutter project.
+  ///
+  /// This list also include deeplinks with custom scheme.
+  final List<AndroidDeeplink> deeplinks;
+}
+
+/// A deep link in a Android build of a Flutter project.
+///
+/// This deeplink are defined in intent filters of AndroidManifest.xml in the
```
kenzieschmoll, line 34: nit: r/are/is

packages/devtools_shared/lib/src/deeplink/app_link_settings.dart
```diff
@@ -0,0 +1,55 @@
...
+/// Android sub-project.
+class AndroidDeeplink {
+  AndroidDeeplink._(this.scheme, this.host, this.path);
+
+  factory AndroidDeeplink._fromJsonObject(Map<String, Object?> object) {
+    return AndroidDeeplink._(
+      object['scheme'] as String,
+      object['host'] as String,
+      object['path'] as String,
```
kenzieschmoll, line 43: please put these strings in constants on this class

packages/devtools_shared/lib/src/deeplink/deeplink_manager.dart
```diff
@@ -8,26 +8,84 @@ import 'dart:io';
...
+  ///
+  /// Example stdout:
+  ///
+  /// {"configurations":["Debug","Release","Profile"],"targets":["Runner","RunnerTests"]}
+  static final _iosBuildOptionsJsonRegex = RegExp(r'({.*})');
+
   static const kErrorField = 'error';
   static const kOutputJsonField = 'json';
 
+  static final _outputFilePathRegex = RegExp(r'result saved in (.*.json)');
```
kenzieschmoll, line 29: can you add dart doc to this?

packages/devtools_shared/lib/src/deeplink/deeplink_manager.dart
```diff
@@ -54,4 +112,69 @@ class DeeplinkManager {
...
+        'analyze',
+        '--android',
+        '--output-app-link-settings',
+        '--build-variant=$buildVariant',
+        rootPath,
+      ],
+    );
+  }
+
+  Future<Map<String, Object?>> getIosBuildOptions({
```
kenzieschmoll, line 131: this method can use `_runFlutterCommandAndReadOutputFile` if we add two parameters `regex` and `void Function(ProcessResult) noMatchError`

```dart
return _runFlutterCommandAndReadOutputFile(
  ['analyze', '--ios', '--list-build-options', rootPath],
  regex: _iosBuildOptionsJsonRegex,
  (result) => 'Can not parse output ${result.stdout}',
);
```

packages/devtools_shared/lib/src/deeplink/universal_link_settings.dart
```diff
@@ -0,0 +1,35 @@
...
+    this.teamIdentifier,
+    this.associatedDomains,
+  );
+
+  factory UniversalLinkSettings.fromJson(String json) {
+    final jsonObject = jsonDecode(json);
+    return UniversalLinkSettings._(
+      jsonObject['bundleIdentifier'] as String? ?? '',
+      jsonObject['teamIdentifier'] as String? ?? '',
+      jsonObject['associatedDomains'].cast<String>() as List<String>,
```
kenzieschmoll, line 20: please put strings in static consts. Here and elsewhere.

packages/devtools_shared/test/server/server_api_test.dart
```diff
@@ -36,9 +36,12 @@ void main() {
     );
     expect(response.statusCode, HttpStatus.ok);
     expect(await response.readAsString(), '["debug", "release]');
+    expect(fakeManager.receivedPath, expectedRootPath);
   });
 
-  test('handle deeplink api returns bad request if no root path', () async {
+  test(
+      'handle deeplink api androidBuildVariants returns bad request if no root path',
```
kenzieschmoll, line 43: nit: in test names, consider using the the enum values directly so the test names don't go stale:
`'${DeeplinkApi.androidBuildVariants} ...'`

here and elsewhere below

packages/devtools_shared/lib/src/deeplink/deeplink_manager.dart
chunhtai: I added the logic to get the proper path and also handler when env var is not set

packages/devtools_shared/lib/src/deeplink/deeplink_manager.dart
```diff
@@ -23,150 +24,140 @@ class DeeplinkManager {
...
-    List<String> arguments,
-  ) async {
-    final ProcessResult result = await _runFlutterCommand(arguments);
+  Future<String> _runFlutterCommand(
+    List<String> arguments, {
+    required RegExp outputMatcher,
+  }) async {
+    final flutterPath = getFlutterBinary();
+    if (flutterPath == null) {
+      throw 'Running Flutter command is not supported in this devtool version';
```
kenzieschmoll, line 73: Is it really the DevTools version that determines this? We should be more clear about which version is preventing the `FLUTTER_ROOT` environment variable from being set. In other words, is this due to the Dart SDK version, the Flutter SDK version, etc.

packages/devtools_shared/lib/src/deeplink/deeplink_manager.dart
```diff
@@ -23,150 +24,140 @@ class DeeplinkManager {
...
+    }
+    final result = await runProcess(flutterPath, arguments: arguments);
     if (result.exitCode != 0) {
-      return <String, String>{
-        kErrorField:
-            'Flutter command exit with non-zero error code ${result.exitCode}\n${result.stderr}',
-      };
+      throw 'Flutter command exit with non-zero error code ${result.exitCode}\n${result.stderr}';
```
kenzieschmoll, line 77: can we throw a specific type of exception instead of throwing a string? e.g. maybe ProcessException here and FormatException below. Here and elsewhere where we are throwing a String

packages/devtools_shared/test/server/server_api_test.dart
```diff
@@ -12,7 +12,7 @@ import 'package:shelf/shelf.dart';
 import 'package:test/test.dart';
 
 void main() {
-  test('handle deeplink api androidBuildVariants', () async {
+  test('handle deeplink api ${DeeplinkApi.androidBuildVariants}', () async {
```
kenzieschmoll, line 15: nit: I think now the "deeplink api" text is redundant since `'${DeeplinkApi.androidBuildVariants}'` should output "DeeplinkApi.androidBuildVariants".

packages/devtools_shared/lib/src/deeplink/deeplink_manager.dart
chunhtai: Hi @DanTup Do you know which dart version the change will be in?

chunhtai: ah I see, there isn't a version we can target in this case, a lot of different cases can result in this situation. For example running the dev tool directly through a standalone dart. I think I will just mention that it can't find a flutter sdk and maybe mention it needs to set FLUTTER_ROOT environment variable

packages/devtools_shared/test/server/server_api_test.dart
chunhtai: The '${DeeplinkApi.androidBuildVariants}' should be evaluate to something like  `'api/androidBuildVariants'`? so I think the deeplink api is still relevant

packages/devtools_shared/lib/src/deeplink/deeplink_manager.dart
kenzieschmoll: Can we just use 'flutter' if we don't find FLUTTER_ROOT? Not as accurate as using FLUTTER_ROOT but we would at least not fail as easily since most users should have `flutter` on their path

kenzieschmoll (approved):
LGTM, left one comment though about defaulting to use 'flutter' if we can't find FLUTTER_ROOT

packages/devtools_shared/lib/src/deeplink/deeplink_manager.dart
DanTup: > Hi @DanTup Do you know which dart version the change will be in?

Version 3.74.0 of the Dart extension for VS Code will include the change to set `FLUTTER_ROOT`. As you noted, this is specific to the server being started by VS Code though, so may not apply if it's run from the command line, etc. so if it's missing, it's worth trying to handle gracefully.

Perhaps before falling back to using `flutter` from `PATH` (which may also not exist, or be a different version if the user is using something like `fvm`), it's worth checking `Platform.resolvedExecutable` (which is the path to the Dart VM that's running) and walk up to see if it's nested inside `bin/cache/dart-sdk/bin` of a Flutter SDK and using that?

chunhtai: I can try to use the flutter from path.

>  it's worth checking Platform.resolvedExecutable (which is the path to the Dart VM that's running) and walk up to see if it's nested inside bin/cache/dart-sdk/bin of a Flutter SDK and using that?

If it is using the dart from the flutter sdk, it should have set the FLUTTER_ROOT I think?

DanTup: > If it is using the dart from the flutter sdk, it should have set the FLUTTER_ROOT I think?

I _think_ if it's using the shell script at `bin/dart` it would, but if it was launched directly using `bin/cache/dart-sdk/bin/dart` I'm not sure it would. I don't know how common each way is though (personally I don't use the shell script because `dart` on my `PATH` is always a standalone Dart SDK, but my setup is likely not the same as most users :-))

kenzieschmoll (approved):

packages/devtools_shared/lib/src/deeplink/deeplink_manager.dart
chunhtai: done

DanTup (approved):
The use of `FLUTTER_ROOT` looks reasonable to me. I'll defer to Kenzie for the rest, as I'm less familiar with this functionality.

Issue #6398, opened 2023-09-15
Make Banner Message behavior configurable by extension


kenzieschmoll:
In DevTools, it is our default that if a user dismisses a banner message that we will not show it to them again. However, this may not be the behavior that an extension author wants. We should make the dismiss behavior configurable in the banner message api for extensions

PR #6406, opened 2023-09-15
Add `ignoreIfAlreadyDismissed` parameter to extensions `showBannerMessage` API

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/6398:

> In DevTools, it is our default that if a user dismisses a banner message that we will not show it to them again. However, this may not be the behavior that an extension author wants. We should make the dismiss behavior configurable in the banner message api for extensions.

With the new API, an extension can call `ExtensionManager.showBannerMessage` with `ignoreIfAlreadyDismissed: false` if they want the banner to be able to be dismissed and re-shown.

This PR also updates the example app to use the new api.


polina-c:
Can you explain in description what is new UX/API?

polina-c (approved):

Issue #6391, opened 2023-09-14
Force reloading extension does not recover from broken state


kenzieschmoll:
Repo - checkout the code in https://github.com/rrousselGit/provider/pull/832.

Run DevTools and connect an app that depends on this extension. Pressing the provider settings button throws an exception, and triggering a force reload does not recover from the broken state
![force-reload-broken](https://github.com/flutter/devtools/assets/43759233/6b8f8dcb-8b4a-4d4e-8f17-69aa06f71112)

kenzieschmoll:
This force reload action is broken in general

PR #6417, opened 2023-09-18
Add `forceReload` endpoint to the extensions API

kenzieschmoll:
We cannot call `location.reload()` directly on the iFrame object due to cross origin protections. In order to do a full reload, we need the iFrame to reload itself. This PR adds an endpoint to the extensions API to do so, and calls this endpoint from the existing force reload action in DevTools.

Part of fixing the hot reload action required fixing a state management issue where we were calling didUpdateWidget and cancelling a stream subscription when we shouldn't be. Fix for this is implementing equality and hashCode for DevToolsExtensionConfig, which is part of this PR.

Fixes https://github.com/flutter/devtools/issues/6391.

### DIFF
--- a/packages/devtools_app/lib/src/extensions/embedded/_controller_web.dart
+++ b/packages/devtools_app/lib/src/extensions/embedded/_controller_web.dart
@@ -100,7 +100,7 @@ class EmbeddedExtensionControllerImpl extends EmbeddedExtensionController
   }) {
     extensionPostEventStream.add(DevToolsExtensionEvent(type, data: data));
   }
-  
+
   @override
   void dispose() async {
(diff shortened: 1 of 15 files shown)


polina-c (approved):
Should integration testing be added on this?

kenzieschmoll:
> Should integration testing be added on this?

The force reload action is pressed in the existing DevTools extensions integration test. Testing what the iFrame itself does is more difficult since we don't have access to the iFrame web app from the FlutterTester device in the devtools integration test. I filed https://github.com/flutter/devtools/issues/6419 to add a test to the simulated environment, where we would have access to what the iFrame is doing. I will test cover the whole extensions API there.

Issue #6416, opened 2023-09-18
Include the VM service uri in the url for the simulated DevTools environment for extensions


kenzieschmoll:
This will ensure that the app connection persists when using hot restart in the simulated devtools environment

Issue #6389, opened 2023-09-14
Extension does not respect devtools theme when force reloading


kenzieschmoll:
on dark theme --> force reload --> extension loads in light theme (inconsistent repro)
![Screenshot 2023-09-14 at 1 10 48 PM](https://github.com/flutter/devtools/assets/43759233/359b047e-4e9c-40ad-8eb1-09a7ec9ebda2)

PR #6420, opened 2023-09-18
Update the extension iFrame's query parameters for theme and vm service uri changes

kenzieschmoll:
The extension iFrame's URL query parameters are now updated when the theme changes and when the vm service uri changes. This ensures the theme and service connection state are preserved on reload of the iFrame window.

Fixes https://github.com/flutter/devtools/issues/6389
Fixes https://github.com/flutter/devtools/issues/6416

### DIFF
--- a/packages/devtools_extensions/lib/src/template/extension_manager.dart
+++ b/packages/devtools_extensions/lib/src/template/extension_manager.dart
@@ -45,8 +45,7 @@ class ExtensionManager {
     // TODO(kenz): handle the ide theme that may be part of the query params.
     final queryParams = loadQueryParams();
     final themeValue = queryParams[ExtensionEventParameters.theme];
-    darkThemeEnabled.value = themeValue == null ||
-        themeValue == ExtensionEventParameters.themeValueDark;
+    _setThemeForValue(themeValue);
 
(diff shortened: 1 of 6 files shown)


packages/devtools_extensions/lib/src/template/extension_manager.dart
```diff
@@ -167,20 +171,40 @@ class ExtensionManager {
...
-      print('Unable to connect to VM service at $vmServiceUri: $e');
+      final errorMessage =
+          'Unable to connect extension to VM service at $vmServiceUri: $e';
+      showNotification('Error: $errorMessage');
+      _log.shout(errorMessage);
     }
   }
 
+  void _setThemeForValue(String? themeValue) {
+    // Default to use dark theme if [value] is null.
```
polina-c, line 192: Do we always default to dark theme? 
Should it be a constant useDarkThemeDefault somewhere in shared constants?

polina-c (approved):
lgtm with one comment

packages/devtools_extensions/lib/src/template/extension_manager.dart
kenzieschmoll: Good idea. Done.

PR #6424, opened 2023-09-19
Tidy up sidebar + use screen metadata

DanTup:
This switches to using the screen metadata for the sidebar menu rather than a hard-coded list. Also has some tidying up suck as detecting web devices and showing tooltips for disabled items on the menu.

https://github.com/flutter/devtools/assets/1078012/b02c6019-0285-47a3-8c1d-895dc7c87f01

### DIFF
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -343,8 +343,7 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
             ),
           );
         },
-      if (FeatureFlags.vsCodeSidebarTooling)
-        if (FeatureFlags.vsCodeSidebarTooling) ..._standaloneScreens,
+      if (FeatureFlags.vsCodeSidebarTooling) ..._standaloneScreens,
     };
   }
 
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -343,14 +343,9 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
             ),
           );
         },
-      if (FeatureFlags.vsCodeSidebarTooling) ..._standaloneScreens,
-    };
-  }
-
-  Map<String, UrlParametersBuilder> get _standaloneScreens {
-    return {
-      for (final type in StandaloneScreenType.values)
-        type.name: (_, __, args, ___) => type.screen,
+      if (FeatureFlags.vsCodeSidebarTooling)
+        for (final type in StandaloneScreenType.values)
+          type.name: (_, __, args, ___) => type.build(_screens),
     };
   }
 
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -111,6 +111,8 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
         (e) => DevToolsScreen<void>(ExtensionScreen(e)).screen,
       );
 
+  // TODO(dantup): This does not take IDE preference into account, so results
+  //  in Dark mode embedded sidebar in VS Code.
   bool get isDarkThemeEnabled => _isDarkThemeEnabled;
   bool _isDarkThemeEnabled = true;
 
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -344,8 +344,14 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
           );
         },
       if (FeatureFlags.vsCodeSidebarTooling)
-        for (final type in StandaloneScreenType.values)
-          type.name: (_, __, args, ___) => type.build(_screens),
+        if (FeatureFlags.vsCodeSidebarTooling) ..._standaloneScreens,
+    };
+  }
+
+  Map<String, UrlParametersBuilder> get _standaloneScreens {
+    return {
+      for (final type in StandaloneScreenType.values)
+        type.name: (_, __, args, ___) => type.screen,
     };
   }
 
--- a/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
@@ -132,29 +132,24 @@ class _DevToolsMenu extends StatelessWidget {
         disabledReason = 'Not available when running on the web';
       }
 
-      Widget button = TextButton.icon(
-        style: TextButton.styleFrom(
-          alignment: Alignment.centerRight,
-          shape: const ContinuousRectangleBorder(),
-        ),
-        onPressed: disabledReason != null
-            ? null
-            : () => unawaited(api.openDevToolsPage(session.id, screen.id)),
-        label: Directionality(
-          textDirection: normalDirection,
-          child: Text(title),
-        ),
-        icon: Icon(screen.icon, size: actionsIconSize),
+      // Because we flipped the direction so the menu is aligned to the end, we
+      // should revert the text direction back to normal for the label.
+      Widget text = Directionality(
+        textDirection: normalDirection,
+        child: Text(title),
       );
 
       if (disabledReason != null) {
-        button =
-            Tooltip(preferBelow: false, message: disabledReason, child: button);
+        text =
+            Tooltip(preferBelow: false, message: disabledReason, child: text);
       }
 
-      return SizedBox(
-        width: double.infinity,
-        child: button,
+      return MenuItemButton(
+        leadingIcon: Icon(screen.icon, size: actionsIconSize),
+        onPressed: disabledReason != null
+            ? null
+            : () => unawaited(api.openDevToolsPage(session.id, screen.id)),
+        child: text,
       );
     }
 
--- a/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
@@ -12,10 +12,11 @@ import '../../shared/screen.dart';
 import '../api/vs_code_api.dart';
 
 class DebugSessions extends StatelessWidget {
-  const DebugSessions(this.api, this.sessions, {super.key});
+  const DebugSessions(this.api, this.sessions, this.deviceMap, {super.key});
 
   final VsCodeApi api;
   final List<VsCodeDebugSession> sessions;
+  final Map<String, VsCodeDevice> deviceMap;
 
   @override
   Widget build(BuildContext context) {
@@ -51,8 +52,7 @@ class DebugSessions extends StatelessWidget {
     final isProfile = mode == 'profile';
     final isRelease = mode == 'release' || mode == 'jit_release';
     final isFlutter = session.debuggerType?.contains('Flutter') ?? false;
-    // TODO(dantup): Detect web so we can handle requiresDartVm correctly.
-    const isWeb = false;
+    final isWeb = deviceMap[session.flutterDeviceId]?.platformType == 'web';
 
     final label = session.flutterMode != null
         ? '${session.name} (${session.flutterMode})'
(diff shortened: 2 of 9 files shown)


packages/devtools_app/lib/src/app.dart
```diff
@@ -343,14 +343,9 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
...
-    };
-  }
-
-  Map<String, UrlParametersBuilder> get _standaloneScreens {
-    return {
-      for (final type in StandaloneScreenType.values)
-        type.name: (_, __, args, ___) => type.screen,
+      if (FeatureFlags.vsCodeSidebarTooling)
+        for (final type in StandaloneScreenType.values)
+          type.name: (_, __, args, ___) => type.build(_screens),
```
DanTup, line 348: @kenzieschmoll I had to inline this to be able to pass screens through - but this doesn't feel very general (other standalone screens might not want screens). Is there something better we could do?

I did wonder if we should move some of the conditions like `requiresDartVm` onto `ScreenMetadata` and then have a full set of those, so we don't actually need the `Screens` to render buttons, only the set of metadata. This could simplify a few places like the screen constructors that all pass name/icon/conditions through to the base - they could just pass their `ScreenMetadata` to the base.

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -112,10 +117,18 @@ class _DevToolsMenu extends StatelessWidget {
...
-      bool enabled = true,
-    }) {
+    Widget? devToolsButton(Screen screen) {
+      // Don't include any screens that aren't appropriate.
+      if (!screen.requiresConnection ||
+          screen.requiresLibrary != null ||
+          screen.requiresVmDeveloperMode ||
+          screen.screenId == 'debugger') {
+        return null;
+      }
```
DanTup, line 127: Should we have a flag somewhere (like `ScreenMetadata` if we made the change above) for whether something should show up in the sidebar menu? It's not appropriate to show things like Home, Debugger.

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -143,36 +157,12 @@ class _DevToolsMenu extends StatelessWidget {
...
-            ScreenMetaData.network,
-            enabled: isDebug,
-          ),
-          devToolsButton(
-            ScreenMetaData.logging,
-          ),
-          // TODO(dantup): Check other screens (like appSize) work embedded and
-          //  add here.
-        ],
+        // TODO(dantup): Why is appSize etc. still missing?
```
DanTup, line 160: When I run in the stager app, we use `defaultScreens()` which doesn't include everything (like app size), so I'm sure that's the right thing to use.

kenzieschmoll:
Given our offline discussion this morning, I'll wait for those changes before reviewing. Thanks!

packages/devtools_app/lib/src/app.dart
DanTup: This was reverted in favour of the menu just using `ScreenMetaData.values`.

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
DanTup: No long applicable, we're using `ScreenMetaData.values` inside the menu.

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
DanTup: We're skipping those that need libraries for now (only provider) and will implement support for extensions in this menu by querying the server later.

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -12,10 +12,11 @@ import '../../shared/screen.dart';
 import '../api/vs_code_api.dart';
 
 class DebugSessions extends StatelessWidget {
-  const DebugSessions(this.api, this.sessions, {super.key});
+  const DebugSessions(this.api, this.sessions, this.deviceMap, {super.key});
```
kenzieschmoll, line 15: nit: can we make these required and named so that constructing `DebugSessions` is more readable?

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -112,26 +119,37 @@ class _DevToolsMenu extends StatelessWidget {
...
+      // Because we flipped the direction so the menu is aligned to the end, we
+      // should revert the text direction back to normal for the label.
+      Widget text = Directionality(
+        textDirection: normalDirection,
+        child: Text(title),
+      );
+
+      if (disabledReason != null) {
+        text =
+            Tooltip(preferBelow: false, message: disabledReason, child: text);
```
kenzieschmoll, line 144: nit: trailing comma at the end of this

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -191,4 +182,16 @@ class _DevToolsMenu extends StatelessWidget {
       ),
     );
   }
+
+  bool _shouldIncludeScreen(ScreenMetaData screen) {
+    // Don't show home screen or debugger in the menu.
+    return screen != ScreenMetaData.home &&
+        screen != ScreenMetaData.debugger &&
+        // Or the generic "simple" screen.
```
kenzieschmoll, line 190: do we also need to hide the provider screen here? We might also want to hide `vmTools` and `deepLinks`.

`vmTools` should be gated behind a DevTools setting `preferences.vmDeveloperModeEnabled`. Since we have the server available we could check for that setting. Or we can just hide for now since that screen is intended for VM authors.

`deepLinks` is behind an experiment flag, so we should check that `FeatureFlags.deepLinkValidation` is true before including it.

DanTup:
@kenzieschmoll I've pushed some changes to use `ScreenMetaData.values`, and also some other minor cleanup.

I've edited a video of how this currently looks into the top comment above.

DanTup:
(@inmatrix @anderdobo FYI - there's a video of the current state at the top if you want to get a headstart on compiling any feedback before we meet)

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
DanTup: > do we also need to hide the provider screen here?

It's hidden by the `requiresLibrary` check further down. We could add it explicitly, but right now we don't support any screens that have library requirements (and I expect we don't bother, and instead will support them via extensions later).

I've hidden `vmTools` for now (to avoid making async calls out to check this), and also added a check for the feature flag for deep links (I meant to ask about this - I noticed it was showing a TODO when I clicked it!).

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
DanTup: Done!

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
DanTup: Done!

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -184,11 +193,15 @@ class _DevToolsMenu extends StatelessWidget {
...
   bool _shouldIncludeScreen(ScreenMetaData screen) {
-    // Don't show home screen or debugger in the menu.
+    // Some screens shouldn't show up in the menu.
     return screen != ScreenMetaData.home &&
         screen != ScreenMetaData.debugger &&
+        screen != ScreenMetaData.vmTools &&
         // Or the generic "simple" screen.
         screen != ScreenMetaData.simple &&
+        // DeepLink is currently behind a feature flag.
+        (screen != ScreenMetaData.deepLinks ||
```
kenzieschmoll, line 203: this should be ==

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -184,11 +193,15 @@ class _DevToolsMenu extends StatelessWidget {
   }
 
   bool _shouldIncludeScreen(ScreenMetaData screen) {
-    // Don't show home screen or debugger in the menu.
+    // Some screens shouldn't show up in the menu.
     return screen != ScreenMetaData.home &&
         screen != ScreenMetaData.debugger &&
+        screen != ScreenMetaData.vmTools &&
```
kenzieschmoll, line 199: can you add a TODO here to consider showing this conditionally based on `preferences.vmDeveloperModeEnabled`? Or we could do that here if it is simple

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
DanTup: I think it's correct - this is returning whether the screen should be included, and it can be included if it's not `deepLinks`, but if it is `deepLinks` we need the flag set.

Perhaps this multiline condition wasn't the simplest to read though, I've changed it to a switch expression which I think is much clearer.

kenzieschmoll: sorry I read this wrong. We might want to split this return statement up to make it more readable, and to make it less likely that logic bugs creep in. Something like this:

```dart
// Hide anything that requires some specific library that we can't check for.
if (screen.requiresLibrary != null) {
  return false;
}

return switch (screen) {
  ScreenMetaData.home ||
  ScreenMetaData.debugger ||
  ScreenMetaData.simple => false,
  ScreenMetaData.vmTools => preferences.vmDeverloperModeEnabled.value,
  ScreenMetaData.deepLinks => FeatureFlags.deepLinkValidation,
  _ => true,
};
```

kenzieschmoll: Returning `preferences.vmDeverloperModeEnabled.value` will only pick up the first value (e.g. won't listen for changes to the notifier), but my guess is that this setting is not something users turn on and off frequently. I expect advanced dart developers on the VM team will always just have this enabled, and the average user will not.

DanTup: Snap 😄

> `preferences.vmDeverloperModeEnabled.value`

I presume the preferences are loaded asynchronously, so if this menu was opened before that completed, this might not be right? Realistically that's unlikely though (you'd need to start a debug session before you could even get to the menu) so if you think it's reasonable to just read the current `value` I'll add this :-)

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -193,18 +193,18 @@ class _DevToolsMenu extends StatelessWidget {
...
-        // check (`requiresLibrary` will be removed for extensions
-        // in future, and those will need some extra work to detect).
-        screen.requiresLibrary == null;
+    return switch (screen) {
+      // Some screens shouldn't show up in the menu.
+      ScreenMetaData.home => false,
+      ScreenMetaData.debugger => false,
+      ScreenMetaData.simple => false, // generic screen isn't a screen itself
+      // TODO(dantup): Check preferences.vmDeveloperModeEnabled
+      ScreenMetaData.vmTools => false,
```
kenzieschmoll, line 202: these cases can be joined with with the `||` operator (see my previous comment)

kenzieschmoll (approved):
one last comment to use || in the switch expression, then lgtm

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
DanTup: I did that originally, but I didn't like how the formatter formatted it 😄

image

If you think that's better though, I'll commit that before merging. Thanks!

kenzieschmoll: Doesn't matter too much. LGTM

PR #6433, opened 2023-09-20
Add an integration test for `package:devtools_extensions`

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/6419


.github/workflows/build.yaml
```diff
@@ -163,6 +163,38 @@ jobs:
...
+      - name: Load Cached Flutter SDK
+        uses: actions/cache@704facf57e6136b1bc63b828d79edcd491f0ee84
+        with:
+          path: |
+            ./flutter-sdk
+          key: flutter-sdk-${{ runner.os }}-${{ needs.flutter-prep.outputs.latest_flutter_candidate }}
+      - name: tool/bots.sh
+        env:
+          BOT: ${{ matrix.bot }}
+          DEVTOOLS_PACKAGE: devtools_extensions
```
CoderDake, line 197: do the golden artifacts, below for this workflow, need to point to devtools_extensions/ instead of devtools_app/?

kenzieschmoll: we don't need these at all actually, removed this step

CoderDake (approved):

PR #6434, opened 2023-09-20
Widget inspector should operate on main isolate, not selected isolate

CoderDake:
Fixes https://github.com/flutter/devtools/issues/6357


packages/devtools_app_shared/lib/src/service/eval_on_dart_library.dart
```diff
@@ -45,7 +45,7 @@ class EvalOnDartLibrary extends DisposableController
     _libraryRef = Completer<LibraryRef>();
 
     // For evals in tests, we will pass the isolateId into the constructor.
-    isolate ??= serviceManager.isolateManager.selectedIsolate;
+    isolate ??= serviceManager.isolateManager.mainIsolate;
```
kenzieschmoll, line 48: I don't think we want to do this for all EvalOnDartLibrary instances. Just for the one that the inspector creates

CoderDake: ah yes, sorry went too fast and only looked at the inits 🖖

packages/devtools_app_shared/lib/src/service/eval_on_dart_library.dart
```diff
@@ -41,11 +41,16 @@ class EvalOnDartLibrary extends DisposableController
     ValueListenable<IsolateRef?>? isolate,
     this.disableBreakpoints = true,
     this.oneRequestAtATime = false,
+    this.forceMainIsolate = false,
```
kenzieschmoll, line 44: instead of adding a new parameter, can we just pass `serviceManager.isolateManager.mainIsolate` to the existing `isolate` parameter?

packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
```diff
@@ -42,6 +42,7 @@ abstract class InspectorServiceBase extends DisposableController
           inspectorLibraryUri,
           serviceConnection.serviceManager.service!,
           serviceManager: serviceConnection.serviceManager,
+          isolate: serviceConnection.serviceManager.isolateManager.mainIsolate,
```
kenzieschmoll, line 45: @matt-ragonese will this break your inspector? Do we need to add an `isolate` param to `InspectorServiceBase` instead of assuming we should use the main isolate here?

kenzieschmoll: lets add the isolate parameter (should be an optional named param) to InspectorServiceBase, and then pass in the main isolate there instead of directly on the EvalOnDartLibrary instance.

packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
```diff
@@ -35,14 +35,16 @@ abstract class InspectorServiceBase extends DisposableController
     required this.clientInspectorName,
     required this.serviceExtensionPrefix,
     required String inspectorLibraryUri,
+    IsolateRef? evalIsolate,
```
kenzieschmoll, line 38: make this match the type of the EvalOnDartLibrary isolate parameter

packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
```diff
@@ -189,11 +191,12 @@ abstract class InspectorServiceBase extends DisposableController
 /// Manages communication between inspector code running in the Flutter app and
 /// the inspector.
 class InspectorService extends InspectorServiceBase {
-  InspectorService()
+  InspectorService(IsolateRef? evalIsolate)
       : super(
           clientInspectorName: 'WidgetInspectorService',
           serviceExtensionPrefix: inspectorExtensionPrefix,
           inspectorLibraryUri: inspectorLibraryUri,
+          evalIsolate: evalIsolate,
```
kenzieschmoll, line 199: this can just be `serviceConnection.serviceManager.isolateManager.mainIsolate`

no need to add the evalIsolate parameter to InspectorService

packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
kenzieschmoll: you should be able to remove the change to make isolateLibrary late then too

packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
CoderDake: 🤦 fair enough :P thats what I get for doing this during a meeting :P

packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
```diff
@@ -35,14 +35,16 @@ abstract class InspectorServiceBase extends DisposableController
...
-          serviceConnection.serviceManager.service!,
-          serviceManager: serviceConnection.serviceManager,
-        ) {
+        clients = {} {
+    inspectorLibrary = EvalOnDartLibrary(
+      inspectorLibraryUri,
+      serviceConnection.serviceManager.service!,
+      serviceManager: serviceConnection.serviceManager,
+      isolate: evalIsolate,
+    );
```
kenzieschmoll, line 47: can this go back to being outside of the constructor block so that inspectorLibrary doesn't have to be late?

packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
```diff
@@ -189,11 +191,12 @@ abstract class InspectorServiceBase extends DisposableController
 /// Manages communication between inspector code running in the Flutter app and
 /// the inspector.
 class InspectorService extends InspectorServiceBase {
-  InspectorService()
+  InspectorService({ValueListenable<IsolateRef?>? evalIsolate})
```
kenzieschmoll, line 194: remove this parameter and pass `evalIsolate: serviceConnection.serviceManager.isolateManager.mainIsolate` directly below

kenzieschmoll (approved):

Issue #6435, opened 2023-09-20
main isolate not reselected after hot restart


CoderDake:
When in vm developer mode, when a hot restart is done, the vm_service isolate ends up being selected after the hot restart.

We don't have a way to hook into the hot restart lifecycle directly, so instead implement the following to ensure smooth behaviour on the main isolate:
- if main isolate exits while main isolate is selected
- re-select main isolate if it comes back alive.

PR #6436, opened 2023-09-20
Reselect main isolate when hot restarting

CoderDake:
Fixes https://github.com/flutter/devtools/issues/6435


packages/devtools_app_shared/lib/src/service/isolate_manager.dart
```diff
@@ -19,6 +19,12 @@ base mixin TestIsolateManager implements IsolateManager {}
 
 final class IsolateManager with DisposerMixin {
   final _isolateStates = <IsolateRef, IsolateState>{};
+
+  /// Signifies whether the main isolate should be selected if it is started.
+  /// This is used to make sure the the main isolate remains selected after
+  /// a hot restart.
```
kenzieschmoll, line 25: dart doc nit: add a blank line after the first sentence

kenzieschmoll (approved):

kenzieschmoll:
@bkonyi can you take a look at this and verify our assumptions here are safe?

packages/devtools_app_shared/lib/src/service/isolate_manager.dart
```diff
@@ -129,6 +136,11 @@ final class IsolateManager with DisposerMixin {
 
       if (_selectedIsolate.value == null) {
         _setSelectedIsolate(event.isolate);
+      } else if (_shouldReselectMainIsolate) {
```
bkonyi, line 139: I think this case should be handled first. In theory, a hot restart could result in a couple of isolates coming up.

If we're already assuming that the first isolate to restart is the main isolate and VM developer mode is disabled, `_selectedIsolate.value` will be null and we'll execute the first case. When the second isolate comes up, `_shouldReselectMainIsolate` will still be `true`, so we'll select the second isolate which, based on our assumptions, isn't the main isolate.

kenzieschmoll: We may also want to check the name of the isolate. @bkonyi is it safe to assume the main isolate will always have a name of 'main'? This is what I have seen while debugging but I'm not sure if my local testing covers all the possible cases.

bkonyi: There's no guarantees provided around isolate names as they're completely dependent on the embedder. However, I'd be surprised if the standalone VM or the Flutter engine deviated from using `main` for the first isolate spawned that's running user code. 

Flutter might have scenarios where there's multiple "main" isolates though, assuming that it's possible for multiple rendering surfaces to have their own UI isolates, but I'm not familiar enough with this area to know.

CoderDake: I'd be ok with us assuming for now and adjusting in the future if it turns out there are cases that it misses :)

packages/devtools_app_shared/lib/src/service/isolate_manager.dart
```diff
@@ -127,7 +134,12 @@ final class IsolateManager with DisposerMixin {
       // but that may not always be a safe assumption.
       _mainIsolate.value ??= event.isolate;
 
-      if (_selectedIsolate.value == null) {
+      if (_shouldReselectMainIsolate && event.isolate?.name == 'main') {
```
bkonyi, line 137: If we're going to explicitly check for the name, we should only assign to `_mainIsolate.value` if `event.isolate?.name == 'main'` as well.

CoderDake: I understand the intention but I worry about our worst case then. If our assumption about isolate naming is wrong then anything using mainIsolate will break, since the mainIsolate would always be left as null. 
With this PR's implementation the only thing that would break is isolate selection, which is a bit more forgivable.

bkonyi: Maybe we should do this instead then:

```
if (_mainIsolate.value == null) {
  _mainIsolate.value = event.isolate;
  if (_shouldReselectMainIsolate) {
     // ...
  }
}
```

This way we can still only execute the `_shouldReselectMainIsolate` path if the isolate is actually what we consider to be the main isolate.

bkonyi (approved):

Issue #5868, opened 2023-06-01
Use DevTools to power the UI for VS Code sidebar tooling


kenzieschmoll:
See [doc](https://docs.google.com/document/d/1V9TRLQ5jt7x0JxHfJCFOUTK-JKOytM6QLyyGW6hBWAE/edit?pli=1#).

PR #6444, opened 2023-09-21
Add analytics to VS Code sidebar panel actions

kenzieschmoll:
Work towards https://github.com/flutter/devtools/issues/5868

@DanTup

### DIFF
--- a/packages/devtools_app/lib/src/shared/analytics/constants.dart
+++ b/packages/devtools_app/lib/src/shared/analytics/constants.dart
@@ -310,8 +310,7 @@ enum MemoryAreas {
 }
 
 enum VsCodeFlutterSidebar {
-  /// Analytics id to track events that come from the Flutter sidebar in VS
-  /// code.
+  /// Analytics id to track events that come from the VS Code Flutter sidebar.
   vsCodeFlutterSidebar,
 
   /// Analytics event that is sent when a device selection occurs from the list
--- a/packages/devtools_app/lib/src/shared/analytics/constants.dart
+++ b/packages/devtools_app/lib/src/shared/analytics/constants.dart
@@ -308,3 +308,20 @@ enum MemoryAreas {
 
   final String name;
 }
+
+enum VsCodeFlutterSidebar {
+  /// Analytics id to track events that come from the Flutter sidebar in VS
+  /// code.
+  vsCodeFlutterSidebar,
+
+  /// Analytics event that is sent when a device selection occurs from the list
+  /// of available devices in the sidebar.
+  changeSelectedDevice;
+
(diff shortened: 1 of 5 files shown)


packages/devtools_app/lib/src/shared/analytics/constants.dart
```diff
@@ -308,3 +308,20 @@ enum MemoryAreas {
 
   final String name;
 }
+
+enum VsCodeFlutterSidebar {
+  /// Analytics id to track events that come from the Flutter sidebar in VS
+  /// code.
```
DanTup, line 314: uber-nit: "Code" with a capital C :-)

DanTup (approved):
LGTM!

packages/devtools_app/lib/src/shared/analytics/constants.dart
polina-c: And may be to be on one line?

polina-c (approved):

kenzieschmoll:
@DanTup please take a look at the query string value we discussed, and let me know what you think.

It should be one of 
`ideFeature=command` // When a user opens DevTools from the command palette 
`ideFeature=sidebarContent` // When a user opens DevTools from the debug session drop down
`ideFeature=sidebarToolbar` // When a user opens DevTools from the action in the sidebar title / tool bar

And this should be passed from VS Code along with the existing (and unchanged) 'ide' query parameter

DanTup:
@kenzieschmoll lgtm - though there are a few additional values you might get because there are quite a few other places we can launch DevTools for (and since I was wiring up these values, it made sense to give them their own).

```
	static commandPalette = "command";
	static sidebarContent = "sidebarContent";
	static sidebarTitle = "sidebarToolbar";
	static touchbar = "touchbar"; // MacOS touchbar button
	static launchConfiguration = "launchConfiguration"; // Configured explicitly in launch configuration
	static onDebugAutomatic = "onDebugAutomatic"; // Configured to always run on debug session start
	static onDebugPrompt = "onDebugPrompt"; // Responded to prompt when running a debug session
	static languageStatus = "languageStatus"; // Launched from the language status popout
```

I suspect the extra 5 are used very little. If having them will cause you any issues, let me know.

image

kenzieschmoll:
Added the values you suggested, @DanTup PTAL

DanTup:
@kenzieschmoll lgtm, though it may be worth noting that that's a non-exhaustive list so if we add additional areas in VS Code we don't necessarily need to keep this list updated?

kenzieschmoll:
added a comment

DanTup (approved):
Still lgtm :)

Issue #6449, opened 2023-09-22
Hot restart causes platformOverride to fail


CoderDake:
To resolve this we will need to handle `SentinelException` around the below and come up with a behaviour for if the service stops existing while we are calling it.

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: [Sentinel kind: Collected, valueAsString: <collected>] from ext.flutter.platformOverride()
#0      new _OutstandingRequest (package:vm_service/src/vm_service.dart:1930:45)
#1      VmService._call (package:vm_service/src/vm_service.dart:2499:21)
#2      VmService.callServiceExtension (package:vm_service/src/vm_service.dart:2470:14)
#3      VmServiceWrapper.callServiceExtension (package:devtools_app/src/service/vm_service_wrapper.dart:163:18)
#4      ServiceExtensionManager._callServiceExtension.callExtension (package:devtools_app_shared/src/service/service_extension_manager.dart:384:25)
#5      ServiceExtensionManager._callServiceExtension (package:devtools_app_shared/src/service/service_extension_manager.dart:412:13)
<asynchronous suspension>
#6      ServiceExtensionManager._addServiceExtension (package:devtools_app_shared/src/service/service_extension_manager.dart:252:14)
<asynchronous suspension>
#7      Future.wait.<anonymous closure> (dart:async/future.dart:523:21)
<asynchronous suspension>
#8      ServiceExtensionManager._onFrameEventReceived (package:devtools_app_shared/src/service/service_extension_manager.dart:151:5)
<asynchronous suspension>
#9      ServiceExtensionManager._maybeCheckForFirstFlutterFrame (package:devtools_app_shared/src/service/service_extension_manager.dart:223:7)
<asynchronous suspension>
```

PR #6451, opened 2023-09-22
Protect service restoration against SentinelException

CoderDake:
Fixes https://github.com/flutter/devtools/issues/6449

When spamming hot restart these service refreshes commonly throw SentinelExceptions. Meaning they stopped existing while being called.

This PR handles those expections with a do nothing, since we will need to recall these for the next service to start anyways

### DIFF
--- a/packages/devtools_app_shared/lib/src/service/service_extension_manager.dart
+++ b/packages/devtools_app_shared/lib/src/service/service_extension_manager.dart
@@ -249,10 +249,14 @@ final class ServiceExtensionManager with DisposerMixin {
       // extension. This will restore extension states on the device after a hot
       // restart. [_enabledServiceExtensions] will be empty on page refresh or
       // initial start.
-      return await _callServiceExtension(
-        name,
-        _enabledServiceExtensions[name]!.value,
-      );
+      try {
+        return await _callServiceExtension(
+          name,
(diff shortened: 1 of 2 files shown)


kenzieschmoll (commented):
can you add a changelog entry to package:devtools_app_shared?

packages/devtools_app_shared/lib/src/service/service_extension_manager.dart
```diff
@@ -249,10 +249,14 @@ final class ServiceExtensionManager with DisposerMixin {
...
-        name,
-        _enabledServiceExtensions[name]!.value,
-      );
+      try {
+        return await _callServiceExtension(
+          name,
+          _enabledServiceExtensions[name]!.value,
+        );
+      } on SentinelException catch (_) {
+        // Service stopped existing while calling, so do nothing.
```
kenzieschmoll, line 258: nit r/Service/Service extension. Can you also give an example of when this would happen (i.e. on hot restart?)

CoderDake, line 258:
```suggestion
        // Service extension stopped existing while calling, so do nothing.
        // This typically happens during hot restarts.
```

kenzieschmoll (approved):

PR #6467, opened 2023-09-25
Add pub ignore comment to devtools extensions readme

kenzieschmoll:
Adding this based on a recommendation from @rrousselGit [here](https://github.com/rrousselGit/provider/pull/832#issuecomment-1732675281). Addressed for `package:provider` in https://github.com/rrousselGit/provider/commit/25a7e18b5316113344d3aa2be82485fa41d50ef1.

FYI @jonasf - do you see any issues with this recommended use of the `.pubignore` file?

### DIFF
--- a/packages/devtools_extensions/README.md
+++ b/packages/devtools_extensions/README.md
@@ -86,7 +86,7 @@ flutter create --template app --platforms web foo_package_devtools_extension
 
 In `foo_package_devtools_extension/pubspec.yaml`, add a dependency on `devtools_extensions`:
 ```yaml
-devtools_extensions: ^0.0.5
+devtools_extensions: ^0.0.7
 ```
 
 In `lib/main.dart`, place a `DevToolsExtension` widget at the root of your app:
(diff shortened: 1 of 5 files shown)


polina-c (approved):

packages/devtools_extensions/README.md
```diff
@@ -210,14 +210,25 @@ dart run devtools_extensions build_and_copy \
...
 
 As a package author, the content that you check into your git repository is completely up to you.
 If you want the contents of `extension/devtools/build/` to be git ignored, then you'll just need
 to ensure that the extension web app is always built and included in `extension/devtools/build/`
-when you publish your package.
+when you publish your package. To do so, add the following to a `.pubignore` file at the root
+of your pub package:
+
+```
+!./extension/devtools/build
```
jonasfj, line 224: @kenzieschmoll 
When you put a `.pubignore` at the root, then the root `.gitignore` won't be considered at all.

So that might be undesirable... or at least the user will have to duplicate patterns they want in both.

A better option could be to do `extensions/devtools/.pubignore` that way it's only for the `extensions/devtools/`.

It won't pollute root, and most likely there is no `.gitignore` in extensions/devtools anyways.

PR #6472, opened 2023-09-26
Fix pubignore instructions and update extension example app

kenzieschmoll:
These changes follow the guidance from @jonasfj  here: https://github.com/flutter/devtools/pull/6467#discussion_r1336701628

FYI @rrousselGit - can you confirm that the setup we are recommending in this PR works as expected for the `package:provider` DevTools extension?

No need to review the updated contents under extension/devtools/build. These were autogenerated.


CoderDake (approved):

rrousselGit:
@kenzieschmoll Yes, this works fine! Good call on moving the pubignore to the devtool folder

PR #6483, opened 2023-09-27
Fixes to release process

CoderDake:
While doing the release process I ran into some issues. This PR addresses adds some documentation and resolves those issues.


tool/lib/commands/release_helper.dart
```diff
@@ -108,19 +108,25 @@ class ReleaseHelperCommand extends Command {
 
       print("Setting the release version.");
       await processManager.runProcess(
-        CliCommand.from('devtools_tool', [
+        CliCommand.from('dart', [
+          'run',
+          'devtools_tool.dart',
```
kenzieschmoll, line 113: don't we expect `devtools_tool` to be activated in other scripts?

CoderDake: for some reason it wasn't working for me and was preventing me from running the helper.
Even though the alias was set it was saying something about a file not existing.
I thought it was weird too since the command was being run with the alias, then inside it was having a hard time finding it

kenzieschmoll: does it work if you add 
`workingDirectory: pathFromRepoRoot("tool"),` to both of those commands?

CoderDake: this is what I have my alias set to
Screenshot 2023-09-27 at 11 33 29 AM

CoderDake: To fix my issue I created a script called `devtools_tool` and put it in my ~/bin directory, which I have on my $PATH.
This helps us avoid the issue where aliases don't seem to play well with the dart command execution stuff
```
#!/bin/bash
dart run /Users/danchevalier/development/devtools/tool/bin/devtools_tool.dart "$@"

```

tool/lib/commands/update_dart_sdk_deps.dart
```diff
@@ -76,7 +76,7 @@ class UpdateDartSdkDepsCommand extends Command {
         ),
         // TODO(kenz): is there a way to automatically close the file that pops up
         // with the commit description?
-        CliCommand('git cl upload -s'),
+        CliCommand('git cl upload -s -f'),
```
kenzieschmoll, line 79: what does -f do?

CoderDake: ` -f, --force           force yes to questions (don't prompt)`
It was the way i found to skip the prompts for a message.
I'll try the -m flag if that is less intense though :)

kenzieschmoll: so with the -f flag, does the file that pops up with the commit description no longer open? if so, can you remove the todo above this line?

tool/RELEASE_INSTRUCTIONS.md
```diff
@@ -28,8 +28,22 @@ Make sure:
...
+          > Replace `<DEVTOOLS_DIR>` with the local path to your DevTools
+          > repo path.
+  - For Windows Users
+    - Ensure the `devtools_tool` executable has been globally activated:
+      - `flutter pub global activate --source path tool`
+        > [!WARNING]  
+        > Always do this activation step before running
+        > `devtools_tool`. If there have been changes to the tool since the
+        > last time you have run the activate, then it needs to be rerun to
+        > pick up those changes.
```
kenzieschmoll, line 46: @DanTup is there a better way to put `devtools_tool` on the global path for windows?

kenzieschmoll (approved):

tool/RELEASE_INSTRUCTIONS.md
DanTup: I don't think so - I think `global activate` is the right way, though you need to then ensure the Pub `bin` path is in your `PATH`. Personally I tend to not bother adding it to `PATH` though and just use `dart pub global run foo` for things like this (and `npm`) because I'm always paranoid about installing something that might have a `ls` executable or something 😄

DanTup: Although - maybe I misunderstood the question.. I'm not sure why the macOS/Windows instructions are so different here - shouldn't they both be approximately the same (ensure the Pub `bin` folder is on `PATH` and then globally activate)?

kenzieschmoll: This PR actually moves away from using pub global activate on mac and linux. Since the `devtools_tool` executable is from the tool/ package, which is activated from source, this can get out of date when working on the tool package. So to prevent having a stale `devtools_tool` exe, Dan added a script `devtools_tool` in the same directory as `devtools_tool.dart`, and then put the `devtools/tool/bin` directory on the global path so that it can be accessed via  `devtools_tool ...`. 

So the question could be, is there a way to do something similar on windows so that whenever you call `devtools_tool ...` you can be sure that you are not getting a stale `devtools_tool` exe

DanTup: Ooooh, I see!

In that case, you could do the same for Windows by having a `devtools_tool.bat` that does the same. I suspect something like (not currently at my PC so can't easily test):

```
dart run "%~dp0\devtools_tool.dart" %*
```

Then the rest would be the same (add this folder to `PATH`).

If this makes sense, I can send another PR to add/update this when back at PC tomorrow.

DanTup: I guess this isn't too different to what Flutter is doing with [dart.bat](https://github.com/flutter/flutter/blob/master/bin/dart.bat).

kenzieschmoll: @DanTup yeah if you want to push up a fix for the window side tomorrow after this lands, that sgtm. Thanks!

DanTup: Since this hasn't landed yet I pointed my PR at Dan's branch, but I can re-point it here if we want to merge this first:

https://github.com/CoderDake/devtools/pull/2

PR #6489, opened 2023-09-28
Improve error running devtools_tool from a different folder

DanTup:
This improves the error if you run the tool from the wrong folder:

Screenshot 2023-09-28 110627

This doesn't apply to the new shell scripts since they change the directory, but invoking it manually (or using `flutter pub global run`) from the wrong folder previously gave an unhelp message.

I don't know if there's much value in `getInstance()` now (every location that called it was using `!` and now calls `requireInstance()`), so we could also just update `getInstance()` to be non-nullable and throw instead. The exception is handled by the command runner and then printed.

### DIFF
--- a/tool/lib/model.dart
+++ b/tool/lib/model.dart
@@ -17,22 +17,15 @@ class DevToolsRepo {
   /// This returns the DevToolsRepo instance based on the current working
   /// directory.
   ///
-  /// This will return null if the current working directory is not
-  /// contained within a git checkout of DevTools. Use [requireInstance] for a
-  /// descriptive exception in this case.
-  static DevToolsRepo? getInstance() {
-    final repoPath = _findRepoRoot(Directory.current);
-    return repoPath == null ? null : DevToolsRepo._create(repoPath);
-  }
-
-  /// This returns the DevToolsRepo instance based on the current working
-  /// directory.
-  ///
   /// Throws if the current working directory is not contained within a git
   /// checkout of DevTools.
-  static DevToolsRepo requireInstance() {
-    return getInstance() ??
-        (throw Exception('devtools_tool must be run from inside of the DevTools repository directory'));
+  static DevToolsRepo getInstance() {
+    final repoPath = _findRepoRoot(Directory.current);
+    if (repoPath == null) {
+      throw Exception(
+          'devtools_tool must be run from inside of the DevTools repository directory');
+    }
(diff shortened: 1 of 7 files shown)


tool/lib/model.dart
```diff
@@ -17,13 +17,24 @@ class DevToolsRepo {
   /// This returns the DevToolsRepo instance based on the current working
   /// directory.
   ///
-  /// This can fail and return null if the current working directory is not
-  /// contained within a git checkout of DevTools.
+  /// This will return null if the current working directory is not
+  /// contained within a git checkout of DevTools. Use [requireInstance] for a
+  /// descriptive exception in this case.
   static DevToolsRepo? getInstance() {
```
CoderDake, line 23: will we ever have a use for this now? I wonder if it might be better to make this private with a `_` prefix, to prevent others from accidentally picking the wrong one in the future.

tool/lib/model.dart
```diff
@@ -17,13 +17,24 @@ class DevToolsRepo {
...
   }
 
+  /// This returns the DevToolsRepo instance based on the current working
+  /// directory.
+  ///
+  /// Throws if the current working directory is not contained within a git
+  /// checkout of DevTools.
+  static DevToolsRepo requireInstance() {
+    return getInstance() ??
+        (throw Exception('devtools_tool must be run from a DevTools checkout'));
```
CoderDake, line 35:
```suggestion
        (throw Exception('devtools_tool must be run from inside of the DevTools repository directory'));
```

tool/lib/model.dart
DanTup: Yeah, I noted this above - we could just remove it, or even put the `requireInstance` impl in here. I wasn't sure if there was a reason it was added as nullable though. If not, I think we just drop the `?` and make this behave like the new method below?

tool/lib/model.dart
DanTup: How do you suggest changes like this? People do it to my PRs all the time and I've never been able to find the UI to do it 🙃

tool/lib/model.dart
DanTup: I just replaced this instead. Now there's get `getInstance` (as before), but it's non-nullable and throws. As long as there's no code outside of this repo trying to use it, I think this is better.

CoderDake (approved):
I dig it :D

Issue #6149, opened 2023-08-03
Some characters not displayed correctly on web


bkonyi:
In a [sample program](https://github.com/dart-lang/sdk/blob/main/runtime/observatory/tests/ui/inspector.dart) meant to create instances of most types that can be displayed in a debugger, `stringTrebleClef` contains the character "𝄞", which is a [UTF-16 surrogate pair](https://learn.microsoft.com/en-us/windows/win32/intl/surrogates-and-supplementary-characters). On desktop, this character is rendered correctly, but on web a black character is rendered instead.
 
On desktop:

![image](https://github.com/flutter/devtools/assets/24210656/df16e742-cf1c-4927-b193-0d6790074615)

On web:

![image](https://github.com/flutter/devtools/assets/24210656/0ae73e7f-7e8e-4252-9d99-38d06a81acfe)

jacob314:
How do other similar debugging tools handle this? Including more fonts would be nice but I don't think we really want to include every possible font.
Possible solutions:
add an extra api registered on the VMService by flutter tools that lets us dynamically load exactly the font we need.
just include an extra 50mb of fonts with devtools covering more unicode characters and call that good enough.

PR #6493, opened 2023-09-29
Fixed missing musical symbols

CoderDake:
Adding a fallback font family for musical symbols so that they show correctly.

Fixes https://github.com/flutter/devtools/issues/6149

More discussion around how many more fallback symbols we would like:
https://discord.com/channels/608014603317936148/1106667330093723668/1157406137545662535

Screenshot 2023-09-29 at 3 32 13 PM


packages/devtools_app/pubspec.yaml
```diff
@@ -163,6 +163,9 @@ flutter:
     - family: Octicons
       fonts:
         - asset: fonts/Octicons.ttf
+    - family: Noto
+      fonts:
+        - asset: fonts/Noto/NotoMusic-Regular.ttf
```
kenzieschmoll, line 168: before we add this we should make sure this already exists in g3 and will not block our roll.

CoderDake:
Closing in lieu of https://github.com/flutter/engine/pull/46720 which solves the symbol fallback in a more central way

Issue #5837, opened 2023-05-22
Tech debt: move features of VmServiceWrapper from DevTools to VmService.


polina-c:
We need to update [VmServiceWrapper](https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/service/service.dart) every time VmService is updated.

We will save this effort by moving the features to VmService.

PR #6508, opened 2023-10-09
Bump to vm_service 13.0.0 and reduce `VmServiceWrapper`.

kenzieschmoll:
With this PR, we would no longer be implementing the `VmService` class in DevTools, but rather extending it. This means that we should not be broken by every breaking change to the `VmService` like we are now. 

This PR depends on both of these changes landing in the SDK and being rolled through the chain package dependencies (`test`, `coverage`, etc.)
- https://dart-review.googlesource.com/c/sdk/+/329800
- https://dart-review.googlesource.com/c/sdk/+/329820

@bkonyi 

Fixes https://github.com/flutter/devtools/issues/5837


bkonyi (approved):

packages/devtools_app/lib/src/service/vm_service_wrapper.dart
```diff
@@ -807,117 +233,59 @@ class VmServiceWrapper implements VmService {
...
   }
-  // Mark: end overrides for the [DdsExtension].
 
-  // Mark: Overrides for [DartIOExtension]. It would help with logical grouping
-  // to make these extension methods, but that makes testing more difficult due
-  // to mocking limitations for extension methods.
-  Future<Version> getDartIOVersion(String isolateId) =>
-      trackFuture('_getDartIOVersion', _vmService.getDartIOVersion(isolateId));
+  // Begin Dart IO extension method wrappers. We wrap these methods so that we
+  // can override them in tests.
```
kenzieschmoll, line 238: I don't love this but I'm not sure if there is another way to do this given that we need to fake these methods in tests.

kenzieschmoll:
I have removed the Provider screen tests as part of this PR. These tests were failing ([example test run](https://github.com/flutter/devtools/actions/runs/6738710930/job/18318900778?pr=6508)), and since this screen is now maintained as part of the provider repository, it is not worth the effort to fix this test in our repo. Many of these tests were already disabled as is. CC @rrousselGit

PR #6535, opened 2023-10-16
Enhance devtools_extension documentation and example app.

kenzieschmoll:
Also fixes https://github.com/flutter/devtools/issues/6539


packages/devtools_extensions/example/app_that_uses_foo/.vscode/launch.json
polina-c, line 1: Is it intentional? Do we want to move these new configurations to our standard location: packages/.vscode/launch.json

packages/devtools_extensions/extension_config_spec.md
```diff
@@ -0,0 +1,13 @@
+The `config.yaml` file for a DevTools extension must follow the format below.
+
```
polina-c, line 2: How about naming it "extension_config.yaml"? 
There can be other things to configure, right?

packages/devtools_extensions/extension_config_spec.md
```diff
@@ -0,0 +1,13 @@
...
+Required fields:
```
polina-c, line 3: How about making it bulleted list?

packages/devtools_extensions/extension_config_spec.md
kenzieschmoll: The name of this file is set. This is plumbed through the DevTools extension framework and `package:extension_discovery`. This markdown file is just to document the expected fields for the file.

packages/devtools_extensions/extension_config_spec.md
```diff
@@ -0,0 +1,13 @@
...
+`name` : the package name that this DevTools extension belongs to. The value of this field
+will be used in the extension page title bar.
+`issue_tracker`: the url for the extension's issue tracker. When a user clicks the “Report an 
+issue” link in the DevTools UI, they will be directed to this url.
+`version`: the version of the DevTools extension. This version number should evolve over time 
+as the extension is developed. The value of this field will be used in the extension page 
+title bar.
+`material_icon_code_point`: corresponds to the codepoint value of an icon from
```
polina-c, line 11: I think we need example here, because format is not clear.

packages/devtools_extensions/example/app_that_uses_foo/.vscode/launch.json
kenzieschmoll: This change was intentional. DevTools extension authors may open the example app directly to run it. I found myself doing this today.

packages/devtools_extensions/extension_config_spec.md
kenzieschmoll: the file lives at:
```
extension/
  devtools/
    config.yaml
```
so the fact that it is a configuration for a devtools extension is implicit in the path

polina-c (approved):

bkonyi (approved):

Issue #6541, opened 2023-10-17
`eval_and_browse_test.dart` integration test is broken or flakes


kenzieschmoll:
This test has been disabled and should be fixed an re-enabled.

Example broken test run: https://github.com/flutter/devtools/actions/runs/6540429539/job/17760352048?pr=6537

PR #6542, opened 2023-10-17
Disable broken integration test `eval_and_browse_test.dart`

kenzieschmoll:
Tracking issue for broken test https://github.com/flutter/devtools/issues/6541


polina-c:
How about renaming it instead: https://github.com/flutter/devtools/pull/6543

With renaming it is easier to see that it is disabled. And, less code to maintain.

PR #6545, opened 2023-10-17
Fix for broken eval_and_browse

polina-c:


packages/devtools_app/integration_test/run_tests.dart
```diff
@@ -22,8 +22,7 @@ const _offlineIndicator = 'integration_test/test/offline';
 /// This list should be empty most of the time, but may contain a broken test
 /// while a fix being worked on.
 const _skipTests = <String>[
-  // TODO(https://github.com/flutter/devtools/issues/6541): re-enable once fixed.
-  'eval_and_browse_test.dart',
+  'my_example_test.dart',
```
kenzieschmoll, line 25: instead of putting an actual value here, can you just add an example in the comment above?

packages/devtools_app/integration_test/test/live_connection/eval_and_browse_test.dart
```diff
@@ -199,13 +200,15 @@ class _EvalAndBrowseTester {
   ///
   /// If [next] is provided, will repeat the tap untill [next] returns results.
   /// If [next] is not null returns [next].
-  Future<Widget?> tapAndPump(
+  Future<Finder?> tapAndPump(
```
kenzieschmoll, line 203: can you update the dart doc for this helper? "If [next] is provided" and "If [next] is not null" describe the same case.

packages/devtools_app/integration_test/run_tests.dart
polina-c: Current implementation is (1) magical, as it will take debugging to understand why the test is skipped and (2) does not account for the same file name in different folders.

So, I suggest to delete it for now and re-implement in differently when we need it.

packages/devtools_app/integration_test/run_tests.dart
```diff
@@ -17,14 +17,6 @@ import 'test_infra/run/run_test.dart';
...
 
-/// The set of test that should be skipped for all devices.
-///
-/// This list should be empty most of the time, but may contain a broken test
-/// while a fix being worked on.
-const _skipTests = <String>[
-  'my_example_test.dart',
-];
-
 void main(List<String> args) async {
```
kenzieschmoll, line 20: why are we removing this skip mechanism? we should leave this here so that we can skip integration tests in the future

packages/devtools_app/integration_test/run_tests.dart
kenzieschmoll: This would catch the same test name in different folders because it checks `testTarget.endsWith(t)`, so the beginning part of the path does not matter. I don't think this is any more magical than setting `skip: true` on a flutter test - that still would require looking at the code to understand why it is skipped. Perhaps we could log "Skipping $testTarget" instead of just returning when `shouldSkip` is true

polina-c: This is what I mean: if there are two tests with the same name, both will be skipped and there is no way to skip just one.
Returned it as it was.

packages/devtools_app/integration_test/run_tests.dart
polina-c: reverted

kenzieschmoll (approved):

Issue #6552, opened 2023-10-18
Create performance benchmarks for DevTools


kenzieschmoll:
See package https://pub.dev/packages/web_benchmarks. We will need somewhere to submit our benchmark numbers to. Flutter uses Skia Perf so maybe we can do this for DevTools too.

CC @mkustermann @kevmoo @sigmundch adding these benchmarks will help with the dart2wasm investigation.

kenzieschmoll:
@kevmoo @mkustermann @srujzs @sigmundch 

What types of performance can we expect dart2wasm to improve? It would be good to get a small set of benchmarks put together that we expect to show improvements with dart2wasm. For example, do we expect IPL to improve? overall UI perf (scrolling, page transitions, etc.)? Business logic? RPC messaging to / from the VM service? Json parsing? etc.

We have some known expensive operations like collecting a heap snapshot over the VM and parsing it, but I'm not sure where we can expect perf improvements from wasm

kevmoo:
Great question, @kenzieschmoll ! Ideally we'd see improvements across the board. Everything you list *should* get better–or, hopefully–not worse!

kevmoo:
Starting with IPL is great. Any work we could do to track janking frames would be great, too.

kenzieschmoll:
Do we have any IPL benchmarks for other Dart / Flutter products we could use as an example?

mkustermann:
As @kevmoo says, we're interested in benchmarks that are a proxy for user experience of DevTools. The set of benchmarks that would (if dart2wasm is on-par of better than dart2js) convince DevTools team to enable WASM for end users. If dart2wasm is meaningfully worse anywhere, we'd like to know about it and work on it.

> Do we have any IPL benchmarks for other Dart / Flutter products we could use as an example?

@yjbanov What benchmarking infrastructure does flutter use for measuring IPL / ...?

kevmoo:
`web_benchmarks` is now published with Wasm support. Are you unblocked @kenzieschmoll ?

PR #6562, opened 2023-10-19
Add a benchmark test for the DevTools web bundle size

kenzieschmoll:
This is modeled after the web bundle size test in flutter gallery: https://github.com/flutter/gallery/blob/master/test_benchmarks/web_bundle_size_test.dart

At this time the bundle sizes for DevTools are:
```
main.dart.js: 4556 kb 
main.dart.js.gz: 1352 kb
```

Work towards https://github.com/flutter/devtools/issues/6552

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -52,6 +52,7 @@ jobs:
           - build_dart2js
           - test_ddc
           - test_dart2js
+          - benchmarks_dart2js
     steps:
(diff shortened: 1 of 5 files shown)


CoderDake (approved):

kenzieschmoll:
We can land this for now, and if we find that we do not get much value out of this, or if we choose to track this on golem instead, we can remove this test.

PR #6567, opened 2023-10-20
Show Performance and CPU Profiler screens without app connection

kenzieschmoll:
Without the app connection, these screens are here to provide users with a way to load offline data. This PR also removes the ImportFileAction from the set of scaffold actions, and replaces the existing "Export" buttons with a new `OpenSaveButtonGroup`, since we will keep these actions local to the screens that support Open / Save.

Screenshot 2023-10-20 at 12 47 44 PM
Screenshot 2023-10-20 at 12 47 53 PM

Error message when importing a DevTools file for the incorrect screen:
Screenshot 2023-10-20 at 12 50 51 PM

New Open / Save actions:
Screenshot 2023-10-20 at 2 26 13 PM
Screenshot 2023-10-20 at 2 26 49 PM

Fixes https://github.com/flutter/devtools/issues/4645


elliette (approved):

packages/devtools_app/lib/src/screens/performance/performance_screen.dart
```diff
@@ -148,6 +158,28 @@ class WebPerformanceScreenBody extends StatelessWidget {
   }
 }
 
+class DisconnectedPerformanceScreenBody extends StatelessWidget {
+  const DisconnectedPerformanceScreenBody({super.key});
+
+  static const importInstructions =
+      'Open a performance data file that was previously saved from DevTools';
```
CoderDake, line 165:
```suggestion
      'Open a performance data file that was previously saved from DevTools.';
```

CoderDake (approved):

PR #6576, opened 2023-10-23
Split deep link exports into ones that require IO and ones that do not

kenzieschmoll:
This is necessary for rolling into g3.


chunhtai (approved):
LGTM

PR #6579, opened 2023-10-24
Migrate generate_code.sh to devtools_tool for Windows compatibility

DanTup:
On Windows I've had to run `build_runner build` manually in each folder so this adds a devtools_tool command in Dart and calls it from the old bash script.


tool/generate_code.sh
```diff
@@ -10,37 +10,9 @@ RELATIVE_PATH_TO_SCRIPT="${BASH_SOURCE[0]}"
 # The directory that this script is located in.
 TOOL_DIR=`dirname "${RELATIVE_PATH_TO_SCRIPT}"`
```
kenzieschmoll, line 11: lets just delete this file altogether and call `devtools_tool generate-code` everywhere we were previously calling generate_code.sh

Screenshot 2023-10-24 at 9 25 51 AM

DanTup: Done! I wasn't sure if others might have scripts/keybinds to call it, but this probably makes more sense. I'll ping a note in Discord about it too.

tool/bots.sh
```diff
@@ -73,7 +73,7 @@ export PATH="$PATH":"$DEVTOOLS_DIR/tool/bin"
 devtools_tool pub-get --only-main
 
 # Generate code.
-bash tool/generate_code.sh
+bash tool/bin/devtools_tool generate-code
```
kenzieschmoll, line 76: this can just be `devtools_tool generate-code`
see the above command

kenzieschmoll (approved):
one comment then lgtm

tool/bots.sh
DanTup: Oops, should've noticed that. Fixed, thanks!

Issue #6580, opened 2023-10-24
Sidebar does not reflect light theme in VS Code


DanTup:
If I set VS Code to the light theme, it embeds the sidebar like this:

http://127.0.0.1:9100/vsCodeFlutterPanel?theme=light&backgroundColor=%23f3f3f3&foregroundColor=%236f6f6f&fontSize=14

However it renders with a black background (darker than even the grey when I have it in the Dark+ theme), so something isn't being applied correctly.

DanTup:
Apparently I may have noticed this before - this might be a dupe of https://github.com/flutter/devtools/issues/5911

Issue #5911, opened 2023-06-13
IDE themes not working in VS Code


DanTup:
If I switch VS Code to a light theme and open the embedded inspector, it's still dark:

![Screenshot 2023-06-13 at 16 16 22](https://github.com/flutter/devtools/assets/1078012/2bc8ed48-0e8c-4387-8dd2-9a183106416f)

The URL being loaded looks like:

> ... /inspector?uri=ws%3A%2F%2F127.0.0.1%3A62171%2F1Bf7h481aGI%3D%2Fws&embed=true&hide=debugger&ide=VSCode&cacheBust=dart-3.1.0-203.0.dev-flutter-3.12.0-3.0.pre.86-custom-1686669373020&theme=light&backgroundColor=%23ffffff&foregroundColor=%23000000&fontSize=12

PR #6581, opened 2023-10-24
Use theme preference from IDE when embedded instead of preference

DanTup:
Currently the theme always comes from the preference and ignores what the IDE passes. Since there's no way to un-set the preference (or change it in the UI when embedded) this may prevent it matching the IDE.

This change always uses the IDE preference when embedded. It requires a change in VS Code to pass embed=true when embedding the sidebar since this was missing originally.

Fixes https://github.com/flutter/devtools/issues/6580
Fixes https://github.com/flutter/devtools/issues/5911

Requires https://github.com/Dart-Code/Dart-Code/issues/4812 in a new Dart-Code release to apply.

### DIFF
--- a/packages/devtools_app/release_notes/NEXT_RELEASE_NOTES.md
+++ b/packages/devtools_app/release_notes/NEXT_RELEASE_NOTES.md
@@ -65,6 +65,10 @@ TODO: Remove this section if there are not any general updates.
 
 TODO: Remove this section if there are not any general updates.
 
+## VS Code Sidebar updates
+
+* When using VS Code with a light theme, the embedded sidebar provided by DevTools will now also show in the light theme
+
 ## Full commit history
 
 To find a complete list of changes in this release, check out the
--- a/packages/devtools_app/release_notes/NEXT_RELEASE_NOTES.md
+++ b/packages/devtools_app/release_notes/NEXT_RELEASE_NOTES.md
@@ -67,7 +67,8 @@ TODO: Remove this section if there are not any general updates.
 
 ## VS Code Sidebar updates
 
-* When using VS Code with a light theme, the embedded sidebar provided by DevTools will now also show in the light theme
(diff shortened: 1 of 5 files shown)


packages/devtools_app/release_notes/NEXT_RELEASE_NOTES.md
```diff
@@ -65,6 +65,10 @@ TODO: Remove this section if there are not any general updates.
 
 TODO: Remove this section if there are not any general updates.
 
+## VS Code Sidebar updates
+
+* When using VS Code with a light theme, the embedded sidebar provided by DevTools will now also show in the light theme
+
```
kenzieschmoll, line 71: please link this PR here in the release notes. thanks!

packages/devtools_app_shared/lib/src/ui/theme/ide_theme.dart
```diff
@@ -16,12 +16,14 @@ final class IdeTheme {
     this.foregroundColor,
```
kenzieschmoll, line 16: are we even using the backgroundColor and foregroundColor in this class?

packages/devtools_app/release_notes/NEXT_RELEASE_NOTES.md
DanTup: Oops, done!

packages/devtools_app_shared/lib/src/ui/theme/ide_theme.dart
DanTup: Yep, it's used around here:

https://github.com/flutter/devtools/blob/66a49c3500ef867662638432c4afebbdd128ffda/packages/devtools_app_shared/lib/src/ui/theme/theme.dart#L40

It only gets set from the querystring if it's dart/light enough (since in VS Code we read it from the theme that we have no control of), and then in the code above we override the colour to that (so that it more closely matches your IDE theme).

kenzieschmoll (approved):

PR #6588, opened 2023-10-25
Fix some "devtools_tool sync" issues on Windows

DanTup:
Some minor issues with needing `.bat` on Windows and the `flutter pub` processes not working/completing unless stdout is drained.

### DIFF
--- a/packages/devtools_extensions/bin/_build_and_copy.dart
+++ b/packages/devtools_extensions/bin/_build_and_copy.dart
@@ -84,6 +84,9 @@ class BuildExtensionCommand extends Command {
 
     _log('Copying built output to the extension destination...');
     await _copyBuildToDestination(source: source, dest: destination);
+
+    // Closes stdin for the entire program.
+    await sharedStdIn.terminate();
   }
 
   Future<void> _copyBuildToDestination({
--- a/packages/devtools_extensions/bin/_build_and_copy.dart
+++ b/packages/devtools_extensions/bin/_build_and_copy.dart
@@ -84,9 +84,6 @@ class BuildExtensionCommand extends Command {
 
     _log('Copying built output to the extension destination...');
     await _copyBuildToDestination(source: source, dest: destination);
-
-    // Closes stdin for the entire program.
-    await sharedStdIn.terminate();
   }
 
   Future<void> _copyBuildToDestination({
--- a/tool/lib/commands/analyze.dart
+++ b/tool/lib/commands/analyze.dart
@@ -46,6 +46,9 @@ class AnalyzeCommand extends Command {
         CliCommand.from(
           sdk.dartToolPath,
           ['analyze', '--fatal-infos'],
+          // Run all so we can see the full set of results instead of stopping
+          // on the first error.
+          throwOnException: false,
         ),
         workingDirectory: p.packagePath,
       );
--- a/tool/lib/commands/analyze.dart
+++ b/tool/lib/commands/analyze.dart
@@ -3,10 +3,10 @@
 // found in the LICENSE file.
 
 import 'dart:async';
-import 'dart:io';
 
 import 'package:args/command_runner.dart';
 import 'package:cli_util/cli_logging.dart';
+import 'package:io/io.dart';
 
 import '../model.dart';
 import '../utils.dart';
@@ -28,6 +28,7 @@ class AnalyzeCommand extends Command {
 
     final log = Logger.standard();
     final repo = DevToolsRepo.getInstance();
+    final processManager = ProcessManager();
     final packages = repo.getPackages();
 
     log.stdout('Running flutter analyze...');
@@ -41,15 +42,17 @@ class AnalyzeCommand extends Command {
 
       final progress = log.progress('  ${p.relativePath}');
 
-      final process = await Process.start(
-        sdk.dartToolPath,
-        ['analyze', '--fatal-infos'],
+      final process = await processManager.runProcess(
+        CliCommand.from(
+          sdk.dartToolPath,
+          ['analyze', '--fatal-infos'],
+        ),
         workingDirectory: p.packagePath,
       );
-      final Stream<List<int>> stdout = process.stdout;
-      final Stream<List<int>> stderr = process.stderr;
 
-      final int exitCode = await process.exitCode;
+      final stdout = process.stdout;
+      final stderr = process.stderr;
+      final int exitCode = process.exitCode;
 
       if (exitCode == 0) {
         progress.finish(showTiming: true);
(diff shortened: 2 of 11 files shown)


tool/lib/commands/pub_get.dart
```diff
@@ -67,6 +68,10 @@ class PubGetCommand extends Command {
         ['pub', command],
         workingDirectory: p.packagePath,
       );
+      // For me (dantup), these processes never complete unless we are
+      // consuming stdout too, so run it through log.trace.
+      process.stdout.transform(utf8.decoder).listen(log.trace);
```
kenzieschmoll, line 73: should we just use the ProcessManager instead like some of the other commands? I think the ProcessManager class automatically manages forwarding the stdout and stderr streams

kenzieschmoll (commented):
this makes me wonder, should we have our windows contributors set an alias `devtools_tool = devtools_tool.bat` so that all our commands in the Contributing guide work as written and we don't always have to conditionally change the instructions for windows users?

tool/lib/commands/pub_get.dart
DanTup: ProcessManager didn't expose stderr, so I had to refactor it a little. I moved the analyzer command over to use it too.

I bumped the language version to 3.0 in tool/ so I could use a record for the return type (the SDK `ProcessResult` has `dynamic` fields!). If this is a problem, lmk and I can switch it to a class.

tool/lib/commands/analyze.dart
```diff
@@ -41,27 +42,26 @@ class AnalyzeCommand extends Command {
...
         failureCount++;
 
         // Display stderr when there's an error.
-        final List<List<int>> out = await stdout.toList();
-        final stdOutput = convertProcessOutputToString(out, '    ');
-
-        final List<List<int>> err = await stderr.toList();
-        final errorOutput = convertProcessOutputToString(err, '    ');
+        final stdOutput = convertProcessOutputToString(stdout, '    ');
+        final errorOutput = convertProcessOutputToString(stderr, '    ');
```
kenzieschmoll, line 64: do we even need these? I thought ProcessManager automatically shows stdout and stderr output

tool/lib/commands/pub_get.dart
```diff
@@ -62,23 +63,23 @@ class PubGetCommand extends Command {
...
 
       if (exitCode == 0) {
         progress.finish(showTiming: true);
       } else {
         failureCount++;
 
         // Display stderr when pub get goes wrong.
-        final List<List<int>> err = await stderr.toList();
-        final errorOutput = convertProcessOutputToString(err, '    ');
+        final errorOutput = convertProcessOutputToString(stderr, '    ');
```
kenzieschmoll, line 82: same question here

tool/lib/utils.dart
```diff
@@ -109,26 +109,24 @@ class CliCommand {
...
     );
-    unawaited(
-      process.stdout
-          .transform(
-            utf8.decoder,
-          )
-          .forEach((x) => stdout = '$stdout$x'),
-    );
+    process.stdout.transform(utf8.decoder).listen(processStdout.write);
+    process.stderr.transform(utf8.decoder).listen(processStderr.write);
```
kenzieschmoll, line 129: maybe I'm mistaken but I don't think we need these. `ProcessManager.spawn` returns `_ForwardingSpawn` which should handle stdout and stderr

tool/lib/commands/analyze.dart
DanTup: Not as far as I can see. It used to run the process and return stdout, but if exitCode is non-zero through. It doesn't seem like it ever prints the output though (I have made some changes to it to return both stdout and stderr, but looking at the original code there was no printing their either).

Some of the commands tun through it (like fetching branches etc.) probably don't want to all be printed anyway?

DanTup: I've done this, though I think the original behaviour for some of these commands (like this) were to _not_ show stdout (and only show stderr if there was a failure) so the output might be more verbose now?

That might not be bad (it might even be good) on the bots, though might be less useful locally. I don't know if we can trivially suppress the stdout using `_ForwardingSpawn` though?

kenzieschmoll: For tools like this, I think more verbosity is better so we can quickly see where things go wrong 👍

packages/devtools_extensions/bin/_build_and_copy.dart
```diff
@@ -84,9 +84,6 @@ class BuildExtensionCommand extends Command {
 
     _log('Copying built output to the extension destination...');
     await _copyBuildToDestination(source: source, dest: destination);
-
-    // Closes stdin for the entire program.
-    await sharedStdIn.terminate();
```
kenzieschmoll, line 89: this should not be removed since this does not use the `DevToolsCommandRunner` from the `tool/` directory

tool/lib/commands/pub_get.dart
```diff
@@ -62,26 +63,19 @@ class PubGetCommand extends Command {
...
         workingDirectory: p.packagePath,
       );
-      final stderr = process.stderr;
-
-      final exitCode = await process.exitCode;
 
+      final exitCode = process.exitCode;
       if (exitCode == 0) {
         progress.finish(showTiming: true);
       } else {
```
kenzieschmoll, line 76: I don't think we will ever hit this else case because the above command will have `throwOnException = true`

tool/lib/commands/analyze.dart
```diff
@@ -41,32 +42,20 @@ class AnalyzeCommand extends Command {
...
       );
-      final Stream<List<int>> stdout = process.stdout;
-      final Stream<List<int>> stderr = process.stderr;
 
-      final int exitCode = await process.exitCode;
-
-      if (exitCode == 0) {
+      if (process.exitCode == 0) {
         progress.finish(showTiming: true);
       } else {
```
kenzieschmoll, line 55: same comment here. We will never hit this else case because an exception would have been thrown by now

kenzieschmoll: since the else case was just there to track the error, can we just remove the if / else entirely and then if the process has a non-zero exit code, an exception will be thrown from `processManager.runProcess`?

tool/lib/commands/pub_get.dart
DanTup: Ah yes - should I remove this code or pass `throwOnException : false`? The current behaviour is to just print and return the failure count from `run` (which presumably becomes the exit code), whereas with `throwOnException: true` it'll presumably always get the same exit code (and not get the same printing we had).

packages/devtools_extensions/bin/_build_and_copy.dart
DanTup: Good catch - fixed

tool/lib/commands/pub_get.dart
kenzieschmoll: How about we change `DevToolsProcessManagerExtension.runProcess` to include the std err output on exceptions? then we just get this stderr handling for free on all the process runs

DanTup: I realise there's another difference if we do that - it'll stop on the first failure. The current behaviour is to continue to run the others so you can see all the results up front (this would avoid getting a single error, fixing it, then finding the next package also has a failure).

So I think `throwOnException:false` might make most sense for these loops?

kenzieschmoll (approved):

PR #6589, opened 2023-10-25
Add some trailing commas to fix build

DanTup:
One of the bots is red because of missing trailing commas. Seems like the autosubmit bot landed https://github.com/flutter/devtools/pull/6584 without all the bot checks. Possibly it checks at the same time as the last commit was pushed after github had cleared the previous checks and was starting the new ones?

image


kenzieschmoll (approved):

DanTup:
Landing with the failure because it's unrelated to this change, and this change fixes at least some of the bot redness.

PR #6591, opened 2023-10-25
Remove integration with leak_tracker.

polina-c:
We will return it in another form when leak_tracker is ready.


packages/devtools_app/lib/src/screens/memory/framework/connected/memory_controller.dart
```diff
@@ -7,7 +7,7 @@ import 'dart:async';
 import 'package:devtools_app_shared/utils.dart';
 import 'package:devtools_shared/devtools_shared.dart';
 import 'package:flutter/foundation.dart';
-import 'package:leak_tracker/devtools_integration.dart';
+
```
bkonyi, line 10: Nit: this space can be removed.

bkonyi (approved):
LGTM. However, out of curiosity, why are we removing this?

polina-c:
> LGTM. However, out of curiosity, why are we removing this?

The integration is eroded significantly and stopped working, as we switched focus from detecting leaks for running app to detecting leaks in regression tests.

It will be easier to recreate the integration in new form, that to maintain it.

Issue #5775, opened 2023-05-11
 copy is not successful NotAllowedError: The Clipboard API has been blocked


CoderDake:
While testing log copying behaviour in VSCode, I noticed that the copy button is no longer working, when using it in the context of VSCode.

Alternatively, I have run Dart Devtools from the command line using `flutter run -d chrome`. When running devtools like this, the copy functionality seems to work just fine.

The following is the log message that is output in the vscode developer console when the copy fails.

```
copy is not successful NotAllowedError: The Clipboard API has been blocked because of a permissions policy applied to the current document. See https://goo.gl/EuHzyv for more details.
main.dart.js:44705 [zoneGuarded]: PlatformException(copy_fail, Clipboard.setData failed, null, null)
cg3 @ main.dart.js:44705
$1 @ main.dart.js:87529
bwd @ main.dart.js:4702
(anonymous) @ main.dart.js:156473
qh @ main.dart.js:55529
qi @ main.dart.js:55506
pp @ main.dart.js:55235
k0 @ main.dart.js:55211
pp @ main.dart.js:54607
I @ main.dart.js:54551
aaH @ main.dart.js:148455
c2 @ main.dart.js:148450
bQC @ main.dart.js:8253
$2 @ main.dart.js:62342
Ax @ main.dart.js:55635
$5 @ main.dart.js:55664
ut @ main.dart.js:55489
ks @ main.dart.js:55523
Lm @ main.dart.js:4548
fZ @ main.dart.js:54840
iY @ main.dart.js:54418
k @ main.dart.js:4402
(anonymous) @ main.dart.js:20137
(anonymous) @ main.dart.js:4411
$2 @ main.dart.js:54430
$2 @ main.dart.js:54427
bwc @ main.dart.js:4705
(anonymous) @ main.dart.js:156474
Ax @ main.dart.js:55531
YT @ main.dart.js:54770
$0 @ main.dart.js:54927
Lm @ main.dart.js:4570
fZ @ main.dart.js:54840
iY @ main.dart.js:54418
k @ main.dart.js:4402
(anonymous) @ main.dart.js:33074
(anonymous) @ main.dart.js:4411
$2 @ main.dart.js:54430
$2 @ main.dart.js:54427
bwc @ main.dart.js:4705
(anonymous) @ main.dart.js:156474
Ax @ main.dart.js:55531
YT @ main.dart.js:54770
$0 @ main.dart.js:54927
Lm @ main.dart.js:4570
[...]

PR #6598, opened 2023-10-26
New VSCode fallback behaviour for copyToClipboard

CoderDake:
Adds a fallback behaviour for copying which will allow copy buttons to actually work in VSCode.

NOTE: success toasts will no longer show in VSCode though.

I've also added some logging just in case a user is looking at their console when a copy fails. Hopefully this provides enough information that the unhide-able error doesn't look suspicious.
Screenshot 2023-10-26 at 10 53 39 AM

Related to https://github.com/Dart-Code/Dart-Code/pull/4814
Fixes https://github.com/flutter/devtools/issues/5775


packages/devtools_app/lib/src/shared/config_specific/copy_to_clipboard/_copy_to_clipboard_web.dart
```diff
@@ -0,0 +1,19 @@
...
+// in the LICENSE file.
+
+// ignore: avoid_web_libraries_in_flutter, as designed
+import 'dart:html';
+
+void copyToClipboardVSCode(String data) {
+  // This post message is only relevant in VSCode. If the parent frame is
+  // listening for this command, then it will attempt to copy the contents
+  // to the clipboard in the context of the parent frame.
+  window.parent?.postMessage(
```
DanTup, line 12: You can probably use the existing `postMessage` in packages\devtools_app\lib\src\shared\config_specific\post_message\post_message.dart to avoid needing multiple stub files for this API? (eg. have a single copy function that just calls the shared postMessage)?

CoderDake: interesting, yeah it doesn't look like it exposed 🤔 
Would you be alright with me just leaving the multi stub files for the copy behaviour?

CoderDake: Issue created :)
https://github.com/flutter/devtools/issues/6613

packages/devtools_app/lib/src/shared/config_specific/copy_to_clipboard/_copy_to_clipboard_stub.dart
```diff
@@ -0,0 +1,7 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: remove the stub file and just use the desktop file in the conditional import

packages/devtools_app/lib/src/shared/config_specific/copy_to_clipboard/_copy_to_clipboard_web.dart
```diff
@@ -0,0 +1,19 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be found
+// in the LICENSE file.
+
+// ignore: avoid_web_libraries_in_flutter, as designed
+import 'dart:html';
```
kenzieschmoll, line 6: please use package:web here. See https://github.com/flutter/devtools/pull/6607

packages/devtools_app/lib/src/shared/config_specific/copy_to_clipboard/copy_to_clipboard.dart
```diff
@@ -0,0 +1,47 @@
...
+  try {
+    await Clipboard.setData(
+      ClipboardData(
+        text: data,
+      ),
+    );
+
+    if (successMessage != null) notificationService.push(successMessage);
+  } catch (e) {
+    _log.warning(
```
kenzieschmoll, line 34: check if we are embedded and in VS code before doing any of this

kenzieschmoll: You can check if we are embedded by using `ideTheme.embed` and you can check which IDE is hosting by checking here: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/shared/analytics/_analytics_web.dart/#L733.

Although, I think we may want to modify the IdeTheme class to include the IDE name so we can look it up from there instead of having to look at our analytics file.

kenzieschmoll (approved):

Issue #6606, opened 2023-10-26
Migrate to pkg:web


kevmoo:
Required to support compilation to wasm

- [x] Migrate pkg:devtools_app_shared - WIP - https://github.com/flutter/devtools/pull/6626
- [x] Migrate pkg:devtools_app

Googlers: [go/wasm-ready-pkg-migrate](https://goto.google.com/wasm-ready-pkg-migrate)

kenzieschmoll:
migrations for `devtools_app` that are not straightforward, or just need a little more attention to figure out:

- [ ] _drag_and_drop_web.dart: uses `MouseEvent.offset`
- [ ] _framework_initialize_web.dart: `for (var element in html.document.body!.querySelectorAll('.legacy-dart'))`
- [ ] _export_web.dart: uses `Url.createObjectUrl`
- [ ] allowed_error_html.dart: uses `window.console`
- [ ] _logger_html.dart: uses `window.console`
- [ ] _server_web.dart: there is an outstanding todo to use package:http instead of dart:html here for http requests
- [ ] post_message_web.dart: uses `html.window.onMessage.map(...)`

kevmoo:
Console:

https://pub.dev/documentation/web/0.3.0/helpers/console.html
https://pub.dev/documentation/web/0.3.0/helpers/$ConsoleExtension.html

srujzs:
- `MouseEvent.offset`: The `dart:html` version returns a `Point` with `offsetX` and `offsetY`. There's some checking to make sure `offsetX` is supported, but that's stale since modern browsers should support it: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/offsetX. If the `Point` class is heavily used in devtools, then maybe we should add it as a helper class in `package:web`. If not, using a record with `offsetX` and `offsetY` should probably suffice.
- `createObjectURL`: The `dart:html` version checks for a `webkitURL`. This is no longer needed as it seems all browsers nowadays support `URL`: https://developer.mozilla.org/en-US/docs/Web/API/URL so it should be safe to url the `package:web` version => `URL.createObjectURL`.
- `html.window.onMessage`: I think this needs `_ElementEventStreamImpl` to be added to the helpers in `package:web`. A workaround is using event listeners, but `Stream`s are nicer.

kenzieschmoll:
@srujzs 
re: MouseEvent, what about the `dataTransfer` property?
`event.dataTransfer.dropEffect = 'move';`

[...]

PR #6607, opened 2023-10-26
devtools_app: migrate _analytics_web.dart to pkg:web

kevmoo:
Towards https://github.com/flutter/devtools/issues/6606

### DIFF
--- a/packages/devtools_app/lib/src/shared/analytics/_analytics_web.dart
+++ b/packages/devtools_app/lib/src/shared/analytics/_analytics_web.dart
@@ -8,12 +8,11 @@ library gtags;
 // ignore_for_file: non_constant_identifier_names
 
 import 'dart:async';
-// ignore: avoid_web_libraries_in_flutter, by design
-import 'dart:html';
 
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/shared/analytics/_analytics_web.dart
```diff
@@ -802,7 +801,7 @@ Future<bool> disableAnalytics() async {
 /// devtoolsChrome.
 void computeDevToolsCustomGTagsData() {
   // Platform
-  final String platform = window.navigator.platform!;
+  final String platform = window.navigator.platform;
   platform.replaceAll(' ', '_');
```
kevmoo, line 805: This is a NO-OP, btw. `replaceAll` returns a `String`!!

@kenzieschmoll

kenzieschmoll: Noted! Let's leave for now since we are switching to unified_analytics soon.

kenzieschmoll (approved):

PR #6608, opened 2023-10-26
Remove io import from devtools_app that was only used for tests

kenzieschmoll:
This was getting tree shaken out of the web build so we didn't see an error, but we should not import `dart:io` under `devtools_app` unless it is for a Desktop specific implementation (usually under `config_specific/`)


jacob314 (approved):
lgtm

bkonyi (approved):

PR #6611, opened 2023-10-27
Fix some Windows issues in update-flutter-sdk

DanTup:
Some minor issues when running on Windows:

- Using `Uri.parse()` doesn't work on a Windows file path like `C:\foo\` (`Unsupported operation: Cannot extract a file path from a c URI`) so changed to `Uri.file()`
- Some calls to `flutter --version` were using `./bin/flutter` which fails on Windows (because programmatic executions require `flutter.bat`) so changed to `CliCommand.flutter`

@kenzieschmoll


kenzieschmoll (commented):
thanks for the fixes!

PR #6616, opened 2023-10-27
Disable the Raster Stats tool for Impeller

kenzieschmoll:
There are three options for how to disable this feature for Impeller:

1) Leave the `Raster Stats` tab in the performance page tab list, but show a message to communicate why the feature is not enabled (current impl of this PR)
Screenshot 2023-10-27 at 2 26 25 PM

This PR also ensures we show an error notification on the fallback case that something goes wrong with collecting raster stats.
Screenshot 2023-10-27 at 2 25 53 PM

The other two options considered are:
2) Remove the Raster Stats tab entirely when the Impeller backend is in use (though this may make users wonder where the tab went and this could be confusing to understand when this feature is and isn't supported)
3) Just show the pictured error notification when a user tried to use this tool on Impeller. I don't like this option either because this makes it seem like something is going wrong with DevTools and that is not the case.


parlough:
I'm not sure if you're actually looking for feedback on the options you proposed, but I agree that the 1st one is the best choice.

Now that Impeller is sometimes enabled by default could users get to this page without knowing a lot about Impeller? If so, consider adding a link after the existing text. Something like: "Learn more about the status of Impeller" or similar that links to https://docs.flutter.dev/perf/impeller.

jonahwilliams:
I agree the first option is the best. Adding a link to impeller docs could be useful, though in particular it won't answer when this particular tab will become useful again.

Perhaps I can file a tracking issue for that, but I'm not sure if linking that in the error message would be a reasonable thing to do for devtools.

packages/devtools_app/lib/src/screens/performance/panes/raster_stats/raster_stats.dart
```diff
@@ -21,12 +21,21 @@ class RasterStatsView extends StatelessWidget {
...
 
   final RasterStatsController rasterStatsController;
 
+  final bool impellerEnabled;
+
   @override
   Widget build(BuildContext context) {
+    if (impellerEnabled) {
+      return const CenteredMessage(
+        'The Raster Stats tool is not yet ready for the Impeller rendering '
```
jonahwilliams, line 35: I would adjust the phrasing here to something like "not currently available", which implies that it may change but I think doesn't imply as much urgency.

kenzieschmoll:
Adjusted based on review comments:
Screenshot 2023-10-27 at 3 04 34 PM

jonahwilliams (approved):
LGTM

parlough (approved):
Makes sense to me! Thanks for adding that link too. We'll likely be updating that page further soon :)

PR #6617, opened 2023-10-29
Make SemanticsNode.isMergedIntoParent Readonly

CaseyHillers:
Follow up from https://github.com/flutter/flutter/pull/137304

Breaks Google testing (b/308325747)

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [X] I signed the [CLA].
- [X] I listed at least one issue that this PR fixes in the description above.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I added new tests to check the change I am making, or there is a reason for not adding tests.

### DIFF
--- a/packages/devtools_app/lib/src/shared/primitives/custom_pointer_scroll_view.dart
+++ b/packages/devtools_app/lib/src/shared/primitives/custom_pointer_scroll_view.dart
@@ -812,7 +812,7 @@ class _RenderScrollSemantics extends RenderProxyBox {
     }
 
     _innerNode ??= SemanticsNode(showOnScreen: showOnScreen);
-    _innerNode!..rect = Offset.zero & node.rect.size;
+    _innerNode!.rect = Offset.zero & node.rect.size;
 
     int? firstVisibleIndex;
     final List<SemanticsNode> excluded = <SemanticsNode>[_innerNode!];
--- a/packages/devtools_app/lib/src/shared/primitives/custom_pointer_scroll_view.dart
+++ b/packages/devtools_app/lib/src/shared/primitives/custom_pointer_scroll_view.dart
@@ -811,8 +811,8 @@ class _RenderScrollSemantics extends RenderProxyBox {
       return;
     }
 
-    _innerNode ??= SemanticsNode(showOnScreen: showOnScreen);
-    _innerNode!.rect = Offset.zero & node.rect.size;
+    (_innerNode ??= SemanticsNode(showOnScreen: showOnScreen)).rect = node.rect;
+    
 
     int? firstVisibleIndex;
     final List<SemanticsNode> excluded = <SemanticsNode>[_innerNode!];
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/shared/primitives/custom_pointer_scroll_view.dart
```diff
@@ -812,9 +812,7 @@ class _RenderScrollSemantics extends RenderProxyBox {
     }
 
     _innerNode ??= SemanticsNode(showOnScreen: showOnScreen);
-    _innerNode!
-      ..isMergedIntoParent = node.isPartOfNodeMerging
-      ..rect = Offset.zero & node.rect.size;
+    _innerNode!..rect = Offset.zero & node.rect.size;
```
LongCatIsLooong, line 815: Looks like the CI is complaining about using `.` instead of the cascade operator `..`.

LongCatIsLooong (commented):
Sorry for the breakage and thank you for the fix! 

The upstream flutter/flutter change removed the `isMergedIntoParent` setter because it's no longer needed. The framework now manages that flag for you.

packages/devtools_app/lib/src/shared/primitives/custom_pointer_scroll_view.dart
CaseyHillers: Thanks!

Jasguerrero:
Is this the correct fix @LongCatIsLooong ? Although not blocking roll since a fix for b/308325747 was already applied we still want this right?

LongCatIsLooong:
Yeah the fix LGTM.

CaseyHillers:
@kenzieschmoll can you take a look? This is going to block the Dart SDK roll into Google.

Let me know if we need release notes for this change. For now, I'm going to mark it as exempt.

packages/devtools_app/lib/src/shared/primitives/custom_pointer_scroll_view.dart
```diff
@@ -812,9 +812,7 @@ class _RenderScrollSemantics extends RenderProxyBox {
     }
 
     _innerNode ??= SemanticsNode(showOnScreen: showOnScreen);
-    _innerNode!
-      ..isMergedIntoParent = node.isPartOfNodeMerging
-      ..rect = Offset.zero & node.rect.size;
+    _innerNode!.rect = Offset.zero & node.rect.size;
```
kenzieschmoll, line 815: can we do instead what is done on flutter master?
`(_innerNode ??= SemanticsNode(showOnScreen: showOnScreen)).rect = node.rect;`

(Most of the code in this file is copied from `scrollable.dart` from the Flutter framework)

kenzieschmoll (approved):

PR #6621, opened 2023-10-31
Prepare DCM workflow for license activation

elliette:
@incendial is ready to enable the DCM license for DevTools, which means we don't need the DCM keys to run the analyze step in our workflow bots.

This PR updates our workflow to remove the DCM keys, and therefore switch the trigger from `pull_request_target` to `pull_request`.


.github/workflows/build.yaml
```diff
@@ -40,6 +40,39 @@ jobs:
...
+        with:
+          ref: "${{ github.event.pull_request.head.sha }}"
+      - name: Load Cached Flutter SDK
+        uses: actions/cache@704facf57e6136b1bc63b828d79edcd491f0ee84
+        with:
+          path: |
+            ./flutter-sdk
+          key: flutter-sdk-${{ runner.os }}-${{ needs.flutter-prep.outputs.latest_flutter_candidate }}
+      - name: Run tool/bots.sh
+        run: ./tool/bots.sh
```
kenzieschmoll, line 61: why do we need to run the bots.sh script for running the DCM check?

kenzieschmoll (commented):
one comment then lgtm

kenzieschmoll (approved):

.github/workflows/build.yaml
elliette: If I remember correctly, it's because that's where we fetch the pub packages. I don't think DCM can analyze the code until the dependencies are fetched.

elliette: Hmm actually, since this is now in build.yaml, we might be able to remove this step (and the cloning step since that's handled above). Will see if it works removed

elliette: Ah no, we can't, see https://github.com/flutter/devtools/actions/runs/6724383676/job/18276582319?pr=6635

PR #6626, opened 2023-10-31
Work towards migrating `devtools_app` to `package:web`

kenzieschmoll:
Migrates most of `devtools_app` to use `package:web`. There are a handful of remaining uses to migrate that are not straightforward or where `package:web` is missing some getters. I documented those on https://github.com/flutter/devtools/issues/6606. 

I also cleaned up some of the conditional imports while I was here (renaming files, removing unnecessary stub files, etc.).


packages/devtools_app/lib/src/extensions/embedded/_view_web.dart
```diff
@@ -108,9 +108,9 @@ class _ExtensionIFrameController extends DisposableController
       }),
     );
 
-    html.window.addEventListener(
+    window.addEventListener(
       'message',
-      _handleMessageListener = _handleMessage,
+      _handleMessageListener = _handleMessage as EventListener,
```
srujzs, line 113: I believe this will fail since you're casting a Dart function to a JS function type. Prefer `_handleMessage.toJS as EventListener` instead.

As general guidance (until we get lints to ensure this!), avoid casts between Dart and JS interop types (casts between JS interop types are generally okay, granted you did the work to type check them or they are trivially true). They may work on one platform but not another.

packages/devtools_app/lib/src/extensions/embedded/_view_web.dart
```diff
@@ -156,14 +156,14 @@ class _ExtensionIFrameController extends DisposableController
...
+      message.jsify(),
+      embeddedExtensionController.extensionUrl.toJS,
     );
   }
 
-  void _handleMessage(html.Event e) {
-    if (e is html.MessageEvent) {
-      final extensionEvent = DevToolsExtensionEvent.tryParse(e.data);
+  void _handleMessage(Event e) {
+    if (e is MessageEvent) {
```
srujzs, line 165: This is one of the big pitfalls of interop alas. For context: in `dart:html`, we did the expensive checking to ensure that a `MessageEvent` is *actually* a `MessageEvent` under the hood. With `package:web`, these are all just different views of a JS object, so this `is` check will always pass, even if `e` is not a `MessageEvent`. This will become a bit more obvious once we generate `package:web` types using extension types with a representation type of `JSObject`. Hopefully there should be a lint here telling you that you're actually doing a `is JSObject` check, reducing the confusion even further.

The right way to do this is `(e as JSObject).instanceOfString('MessageEvent')` instead. When `Event` is an extension type on `JSObject`, you won't need the cast.

packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_controller_web.dart
```diff
@@ -166,7 +165,7 @@ class PerfettoControllerImpl extends PerfettoController {
     );
     _initialized = true;
 
-    _perfettoIFrame = html.IFrameElement()
+    _perfettoIFrame = createElementTag('iframe') as HTMLIFrameElement
```
srujzs, line 168: Note that this cast is okay in contrast with the cast above. You're casting between interop types and you know it's that type. I *think* this isn't a lint in the future with extension types.

packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_web.dart
```diff
@@ -137,9 +137,9 @@ class _PerfettoViewController extends DisposableController
       }),
     );
 
-    html.window.addEventListener(
+    window.addEventListener(
       'message',
-      _handleMessageListener = _handleMessage,
+      _handleMessageListener = _handleMessage as EventListener,
```
srujzs, line 142: Same comment as above.

packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_web.dart
```diff
@@ -261,14 +261,14 @@ class _PerfettoViewController extends DisposableController
     _postMessage(message);
   }
 
-  void _handleMessage(html.Event e) {
-    if (e is html.MessageEvent) {
-      if (e.data == EmbeddedPerfettoEvent.pong.event &&
+  void _handleMessage(Event e) {
+    if (e is MessageEvent) {
```
srujzs, line 265: Same comment as above.

packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_web.dart
```diff
@@ -261,14 +261,14 @@ class _PerfettoViewController extends DisposableController
...
+      if (e.data.toString() == EmbeddedPerfettoEvent.pong.event &&
```
srujzs, line 266: Hah! This is funny, because I think this will work since `toString` is equivalent to `toDart` for JS strings on all backends. Still, we should avoid relying on that and use `toDart` conversions e.g. `(e.data! as JSString).toDart`.

packages/devtools_app/lib/src/shared/config_specific/framework_initialize/_framework_initialize_web.dart
```diff
@@ -15,7 +17,7 @@ import '../../server_api_client.dart';
 /// Return the url the application is launched from.
 Future<String> initializePlatform() async {
   // Clear out the unneeded HTML from index.html.
-  for (var element in document.body!.querySelectorAll('.legacy-dart')) {
+  for (var element in html.document.body!.querySelectorAll('.legacy-dart')) {
```
srujzs, line 20: Is this kept as `dart:html` because `NodeList` is not a `List`?

packages/devtools_extensions/lib/src/template/_simulated_devtools_environment/_simulated_devtools_controller.dart
```diff
@@ -17,12 +17,12 @@ class SimulatedDevToolsController extends DisposableController
...
   /// removed in [dispose].
-  html.EventListener? _handleMessageListener;
+  EventListener? _handleMessageListener;
 
   void init() {
-    html.window.addEventListener(
+    window.addEventListener(
       'message',
-      _handleMessageListener = _handleMessage,
+      _handleMessageListener = _handleMessage as EventListener,
```
srujzs, line 25: Same comment as above.

packages/devtools_extensions/lib/src/template/_simulated_devtools_environment/_simulated_devtools_controller.dart
```diff
@@ -32,9 +32,9 @@ class SimulatedDevToolsController extends DisposableController
     });
   }
 
-  void _handleMessage(html.Event e) {
-    if (e is html.MessageEvent) {
-      final extensionEvent = DevToolsExtensionEvent.tryParse(e.data);
+  void _handleMessage(Event e) {
+    if (e is MessageEvent) {
```
srujzs, line 36: Same comment as above.

packages/devtools_extensions/lib/src/template/extension_manager.dart
```diff
@@ -39,13 +39,13 @@ class ExtensionManager {
...
-  html.EventListener? _handleMessageListener;
+  EventListener? _handleMessageListener;
 
   // ignore: unused_element, false positive due to part files
   void _init({required bool connectToVmService}) {
-    html.window.addEventListener(
+    window.addEventListener(
       'message',
-      _handleMessageListener = _handleMessage,
+      _handleMessageListener = _handleMessage as EventListener,
```
srujzs, line 48: Same comment as above.

packages/devtools_extensions/lib/src/template/extension_manager.dart
```diff
@@ -73,13 +73,13 @@ class ExtensionManager {
...
-    html.window.removeEventListener('message', _handleMessageListener);
+    window.removeEventListener('message', _handleMessageListener);
     _handleMessageListener = null;
   }
 
-  void _handleMessage(html.Event e) {
-    if (e is html.MessageEvent) {
-      final extensionEvent = DevToolsExtensionEvent.tryParse(e.data);
+  void _handleMessage(Event e) {
+    if (e is MessageEvent) {
```
srujzs, line 81: Same comment as above. :)

srujzs (commented):
Thanks for doing this migration, and sorry for all the feedback!

packages/devtools_app/lib/src/shared/config_specific/notifications/_notifications_web.dart
```diff
@@ -0,0 +1,27 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
```
CoderDake, line 1:
```suggestion
// Copyright 2023 The Chromium Authors. All rights reserved.
```

CoderDake (approved):

packages/devtools_app/lib/src/extensions/embedded/_view_web.dart
kenzieschmoll: Since `Event` implements `JSObject`, do I still need the cast? I'm getting an unnecessary cast lint when I add it.

packages/devtools_app/lib/src/shared/config_specific/notifications/_notifications_web.dart
kenzieschmoll: this was just a rename of the file, so will leave the imports as is.

packages/devtools_app/lib/src/extensions/embedded/_view_web.dart
srujzs: Ah no, I forgot that we subtype it. You can just do `e.instanceOfString('MessageEvent')`.

packages/devtools_app/lib/src/shared/config_specific/framework_initialize/_framework_initialize_web.dart
kenzieschmoll: Yep:
```
The type 'NodeList' used in the 'for' loop must implement 'Iterable'.dart[for_in_of_invalid_type](https://dart.dev/diagnostics/for_in_of_invalid_type)
```

srujzs: I think long-term we're still deciding the right way to handle list-like JS interfaces. One option would be a conversion function `toDartListLike` that would allow users to treat these interfaces like a `List` with `external` accesses.

For now though, we can use interop in this file to iterate:

```dart
extension NodeListExtension on NodeList {
  external void forEach(JSFunction callback);
}

...

document.body!.querySelectorAll('.legacy-dart').forEach((Node element) {
  // This should be the equivalent of `element.remove()` from `dart:html`.
  if (element.parent != null) {
    element.parent!.removeChild(element);
  }
}.toJS);

```

There's a bug where `forEach` is not emitted on `iterable` types so that's why there's a separate definition here: https://github.com/dart-lang/web/issues/21

kenzieschmoll: I think `Node` is the wrong type here. Do I need to cast `element` to something else for this to work?
Screenshot 2023-10-31 at 3 05 33 PM

srujzs: Egg on my face for not validating that API exists first. I believe the right API is `parentNode`, not `parent`. :)

kevmoo (approved):

srujzs (approved):

Issue #6324, opened 2023-09-06
Convert shell scripts to dart scripts


CoderDake:
Look into converting the following over to dart scripts:
- [ ] bots.sh
- [ ] build_release.sh
- [ ] fix_goldens.sh
- [ ] flutter_tests.sh
- [ ] generate_code.sh
- [ ] latest_flutter_candidate.sh
- [ ] pub_upgrade.sh
- [ ] pull_and_refresh.sh
- [ ] refresh.sh
- [ ] release_helper.sh
- [ ] tag_version.sh
- [ ] update_flutter_sdk.sh
- [ ] update_perfetto.sh
- [ ] upgrade.sh

PR #6634, opened 2023-11-01
Convert the build_release.sh script to a dart script

kenzieschmoll:
work towards: https://github.com/flutter/devtools/issues/6324

@DanTup this should enable this script to work on windows.

### DIFF
--- a/tool/lib/commands/build_release.dart
+++ b/tool/lib/commands/build_release.dart
@@ -1,0 +1,84 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:args/command_runner.dart';
+import 'package:devtools_tool/model.dart';
+import 'package:io/io.dart';
+import 'package:path/path.dart' as path;
+
+import '../utils.dart';
+
+const _useLocalFlutterFlag = 'use-local-flutter';
+const _updatePerfettoFlag = 'update-perfetto';
+
+class BuildReleaseCommand extends Command {
+  BuildReleaseCommand() {
+    argParser
+      ..addFlag(
+        _useLocalFlutterFlag,
+        negatable: false,
+        defaultsTo: false,
+        help:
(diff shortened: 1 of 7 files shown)


tool/lib/commands/build_release.dart
```diff
@@ -0,0 +1,84 @@
...
+      ],
+      workingDirectory: repo.devtoolsAppDirectoryPath,
+    );
+
+    final canvaskitDir = Directory(
+      path.join(repo.devtoolsAppDirectoryPath, 'build', 'web', 'canvaskit'),
+    );
+    for (final file in canvaskitDir.listSync()) {
+      if (RegExp(r'canvaskit\..*').hasMatch(file.path)) {
+        await processManager.runProcess(CliCommand('chmod 0755 ${file.path}'));
```
kenzieschmoll, line 80: @DanTup do you know the preferred way to do this on windows? I searched on Stack overflow but didn't see one definitive answer

DanTup: Is it necessary?

I can test it tomorrow when back on PC, but for the chmod issue in the build_and_copy script I just removed it and didn't see any issues. It's not obvious to me what this is doing though - why is the permissions change necessary for non-Windows?

kenzieschmoll: I added a TODO to investigate whether changing file permissions is necessary on windows or at all anymore.

tool/lib/commands/build_release.dart
```diff
@@ -0,0 +1,89 @@
...
+    final useLocalFlutter = argResults![_useLocalFlutterFlag];
+    final updatePerfetto = argResults![_updatePerfettoFlag];
+
+    if (!useLocalFlutter) {
+      logStatus('updating tool/flutter-sdk to the latest flutter candidate');
+      await processManager.runProcess(CliCommand.tool('update-flutter-sdk'));
+    }
+
+    if (updatePerfetto) {
+      logStatus('updating the bundled Perfetto assets');
```
elliette, line 56: suggestion: should the comment for now to be "skipping updating the bundled Perfetto assets" so we aren't confused why the perfetto assets don't get updated?

elliette (approved):

Issue #6633, opened 2023-11-01
DevTools extensions feedback


DanTup:
I went through the readme at https://pub.dev/packages/devtools_extensions on Windows to build and test a sample extension and hit some bugs/had some feedback.

(note: I haven't yet tried to have my extension interact with the app, this testing was only getting the Flutter app working correctly inside DevTools both in the simulated and real environment).

- [x] The sample code for `FooPackageDevToolsExtension` is missing Flutter imports and also references a non-existent widget `FooDevToolsExtension`
  I think it'd be better to make it a complete file and use `Placeholder` or something so while trying to get the basics working the user doesn't need to create any additional widgets of their own
  - addressed by https://github.com/flutter/devtools/pull/6639
- [x] The path in the `launch.json` example doesn't seem to match the structure (`"foo/extension/foo_devtools_extension/lib"` vs `"packages/foo_devtools_extension/lib"`)
  While the user will need to tweak this, it should be consistent so it's clearer what should be changed
  - addressed by https://github.com/flutter/devtools/pull/6639
- [x] `Member not found: 'profilePlatformChannels'` when using current beta
  Fix in progress at https://github.com/flutter/devtools/pull/6632
- [x] The example command for `build_and_copy` is written using && and backslash continuations so is awkward to copy/paste for Windows
  Split across multiple lines?
  - addressed by https://github.com/flutter/devtools/pull/6639
- [x] build_and_copy fails due to trying to run `flutter` but needing to be `flutter.bat` on Windows
  https://github.com/flutter/devtools/blob/d5ce5a879060e515cfd52bb1af2c48ba8f04bb17/packages/devtools_extensions/bin/_build_and_copy.dart#L59
  - addressed by https://github.com/flutter/devtools/pull/6639
- [x] build_and_copy fails because it tries to run `chmod` which doesn't exist on Windows
  https://github.com/flutter/devtools/blob/d5ce5a879060e515cfd52bb1af2c48ba8f04bb17/packages/devtools_extensions/bin/_build_and_copy.dart#L75
  - addressed by https://github.com/flutter/devtools/pull/6639
- [x] Probably known, but doc has `<TODO: insert version>` - fixed by https://github.com/flutter/devtools/pull/6631
- [x] Initially I misread the docs and entered `name: "Danny's First Extension"` in my `config.yaml` (instead of the name of my package like `name: danny_library`). This prevented the "Enable" button from working when I loaded in DevTools
[...]

PR #6639, opened 2023-11-01
Address some devtools extensions feedback related to Windows

kenzieschmoll:
Address some of the feedback from https://github.com/flutter/devtools/issues/6633

### DIFF
--- a/packages/devtools_extensions/README.md
+++ b/packages/devtools_extensions/README.md
@@ -113,16 +113,16 @@ In `lib/main.dart`, place a `DevToolsExtension` widget at the root of your app:
 import 'package:devtools_extensions/devtools_extensions.dart';
 
 void main() {
-  runApp(const FooPackageDevToolsExtension());
+  runApp(const FooDevToolsExtension());
 }
 
-class FooPackageDevToolsExtension extends StatelessWidget {
-  const FooPackageDevToolsExtension({super.key});
+class FooDevToolsExtension extends StatelessWidget {
+  const FooDevToolsExtension({super.key});
 
   @override
   Widget build(BuildContext context) {
     return const DevToolsExtension(
-      child: FooDevToolsExtension(),
+      child: Placeholder(),
     );
   }
 }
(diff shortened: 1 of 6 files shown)


packages/devtools_extensions/README.md
```diff
@@ -185,11 +186,13 @@ real DevTools environment. Build your flutter web app and copy the built assets
...
-dart run devtools_extensions build_and_copy \
-  --source=. \
-  --dest=path/to/your_pub_package/extension/devtools 
+cd your_extension_web_app
+```
+```sh
+flutter pub get
+```
+```sh
+dart run devtools_extensions build_and_copy --source=. --dest=path/to/your_pub_package/extension/devtools
```
DanTup, line 195: Maybe `--dest=../foo/extension/devtools` to more closely match the other examples? (They'll likely need to change "foo", but possible the rest would be correct).

packages/devtools_extensions/example/foo/.vscode/launch.json
```diff
@@ -0,0 +1,57 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "foo_devtools_extension",
```
DanTup, line 8: Is there much value in running without the simulated environment? If not, maybe we should remove this or put the simulated environment at the top (which is basically the default if it's the first time you open the example).

packages/devtools_extensions/extension_config_spec.md
```diff
@@ -3,7 +3,7 @@ The `config.yaml` file for a DevTools extension must follow the format below.
 ## Required fields
 
 - `name` : the package name that this DevTools extension belongs to. The value of this field
-will be used in the extension page title bar.
+will be used in the extension page title bar. This name should not contain spaces or special characters.
```
DanTup, line 6: Does "special characters" include underscores? If so, if your package is called "foo_bar", what should you put here (it says "package name").

DanTup (approved):
LGTM! Some minor nits/questions

PR #6647, opened 2023-11-02
Add service extension example to devtools_extensions

kenzieschmoll:
This adds an example of calling `registerServiceExtension` from `package:foo` and calling that service extension from the `foo_devtools_extension`.

This cleans up the extension UI to make it more clear what each example is for:
Screenshot 2023-11-02 at 12 50 01 PM


packages/devtools_extensions/example/foo/packages/foo_devtools_extension/lib/src/service_extension_example.dart
```diff
@@ -0,0 +1,235 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
```
bkonyi, line 1: Consider adding some documentation at the top of this file describing what's going on in this file at a high level. Including references to service extension documentation would also be good.

bkonyi (approved):
Overall LGTM.

I think we should revisit these examples and over-document them. Developers will be working off of these examples, so we should make a point to be super clear as to what functionality each example is exercising and why that functionality may be useful to an extension author. The documentation with the default Flutter counter app is a good example.

PR #6651, opened 2023-11-02
Fix DevTools extension analytics

kenzieschmoll:
@jonasfj this PR adds a field `isPublic` for a DevTools extension, meaning this extension was shipped in a public pub package. To determine this we are checking:
```dart
final isPublic = extension.rootUri.path.contains(path.join('.pub-cache', 'hosted'));
```

Is there a better way to do this? Would it be possible to add a field in the `package_config.json` so that we do not have to infer from the root uri?


packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -78,6 +78,8 @@ class ExtensionsManager {
       }
       for (final extension in extensions) {
         final config = extension.config;
+        final isPublic =
+            extension.rootUri.path.contains(path.join('.pub-cache', 'hosted'));
```
jacob314, line 82: add a TODO and file a bug to not depend on the internals of how the pub-cache directories are specified.

jacob314 (approved):
lgtm_dog

jacob314:
> @jonasfj this PR adds a field `isPublic` for a DevTools extension, meaning this extension was shipped in a public pub package. To determine this we are checking:
> 
> ```dart
> final isPublic = extension.rootUri.path.contains(path.join('.pub-cache', 'hosted'));
> ```
> 
> Is there a better way to do this? Would it be possible to add a field in the `package_config.json` so that we do not have to infer from the root uri?
+ @sigurdm  as @jonasfj is OOO

packages/devtools_app/lib/src/shared/analytics/constants/_extension_constants.dart
```diff
@@ -21,34 +21,38 @@ enum DevToolsExtensionEvents {
...
 
   /// Event sent when an extension is force reloaded from the extension screen
   /// context menu.
-  static String extensionForceReload(String name) =>
-      'extensionForceReload-$name';
+  static String extensionForceReload(DevToolsExtensionConfig ext) =>
+      'extensionForceReload-${_extensionName(ext)}';
 }
+
+String _extensionName(DevToolsExtensionConfig ext) => ext.gaName;
```
jacob314, line 58: nit: why add this private method. Its name seems less clear than ext.gaName and it is doing nothing but passing though the call.

packages/devtools_shared/lib/src/extensions/extension_model.dart
```diff
@@ -104,14 +108,21 @@ class DevToolsExtensionConfig implements Comparable {
   /// See https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/material/icons.dart.
   final int materialIconCodePoint;
 
+  /// Whether this extension is distrubuted in a public package on pub.dev.
+  final bool isPublic;
+
   String get displayName => name.toLowerCase();
 
+  /// The extension name that will be used in analytics.
```
jacob314, line 116: nit: I would change this method name to make the method comment unneeded. E.g. 
analyticsSafeName

sigurdm:
> ```dart
> final isPublic = extension.rootUri.path.contains(path.join('.pub-cache', 'hosted'));
> ```
> Is there a better way to do this? Would it be possible to add a field in the package_config.json so that we do not have to infer from the root uri?

I think I need to understand the use-case better. What do you mean by public? Pub doesn't really have such a concept.
What behavior is gonna depend on the package being public?

Pub has a concept of the "source" of a dependency - where can it find the package.

It can be one of `path`, `git` and `hosted`.

If it is hosted it can either be from `https://pub.dev` (the default) or from some other package repository.

Currently there is no way of knowing the source from only the package_config.json, except by relying on pub's cache conventions (which you probably shouldn't). And I would be somewhat reluctant to add this info, as it kind of ruins the separation of concerns that package_config.json is helping us maintain.

You could probably tease this information from pubspec.lock, it contains a machine-independent version of the package-resolution including the source...

Note also that the default pub-cache is not at `~/.pub-cache` on windows, but at `%LOCALAPPDATA%/Pub/Cache`. So this code will probably not work on windows. Also the pub-cache location can be overridden by the `$PUB_CACHE` variable.

bkonyi (approved):

kenzieschmoll:
Discussed offline with @sigurdm. To fix this the right way, we need to pull data from the `pubspec.lock` file, but to do that we need to consider the possibility that the `pubspec.lock` file does not exist. For now, we will use this check as a workaround, and we will migrate to a more robust solution later:
```dart
final isPubliclyHosted = extension.rootUri.path.contains('pub.dev') ||
     extension.rootUri.path.contains('pub.flutter-io.cn');
```

Documented at https://github.com/dart-lang/pub/issues/4042.

PR #6666, opened 2023-11-06
Validate extension names and warn in devtools_app for server-side extension errors

kenzieschmoll:
Addresses remaining items at https://github.com/flutter/devtools/issues/6633.

This PR adds validation for the "name" field in an extension's `config.yaml` file and logs a warning to the devtools app console when issues have been encountered:

Screenshot 2023-11-06 at 3 45 23 PM

### DIFF
--- a/packages/devtools_shared/lib/src/extensions/extension_manager.dart
+++ b/packages/devtools_shared/lib/src/extensions/extension_manager.dart
@@ -117,7 +117,7 @@ class ExtensionsManager {
     if (parsingErrors.isNotEmpty) {
       throw ExtensionParsingException(
         'Encountered errors while parsing extension config.yaml '
-        'files:\n${parsingErrors.toString()}',
+        'files:\n$parsingErrors',
       );
     }
   }
--- a/packages/devtools_shared/lib/src/extensions/extension_manager.dart
+++ b/packages/devtools_shared/lib/src/extensions/extension_manager.dart
@@ -15,10 +15,6 @@ import 'extension_model.dart';
 /// DevTools assets are served (build/).
 const extensionRequestPath = 'devtools_extensions';
 
-/// The location for the DevTools extension, relative to the parent package's
-/// root.
-const extensionLocation = 'extension/devtools';
-
 /// The default location for the DevTools extension, relative to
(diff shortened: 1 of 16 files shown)


packages/devtools_shared/test/extensions/extension_model_test.dart
```diff
@@ -181,5 +181,56 @@ void main() {
...
+            'path': 'path/to/foo/extension',
+            'issueTracker': 'www.google.com',
+            'version': '1.0.0',
+            'isPubliclyHosted': 'false',
+          });
+        },
+        throwsInvalidNameError(),
+      );
+
+            expect(
```
DanTup, line 222: nit: indenting looks wrong here

DanTup (approved):

packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -114,6 +113,13 @@ class ExtensionsManager {
       for (final extension in devtoolsExtensions)
         _moveToServedExtensionsDir(extension.name, extension.path),
     ]);
+
+    if (parsingErrors.isNotEmpty) {
+      throw ExtensionParsingException(
+        'Encountered errors while parsing extension config.yaml '
+        'files:\n${parsingErrors.toString()}',
```
bkonyi, line 120: Nit: `$parsingErrors` should be equivalent to `${parsingErrors.toString()}`.

packages/devtools_shared/lib/src/extensions/extension_model.dart
```diff
@@ -36,6 +36,18 @@ class DevToolsExtensionConfig implements Comparable {
           versionKey: final String version,
           isPubliclyHostedKey: final String isPubliclyHosted,
         }) {
+      final underscoresAndLetters = RegExp(r'[a-z0-9_]*');
```
bkonyi, line 39: Consider using this expression instead so you don't have to deal with matches on substrings of `name`: `RegExp(r'^[a-z0-9]*$')`

Then you can just do this:

```dart
if (!underscoresAndLetters.hasMatch(name)) {
  throw StateError(
    // ...
  );
}
```

bkonyi (approved):

PR #6668, opened 2023-11-07
Fix an issue locating the Flutter SDK on Windows when not invoked via 'dart' inside a Flutter SDK

DanTup:
If the current Dart VM isn't inside a Flutter SDK, we fall back to trying to find it on `PATH` with `which`. The equiv on Windows is `where.exe`.

### DIFF
--- a/tool/lib/model.dart
+++ b/tool/lib/model.dart
@@ -128,7 +128,6 @@ class FlutterSdk {
     }
 
     // Look to see if we can find the 'flutter' command in the PATH.
-    // TODO(dantup): This won't work on Windows.
     final whichCommand = Platform.isWindows ? 'where.exe' : 'which';
     final result = Process.runSync(whichCommand, ['flutter']);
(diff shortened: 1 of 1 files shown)


DanTup:
Question: Can I add the autosubmit label before there's a review, and the bot will merge it once reviews - or does it only work on the bot checks? @CoderDake ?

CoderDake:
@DanTup I'm not 100% sure but we are about to find out 😄

CoderDake (approved):

DanTup:
I guess my real concern was whether it would merge it when everything went green but there still wasn't an approving review (I don't know whether GitHub's rules apply to the bot). So to test, we should add the tag to a review that is good, but without an approval.

I'll test with https://github.com/flutter/devtools/pull/6669 since it'd be handy to know :)

CoderDake:
@DanTup looks like it submitted on it's own after the approval :D

PR #6669, opened 2023-11-07
Use located Flutter instead of PATH flutter for Command.flutter() + support FLUTTER_ROOT override

DanTup:
This fixes a TODO where Command.flutter() would always run the version of Flutter from PATH and not the same one we would detect.

This will allow me to fix an issue when using a custom version of DevTools in VS Code (calling "devtools_tool serve") without having to ensure my PATH points at the same version (which it probably doesn't when I'm just opening test apps that use my "normal" checkout of Flutter).

### DIFF
--- a/tool/lib/commands/analyze.dart
+++ b/tool/lib/commands/analyze.dart
@@ -20,7 +20,7 @@ class AnalyzeCommand extends Command {
 
   @override
   Future run() async {
-    final sdk = FlutterSdk.getSdk();
+    final sdk = FlutterSdk.current;
     if (sdk == null) {
       print('Unable to locate a Flutter sdk.');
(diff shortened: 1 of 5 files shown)


CoderDake (approved):

DanTup:
@CoderDake I'm resetting your review and adding auto-submit as a test to see whether the bot will/can merge without the reviews. Assuming it hasn't happened in ~10min or so, we can probably assume it's save (and you can re-approve this) :)

DanTup:
Oh, that didn't work... re-requesting your review doesn't clear the previous approval for these rules.

image

DanTup:
Apparently I confused GitHub too :-)

image

PR #6670, opened 2023-11-07
Allow --machine/--allow-embedding to be passed to devtools_tool serve

DanTup:
The arg parser allows additional arguments, but not additional "options", so we need to add this explicitly* to allow `--machine` to work. The code already passes the extra args through to the server correctly, it just wasn't allowing this option to be provided on the command line.

\* Or use the `AllowAnythingParser`, but looks like that would affect the existing options.

### DIFF
--- a/tool/lib/commands/serve.dart
+++ b/tool/lib/commands/serve.dart
@@ -14,6 +14,8 @@ import '../utils.dart';
 const _useLocalFlutterFlag = 'use-local-flutter';
 const _updatePerfettoFlag = 'update-perfetto';
 const _buildAppFlag = 'build-app';
+const _machineFlag = 'machine';
+const _allowEmbeddingFlag = 'allow-embedding';
 
 /// This command builds DevTools in release mode by running the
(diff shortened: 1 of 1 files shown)


DanTup:
(converted to draft... there are other flags we pass in VS Code like `--try-ports` we also need to support)

DanTup:
Actually, I'm ditching `--try-ports` from VS Code since we're passing 10 which is the default value, but we did need allow-embedding.

DanTup:
(this is ready for review now and hopefully the bots will be green 🤞)

kenzieschmoll (approved):
I'm surprised that these weren't passed through automatically since we pass any unhandled arguments along to the dds command. but lgtm

DanTup:
> I'm surprised that these weren't passed through automatically since we pass any unhandled arguments along to the dds command

I was too.. seems to be a difference between arguments and the named options. There's an `AllowAnythingParser` that would allow these, but I think we'd then have to also read the other options ourselves, so since we only need these two I figured this was better. We can always revisit if we want to pass more in future.

PR #6671, opened 2023-11-07
Remove unnecessary async keywords

DanTup:
Latest version of DCM flags these so the bots turned red. This fixes the build by removing the unnecessary `async`s.

### DIFF
--- a/packages/devtools_app/lib/src/service/vm_service_wrapper.dart
+++ b/packages/devtools_app/lib/src/service/vm_service_wrapper.dart
@@ -114,7 +114,7 @@ class VmServiceWrapper extends VmService {
     String isolateId, {
     bool? reset,
     bool? gc,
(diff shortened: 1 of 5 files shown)


DanTup:
I feel like this shouldn't change anything (`async` just says `await` is a keyword within the method?), but it's failing with this:

```
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞═════════════════
The following _Exception was thrown running a test:
Exception: There is no registered method for service "fakeMethod"
```

I re-ran the bot thinking maybe it was a flake, but it failed a second time so now I'm unsure 🤔

kenzieschmoll:
Sorry just seeing this PR Danny. The DCM errors were fixed and merged in https://github.com/flutter/devtools/pull/6666.

PR #6678, opened 2023-11-08
Prepare for roll to G3.

polina-c:


kenzieschmoll:
In the future - we should only be bumping the minor version for releases into the Dart SDK, and we do not need to do a version bump before rolling in to google3.

Issue #6679, opened 2023-11-08
devtools_tool release-helper does not create PR


polina-c:
Used guidance: https://github.com/flutter/devtools/blob/master/tool/RELEASE_INSTRUCTIONS.md#create-a-release-pr

Output:

```
...

 > devtools_tool update-version current-version
2.30.0
 > git commit -a -m Prepare for release (exitCode: 0, stderr: , stdout: 2.30.0
)
fatal: paths 'for ...' with -a does not make sense
ProcessException: Failed with exit code: 128. 
  Command: git commit -a -m Prepare for release (exitCode: 0, stderr: , stdout: 2.30.0
)
```

polina-c:
This is what was changed: https://github.com/flutter/devtools/pull/6678
Commit: https://github.com/flutter/devtools/pull/6678/commits/43d8fe8f4e9eb24b1fcd2e8c4c9692d39d6c7584

DanTup:
I think the issue is that the commit message doesn't have quotes - but adding quotes raises another issue. I have a possible fix here:

https://github.com/flutter/devtools/pull/6684

But to avoid similar issues I think we should remove all of the splitting of args by spaces.

PR #6684, opened 2023-11-08
Fix release-helper issues

DanTup:
- Fix a Windows issue caused by trying to run `devtools_tool` without `.bat`
- Fix handling of arguments to git for commit message (Fixes https://github.com/flutter/devtools/issues/6679)

I think we should probably remove all the code that's taking strings for arguments and splitting on spaces because it seems unsafe. It is convenient to be able to do:

```dart
CliCommand.git('fetch upstream/master')
```

instead of:

```dart
CliCommand.git(['fetch', 'upstream/master'])
```

However it becomes easy to pass variables that fail (like the commit message in the example here). Rather than doubling up the constructors (as I have here for the git one), it might be better to just be consistent and always use a `List`? (@kenzieschmoll?)


DanTup:
I've changed this to just be the Windows fix now (@CoderDake @kenzieschmoll).

CoderDake (approved):

PR #6685, opened 2023-11-08
Fix release helper git command

CoderDake:
Since some git commands take strings, with spaces, I've added a way for us to avoid splitting the args.

Fixes https://github.com/flutter/devtools/issues/6679

### DIFF
--- a/tool/lib/utils.dart
+++ b/tool/lib/utils.dart
@@ -101,14 +101,13 @@ class CliCommand {
     bool throwOnException = true,
     bool split = true,
   }) {
-    assert(
-      (cmd != null) || (args != null),
-      'One of `cmd` and `args` must be specified',
-    );
-    assert(
-      (cmd == null) != (args == null),
-      'Only one of `cmd` and `args`, at a time, may be specified',
-    );
+    if ((cmd != null) || (args != null)) {
+      throw ('One of `cmd` and `args` must be specified');
+    }
+
+    if ((cmd == null) != (args == null)) {
+      throw ('Only one of `cmd` and `args`, at a time, may be specified');
+    }
 
     if (cmd != null) {
       args = cmd.split(' ');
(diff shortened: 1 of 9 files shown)


tool/lib/utils.dart
```diff
@@ -88,13 +88,27 @@ class CliCommand {
...
+  /// CliCommand helper for running git commands.
+  ///
+  /// Arguments can be passed in as a single string, this helper will split them
+  /// up using their spaces. e.g. CliCommand.git(['checkout test-branch'])
+  ///
+  /// If you don't want the args to be split by their spaces then you can set
+  /// [split] to false.
   factory CliCommand.git(
-    String args, {
+    List<String> args, {
```
kenzieschmoll, line 99: wdyt about instead having two parameters: 
```
factory CliCommand.git({
  String? cmd,
  List<String>? args,
  bool throwOnException = true,
}) : assert((cmd == null) != (args == null), 'Only one of `cmd` and `args` may be specified) {
...
}
```

I worry that it isn't intuitive to be passing a single string to a list in the most common case. Having named parameters would make this more clear.

CoderDake: I dig it!

kenzieschmoll:
FYI looks like you and @DanTup  are trying to fix the same issue: https://github.com/flutter/devtools/pull/6684

DanTup:
Yeah, I mentioned that to Dan earlier 😄

Let's go with this one and I'll dump mine (or switch it to just the Windows fixes unless @CoderDake wants to include them here).

tool/lib/utils.dart
```diff
@@ -90,22 +90,29 @@ class CliCommand {
...
-      args = args.map((e) => e.split(' ')).reduce((value, element) {
-        value.addAll(element);
-        return value;
-      });
+    if ((cmd == null) && (args == null)) {
+      throw ('One of `cmd` and `args` must be specified');
+    }
+
+    if ((cmd != null) && (args != null)) {
+      throw ('Only one of `cmd` and `args`, at a time, may be specified');
```
kenzieschmoll, line 109: this can all be combined
```dart
if ((cmd == null) == (args == null)) {
  throw Exception('Exactly one of `cmd` and `args` must be specified.);
}
```

kenzieschmoll (approved):
one comment then lgtm

Issue #6563, opened 2023-10-19
Feedback on Flutter Sidebar


anderdobo:
**Setup**
1. Flutter beta channel
2. VS Code with pre-release Flutter extension v3.75.20231002

**Feedback**
1. Recommendation: For the DevTools options in the Debug Sessions section, add **Open DevTools in Web browser**  option
3. In Devices section, Chrome and macOS show even when a project doesn't have these enabled. Consider having the behavior be  like in the command palette, where platforms that are not enabled say "Enable [platform] for this project"
4. Android virtual device in Devices section is titled "sdk gphone64 arm64". Ideally should be titled the same as in VS Code (command palette and device picker in the bottom right) which was "Pixel 7 Pro API 34" in this case. Interestingly, the iPhone simulator was correctly titled "iPhone 15 Pro Max" in the sidebar.
5. Switching between devices in the Devices section of the sidebar changes the device in the device picker in the bottom right of the window, including "selecting" platforms that are not enabled in the project. Hitting F5 doesn't do anything when a platform that's not enabled is selected.

<img width="2344" alt="Screenshot 2023-10-19 at 1 47 34 PM" src="https://github.com/flutter/devtools/assets/5350105/baa2d3f9-06a7-4864-a9c4-40b5be77aac6">

anderdobo:
FYI @kenzieschmoll and @DanTup

DanTup:
I've pushed some changes to Dart-Code to:

1. Use the emulators names when a device is an emulator
2. Hide unavailable platforms from the device list

Adding an option to "Enable [platform]" is a little more involved (it requires work in both Dart-Code and DevTools). I've opened https://github.com/Dart-Code/Dart-Code/issues/4820 for the Dart-Code work that needs doing first.

> 1. Recommendation: For the DevTools options in the Debug Sessions section, add Open DevTools in Web browser option

This also needs some tweaks on both sides. I've opened https://github.com/Dart-Code/Dart-Code/issues/4821 for the Dart-Code side to extend the API to allow an additional flag to be passed (and the capability for the sidebar to know it's supported).

Nancyhu2023:
I have tried to test the Sidebar demo, and found the color of the FLUTTER SIDEBAR stays in light mode while my VScode is in night mode. refer to the image attached below. Not sure if is my local VScode setup problem or the demo problem. @anderdobo @kenzieschmoll @DanTup 
[...]

PR #6688, opened 2023-11-08
Add the ability to enable platform types from the device list in the sidebar

DanTup:
For each unique platform type in the set of unsupported devices, this adds a new row ("Enable web for this project"). Clicking it triggers the same flow as VS Code (which currently involves a toast in the corner to confirm we can run `flutter create`).

In future we might want to change this prompt to be in the sidebar (so it's less likely you'll miss it).

Related PR for Dart-Code is at https://github.com/Dart-Code/Dart-Code/pull/4831

See https://github.com/Dart-Code/Dart-Code/issues/4820
See https://github.com/flutter/devtools/issues/6563

https://github.com/flutter/devtools/assets/1078012/aaeb254f-a227-4038-8503-0edc7f823f9f

### DIFF
--- a/packages/devtools_app/lib/src/shared/analytics/constants/_vs_code_sidebar_constants.dart
+++ b/packages/devtools_app/lib/src/shared/analytics/constants/_vs_code_sidebar_constants.dart
@@ -10,7 +10,11 @@ enum VsCodeFlutterSidebar {
 
   /// Analytics event that is sent when a device selection occurs from the list
   /// of available devices in the sidebar.
-  changeSelectedDevice;
+  changeSelectedDevice,
+
+  /// Analytics event that is sent when a platform type is selected to be enable
+  /// from the list of devices in the sidebar.
+  enablePlatformType;
 
   static String get id => VsCodeFlutterSidebar.vsCodeFlutterSidebar.name;
 
--- a/packages/devtools_app/lib/src/shared/analytics/constants/_vs_code_sidebar_constants.dart
+++ b/packages/devtools_app/lib/src/shared/analytics/constants/_vs_code_sidebar_constants.dart
@@ -10,13 +10,14 @@ enum VsCodeFlutterSidebar {
 
   /// Analytics event that is sent when a device selection occurs from the list
   /// of available devices in the sidebar.
-  changeSelectedDevice,
-
-  /// Analytics event that is sent when a platform type is selected to be enable
-  /// from the list of devices in the sidebar.
-  enablePlatformType;
+  changeSelectedDevice;
 
   static String get id => VsCodeFlutterSidebar.vsCodeFlutterSidebar.name;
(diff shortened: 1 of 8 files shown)


packages/devtools_app/lib/src/shared/analytics/constants/_vs_code_sidebar_constants.dart
```diff
@@ -10,7 +10,11 @@ enum VsCodeFlutterSidebar {
 
   /// Analytics event that is sent when a device selection occurs from the list
   /// of available devices in the sidebar.
-  changeSelectedDevice;
+  changeSelectedDevice,
+
+  /// Analytics event that is sent when a platform type is selected to be enable
+  /// from the list of devices in the sidebar.
+  enablePlatformType;
```
kenzieschmoll, line 17: should we record the platform type that is enabled here?
`String enablePlatformType(String platformType) => 'enablePlatformType-$platformType';`

packages/devtools_app/lib/src/standalone_ui/vs_code/devices.dart
```diff
@@ -12,15 +12,21 @@ import '../../shared/analytics/constants.dart' as gac;
...
     super.key,
-  });
+  }) : unsupportedDevicePlatformTypes = unsupportedDevices
+            .map((device) => device.platformType)
+            .nonNulls
+            .toSet();
 
   final VsCodeApi api;
   final List<VsCodeDevice> devices;
+  final List<VsCodeDevice> unsupportedDevices;
```
kenzieschmoll, line 28: nit: what about a single list of devices with a boolean field 'supported' so that we don't have to maintain two lists?

DanTup: It'll be slightly more complicated because we'd need DevTools to tell VS Code that it knows to check the flag (otherwise, new VS Code would include unsupported devices to old DevTools, and users would see "broken" devices in the list because the old Sidebar didn't filter them out).

We don't currently have capabilities in this direction (for VS Code to know what DevTools supports), but adding support for it might not be a bad idea (we're probably going to want it in future for other things).

(we could also base it on an SDK version number, but that's a little more awkward because we'd have to land the DevTools changes before we have a version number to code into VS Code)

kenzieschmoll: Gotcha. Keeping two lists sgtm then.

packages/devtools_app/lib/src/shared/analytics/constants/_vs_code_sidebar_constants.dart
DanTup: Done!

kenzieschmoll (approved):

PR #6693, opened 2023-11-09
Fix `tool/flutter_customer_tests/setup.sh` script

elliette:
* We need to run pub-get first before gnerating code 
* We should be calling the `.dart` scripts explicitly 

Trying to unblock https://github.com/flutter/tests/pull/310


kenzieschmoll (approved):

kenzieschmoll:
If you merge master the CI errors should go away

PR #6695, opened 2023-11-09
Add `build-mode` and `pub-get` argument to the `serve` and `build` CLI tools

kenzieschmoll:
This changes the `build-release` command to `build` and adds arguments:
- `build-mode`: this can be one of 'debug', 'profile', or 'release' (defaults to 'release')
- `pub-get`: this flag can be negated (`--no-pub-get`) to prevent `devtools_tool pub-get --only-main` from being run before building DevTools. This is helpful for faster development cycles when building DevTools with the DevTools server.

This PR also:
- de-duplicates code between the `BuildCommand` and `ServeCommand` classes. Since the majority of the arguments added to the `ServeCommand` arg parser are just being passed along to the `devtools_tool build` command.
- fixes a bug in the `update-flutter-sdk` script


tool/lib/commands/shared.dart
```diff
@@ -0,0 +1,70 @@
...
+    );
+  }
+
+  void addUseLocalFlutterFlag() {
+    addFlag(
+      BuildCommandArgs.useLocalFlutter.flagName,
+      negatable: false,
+      defaultsTo: false,
+      help: 'Whether to use the Flutter SDK on PATH instead of the Flutter SDK '
+          'contained in the "tool/flutter-sdk" directory.',
```
DanTup, line 50: I didn't realise we had this flag elsewhere, but the name seems confusing to me. The word "local" would make me think it uses the version local to DevTools (`tools/flutter-sdk`) though this text suggests it's the other way around. Is it worth making more explicit (`--use-flutter-from-path` or `--use-flutter-from-devtools` or something) with an alias for the old name?

(it's possible we can remove the change I added to support FLUTTER_ROOT recently if we have this - I added it specifically to force using the version from `tool/flutter-sdk` because I was getting the one from `PATH` that I didn't want).

tool/lib/commands/update_flutter_sdk.dart
```diff
@@ -133,35 +134,28 @@ class UpdateFlutterSdkCommand extends Command {
...
-        workingDirectory: repo.toolDirectoryPath,
-      );
-      await processManager.runAll(
-        commands: [
-          CliCommand.git('checkout $flutterTag -f'),
-          CliCommand.flutter('--version'),
-        ],
-        workingDirectory: toolFlutterSdk.path,
-      );
+      toolFlutterSdk.deleteSync(recursive: true);
```
DanTup, line 141: What's the reason for deleting and re-cloning this? (seems like that would be way slower than the original code that just pulled latest commits)

tool/lib/commands/update_flutter_sdk.dart
```diff
@@ -133,35 +134,28 @@ class UpdateFlutterSdkCommand extends Command {
       repo.toolDirectoryPath,
       flutterSdkDirName,
     );
-    final toolFlutterSdk = Directory.fromUri(Uri.file(toolSdkPath));
+    final toolFlutterSdk = Directory(toolSdkPath);
```
DanTup, line 137: nit: `toolFlutterSdkDirectory`? (since we have a `FlutterSdk` class and this isn't an instance of that)

(or maybe just `toolSdkDirectory` to be consistent with `toolSdkPath`)

DanTup (approved):
lgtm!

tool/lib/commands/update_flutter_sdk.dart
kenzieschmoll: I kept getting process exceptions on the series of `git` calls, so figured deleting and re-cloning would be a robust way to make sure we get the branch we need and aren't getting in a bad state from git.

tool/lib/commands/shared.dart
kenzieschmoll: changed to --use-flutter-from-path to make this more clear.

tool/lib/commands/update_flutter_sdk.dart
kenzieschmoll: Reverted back to what we had. If I see the errors again we can dig back into this issue. maybe I got myself in a weird state by cancelling the script in the middle of execution.

tool/lib/commands/shared.dart
DanTup: Cool, thanks! I'll make a note to look back at the FLUTTER_ROOT change I made.. I should be able to remove it if we're generally defaulting to tool/flutter-sdk by default. If there are places that's not happening, I'll fix them up.

PR #6696, opened 2023-11-09
Update contributing guide to point to `devtools_tool serve`

kenzieschmoll:
@DanTup can you review for accuracy? Also, the `tool/flutter-sdk` directory isn't necessarily guaranteed to be up to date. Does the dart-code plugin run `devtools_tool update-flutter-sdk` before running `devtools_tool serve`?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -154,7 +154,7 @@ dependency_overrides:
 
 Then you can run DevTools with the server by running the following from the top-level `devtools` directory:
 ```
-dart ./tool/build_e2e.dart
+devtools_tool serve
 ```
 
 ### DevTools + VS Code integration (IDE-embedded DevTools experience)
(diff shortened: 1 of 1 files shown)


DanTup (approved):
Didn't see this before I replied on the other issue.

This LGTM, though I'm gonna make some more tweaks to this after your other PR lands (because I'd like to be able to control the flags like `--no-pub-get` from this config, so I can make it faster if I know it's up-to-date, but without removing the ability for it to run the whole thing).

> Does the dart-code plugin run `devtools_tool update-flutter-sdk` before running `devtools_tool serve`?

It does not, it _only_ runs `devtools serve`. I think we should control anything like that with flags to `serve` and then users can easily enable/disable them from the `dart.customDevTools` config based on their needs (this will require a tweak in Dart-Code).

kenzieschmoll:
The tweaks you want to make sgtm. My other PR should land shortly: https://github.com/flutter/devtools/pull/6695.

PR #6700, opened 2023-11-09
Fix a bug where we were unnecessarily calling `manualDisconnect`

kenzieschmoll:
Found an issue while investigating https://github.com/Dart-Code/Dart-Code/issues/4832#issuecomment-1804186076. We do not need to perform a disconnect if there is not already an existing connection.

### DIFF
--- a/packages/devtools_app/lib/src/shared/routing.dart
+++ b/packages/devtools_app/lib/src/shared/routing.dart
@@ -60,7 +60,9 @@ class DevToolsRouteInformationParser
       // If the uri has been modified and we do not have a vm service uri as a
       // query parameter, ensure we manually disconnect from any previously
       // connected applications.
-      await serviceConnection.serviceManager.manuallyDisconnect();
+      if (serviceConnection.serviceManager.hasConnection) {
+        await serviceConnection.serviceManager.manuallyDisconnect();
+      }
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/shared/routing.dart
```diff
@@ -60,7 +60,9 @@ class DevToolsRouteInformationParser
       // If the uri has been modified and we do not have a vm service uri as a
       // query parameter, ensure we manually disconnect from any previously
       // connected applications.
-      await serviceConnection.serviceManager.manuallyDisconnect();
+      if (serviceConnection.serviceManager.hasConnection) {
+        await serviceConnection.serviceManager.manuallyDisconnect();
+      }
```
DanTup, line 65: I'm not too familiar with this, but is it possible we'd have a race here if there is a connection in progress (`hasConnection` being false so we skip this, but a connection ends up being formed shortly after?)

(if it is possible, I don't know if calling `manuallyDisconnect()` is the right thing to do to cancel that anyway, so it might be an unrelated issue if there is such an issue)

DanTup (approved):

kenzieschmoll:
closing this for now since https://github.com/Dart-Code/Dart-Code/issues/4832 has been resolved. If this is causing other issues we can investigate then.

PR #6713, opened 2023-11-11
Refactor server APIs by logical groups

kenzieschmoll:
This also removes the server stub for desktop since we can now use a single server implementation for both platforms

All the new part files were copy & pasted from server.dart


bkonyi (approved):
LGTM, but is there any particular reason that you're using part files instead of just importing the individual components of the server API where needed?

kenzieschmoll:
> LGTM, but is there any particular reason that you're using part files instead of just importing the individual components of the server API where needed?

We import the main server file everywhere with the prefix ` as server;` and then call APIs with `server.xyz`. So using part files allows us to still use the same syntax everywhere to interact with the DevTools server.

PR #6724, opened 2023-11-14
Don't render not-found page during initial setup

DanTup:
The router is always called with a `null` page when the app first loads. Previously this would render a not-found page and because there are also no args, it would be marked as not-embedded and would register with the server as a page that can be reused which (partly) caused https://github.com/Dart-Code/Dart-Code/issues/4832.

In my testing, this change does not seem to impact any behaviour of loading the sidebar, homepage or directly accessing other pages.

I added a TODO about standalone screens not transmitting their pages. This is not causing any problems, but it feels a little odd. We could:

a) ignore this, it doesn't affect much since they're always embedded so the server will never try to reuse them
b) add a wrapper around standalone screens that just sends `currentPage` inside its `initState`
c) split `DevToolsScaffold` up into the parts that are only used by non-embedded pages (the nav, etc.) and the parts that should be used by everything (initialisation of things like `currentPage`)

### DIFF
--- a/packages/devtools_app/lib/src/app.dart
+++ b/packages/devtools_app/lib/src/app.dart
@@ -204,13 +204,15 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
     Map<String, String?> args,
     DevToolsNavigationState? state,
   ) {
-    if (FrameworkCore.initializationInProgress) {
+    // `page` will initially be null while the router is set up, then we will
+    // be called again with an empty string for the root.
+    if (FrameworkCore.initializationInProgress || page == null) {
       return const MaterialPage(child: CenteredCircularProgressIndicator());
     }
 
     // Provide the appropriate page route.
     if (pages.containsKey(page)) {
-      Widget widget = pages[page!]!(
+      Widget widget = pages[page]!(
         context,
         page,
         args,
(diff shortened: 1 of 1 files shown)


DanTup:
Re-running tests to see if this is a flake - I can't reproduce it locally (Windows) and don't think it's related:

```
FlutterDriveProcess - Failure Details:
FlutterDriveProcess - Failure in method: Debugger panel
FlutterDriveProcess - ══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞═════════════════
FlutterDriveProcess - The following TestFailure was thrown running a test:
FlutterDriveProcess - Expected: no matching candidates
FlutterDriveProcess -   Actual: _TypeWidgetFinder:<Found 1 widget with type
Unhandled exception:
FlutterDriveProcess - "ConnectDialog": [
FlutterDriveProcess -             ConnectDialog(dependencies: [_InheritedTheme,
Exception: Test 'Debugger panel' failed: ══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞═════════════════
FlutterDriveProcess - _LocalizationsScope-[GlobalKey#667a2], _RouterScope], state:
The following TestFailure was thrown running a test:
FlutterDriveProcess - _ConnectDialogState#6774a),
Expected: no matching candidates
FlutterDriveProcess -           ]>
  Actual: _TypeWidgetFinder:<Found 1 widget with type
FlutterDriveProcess -    Which: means one was found but none were expected
"ConnectDialog": [
FlutterDriveProcess - 
            ConnectDialog(dependencies: [_InheritedTheme,
_LocalizationsScope-[GlobalKey#667a2], _RouterScope], state:
FlutterDriveProcess - When the exception was thrown, this was the stack:
_ConnectDialogState#6774a),
FlutterDriveProcess - dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 297:3        throw_
          ]>
FlutterDriveProcess - packages/matcher/src/expect/prints_matcher.dart.js 436:22                          fail
```

kenzieschmoll (approved):

PR #6728, opened 2023-11-14
[deep link] add isRowHovered in column render

hangyujin:
The last column in deep link table is a navigation column, when the row is hovered, the column content is an icon, otherwise it's empty.
  
WeChat23c7fe6bf78ef1ede95094d70f84523c

*List which issues are fixed by this PR.*

*Please add a note to `packages/devtools_app/release_notes/NEXT_RELEASE_NOTES.md` if your change requires release notes. Otherwise, add the 'release-notes-not-required' label to the PR.*

## Pre-launch Checklist

- [ ] I signed the [CLA].

![build.yaml badge]

### DIFF
--- a/packages/devtools_app/lib/src/shared/table/table.dart
+++ b/packages/devtools_app/lib/src/shared/table/table.dart
@@ -1149,6 +1149,7 @@ abstract class ColumnRenderer<T> {
     BuildContext context,
     T data, {
     bool isRowSelected = false,
+    bool isRowHovered = false,
     VoidCallback? onPressed,
   });
 }
@@ -1335,6 +1336,8 @@ class _TableRowState<T> extends State<TableRow<T>>
 
   bool isActiveSearchMatch = false;
 
+  bool isHovering = false;
+
   @override
   void initState() {
     super.initState();
@@ -1497,6 +1500,7 @@ class _TableRowState<T> extends State<TableRow<T>>
             context,
             node,
             isRowSelected: widget.isSelected,
+            isRowHovered: isHovered,
             onPressed: onPressed,
           );
         }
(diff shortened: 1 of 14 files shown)


packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -1640,30 +1644,34 @@ class _TableRowState<T> extends State<TableRow<T>>
...
-                width: columnSpacing,
-                child: VerticalDivider(width: columnSpacing),
-              );
-            case _TableRowPartDisplayType.columnGroupSpacer:
-              return const _ColumnGroupSpacer();
-          }
-        },
+    final rowContent = MouseRegion(
+      onEnter: (_) => setState(() => isHovering = true),
+      onExit: (_) => setState(() => isHovering = false),
```
kenzieschmoll, line 1649: adding a MouseRegion for every row may be expensive. We have hit performance issues in the past by doing stuff like this (anything that adds a bunch of listeners). Instead of adding a MouseRegion for each row, add a single MouseRegion for the _Table widget and then calculate the hovered rowed based on the event's dy value.

Example: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/shared/charts/flame_chart.dart/#L283

kenzieschmoll: Or another option is to make hover handling opt-in for the table widget, so that this `MouseRegion` is only added for the deep links table. At least then we can limit the performance impact and address this if it becomes a performance concern for the deep links table.

hangyujin: changed it to opt-in

packages/devtools_app/lib/src/shared/table/table.dart
```diff
@@ -1195,6 +1207,7 @@ abstract class ColumnRenderer<T> {
     BuildContext context,
     T data, {
     bool isRowSelected = false,
+    bool isRowHovered = false,
```
kenzieschmoll, line 1210: add a note to this dart doc that states that `isRowHovered` is only used when `enableHoverHandling` is true on the table that this column belongs to.

kenzieschmoll (approved):
one dartdoc request then lgtm

PR #6730, opened 2023-11-14
Set version constraints for pkg:web >=0.3.0 <0.5.0

srujzs:
package:web 0.4.0 restricts types to not be subtypes of JSObject until we migrate those types to extension types. Flutter pins package:web to 0.3.0, so packages targeting 3.2 should support this version range for now.


kenzieschmoll:
Does this PR mean we can use helpers like `createIFrameElement` and address the final package:web migration todo here? https://github.com/flutter/devtools/issues/6606#issuecomment-1809080582

srujzs:
Maybe. It really depends on whether we can move this straight to `^0.4.0` or if we need the version range. If the former, then yes, and if the latter, then not yet.

kenzieschmoll (approved):

Issue #6492, opened 2023-09-29
Service extension integration test flake


CoderDake:
Flake encountered in https://github.com/flutter/devtools/pull/6490

```
2023-09-28T15:14:28.2586330Z FlutterDriveProcess - Got dependencies!
2023-09-28T15:14:28.2587400Z FlutterDriveProcess - 21 packages have newer versions incompatible with dependency constraints.
2023-09-28T15:14:28.2588500Z FlutterDriveProcess - Try `flutter pub outdated` for more information.
2023-09-28T15:14:29.0733960Z FlutterDriveProcess - Launching integration_test/test/live_connection/service_connection_test.dart on Web Server in debug mode...
2023-09-28T15:15:31.2432190Z FlutterDriveProcess - Waiting for connection from debug service on Web Server...         61.6s
2023-09-28T15:15:31.4831880Z FlutterDriveProcess - integration_test/test/live_connection/service_connection_test.dart is being served at http://localhost:49431
2023-09-28T15:15:31.4833420Z FlutterDriveProcess - The web-server device requires the Dart Debug Chrome extension for debugging. Consider using the Chrome or Edge devices for an improved development workflow.
[...]

PR #6736, opened 2023-11-15
Add support for "Open DevTools in Browser" in Sidebar menu

DanTup:
This adds a new entry to the DevTools menu in the sidebar to "Open in Browser" (matching the extra option for this when you run the "Open DevTools" command from the VS Code command palette):

image

This will open whatever the default page is in DevTools for this type of app, and will always open in a browser window regardless of the user-settings for embedded/external.

See https://github.com/flutter/devtools/issues/6563


bkonyi (approved):

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -139,19 +142,13 @@ class _DevToolsMenu extends StatelessWidget {
...
-      } else if (screen.requiresDartVm && isWeb) {
-        disabledReason = 'Not available when running on the web';
-      }
-
+    Widget devToolsButton(
+      String title, {
+      IconData? icon,
+      String? screenId,
+      String? disabledReason,
+      bool? forceExternal,
```
kenzieschmoll, line 150: give this a default value of 'false' so that we can write `screenId != null && !forceExternal` below. I think that would make the ternary check below a bit more readable.

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -139,19 +142,13 @@ class _DevToolsMenu extends StatelessWidget {
...
-        disabledReason = 'Not available in release mode';
-      } else if (screen.requiresFlutter && !isFlutter) {
-        disabledReason = 'Only available for Flutter applications';
-      } else if (screen.requiresDebugBuild && !isDebug) {
-        disabledReason = 'Only available in debug mode';
-      } else if (screen.requiresDartVm && isWeb) {
-        disabledReason = 'Not available when running on the web';
-      }
-
+    Widget devToolsButton(
```
kenzieschmoll, line 145: instead of using a helper method, can you create a helper widget like in my draft PR here: https://github.com/flutter/devtools/pull/6709/files#diff-e976c2386d5c7dc8f65bcad658ee5fba12faf0c519c2ad34664c78465fd9129fR275-R306

kenzieschmoll (approved):
couple comments but lgtm

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
DanTup: Done!

DanTup: (Actually this condition went away when I extracted a widget because they pass their own `onPressed` now)

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
DanTup: Done! (unfortunately you're going to have some conflicts here)

DanTup:
I made the requested changes - if there's any feedback on those changes, I'll address in another PR (I figure it's good to land this before our two conflicting PRs diverge further).

DanTup:
Failure is https://github.com/flutter/devtools/issues/6492, re-running...

PR #6742, opened 2023-11-15
Create setup.sh script for the CI

kenzieschmoll:


tool/ci/setup.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
+
+# Copyright 2018 The Chromium Authors. All rights reserved.
```
CoderDake, line 3:
```suggestion
# Copyright 2023 The Chromium Authors. All rights reserved.
```
not sure what we are supposed to do with new files but should this be 2023?

CoderDake (approved):

PR #6750, opened 2023-11-16
Pin package:web 0.4.0

srujzs:
Now that Flutter main has moved to ^0.4.0, we can pin this version.

### DIFF
--- a/packages/devtools_app/lib/src/shared/config_specific/post_message/_post_message_web.dart
+++ b/packages/devtools_app/lib/src/shared/config_specific/post_message/_post_message_web.dart
@@ -2,11 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be found
 // in the LICENSE file.
 
-// TODO(https://github.com/flutter/devtools/issues/6606): remove this import.
-// This is the final dart:html import in DevTools. In order to remove it, we
-// need to bump the `package:web` version in DevTools to > 0.3.1, but we are
-// blocked on `package:web` rolling into the Flutter SDK.
-import 'dart:html' as html;
 import 'dart:js_interop';
 
 import 'package:web/helpers.dart';
@@ -14,7 +9,7 @@ import 'package:web/helpers.dart';
 import 'post_message.dart';
 
 Stream<PostMessageEvent> get onPostMessage {
-  return html.window.onMessage.map(
+  return window.onMessage.map(
     (message) => PostMessageEvent(
       origin: message.origin,
       data: message.data,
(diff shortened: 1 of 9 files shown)


srujzs:
Hmm, maybe we need to update the Flutter version constraint in the pubspecs?

kevmoo:
Do you want to follow-up with https://github.com/flutter/devtools/pull/6753 ?

kevmoo:
> Hmm, maybe we need to update the Flutter version constraint in the pubspecs?

That shouldn't matter. The version of Flutter that's tested here just needs to roll forward to the version where you updated the pin dep

srujzs:
Ah yes, we should remove the last imports of `dart:html` in this PR too. I'll incorporate your changes into here.

> The version of Flutter that's tested here just needs to roll forward to the version where you updated the pin dep.

I thought the Flutter environment constraint *was* the version of Flutter that's being tested. Is that defined elsewhere?

flutter-candidate.txt
```diff
@@ -1 +1 @@
-3.17.0-9.0.pre
```
kenzieschmoll, line 1: we shouldn't update this file manually - it is automatically updated by a github workflow when a new candidate branch is ready. The workflow runs 2x per day.

packages/devtools_app/lib/src/shared/config_specific/post_message/_post_message_web.dart
```diff
@@ -2,14 +2,9 @@
...
 
-// TODO(https://github.com/flutter/devtools/issues/6606): remove this import.
-// This is the final dart:html import in DevTools. In order to remove it, we
-// need to bump the `package:web` version in DevTools to > 0.3.1, but we are
-// blocked on `package:web` rolling into the Flutter SDK.
-import 'dart:html' as html;
 import 'dart:js_interop';
 
-import 'package:web/helpers.dart';
+import 'package:web/helpers.dart' as html;
```
kenzieschmoll, line 7: why prefix this with html?

srujzs: I think it's just keeping consistent with the old syntax, but since we don't need `dart:html` anymore, there are no conflicts, and we can just remove the prefix.

flutter-candidate.txt
srujzs: Got it, let's wait until it automatically updates to the needed version then.

kevmoo (approved):
LGTM

PR #6753, opened 2023-11-16
WIP

kevmoo:


packages/devtools_app/lib/src/shared/config_specific/notifications/_notifications_web.dart
```diff
@@ -17,9 +17,8 @@ class Notification {
 
   late final web_helpers.Notification _impl;
 
-  static Future<String> requestPermission() {
-    return promiseToFuture(web_helpers.Notification.requestPermission());
-  }
+  static Future<String> requestPermission() async =>
+      (await web_helpers.Notification.requestPermission().toDart) as String;
```
srujzs, line 21: Note that this will fail on dart2wasm since it isn't going through a conversion: `toJS`. There's a lint proposal on catching such instances, but we want some lint telling users here that `JSString as String` is going to fail on dart2wasm.

kevmoo: So this will be `Future<JsString>` – then we need to to a toDart on it?

kevmoo: see the last commit

srujzs: Your change will work. The other alternative is to do a `.then` and cast.

Once I land the change to move `dart:js_interop` to extension types, I'm going to add generic types to `JSArray` and `JSPromise`, and then we can update `package:web` to use generic `JSPromise`s and `JSArray`s, so you'll just need to do a `web_helpers.Notification.requestPermission().toDart` in the future to make this work.

kevmoo:
subsumed by https://github.com/flutter/devtools/pull/6750

Issue #6751, opened 2023-11-16
Enhance DevTools survey logic


kenzieschmoll:
- [x] Include the DevTools version in the survey url query parameters
- [x] Add the ability to set a minimum DevTools version in the survey json file. Do not show the survey prompt if the current version of DevTools is below the min version.
- [x] Remove the "from" query parameter - it is not useful
- [x] Add a way to specify a platform so that we can perform more targeted surveys (VS Code, Android Studio, CLI, etc.)

Other feedback is to show page screenshots in the survey questions so users know which pages we are referring to (this would help avoid issues like the user thinking that we are talking about their IDE's debugger instead of the DevTools debugger)

PR #6754, opened 2023-11-16
Enhance DevTools survey logic

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/6751

- [x]  Include the DevTools version in the survey url query parameters
- [x]  Add the ability to set a minimum DevTools version in the survey json file. Do not show the survey prompt if the current version of DevTools is below the min version.
- [x]  Remove the "from" query parameter - it is not useful
- [x]  Add a way to specify a platform so that we can perform more targeted surveys (VS Code, Android Studio, CLI, etc.)

### DIFF
--- a/packages/devtools_app/lib/src/shared/analytics/_analytics_stub.dart
+++ b/packages/devtools_app/lib/src/shared/analytics/_analytics_stub.dart
@@ -15,6 +15,8 @@ import 'analytics_common.dart';
 
 final _log = Logger('_analytics_stub');
 
+String ideLaunched = '';
+
 Future<void> setAnalyticsEnabled(bool value) async {}
 
 FutureOr<bool> isAnalyticsEnabled() => false;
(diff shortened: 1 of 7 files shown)


packages/devtools_app/lib/src/shared/analytics/_analytics_stub.dart
```diff
@@ -15,6 +15,8 @@ import 'analytics_common.dart';
 
 final _log = Logger('_analytics_stub');
 
+String ideLaunched = '';
```
polina-c, line 18: Doc comment?

packages/devtools_app/lib/src/shared/survey.dart
```diff
@@ -177,5 +200,47 @@ class DevToolsSurvey {
...
+
+  /// A list of development environments to show the survey for (e.g. 'VSCode',
+  /// 'Android-Studio', 'IntelliJ-IDEA', 'CLI', etc.).
+  ///
+  /// If [environments] is null, the survey can be shown to any platform.
+  ///
+  /// The possible values for this list correspond to the possible values of
+  /// [_ideLaunched] from [shared/analytics/_analytics_web.dart].
+  final List<String>? environments;
+}
```
polina-c, line 224: environments is too generic. How about devEnvironments? Or ides?

polina-c (approved):
lgtm with coule comments

Issue #6767, opened 2023-11-17
Improve messaging when we do not have timeline data available for a frame


kenzieschmoll:
Right now we just say "No analysis data available for this frame" but we need to explain why so that users understand that our tool is not broken, but that they instead need to open the tool sooner so that we can collect the timeline events we need to create the analysis data.

PR #6768, opened 2023-11-17
Improve messaging when we don't have analysis data available for a frame

kenzieschmoll:
Screenshot 2023-11-17 at 12 07 17 PM

Fixes https://github.com/flutter/devtools/issues/6767


polina-c:
'earlier' sounds better then 'sooner'

polina-c (approved):
lgtm with one comment

Issue #4467, opened 2022-09-12
Show whether app is running with skia or impeller in the DevTools performance page


kenzieschmoll:
service extension added here: https://github.com/flutter/engine/pull/36103

PR #6771, opened 2023-11-17
Show whether the app is using Impeller or Skia in the performance page

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/4467

Currently this only shows one of two values: 'Impeller' or 'Skia'. @jonahwilliams is this accurate?

flutter_frames_engine_text

I am open to suggestions for a better spot to put this data.

This PR also:
- removes the Shader Compilation entry from the frames chart legend when impeller is enabled, since shader compilation is not relevant for impeller


packages/devtools_app/lib/src/screens/performance/panes/flutter_frames/flutter_frames_chart.dart
```diff
@@ -332,6 +343,14 @@ class _FramesChartState extends State<FramesChart> with AutoDisposeMixin {
           child: chart,
         ),
         fpsLinePainter,
+        Positioned(
+          right: denseSpacing,
+          top: densePadding,
+          child: Text(
+            'Engine: ${widget.impellerEnabled ? 'Impeler' : 'Skia'}',
```
jonahwilliams, line 350:
```suggestion
            'Engine: ${widget.impellerEnabled ? 'Impeller' : 'Skia'}',
```

jonahwilliams (approved):
LGTM with nit

PR #6776, opened 2023-11-20
Always use tool/flutter-sdk for devtools_tool + remote FLUTTER_ROOT

DanTup:
This removes some code I added to support using `FLUTTER_ROOT` for the SDK and instead just updates the `devtools_tool` shell scripts to always run from `tools/flutter-sdk` and just locate the Flutter SDK from the running VM.

This makes the assumption that we _always_ want to use the DevTools version of Flutter and avoids any issues where you might have something different on `PATH`.

It also renames the `--use-flutter-from-path` flag to `--update-flutter`/`--no-update-flutter` (and retains the default of updating).


tool/lib/commands/shared.dart
```diff
@@ -41,12 +41,12 @@ extension BuildCommandArgsExtension on ArgParser {
...
+  void addUpdateFlutterFlag() {
     addFlag(
-      BuildCommandArgs.useFlutterFromPath.flagName,
-      negatable: false,
-      defaultsTo: false,
-      help: 'Whether to use the Flutter SDK on PATH instead of the Flutter SDK '
+      BuildCommandArgs.updateFlutter.flagName,
+      negatable: true,
+      defaultsTo: true,
+      help: 'Whether to update the Flutter SDK before building DevTools '
```
kenzieschmoll, line 49: nit: remove "before building DevTools" because this flag can be used for more than just the build command.

After removing, it should say 'Whether to update the Flutter SDK contained in the "tool/flutter-sdk" directory'

kenzieschmoll (approved):
one comment but lgtm

tool/lib/commands/shared.dart
DanTup: Done!

+ another minor fix I forgot to push earlier, which ensures if you pass `--no-update-flutter` we remove it before passing on to the local server.

PR #6782, opened 2023-11-20
Increase the pump duration for app connection in integration test

kenzieschmoll:
Attempt to fix test flake by increasing the settle duration. We sometimes get the exception that the state after we click "connect" is not what we expect. Increasing the settle duration may reduce this flake.

### DIFF
--- a/packages/devtools_app/lib/src/shared/config_specific/import_export/_export_web.dart
+++ b/packages/devtools_app/lib/src/shared/config_specific/import_export/_export_web.dart
@@ -2,9 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-// TODO(kenz): remove this once the package:web version has been upgraded.
-// ignore_for_file: avoid-unused-ignores
-
 import 'dart:js_interop';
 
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/shared/config_specific/import_export/_export_web.dart
```diff
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+// ignore_for_file: avoid-unused-ignores, package:web upgrade is in progress.
+
```
kenzieschmoll, line 6: @srujzs FYI the DCM check started failing on the `avoid-unnecessary-type-casts` in this file. I ignored the
`avoid-unused-ignores` check instead of trying to fix the type casts since you are actively working on upgrading our package:web dep

CoderDake (approved):

packages/devtools_app/lib/src/shared/config_specific/import_export/_export_web.dart
srujzs: I've landed the pin to 0.4.0, so we don't need this anymore!

PR #6783, opened 2023-11-20
Fix some minor Windows issues

DanTup:
`Uri.parse()` shouldn't be used with file paths for Windows (should use `Uri.file()`) + tweak some tests to always use posix paths in strings that are checked so that they're consistent across platforms.


kenzieschmoll (commented):
can you add a changelog entry for devtools_shared and bump the pubspec / changelog version to `6.0.2-wip`

DanTup:
Done!

kenzieschmoll (approved):

PR #6820, opened 2023-11-22
Revert "Always use tool/flutter-sdk for devtools_tool + remote FLUTTE…

zanderso:
…R_ROOT (#6776)"

This reverts commit 20f3dc671e7658e2b61f2970aed5e722c12c95c1.

Reason for the revert: Rolls to flutter/tests are failing. See discussion in https://github.com/flutter/tests/pull/314.

PR #6824, opened 2023-11-23
Re-land SDK detection changes with support for running from PATH

DanTup:
The first commit (f21229bcd) here is a straight re-land of https://github.com/flutter/devtools/pull/6776 (a revert of the revert at https://github.com/flutter/devtools/pull/6820 / 79cde66).

The second commit (a5103ad61) adds an env variable you can set to force devtools_tool to use Dart/Flutter from `PATH` rather than the one from tool/flutter-sdk, and sets it in the flutter_customer_tests script.

I made this explicit so it's difficult to do accidentally, because I think outside of the Flutter customer tests, we do generally want everything to use the pinned version.

@kenzieschmoll

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -32,7 +32,7 @@ jobs:
         uses: actions/cache@704facf57e6136b1bc63b828d79edcd491f0ee84
         with:
           path: |
-            ./flutter-sdk
+            ./tool/flutter-sdk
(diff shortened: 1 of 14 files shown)


kenzieschmoll (approved):

DanTup:
I just realised the green status here was probably stale from before master broke. Let me know if I should revert this while you're fixing the build or not.

Issue #6841, opened 2023-11-28
Widget Inspector shows too many widgets


johnpryan:
When I open the Widget Inspector for the counter app, I get a simplified version of the widget tree, but when I open it for DartPad  ([pkgs/sketch_pad](https://github.com/dart-lang/dart-pad/tree/main/pkgs/sketch_pad)), I see a very complex widget tree, with widgets like View and MediaQuery, that aren't part of my application code.

**Counter app:**

![Screenshot 2023-11-28 at 10 01 05 AM](https://github.com/flutter/devtools/assets/1145719/887ebad1-150f-404b-83c2-17c6ebd752af)

**DartPad:**

![Screenshot 2023-11-28 at 10 05 04 AM](https://github.com/flutter/devtools/assets/1145719/88728157-12d9-4be6-81ce-961713d45078)

kenzieschmoll:
Can you check your Package Directories in the Inspector settings and post a screenshot of what is there?

kenzieschmoll:
Delete the flutter SDK from the list of package directories. I'm not sure how this was added to the list. @CoderDake could there be a bug in our package directory detection logic?

kenzieschmoll:
> @jacob314 and I were just discussing something similar where an internal customer's widget tree is comprised of primarily Provider widgets, making it quite user-unfriendly.

Were these Provider widgets ones that they defined in their app? Or were they widgets from the internals of package:provider?

johnpryan:
> Delete the flutter SDK from the list of package directories. I'm not sure how this was added to the list. @CoderDake could there be a bug in our package directory detection logic?

I'm not sure either, but that fixed it for me.

> @jacob314 and I were just discussing something similar where an internal customer's widget tree is comprised of primarily Provider widgets, making it quite user-unfriendly.

I'm seeing the correct number of Provider widgets I would expect in DartPad, so I'm not sure if this is related or not.

elliette:
Summarizing what @jacob314 just told me in the office:

In `inspector_service.dart`, we make the assumption that the first widget in your widget tree is owned by your application. However, @goderbauer recently landed a change to add a `View` widget as the first widget in the tree. (I'm seeing the `View` widget both in DartPad's widget tree, and in the widget tree of an internal customer that is incorrectly getting the flutter package added to their project directories.) We will need to update our detection logic to non assume that the first widget in the tree is owned by the application itself.

@goderbauer did the `View` widget changes make it into this latest Flutter stable release? Thanks!

[...]

Issue #6853, opened 2023-11-29
Flutter inspector settings auto-detects the wrong project directory


elliette:
~~This is really two issues that an internal customer is hitting:~~

~~1. For large apps, auto-detection of the package directory fails, so there is an infinite spinner next to the package directory settings (because of this the widget tree is empty)
2. If you then try to add a package directory manually, that fails (I'm guessing because we are stuck trying to detect the package directory)~~

![package loading](https://github.com/flutter/devtools/assets/21270878/e1e79d0c-9eaa-4026-9d07-1cb728f22c16)

elliette:
Fixed in https://github.com/flutter/devtools/pull/6857

PR #6857, opened 2023-11-30
Fix project directory inference for inspector panel

elliette:
Fixes https://github.com/flutter/devtools/issues/6841, https://github.com/flutter/devtools/issues/6853

Instead of using the top level widget in the widget tree to determine the project directory, we instead use the main isolate's root library. 

For google3, we also have extra logic to strip out everything before `/google3`, and to only use top-level directory after `google3` directory (or the top-level directory after `third_party`)


packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
```diff
@@ -483,60 +484,62 @@ class InspectorService extends InspectorServiceBase {
...
     // package root directory given that the root script of this project is in
     // this directory rather than guessing based on url structure.
     final parts = path.split('/');
     String? pubRootDirectory;
-    for (int i = parts.length - 1; i >= 0; i--) {
-      final part = parts[i];
-      if (part == 'lib' || part == 'web') {
-        pubRootDirectory = parts.sublist(0, i).join('/');
-        break;
+    // For google3, we grab the top-level directory in the google3 directory
```
jacob314, line 499: I would move this logic into a separate helper method and add a couple unittests for google3 paths.
as the google3 logic is otherwise hard to test.

jacob314 (approved):
Thanks for cleaning up this bad tech debt that I created. Before submitting, lets add a unittests to make sure the logic for g3 paths does what we expect for a couple example cases.

jacob314:
lgtm

packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
```diff
@@ -483,60 +485,69 @@ class InspectorService extends InspectorServiceBase {
...
+  String? _pubRootDirectoryForGoogle3(List<String> pathParts) {
+    final strippedParts = _stripGoogle3(pathParts);
+    if (strippedParts.isEmpty) return null;
+
+    final topLevelDirectory = strippedParts.first;
+    if (topLevelDirectory == _thirdPartyPathSegment &&
+        strippedParts.length >= 2) {
+      return '/${strippedParts.sublist(0, 2).join('/')}';
+    } else {
+      return '/${strippedParts.first}';
```
kenzieschmoll, line 547: do we need to be concerned with how these paths will work with Windows at all? WRT forward slashes vs back slashes?

kenzieschmoll: I agree we should use package:path to be safe. I'll leave that up to you whether you want to do that here or file an issue to track this as a follow up. LGTM from me.

kenzieschmoll (approved):

packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
elliette: done!

packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
```diff
@@ -28,6 +28,8 @@ import 'primitives/instance_ref.dart';
 import 'primitives/source_location.dart';
 
 const inspectorLibraryUri = 'package:flutter/src/widgets/widget_inspector.dart';
+const google3PathSegment = 'google3';
```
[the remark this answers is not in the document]
elliette: done!

packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
elliette: SG, will do as follow up shortly (tracking in https://github.com/flutter/devtools/issues/6868)

PR #6861, opened 2023-11-30
Focus inspector when done typing

CoderDake:
Reverts https://github.com/flutter/devtools/pull/6677

This only brings back the "Focus when done typing behaviour" and removes the "enter and shift+enter" convenience functionality.

It turns out that there is some autocomplete behaviour that overlaps with this. 

I will open up a new issue to discuss modifying the behaviour of the autocomplete so that we can still have the convenience functions


CoderDake:
Opened up https://github.com/flutter/devtools/issues/6864 to address adding the enter and shift+enter behaviour back by modifying the autocomplete behaviour.

PR #6862, opened 2023-11-30
Bump package versions to match the changelogs.

kenzieschmoll:
@pq could be useful to have a lint that verifies your changelog version matches your pubspec version, for packages that maintain a changelog.

### DIFF
--- a/packages/devtools_app_shared/pubspec.yaml
+++ b/packages/devtools_app_shared/pubspec.yaml
@@ -1,6 +1,6 @@
 name: devtools_app_shared
 description: Package of Dart & Flutter structures shared between devtools_app and devtools extensions.
-version: 0.0.7
+version: 0.0.8
 repository: https://github.com/flutter/devtools/tree/master/packages/devtools_extensions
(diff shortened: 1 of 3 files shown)


CoderDake (approved):

pq:
> @pq could be useful to have a lint that verifies your changelog version matches your pubspec version, for packages that maintain a changelog.

Oh yeah. In fact, when we were publishing the linter I had a test that did something similar. A lint would have been nicer still. One trick here is that we don't currently analyze CHANGELOGs and we'd have to change that. (fyi @bwilkerson)

parlough:
Instead of a lint, a GitHub workflow might be a better choice.

@mosuem has some workflows in [firehose](https://github.com/dart-lang/ecosystem/tree/main/pkgs/firehose) that run verification and adds comments like https://github.com/dart-lang/ecosystem/pull/201#issuecomment-1822672467. Perhaps with some updates that tooling could benefit DevTools too now that there is more publishing from this repo.

Issue #6882, opened 2023-12-02
Too aggressively caching pub root directories


elliette:
This is a follow-up to https://github.com/flutter/devtools/issues/6841 (which was fixed by https://github.com/flutter/devtools/pull/6857)

This issue explains why our `inspector_service` tests failed to detect the regression. Once we infer the pub root directory for a project, we cache it across subsequent loads of DevTools, and never try to re-infer it.  

### When DevTools is opened for the first time, or after the cache has been cleared:

1. No cached directories are added to the widget inspector service ([code link](https://github.com/flutter/devtools/blob/58c6e1c8fc20c78979ced06c671c2ed6ff51c947/packages/devtools_app/lib/src/shared/preferences.dart#L343))
2. Fetching directories from the widget inspector service returns nothing ([code](https://github.com/flutter/devtools/blob/58c6e1c8fc20c78979ced06c671c2ed6ff51c947/packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart#L475))
3. Therefore, DevTools tries to infer the pub root directory (“Directory A”) and adds it to the widget inspector service ([code](https://github.com/flutter/devtools/blob/58c6e1c8fc20c78979ced06c671c2ed6ff51c947/packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart#L523))
4. Directory A is then cached ([code](https://github.com/flutter/devtools/blob/58c6e1c8fc20c78979ced06c671c2ed6ff51c947/packages/devtools_app/lib/src/shared/preferences.dart#L324))

### On any subsequent loads of DevTools:

1. The cached Directory A is added to the widget inspector service ([code link](https://github.com/flutter/devtools/blob/58c6e1c8fc20c78979ced06c671c2ed6ff51c947/packages/devtools_app/lib/src/shared/preferences.dart#L343))
2. Directory A is fetched from the widget inspector service ([code](https://github.com/flutter/devtools/blob/58c6e1c8fc20c78979ced06c671c2ed6ff51c947/packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart#L475))
3. Because Directory A was fetched, we do not try to infer the pub root directory ([code](https://github.com/flutter/devtools/blob/58c6e1c8fc20c78979ced06c671c2ed6ff51c947/packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart#L523))

## Expected behavior 

[...]

PR #6897, opened 2023-12-06
Only cache pub root directories set by the user

elliette:
Work towards https://github.com/flutter/devtools/issues/6882

This PR changes our caching logic around pub root directories to:
1. Always try to infer the app's pub root directory when we connect to an app
2. Only cache pub root directories that have been explicitly added by the user

This PR also:
* Renames a lot of variables that were prefixed by "customPubRootDirectoriesX" to "pubRootDirectoriesX", since the **custom** pub root directories only refer to those that have been explicitly added by the user and saved in the cache.
* Moves the logic for inferring the pub root directory from `InspectorService` to `InspectorPreferencesController` (with https://github.com/flutter/devtools/pull/6857, we no longer use the widget tree to infer the pub root directory, therefore there is no need for it to be in `InspectorService`)
* Adds test cases, including test cases to make sure that our google3-path parsing logic is sound (TODO from https://github.com/flutter/devtools/pull/6857)

Follow-up work (not included in this PR):
- Remove the Flutter pub root from the cache if we detect it there. (remediation for https://github.com/flutter/devtools/issues/6841)
- Invalidate the user's cache, so that any stale inferred directories are removed

### DIFF
--- a/packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
+++ b/packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
@@ -583,8 +583,12 @@ class PubRootDirectorySection extends StatelessWidget {
             textFieldLabel: 'Enter a new package directory',
             isRefreshing:
                 preferences.inspector.isRefreshingCustomPubRootDirectories,
-            onEntryAdded: (p0) =>
-                unawaited(preferences.inspector.addPubRootDirectories([p0])),
+            onEntryAdded: (p0) => unawaited(
+              preferences.inspector.addPubRootDirectories(
+                [p0],
+                shouldCache: true,
+              ),
+            ),
             onEntryRemoved: (p0) =>
                 unawaited(preferences.inspector.removePubRootDirectories([p0])),
             onRefreshTriggered: () =>
--- a/packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
+++ b/packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
@@ -126,7 +126,7 @@ class InspectorScreenBodyState extends State<InspectorScreenBody>
         searchPreventClose = false;
       }
     });
-    addAutoDisposeListener(preferences.inspector.customPubRootDirectories, () {
+    addAutoDisposeListener(preferences.inspector.pubRootDirectories, () {
       if (serviceConnection.serviceManager.hasConnection &&
           controller.firstInspectorTreeLoadCompleted) {
         _refreshInspector();
--- a/packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
+++ b/packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
@@ -579,10 +579,9 @@ class PubRootDirectorySection extends StatelessWidget {
           child: EditableList(
             gaScreen: gac.inspector,
             gaRefreshSelection: gac.refreshPubRoots,
-            entries: preferences.inspector.customPubRootDirectories,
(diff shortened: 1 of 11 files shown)


packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -579,16 +579,19 @@ class PubRootDirectorySection extends StatelessWidget {
...
-                unawaited(preferences.inspector.addPubRootDirectories([p0])),
+            isRefreshing: preferences.inspector.isRefreshingPubRootDirectories,
+            onEntryAdded: (p0) => unawaited(
+              preferences.inspector.addPubRootDirectories(
+                [p0],
+                shouldCache: true,
+              ),
+            ),
             onEntryRemoved: (p0) =>
                 unawaited(preferences.inspector.removePubRootDirectories([p0])),
```
kenzieschmoll, line 592: do we need to remove this from the cache here?

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -306,22 +380,20 @@ class InspectorPreferencesController extends DisposableController
...
+        final oldSet = Set<String>.of(_pubRootDirectories.value);
         final directoriesToAdd = newSet.difference(oldSet);
         final directoriesToRemove = oldSet.difference(newSet);
 
-        _customPubRootDirectories.removeAll(directoriesToRemove);
-        _customPubRootDirectories.addAll(directoriesToAdd);
-
-        _persistCustomPubRootDirectoriesToStorage();
+        _pubRootDirectories.removeAll(directoriesToRemove);
+        _pubRootDirectories.addAll(directoriesToAdd);
```
kenzieschmoll, line 396: nit use cascades `..`

elliette: Done!

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -331,14 +332,28 @@ class InspectorPreferencesController extends DisposableController
     }
   }
 
-  void _cacheCustomPubRootDirectories(
+  void _cachePubRootDirectories(
```
kenzieschmoll, line 335: return `Future<void>` now that we are awaiting in this method so that we can catch unawaited futures

packages/devtools_test/lib/src/mocks/fake_isolate_manager.dart
```diff
@@ -14,6 +14,12 @@ import 'package:vm_service/vm_service.dart';
 import 'generated.mocks.dart';
 
 base class FakeIsolateManager extends Fake with TestIsolateManager {
+  FakeIsolateManager({
+    this.rootLibrary,
+  });
+
+  String? rootLibrary;
```
kenzieschmoll, line 21: pass this all the way through to line 60.  We can add a default value for `rootLibrary` 

```
  FakeIsolateManager({
    this.rootLibrary = 'package:fake_app/main.dart',
  });

  final String rootLibrary;
  ```

and then below in the `isolateState` method: 
`final rootLib = LibraryRef(id: '0', uri: rootLibrary);`

Then you may not have to override `rootLibraryForMainIsolate` here at all.

elliette: Passing it now to the isolateState method, but switching over to using the mock service connection manager instead of the fake service connection manager required mocking a lot more of the functionality

kenzieschmoll (approved):

PR #6901, opened 2023-12-06
Move some integration test utilities to devtools_test for sharing with benchmark tests

kenzieschmoll:


packages/devtools_test/lib/src/test_data/sample_data.dart
```diff
@@ -0,0 +1,19 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
```
CoderDake, line 1:
```suggestion
// Copyright 2023 The Chromium Authors. All rights reserved.
```

Issue #6867, opened 2023-11-30
`devtools_tool update-flutter-sdk --from-path` is broken


kenzieschmoll:
I'm pretty sure this has to do with the change to always use the flutter-sdk from tools/. CC @DanTup 
```
 ~/develop/devtools/tool/ [master] devtools_tool update-flutter-sdk --from-path
Updating to Flutter version from cache: tags/3.17.0-18.0.pre 
Using Flutter SDK from /Users/kenzieschmoll/develop/devtools/tool/flutter-sdk
Updating local flutter/flutter repository...
Searching for a remote that points to flutter/flutter.git.
/Users/kenzieschmoll/develop/devtools/tool/flutter-sdk > git remote -v
origin  https://github.com/flutter/flutter (fetch)
origin  https://github.com/flutter/flutter (push)
Bad state: Couldn't find a remote that points to flutter/devtools.git. Instead got: 
origin  https://github.com/flutter/flutter (fetch)
origin  https://github.com/flutter/flutter (push)
```

DanTup:
I think the error is caused by a mix of having `flutter/flutter` and `flutter/flutter.git` in the code vs the origin in the clone.

However, the reason these don't match now is probably because we're trying to update a different SDK to what we did before. My assumption with the previous changes was that we basically never wanted to use Flutter from `PATH` (except the flutter customer tests) and I forgot about this flag here.

I guess my first question is why do we want this - why would we use the DevTools scripts to update a different Flutter than the one that we'll use? (eg., why can't we just remove the `--from-path` flag and this code path and _everything_ for DevTools uses the `tool/flutter-sdk`)?

kenzieschmoll:
> I guess my first question is why do we want this - why would we use the DevTools scripts to update a different Flutter than the one that we'll use? (eg., why can't we just remove the --from-path flag and this code path and everything for DevTools uses the tool/flutter-sdk)?

For local development, it is convenient to use the Flutter SDK on path (your local git checkout of flutter). For example, I often check out a specific flutter branch or have local changes in flutter/flutter that I want to develop with. The `--from-path` is supposed to be a convenient way to get your local flutter sdk to match what the bots will use.

DanTup:
> The `--from-path` is supposed to be a convenient way to get your local flutter sdk to match what the bots will use.

[...]

PR #6908, opened 2023-12-07
Add a global --flutter-from-path argument that overrides SDK selection

DanTup:
This promotes the `--flutter-from-path` flag to a global arg for all commands and uses it to locate the Flutter SDK before executing commands.

If set, we'll search for the SDK on PATH and fail if we don't find one.

If not set, we will infer an SDK from the one currently running the devtools_tool script (and again, fail if we're running from an SDK that doesn't have one).

When we call another instance of the tool, we bypass the shell script and instead execute it using Dart from the selected SDK (so that child processes use the selected SDK and we don't need to propagate the --flutter-from-path flag).

The update-flutter-sdk script was changed slightly to check whether the selected SDK is different to the tool/flutter-sdk one to decide whether to perform the extra update. This avoids some minor edge cases if:

- you have tool/flutter-sdk on PATH and pass the flag (it would update the Flutter SDK twice)
- you run devtools_tool from another SDK and don't pass the flag (it would not update the version being used to run the script)

Fixes https://github.com/flutter/devtools/issues/6867

### DIFF
--- a/tool/lib/commands/update_flutter_sdk.dart
+++ b/tool/lib/commands/update_flutter_sdk.dart
@@ -12,7 +12,6 @@ import 'package:path/path.dart' as path;
 
 import '../utils.dart';
 
-const _fromPathFlag = 'from-path';
 const _useCacheFlag = 'use-cache';
 
 /// This command updates the the Flutter SDK contained in the 'tool/' directory
@@ -33,22 +32,14 @@ const _useCacheFlag = 'use-cache';
 /// `devtools_tool update-flutter-sdk [--from-path] [--no-use-cache]`
 class UpdateFlutterSdkCommand extends Command {
   UpdateFlutterSdkCommand() {
-    argParser
-      ..addFlag(
-        _fromPathFlag,
-        abbr: 'p',
-        negatable: false,
-        help: 'Update the Flutter SDK that is on PATH (your local '
-            'flutter/flutter git checkout)',
-      )
-      ..addFlag(
-        _useCacheFlag,
-        negatable: true,
-        defaultsTo: true,
-        help: 'Update the Flutter SDK(s) to the cached Flutter version stored '
-            'in "flutter-candidate.txt" instead of the latest version at '
-            '"https://flutter.googlesource.com/mirrors/flutter/"',
-      );
+    argParser.addFlag(
+      _useCacheFlag,
+      negatable: true,
+      defaultsTo: true,
+      help: 'Update the Flutter SDK(s) to the cached Flutter version stored '
+          'in "flutter-candidate.txt" instead of the latest version at '
+          '"https://flutter.googlesource.com/mirrors/flutter/"',
+    );
   }
   @override
   String get name => 'update-flutter-sdk';
@@ -61,7 +52,6 @@ class UpdateFlutterSdkCommand extends Command {
 
   @override
   Future run() async {
-    final updateFlutterFromPath = argResults![_fromPathFlag];
     final useCachedVersion = argResults![_useCacheFlag];
     final log = Logger.standard();
 
@@ -100,16 +90,28 @@ class UpdateFlutterSdkCommand extends Command {
       '${useCachedVersion ? 'from cache' : 'from upstream'}: $flutterTag ',
     );
 
-    if (updateFlutterFromPath) {
-      final sdk = FlutterSdk.current;
+    final sdk = FlutterSdk.current;
+    final flutterSdkDirName = 'flutter-sdk';
+    final toolSdkPath = path.join(repo.toolDirectoryPath, flutterSdkDirName);
+    final toolFlutterSdkDirectory = Directory(toolSdkPath);
 
-      log.stdout('Updating local flutter/flutter repository...');
+    // Check if the Flutter SDK we're using is not the local tool/flutter-sdk
+    // one and if so update that too.
+    //
+    // Check paths case-insensitively because of potential differences like
+    // windows drive letters.
+    if (sdk.sdkPath.toLowerCase() !=
+        toolFlutterSdkDirectory.path.toLowerCase()) {
+      log.stdout(
+        'Current Flutter SDK is not tool/flutter-sdk, updating repository '
+        'at ${sdk.sdkPath}',
+      );
 
       // Verify we have an upstream remote to pull from.
       await findRemote(
         processManager,
         remoteId: 'flutter/flutter.git',
-        workingDirectory: sdk.sdkPath,
+        workingDirectory: FlutterSdk.current.sdkPath,
       );
 
       await processManager.runAll(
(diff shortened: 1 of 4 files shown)


tool/lib/commands/update_flutter_sdk.dart
```diff
@@ -100,16 +90,28 @@ class UpdateFlutterSdkCommand extends Command {
       '${useCachedVersion ? 'from cache' : 'from upstream'}: $flutterTag ',
     );
 
-    if (updateFlutterFromPath) {
-      final sdk = FlutterSdk.current;
+    final sdk = FlutterSdk.current;
+    final flutterSdkDirName = 'flutter-sdk';
+    final toolSdkPath = path.join(repo.toolDirectoryPath, flutterSdkDirName);
+    final toolFlutterSdkDirectory = Directory(toolSdkPath);
```
kenzieschmoll, line 96: nit: add a getter `toolFlutterSdkPath` to the DevToolsRepo class

tool/lib/commands/update_flutter_sdk.dart
```diff
@@ -33,22 +32,14 @@ const _useCacheFlag = 'use-cache';
...
 class UpdateFlutterSdkCommand extends Command {
   UpdateFlutterSdkCommand() {
-    argParser
-      ..addFlag(
-        _fromPathFlag,
-        abbr: 'p',
-        negatable: false,
-        help: 'Update the Flutter SDK that is on PATH (your local '
-            'flutter/flutter git checkout)',
-      )
```
kenzieschmoll, line 43: I think this should still be optional. We may not always want to wipe our local flutter checkout. Maybe we should change it to `--on-path` too.

tool/lib/devtools_command_runner.dart
```diff
@@ -20,6 +22,9 @@ import 'commands/rollback.dart';
 import 'commands/update_dart_sdk_deps.dart';
 import 'commands/update_version.dart';
 
+const _flutterFromPathFlag = 'flutter-from-path';
+const _flutterFromPathAlias = 'from-path';
```
kenzieschmoll, line 26: nit: is the alias necessary? can we remove to simplify?

tool/lib/model.dart
```diff
@@ -97,17 +97,26 @@ class DevToolsRepo {
...
+  /// The current located Flutter SDK.
   ///
   /// Tries to locate from the running Dart VM. If not found, will print a
   /// warning and use Flutter from PATH.
-  static final current = _findSdk();
+  static FlutterSdk get current {
+    if (_current == null) {
+      throw Exception(
+          'Cannot use FlutterSdk.current before SDK has been selected.'
+          'SDK selection is done by DevToolsCommandRunner.runCommand().');
```
kenzieschmoll, line 110: nit: trailing comma

tool/lib/commands/update_flutter_sdk.dart
DanTup: Done!

tool/lib/model.dart
DanTup: Done!

tool/lib/devtools_command_runner.dart
kenzieschmoll: I think the 'p' abbreviation is sufficient for short hand 👍

tool/lib/commands/update_flutter_sdk.dart
DanTup: > Uses the Flutter SDK from PATH to run the Processes started by the update-flutter-sdk command.
> vs
> Uses the `tool/flutter-sdk` to run the Processes started by the update-flutter-sdk command.

I guess my confusion is that it's not clear why we need this distinction. The two things you list that those commands do appears the same (update both Flutter SDKs). What's the purpose of running with the local tool/flutter-sdk SDK when updating the one from PATH?

I was trying to keep things simple by saying that basically you're always using one SDK or the other... Only one SDK exists in the world of the script, and you can choose the one from PATH or the one from the current running VM (which is essentially always tool/flutter-sdk because the shell scripts force that). Mixing them feels like it could cause confusion, because you need to know/remember which SDK is being used for what.


> Another way to summarize all this, if the intent of --flutter-from-path for all devtools_tool commands means "use my Flutter SDK on PATH for all processes and subprocesses started by this command", it feels strange for this flag to also mean "but when you run with this particular command update-flutter-sdk it actually does other things too".

I agree, and my original goal was for it not to do anything different at all, but rather just update whichever SDK was being used (and updated). However the original implementation seemed to update _both_ SDKs and I wasn't sure if it reasonable to change that. In summary:

- `--flutter-from-path` means always use the Flutter SDK from path for all spawned processes - Flutter or Dart, including devtools_tool. This occurs all the way down.
- `update-flutter-sdk` will always update the "current" SDK (based on that flag above), but additionally it will ensure "tool/flutter-sdk" is always up-to-date if it wasn't the one already being updated.

My preference would've been to remove that last part, (so update-flutter-sdk only ever updates one SDK - the one being run) because I think it's much easier to understand. If you want to update both SDKs, you could run `devtools_tool update-flutter-sdk && devtools_tool update-flutter-sdk --flutter-from-path`). I'm not sure if that addresses everything you wanted though?

tool/lib/commands/update_flutter_sdk.dart
```diff
@@ -100,16 +98,23 @@ class UpdateFlutterSdkCommand extends Command {
...
-      final sdk = FlutterSdk.current;
+    final flutterSdkDirName = repo.sdkDirectoryName;
+    final toolSdkPath = repo.toolFlutterSdkPath;
 
-      log.stdout('Updating local flutter/flutter repository...');
+    // Check if the Flutter SDK we're using is not the local tool/flutter-sdk
+    // one and if so update that too.
+    //
+    // Check paths case-insensitively because of potential differences like
+    // windows drive letters.
```
kenzieschmoll, line 108: this comment seems outdated

tool/lib/devtools_command_runner.dart
```diff
@@ -38,5 +42,25 @@ class DevToolsCommandRunner extends CommandRunner {
     addCommand(UpdateDevToolsVersionCommand());
     addCommand(UpdateFlutterSdkCommand());
     addCommand(UpdatePerfettoCommand());
+
+    argParser.addFlag(
+      _flutterFromPathFlag,
+      abbr: 'p',
+      negatable: false,
+      help: 'Use the Flutter SDK on PATH rather than the one being used to run '
+          'devtools_tool (usually tool/flutter-sdk)',
```
kenzieschmoll, line 51: nit this makes it sound like the tool/flutter-sdk is still being used to run `devtools_tool` even if the flutter from path flag is passed. Maybe reword: "Use the Flutter SDK on PATH to run devtools_tool commands and sub commands, instead of the Flutter SDK from tool/flutter-sdk which is used by default."

kenzieschmoll (approved):

tool/lib/commands/update_flutter_sdk.dart
DanTup: Oops, updated!

tool/lib/devtools_command_runner.dart
DanTup: Does my update make sense? Something that's slightly odd but I don't think matters enough to mention (because it makes things more verbose) is that this only applies to the spawned processes - the current process is using whatever the shell script decided to use (because we're already dunning a VM by the time we parse this flag).

I did consider once having the top level runner detect if this flag was set but it was running with a different SDK, and just re-run itself with the new one, but I'm not sure there would be any real benefit to it.

kenzieschmoll: If the only thing that could be run with a different SDK is the entrance to this script, then I think this is fine because all sub commands will respect the Flutter SDK we select.

DanTup: Yep, that's correct - as long as we don't add new code that accidentally runs things from path instead of using the current SDK, but as long as we go through `CliCommand.flutter` etc., then this should be hard to do accidentally.

PR #6910, opened 2023-12-07
Add a benchmark test for the Performance screen

kenzieschmoll:
Work towards https://github.com/flutter/devtools/issues/6552


packages/devtools_test/lib/src/helpers/actions.dart
```diff
@@ -119,3 +119,18 @@ Future<void> loadSampleData(
   await controller.tap(find.text('Load sample data'));
   await controller.pump(longPumpDuration);
 }
+
+/// Scrolls to the end of the first [Scrollable] descendant of the [T] widget.
```
polina-c, line 123: Can you give example?
Dos is confusing.
If my widget is list of 1000 images, will it scroll to the end of first image? Then it means scroll to start, not to end.
Will this end of first image be positioned at the start or at end of the parent widget?

elliette (approved):

PR #6911, opened 2023-12-07
Remove the Flutter pub root directory if it was accidentally cached

elliette:
Work towards https://github.com/flutter/devtools/issues/6882


packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -154,6 +154,7 @@ class InspectorPreferencesController extends DisposableController
   static const _customPubRootDirectoriesStoragePrefix =
       'inspector.customPubRootDirectories';
   String? _mainScriptDir;
+  bool _checkedFlutterPubRoot = false;
```
kenzieschmoll, line 157: will `readCachedPubRootDirectories` ever need to be called more than once? What about when we disconnect and connect to a new app - in that case does `_checkedFlutterPubRoot` need to be reset?

elliette: Ah good point! Resetting now on a new app connection (added a test for that as well)

kenzieschmoll (approved):

PR #6912, opened 2023-12-08
Bump to web_benchmarks 1.0.0

kenzieschmoll:


packages/devtools_app/benchmark/devtools_benchmarks_test.dart
```diff
@@ -30,57 +30,67 @@ final valueList = <String>[
 /// Tests that the DevTools web benchmarks are run and reported correctly.
 void main() {
   test(
-    'Can run a web benchmark',
+    'Can run web benchmarks',
     () async {
-      stdout.writeln('Starting web benchmark tests ...');
+      await _runBenchmarks();
+    },
+    timeout: const Timeout(Duration(minutes: 10)),
```
elliette, line 37: OOC, how long does this (and test below) usually take to run? Will this be running on CI, and if so would a shorter timeout be possible?

elliette (approved):
LGTM with a question

packages/devtools_app/benchmark/devtools_benchmarks_test.dart
kenzieschmoll: These do take a little while to run since all the benchmarks are being run as a single test case. The longest part is the time it takes to build the benchmark app (which is a profile build of DevTools). Locally these only take a couple of minutes but on the bots I have seen them take up to 6 minutes. 

I chose 10 to avoid timeout flakiness and to leave room for the runtime of another benchmark I plan on adding.

Issue #6929, opened 2023-12-11
Dynamic invocations in source code


srawlins:
There are some dynamic invocations in the source code, I think > 90% of it relates to JSON-parsing, so that's to be expected.

But it would be nice to clean this up. I found it while fixing `operator ==` overrides, and saw the "ignore: avoid-dynamic, it's necessary" comments, and was curious to see how much was necessary.

I enabled the `avoid_dynamic_calls` lint rule, and it found just over 400 violations.

PR #6940, opened 2023-12-11
Add type arguments on class declaration clauses

srawlins:

Without explicit type arguments, `dynamic` is implicit. Adding type arguments prevents dynamic calls, and you can see the bug catch at the end of the diff, where `preReleaseMajor` is no longer `dynamic`, and the type system can determine that it needs to be null checked.

## Pre-launch Checklist

- [x] I signed the [CLA].

![build.yaml badge]

### DIFF
--- a/packages/devtools_shared/lib/src/utils/semantic_version.dart
+++ b/packages/devtools_shared/lib/src/utils/semantic_version.dart
@@ -6,7 +6,7 @@ import 'dart:math' as math;
 
 import 'compare.dart';
 
-class SemanticVersion with CompareMixin {
+class SemanticVersion with CompareMixin<SemanticVersion> {
   SemanticVersion({
     this.major = 0,
     this.minor = 0,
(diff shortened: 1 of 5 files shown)


packages/devtools_shared/lib/src/utils/semantic_version.dart
```diff
@@ -6,7 +6,7 @@ import 'dart:math' as math;
 
 import 'compare.dart';
 
-class SemanticVersion with CompareMixin {
```
srawlins, line 9: Should this have release notes? Is this code imported by users?

Any code that was passing in an object other than a `SemanticVersion` into `compareTo` was a bug, but this could still be breaking?

kenzieschmoll: According to pub, these are the only packages that depend on devtools_shared (in alignment with what I would expect). None of those packages break as far as I can tell. Can you add a Changelog entry for this change?

kenzieschmoll (approved):

packages/devtools_shared/lib/src/utils/semantic_version.dart
srawlins: Sounds good; done!

kenzieschmoll (approved):

Issue #6874, opened 2023-11-30
Add a setting to make Widget Details Tree the default details view


kenzieschmoll:
Some users complained that they do not like that the Layout Explorer is the default view. It would be cheap to add a setting to allow users to pick which details view is their default.

Longer term, we could remember tab order if the order was configurable from the tab bar: https://github.com/flutter/flutter/issues/127693

kenzieschmoll:
We could use a radio button selector like we have on the performance page:
![Screenshot 2023-11-30 at 3 44 14 PM](https://github.com/flutter/devtools/assets/43759233/50b652d2-6019-4b32-bdab-2832dba75f94)

something in the inspector settings like
```
Default inspector details view
    [ ] Layout Explorer     [ ] Widget Details View
```

PR #6949, opened 2023-12-11
Add an inspector preference for the default inspection tab

CoderDake:
I'm not sure if my copy is very good so it may need some love.

Screenshot 2023-12-11 at 6 42 11 PM

Fixes https://github.com/flutter/devtools/issues/6874


kenzieschmoll:
Can we get rid of the indent before the radio buttons? or at least make it more subtle so that it is slightly indented from the description text. 

Also, we can put this setting under general, it doesn't need its own header / subsection

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -137,9 +137,13 @@ class PreferencesController extends DisposableController
   }
 }
 
+enum InspectorDetailsViewOptions { layoutExplorer, widgetDetailsView }
```
kenzieschmoll, line 140: supernit: InspectorDetailsViewType to match singular and "Type" word pattern we use elsewhere in DevTools

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -150,7 +154,14 @@ class InspectorPreferencesController extends DisposableController
...
   final _pubRootDirectoriesAreBusy = ValueNotifier<bool>(false);
   final _busyCounter = ValueNotifier<int>(0);
+  final _defaultInspectorDetailsView =
+      ValueNotifier<InspectorDetailsViewOptions>(
+    InspectorDetailsViewOptions.layoutExplorer,
+  );
+
   static const _hoverEvalModeStorageId = 'inspector.hoverEvalMode';
+  static const _defaultInspectorDetailsViewStorageId =
+      'inspector.detailsViewSelection';
```
kenzieschmoll, line 164: nit: let's name "inspector.defaultDetailsViewType" so that it is clear that this isn't updated each time they select a tab, but rather indicates their selection of a setting

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -194,6 +206,27 @@ class InspectorPreferencesController extends DisposableController
...
+    });
+  }
+
+  Future<void> _updateInspectorDetailsViewSelection() async {
+    final inspectorDetailsView =
+        await storage.getValue(_defaultInspectorDetailsViewStorageId);
+
+    if (inspectorDetailsView != null) {
+      _defaultInspectorDetailsView.value = InspectorDetailsViewOptions.values
+          .firstWhere((e) => e.toString() == inspectorDetailsView);
```
kenzieschmoll, line 226: should we use e.name here instead?

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -565,6 +569,59 @@ class _ErrorNavigatorButton extends StatelessWidget {
...
+      builder: (context, selection, _) {
+        final theme = Theme.of(context);
+        return Column(
+          children: [
+            Text(
+              'Select the tab which opens by default when opening the inspector.',
+              style: theme.subtleTextStyle,
+            ),
+            const SizedBox(
+              height: denseSpacing,
```
kenzieschmoll, line 588: style nit: no need for trailing comma on SizedBox widgets that only take one line.

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -565,6 +569,59 @@ class _ErrorNavigatorButton extends StatelessWidget {
...
+            Row(
+              mainAxisSize: MainAxisSize.min,
+              children: [
+                Radio<InspectorDetailsViewOptions>(
+                  value: InspectorDetailsViewOptions.layoutExplorer,
+                  materialTapTargetSize: MaterialTapTargetSize.shrinkWrap,
+                  groupValue: selection,
+                  onChanged: _onChanged,
+                ),
+                const Text('Layout Explorer'),
```
kenzieschmoll, line 599: these strings should be added to the InspectorDetailsViewOptions enum, and should also be used for the tab titles so these values stay in sync

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -565,6 +569,59 @@ class _ErrorNavigatorButton extends StatelessWidget {
...
+              ],
+            ),
+          ],
+        );
+      },
+    );
+  }
+
+  void _onChanged(InspectorDetailsViewOptions? value) {
+    if (value != null) {
```
kenzieschmoll, line 619: we should be sending analytics events here

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -356,6 +357,9 @@ class FlutterInspectorSettingsDialog extends StatelessWidget {
               gaItem: gac.inspectorHoverEvalMode,
             ),
             const SizedBox(height: denseSpacing),
+            ...dialogSubHeader(theme, 'Default Inspector Details View'),
```
elliette, line 360: Looks like Kenzie wanted this subheading removed (so that the options fall under general)

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -565,6 +569,63 @@ class _ErrorNavigatorButton extends StatelessWidget {
   }
 }
 
+class InspectorDetailsOption extends StatelessWidget {
```
elliette, line 572: Could we rename to something more descriptive like "DefaultTabOption"?

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -565,6 +569,63 @@ class _ErrorNavigatorButton extends StatelessWidget {
...
+  const InspectorDetailsOption({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder(
+      valueListenable: preferences.inspector.defaultInspectorDetailsView,
```
elliette, line 578: Whatever the class gets renamed to, lets have the value listenable match (e.g. `preferences.inspector.defaultTab`)

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -565,6 +569,63 @@ class _ErrorNavigatorButton extends StatelessWidget {
...
+      builder: (context, selection, _) {
+        final theme = Theme.of(context);
+        return Column(
+          crossAxisAlignment: CrossAxisAlignment.start,
+          children: [
+            Text(
+              'Select the tab which opens by default when opening the inspector.',
```
elliette, line 585: Suggestion: "Set the default tab for the inspector on opening"

elliette: Or "Select the default tab for the inspector on opening" (mostly trying to avoid having the word "open" appear twice)

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
elliette: Ahh I think just renaming to InspectorDetailsViewOption would be more clear (then everything else wouldn't have to be renamed)

packages/devtools_app/lib/src/shared/preferences.dart
```diff
@@ -195,6 +216,27 @@ class InspectorPreferencesController extends DisposableController
     setHoverEvalMode(hoverEvalModeEnabledValue == 'true');
   }
 
+  Future<void> _initDefaultInspectorDetailsView() async {
+    await _updateInspectorDetailsViewSelection();
```
elliette, line 220: - `_updateInspectorDetailsViewSelection` sets `_defaultInspectorDetailsView.value` to be whatever was set in storage
- and then in this method, we then set the value in storage to be whatever is in `_defaultInspectorDetailsView.value`? 

Maybe I'm misunderstanding, but is setting the value in storage again necessary?

I'm also curious what happens if a value is saved in storage, and then the user updates the setting. Where does the new setting get set in storage?

CoderDake: The _updateInspector... checks if anyvalue is already set in storage, if so it sets the current default to that one. Otherwise the default default is used. I guess default is a bit confusing here since it is a double default :P

Line 222 here is a listener that will update the value in storage any time the listener gets updated.

elliette: Ah got it! Makes sense

kenzieschmoll:
Make this text have the regular text style instead of the subtle text style
Screenshot 2024-01-03 at 12 30 59 PM

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -333,7 +334,7 @@ class FlutterInspectorSettingsDialog extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
-    final dialogHeight = scaleByFontFactor(400.0);
+    final dialogHeight = scaleByFontFactor(600.0);
```
kenzieschmoll, line 337: maybe 500 now that you got rid of the second heading?

packages/devtools_app/lib/src/screens/inspector/inspector_screen.dart
```diff
@@ -580,10 +580,7 @@ class InspectorDefaultDetailsViewOption extends StatelessWidget {
         return Column(
           crossAxisAlignment: CrossAxisAlignment.start,
           children: [
-            Text(
-              'Select the default tab for the inspector.',
-              style: theme.subtleTextStyle,
-            ),
+            const Text('Select the default tab for the inspector.'),
```
kenzieschmoll, line 583: manually apply theme.regularTextStyle to this text (I'm working on another PR where not having the style  will cause theming bugs)

kenzieschmoll (approved):

PR #6961, opened 2023-12-14
[DevTools extensions] Only use `Uri.file` for windows platforms

kenzieschmoll:
This was changed from `Uri.parse` to `Uri.file` in https://github.com/flutter/devtools/commit/89c7a4861fbd874c5af5c919a95a6e32af0fedcc. While this fixed extension detection on windows configuration, it broke extension detection on macos. The reasons why this behaves differently between platforms is a mystery to me (seems like a bug with how we parse URIs, IMO).

More context at https://github.com/dart-lang/tools/issues/220. This issue breaking DevTools extensions on flutter master. Once this lands, I will do an SDK roll.

I filed https://github.com/flutter/devtools/issues/6962 to track adding test coverage for this regression.

### DIFF
--- a/packages/devtools_shared/CHANGELOG.md
+++ b/packages/devtools_shared/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 6.0.3
+* Fix an issue parsing file paths on Mac that could prevent extensions from being detected.
+
 # 6.0.2
 * Fix an issue parsing file paths on Windows that could prevent extensions from being detected.
 
--- a/packages/devtools_shared/CHANGELOG.md
+++ b/packages/devtools_shared/CHANGELOG.md
@@ -1,4 +1,9 @@
 # 6.0.3
+* `CompareMixin` is now generic, implementing `Comparable<T>` instead of
+  `Comparable<dynamic>`, and it's operators each therefore accept a `T`
+  argument.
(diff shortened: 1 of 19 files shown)


packages/devtools_shared/CHANGELOG.md
```diff
@@ -1,9 +1,10 @@
-# 6.0.3-wip
+# 6.0.3
 * `CompareMixin` is now generic, implementing `Comparable<T>` instead of
   `Comparable<dynamic>`, and it's operators each therefore accept a `T`
   argument.
 * `SemanticVersion` now mixes in `CompareMixin<SemanticVersion>`, and it's
   `compareTo` method therefore now accepts a `SemanticVersion`.
+* Fix an issue parsing file paths on Mac that could prevent extensions from being detected.
```
polina-c, line 7: Any way to test cover the fix?

polina-c (approved):
lgtm with one comment

packages/devtools_shared/CHANGELOG.md
kenzieschmoll: Theres a tracking issue for that: https://github.com/flutter/devtools/issues/6962. First priority is to land fix to Dart SDK to unblock extension authors.

PR #6968, opened 2023-12-14
[DevTools extensions] Use `Uri.parse` for windows platforms too

kenzieschmoll:
context: https://github.com/dart-lang/tools/issues/220.
The fix committed here should actually apply to windows too: https://github.com/flutter/devtools/pull/6961

### DIFF
--- a/packages/devtools_shared/lib/src/extensions/extension_manager.dart
+++ b/packages/devtools_shared/lib/src/extensions/extension_manager.dart
@@ -44,33 +44,33 @@ class ExtensionsManager {
   /// [serveAvailableExtensions] is called.
   final devtoolsExtensions = <DevToolsExtensionConfig>[];
 
-  /// Serves any available DevTools extensions for the given [rootPath], where
-  /// [rootPath] is the root for a Dart or Flutter project containing the
-  /// `.dart_tool/` directory.
+  /// Serves any available DevTools extensions for the given [rootPathFileUri],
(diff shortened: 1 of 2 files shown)


packages/devtools_shared/lib/src/extensions/extension_manager.dart
```diff
@@ -44,33 +44,33 @@ class ExtensionsManager {
...
-        );
-        // Only use [Uri.file] for windows platforms (https://github.com/dart-lang/tools/issues/220).
-        final packageConfigUri = Platform.isWindows
-            ? Uri.file(packageConfigPath)
-            : Uri.parse(packageConfigPath);
         extensions = await findExtensions(
           'devtools',
-          packageConfig: packageConfigUri,
+          packageConfig: Uri.parse(
+            path.join(
```
DanTup, line 68: Using `path.join` on the URI string feels slightly weird here. On Windows it will use backslashes and will end up sending this to `Uri.parse()`:

```
file:///C:/Dev/Test%20Projects/dart_sample/bin/uri.dart\foo
```

It seems to work, but it feels odd. The easiest thing is probably to use `path.posix.join` instead (otherwise, we could `.toFilePath()`, `join`, then `Uri.file`).

DanTup (approved):
Tested the fix on Windows and it's good, although I have two small suggestions.

Issue #6969, opened 2023-12-14
`devtools_tool` does not work on first run


kenzieschmoll:
The `devtools_tool` script assumes you have the `devtools/tool/flutter-sdk` directory in your DevTools repo: https://github.com/flutter/devtools/blob/master/tool/bin/devtools_tool/#L10

However, for a fresh clone of DevTools, this directory does not exist. To set up this Flutter SDK under `devtools/tool/flutter-sdk`, a user should use the `devtools_tool update-flutter-sdk` command, but they cannot use this command since the `devtools/tool/flutter-sdk` directory does not exist.

This script should be smarter and check if the `devtools/tool/flutter-sdk` exists, and if it does not, it should check it out.

PR #6970, opened 2023-12-14
Fix first run for `devtools_tool`

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/6969 so that the first execution of `devtools_tool` after a fresh clone of DevTools works as intended. As part of our Contributing guide, we first tell users to get Flutter and Dart set up on their path, so at the point they begin to interact with the `devtools_tool` script, we can assume that `dart` will be available.

### DIFF
--- a/tool/bin/devtools_tool
+++ b/tool/bin/devtools_tool
@@ -4,17 +4,16 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
 if [ ! -z "$DEVTOOLS_TOOL_FLUTTER_FROM_PATH" ]
 then
-	echo Running devtools_tool using Dart/Flutter from PATH because DEVTOOLS_TOOL_FLUTTER_FROM_PATH is set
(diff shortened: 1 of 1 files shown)


tool/bin/devtools_tool
```diff
@@ -7,5 +7,14 @@ then
 	echo Running devtools_tool using Dart/Flutter from PATH because DEVTOOLS_TOOL_FLUTTER_FROM_PATH is set
 	dart run "$SCRIPT_DIR/devtools_tool.dart" "$@"
 else
+	
+	if [ ! -d $SCRIPT_DIR/../flutter-sdk ]
+	then
+		# If the `devtools/tool/flutter-sdk` directory does not exist yet, use whatever Dart
+		# is on the user's path to update it before proceeding.
+		echo "Running devtools_tool using the Dart SDK from `which dart` to create the Flutter SDK in tool/flutter-sdk."
+    	dart run "$SCRIPT_DIR/devtools_tool.dart" update-flutter-sdk
```
DanTup, line 16: nit: indenting seems wrong

DanTup (approved):
LGTM! We'll need to do the same in the `.bat` for Windows - I can do that and test it.

PR #6973, opened 2023-12-15
Rename ToCsvButton to  DownloadButton to enable different types of download/save/export buttons.

polina-c:


kenzieschmoll (approved):
Just so you know, we already have an existing button `OpenSaveButtonGroup` that you might be able to use for your use case.

PR #6948, opened 2023-12-11
Cast JSON objects to their expected type avoiding follow-on dyamic calls

srawlins:
Work towards https://github.com/flutter/devtools/issues/6929. This is the big one.

This diff is unpleasant, with so many explicit casts. But this is what is required to avoid dynamic calls in this JSON-processing-heavy code. Feedback about whether we want dynamic-free, explicit cast code, _as a requirement_ (like, guarded with a lint rule), is 100% welcome please let me know 😁 .

This diff is not too large (+75/-72 lines) but it touches a lot of files, and I'm happy to break it up, if that helps the review.

The new extension types _should be_ a real alternative to casting JSON-parsed objects like this. They require Dart 3.3, so we can't land any yet (I think?). But as an example, let's look at the diff in `logging_controller.dart`. Instead of casting each value like `entry['source'] as String?`, we can type `entry` as an Extension Type called `Entry`:

```dart
extension type Entry(Map<String, Object?> data) {
  String? get source => data['source'];
  Map<String, Object?>? get displaySize => (data['displaySize'] as Map).cast<String, Object?>();
  Map<String, Object?>? get imageSize => (data['imageSize'] as Map).cast<String, Object?>();
}
```

Then all access on this object is restricted to the API declared above. If folks are OK with the code as is, and want to explore extension types later (soon), we could also kick the can down the road.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or there is a reason for not adding tests.

![build.yaml badge]

If you need help, consider asking for help on [Discord].

[Contributor Guide]: https://github.com/flutter/devtools/blob/master/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[CLA]: https://cla.developers.google.com/
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[build.yaml badge]: https://github.com/flutter/devtools/actions/workflows/build.yaml/badge.svg

PR #6977, opened 2023-12-18
Add extension type for profile report range JSON

srawlins:
An alternative to some of the code in https://github.com/flutter/devtools/pull/6948

This one is, I think, a fairly straightforwards extension type. It contains the safe access of JSON fields, and the const field names. We wrap access to the JSON as early as possible, at the very end of the diff, line ~1069, before we enter `ProfileReportRange._fromJson`. Other notes:

* _ProfileReportRangeJson is private. I think this makes sense; it is only accessed from this library.
* The representation field in _ProfileReportRangeJson is public (`json`). Shrug. It could be private. Maybe it should be.
* The `_profile` field, the outermost access on `json` is private, because it is only ever accessed _from within the extension type_. But if this clashes with local style, we can change.
* The other extension type fields are public.
* No verification is done. It is possible, and maybe desirable, to have a constructor that verifies the structure of the JSON. But I don't know what that structure should be (aside from the fields used), so I didn't add any verification.

Other cleanups here:

* ProfileReportMetaData had one field, was only accessed within the same library, and was only accessed to access that field; removed.
* None of the `metadata`, `inclusiveTicks`, `exclusiveTicks`, or `list` fields on ProfileReportRange were accessed outside the constructor. So, removed.

## Pre-launch Checklist

- [x] I signed the [CLA].

![build.yaml badge]

### DIFF
--- a/packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
+++ b/packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
@@ -1022,6 +1022,9 @@ class ProfileReportRange {
         .toList();
     for (int i = 0; i < lines.length; ++i) {
       final line = lines[i];
+      // In a `Map<int, ProfileReportEntry>`, we're mapping an `int` to a
+      // `ProfileReportEntry`. No bug.
+      // ignore: avoid-collection-methods-with-unrelated-types
       entries[line] = ProfileReportEntry(
         sampleCount: sampleCount,
         line: line,
--- a/packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
+++ b/packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
@@ -1006,7 +1006,6 @@ class ProfileReportEntry {
 /// Profiling information for a range of token positions in a [Script].
 class ProfileReportRange {
   ProfileReportRange._fromJson(Script script, _ProfileReportRangeJson json) {
-    final sampleCount = json.metadata[_kSampleCountKey] as int;
     final inclusiveTicks = json.inclusiveTicks;
     final exclusiveTicks = json.exclusiveTicks;
     final lines = json.positions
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
```diff
@@ -1010,47 +1005,58 @@ class ProfileReportEntry {
...
-              // either be a negative value or a String (e.g., 'ParallelMove'
-              // or 'NoSource'). We'll just use -1 as a placeholder since we
-              // won't display anything for these tokens anyway.
-              (e) => e is int
-                  ? script.getLineNumberFromTokenPos(e) ?? _kNoSourcePosition
-                  : _kNoSourcePosition,
-            )
-            .toList() {
+  ProfileReportRange._fromJson(Script script, _ProfileReportRangeJson json) {
+    final sampleCount = json.metadata[_kSampleCountKey] as int;
```
kenzieschmoll, line 1009: should sampleCount be a getter in the extension type?

packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
```diff
@@ -1010,47 +1005,58 @@ class ProfileReportEntry {
...
-        sampleCount: metadata.sampleCount,
+        sampleCount: sampleCount,
         line: line,
         inclusive: inclusiveTicks[i],
         exclusive: exclusiveTicks[i],
       );
     }
   }
 
+  static const _kSampleCountKey = 'sampleCount';
```
kenzieschmoll, line 1037: if we add a sampleCount getter below, this const could move into the extension type too

kenzieschmoll (approved):

packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
srawlins: Done

packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
srawlins: done

PR #6980, opened 2023-12-18
Add CpuProfileDataJson extension type for handling this data

srawlins:
An alternative to some of the code in https://github.com/flutter/devtools/pull/6948

This one is again fairly simple. `CpuProfileData.parse` used to handle all of the parsing as statements, and relies on a fair amount of dynamic calling. When we extract the field calls into extension type getters, the casting etc is isolated to that section of the code. The statements in `CpuProfileData.parse` get simpler.

This could probably be improved more; I think `CpuProfileMetaData()` should just accept a `_CpuProfileDataJson` and get the data it needs. But that can be for later. Or now. Also all of the null-aware code could also be moved into the extension type. Depending on the business logic. Is it ever important to know if `sampleCount` is `null`? Or can we default it to `0` at the root? Etc.

## Pre-launch Checklist

- [x] I signed the [CLA].

![build.yaml badge]

### DIFF
--- a/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
+++ b/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
@@ -569,7 +569,7 @@ class CpuProfileData {
       if (tree.frameId == kRootId) {
         continue;
       }
-      (traceObject[_traceEventsKey]! as List<Object?>).add({
+      (traceObject[CpuProfileData._traceEventsKey]! as List<Object?>).add({
         'ph': 'P', // kind = sample event
         'name': '', // Blank to keep about:tracing happy
         'pid': cpuSamples.pid,
--- a/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
+++ b/packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
@@ -168,16 +168,17 @@ class CpuProfileData {
     _cpuProfileRoot = CpuStackFrame.root(profileMetaData);
   }
 
-  factory CpuProfileData.parse(Map<String, dynamic> json) {
+  factory CpuProfileData.parse(Map<String, dynamic> json_) {
+    final json = _CpuProfileDataJson(json_);
     final profileMetaData = CpuProfileMetaData(
-      sampleCount: json[sampleCountKey] ?? 0,
-      samplePeriod: json[samplePeriodKey] ?? 0,
-      stackDepth: json[stackDepthKey] ?? 0,
-      time: (json[timeOriginKey] != null && json[timeExtentKey] != null)
+      sampleCount: json.sampleCount ?? 0,
+      samplePeriod: json.samplePeriod ?? 0,
+      stackDepth: json.stackDepth ?? 0,
+      time: (json.timeOriginMicros != null && json.timeExtentMicros != null)
           ? (TimeRange()
-            ..start = Duration(microseconds: json[timeOriginKey])
+            ..start = Duration(microseconds: json.timeOriginMicros!)
             ..end = Duration(
-              microseconds: json[timeOriginKey] + json[timeExtentKey],
+              microseconds: json.timeOriginMicros! + json.timeExtentMicros!,
             ))
           : null,
     );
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -211,12 +212,7 @@ class CpuProfileData {
     }
 
     // Initialize all CPU samples.
-    final stackTraceEvents =
-        (json[traceEventsKey] ?? []).cast<Map<String, dynamic>>();
-    final samples = stackTraceEvents
-        .map((trace) => CpuSampleEvent.parse(trace))
-        .toList()
-        .cast<CpuSampleEvent>();
+    final samples = json.traceEvents ?? [];
```
kenzieschmoll, line 215: can we give the empty list a type?

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -573,7 +569,7 @@ class CpuProfileData {
       if (tree.frameId == kRootId) {
         continue;
       }
-      (traceObject[CpuProfileData.traceEventsKey]! as List<Object?>).add({
+      (traceObject[_traceEventsKey]! as List<Object?>).add({
```
kenzieschmoll, line 572: should this be CpuProfileData._traceEventsKey for consistency?

kenzieschmoll (approved):

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
srawlins: I'm sort of following ["DON’T write type arguments on generic invocations that are inferred"](https://dart.dev/effective-dart/design#dont-write-type-arguments-on-generic-invocations-that-are-inferred) here. Since `json.traceEvents` is a `List<CpuSampleEvent>?`, the type argument is inferred on `[]`, and `samples` has the correct type of `List<CpuSampleEvent>`.

All that being said though, I'm happy to add a type argument if it makes the code more readable.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
srawlins: Done

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -168,24 +168,25 @@ class CpuProfileData {
     _cpuProfileRoot = CpuStackFrame.root(profileMetaData);
   }
 
-  factory CpuProfileData.parse(Map<String, dynamic> json) {
+  factory CpuProfileData.parse(Map<String, dynamic> json_) {
+    final json = _CpuProfileDataJson(json_);
```
jacob314, line 172: I wonder if we can go a step further and just have a single CpuProfileData class that is is an extension type on Map<String, dynamic>
?
profileMetaData and individual frames could be computed as needed from the json. This might be horrible for performance as some of the logic for computing frames looks a bit expensive.

packages/devtools_app/lib/src/screens/profiler/cpu_profile_model.dart
```diff
@@ -168,24 +168,25 @@ class CpuProfileData {
...
+              microseconds: json.timeOriginMicros! + json.timeExtentMicros!,
             ))
           : null,
     );
 
     // Initialize all stack frames.
     final stackFrames = <String, CpuStackFrame>{};
     final Map<String, Object?> stackFramesJson =
-        jsonDecode(jsonEncode(json[stackFramesKey] ?? <String, Object?>{}));
+        jsonDecode(jsonEncode(json.stackFrames ?? <String, Object?>{}));
```
jacob314, line 189: @kenzieschmoll do you remember why we are jsonDecoding + encoding? those are expensive operations that are definitely not worth it if we are just trying to get a nice default value for the case where there are no frames. Aside from this decode+encode I don't see anything that is clearly expensive enough that we can't do the parsing more lazily. More lazy parsing might also be faster as right now we parse all the fields and I don't know if we need all of them right away or just a few.

Issue #6981, opened 2023-12-18
Flutter sidebar infinite spinner


kenzieschmoll:
![Screenshot 2023-12-18 at 9 53 21 AM](https://github.com/flutter/devtools/assets/43759233/339f15ea-bd95-4d0c-b0d9-528b150ef562)

@DanTup 

![Screenshot 2023-12-18 at 9 54 43 AM](https://github.com/flutter/devtools/assets/43759233/3f681e53-cd01-458d-9d63-07d06e92d7c5)

VS code version info:

Version: 1.85.0 (Universal)
Commit: af28b32d7e553898b2a91af498b1fb666fdebe0c
Date: 2023-12-06T18:18:04.614Z
Electron: 25.9.7
ElectronBuildId: 25551756
Chromium: 114.0.5735.289
Node.js: 18.15.0
V8: 11.4.183.29-electron.0
OS: Darwin x64 23.2.0

DanTup:
Yeah, I'm looking into this. Seems to only be on master and not stable. I can repro running from DevTools from source, but I'm not _certain_ I'm hitting the same issue because the VS Code API isn't available running this way, but I'll dig into it.

DanTup:
Running standalone is probably a red herring, because we don't have anyone listening to our `postMessage` calls.

When running inside VS Code, DevTools logs that it sends:

```
==> {jsonrpc: 2.0, method: vsCode.getCapabilities, id: 0}
```

VS Code logs that it receives and replies:

```
[Extension Host] VS CODE GOT: {"jsonrpc":"2.0","method":"vsCode.getCapabilities","id":0}
[Extension Host] Sending response... {"id":0,"result":{"executeCommand":true,"hotReload":true,"hotRestart":true,"openDevToolsExternally":true,"openDevToolsPage":true,"selectDevice":true}}
```

However DevTools never gets the response via `postMessage`. So something is probably wrong in VS Code sending with `postMessage` or DevTools receiving with `postMessage`. It's not clear which, although since it works with stable it might be in the DevTools side.

It would be easier to debug if Chrome DevTools showed `postMessage`s :)

PR #6984, opened 2023-12-18
Fix incoming postMessages for package:web

DanTup:
`message.data` here is `LegacyJavaScriptObject` which is different to the `dart:html` version. We need to call `dartify()` to get this back to the Dart map we expect (which matches the `jsify()` call on the way out.

Fixes https://github.com/flutter/devtools/issues/6981

We need to look at whether we can have better automated tests for things that go over postMessage (our current mock VS Code environment doesn't go via postMessage because it's just hosted as a widget inside a single Flutter page - but perhaps it should be real iframes), but we should get this fix landed first.

@jacob314 @kenzieschmoll

### DIFF
--- a/packages/devtools_app/lib/src/shared/config_specific/post_message/_post_message_web.dart
+++ b/packages/devtools_app/lib/src/shared/config_specific/post_message/_post_message_web.dart
@@ -12,7 +12,7 @@ Stream<PostMessageEvent> get onPostMessage {
   return window.onMessage.map(
     (message) => PostMessageEvent(
       origin: message.origin,
-      data: message.data,
+      data: message.data.dartify(),
     ),
   );
 }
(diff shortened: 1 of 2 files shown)


jacob314 (approved):
lgtm

kenzieschmoll:
We have a section for VS Code Sidebar Updates on our NEXT_RELEASE_NOTES.md section. It seems like these changes should be announced with the Dart Code announcements, but the change itself is pinned to the DevTools / Flutter version. So for now probably fine to document the release notes in both places. Yes, if you could add a release note, that would be great. Thanks.

kenzieschmoll (approved):

kenzieschmoll:
Rolling this commit into the Dart SDK with https://dart-review.googlesource.com/c/sdk/+/342384

Issue #7025, opened 2024-01-05
Add a checkbox filter to the logging filter view to hide Flutter framework logs 


kenzieschmoll:
The logs are cluttered with `Flutter.Frame` and `Flutter.imagesize...` logs that aren't interesting to the user.

PR #7026, opened 2024-01-05
Add toggle filters to the logging page for Flutter and Dart noise

kenzieschmoll:
Screenshot 2024-01-05 at 11 46 46 AM

Fixes https://github.com/flutter/devtools/issues/7025


packages/devtools_app/lib/src/screens/logging/logging_controller.dart
```diff
@@ -548,40 +578,49 @@ class LoggingController extends DisposableController
...
-              !kindArg.matchesValue(log.kind.toLowerCase())) {
-            return false;
-          }
 
-          if (filter.queryFilter.substrings.isNotEmpty) {
-            for (final substring in filter.queryFilter.substrings) {
-              final matchesKind = log.kind.caseInsensitiveContains(substring);
-              if (matchesKind) return true;
+    bool filterCallback(LogData log) {
+      for (final toggleFilter in filter.toggleFilters) {
```
elliette, line 583: Could we set this check to a variable to make it easier to understand what it's checking for?

Something like (might be a better way to write this):
```
final matchesToggledFilter = filter.toggleFilters.any(toggleFilter.enabled.value && !toggleFilter.includeCallback(log));
if (matchesToggledFilter) return false;
```

packages/devtools_app/lib/src/screens/logging/logging_controller.dart
```diff
@@ -548,40 +578,49 @@ class LoggingController extends DisposableController
...
+        if (toggleFilter.enabled.value && !toggleFilter.includeCallback(log)) {
+          return false;
+        }
+      }
+
+      final queryFilter = filter.queryFilter;
+      if (!queryFilter.isEmpty) {
+        final kindArg = queryFilter.filterArguments[kindFilterId];
+        if (kindArg != null &&
```
elliette, line 592: Ditto for making this a variable

elliette (approved):
LGTM with a couple suggestions!

packages/devtools_app/lib/src/screens/logging/logging_controller.dart
kenzieschmoll: Added a TODO. I will address in a follow up PR to clean up the query filter arguments in general.

PR #7027, opened 2024-01-05
Add support for filtering by regular expressions

kenzieschmoll:
Screenshot 2024-01-05 at 4 40 53 PM

This mimics the behavior in VS Code search / filter text fields. Fixes https://github.com/flutter/devtools/issues/7024

This PR also cleans up some query filter logic to simplify filterData overrides.


packages/devtools_app/lib/src/shared/primitives/utils.dart
```diff
@@ -1061,9 +1061,20 @@ extension StringExtension on String {
     return true;
   }
 
-  /// Whether [other] is a case insensitive match for this String
-  bool caseInsensitiveEquals(String? other) {
-    return toLowerCase() == other?.toLowerCase();
+  /// Whether [other] is a case insensitive match for this String.
+  ///
+  /// If [pattern] is a [RegExp], this method will return true iff this String
```
elliette, line 1066: nit: if (typo)

packages/devtools_app/lib/src/shared/primitives/utils.dart
```diff
@@ -1061,9 +1061,20 @@ extension StringExtension on String {
...
+  /// is a complete [RegExp] match, meaning that the regular expression finds a
+  /// match with starting index 0 and ending index [this.length].
+  bool caseInsensitiveEquals(Pattern? pattern) {
+    if (pattern is RegExp) {
+      assert(!pattern.isCaseSensitive);
```
elliette, line 1071: What happens when this assert is  triggered?

elliette (approved):

packages/devtools_app/lib/src/shared/primitives/utils.dart
kenzieschmoll: Right now this would just fail silently but we could catch it in development. This is what we have in caseInsensitiveContains too. Maybe we should provide some validation in the search / filter text fields or add a note when reg exp is enabled.

packages/devtools_app/lib/src/shared/primitives/utils.dart
kenzieschmoll: I wrote iff to mean if and only iff. I will expand to if and only if for clarity.

elliette (approved):

bkonyi (approved):

Issue #6761, opened 2023-11-17
Make the DevTools UI more dense


kenzieschmoll:
Users can use DevTools in the browser or embedded in the IDE. Either way, a more dense UI would help with being able to see more data at once. Our flutter UI is still too "mobile-first" in some areas WRT tap target sizes, row heights, font sizes, etc. 

We've had user's complain that they don't have enough screen real estate to visualize the data they are looking for (for example, wanting to be able to see the live memory chart and look at a memory profile without being constrained by screen space.)

The layout explorer is another example of a UI component that could be made considerably more dense

kenzieschmoll:
- [ ] smaller default font size
- [ ] smaller default table row size
- [ ] smaller table header sizes
- [ ] smaller default padding sizes 
- [ ] smaller default button heights / sizes

One example comparison is the Android Studio profiler vs the DevTools profiler (keep in mind the android studio screenshot is from an embedded IDE tool window and the DevTools screen is full screen on a laptop):

![Screenshot 2023-12-14 at 3 47 41 PM](https://github.com/flutter/devtools/assets/43759233/b03ebc94-8ad3-4eaf-b697-6cebc8b71e3a)

![Screenshot 2023-12-14 at 3 47 47 PM](https://github.com/flutter/devtools/assets/43759233/f5824e35-42f1-42cc-adc4-29c3ce12c947)

PR #7030, opened 2024-01-08
Densify DevTools UI

kenzieschmoll:
Many styling issues were resolved by changing constants in `theme.dart` and applying styling to the default textTheme.

In this PR we set the default font size for `textTheme.bodyMedium`, which is the default font style applied to all `Text` widgets.

Fixes https://github.com/flutter/devtools/issues/6761

Some before (left) and after (right) examples:
Screenshot 2024-01-08 at 3 09 26 PM
Screenshot 2024-01-08 at 3 06 05 PM
Screenshot 2024-01-08 at 3 08 27 PM
Screenshot 2024-01-08 at 3 02 41 PM


kenzieschmoll:
This PR does have release notes, but it appears the release notes check may not be working.

packages/devtools_app/lib/src/framework/notifications_view.dart
```diff
@@ -318,7 +318,7 @@ class _NotificationMessage extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
-    final textStyle = theme.textTheme.bodyMedium;
```
polina-c, line 321: Can you document in our style guide not to use `theme.textTheme.bodyMedium`

elliette (approved):

jacob314:
Looks like a big improvement!
Only nit is the close button on the warning message looks a bit close to the RHS of the window.
You might want to also verify that the tree table horizontal scrolling is still robust when using a variable width font. Perhaps we used a fixed width font to make it easier to exactly compute the maximum width of all rows visible in the table. If that was the case, you should see excessive horizontal width of the table when expanding out a large tree table.

polina-c (approved):
LGTM

packages/devtools_app/lib/src/screens/vm_developer/vm_developer_common_widgets.dart
```diff
@@ -738,9 +737,9 @@ class VmServiceObjectLink extends StatelessWidget {
 
     final TextStyle style;
     if (isServiceObject) {
-      style = theme.fixedFontLinkStyle;
+      style = theme.linkTextStyle;
```
bkonyi, line 740: Are these still using monospace fonts?

kenzieschmoll: No. Chrome DevTools does not use a monospace font for links either:
Screenshot 2024-01-04 at 11 21 42 AM

kenzieschmoll:
> You might want to also verify that the tree table horizontal scrolling is still robust when using a variable width font. Perhaps we used a fixed width font to make it easier to exactly compute the maximum width of all rows visible in the table. If that was the case, you should see excessive horizontal width of the table when expanding out a large tree table.

@jacob314 

You are right. We assumed a monospace font for calculating the width of the variable width columns. Other options:
1. I can use a TextPainter instead to measure the size of the text with a variable width font, but that may hurt performance when we have to do that for every row when the table is resized or nodes are expanded.
2. Continue to use a fixed value as the assumed width of each character. Error on the high side so that we do not overflow, and ellipsize the text in case we do. This will mean there will be some empty space to the right if the user scrolls all the way over.

@Piinks does the new Flutter table you built handle variable width columns well? TL;DR is we used a monospace font so that we could accurately guess the size of a text display (`column.getDisplayValue(node).length * characterWidth`), and now that we are switching to a variable width font, calculating the width that the column needs to be to fix the content is trickier.

Piinks:
> does the new Flutter table you built handle variable width columns well?

If you mean does the column automatically size itself based on the content in a given cell? No. That would mean we would need to lay out every cell for a given column to determine how wide it should be, which would make the table not lazy and have poor performance.

However, there are many ways to define the size of a column or row, all of the subclasses of TableSpanExtent are listed here: https://pub.dev/documentation/two_dimensional_scrollables/latest/table_view/TableSpanExtent-class.html

jacob314:
I think option 3 makes sense if we add back in tooltips or another affordance so there is a way to view the full content without truncation.
Clarified offline: option 3 wouldn't be a truly fixed width but instead fixed width + max indent of any row currently visible.

kenzieschmoll:
This discussed table changes are implemented as a separate PR: https://github.com/flutter/devtools/pull/7033

Issue #6264, opened 2023-08-28
Allow for developers to inspect overall process memory footprint


bkonyi:
The [getProcessMemoryUsage](https://github.com/dart-lang/sdk/blob/main/runtime/vm/service/service.md#getprocessmemoryusage) RPC surfaces a high-level view of major uses of memory known to the VM. Currently, the VM's memory footprint is broken down into buckets including memory usage by isolate groups, profiler and timeline buffers, etc. This can be used to determine the overhead of developer instrumentation and services.

This information is currently available in Observatory through the Process Snapshot tool:

![image](https://github.com/flutter/devtools/assets/24210656/544bde2b-ac5a-4d90-8475-343c4d936c3b)

PR #7032, opened 2024-01-09
Add a Process Memory view to VM Tools

bkonyi:
Displays a breakdown of the target application's overall memory footprint. This allows for developers to determine:

  - The total resident set size (RSS)
  - How much memory can be attributed to each isolate group's heap, including system isolate groups like the vm-service and kernel-service
  - How much memory can be attributed to developer tooling functionality (e.g., CPU profiler buffer size, size of all recorded timeline events)

This view provides both a tree table and a tree map to explore the process's memory footprint.

Fixes https://github.com/flutter/flutter/issues/87816
Fixes https://github.com/flutter/devtools/issues/6264

**Screenshots:**

image

image

### DIFF
--- a/packages/devtools_app/lib/src/screens/vm_developer/process_memory/process_memory_tree_columns.dart
+++ b/packages/devtools_app/lib/src/screens/vm_developer/process_memory/process_memory_tree_columns.dart
@@ -20,11 +20,11 @@ class DescriptionColumn extends ColumnData<TreemapNode> {
   String getValue(TreemapNode dataObject) => dataObject.caption ?? '';
 }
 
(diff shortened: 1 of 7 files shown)


kenzieschmoll:
General comment: at some point, this may be useful to pull out into the memory screen instead of having this tucked away in to the VM developer tools screen.

bkonyi:
> General comment: at some point, this may be useful to pull out into the memory screen instead of having this tucked away in to the VM developer tools screen.

For sure. I didn't want to try and shove it into the memory screen right now since screen real estate is pretty limited. I figured we may as well just get the functionality implemented before trying to fight with the UI.

kenzieschmoll (approved):

PR #7037, opened 2024-01-11
Update to avoid deprecated library in `package:web` v0.4.1

parlough:
\cc @devoncarew @srujzs Since this is breaking the Flutter tree.

### DIFF
--- a/packages/devtools_app/lib/src/extensions/embedded/_controller_web.dart
+++ b/packages/devtools_app/lib/src/extensions/embedded/_controller_web.dart
@@ -8,7 +8,7 @@ import 'dart:ui_web' as ui_web;
 import 'package:devtools_app_shared/utils.dart';
 import 'package:devtools_extensions/api.dart';
 import 'package:path/path.dart' as path;
-import 'package:web/helpers.dart';
+import 'package:web/web.dart';
 
 import '../../shared/development_helpers.dart';
 import '../../shared/globals.dart';
--- a/packages/devtools_app/lib/src/extensions/embedded/_view_web.dart
+++ b/packages/devtools_app/lib/src/extensions/embedded/_view_web.dart
@@ -10,7 +10,7 @@ import 'package:devtools_app_shared/utils.dart';
 import 'package:devtools_extensions/api.dart';
 import 'package:devtools_extensions/utils.dart';
 import 'package:flutter/material.dart';
-import 'package:web/helpers.dart';
+import 'package:web/web.dart';
 
 import '../../shared/banner_messages.dart';
 import '../../shared/common_widgets.dart';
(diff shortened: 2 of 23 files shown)


srujzs (approved):

srujzs:
@devoncarew It's possible we might end up breaking more stuff here - should we release this as a `0.5.0` instead? We can redact and then revert the change in the Dart SDK, but maybe that boat has sailed.

parlough:
Yeah I guess this might cause issues for `flutter/packages` as well, but I'm not sure what their usages look like.

I suppose we should have released `0.4.1` with `package:web/web.dart` exporting the helpers, migrate the Flutter usages, then release `0.4.2` or `0.5.0` deprecating the helpers?

Do you think you want to redact or should I merge this?

QuncCccccc:
Waiting for some advice to unblock the tree🙂

srujzs:
I think a lot of `flutter/packages` use `package:web`, but I think their constraints look more like `>=0.3.0 <0.5.0` so I'm not sure if they'll hit this or not.

For now, let's push this through and fix the breakage first. @kevmoo

devoncarew:
@parlough - thanks for the fix!

I'm inclined to fix this forward in the breakages we see. We do need to be able to message that there will be future API changes via `deprecation` annotations. That's challenging to reconcile w/ CIs that fail w/ new deprecation usages.

Perhaps for package:web we should deprecate in a new major version, and remove in the next? Or, default deprecation usages to not fail analysis w/o a `--fail-on-deprecation` flag to `dart analyze`.

srujzs:
> Perhaps for package:web we should deprecate in a new major version, and remove in the next? Or, default deprecation usages to not fail analysis w/o a --fail-on-deprecation flag to dart analyze.

Going forward that makes sense, but that may lead to excessive versions. Flutter might end up likely using that flag anyways, so it'll still be a matter of fixing when updating versions.

parlough:
I don't think Flutter's testing should prevent `package:web` from evolving, but yeah I think most of Flutter's CI checks don't allow deprecation.

So it'll end up like the SDK which has to fix all deprecations in Flutter before landing the deprecations, or the packages will need tighter constraints.

devoncarew:
> package:web's CI might benefit from the customer testing that the analyzer has?

That would certainly be useful - it would give us advance awareness of the impact of breaking changes.

> So it'll end up like the SDK which has to fix all deprecated usages in Flutter (and siblings) before landing the deprecations

I think that's reasonable, but we wouldn't need to do that on a publish. Flutter uses specific versions of packages, so won't be broken on new deprecation messages (on publish); just when you go to use that specific version in the repo.

I suspect that the devtools repo is not using pubspec.lock files, so is getting the last published compatible version of a package.

parlough:
> I think that's reasonable, but we wouldn't need to do that on a publish. Flutter uses specific versions of packages, so won't be broken on new deprecation messages (on publish); just when you go to use that specific version in the repo.

It looks like `flutter/packages` is blocked now as well since they use `package:web` with a wider constraint and run an analysis that fails on deprecated usages: https://ci.chromium.org/ui/p/flutter/builders/try/Linux%20analyze%20master/5315/overview

PR #7054, opened 2024-01-17
Adding api to get dtd uri from devtools server

CoderDake:
This change will support the devtools server being able to server a dtd uri to devtools when it starts up.
If the uri isn't set then null will be returned.


packages/devtools_app/lib/src/shared/server/_dtd_api.dart
```diff
@@ -0,0 +1,23 @@
+// Copyright 2020 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: 2024

packages/devtools_shared/lib/src/server/server_api.dart
```diff
@@ -10,6 +10,7 @@ import 'dart:io';
 
 import 'package:shelf/shelf.dart' as shelf;
 
+import '../../devtools_shared.dart';
```
kenzieschmoll, line 13: import the file you need directly instead of importing the library that exports others

packages/devtools_shared/lib/src/server/server_api.dart
```diff
@@ -252,7 +254,12 @@ class ServerApi {
           queryParams,
           deeplinkManager,
         );
-
+      case DtdApi.apiGetDtdUri:
+        return api.setCompleted(
+          json.encode({
+            DtdApi.uriPropertyName: dtdUri == null ? null : dtdUri(),
```
kenzieschmoll, line 260: this can be `dtdUri?.call()`

packages/devtools_shared/lib/src/server/server_api.dart
```diff
@@ -36,6 +37,7 @@ class ServerApi {
     required ExtensionsManager extensionsManager,
     required DeeplinkManager deeplinkManager,
     ServerApi? api,
+    String? Function()? dtdUri,
```
kenzieschmoll, line 40: is this a function instead of a String so that it will always be up to date with the latest value on the server?

CoderDake: you are correct :)

kenzieschmoll (approved):

Issue #7066, opened 2024-01-19
App size screen text is too large


kenzieschmoll:
![Screenshot 2024-01-19 at 10 42 07 AM](https://github.com/flutter/devtools/assets/43759233/7ef73a5f-4ba7-4106-9766-c73f2a3f25d0)

PR #7070, opened 2024-01-19
Clean up treemap code by using helper widgets and fix text styling

kenzieschmoll:
fixes https://github.com/flutter/devtools/issues/7066

### DIFF
--- a/packages/devtools_app/lib/src/shared/charts/treemap.dart
+++ b/packages/devtools_app/lib/src/shared/charts/treemap.dart
@@ -369,7 +369,7 @@ class _TreemapState extends State<Treemap> {
       return buildSubTreemaps();
     } else {
       // If constructed with Treemap.fromRoot
-      return buildTreemap(context);
+      return buildTreemap();
     }
   }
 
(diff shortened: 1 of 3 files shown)


elliette (approved):

packages/devtools_app/lib/src/shared/charts/treemap.dart
```diff
@@ -464,71 +474,162 @@ class _TreemapState extends State<Treemap> {
...
+    List<PositionedCell> positionedCells,
+  ) {
+    final x = event.localPosition.dx;
+    final y = event.localPosition.dy;
+    final touchPoint = Offset(x, y);
+    // TODO(peterdjlee): Optimize with more efficient algorithm to find the overlapping cell.
+    //                   Currently O(positionedCells.length) but an optimized algorithm with
+    //                   O(log(positionedCells.length)) is possible.
+    for (final cell in positionedCells) {
+      if (cell.rect.contains(touchPoint)) {
```
bkonyi, line 519: Nit: should we break this loop if we find a match?

bkonyi (approved):

packages/devtools_app/lib/src/shared/charts/treemap.dart
kenzieschmoll: There were no logic changes in this CL, so if there's a logic bug here we should address separately.

Issue #7083, opened 2024-01-22
☔ Migration to `package:unified_analytics` for analytics collection


eliasyishak:
This serves as a tracker issue for the migration efforts for devtools to `package:unified_analytics`

Tasks:
- Setup the analytics instance and integrate the necessary APIs for onboarding a new user
  - [x] PR in `flutter/devtools` to add logic: https://github.com/flutter/devtools/pull/7084
  - [x] PR in `dart-lang/tools` to check legacy file: https://github.com/dart-lang/tools/pull/227
  - [x] PR in `dart-lang/tools` to fix behavior for showing message: https://github.com/dart-lang/tools/pull/228
  - [ ] ~[SWAP WITH PR] https://github.com/flutter/devtools/issues/7230~ tracked separately in https://github.com/flutter/devtools/issues/7230

Migration for events
- [x] Select (selecting tabs)
  - [x] PR in dart-lang/tools to add event: https://github.com/dart-lang/tools/pull/258
  - [x] PR in flutter/devtools to use event: https://github.com/flutter/devtools/pull/7479
- [x] Timing events (https://github.com/flutter/devtools/pull/7676) 
- [x] Screen views (https://github.com/flutter/devtools/pull/7676)
- [x] Impressions (https://github.com/flutter/devtools/pull/7676) 
- [x] Exceptions (https://github.com/flutter/devtools/pull/7676)

PR #7084, opened 2024-01-22
Setting up `package:unified_analytics` within devtools

eliasyishak:
Part of tracker issue:
- https://github.com/flutter/devtools/issues/7083

## Pre-launch Checklist

- [x] I signed the [CLA].

![build.yaml badge]


packages/devtools_shared/lib/src/server/server_api.dart
```diff
@@ -17,6 +18,17 @@ import '../extensions/extension_manager.dart';
...
+// from development constructor too
+final Analytics analytics = Analytics.development(
+  tool: DashTool.devtools,
+  dartVersion: 'TBD',
+  clientIde: 'TBD',
+  enabledFeatures: 'feature1-feature2',
+  flutterChannel: 'TBD',
+  flutterVersion: 'TBD',
+);
+
```
eliasyishak, line 31: Swap this out

packages/devtools_app/lib/src/shared/analytics/analytics_controller.dart
```diff
@@ -43,8 +46,17 @@ class AnalyticsController {
...
+  Future<void> confirmConsentMessageShown() async {
+    if (_confirmConsentMessageShown != null) {
+      await _confirmConsentMessageShown();
+    }
+  }
+
   final VoidCallback? onSetupAnalytics;
 
+  final String? consentMessage;
+
```
eliasyishak, line 59: I made this non-nullable because I didn't want to assume that I can pass empty strings to the constructors within the test files.

Would it be okay if I made this non-nullable and for all tests, I can provide an empty string?

```dart
AnalyticsController _controller =
    AnalyticsController(enabled: false, firstRun: false, consentMessage: '');
```

packages/devtools_app/lib/src/shared/analytics/_analytics_stub.dart
```diff
@@ -32,6 +32,10 @@ Future<void> enableAnalytics() async {}
 
 Future<void> disableAnalytics() async {}
 
+Future<String> getConsentMessage() async => 'stubbed consent message';
```
kenzieschmoll, line 35: nit: https://dart.dev/effective-dart/design#avoid-starting-a-method-name-with-get

packages/devtools_app/lib/src/shared/analytics/prompt.dart
```diff
@@ -89,34 +96,9 @@ class _AnalyticsPromptState extends State<AnalyticsPrompt>
...
-              },
-          ),
-          TextSpan(
-            text: '.',
-            style: textTheme.titleMedium,
-          ),
-        ],
-      ),
+    return SelectableText(
+      controller.consentMessage!,
```
kenzieschmoll, line 100: looks like now we won't be able to have a link to show google's privacy policy. Before we had a link text span for 'privacy policy'. Is this okay?

packages/devtools_app/lib/src/shared/analytics/prompt.dart
```diff
@@ -89,34 +96,9 @@ class _AnalyticsPromptState extends State<AnalyticsPrompt>
...
+      style: textTheme.titleMedium,
```
kenzieschmoll, line 101: instead of titleMedium can we just do `Theme.of(context).regularTextStyle`?

packages/devtools_app/lib/src/shared/server/_analytics_api.dart
```diff
@@ -56,6 +56,32 @@ Future<bool> setAnalyticsEnabled([bool value = true]) async {
...
+  }
+
+  return consentMessage;
+}
+
+/// Confirm with package:unified_analytics that the consent message
+/// has been shown to the user.
+Future<bool> confirmConsentMessageShown() async {
+  if (isDevToolsServerAvailable) {
+    final resp = await request(apiSetConsentMessageAsShown);
```
kenzieschmoll, line 76: I'm a bit confused by the contrasting api name and this method name. Are we setting a value or getting a value? 

Perhaps instead of `apiSetConsentMessageAsShown`, we could use `apiAnalyticsConsentMessageShown` and if that api call has an argument setting the value, then it is a set call, and if it does not, then it is a get call. This is similar to what is done for several boolean service extensions throughout the Flutter framework.

packages/devtools_app/lib/src/shared/server/_analytics_api.dart
```diff
@@ -56,6 +56,32 @@ Future<bool> setAnalyticsEnabled([bool value = true]) async {
   return false;
 }
 
+/// Fetch the consent message for package:unified_analytics.
+Future<String?> getConsentMessage() async {
+  String? consentMessage;
+  if (isDevToolsServerAvailable) {
+    final resp = await request(apiGetConsentMessage);
```
kenzieschmoll, line 63: nit: rename `apiGetConsentMessage` to `apiAnalyticsConsentMessage`? to make it clear this is for analytics

packages/devtools_shared/lib/src/devtools_api.dart
```diff
@@ -14,6 +14,8 @@ const apiResetDevTools = '${apiPrefix}resetDevTools';
 const apiGetDevToolsFirstRun = '${apiPrefix}getDevToolsFirstRun';
 const apiGetDevToolsEnabled = '${apiPrefix}getDevToolsEnabled';
 const apiSetDevToolsEnabled = '${apiPrefix}setDevToolsEnabled';
+const apiGetConsentMessage = '${apiPrefix}getConsentMessage';
+const apiSetConsentMessageAsShown = '${apiPrefix}setConsentMessageAsShown';
```
kenzieschmoll, line 18: nit: maybe separate these and add a new header `/// package:unified_analytics properties APIs`

packages/devtools_shared/lib/src/server/server_api.dart
```diff
@@ -68,23 +80,35 @@ class ServerApi {
...
+        final isFirstRun =
+            _devToolsUsage.isFirstRun || analytics.shouldShowMessage;
         return api.getCompleted(
-          json.encode(_devToolsUsage.isFirstRun),
+          json.encode(isFirstRun),
         );
       case apiGetDevToolsEnabled:
         // Is DevTools Analytics collection enabled?
+        final isEnabled =
+            _devToolsUsage.analyticsEnabled && analytics.telemetryEnabled;
```
kenzieschmoll, line 91: will this mess up existing DevTools users who are opted into analytics for their first run? IIRC `analytics.telemetryEnabled` won't be true for users on their first run no matter what their opt in state is - is that correct?

packages/devtools_app/lib/src/shared/server/_analytics_api.dart
kenzieschmoll: or maybe I'm misunderstanding this - is there a need for getting the value of consent message shown? Or only setting?

packages/devtools_app/lib/src/shared/analytics/analytics_controller.dart
kenzieschmoll: fine to make required. you can just set some bogus value for tests "Test consent message."

packages/devtools_app/lib/src/shared/analytics/prompt.dart
eliasyishak: Yeah we don't have the hyperlink anymore because the text is now getting fetched, but I did wrap the text with the `SelectableText` widget so that users can copy and paste the url still if they wanted to read about the policy.

Is that good enough?

packages/devtools_app/lib/src/shared/server/_analytics_api.dart
eliasyishak: > is there a need for getting the value of consent message shown? Or only setting?

No, we don't need to get any value, this was a good catch because I made stub version of this function return `Future<void>`. I can swap out the web version of the api to be the same

packages/devtools_shared/lib/src/server/server_api.dart
eliasyishak: So there a few different scenarios and if the user has already interacted with `package:unified_analytics` through other Flutter and Dart Tooling.

Each of the scenarios is for the first time a developer interacts with analytics from devtools.

1. If the developer has interacted with `package:unified_analytics` and is opted in --> `analytics.telemetryEnabled` = `true`
2. If the developer has interacted with `package:unified_analytics` and is opted out --> `analytics.telemetryEnabled` = `false`

3. The developer has never interacted with `package:unified_analytics` --> `analytics.telemetryEnabled` = `true`

For situation 3, we set the `telemetryEnabled` value to `true` the first time the tool is run, but this still doesn't allow events to be sent during that run.

> will this mess up existing DevTools users who are opted into analytics

So if they were opted into devtools historically, but they opted out of analytics with `package:unified_analytics` in a different tool, this would now disable events from those users. This is referring to scenario 2 above

packages/devtools_app/lib/src/shared/analytics/prompt.dart
kenzieschmoll: Does the hyperlink come as part of the message? What do you think about using a RegExp to pull the link out of the text and put it in a link text span? I think this would be relatively easy to do.

eliasyishak: yeah we can do that, i can add that in

packages/devtools_app/lib/src/shared/analytics/_analytics_controller_stub.dart
```diff
@@ -7,5 +7,8 @@ import 'dart:async';
 import 'analytics_controller.dart';
 
 FutureOr<AnalyticsController> get devToolsAnalyticsController => _controller;
-AnalyticsController _controller =
-    AnalyticsController(enabled: false, firstRun: false);
+AnalyticsController _controller = AnalyticsController(
+  enabled: false,
+  firstRun: false,
+  consentMessage: 'fake message',
```
kenzieschmoll, line 13: nit: since you implemented `fetchAnalyticsConsentMessage,` in  _analytics_stub.dart, can we return `consentMessage: await ga.fetchAnalyticsConsentMessage()` here?

packages/devtools_app/lib/src/shared/analytics/_analytics_web.dart
```diff
@@ -795,11 +795,11 @@ Future<bool> disableAnalytics() async {
...
 }
 
-Future<String?> getConsentMessage() async {
-  return await server.getConsentMessage();
+Future<String> fetchAnalyticsConsentMessage() async {
+  return await server.fetchAnalyticsConsentMessage();
 }
 
-Future<bool> confirmConsentMessageShown() async {
+Future<void> confirmConsentMessageShown() async {
```
kenzieschmoll, line 802: can you add dart doc on these methods?

kenzieschmoll, line 802: nit: "confirm" makes me think we are asking a question, or in other words that this would return a value of whether the message has been shown. Maybe `markConsentMessageAsShown` would make what this method does more clear?

packages/devtools_app/lib/src/shared/server/_analytics_api.dart
```diff
@@ -71,15 +71,10 @@ Future<String?> getConsentMessage() async {
 
 /// Confirm with package:unified_analytics that the consent message
 /// has been shown to the user.
-Future<bool> confirmConsentMessageShown() async {
+Future<void> confirmConsentMessageShown() async {
```
kenzieschmoll, line 74: same comments here about naming. Maybe `markConsentMessageAsShown` if that is what this method does.

packages/devtools_app/lib/src/shared/server/_analytics_api.dart
```diff
@@ -71,15 +71,10 @@ Future<String?> getConsentMessage() async {
...
   if (isDevToolsServerAvailable) {
-    final resp = await request(apiSetConsentMessageAsShown);
-    if (resp?.statusOk ?? false) {
-      return true;
-    }
+    await request(apiAnalyticsConsentMessageShown);
```
kenzieschmoll, line 76: same with name here. I think I was confused before thinking that this API would also be used to get a value and not just set it. Having the name make it clear that it is only a one-way api (set only) would be good.

packages/devtools_shared/lib/src/server/server_api.dart
```diff
@@ -80,6 +80,10 @@ class ServerApi {
         return api.getCompleted(json.encode(true));
       case apiGetDevToolsFirstRun:
         // Has DevTools been run first time? To bring up analytics dialog.
+        //
+        // Additionally if package:unified_analytics will show a message if it
```
kenzieschmoll, line 84: typo: remove first 'if'

packages/devtools_shared/lib/src/server/server_api.dart
eliasyishak: @kenzieschmoll any recommendations for where I should create the `Analytics` instance? Ideally it can be somewhere it can get imported from or somewhere near the entrypoint so you can direct inject it into where it is needed

kenzieschmoll: How do we get these values from other tools? 

Since this code is on the `devtools_server` where we have access to `dart:io`, one idea is to run `flutter --version` and parse the output. We have a class that parses the output from Flutter Tools' registered `flutterVersion` service here: https://github.com/flutter/devtools/blob/master/packages/devtools_app_shared/lib/src/service/flutter_version.dart/#L16. We could move this class to devtools_shared if it was useful for parsing the output from `flutter --version`.

A couple things we will need to be careful about:
- using the `flutter` that this DevTools server instance is being run from, since a user could have multiple `flutter`'s on their machine / path.
- ensuring these values can be null when `flutter` is not on the user's machine (for example if DevTools was being used from a pure Dart user who only writes Dart CLI apps or Dart web apps.

WRT for where to create the Analytics instance, where do we create the Analytics instance for other tools? Near the root of initialization? Some ideas I have:
- create it in the constructor of the DevToolsServer class: https://github.com/dart-lang/sdk/blob/main/pkg/dds/lib/devtools_server.dart#L23
- make it a late final variable that is created and passed into `defaultHandler` similar to what we do for `clientManager`: https://github.com/dart-lang/sdk/blob/main/pkg/dds/lib/devtools_server.dart#L265-L271

both of these scenarios would require that the `Analytics` instance is passed in as a parameter to `ServerApi.handle` (similar to how `extensionsManager` and `deepLinkManager` are passed in): https://github.com/flutter/devtools/blob/master/packages/devtools_shared/lib/src/server/server_api.dart/#L37

It is also worth noting that `ServerApi.handle` is called from google3 on a separate path, so we'd have to make sure we create the `Analytics` instance properly there too.

kenzieschmoll (commented):
I had another thought WRT opt-in logic. Since we are going to continue to send events to the old analytics ID and through the new unified_analytics package, should we keep the opt in states separate for those? So old analytics would check out existing logic and new analytics would use the unified_analytics opt in logic? This would mean we'd need different `apiGetDevToolsEnabled` and `apiSetDevToolsEnabled` endpoints for both legacy analytics and new analytics.

eliasyishak:
hmm so are you suggesting we would have to perform 2 separate actions if we wanted to opt out of analytics completely as a devtools user? Or is this more of a suggestion to have another `case` clause in the `server_api.dart`?

Having to perform 2 separate actions to opt out may confuse some of our users

kenzieschmoll:
Discussed offline. Defaulting to unified_analytics opt-in state for both legacy and UA when they diverge sounds sgtm (what is already done for other tools).

packages/devtools_app/lib/src/shared/analytics/prompt.dart
eliasyishak: image

I was able to swap out the text and user a hyperlink

packages/devtools_app/test/shared/analytics_prompt_test.dart
```diff
@@ -11,13 +11,15 @@ import 'package:devtools_test/helpers.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:provider/provider.dart';
+import 'package:unified_analytics/src/constants.dart' as unified_analytics;
```
eliasyishak, line 14: @kenzieschmoll is this okay? We aren't explicitly adding the package into the pubspec for `devtools_app` but I was wondering if it would be okay i pulled in the constant like this?

packages/devtools_shared/lib/src/server/server_api.dart
eliasyishak: > using the flutter that this DevTools server instance is being run from, since a user could have multiple flutter's on their machine / path.

Yeah this was the first thing I was concerned about, since devtools also downloads its own separate flutter sdk in `/devtools/tool/flutter-sdk`, wouldn't running `flutter --version --machine` return versioning information for `flutter` that is on the path?

For example, the flutter sdk that is in the devtools directory is version `"frameworkVersion": "3.20.0-0.0.pre",` while my flutter in path resolves to `"frameworkVersion": "3.20.0-2.0.pre.9",`... which of these values would we want to report for analytics?

packages/devtools_app/test/shared/analytics_prompt_test.dart
kenzieschmoll: Can we add this as a dependency in the pubspec? Won't we need to access the constants anyway as we migrate all of our devtools events to package:unified_analytics?

packages/devtools_shared/lib/src/server/server_api.dart
kenzieschmoll: The `tool/flutter-sdk` Fluter isn't used in production. This is just for all of our repository management scripts as well as for the workflows we run on the CI.

packages/devtools_app/test/shared/analytics_prompt_test.dart
eliasyishak: oh right! i'll add it in

packages/devtools_shared/lib/src/server/server_api.dart
```diff
@@ -70,23 +72,39 @@ class ServerApi {
...
+        final isFirstRun =
+            _devToolsUsage.isFirstRun || analytics.shouldShowMessage;
         return api.getCompleted(
-          json.encode(_devToolsUsage.isFirstRun),
+          json.encode(isFirstRun),
         );
       case apiGetDevToolsEnabled:
         // Is DevTools Analytics collection enabled?
+        final isEnabled =
+            _devToolsUsage.analyticsEnabled && analytics.telemetryEnabled;
```
kenzieschmoll, line 87: when these diverge, should we be setting the value of `_devtoolsUsage.analyticsEnabled` to `analytics.telemetryEnabled`?

eliasyishak: I think once we have fully migrated, we could probably get rid of `_devtoolsUsage` all together since the package will check itself when an event is attempted to be sent. Example below of how we do it in the flutter-tool by just calling the send method without any wrapping class

https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/runner.dart#L190

packages/devtools_app/lib/src/shared/analytics/_analytics_web.dart
```diff
@@ -795,6 +795,19 @@ Future<bool> disableAnalytics() async {
...
 
+/// Fetch the legal consent message for telemetry collection for
+/// package:unified_analyitcs from the server
+Future<String> fetchAnalyticsConsentMessage() async {
+  return await server.fetchAnalyticsConsentMessage();
+}
+
+/// Communicates with the server to confirm with package:unified_analyitcs
+/// that the consent message has successfully been shown and to allow events
+/// to be recorded if the user has decided to remain opted in.
```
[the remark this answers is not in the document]
eliasyishak: So I made the decision to automatically mark the message as shown when we show the text. If the user was previously opted out through other dash tools, then they will remain opted out. If they press the accept button, we will change the their telemetry status to opted in for all dash tools.

But in both cases, marking the message as shown just tells the package to add it to the configuration file and log the date the message was shown

packages/devtools_app/lib/src/shared/analytics/prompt.dart
```diff
@@ -89,30 +96,46 @@ class _AnalyticsPromptState extends State<AnalyticsPrompt>
...
       text: TextSpan(
         children: [
           TextSpan(
-            text: 'DevTools reports feature usage statistics and basic '
-                'crash reports to Google in order to help Google improve '
-                'the tool over time. See Google\'s ',
+            text: consentMessageRegExpResults[0],
             style: textTheme.titleMedium,
           ),
           TextSpan(
```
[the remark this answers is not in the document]
eliasyishak: Sounds good, i made the fields `Link.gaScreenName` and `Link.gaSelectedItemDescription` nullable. I also added conditionals for where we call `ga.select()` with those two fields to not send anything if they are null

kenzieschmoll (approved):

PR #7085, opened 2024-01-22
Create `CliCommand.dart` that uses the Dart executable from flutter

kenzieschmoll:
Fixes https://github.com/Dart-Code/Dart-Code/issues/4953

PR #7093, opened 2024-01-23
Remove splitting by spaces from process/command helpers

DanTup:
Accepting strings and splitting by spaces is unsafe if paths or arguments could contain spaces. Accepting strings makes it really easy to make this mistake.

I propose we stop taking strings in this way and if we want to use strings for convenience we force calling `.split(' ')` at the call site (on the literal string). This makes it easier to spot these kinds of bugs... if you are calling `.split(' ')` on a string that contains a variable, you might have a bug. It's safe to call it on a simple literal string where you can easily see that it can't split incorrectly.

An example of a previous bug because of splitting args is https://github.com/flutter/devtools/issues/6679 and a recent change that accidentally adds this elsewhere is https://github.com/flutter/devtools/pull/7085#pullrequestreview-1838567159

@kenzieschmoll thoughts?

### DIFF
--- a/tool/lib/commands/analyze.dart
+++ b/tool/lib/commands/analyze.dart
@@ -38,7 +38,7 @@ class AnalyzeCommand extends Command {
 
       final process = await processManager.runProcess(
         CliCommand.dart(
-          'analyze --fatal-infos',
+          'analyze --fatal-infos'.split(' '),
           // Run all so we can see the full set of results instead of stopping
           // on the first error.
           throwOnException: false,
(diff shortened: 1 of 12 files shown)


kenzieschmoll (commented):
instead of using `.split()` calls on Strings to pass to the commands, I'm in favor of using an explicit list of Strings everywhere.

DanTup:
> I'm in favor of using an explicit list of Strings everywhere.

SGTM, I've made this change and removed the remaining constructor that was splitting a command.

kenzieschmoll (approved):
LGTM thanks for the cleanup

Issue #7115, opened 2024-01-25
serveAvailableExtensions returns 500 Internal Server Error


DanTup:
Looking into https://github.com/Dart-Code/Dart-Code/issues/4952, I'm not seeing the extensions show up in the menu:

![image](https://github.com/Dart-Code/Dart-Code/assets/1078012/f3e31d10-adeb-4282-a6ed-c7546a8e7257)

Seems like there's a web request failing with a 500:

![Screenshot 2024-01-25 155149](https://github.com/flutter/devtools/assets/1078012/a61178c6-1029-400f-97e6-ab4612fc4cf2)

Not sure if it's due to a bad version of a package or something, or maybe a Windows-specific bug. Will try to find more about the issue (doesn't seem like there's anything useful in the http response).

DanTup:
Ah, when I open in the browser the request has a file URI in rootPath:

```
http://127.0.0.1:9100/api/serveAvailableExtensions?rootPath=file%3A%2F%2F%2FC%3A%2FDev%2FTest%2520Projects%2Fflutter_counter
```

But inside VS Code it appears to just be a file path...

DanTup:
@kenzieschmoll I'm looking through the code and there are a lot of things called "rootPath" and in some cases they definitely seem to be URIs. It would be good to add some comments on these (or rename them where non-breaking) where the names are misleading.

VS Code sends the file path in the session information, but it seems like it becomes a URI at:

https://github.com/flutter/devtools/blob/65adbe2e94c478778a5fcc4fe569f7815ef2c0e3/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart#L157

So it's not yet clear to me how we're ending up with just the file path in this web request, but I'll keep digging.

DanTup:
Ok, here is the problem..

```dart
print('init extensions with sessionRootPath=$sessionRootPath');
final fileUri = Uri.file(sessionRootPath);
print('file URI is $fileUri');
```

logs:

```
init extensions with sessionRootPath=C:\Dev\Test Projects\flutter_counter
file URI is C%3A%5CDev%5CTest%20Projects%5Cflutter_counter
```

It appears it is unsafe for DevTools to use `Uri.file()` in code that runs in the browser with a Windows path. The resulting URI is not a `file://` URI at all...

😬

This can be preproduced in DartPad:

```
  final path = r"C:\foo\bar.txt";
  final uri = Uri.file(path);
  print('scheme: ${uri.scheme}');
[...]

PR #7116, opened 2024-01-25
Fix extensions not showing in sidebar on Windows

DanTup:
+ remove "Path" from variables that are URIs and change them from Strings to Uris to avoid confusion.

Fixes https://github.com/flutter/devtools/issues/7115

Unfortunately despite the change to the tests, they don't currently catch the bug because when running in the VM `Uri.file()` does the expected thing on Windows. To cover this with tests, we really need to have some real integration tests testing through the browser (@kenzieschmoll do we already have the ability to do this? if not and it's non trivial, I think we should land this fix first).

Note: I have assumed that renaming variables and changing `String` ->`Uri`s is non-breaking for `devtools_app` but let me know if that's not the case and I'll revert those and document this in comments. It would be nice to change though, variables named "Path" that are `Strings` but are actually URIs is really confusing.

### DIFF
--- a/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
@@ -157,12 +157,11 @@ class _DevToolsMenuState extends State<_DevToolsMenu> {
       // This file path might be a Windows path but because this code runs in
       // the web, Uri.file() will not handle it correctly.
       //
-      // Since all paths are absolute, assume that if the path does not start
-      // with '/' that it's Windows, and parse with Windows semantics.
-      final fileUri = Uri.file(
-        sessionRootPath,
-        windows: !sessionRootPath.startsWith('/'),
-      );
+      // Since all paths are absolute, assume that if the path contains `\` and
+      // not `/` then it's Windows.
+      final isWindows =
+          sessionRootPath.contains(r'\') && !sessionRootPath.contains(r'/');
+      final fileUri = Uri.file(sessionRootPath, windows: isWindows);
       assert(fileUri.isScheme('file'));
       _extensionServiceForSession = ExtensionService(fixedAppRoot: fileUri);
       unawaited(_extensionServiceForSession!.initialize());
--- a/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
@@ -154,9 +154,17 @@ class _DevToolsMenuState extends State<_DevToolsMenu> {
   void _initExtensions() {
     final sessionRootPath = widget.session.projectRootPath;
     if (sessionRootPath != null) {
-      final fileUri = Uri.file(sessionRootPath);
-      _extensionServiceForSession =
-          ExtensionService(fixedAppRootPath: fileUri.toString());
+      // This file path might be a Windows path but because this code runs in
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -154,9 +154,17 @@ class _DevToolsMenuState extends State<_DevToolsMenu> {
...
-          ExtensionService(fixedAppRootPath: fileUri.toString());
+      // This file path might be a Windows path but because this code runs in
+      // the web, Uri.file() will not handle it correctly.
+      //
+      // Since all paths are absolute, assume that if the path does not start
+      // with '/' that it's Windows, and parse with Windows semantics.
+      final fileUri = Uri.file(
+        sessionRootPath,
+        windows: !sessionRootPath.startsWith('/'),
+      );
```
DanTup, line 165: This is the actual fix.. There are a few other ways we could do this:

```dart
Uri.file(x, windows: !x.startsWith('/')) // assume if doesn't start with / it's Windows
Uri.file(x, windows: userAgent.contains('win')) // assume Windows if browser is Windows
Uri.file(x, windows: x.startsWith('\[a-zA-Z](?::|%3a|%3A)]')) // assume Windows path starts with drive letter and then colon (or encoded colon)
```

In hindsight, it would've been better if the tooling API was passing file URIs in the protocol.. that's more of a breaking change though. We could add it as a new field and use that in preference with the goal of dropping the "path" version (far into the) future.. lmk if you think that's worthwhile.

DanTup:
Added release-notes-not-required because the previous work to include the extensions has never shipped, so the bug and fix will ship together in the same version.

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
kenzieschmoll: > Uri.file(x, windows: userAgent.contains('win')) // assume Windows if browser is Windows

I actually like this one. I would worry that somehow we'd be missing a leading slash on a non-windows uri and then parse it incorrectly as windows. Another option is to OR these checks if we don't trust the robustness of any individually.

> In hindsight, it would've been better if the tooling API was passing file URIs in the protocol

What about making this the default when we switch from using postMessage to using the DTD to communicate between VS code and DevTools?

DanTup: > > ```
> > Uri.file(x, windows: userAgent.contains('win')) // assume Windows if browser is Windows
> > ```
> 
> I actually like this one. 

My concern doing only this is that "win" is quite short and who knows what other things might be in peoples user agents. Maybe for now we should do `userAgent.contains('win') && !sessionRootPath.startsWith('/')`. I'll verify the user agent when embedded in VS Code and push that unless you'd prefer something else.

> What about making this the default when we switch from using postMessage to using the DTD to communicate between VS code and DevTools?

Sounds like a plan to me. If we can standardise on URIs everywhere (and ensuring DevTools in the browser never gets paths so it doesn't need to use `Uri.file()` or `Uri.toFilePath()`) it should avoid a number of gotchas here.

And ofc, we should work on getting Windows bots running. I suspect a lot of the shell scripts can be converted to devtools_tool. but if there are bits that need to be Windows shell scripts (to bootstrap setting up `devtools_tool`) I'm happy to work on them.

kenzieschmoll (approved):

Issue #7099, opened 2024-01-24
debugger_evaluation_test is flaky


whesse:
The test "returns privates only from library" in
packages/devtools_app/test/debugger/debugger_evaluation_test.dart
is failing flakily on the builder https://ci.chromium.org/ui/p/dart/builders/ci.sandbox/flutter-linux-customer-testing

It is only this test that is failing, always with the same stack trace, and it is happening more than 10% of the time.
The stack trace says that the test is failing and always has the message: "This test failed after it had already completed.
|   Make sure to use a matching library which informs the test runner
|   of pending async work."

This flakiness is affecting our CI, so if the problem with the test can't be fixed easily, it would be good to disable it until it can be fixed.

The full stack trace from the logs is:
```
02:21 +566 ~8 -1: /b/s/w/ir/x/t/flutter_customer_testing.flutter_devtools.MFWIAI/tests/packages/devtools_app/test/debugger/debugger_evaluation_test.dart: EvalOnDartLibrary returns privates only from library [E]
|   [Sentinel kind: Collected, valueAsString: <collected>] from getObject()
|   package:vm_service/src/vm_service.dart 265:34                              new _OutstandingRequest
|   package:vm_service/src/vm_service.dart 1771:25                             VmService._call.<fn>
|   package:vm_service/src/vm_service.dart 1783:8                              VmService._call
|   package:vm_service/src/vm_service.dart 1003:7                              VmService.getObject
|   package:devtools_app/src/service/vm_service_wrapper.dart 164:18            VmServiceWrapper.getObject
|   package:devtools_app/src/shared/scripts/script_manager.dart 125:10         _ScriptCache.getScript
|   package:devtools_app/src/shared/scripts/script_manager.dart 77:25          ScriptManager.getScript
|   package:devtools_app/src/screens/debugger/debugger_controller.dart 518:40  DebuggerController._createStackFrameWithLocation
|   dart:async                                                                 Future.wait
|   package:devtools_app/src/screens/debugger/debugger_controller.dart 461:20  DebuggerController._getStackInfo
|   ===== asynchronous gap ===========================
|   dart:async                                                                 _Completer.completeError
|   package:vm_service/src/vm_service.dart 272:18                              _OutstandingRequest.completeError
|   package:vm_service/src/vm_service.dart 1860:17                             VmService._processResponse
|   package:vm_service/src/vm_service.dart 1840:9                              VmService._processMessageStr
[...]

PR #7134, opened 2024-01-30
Re-enable `debugger_evaluation_test`

elliette:
Hopefully fixes https://github.com/flutter/devtools/issues/7099

This test case was flaking with the following error: 
```
|   This test failed after it had already completed.
|   Make sure to use a matching library which informs the test runner
|   of pending async work.
```

Only difference between it and the other test cases in the file is that it was using a non-standard matcher (`collectionEquals` which is defined in DevTools) instead of a standard matcher like `equals`. 

I've switched to using `unorderedEquals` instead. I can't reproduce the flake locally to verify that it's fixed unfortunately.

### DIFF
--- a/packages/devtools_app/test/debugger/debugger_evaluation_test.dart
+++ b/packages/devtools_app/test/debugger/debugger_evaluation_test.dart
@@ -169,13 +169,11 @@ void main() {
       );
       test(
         'returns privates only from library',
-        skip: true,
         () async {
           await runMethodAndWaitForPause(
             'AnotherClass().pauseWithScopedVariablesMethod()',
           );
           expect(
-            collectionEquals(
               await autoCompleteResultsFor(
                 EditingParts(
                   activeWord: '_',
@@ -184,14 +182,13 @@ void main() {
                 ),
                 evalService,
               ),
+            unorderedEquals(
               [
                 '_privateField2',
                 '_privateField1',
                 '_PrivateClass',
               ],
-              ordered: false,
             ),
-            isTrue,
           );
         },
         timeout: const Timeout.factor(8),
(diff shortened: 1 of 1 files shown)


packages/devtools_app/test/debugger/debugger_evaluation_test.dart
```diff
@@ -184,14 +182,13 @@ void main() {
                 ),
                 evalService,
               ),
+            unorderedEquals(
```
kenzieschmoll, line 185: formatting

kenzieschmoll (approved):

kenzieschmoll:
lgtm. Though it seems like this flake was due to some async work not being awaited in the test. Perhaps we need to instead use `expectAsync` or `tester.runAsync` to make sure we are awaiting the async work properly?

whesse:
When this lands, the pinned devtools tests in flutter customer testing should be updated to the fixed version at 
registry/flutter_devtools.test

I haven't yet updated it to the version that skips the test, so the skip hasn't actually happened on our CI yet, and it never will, if the pin advances to this revert instead.

kenzieschmoll (approved):

elliette:
> lgtm. Though it seems like this flake was due to some async work not being awaited in the test. Perhaps we need to instead use `expectAsync` or `tester.runAsync` to make sure we are awaiting the async work properly?

SG, switched to `expectLater`

elliette:
> When this lands, the pinned devtools tests in flutter customer testing should be updated to the fixed version at https://github.com/flutter/tests/blob/main/registry/flutter_devtools.test#L6
> 
> I haven't yet updated it to the version that skips the test, so the skip hasn't actually happened on our CI yet, and it never will, if the pin advances to this revert instead.

SG, will update it!

Issue #7114, opened 2024-01-25
DevTools makes lots of requests for release notes that are 404s


DanTup:
When I have the network tools open I see DevTools make a lot of requests for release notes that are 404s:

![image](https://github.com/flutter/devtools/assets/1078012/4e75b9a6-cbbd-4a60-8957-81adff4107e9)

I thought this might have been known and just how this works, but @jacob314 was surprised by it and suggested I file an issue.

parlough:
Without looking at the code, I'm assuming it keeps looking until it finds the most recent release notes on the website, since not all versions have one or at least not right away. Stable versions usually have a corresponding entry, so I imagine it doesn't happen too much for developers on stable.

If it's helpful, the site build could generate(and host) a file with information about what release notes are available?

kenzieschmoll:
> If it's helpful, maybe the site build could generate(and host) a file with information about what release notes are available?

This would be helpful. Then DevTools could just request this file and do the logic on our side to find the most recent version of release notes to show instead of relying on http requests in a try catch to perform this logic.

PR #7141, opened 2024-01-31
Adjust release notes retrieval and error handling

parlough:
Corresponds to https://github.com/flutter/website/pull/10099.

Before I land that, I'd like to get this reviewed to verify it's the right direction.

In switching to use the new index file I made some other changes:

- Restructured to not rely on try-catch for every state of release notes missing
- Handled the staging/debug case completely separately to keep the code a bit easier to follow. It's all tree-shaken cleanly in production builds.
- Avoids a variety of Uri and semantic version parsing and calculations.

Fixes https://github.com/flutter/devtools/issues/7114


packages/devtools_app/lib/src/framework/release_notes/release_notes.dart
```diff
@@ -23,11 +24,11 @@ bool debugTestReleaseNotes = false;
 // from the flutter/website PR, which has a GitHub action that automatically
 // stages commits to firebase. Example:
 // https://flutter-docs-prod--pr8928-dt-notes-links-b0b33er1.web.app/tools/devtools/release-notes/release-notes-2.24.0-src.md.
-String? _debugReleaseNotesUrl;
+const String? _debugReleaseNotesUrl = null;
 
-const _flutterDocsSite = 'https://docs.flutter.dev';
-
-const releaseNotesKey = Key('release_notes');
+const String _unsupportedPathSyntax = '{{site.url}}';
```
kenzieschmoll, line 29: nit: replace String with final

packages/devtools_app/lib/src/framework/release_notes/release_notes.dart
```diff
@@ -76,66 +75,121 @@ class ReleaseNotesController extends SidePanelController {
   Future<void> _fetchAndShowReleaseNotes({
     SemanticVersion? versionFloor,
   }) async {
+    if (_debugReleaseNotesUrl case final debugUrl?) {
+      // Specially handle the case where a debug release notes URL is specified.
+      final debugUri = Uri.parse(debugUrl);
+      final releaseNotesMarkdown = await http.read(
+        debugUri,
```
kenzieschmoll, line 82: style nit: you can remove trailing comma from dart code as long as this fits on one line

packages/devtools_app/lib/src/framework/release_notes/release_notes.dart
parlough: This variable is already `const`, do you just want me to rely on type inference here?

packages/devtools_app/lib/src/framework/release_notes/release_notes.dart
```diff
@@ -76,66 +75,121 @@ class ReleaseNotesController extends SidePanelController {
...
+    if (releases is! Map<String, Object?>) {
+      _emptyAndClose(
+        'The DevTools release index file was incorrectly formatted.',
+      );
+      return;
+    }
+
+    // If the version floor has the same major and minor version,
+    // don't check below its patch version.
+    final int minimumPatch;
```
kenzieschmoll, line 132: does this need to be late final?

parlough: Nope! Flow analysis is great these days :D

packages/devtools_app/lib/src/framework/release_notes/release_notes.dart
```diff
@@ -76,66 +75,121 @@ class ReleaseNotesController extends SidePanelController {
...
+    final Map<String, Object?> releaseIndex;
     try {
-      // Try all patch versions for this major.minor combination until we find
-      // release notes (e.g. 2.11.4 -> 2.11.3 -> 2.11.2 -> ...).
-      final attemptedVersions = <String>[];
-      var attempts = notesVersion.patch;
-      while (attempts >= 0 && notesVersion > versionFloor) {
-        final versionString = notesVersion.toString();
+      final releaseIndexUrl =
+          _flutterDocsSite.replace(path: '/f/devtools-releases.json');
```
kenzieschmoll, line 114: put "'/f/devtools-releases.json'" into a named const at top of file with _flutterDocsSite

packages/devtools_app/lib/src/framework/release_notes/release_notes.dart
```diff
@@ -76,66 +75,121 @@ class ReleaseNotesController extends SidePanelController {
...
+    // Try each patch version in this major.minor combination until we find
+    // release notes (e.g. 2.11.4 -> 2.11.3 -> 2.11.2 -> ...).
+    while (patchToCheck >= minimumPatch) {
+      final releaseToCheck = '$majorMinor.$patchToCheck';
+      if (releases[releaseToCheck] case final String releaseNotePath) {
+        final String releaseNotesMarkdown;
         try {
-          String releaseNotesMarkdown = await http.read(
-            Uri.parse(_debugReleaseNotesUrl ?? _releaseNotesUrl(versionString)),
+          releaseNotesMarkdown = await http.read(
```
kenzieschmoll, line 150: will this still print errors to the command line? When I've looked at this before, I could never get the try / catch to swallow the exception that is printed to the chrome console

parlough: Yeah it does, but now this is handling an error case rather than part of the core discovery behavior, so it should never happen.

I'm assuming Chrome DevTools always adds that, but I'll see if I can prevent it somehow.

kenzieschmoll: Since `releaseToCheck` has to be a key in the `releases` map, this shouldn't fail though right? This would really only catch the case where we had a broken url for the `releaseToCheck` entry on the flutter.dev lookup

packages/devtools_app/lib/src/framework/release_notes/release_notes.dart
```diff
@@ -23,11 +24,12 @@ bool debugTestReleaseNotes = false;
...
-String? _debugReleaseNotesUrl;
+const String? _debugReleaseNotesUrl = null;
 
-const _flutterDocsSite = 'https://docs.flutter.dev';
-
-const releaseNotesKey = Key('release_notes');
+const Key releaseNotesKey = Key('release_notes');
+const String _unsupportedPathSyntax = '{{site.url}}';
+const String _releaseNotesPath = '/f/devtools-releases.json';
+final Uri _flutterDocsSite = Uri.https('docs.flutter.dev');
```
kenzieschmoll, line 32: devtools style: remove types on left hand side

kenzieschmoll (approved):
a couple more comments, but lgtm. Thank you for fixing this!

packages/devtools_app/lib/src/framework/release_notes/release_notes.dart
parlough: I was thinking more about caching or strange CDN issues/delays as the broken URL would be caught by the website's CI currently. Or maybe if the user lost connection momentarily. So you're right, in 99.99% of the cases, it shouldn't fail.

I added a comment to expand on when it could somehow fail and added a log message with the version number and the link tried so it's easier to identify what caused it to fail, if someone ever reports it.

parlough:
Thanks for the review @kenzieschmoll! I made adjustments according to your remaining comments.

I'm not familiar with Flutter app testing, but if there's anything specific you'd like me to account for, I can figure something out :)

kenzieschmoll:
Can you add testing to `test/shared/release_notes_test.dart`? Would be good to check that we can read the index file from the flutter website and parse it as expected. This would catch changes where changes to that file unexpectedly break DevTools.

kenzieschmoll:
Hi @parlough is this still on your radar? Need any help with adding the test?

parlough:
@kenzieschmoll Sorry, it's been a busy week! I'll try to get back to this today and let you know if I have any questions :)

packages/devtools_app/lib/src/framework/release_notes/release_notes.dart
```diff
@@ -52,6 +52,10 @@ class ReleaseNotesController extends SidePanelController {
     _init();
   }
 
+  @visibleForTesting
+  static Uri get releaseIndexUrl =>
```
kenzieschmoll, line 56: sytle-nit: remove type on LHS

kenzieschmoll (approved):

packages/devtools_app/lib/src/framework/release_notes/release_notes.dart
parlough: Since this is a getter, doing so will trigger [`always_declare_return_types`](https://dart.dev/tools/linter-rules/always_declare_return_types).

parlough:
Thanks for the review and feedback :D

Issue #5874, opened 2023-06-02
Support eval-and-browse on web platform


annagrin:
Currently eval-and-browse feature (https://github.com/flutter/devtools/issues/4962) is not fully supported on web platform due to dwds API limitations (not allowing scope in `VmServiceInterface.evaluateInFrame`)

Work to allow scope passing is in progress in dwds: https://github.com/dart-lang/webdev/pull/2122

After the work is done and a new dwds version is released, we need to remove the following check from DevTools: https://github.com/flutter/devtools/blob/260c266576eb7bb1b39ff2df026bd3ac85a7efe6/packages/devtools_app/lib/src/shared/console/eval/eval_service.dart#L190

annagrin:
Update: dwds work is done in https://github.com/dart-lang/webdev/pull/2122

Should be included in dwds version 19.0.1 (to be released)

elliette:
Ideally when we enable this for web it would be good to have a test to catch any regressions. The `eval_and_browse_test` currently cannot be enabled for web, because it is using the memory panel which is not supported for web devices: https://github.com/flutter/devtools/blob/2a106f3bda53d6925a55599060a4cd7b29933229/packages/devtools_app/integration_test/test_infra/run/_test_app_driver.dart#L428

polina-c:
I think you need to create different test like 'eval_web_test'. You can start with copying `eval_and_browse_test`.
The test first checks basic console functionality and then switches to memory screen.
So, you need to remove the second part and configure the bot to test the feature for web, not for headless.

Makes sense?

polina-c:
It will be just `eval`. `browse` part will not be available for web, because there is no VM, that is needed for this.

elliette:
@polina-c Would you be willing to add the test? I don't have much context on this feature since I didn't write it. Thanks!

annagrin:
We need to enable this for our customers, I'll give it a try.

PR #6190, opened 2023-08-15
Enable eval-and-browse feature for web

annagrin:
*What changed*

- Remove restrictions on using the feature for web platform now that dwds supports evaluation with scope.

- Add tests for eval-and-browse in the inspector panel.

*Issues are fixed by this PR.*

Closes: https://github.com/flutter/devtools/issues/5874

*Please add a note to `packages/devtools_app/release_notes/NEXT_RELEASE_NOTES.md` if your change requires release notes. Otherwise, add the 'release-notes-not-required' label to the PR.*

## Pre-launch Checklist

- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ ] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [ ] I signed the [CLA].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or there is a reason for not adding tests.

![build.yaml badge]

If you need help, consider asking for help on [Discord].

[Contributor Guide]: https://github.com/flutter/devtools/blob/master/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[CLA]: https://cla.developers.google.com/
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[build.yaml badge]: https://github.com/flutter/devtools/actions/workflows/build.yaml/badge.svg

PR #7144, opened 2024-01-31
Enable eval with scope for web

elliette:
Fixes https://github.com/flutter/devtools/issues/5874

Split off from https://github.com/flutter/devtools/pull/6190

Allows web users to evaluate inspected widgets in the console:

Jan-31-2024 11-57-42

And to use scope for any expression evaluation:

Jan-31-2024 12-03-52


packages/devtools_app/integration_test/test_infra/run/_test_app_driver.dart
```diff
@@ -430,15 +430,18 @@ enum TestAppDevice {
   static final _unsupportedTestsForDevice = <TestAppDevice, List<String>>{
     TestAppDevice.flutterTester: [],
     TestAppDevice.flutterChrome: [
-      // TODO(https://github.com/flutter/devtools/issues/5874): Remove once supported on web.
       'eval_and_browse_test.dart',
```
kenzieschmoll, line 433: should we remove this now?

kenzieschmoll (approved):
lgtm with one question

packages/devtools_app/integration_test/test_infra/run/_test_app_driver.dart
elliette: The `eval_and_browse_test` isn't supported on web because it relies on the memory screen (which isn't supported for web). The new test `eval_and_inspect_test` tests for similar functionality on the inspector screen.

PR #7162, opened 2024-02-05
Add a flag to `devtools_tool serve` to support debugging the devtools server

kenzieschmoll:
CC @hangyujin this PR combined with https://dart-review.googlesource.com/c/sdk/+/350360 will unblock profiling the deep links code that runs on the devtools server.

### DIFF
--- a/tool/lib/commands/serve.dart
+++ b/tool/lib/commands/serve.dart
@@ -15,6 +15,7 @@ import 'shared.dart';
 const _buildAppFlag = 'build-app';
 const _machineFlag = 'machine';
 const _allowEmbeddingFlag = 'allow-embedding';
+const _debugServerFlag = 'debug-server';
 
 /// This command builds DevTools in release mode by running the
 /// `devtools_tool build` command and then serves DevTools with a locally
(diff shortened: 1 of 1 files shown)


tool/lib/commands/serve.dart
```diff
@@ -163,6 +176,10 @@ class ServeCommand extends Command {
     await processManager.runProcess(
       CliCommand.dart(
         [
+          if (debugServer) ...[
+            'run',
+            '--observe',
```
bkonyi, line 181: This will fail if another process is already bound to port 8181. Do you want to provide a way to override the port or just bind to a random available port?

kenzieschmoll: binding to random available port is fine.

tool/lib/commands/serve.dart
```diff
@@ -26,6 +27,12 @@ const _allowEmbeddingFlag = 'allow-embedding';
 /// to this command. All of the following commands are passed along to the
 /// `devtools_tool build` command.
 ///
+/// If the [_debugServerFlag] argument is present, the DevTools server will be
+/// started with the `--observe` flag. This will allow you to debug and profile
+/// the server with a local VM service connection. By default, this will set
+/// `--pause-osolates-on-start` and `--pause-isolates-on-unhandled-exception`
```
bkonyi, line 33: `--pause-isolates-on-start` and `--pause-isolates-on-unhandled-exceptions`.

bkonyi (approved):

PR #7164, opened 2024-02-05
Change dtdUri handle parameter to be a string

CoderDake:
For now we will only ever get one dtdUri so we don't need the ability to get a fresh version in the future.

This PR changesthe devtools server handler so it just takes a String instead of a callback for that same string .

### DIFF
--- a/packages/devtools_shared/lib/src/server/server_api.dart
+++ b/packages/devtools_shared/lib/src/server/server_api.dart
@@ -255,9 +255,7 @@ class ServerApi {
         );
       case DtdApi.apiGetDtdUri:
         return api.setCompleted(
-          json.encode({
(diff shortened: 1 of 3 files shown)


packages/devtools_shared/lib/src/server/server_api.dart
```diff
@@ -256,7 +256,7 @@ class ServerApi {
       case DtdApi.apiGetDtdUri:
         return api.setCompleted(
           json.encode({
-            DtdApi.uriPropertyName: dtdUri?.call(),
+            DtdApi.uriPropertyName: dtdUri,
```
kenzieschmoll, line 259: probably can remove trailing comma now if it fits on one line

kenzieschmoll (approved):

PR #7174, opened 2024-02-06
Fix regular expression that is responsible for detecting app root

kenzieschmoll:
@DanTup can you confirm that should work as expected for Windows paths?

### DIFF
--- a/packages/devtools_app/lib/src/extensions/extension_service.dart
+++ b/packages/devtools_app/lib/src/extensions/extension_service.dart
@@ -182,19 +182,21 @@ class ExtensionService extends DisposableController
 // TODO(kenz): consider caching this for the duration of the VM service
 // connection.
 Future<Uri?> _connectedAppRoot() async {
-  var fileUriString = await serviceConnection.rootLibraryForMainIsolate();
+  final fileUriString = await serviceConnection.rootLibraryForMainIsolate();
   if (fileUriString == null) return null;
+  return Uri.parse(rootFromFileUriString(fileUriString));
+}
 
+@visibleForTesting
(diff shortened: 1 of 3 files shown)


DanTup (approved):
> can you confirm that should work as expected for Windows paths?

As far as I can tell, this is all operating on URIs, so it should work fine for Windows because file URIs have the same direction slashes:

```
file:///c:/foo/bar
```

That said, while looking through the code to confirm that's the case, I see a lot of APIs that use the term "path" that I believe are using URIs, for example:

```
/// Helper class for storing 1:1 mappings for full file paths to package paths.
class _PackagePathMappings {
  final Map<String, Map<String, String?>> _isolatePackageToFullPathMappings =
      <String, Map<String, String?>>{};
  final Map<String, Map<String, String?>> _isolateFullPathToPackageMappings =
      <String, Map<String, String?>>{};
```

This makes it a bit confusing to read, and could lead to bugs. It would be good if we could generally settle on using "paths" only to mean native file paths (which look like `C:\foo` on Windows) and use things like "uri", "fileUri", "uriString" (for URIs in `String`s) to reduce the chance of bugs related to path differences.

(It would also probably be better to use the `Uri` class anywhere we can for URIs instead of Strings... or maybe an extension type for FileUri that is just Uri guaranteed to have a File scheme)

kenzieschmoll:
Maybe we should add a section here to document this: https://github.com/flutter/devtools/blob/master/STYLE.md. That why we always have a source of truth to refer to for naming practices.

DanTup:
I opened https://github.com/flutter/devtools/pull/7229 with some notes about paths/URIs.

Issue #7059, opened 2024-01-18
Feedback about stats in devools debugger


osaxma:
Hi there,

I have two feedbacks about the stats in the debugger:
- The green stats is a bit hard to read when using Dark theme
- The stat that is shown in line 37 should be in line 44 (that's where the code is waiting the majority of the time).
  - I noticed that most of the stats are off by one line or so, but this one seems to be a bit off (maybe because of the commented code?). 

<img width="948" alt="Screen Shot 2024-01-18 at 7 44 42 PM" src="https://github.com/flutter/devtools/assets/46427323/ed7e7a6b-be2b-41cf-9eb6-a6dd4f00e188">

**DevTools version**: 2.28.4
**Connected Device**:
CPU / OS: x64 (64 bit) macos
Connected app type: Dart CLI
Dart Version: 3.2.3

osaxma:
For reference, the yellow one is also hard to read (not sure if there are other colors in between):

<img width="214" alt="Screen Shot 2024-01-19 at 8 41 58 AM" src="https://github.com/flutter/devtools/assets/46427323/ed2ea2ed-9809-44bc-b1fd-347696f85016">

bkonyi:
> For reference, the yellow one is also hard to read (not sure if there are other colors in between):
> 
[...]

PR #7178, opened 2024-02-07
Fix various issues with coverage and profiling gutters in debugger

bkonyi:
The profiler gutters in the debugger view were displaying profiling information for LINE at LINE + 1 as we weren't properly accounting for the "Total (%) | Self (%)" header taking up the first line's gutter position. This change adds logic to shift the code view and line number gutter down by the height of a single line to account for this header when profiling information is visible.

Line numbers and profiling details were difficult to read in dark mode on top of the red / yellow / green backgrounds used to signify coverage and profile information, so these widgets now use the default light theme text color for rows displaying coverage or profiling information.

This change also fixes some minor padding issues that were visible while the profiler information was made visible.

Fixes https://github.com/flutter/devtools/issues/7059

**Profiler hits enabled adds extra line to the top of the gutter and code view:**

image

**Disabling profiler hits removes the extra line:**

image

**Fixed rendering of profiler hits:**

image

**Verified against Observatory UI:**

image


packages/devtools_app/lib/src/screens/debugger/codeview.dart
```diff
@@ -1051,27 +1082,41 @@ class _LinesState extends State<Lines> with AutoDisposeMixin {
...
         physics: const ClampingScrollPhysics(),
         itemExtent: CodeView.rowHeight,
-        itemCount: widget.lines.length,
+        itemCount: widget.lines.length + profileInformationHeaderOffset,
         itemBuilder: (context, index) {
-          final lineNum = index + 1;
+          if (widget.showProfileInformation && index == 0) {
+            return SizedBox(height: CodeView.rowHeight);
+          }
+          final lineNum = index - profileInformationHeaderOffset + 1;
```
kenzieschmoll, line 1102: nit: create a local var to reduce duplication:
```
final dataIndex = index - profileInformationHeaderOffset;
final lineNumber = dataIndex + 1;
```

then below replace all uses of `index - profileInformationHeaderOffset` with `dataIndex`

kenzieschmoll (approved):

Issue #7210, opened 2024-02-13
Vm Tool throwing exceptions


kenzieschmoll:
This happened when running a dart test ` dart test test/example_test.dart --pause-after-load` and connecting to DevTools
```
flutter: [ERROR]: [zoneGuarded]: type 'Null' is not a subtype of type 'Map<String, dynamic>'
flutter: [ERROR]: [PlatformDispatcher]: type 'Null' is not a subtype of type 'Map<String, dynamic>'
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: type 'Null' is not a subtype of type 'Map<String, dynamic>'
#0      IsolatePrivateViewExtension.tagCounters (package:devtools_app/src/screens/vm_developer/vm_service_private_extensions.dart:34:48)
#1      IsolateStatisticsViewController._updateTagCounters (package:devtools_app/src/screens/vm_developer/isolate_statistics/isolate_statistics_view_controller.dart:75:33)
#2      IsolateStatisticsViewController.switchToIsolate (package:devtools_app/src/screens/vm_developer/isolate_statistics/isolate_statistics_view_controller.dart:65:5)
<asynchronous suspension>
```

PR #7214, opened 2024-02-13
Fix type error in `IsolatePrivateViewExtension.tagCounters`

bkonyi:
This field can be null if the profiler isn't enabled.

Fixes https://github.com/flutter/devtools/issues/7210

### DIFF
--- a/packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
+++ b/packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
@@ -31,7 +31,7 @@ extension VMPrivateViewExtension on VM {
 
 /// An extension on [Isolate] which allows for access to VM internal fields.
 extension IsolatePrivateViewExtension on Isolate {
-  Map<String, dynamic> get tagCounters => json!['_tagCounters'];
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
```diff
@@ -31,7 +31,7 @@ extension VMPrivateViewExtension on VM {
 
 /// An extension on [Isolate] which allows for access to VM internal fields.
 extension IsolatePrivateViewExtension on Isolate {
-  Map<String, dynamic> get tagCounters => json!['_tagCounters'];
+  Map<String, dynamic>? get tagCounters => json!['_tagCounters'];
```
kenzieschmoll, line 34: should this be `json?['_tagCounters']` or do we know `json` is non null?

kenzieschmoll (approved):

packages/devtools_app/lib/src/screens/vm_developer/vm_service_private_extensions.dart
bkonyi: `json` is definitely not null, otherwise the `Isolate` would have no backing data.

Issue #7231, opened 2024-02-15
Tracking bug for re-establishing breakpoints in debuggers


elliette:
Spinning this off of https://github.com/dart-lang/webdev/issues/2257, although this issue involves work for other systems besides DevTools. But I would like everything to be tracked in one place. 

This all began because on web we were experiencing an issue of "phantom breakpoints" where a breakpoint that a user had removed was still being hit in Chrome, see https://github.com/dart-lang/webdev/issues/2257. We would like to remove that behavior and have the debugging clients be the source of truth for breakpoints. However, while DAP clients re-establish breakpoints on hot-restart, DevTools does not. 

On further investigation, we discovered that breakpoints set along the app's initial startup path in DevTools and in DAP clients don't always get hit on hot-restart or (for web) when users refresh the page. This is because breakpoints are re-set by the debugging client after the app's isolate has already started. Therefore, we want to use a combination of:
- `set_isolates_paused_on_start` - make sure the app initially starts out paused
- `requirePermissionToResume(onPauseStart: true)` - make sure that all connected debugging clients using `set_isolates_paused_on_start` send `resume` event before resuming
- debugging clients send `resume` after they have set breakpoints

Therefore, the work required here:

*Note: the following needs to be done before rolling unpinning DWDS and rolling it into google3. It was pinned here: https://github.com/dart-lang/sdk/commit/4642c73347c3d9e04b0d5a4078bd4ba40cfd56d3*

**In DWDS:**
- [x] Remove the logic to re-establish breakpoints after a hot-restart from DWDS (**DONE:**  https://github.com/dart-lang/webdev/pull/2371)
- [x] DWDS needs to implement the VM Service `setFlag` method for `pause_isolates_on_start` (**DONE:** https://github.com/dart-lang/webdev/pull/2373)
- [x] During hot-restart, wait to run `main` if `pause-isolates-on-start` is `true` (**DONE:** https://github.com/dart-lang/webdev/pull/2378)
- [x] After the debugger disconnects, ensure that we no longer pause for hot-restarts/page reloads (**DONE**: https://github.com/dart-lang/webdev/pull/2398)
- [x] Follow-up: Make `pauseIsolatesOnStart` optional (**DONE:** https://github.com/dart-lang/webdev/pull/2397)

**In DevTools:**
- [x] Add logic to re-establish breakpoints after a hot-restart to DevTools (**DONE:** https://github.com/flutter/devtools/pull/7205)
[...]

PR #7234, opened 2024-02-16
Configure DevTools to pause isolates on hot restart, and then only resume once breakpoints have been set

elliette:
Work towards https://github.com/flutter/devtools/issues/7231


packages/devtools_app_shared/lib/src/service/service_manager.dart
```diff
@@ -269,6 +272,20 @@ class ServiceManager<T extends VmService> {
     _connectedState.value = connectionState;
   }
 
+  Future<void> _configureIsolateSettings() async {
+    if (service != null) {
+      try {
+        final vmService = service!;
+        await vmService.setFlag('pause_isolates_on_start', 'true');
+        await vmService.requirePermissionToResume(
+          onPauseStart: true,
```
elliette, line 281: @bkonyi do we need `onPauseReload` here as well? What is the difference between them? (Is `onPauseStart` for hot-restarts and `onPauseReload` for hot-reloads?)

packages/devtools_app_shared/lib/src/service/service_manager.dart
```diff
@@ -269,6 +272,20 @@ class ServiceManager<T extends VmService> {
     _connectedState.value = connectionState;
   }
 
+  Future<void> _configureIsolateSettings() async {
+    if (service != null) {
+      try {
+        final vmService = service!;
+        await vmService.setFlag('pause_isolates_on_start', 'true');
```
bkonyi, line 279: We need to be a bit careful if we're going to set this flag.

If the user starts their process without `--pause_isolates_on_start` set, they connect to DevTools, and then their application spawns a new isolate, they're going to be surprised if their isolate pauses on start.

Similarly, if the user starts their application with `--pause_isolates_on_start`, they connect to DevTools, and their program spawns another isolate, they'll be surprised if DevTools automatically resumes the newly spawned isolate.

We'll want to test both these cases to make sure we can handle them correctly. I'm fairly certain that the second case will be broken with this PR, but I'm not 100% sure about the first case.

elliette: Hmm these are good points. 

I've pushed up changes to resolve the first case:
> "If the user starts their process without --pause_isolates_on_start set, they connect to DevTools, and then their application spawns a new isolate, they're going to be surprised if their isolate pauses on start."

Regarding the second case:
> "Similarly, if the user starts their application with --pause_isolates_on_start, they connect to DevTools, and their program spawns another isolate, they'll be surprised if DevTools automatically resumes the newly spawned isolate."

I think we can handle that case by DevTools getting the flag list from the VM to check if `pause-isolates-on-start` flag has already been set. However, I think there is another issue. If a user is starting their application with `--pause_isolates_on_start`, they will also be surprised if their main isolate is resumed by DevTools (after the breakpoints have been set). There is no way for DevTools to tell if the flag was set by another debugging client (e.g. DAP), in which case as long as we are using `requirePermissionsToResume` it is safe to resume the isolate, or if the flag was set by a user, in which case resuming the isolate would cause unexpected behavior. Adding @DanTup as well since whatever is done here will also need to be done at https://dart-review.googlesource.com/c/sdk/+/350649 (will get back to that CL eventually I promise!!)

One idea: if a user starts their app with `pause-isolates-on-start`, could we also automatically set `requirePermissionsToResume`?

elliette: > "Hmm, what happens if the user started their app via VS Code, and it set pause-on-start. DevTools wouldn't know whether the user has set it (and expects them to stay paused), or if VS Code did it (and is auto-resuming them, so the user expects them not to stay paused)?"

Yes, that's the problem. I'm not sure what a good solution is. 

If both DevTools and DAP set pause-isolates-on-start, they can use `requirePermissionsToResume(onStart)` to both safely resume, knowing that DDS will wait until they have both resumed to send the resume event. 

However, if the user has also set pause-isolates-on-start, there is (as far as I know) no way for DevTools and DAP to know that it was set by the user and not by another debugging client. In which case both of them using `requirePermissionsToResume(onStart)` is not sufficient, because they would both send the resume event and DDS would not wait for the user to also resume. 

That's why I'm proposing that when a user sets `pause-isolates-on-start`, `requirePermissionsToResume` is automatically called for them. Basically, DDS would treat the user like another debugging client, and wait for DevTools, DAP, and the user to all resume before the VM service resumes.

DanTup: > and the user to all resume before the VM service resumes

I think the issue here is that the user doesn't have a connection to the VM, so they can't "resume". The users request to resume goes via DAP (or DevTools) and looks to DDS as a request from them. I think to do this, we'd need to distinguish between a user-initiated resume and a tool asking to resume because it's done with its configuration?

packages/devtools_app_shared/lib/src/service/isolate_manager.dart
```diff
@@ -104,7 +104,16 @@ final class IsolateManager with DisposerMixin {
...
+      final isWebApp =
+          serviceConnection.serviceManager.connectedApp?.isDartWebAppNow ??
+              false;
+      // TODO(bug link) Once we can reliably determine whether a web app is
+      // paused, then switch to resumeIsolateIfPaused.
+      if (isWebApp) {
+        await resumeIsolate(isolateRef);
+      } else {
+        await resumeIsolateIfPaused(isolateRef);
+      }
```
DanTup, line 116: In DAP we tend to just always resume and discard the error if it was not paused. Checking like this can have race conditions if an isolate is unpaused between you checking and the resume request being processed (for example because the user hit resume, or the isolate was only paused temporarily while breakpoints were being set or something).

packages/devtools_app_shared/lib/src/service/service_manager.dart
elliette: Ok sounds good! This is just for hot-restart, not hot-reload.

packages/devtools_app_shared/lib/src/service/service_manager.dart
elliette: Ok - I think this is finally ready for review. 

The second case is now handled:

> ""Similarly, if the user starts their application with --pause_isolates_on_start, they connect to DevTools, and their program spawns another isolate, they'll be surprised if DevTools automatically resumes the newly spawned isolate."

However the first case is not (as discussed offline, this requires DDS to set `requireUserPermissionsToResume` to `true` on start up if the flag `pause-isolates-on-start` is set). However, that change has been made to DDS this will work as well.

elliette: Now that https://github.com/dart-lang/sdk/commit/9382e58fc9ff8a2978037a25161b2bbf2d37c1ea has landed and DDS 4.1.0 has been published the first case is now handled as well.

packages/devtools_app_shared/lib/src/service/isolate_manager.dart
```diff
@@ -237,6 +245,32 @@ final class IsolateManager with DisposerMixin {
...
+    } catch (error) {
+      _log.warning(error);
+    }
+  }
+
+  Future<void> _readyToResume(String isolateId) async {
+    final service = _service!;
+    try {
+      await service.readyToResume(isolateId);
+    } on UnimplementedError {
```
bkonyi, line 267: Does this actually throw `UnimplementedError` on this end? I would assume this would actually result in an `RPCError` with code `kMethodNotFound`.

bkonyi (approved):
LGTM with one comment.

packages/devtools_app_shared/lib/src/service/isolate_manager.dart
elliette: Just verified that is does throw `UnimplementedError` (for both web and non-web apps)

maxfrees:
I'm currently encountering such a problem. Is it caused by this pr?

https://github.com/user-attachments/assets/87e149ac-3274-41b8-9a25-7cc3e7bd46bd

Each time, it needs to be clicked twice to restore before it can run normally. Is there any way to solve this problem

@elliette  @bkonyi @DanTup  please help

DanTup:
@maxfrees it's hard to say without more information. Does it occur in a newly created project? Are you able to share exact steps that would help reproduce the issue? It's probably better to file a new issue with the details.

Issue #7256, opened 2024-02-21
Add a validation command to the `devtools_extensions` executable


kenzieschmoll:
This command should help and extension author validate that their extension meets the requirements for loading into DevTools.

This will perform similar logic to the existing pub validator: https://github.com/dart-lang/pub/blob/master/lib/src/validator/devtools_extension.dart

PR #7257, opened 2024-02-22
Add a `devtools_extensions validate` command

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/7256

### DIFF
--- a/packages/devtools_extensions/bin/_build_and_copy.dart
+++ b/packages/devtools_extensions/bin/_build_and_copy.dart
@@ -102,7 +102,7 @@ class BuildExtensionCommand extends Command {
     if (destinationDirectory.existsSync()) {
       destinationDirectory.deleteSync(recursive: true);
     }
-    Directory(destinationBuildPath)..createSync(recursive: true);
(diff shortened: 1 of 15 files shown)


packages/devtools_extensions/bin/_build_and_copy.dart
```diff
@@ -116,6 +116,7 @@ class BuildExtensionCommand extends Command {
     );
   }
 
+  // ignore: avoid_print, logging command output.
```
bkonyi, line 119: Can you use `stdout.writeln('...')` instead to avoid this lint?

packages/devtools_shared/lib/src/extensions/yaml_utils.dart
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
bkonyi, line 1: Copyright should be 2024 for new files.

bkonyi (approved):

PR #7266, opened 2024-02-23
Initial network tab improvements

CoderDake:
- Fix an oversight where the data refresh doesn't start until after the first timeout
- Reduce processing by removing usage of toList() by keeping track of an array separately from the ID map in CurrentNetworkRequests.
- Ensure a tight coupling of CurrentNetworkRequests to widgets by making it a ValueNotifier
- Optimize addition of new network requests to avoid refreshing more than once per update.
- Update NetworkRequestInspector (side pane) so that it update's it's contents if the value changes
- Change Network Request to a ChangeNotifier so that sub classes can use notififyListeners any time the values change.


packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -361,28 +350,50 @@ class NetworkController extends DisposableController
...
 /// Class for managing the set of all current websocket requests, and
 /// http profile requests.
-class CurrentNetworkRequests {
-  CurrentNetworkRequests({required this.onRequestDataChange});
+class CurrentNetworkRequests extends ValueNotifier<List<NetworkRequest>> {
+  CurrentNetworkRequests() : super([]);
 
-  List<NetworkRequest> get requests => _requestsById.values.toList();
+  ValueListenable<List<NetworkRequest>> get requests =>
+      this; // todo: remove this and just let callers use .value
```
kenzieschmoll, line 357: nit: todos should be formatted as `TODO(<your username, or a github issue>): blah blah` and placed above the code they refer to

kenzieschmoll: also, do we care that the `CurrentNetworkRequests` change notifier is now public? the list of requests could be modified directly outside of this class

packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -155,23 +153,15 @@ class NetworkController extends DisposableController
...
-    );
   }
 
   void processNetworkTraffic({
     required List<SocketStatistic> sockets,
     required List<HttpProfileRequest>? httpRequests,
   }) {
     // Trigger refresh.
-    _requests.value = processNetworkTrafficHelper(
+    // we reassign this every time which
```
kenzieschmoll, line 163: should this comment be removed?

packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -183,11 +173,11 @@ class NetworkController extends DisposableController
 
   void _updatePollingState(bool recording) {
     if (recording) {
-      _pollingTimer ??= Timer.periodic(
+      _pollingTimer ??= DebounceTimer.periodic(
         // TODO(kenz): look into improving performance by caching more data.
         // Polling less frequently helps performance.
         const Duration(milliseconds: 2000),
-        (_) => unawaited(_networkService.refreshNetworkData()),
+        (_) async => await _networkService.refreshNetworkData(),
```
kenzieschmoll, line 180: does awaiting here block the UI thread?

packages/devtools_app/lib/src/screens/network/network_request_inspector.dart
```diff
@@ -45,9 +45,14 @@ class NetworkRequestInspector extends StatelessWidget {
                     style: Theme.of(context).regularTextStyle,
                   ),
                 )
-              : AnalyticsTabbedView(
-                  tabs: _generateTabs(data),
-                  gaScreen: gac.network,
+              : ListenableBuilder(
+                  listenable: data,
+                  builder: (context, _) {
+                    return AnalyticsTabbedView(
```
kenzieschmoll, line 51: this worries me a bit. Read the dart doc for `AnalyticsTabbedView`. If we are rebuilding this unnecessarily, we will get a bunch of analytics events.

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -228,3 +230,31 @@ List<String> stripGoogle3(List<String> pathParts) {
 extension IsKeyType on KeyEvent {
   bool get isKeyDownOrRepeat => this is KeyDownEvent || this is KeyRepeatEvent;
 }
+
+class DebounceTimer {
```
kenzieschmoll, line 234: add dart doc

kenzieschmoll (approved):

packages/devtools_app/lib/src/shared/ui/tab.dart
```diff
@@ -10,6 +10,12 @@ import '../analytics/analytics.dart' as ga;
 double get _tabHeight => scaleByFontFactor(46.0);
 double get _textAndIconTabHeight => scaleByFontFactor(72.0);
 
+enum AnalyticsSendSettings {
```
kenzieschmoll, line 13: delete this

packages/devtools_app/lib/src/screens/network/network_controller.dart
CoderDake: I'm not sure I'm too concerned about this. normal valuelisteners for Lists suffer from the same issue. Anyone could modify the value without others being notified of the changes.

Since this is internal to Devtools only I don't think we need to be too concerned about abuse of this.

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -243,25 +243,40 @@ class DebounceTimer {
...
-
-      try {
-        _isRunning = true;
-        await callback(timer);
-      } finally {
-        _isRunning = false;
-      }
+    // Start a timer with no duration so that the callback will be triggered
+    // immediately.
+    _timer = Timer(const Duration(), () async {
```
kenzieschmoll, line 248: we don't need another timer for this.  just call `callback` directly once before starting the periodic timer

packages/devtools_app/test/shared/utils_test.dart
```diff
@@ -15,32 +14,30 @@ void main() {
...
       fakeAsync((async) {
         int callbackCounter = 0;
         DebounceTimer.periodic(
           const Duration(seconds: 1),
           (_) async {
             callbackCounter++;
-            await Future<void>.delayed(
-              const Duration(milliseconds: 1),
-            );
+            await Future<void>.delayed(const Duration(seconds: 30));
```
kenzieschmoll, line 29: are we actually waiting this much time on the test? Make these durations much shorter if so

CoderDake: No, these tests use fake async to mock the passage of time. So everything happens almost instantly :)

packages/devtools_app/lib/src/shared/utils.dart
CoderDake: Since it is a FutureOr, I couldn't find a way to trigger the callback inside the constructor, as they can't be async.

kenzieschmoll: Why does it need to be a `FutureOr`? if you change the parameter type to `Future`, you can do something like this:
```
unawaited(callback().then((_) {
  /* start periodic timer */
));
```

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -228,3 +230,51 @@ List<String> stripGoogle3(List<String> pathParts) {
...
+
+        // Now a periodic timer is used to repeatedly trigger the callback after
+        // the duration.
+        _timer = Timer.periodic(duration, (timer) async {
+          // If the previous callback is still running, then don't trigger another
+          // callback. (debounce)
+          if (_isRunning) {
+            return;
+          }
+
```
kenzieschmoll, line 260: should we also check isCancelled here?

kenzieschmoll (approved):
lgtm with one comment

packages/devtools_app/lib/src/shared/utils.dart
CoderDake: The time would be cancelled before it is ever run, but this makes me realize I should probably start the timer at the same time as starting the callback so the duration gap is always the same! Thanks for asking this question!

CoderDake: Your comment helped me realize that this could be written without _isCancelled all together :D

PR #7286, opened 2024-02-27
Small release helper fixes

CoderDake:

### DIFF
--- a/tool/lib/commands/release_helper.dart
+++ b/tool/lib/commands/release_helper.dart
@@ -86,11 +86,11 @@ class ReleaseHelperCommand extends Command {
 
       final getNewVersionResult = await processManager.runProcess(
         CliCommand.tool(
-          ['update-version', 'current-version', '|', 'tail', '-n1'],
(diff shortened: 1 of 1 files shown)


tool/lib/commands/release_helper.dart
```diff
@@ -86,11 +86,11 @@ class ReleaseHelperCommand extends Command {
 
       final getNewVersionResult = await processManager.runProcess(
         CliCommand.tool(
-          ['update-version', 'current-version', '|', 'tail', '-1'],
+          ['update-version', 'current-version'],
         ),
       );
 
-      final newVersion = getNewVersionResult.stdout.trim();
+      final newVersion = getNewVersionResult.stdout.split('\n').last.trim();
```
kenzieschmoll, line 93: curious: why is this change necessary?

CoderDake: I don't think the pipe works in the run process command. It was failing when I ran.
So instead of ping to tail, I'm just grabbing the last line on our side.

Issue #7192, opened 2024-02-08
Meta tracking issue: "Reporting error: Failed to parse package:xyz/main.dart"


kenzieschmoll:
We have gotten several reports of this error. The stack traces are minified so they aren't helpful, but looks like it is coming from `codeview_controller.dart` based on the String:
https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/debugger/codeview_controller.dart/#L251

kenzieschmoll:
One question we should consider once we have a repro is whether we want to be notifying the user with this parsing error happens (`notifyUser: true` is set currently). If this error is expected (potentially on a hot restart or app disconnect?) then we might want to fail gracefully and not give the users an error notification that they can report.

kenzieschmoll:
Some repro details shared on https://github.com/flutter/devtools/issues/7305.

elliette:
**Update:** With https://github.com/flutter/devtools/pull/8385 we're finally starting to get some useful stack traces for these errors:

* https://github.com/flutter/devtools/issues/8879

kenzieschmoll:
I just closed 9 more duplicates of this issue. We should be sure to fix this before the next beta cut.

elliette:
**NOTE: Fixed by https://github.com/flutter/devtools/pull/9104**

Issue #7305, opened 2024-03-04
Reporting error: Failed to parse package:vocabugator/main.dart.


pabluez:
Stack trace:
main.dart.js 6212:18    Object.qn
main.dart.js 69298:76   <fn>
main.dart.js 4775:63    bFj.a
main.dart.js 58094:14   bFj.$2
main.dart.js 58088:21   bDc.$1
main.dart.js 5071:19    Object.bEY
main.dart.js 168428:68  al8.<fn>
main.dart.js 59293:12   a_M.rF
main.dart.js 58598:11   bpI.$0
main.dart.js 4940:40    Object.Nj

I get an error everytime I try to reuse an already open but disconnect DevTools for web tab by just copying and pasting the URL I see in my debug output window.
![image](https://github.com/flutter/devtools/assets/777052/b3af97b7-7522-43dc-b2ca-1bdbe9d693e6)

___
**DevTools version**: 2.31.1
**IDE**: VSCode

kenzieschmoll:
duplicate of https://github.com/flutter/devtools/issues/7192

elliette:
Hi @pabluez!

Thank you for the report! I haven't been able to reproduce this, but I'm not sure I'm trying the same thing you are describing. Could you please give a few more details? Specifically:

- how are you running the app? are you running it from the command line or from an IDE (e.g. VS Code, or?)
- how are you opening DevTools? are you launching it from an IDE, or copy-pasting the URL in your terminal?
- how does DevTools get into a disconnected state? Is the app no longer running, or are you clicking "Connect to a new app" from DevTools?
- which URL are you copy-pasting (what "debug output window" is it in), and where are you pasting it to?

PR #7335, opened 2024-03-08
Compile release build of DevTools with default dart2js optimizations 

elliette:
Work towards https://github.com/flutter/devtools/issues/7192

It would be helpful for DevTools bug reports to have more intelligible stack traces. This PR changes DevTools to use unminified dart2js. See https://dart.dev/tools/dart-compile#js for information about the varous optimization levels.


kenzieschmoll (approved):

kenzieschmoll (commented):
I think we need to update build_release.sh too. The automated builder that uploads DevTools builds to CIPD uses that script

hangyujin:
This pr broke `devtools_tool serve`, it launches an empty website page now.

elliette:
> I think this pr broke `devtools_tool serve`, it launches an empty website page now.

Uh oh, thanks! Will take a look

sigmundch:
I'm curious, is the use O1 is meant to be temporary while debugging the other issue or a permanent change?

Just in case this is not temporary, let me suggest a couple other strategies that could provide a better balance of performance and debugability.

1. Only disable minification. Since `-O4` does a lot more than `-O1`, there is a risk that running at `-O1` you'd see performance issues on DevTools itself come up. Internally, dart2js enables minification from `-O2` and up. However, it also has a `--no-minify` flag that can turn it off while keeping all other optimizations enabled. Using `--no-minify` will keep DevTools performing at it's current performance level. Note: this is what flutter uses when you enable "profile" mode on web: `fluter build web --profile`. 

2. Keep minification, but use our stacktrace [deobfuscation tools](https://github.com/dart-lang/sdk/blob/e79131df0f8994754dc6c9323756fb69681f5a57/pkg/dart2js_tools/bin/deobfuscate.dart). We created a few tools that are mainly used by internal apps today. These tools can desugar traces reported from a minified app by using the source-maps generated by dart2js. The deobfuscator can unmangle names, reverse error messages, and even split out inlined call frames (this requires dart2js source map [extensions](https://github.com/dart-lang/sdk/blob/e79131df0f8994754dc6c9323756fb69681f5a57/pkg/compiler/doc/sourcemap_extensions.md), which are enabled by default today).

The latter requires a bit more setup, but it's the recommended approach to ship minified apps and still be able to get meaningful stack traces from production runs.

sigmundch:
Re option (2), I was trying to see if I could lookup the stack provided in https://github.com/flutter/devtools/issues/7305,  but I noticed that devtools currently doesn't ship it's sourcemap file. Is that correct? Is it by chance built and saved somewhere that could be accessed retroactively?

If not, that would be the main thing we need to do to make it possible to use the deobfuscation tools. We could try to rebuild the exact same state, but it would require using the exact same version of flutter/dart as it was used at the time the release was made (in this example, for devtools 2.31.1), which is likely hard to do.

PR #7342, opened 2024-03-09
Disconnected memory screen.

polina-c:
Screenshot 2024-03-11 at 12 12 51 PM


polina-c:
I wanted to make it under flag, but enum ScreenMetaData requires everything to be const. I can hack it if it is important though.

STYLE.md
```diff
@@ -16,9 +16,9 @@ in compliance with
 2. Private field
 3. Public setter (when needed)
 
-## Naming for typedefs and function variables
+## Naming for typedef`s and function variables
 
-Follow [Flutter repo naming rules for typedefs and function variables](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#naming-rules-for-typedefs-and-function-variables).
+Follow [Flutter repo naming rules for typedef`s and function variables](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#naming-rules-for-typedefs-and-function-variables).
```
kenzieschmoll, line 21: These apostrophes are not needed. Instead you can use backtics: `typedef`s

packages/devtools_app/lib/src/screens/memory/framework/offline/screen_data.dart
```diff
@@ -0,0 +1 @@
+class MemoryScreenOfflineData {}
```
kenzieschmoll, line 1: copyright and dartdoc

packages/devtools_app/lib/src/shared/offline_mode.dart
```diff
@@ -43,6 +43,12 @@ class OfflineModeController {
   }
 }
 
+/// This mixin adds offline functionality to a screen.
+///
+/// For offline mode, the screen controller should add this mixing,
```
kenzieschmoll, line 48: sp: mixin

packages/devtools_app/lib/src/shared/screen.dart
```diff
@@ -53,6 +53,8 @@ enum ScreenMetaData {
     'memory',
     title: 'Memory',
     icon: Octicons.package,
+    worksOffline: true,
+    requiresConnection: false,
```
kenzieschmoll, line 57: let's not add these until support is ready.

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
```diff
@@ -65,19 +68,46 @@ class MemoryFeatureControllers {
...
+        offlineController.shouldLoadOfflineData(ScreenMetaData.memory.id);
+    if (!isOffline) {
+      _controllers = MemoryFeatureControllers(
+        diffPaneController,
+        profilePaneController,
+      );
+    }
+  }
+
+  late final bool isOffline;
```
kenzieschmoll, line 88: we should not store this value on the memory controller. it may get out of sync. Use the global offlineController as the source of truth

packages/devtools_app/lib/src/screens/memory/framework/memory_screen.dart
```diff
@@ -5,10 +5,10 @@
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 
+import '../../../../devtools_app.dart';
```
kenzieschmoll, line 8: import libraries directly

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
```diff
@@ -65,19 +68,46 @@ class MemoryFeatureControllers {
...
+  FutureOr<void> processOfflineData(MemoryScreenOfflineData offlineData) {
+    offlineData = offlineData;
+  }
+
+  @override
+  OfflineScreenData screenDataForExport() {
+    return OfflineScreenData(
+      screenId: ScreenMetaData.memory.id,
+      data: {
+        '$MemoryScreenOfflineData': MemoryScreenOfflineData(),
```
kenzieschmoll, line 100: just return an empty map for now:
`data: {}`

This will eventually be something like:
```
data: OfflineMemoryData(profile: {}, snapshots: [...], ...).toJson(),
```

packages/devtools_app/lib/src/app.dart
```diff
@@ -579,6 +566,7 @@ List<DevToolsScreen> defaultScreens({
     DevToolsScreen<MemoryController>(
       MemoryScreen(),
       createController: (_) => MemoryController(),
+      supportsOffline: true,
```
kenzieschmoll, line 569: don't enable this until support is ready

packages/devtools_app/lib/src/screens/memory/framework/memory_screen.dart
```diff
@@ -38,16 +40,27 @@ class MemoryBody extends StatefulWidget {
...
+
   @override
   Widget build(BuildContext context) {
-    // TODO(polina-c): load static body if not connected.
-    return const ConnectedMemoryBody();
+    if (controller.isOffline ||
+        !serviceConnection.serviceManager.hasConnection) {
+      return const OfflineMemoryBody();
+    } else {
+      return const ConnectedMemoryBody();
```
kenzieschmoll, line 63: This is fine for this PR, but I think we should try to reuse a single widget. In an ideal world, the MemoryBody widget would be fed data from the memory controller, and it should not need to have many checks for whether we are online or off.

packages/devtools_app/lib/src/screens/memory/framework/memory_screen.dart
```diff
@@ -38,16 +40,27 @@ class MemoryBody extends StatefulWidget {
   MemoryBodyState createState() => MemoryBodyState();
 }
 
-class MemoryBodyState extends State<MemoryBody> {
+class MemoryBodyState extends State<MemoryBody>
+    with ProvidedControllerMixin<MemoryController, MemoryBody> {
   @override
   void initState() {
     super.initState();
     ga.screen(MemoryScreen.id);
```
kenzieschmoll, line 48: add a listener for offline mode:
`addAutoDisposeListener(offlineController.offlineMode);`

packages/devtools_app/lib/src/screens/memory/framework/memory_screen.dart
```diff
@@ -38,16 +40,27 @@ class MemoryBody extends StatefulWidget {
...
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+    initController();
+  }
+
   @override
   Widget build(BuildContext context) {
-    // TODO(polina-c): load static body if not connected.
-    return const ConnectedMemoryBody();
+    if (controller.isOffline ||
```
kenzieschmoll, line 59: check `offlineController.offlineMode.value`

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
polina-c: If the entire MemoryScreenOfflineData is serializable, why I cannot pass it as is?
It seems to be convenient to have all the data as one object.

packages/devtools_app/lib/src/screens/memory/framework/memory_screen.dart
polina-c: Most likely yes.

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
kenzieschmoll: The value passed to data will be encoded as json here: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/shared/offline_mode.dart#L71.

So I don't think the `MemoryScreenOfflineData` object will serialize well for `jsonEncode`.

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
kenzieschmoll: The only way to guarantee consistency is by using the `offlineController.offlineMode` value notifier. This is the source of truth for offline mode. Any piece of code that needs to check offline mode should check `offlineController.offlineMode.value`. If you search the codebase for "offlineController.offlineMode.value", you will see a lot of uses of this you can use for examples.

packages/devtools_app/lib/src/shared/offline_mode.dart
```diff
@@ -43,6 +43,19 @@ class OfflineModeController {
...
 }
 
+/// This mixin adds offline functionality to a screen.
+///
+/// For offline mode, the screen controller should add this mixin,
+/// implement its abstract methods.
+///
+/// To detect if the app is in offline mode, check `offlineController.offlineMode.value`.
+///
+/// To detect if offline data are available check the value of the
```
kenzieschmoll, line 53: r/are/is

packages/devtools_app/lib/src/shared/offline_mode.dart
```diff
@@ -43,6 +43,19 @@ class OfflineModeController {
   }
 }
 
+/// This mixin adds offline functionality to a screen.
+///
+/// For offline mode, the screen controller should add this mixin,
+/// implement its abstract methods.
+///
+/// To detect if the app is in offline mode, check `offlineController.offlineMode.value`.
```
kenzieschmoll, line 51: nit: r/app/DevTools

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
polina-c: Made memory controller listening for offlineController.offlineMode, and widget listening for initialization status of the controller. 
How does it look?

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
polina-c: Added hello world code for it. It works.

kenzieschmoll: This fails on DartPad
Screenshot 2024-03-12 at 1 14 55 PM

PR #7382, opened 2024-03-18
Don't compile DevTools with unminified JS

elliette:
Reverts the changes in https://github.com/flutter/devtools/pull/7335

Using the unminified dart2js optimization levels breaks serving DevTools due to an unexpected type error. I've opened https://github.com/dart-lang/sdk/issues/55234 in the SDK with details.

### DIFF
--- a/tool/build_release.sh
+++ b/tool/build_release.sh
@@ -62,7 +62,9 @@ flutter pub get
 flutter build web \
   --web-renderer canvaskit \
   --pwa-strategy=offline-first \
-  --dart2js-optimization=O1 \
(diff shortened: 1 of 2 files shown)


kenzieschmoll:
Are we sure compiling with unminified js was the issue? @bkonyi submitted a fix for this error this morning: https://github.com/flutter/devtools/pull/7374

elliette:
> Are we sure compiling with unminified js was the issue? @bkonyi submitted a fix for this error this morning: #7374

Hmm let me pull in his fix and see

PR #7389, opened 2024-03-19
Fix multiline text in display JsonViewer

CoderDake:
There is only 1 line to show text in the tree viewer, so I've updated the display provider to ensure that the text is only 1 line.

The original text can be viewed in a tooltip now.

The text copied from the copy button is not effected
## Open Questions
* Do we care about having tooltips in the other places where TreeView is used?
* Is the \n approach reasonable? 
    * in the json itself newlines are represented like this so i feel like it is reasonable.

## Before
Screenshot 2024-03-19 at 10 36 41 AM

## After
Screenshot 2024-03-19 at 10 35 14 AM


packages/devtools_app/lib/src/shared/console/widgets/display_provider.dart
```diff
@@ -56,30 +56,42 @@ class DisplayProvider extends StatelessWidget {
...
       menuButtons: _getMenuButtons(context),
-      child: Text.rich(
-        TextSpan(
-          text: hasName ? variable.name : null,
-          style: variable.artificialName
-              ? theme.subtleFixedFontStyle
-              : theme.fixedFontStyle.apply(
-                  color: theme.colorScheme.controlFlowSyntaxColor,
+      child: DevToolsTooltip(
+        message: variable.displayValue.toString(),
```
kenzieschmoll, line 70: use originalDisplayValue here

packages/devtools_app/lib/src/shared/console/widgets/display_provider.dart
```diff
@@ -56,30 +56,42 @@ class DisplayProvider extends StatelessWidget {
...
+        child: Text.rich(
+          maxLines: 1,
+          overflow: TextOverflow.ellipsis,
```
kenzieschmoll, line 73: instead of showing the \n chars, can we just make this a multiline text widget? With some reasonable number of max lines like 3 or 4? We would still ellipsize and have a tooltip.

packages/devtools_app/release_notes/NEXT_RELEASE_NOTES.md
```diff
@@ -40,6 +40,8 @@ TODO: Remove this section if there are not any general updates.
 
 * Improved Network profiler performance. [#7266](https://github.com/flutter/devtools/pull/7266)
 * Fixed a bug where selected pending requests weren't refreshing the tab once updated. [#7266](https://github.com/flutter/devtools/pull/7266)
+* Fixed the JSON viewer so multiline strings are visible in their row, and
+  through a tooltip [#7389](https://github.com/flutter/devtools/pull/7389)
```
kenzieschmoll, line 44: nit: period at end of sentence and separate PR links from description with a `-` character

kenzieschmoll (approved):

PR #7397, opened 2024-03-20
Add DCP refresh guidance.

polina-c:


DanTup (approved):
LGTM, though updating the global version on your machine to match DevTools if you also work on other projects might not be ideal - it might be useful to include how you could use a specific version for DevTools that might be different to other projects (DevTools could be pinned to an older version than you want to use for other projects)?

PR #7400, opened 2024-03-20
Cleanup desktop platform configuration

kenzieschmoll:
Setting the default target platform is no longer necessary, and this causes a crash when building desktop in profile mode.


CoderDake (approved):
🔥

PR #7406, opened 2024-03-20
Refactor and cleanup byte utilities

kenzieschmoll:
This refactor / cleanup is in preparation of an upcoming PR.


packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
```diff
@@ -318,7 +318,11 @@ class _GCHeapCapacityColumn extends _GCHeapStatsColumn {
 
   @override
   String getDisplayValue(AdaptedProfile dataObject) {
-    return prettyPrintBytes(getValue(dataObject), includeUnit: true)!;
+    return prettyPrintBytes(
+      getValue(dataObject),
+      includeUnit: true,
+      kbFractionDigits: 0,
```
kenzieschmoll, line 324: @polina-c should these columns actually have KB fraction digits of size 1? This PR changes the default value for this parameter so I added the 0 here to keep the current behavior, but wanted to check if this is the current behavior by accident or if it is intentional.

jacob314: agree that this should probably be changed to 1.

packages/devtools_app/lib/src/screens/network/network_request_inspector_views.dart
```diff
@@ -421,6 +422,7 @@ class ImageResponseView extends StatelessWidget {
               prettyPrintBytes(
                 encodedResponse.lengthInBytes,
                 includeUnit: true,
+                kbFractionDigits: 0,
```
jacob314, line 425: doesn't really matter here as images are large but could probably be 1 for consistency.

packages/devtools_app/lib/src/shared/primitives/byte_utils.dart
```diff
@@ -0,0 +1,120 @@
...
+  int mbFractionDigits = 1,
+  int gbFractionDigits = 1,
+  bool includeUnit = false,
+  num roundingPoint = 1.0,
+  int maxBytes = 52,
+}) {
+  if (bytes == null) {
+    return null;
+  }
+  // TODO(peterdjlee): Generalize to handle different kbFractionDigits.
```
jacob314, line 23: nit: change TODO to your name while moving code as peter is unlikely to resolve the todo.

jacob314 (approved):
LGTM with a couple suggestions that are better for a followup than this pull request.

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
kenzieschmoll: done

jacob314 (approved):
lgtm

Issue #7366, opened 2024-03-14
Use DTD to project roots for the deep link tool


kenzieschmoll:
Right now, users have to manually enter a path to their project root. We can simplify this user journey by populating a list of project roots from the Dart Tooling Daemon. We can still show the text field for entering a path if DTD is not available.

Some additional context: https://github.com/flutter/devtools/issues/7253#issuecomment-1997440951

This would be a great feature to get in to the next beta (about 2 weeks until the code cutoff).

DanTup:
> workspace roots

I'm not familiar with the tool requirements just to make sure it's not accidentally overlooked, your workspace roots might not be the same as the project/package roots. Eg. your workspace root might be:
```js
[
  "/mono_repo_1",
  "/mono_repo_2/"
]
```

but your packages might be in

```
[
  "/mono_repo_1/pkg1",
  "/mono_repo_1/pkg2",
  "/mono_repo_1/pkg2",
  "/mono_repo_2/pkg2"
]
````

kenzieschmoll:
Reopening since https://github.com/flutter/devtools/pull/7415 only partially fixed this issue, but it will have accuracy issues when the flutter project is not the project root. Moving discussion from https://github.com/flutter/devtools/pull/7415 to here:

This approach will work well for Flutter projects where the project root is opened in the IDE, but it will not work well for projects where the flutter project is deeper in the tree. Here is an example from DevTools, where the DevTools flutter app is actually under devtools/packages/devtools_app, but it is not in the list here:

![Screenshot 2024-03-21 at 4 13 37 PM](https://github.com/flutter/devtools/assets/43759233/0673cb72-0fe8-4bd8-ac28-6ecf45450658)

@DanTup, @helin24 - does the IDE know where all the Flutter project roots are within a monorepo? If so could the IDE register a method on DTD that DevTools can call to get the Flutter project roots instead of the IDE workspace roots?

DanTup:
> @DanTup, @helin24 - does the IDE know where all the Flutter project roots are within a monorepo?

Technically, it can find them (we don't track them, but we scan the workspace for them* when we need them such as during startup or if the user runs "Get Packages" with no file for context).

However, I think we would be better reversing this and doing it in DTD. We've talked about this in the past (potentially having the analysis server do it), because the IDEs might implement this slightly differently, and the rules can change across SDK versions (for example the upcoming monorepo changes may change the rules for how to decide what a project root is).

[...]

PR #7415, opened 2024-03-21
Get workspace roots from DTD for Deep Links tool

kenzieschmoll:
This will automatically populate the IDE workspace roots from the IDE that the user opened devtools in / from. As a fallback, the text field is still an option.

Screenshot 2024-03-21 at 3 17 25 PM

@hangyujin @zoeyfan 

Fixes https://github.com/flutter/devtools/issues/7366


kenzieschmoll:
This approach will work well for Flutter projects where the project root is opened in the IDE, but it will not work well for projects where the flutter project is deeper in the tree. Here is an example from DevTools, where the DevTools flutter app is actually under packages/devtools_app, but it is not in the list here:

Screenshot 2024-03-21 at 4 13 37 PM

@DanTup, @helin24 - does the IDE know where all the Flutter project roots are within a monorepo? If so could the IDE register a method on DTD that DevTools can call to get the Flutter project roots instead of the IDE workspace roots?

hangyujin (approved):
LGTM, thank you!

kenzieschmoll:
This will really improve the UX of loading a project by automating this step for the user 🎉 
deep-links-project-selector

We still need to find a solution for the monorepo case, but for the case where the user opens a project to the root of their flutter app, this will work nicely.

Issue #7425, opened 2024-03-22
Export snapshot fails with exception


kenzieschmoll:
i got an exception when exporting a snapshot:
![Screenshot 2024-03-22 at 10 45 55 AM](https://github.com/flutter/devtools/assets/43759233/25bb2828-8660-401e-a462-9df76c13c05a)

![Screenshot 2024-03-22 at 10 45 20 AM](https://github.com/flutter/devtools/assets/43759233/66941a0a-a702-41b2-93ae-38735074f7bd)

This was running against a Dart CLI app.

polina-c:
Remaining: add test coverage

polina-c:
To add test coverage waiting for https://github.com/flutter/devtools/issues/7425 to be addressed.

PR #7439, opened 2024-03-25
Fix export.

polina-c:
Contributes to: https://github.com/flutter/devtools/issues/7425

Minimal fix for snapshot export for web.
Next PRs will:
1. Create fix for macos
2. Create integration test

https://github.com/flutter/devtools/assets/12115586/bb9259c6-789e-4bfe-82d0-68acafb6a8ea


packages/devtools_app/lib/src/shared/config_specific/import_export/_export_desktop.dart
```diff
@@ -21,4 +23,9 @@ class ExportControllerDesktop extends ExportController {
   }) {
     _fs.writeStringToFile(fileName, content);
   }
+
+  @override
+  void saveDataFile({required Uint8List content, required String fileName}) {
+    throw UnimplementedError();
```
kenzieschmoll, line 29: can you write this to file (like above) so that we can still use this feature on the Desktop build of DevTools?

packages/devtools_app/lib/src/shared/config_specific/import_export/_export_web.dart
```diff
@@ -15,20 +16,37 @@ ExportControllerWeb createExportController() {
...
+    // ignore: avoid-dynamic, there is not better way in this case
+    required dynamic content,
     required String fileName,
   }) {
     final element = document.createElement('a') as HTMLAnchorElement;
     element.setAttribute(
       'href',
-      URL.createObjectURL(Blob([content.toJS].toJS) as JSObject),
+      // ignore: avoid-dynamic, avoid_dynamic_calls, there is not better way in this case,
+      URL.createObjectURL(Blob(([content.toJS] as dynamic).toJS) as JSObject),
```
kenzieschmoll, line 28: @srujzs is this what you would expect we'd need to do here?

packages/devtools_app/lib/src/shared/config_specific/import_export/_export_web.dart
```diff
@@ -15,20 +16,37 @@ ExportControllerWeb createExportController() {
...
+  @override
+  void saveFile({
+    required String content,
+    required String fileName,
+  }) {
+    _downloadJsObject(content: content, fileName: fileName);
+  }
+
+  @override
+  void saveDataFile({
```
kenzieschmoll, line 46: can these be merged into one method?
```
void saveFile<T>({
  required T content,
  required String fileName,
})
```

packages/devtools_app/lib/src/shared/config_specific/import_export/_export_desktop.dart
polina-c: YEs, this is the plan, but can i do it is separate PR. I want to squise the fix to release.

packages/devtools_app/lib/src/shared/config_specific/import_export/_export_web.dart
srujzs: I don't think this will work as intended. `toJS` doesn't exist dynamically, and if it does, it won't be consistent across all web compilers.

Does `content` need to be `dynamic`? JS interop requires that we know the type in order to convert it. If it's a set of types, we can do type checks to convert each case.

srujzs: For example if `content` is either a `String` or a `Uint8List`, we can do the following:

```dart
JSAny jsContent = content is String ? content.toJS : (content as Uint8List).toJS;
URL.createObjectURL(Blob([content].toJS]) as JSObject);
```

packages/devtools_app/lib/src/shared/config_specific/import_export/_export_desktop.dart
kenzieschmoll: actually, if you merge to use one method (see my comment below), then you can use the existing override above.

kenzieschmoll:
We should add an integration test with this fix. The test can be simple - load the memory page, take a snapshot, and export it. If that succeeds without exception, then this would be test covered. Later, we can build upon this test to re-import the exported snapshot.

packages/devtools_app/lib/src/shared/config_specific/import_export/_export_web.dart
bkonyi: +1. Can we not give `_downloadJsObject` a type argument as well to avoid the dynamic in the signature?

packages/devtools_app/lib/src/shared/config_specific/import_export/_export_web.dart
polina-c: Thanks. This way .toJS does not resolve for Blob. So, I made it little differently.

packages/devtools_app/lib/src/shared/config_specific/import_export/_export_desktop.dart
polina-c: May be. I need to test it.

packages/devtools_app/lib/src/shared/config_specific/import_export/_export_web.dart
srujzs: Looks good to me!

polina-c:
> Member

working on it

packages/devtools_app/integration_test/test/live_connection/memory_screen_helpers.dart
```diff
@@ -0,0 +1,135 @@
...
+import 'package:devtools_app/devtools_app.dart';
+import 'package:devtools_app/src/screens/memory/panes/control/primary_controls.dart';
+import 'package:devtools_app/src/screens/memory/panes/diff/widgets/snapshot_list.dart';
+import 'package:devtools_app/src/shared/console/widgets/console_pane.dart';
+import 'package:devtools_test/helpers.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+// / Prepares the UI of the memory screen so that the eval-related elements are
+// / visible on the screen for testing.
```
kenzieschmoll, line 14: dart doc formatting

packages/devtools_app/integration_test/test/live_connection/memory_screen_helpers.dart
```diff
@@ -0,0 +1,135 @@
...
+    const dragDistance = -320.0;
+    await tester.drag(
+      find.byType(ConsolePaneHeader),
+      const Offset(0, dragDistance),
+    );
+    await tester.pumpAndSettle();
+  }
+
+  // For the sake of this test, do not show extension screens by default.
+  preferences.devToolsExtensions.showOnlyEnabledExtensions.value = true;
```
kenzieschmoll, line 55: why do we need this?

packages/devtools_app/lib/src/shared/config_specific/import_export/import_export.dart
```diff
@@ -146,29 +145,17 @@ abstract class ExportController {
...
-      case String:
-        saveFile(content: content as String, fileName: fileName);
-      case Uint8List:
-        saveDataFile(content: content as Uint8List, fileName: fileName);
-      default:
-        throw 'Unsupported content type: $T';
-    }
-
+    saveFile(
+      content: content as String,
```
kenzieschmoll, line 149: should this cast be removed here since content could be String or Uint8List?

also, add the type to saveFile: `saveFile<T>(...)`

kenzieschmoll (approved):

packages/devtools_app/integration_test/test/live_connection/memory_screen_helpers.dart
polina-c: I do not know. I just moved it as it was, from a test to helpers.

kenzieschmoll: I do not see where in the diff this was copied from. This shouldn't be necessary so let's remove it.

polina-c:
Local integration test failure during snapshot:

```
══╡ EXCEPTION CAUGHT BY SCHEDULER LIBRARY ╞═════════════════════════════════════════════════════════
The following TypeErrorImpl was thrown during a scheduler callback:
Unexpected null value.

dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 296:3       throw_
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 586:18  nullCheck
lib/_engine/engine/semantics/semantics.dart 1421:82                               updateChildren
lib/_engine/engine/semantics/semantics.dart 2377:13                               updateSemantics
lib/_engine/engine/window.dart 112:5                                              updateSemantics
packages/flutter_test/src/window.dart 909:5                                       updateSemantics
packages/flutter/src/rendering/view.dart 330:5                                    updateSemantics
packages/flutter/src/rendering/binding.dart 245:48                                <fn>
packages/flutter/src/semantics/semantics.dart 3466:22                             sendSemanticsUpdate
packages/flutter/src/rendering/object.dart 1306:7                                 flushSemantics
packages/flutter/src/rendering/object.dart 1308:14                                flushSemantics
packages/flutter/src/rendering/binding.dart 584:7                                 drawFrame
packages/flutter/src/widgets/binding.dart 1023:13                                 drawFrame
packages/flutter/src/rendering/binding.dart 443:5                                 [_handlePersistentFrameCallback]
packages/flutter/src/scheduler/binding.dart 1392:7                                [_invokeFrameCallback]
packages/flutter/src/scheduler/binding.dart 1313:9                                handleDrawFrame
packages/flutter_test/src/binding.dart 1789:13                                    handleDrawFrame
packages/flutter/src/scheduler/binding.dart 1171:5                                [_handleDrawFrame]
dart-sdk/lib/async/zone.dart 1399:13                                              _rootRun
dart-sdk/lib/async/zone.dart 1301:19                                              run
dart-sdk/lib/async/zone.dart 1209:7                                               runGuarded
lib/_engine/engine/platform_dispatcher.dart 1406:9                                invoke
lib/_engine/engine/platform_dispatcher.dart 307:5                                 invokeOnDrawFrame
lib/_engine/engine/initialization.dart 187:36                                     <fn>
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 426:37  _checkAndCall
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 431:39  dcall
════════════════════════════════════════════════════════════════════════════════════════════════════
```

polina-c:
Moved both menu item 'Export' and integration test to under the feature flag for now.

polina-c:
The exception is reported on stack overflow: https://stackoverflow.com/questions/69677878/exception-caught-by-scheduler-library-unexpected-null-value

kenzieschmoll:
Before landing this, can you manually test that exporting and importing files from the Performance and CPU profiler page is not broken with this change?

polina-c:
> Before landing this, can you manually test that exporting and importing files from the Performance and CPU profiler page is not broken with this change?

Checked. Works.

PR #7458, opened 2024-03-26
Update flutter-candidate.txt

polina-c:
Fix dev build:

Screenshot 2024-03-26 at 2 24 50 PM


flutter-candidate.txt
```diff
@@ -1 +1 @@
-3.21.0-15.0.pre.11
+3.21.0-15.0.pre
```
kenzieschmoll, line 1: does this still have Srujan's fix?

polina-c: nope, but the build will not be broken

kenzieschmoll (approved):

Issue #7429, opened 2024-03-22
Selecting frame after "no timeline events for this time range" message is broken


kenzieschmoll:
repro:
1. Run a flutter app for a while
2. open the devtools timeline at the point where no timeline events are available for the flutter frames in the frame chart
3. clear data
4. do something to trigger frames (new events should have been recorded)
5. select a frame
6. nothing happens 
![no-timeline-events-loaded](https://github.com/flutter/devtools/assets/43759233/a1460159-9fd6-4a74-a1c0-77ca3e70253c)

What should happen is the timeline should refresh to try to bring in events from these new frames

PR #7461, opened 2024-03-26
Fix bug with selecting flutter frames

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/7429

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/panes/timeline_events/timeline_events_controller.dart
+++ b/packages/devtools_app/lib/src/screens/performance/panes/timeline_events/timeline_events_controller.dart
@@ -354,19 +354,22 @@ class TimelineEventsController extends PerformanceFeatureController
     debugTraceCallback(
       () => _log.info('[handleSelectedFrame]\n${frame.toStringVerbose()}'),
     );
-    await _perfettoSelectFrame(frame);
-  }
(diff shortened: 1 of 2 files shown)


packages/devtools_app/lib/src/screens/performance/panes/timeline_events/timeline_events_controller.dart
```diff
@@ -375,46 +378,34 @@ class TimelineEventsController extends PerformanceFeatureController
       hasProcessedTimelineEventsForFrame =
           perfettoController.processor.hasProcessedEventsForFrame(frame.id);
       if (!hasProcessedTimelineEventsForFrame) {
+        final timelineEventsUnavailable =
```
bkonyi, line 381: Why do we do this twice? Is it worth putting into a helper?

bkonyi (approved):

CoderDake (approved):

packages/devtools_app/lib/src/screens/performance/panes/timeline_events/timeline_events_controller.dart
kenzieschmoll: added comments in the code to explain. and added a helper.

Issue #7471, opened 2024-03-27
Error in `_file_web.dart`


kenzieschmoll:
`FileSystemWeb._files` expects values of type String but we now use a generic T.
```
devtools_app/lib/src/shared/config_specific/file/_file_web.dart:30:36: Error: The argument type 'T Function()' can't be assigned to the parameter type 'String Function()'.
      _files.putIfAbsent(filename, () => contents);
```

PR #7474, opened 2024-03-27
Fix file system tech debt

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/7471


packages/devtools_app/lib/src/shared/config_specific/import_export/_file_desktop.dart
```diff
@@ -46,7 +39,7 @@ class FileSystemDesktop implements FileIO {
     return _fs.systemTempDirectory;
   }
 
-  @override
+  /// Create file in a directory (default Downloads).
```
kenzieschmoll, line 42: dartdoc transfered from file.dart

polina-c (approved):

PR #7476, opened 2024-03-27
Switch back to dart:html for postMessage

DanTup:
This is a workaround for https://github.com/dart-lang/sdk/issues/54938 because the `package:web` version of `postMessage` causes security exceptions.

Fixes https://github.com/Dart-Code/Dart-Code/issues/5049

@kenzieschmoll @jacob314 I verified running DevTools from source had the issue shown in the bug above, and with this change it can communicate with VS Code correctly:

image

I don't know if a release has been made for the next beta already, but if so this might need to be included (I'm assuming https://github.com/dart-lang/sdk/issues/54938 won't be fixed before then.. if it is, we can ofcourse drop this).

I don't know if there are other implications of using `dart:html` here (since work had been done to remove it).

(This also reinforces that we need a better way to test all of this end-to-end against bleeding-edge code!).

### DIFF
--- a/packages/devtools_app/lib/src/shared/config_specific/copy_to_clipboard/_copy_to_clipboard_web.dart
+++ b/packages/devtools_app/lib/src/shared/config_specific/copy_to_clipboard/_copy_to_clipboard_web.dart
@@ -2,18 +2,17 @@
 // Use of this source code is governed by a BSD-style license that can be found
 // in the LICENSE file.
 
-import 'dart:js_interop';
-import 'package:web/web.dart';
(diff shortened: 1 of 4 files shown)


kevmoo:
@srujzs – anything we could do instead? Maybe inline stuff?

srujzs:
We may have a couple options here to avoid this that I want to try so we don't have to do this. Centralizing the discussion in https://github.com/Dart-Code/Dart-Code/issues/5049.

DanTup:
This was better fixed by https://github.com/flutter/devtools/pull/7485.

PR #7479, opened 2024-03-27
Migrate select event

eliasyishak:
Part of tracker issue:
- https://github.com/flutter/devtools/issues/7083

Migrates the select analytics event to package:unified_analytics

### DIFF
--- a/packages/devtools_app/lib/src/shared/analytics/_analytics_web.dart
+++ b/packages/devtools_app/lib/src/shared/analytics/_analytics_web.dart
@@ -12,6 +12,7 @@ import 'dart:async';
 import 'package:devtools_app_shared/ui.dart';
 import 'package:js/js.dart';
 import 'package:logging/logging.dart';
+import 'package:unified_analytics/unified_analytics.dart' as ua;
 import 'package:web/web.dart';
 
 import '../../../devtools.dart' as devtools show version;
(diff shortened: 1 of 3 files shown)


packages/devtools_app/lib/src/shared/analytics/_analytics_web.dart
```diff
@@ -642,18 +644,52 @@ void select(
...
+        gtagEvent.shader_compilation_duration_micros,
+    traceEventCount: gtagEvent.trace_event_count,
+    cpuSampleCount: gtagEvent.cpu_sample_count,
+    cpuStackDepth: gtagEvent.cpu_stack_depth,
+    heapDiffObjectsBefore: gtagEvent.heap_diff_objects_before,
+    heapDiffObjectsAfter: gtagEvent.heap_diff_objects_after,
+    heapObjectsTotal: gtagEvent.heap_objects_total,
+    rootSetCount: gtagEvent.root_set_count,
+    rowCount: gtagEvent.row_count,
+    inspectorTreeControllerId: gtagEvent.inspector_tree_controller_id,
```
eliasyishak, line 689: Even though we include all of the possible fields, most of these are nullable and if they are null when passed to `Event.fromJson`, those fields won't get sent to GA4

kenzieschmoll (approved):

kenzieschmoll:
Will this PR also need a bump to the new unified_analytics version 6.1.0?

eliasyishak:
Yep, once https://github.com/dart-lang/tools/pull/258 merges and publishes i'll bump the version here

PR #7484, opened 2024-03-27
Fix a broken link in the release instructions

kenzieschmoll:


polina-c (approved):

kenzieschmoll:
landing on red since this is just a readme update. We will use this hash for the patch release.

PR #7485, opened 2024-03-27
Compile devtools with `release` mode (no dart2js optimization set)

elliette:
Fixes https://github.com/Dart-Code/Dart-Code/issues/5049

Without this change, seeing infinite spinner:

Screenshot 2024-03-27 at 2 33 21 PM

With this change, Flutter sidebar loads:

Screenshot 2024-03-27 at 2 39 49 PM

### DIFF
--- a/tool/lib/commands/build.dart
+++ b/tool/lib/commands/build.dart
@@ -85,7 +85,6 @@ class BuildCommand extends Command {
     );
 
     logStatus('building DevTools in release mode');
-    final mode = buildMode == 'release' ? 'profile' : buildMode;
     await processManager.runAll(
       commands: [
         if (runPubGet) CliCommand.tool(['pub-get', '--only-main']),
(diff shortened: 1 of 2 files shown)


elliette:
@kenzieschmoll pointed out we have a lot of checks for whether we are in release mode (e.g. to collect analytics). Switching to just plain release mode without setting the dart2js optimization level.

tool/lib/commands/build.dart
```diff
@@ -95,8 +95,6 @@ class BuildCommand extends Command {
             '--web-renderer',
             'canvaskit',
             '--pwa-strategy=offline-first',
-            // Enable default optimizations: https://dart.dev/tools/dart-compile#js
-            '--dart2js-optimization=O1',
```
kenzieschmoll, line 99: do we need the `--release` arg here?

elliette: Nope, that's passed as the `buildMode` (below)

PR #7486, opened 2024-03-27
List commands for publishing.

polina-c:


tool/RELEASE_INSTRUCTIONS.md
```diff
@@ -184,14 +184,26 @@ you may need to hard reload and clear your browser cache.
 
 If `package:devtools_app_shared`, `package:devtools_extensions`, or
 `package:devtools_shared` have unreleased changes, these packages may need to be
-published to pub.
+published to pub. Publish the backages whith versions they have, even if it is
+`.dev` versions.
```
kenzieschmoll, line 188: let's remove this. This release was more of a one-off note since I upgraded the minimum SDK versions for these packages to pre-release versions. When you have prerelease versions for the dart and flutter sdk mins, you will get a warning when you publish to publish the package as a pre-release version.

tool/RELEASE_INSTRUCTIONS.md
```diff
@@ -184,14 +184,26 @@ you may need to hard reload and clear your browser cache.
...
 **Before publishing these packages, please message the DevTools Team chat room to ask if there
 are any reasons why we should wait.** Since these packages follow their own release schedules,
 it is possible that there are changes that are not ready to publish.
 
-From the respective `devtools/packages/devtools_*` directories, run:
+From the respective `devtools/packages/devtools_*` directories, run `flutter pub publish`. You can copy
+commands from here:
+
    ```shell
+   # Assuming you are in one of packages root.
```
kenzieschmoll, line 198: "...one of the directories under packages/ (e.g packages/devtools_app)"

tool/RELEASE_INSTRUCTIONS.md
```diff
@@ -184,14 +184,26 @@ you may need to hard reload and clear your browser cache.
...
+
+   cd ../devtools_app_shared
+   flutter pub publish
+
+   cd ../devtools_extensions
+   flutter pub publish
+
+   cd ../devtools_shared
```
kenzieschmoll, line 206: publish devtools_shared before devtools_app_shared, since devtools_shared is a dep of devtools_app_shared

tool/RELEASE_INSTRUCTIONS.md
polina-c: So, do we want to instruct not to publish if version is .dev? Or ask in team chat?

kenzieschmoll: There is already a comment below to ask in team chat for context around publishing these packages.

tool/RELEASE_INSTRUCTIONS.md
```diff
@@ -184,14 +184,25 @@ you may need to hard reload and clear your browser cache.
 
 If `package:devtools_app_shared`, `package:devtools_extensions`, or
 `package:devtools_shared` have unreleased changes, these packages may need to be
-published to pub.
+published to pub. Publish the backages whith versions they have, even if it is
+`.dev` versions.
```
kenzieschmoll, line 188:
```suggestion
published to pub.
```

tool/RELEASE_INSTRUCTIONS.md
polina-c: So if team said yes and versions are dev, they are still ok to publish, right? Or we want it to be said explicitly in chat? 
I want to give instructions, because dev versions look alerting and create extra mental effort for the release process.

kenzieschmoll: After thinking about this a bit more, perhaps we should have different instructions for the devtools packages that are shipped on pub. They should follow their own release cadence and each have their own considerations. For example, because package:dds depends on package:devtools_shared, we have to be careful about timing devtools_shared publishes with DevTools rolls into google3 and into the Dart SDK.

So my recommendation is to leave this part as is for now, and follow up with better docs for publishing the DevTools pub packages. Adding the commands is a nice addition though.

kenzieschmoll (approved):

PR #7488, opened 2024-03-27
Migrate to using unified_analytics from DTD

kenzieschmoll:
Work towards https://github.com/flutter/devtools/issues/7496. Depends on https://dart-review.googlesource.com/c/sdk/+/360221.

This PR also includes breaking changes to `package:devtools_shared` and bumps the version to 9.0.0.


packages/devtools_app/lib/src/shared/analytics/analytics_controller.dart
```diff
@@ -49,14 +49,13 @@ class AnalyticsController {
 
   /// Method to call to confirm with package:unified_analytics the user has
   /// seen the consent message.
-  final AsyncAnalyticsCallback? _markConsentMessageAsShown;
   Future<void> markConsentMessageAsShown() async =>
-      await _markConsentMessageAsShown?.call();
+      await dtdManager.analyticsConsentMessage();
```
CoderDake, line 53: Is the right function being called here?

```suggestion
      await dtdManager.analyticsClientShowedMessage();
```

packages/devtools_shared/lib/src/server/server_api.dart
```diff
@@ -98,13 +96,11 @@ class ServerApi {
         // Additionally, package:unified_analytics will show a message if it
         // is the first run with the package or the consent message version has
         // been updated
```
CoderDake, line 98: I think this comment needs to be updated as well

CoderDake (commented):
a couple of comments to start. I'm also curious to see the package:dtd changes

packages/devtools_app/lib/src/shared/server/_analytics_api.dart
```diff
@@ -28,7 +32,8 @@ Future<bool> isAnalyticsEnabled() async {
   if (isDevToolsServerAvailable) {
```
eliasyishak, line 32: We probably want to fix the dartdoc here and not mention the legacy file and speak generally

packages/devtools_app/lib/src/shared/server/_analytics_api.dart
```diff
@@ -12,7 +12,11 @@ Future<bool> isFirstRun() async {
...
     final resp = await request(apiGetDevToolsFirstRun);
     if (resp?.statusCode == 200) {
-      firstRun = json.decode(resp!.body);
+      // Additionally, package:unified_analytics will show a message if it
+      // is the first run with the package or the consent message version has
+      // been updated.
+      firstRun = json.decode(resp!.body) ||
+          await dtdManager.shouldShowAnalyticsConsentMessage();
```
eliasyishak, line 19: This is just for my awareness, but does `resp!.body` always equal a string of `'false'` or `'true'`?

packages/devtools_shared/lib/src/server/server_api.dart
```diff
@@ -94,17 +92,11 @@ class ServerApi {
...
       case apiGetDevToolsFirstRun:
         // Has DevTools been run first time? To bring up analytics dialog.
-        //
-        // Additionally, package:unified_analytics will show a message if it
-        // is the first run with the package or the consent message version has
-        // been updated
-        final isFirstRun =
-            _devToolsUsage.isFirstRun || analytics.shouldShowMessage;
+        final isFirstRun = _devToolsUsage.isFirstRun;
         return _encodeResponse(isFirstRun, api: api);
```
eliasyishak, line 96: Will we ever get in a state now where the legacy and new analytics systems are in disagreements with each other and possibly causing errors?

eliasyishak (commented):
First pass of review, I want to also check this out locally and test the workflows tomorrow

packages/devtools_app/lib/src/shared/server/_analytics_api.dart
kenzieschmoll: For this API endpoint, yes. You can look at the handler for this in server_api.dart.

packages/devtools_app/lib/src/shared/server/_analytics_api.dart
kenzieschmoll: Changed things around a bit so we don't interact with the DTD unified analytics server at all from the DevTools server wrapper. These server APIs are now exclusively for the legacy analytics.

packages/devtools_shared/lib/src/server/server_api.dart
kenzieschmoll: There should not be any logic changes WRT analytics in this PR; only changes to where the unified analytics logic is accessed (by proxy of DTD now). If you see a case where there is a logic change, this is not intentional so please flag it.

packages/devtools_app/lib/src/shared/server/_analytics_api.dart
eliasyishak: ahh i see now, that makes sense, that's a clean solution!

packages/devtools_app/lib/src/shared/analytics/_analytics_web.dart
```diff
@@ -425,16 +427,42 @@ class GtagExceptionDevTools extends GtagException {
...
+Future<bool> shouldShowAnalyticsConsentMessage() async {
+  bool shouldShow = false;
+  if (kReleaseMode) {
+    // When asked if the consent message should be shown,
+    // package:unified_analytics will return true if this the user's first run
+    // (unlikely for the case of DevTools since the user is likely to have
+    // already seen the message from another Dash tool (Dart or Flutter CLI,
+    // IDE, etc.) or when the consent message version has been updated.
+    shouldShow = await dtdManager.shouldShowAnalyticsConsentMessage();
+  }
```
eliasyishak, line 457: This comment may be wrong, even if a user has already interacted with package:unified_analytics through the flutter-cli, we will still have to show the consent message because it is the first time it was run on devtools.

So the package will return true if this is the first time we added devtools into the config file for UA

packages/devtools_app/lib/src/shared/analytics/analytics_controller.dart
```diff
@@ -6,27 +6,54 @@ import 'dart:async';
...
+
+  // TODO(https://github.com/flutter/devtools/issues/7083): when the legacy
+  // analytics are fully removed, this try-catch is unnecessary because we will
+  // get these values directly from [dtdManater], like how the consentMessage
+  // parameter is specified below.
+  var enabled = false;
+  var shouldShowConsentMessage = false;
+  try {
+    enabled = await ga.isAnalyticsEnabled();
+    shouldShowConsentMessage = await ga.shouldShowAnalyticsConsentMessage();
```
eliasyishak, line 25: Do we have any legal concerns about showing the consent message for UA even if it was the legacy analytics that reported we should be showing the consent message?

This would only happen if UA reports false under `ga.shouldShowAnalyticsConsentMessage` but legacy reports true since this function checks both analytics systems.

eliasyishak:
I mentioned over chat, but i did some testing of this code after the below steps:
1. Go to my local dart sdk and make an update to dds pubspec to point to my local `devtools_shared`
```yaml
  devtools_shared:
    path: /Users/eliasyishak/Desktop/devtools/packages/devtools_shared
```
2. Update `pkg/dds/lib/src/devtools/handler.dart` so that i remove the `analytics` parameter from the `ServerApi.handle` callsite
3. From the `devtools` repo, i ran `devtools_tool sync`
4. From the `devtools` repo, i ran `devtools_tool serve` --> this launches the web app as expected but I don't see any consent message, and this is after confirming that devtools has not been added to the local config file
image

If i'm missing any step for the local testing, please lmk and i can try it but is it working for you? You should be able to run `rm ~/.dart-tool/dart-flutter-telemetry.config` to ensure that it will be the first run to prompt the consent message

packages/devtools_app/lib/src/shared/analytics/analytics_controller.dart
kenzieschmoll: This does not change any behavior from what we had before this PR. Should we instead stop checking the legacy value for this one? We don't show the legacy message anymore, so maybe we should just defer to the UA logic for when to show the message. WDYT?

kenzieschmoll:
The issue described is due to a limitation with our `devtools_tool serve` command. That is fixed here: https://github.com/flutter/devtools/pull/7532. This will help with testing changes that have not made their way into Flutter master yet.

packages/devtools_app/lib/src/shared/analytics/_analytics_web.dart
kenzieschmoll: Updated comment.

eliasyishak (approved):
Locally testing this worked as expected! LGTM

Issue #7424, opened 2024-03-22
Misleading snapshot sizes


kenzieschmoll:
When looking at a list of snapshots with different sizes:
![Screenshot 2024-03-22 at 10 41 03 AM](https://github.com/flutter/devtools/assets/43759233/d1f191d9-2eaa-4b04-a345-d8044ebfe5ae)

I would expect diffing two of these snapshots to show a meaningful comparison highlighting the size difference in the two snapshots. Instead, I got empty diffs (no filters applied). This was because the reachable size of these snapshots was the same:
![Screenshot 2024-03-22 at 10 41 07 AM](https://github.com/flutter/devtools/assets/43759233/ec653803-9ed4-453f-9d7a-53a94bff7f23)
![Screenshot 2024-03-22 at 10 41 12 AM](https://github.com/flutter/devtools/assets/43759233/6fa360c7-c839-442e-a46a-baf054313b5d)

I confirmed that triggering a garbage collection brought the heap size down to the reachable size:
![Screenshot 2024-03-22 at 10 43 14 AM](https://github.com/flutter/devtools/assets/43759233/bdd7a8a8-98f2-42c0-b55a-ae7ceb033ba7)

We should consider showing reachable size in the snapshot list so that the user does not get confused why there is a size difference between two snapshots but an unexpected empty diff.

PR #7493, opened 2024-03-29
Show reachable size in list.

polina-c:
Fixes: https://github.com/flutter/devtools/issues/7424

### DIFF
--- a/packages/devtools_app/release_notes/helpers/release_notes_template.md
+++ b/packages/devtools_app/release_notes/helpers/release_notes_template.md
@@ -1,6 +1,12 @@
 This is draft for future release notes, that are going to land on
 [the Flutter website](https://docs.flutter.dev/tools/devtools/release-notes).
 
+Template for a release notes item:
+
+
+Put images to the folder `images`.
+
+
 # DevTools <number> release notes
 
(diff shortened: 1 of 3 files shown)


CoderDake (approved):

packages/devtools_app/release_notes/helpers/release_notes_template.md
```diff
@@ -1,6 +1,17 @@
...
+
+------------
+* Enabled export of snapshots and improved snapshotting performance. -
+[#7197](https://github.com/flutter/devtools/pull/7197),
+[#7439](https://github.com/flutter/devtools/pull/7439),
+[#7449](https://github.com/flutter/devtools/pull/7449)
+
+    ![Export snapshot](/tools/devtools/release-notes/images-2.34.1/7197-export.png "Export snapshot")
+------------
+
```
kenzieschmoll, line 14: why are we adding this to the template? this will get copied over to NEXT_RELEASE_NOTES.md each time we have a new set of release notes, which we don't wan't.

kenzieschmoll: the release notes readme might be a better place for an example like this. It already has an example for adding an image.

PR #7522, opened 2024-04-04
Add a `debugDtdUri` development helper and clean up memory debug flag

kenzieschmoll:


polina-c (approved):
Lgtm with one comment.

kenzieschmoll:
> In fact all flags here are for debugging and can be moved to development_helpers as well:
> 
> ```
>   static Future<HeapData> calculate(
>     HeapSnapshotGraph graph,
>     DateTime created, {
>     @visibleForTesting bool calculateRetainingPaths = true,
>     @visibleForTesting bool calculateRetainedSizes = true,
>     @visibleForTesting bool calculateClassData = true,
>   })
> ```

These flags look like they are only used in tests, so seems like they are properly tagged as `visibleForTesting`. The development_helpers flags are more for things that will be temporarily enabled during development, but should never be checked in in an "enabled" state.

Issue #7523, opened 2024-04-04
DTD select project view does not layout well for narrow views


kenzieschmoll:
This makes the IDE embedded view difficult to use:
![Screenshot 2024-04-04 at 9 51 11 AM](https://github.com/flutter/devtools/assets/43759233/1f6321ea-fbd1-45bb-b1ad-28edac8a7320)

PR #7524, opened 2024-04-04
Fix deep links project selection view for narrow screens

kenzieschmoll:
Fixes https://github.com/flutter/devtools/issues/7523

select-project-view


hangyujin (approved):
37737237-66e9c9dc-2d10-11e8-8797-13a405eace49

PR #7525, opened 2024-04-04
Cleanup old offline memory code.

polina-c:


packages/devtools_app/lib/src/shared/feature_flags.dart
```diff
@@ -81,7 +81,7 @@ abstract class FeatureFlags {
   /// well.
   static final _allFlags = <String, bool>{
     'widgetRebuildStats': widgetRebuildStats,
-    'memoryAnalysis': memoryAnalysis,
+    'memoryAnalysis': memoryOffline,
```
kenzieschmoll, line 84: nit rename to 'memoryOffline' to match

kenzieschmoll (approved):

PR #7526, opened 2024-04-04
Enable experiment as constant.

polina-c:


packages/devtools_app/lib/src/shared/screen.dart
```diff
@@ -55,6 +56,8 @@ enum ScreenMetaData {
     icon: Octicons.package,
     requiresDartVm: true,
     tutorialVideoTimestamp: '?t=420',
+    // ignore: avoid_redundant_argument_values, false positive
+    worksWithOfflineData: kMemoryOfflineExperiment,
```
kenzieschmoll, line 60: use `FeatureFlags.memoryOffline` instead of the "k" constant

packages/devtools_app/lib/src/shared/feature_flags.dart
```diff
@@ -35,6 +35,9 @@ void setEnableExperiments() {
 @visibleForTesting
 bool get enableBeta => enableExperiments || !isExternalBuild;
 
+const bool kMemoryOfflineExperiment =
```
kenzieschmoll, line 38: make this private.

packages/devtools_app/lib/src/shared/screen.dart
polina-c: Enums require constants. FeatureFlags.memoryOffline is not constant.

packages/devtools_app/lib/src/shared/feature_flags.dart
polina-c: It is needed in screen.dart

packages/devtools_app/lib/src/shared/screen.dart
kenzieschmoll: make memoryOffline constant and then you should be able to use it.

packages/.vscode/launch.json
```diff
@@ -22,7 +22,9 @@
             "type": "dart",
             "program": "devtools_app/lib/main.dart",
             "args": [
-                "--dart-define=enable_experiments=true"
+                "--dart-define",
+                "memory_offline_experiment=true",
```
kenzieschmoll, line 26: does this work? we might need a comma after '=true'

kenzieschmoll (approved):
one question, but lgtm

packages/.vscode/launch.json
polina-c: It worked. But it seems there is better way: https://dartcode.org/docs/using-dart-define-in-flutter/

PR #7529, opened 2024-04-05
Convert relative site URLs in release notes

parlough:
We no longer use `{{site.url}}` in Markdown files on the site to enable easier contributing, so `{{site.url}}` doesn't work as a replacement target anymore. Instead the site-relative image links are formatted as `alt text`. Same goes for other non-image links.

For DevTools to handle this case, this PR uses a regex to recognize Markdown link format to insert the docs site URL or the current debug URL. I used `replaceAllMapped` over lookahead/lookbehind assertions due to browser support still being limited.


kenzieschmoll:
Thanks for the PR! Have you verified that the markdown viewer in DevTools still works as expected? (for both the local debugging case and for loading notes from the flutter website in production)? You can click the release notes button here to trigger opening the viewer:
Screenshot 2024-04-05 at 1 27 52 PM

packages/devtools_app/lib/src/framework/release_notes/release_notes.dart
```diff
@@ -188,13 +191,21 @@ class ReleaseNotesController extends SidePanelController {
     return;
   }
 
+  /// Convert all site-base-url relative links in [markdownContent]
+  /// to absolute links from the specified [baseUrl].
```
kenzieschmoll, line 195: nit: could you add an example transformation to this dart doc? That would help with maintaining this code moving forward if things were to change.

kenzieschmoll (approved):
one comment, but lgtm

Issue #7240, opened 2024-02-17
Devtool Extension does not load in Devtools for git dependencies (with build generated)


mkorbel1:
We are attempting a workflow where we use our CD flow in GitHub Actions to build the artifacts for our devtool extension on rohd and push them to a separate branch.  The goal is that users can point to a pre-release version of our package hosted on git but still benefit from the devtools extension by pointing to this special branch instead of main.  But, we're not seeing the devtools extension show up in devtools when we point it there.

If helpful, this PR (https://github.com/intel/rohd/pull/461) has the relevant changes we're making, and here's a dependency override that we expected would work (on a fork, for now):
```yaml
dependency_overrides:
  rohd:
    git:
      url: https://github.com/quekyj/rohd.git
      ref: artifacts
```

You can see all the build output located here: https://github.com/quekyj/rohd/tree/artifacts/extension/devtools/build

This isn't working for me on `Dart SDK version: 3.3.0 (stable) (None) on "linux_x64"` (in windows 11 WSL2 ubuntu)
Here's a tiny example program that you can set a breakpoint on after the call to `.build()`, and the extension should show up in the devtools and show the built module within it

```dart
import 'package:rohd/rohd.dart';

class MyMod extends Module {
  MyMod(Logic x) {
    x = addInput('x', x);
  }
}

void main(List<String> arguments) async {
  final m = MyMod(Logic());
  await m.build();

  print(m.generateSynth());
}
```

This is a follow-up to this Discord discussion: https://discord.com/channels/608014603317936148/1159561514072690739/1206721049430069248

mkorbel1:
> After you run `dart pub get` on the package with the git dependency, can you check your `.dart_tool/package_config.json` file for the `rohd` entry. The pub cache location it points to is where DevTools looks for the `rohd/extension/devtools` folder. Do you see the content you expect in that location?

Yes, I see the built devtools extension in the downloaded git dependency

```json
{
  "name": "rohd",
  "rootUri": "file:///home/max/.pub-cache/git/rohd-aba2c2ffac83c318615160b4172aa771f8422d45/",
  "packageUri": "lib/",
  "languageVersion": "2.19"
},
```

```bash
$ ls /home/max/.pub-cache/git/rohd-aba2c2ffac83c318615160b4172aa771f8422d45/extension/devtools/build/
assets  canvaskit  favicon.png  flutter.js  flutter_service_worker.js  icons  index.html  main.dart.js  manifest.json  version.json
```

[...]

PR #7535, opened 2024-04-05
Use DTD to detect the package root for a file uri string

kenzieschmoll:
This will prevent bugs like https://github.com/flutter/devtools/issues/7240, where an extension failed to be detected because it is under a subdirectory that we do check in the regular expression.


packages/devtools_shared/test/utils/file_utils_test.dart
```diff
@@ -2,81 +2,183 @@
...
-      expect(
-        packageRootFromFileUriString(
-          'file:///Users/me/foo/my_app_root/integration_test/sub/some_test.dart',
+    setUp(() async {
+      dtd = await startDtd();
+      expect(dtd!.uri, isNotNull, reason: 'Error starting DTD for test');
+      testDtdConnection = await DartToolingDaemon.connect(Uri.parse(dtd!.uri!));
+
+      // Set up the project root.
+      final tmpDirectory = Directory.systemTemp.createTempSync();
```
bkonyi, line 43: Consider adding a comment describing the directory structure. It's hard to visualize just based on the code.

kenzieschmoll:
@DanTup can you take a look at this and see if I'm introducing any Windows-related bugs with this change?

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -212,10 +212,17 @@ class ServiceConnectionManager {
 
   // TODO(kenz): consider caching this value for the duration of the VM service
   // connection.
+  /// Returns the root package directory for the main isolate.
+  ///
+  /// If a non-null value is returned, the value will be a file URI String and
+  /// it will NOT have a trailing slash.
```
DanTup, line 218: Is it worth adding asserts for this so if it happens during tests it would throw?

packages/devtools_shared/lib/src/utils/file_utils.dart
```diff
@@ -2,15 +2,78 @@
...
+      // URI scheme.
+      // Include the trailing slash because the IDE workspace root this URI is
+      // a part of may have a trailing slash, and DTD checks that the URI we
+      // pass to `listDirectoryContents` starts with at least one IDE workspace
+      // root.
+      var pathString = p.joinAll(parts);
+      if (!pathString.startsWith('/')) pathString = '/$pathString';
+      if (!pathString.endsWith('/')) pathString = '$pathString/';
+
+      final currentUri = Uri.file(pathString);
```
DanTup, line 52: This doesn't seem right for Windows, we shouldn't be calling `Uri.file()` with a string that contains `/` (or starts with) for Windows paths.

Since we're sending a URI to DTD anyway, I think we could just manipulate the URIs something like:

```dart
while (uri.pathSegments.length > 1) {
  // Take the last path segment off
  uri = uri.replace(pathSegments: uri.pathSegments.sublist(0, uri.pathSegments.length - 1));

  // uri is now the parent directory
}
```

That way we don't need to worry about which kind of paths they are (so if this code runs in the browser, we don't need to worry that it will assume non-Windows).

packages/devtools_shared/lib/src/utils/file_utils.dart
```diff
@@ -2,15 +2,78 @@
...
+      } catch (e) {
+        // Fail gracefully on exception, and proceed to using heuristic below.
+        exception = e;
+        break;
+      }
+    }
+    if (throwOnDtdSearchFailed) {
+      throw Exception('Expected DTD to detect the package root, but it failed'
+          '${exception != null ? ' with exception: $exception' : '.'}');
+    }
```
DanTup, line 72: You probably shouldn't, but if you call this method with `throwOnDtdSearchFailed=true` and `dtd==null` I think it _won't_ throw (because this is inside the `if` condition), but maybe it should because that's using it incorrectly?

kenzieschmoll: this is intentional because this parameter is used to verify that we can properly find the package using only the DTD path. We don't want to throw if the regexp search fails, because we will already be able to verify that by the value returned.

packages/devtools_shared/lib/src/utils/file_utils.dart
kenzieschmoll: Applied requested changes, PTAL.

packages/devtools_shared/lib/src/utils/file_utils.dart
DanTup: Oops, I missed the `@visibleForTesting` on that parameter and thought it might be used by other code :)

packages/devtools_shared/lib/src/utils/file_utils.dart
kenzieschmoll: Thanks, yes if you wouldn't mind running all devtools_shared/test on windows with this change, that would be great, thank you!

packages/devtools_shared/lib/src/utils/file_utils.dart
```diff
@@ -2,20 +2,88 @@
...
+/// * bin
+/// * integration_test
+/// * test
+/// * benchmark
+/// * example
+///
+/// The URI returned will be a file URI String and will NOT have a trailing
+/// slash.
+Future<String> packageRootFromFileUriString(
+  String fileUriString, {
```
bkonyi, line 29: Is this always assumed to be a `file://` URI? Do we know it is? `Uri.parse` will interpret Windows drive letters as the scheme if it isn't prefixed with `file://`.

packages/devtools_shared/test/utils/file_utils_test.dart
```diff
@@ -2,81 +2,216 @@
...
       );
-      expect(
-        packageRootFromFileUriString(
-          'file:///Users/me/foo/my_app_root/benchmark/sub/some_test.dart',
-        ),
-        equals('file:///Users/me/foo/my_app_root'),
+      expect(result, equals(expected ?? projectRoot));
+    }
+
+    for (final useDtd in [true, false]) {
```
bkonyi, line 70: Ubernit: this list can be const

bkonyi (approved):

packages/devtools_shared/lib/src/utils/file_utils.dart
kenzieschmoll: It is expected to be a file URI. I can add an assert at the top of this method to make sure.

kenzieschmoll: I added an assertion and a test case.

packages/devtools_shared/lib/src/utils/file_utils.dart
DanTup: As-is I see a bunch of errors trying to clean delete previous test apps:

```
00:02 +38 -1: test/server\general_api_test.dart: General DevTools server API endpoints that require DTD detectRootPackageForVmService succeeds for a connect event [E]
  PathAccessException: Deletion failed, path = 'tmp/test_app' (OS Error: The process cannot access the file because it is being used by another process.
  , errno = 32)
  dart:io                                   FileSystemEntity.deleteSync
  test\helpers.dart 111:43                  TestDartApp.kill
  ===== asynchronous gap ===========================
  test\server\general_api_test.dart 176:11  main.<fn>.<fn>.<fn>.<fn>
```

This isn't uncommon on Windows - things are often locked for a while after a process dies and many of our tools retry these kinds of deletes.

I put a loop around the delete until it succeeded and everything passes:

```
PS C:\Dev\Google\devtools\packages\devtools_shared> ..\..\tool\flutter-sdk\bin\dart test test/
Resolving dependencies in `C:\Dev\Google\devtools\packages\devtools_shared`...
Downloading packages...
Got dependencies in `C:\Dev\Google\devtools\packages\devtools_shared`.
Building package executable...
Built test:test.
00:00 +22: test/server\deeplink_api_test.dart: DeeplinkApi handle deeplink api api/androidBuildVariants returns bad request if no root path
[api/androidBuildVariants] missing required query parameters: [rootPath]
00:00 +31: test/server\general_api_test.dart: General DevTools server API api/notifyForVmServiceConnection returns badRequest for invalid VM service argument
Cannot normalize VM service URI: fake_uri
00:00 +34: test/server\general_api_test.dart: General DevTools server API api/notifyForVmServiceConnection returns badRequest for invalid connected argument
Cannot parse connected parameter:
FormatException: Invalid boolean
bad_arg
00:02 +38: test/server\general_api_test.dart: General DevTools server API endpoints that require DTD detectRootPackageForVmService succeeds for a connect event
Failed to delete old test app on attempt 2... retrying...
Failed to delete old test app on attempt 3... retrying...
Failed to delete old test app on attempt 4... retrying...
Failed to delete old test app on attempt 5... retrying...
Failed to delete old test app on attempt 6... retrying...
00:08 +39: test/server\general_api_test.dart: General DevTools server API endpoints that require DTD detectRootPackageForVmService succeeds for a disconnect event when cache is empty
Failed to delete old test app on attempt 2... retrying...
00:10 +40: test/server\general_api_test.dart: General DevTools server API endpoints that require DTD detectRootPackageForVmService succeeds for a disconnect event when cache contains entry for VM service
Failed to delete old test app on attempt 2... retrying...
Failed to delete old test app on attempt 3... retrying...
Failed to delete old test app on attempt 4... retrying...
Failed to delete old test app on attempt 5... retrying...
Failed to delete old test app on attempt 6... retrying...
Failed to delete old test app on attempt 7... retrying...
00:17 +41: All tests passed!
```

I'll make a note to open a PR with a retry like this so it doesn't need to be done ad-hoc to run tests.

DanTup: I opened https://github.com/flutter/devtools/pull/7544 with a retry for the deletes so we only fail if we can't delete for some period (currently 10s).

PR #7539, opened 2024-04-08
Clean up styling for analytics prompt

kenzieschmoll:
This PR:
- adjusts the font sizing and button styling to better match with DevTools theme defaults
- makes the text selectable so users can copy the disable command from the prompt
- adjusts the new line characters in the consent message so that the text is responsive for different screen widths

Screenshot 2024-04-08 at 11 01 23 AM


eliasyishak (approved):
LGTM, clever use of look back/ahead regex to combine single new lines!

PR #7540, opened 2024-04-08
Add getter for devtools mode.

polina-c:


packages/devtools_app/lib/src/shared/primitives/simple_items.dart
```diff
@@ -57,3 +57,14 @@ enum DocLinks {
...
+enum DevToolsMode {
+  /// No connection to app and no data from a previous session.
+  disconnected,
+
+  /// Interacting with a connected application.
+  connected,
+
+  /// Showing data saved from a previous session.
+  offlineData,
+}
```
kenzieschmoll, line 70: hmm. This still seems inaccurate / misleading because these are not disjoint modes. DevTools can be disconnected and showing offline data or connected and showing offline data.

packages/devtools_app/lib/src/shared/globals.dart
```diff
@@ -87,3 +88,12 @@ void setStagerMode() {
...
 }
+
+/// Current mode of DevTools.
+DevToolsMode get devToolsMode {
+  return offlineDataController.showingOfflineData.value
+      ? DevToolsMode.offlineData
+      : serviceConnection.serviceManager.hasConnection
+          ? DevToolsMode.connected
+          : DevToolsMode.disconnected;
+}
```
kenzieschmoll, line 99: where will this be used? This seems like it is more reflective of the DevTools "screen" mode than it is the global state of DevTools. Can we move this to the screen class if that is where this getter is going to be used?

polina-c: It will be used in controller's constructor.

packages/devtools_app/lib/src/shared/primitives/simple_items.dart
polina-c: Updated comment for offlineData.

DevToolsMode is not about existence of connection, but about what DevTools is doing.

packages/devtools_app/lib/src/shared/globals.dart
polina-c: May be it is better to put in its own lib?

kenzieschmoll: Maybe put this in `utils.dart` instead of globals

polina-c: done

kenzieschmoll (approved):

Issue #7547, opened 2024-04-09
Failure serving DevTools extensions when DevTools build is not writable


kenzieschmoll:
This happens when using vendored Dart SDKs on Windows (like when getting dart through `apt-get`).

```
[3:44:32 PM] [DevTools] [Info] <== PathAccessException: Creation failed, path = '/usr/lib/dart/bin/resources/devtools/devtools_extensions' (OS Error: Permission denied, errno = 13)
[3:44:32 PM] [DevTools] [Info] <== PathAccessException: Creation failed, path = '/usr/lib/dart/bin/resources/devtools/devtools_extensions' (OS Error: Permission denied, errno = 13)
```

Proposed solution. Stop copying extension assets to the DevTools `build/` directory, and instead serve extensions from their absolute location in the pub cache.

kenzieschmoll:
The changes for this fix have landed in flutter/devtools and in the Dart SDK

PR #7548, opened 2024-04-09
Serve DevTools extensions from pub cache instead of copying assets.

kenzieschmoll:
Work towards fixing https://github.com/flutter/devtools/issues/7547. TL;DR is that copying assets will fail when the DevTools build directory is not writable. We do not need to copy assets, but can instead serve the extension assets directly from their location on disk.

This change requires https://dart-review.googlesource.com/c/sdk/+/362062, so to roll this through, we will:
1. Land this PR
2. Publish `devtools_shared` 9.0.0
3. Add a DEPS update in https://dart-review.googlesource.com/c/sdk/+/362062 that includes this change.
4. Land https://dart-review.googlesource.com/c/sdk/+/362062.

@DanTup


DanTup (approved):
lgtm!

Tested on Windows along with the SDK change using `dart.customDevTools` and the rohd extension and everything still works.

bkonyi (approved):

PR #7564, opened 2024-04-11
Return available extensions if any were successfully served

kenzieschmoll:
This ensures that we return any extensions that were successfully detected and served, even if an exception was thrown.


packages/devtools_shared/lib/src/server/handlers/_devtools_extensions.dart
```diff
@@ -23,24 +23,38 @@ abstract class _ExtensionsApiHandler {
...
       await extensionsManager.serveAvailableExtensions(rootPath, logs);
     } on ExtensionParsingException catch (e) {
       // For [ExtensionParsingException]s, we should return a success response
       // with a warning message.
       result[ExtensionsApi.extensionsResultWarningPropertyName] = e.message;
     } catch (e) {
-      // For all other exceptions, return an error response.
+      // If any extensions were successfully detected, return a success response
+      // with a warning.
+      if (extensionsManager.devtoolsExtensions.isNotEmpty) {
```
polina-c, line 51: Should it be test covered?

kenzieschmoll: Added tests.

CoderDake (approved):

PR #7572, opened 2024-04-12
Reorganize chart pane controller code and prepare structure for offline mode.

polina-c:


packages/devtools_app/lib/src/screens/memory/panes/chart/controller/memory_tracker.dart
```diff
@@ -85,9 +61,34 @@ class MemoryTracker {
     _updateGCEvent(event.isolate!.id!, memoryUsage);
   }
 
-  void _pollMemory() async {
-    _pollingTimer = null;
+  void onMemoryData(Event data) {
```
polina-c, line 64: This piece is moved as is from MemoryChartPaneController.

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
```diff
@@ -65,26 +75,82 @@ class MemoryController extends DisposableController
...
+          profilePaneController: connectedProfile,
+        );
+      case DevToolsMode.offlineData:
+        assert(connectedDiff == null && connectedProfile == null);
+        await maybeLoadOfflineData(
+          PerformanceScreen.id,
+          createData: (json) => OfflineMemoryData.parse(json),
+          shouldLoad: (data) => !data.isEmpty,
+        );
+        // If shouldLoad returns false, previous line is noop, so data should be initialized.
```
kenzieschmoll, line 98: nit: can we make this comment more clear 
"[maybeLoadOfflineData] will be a noop if we should not load offline data for the memory screen, so ensure we still call [_initializedData] if it has not been called."

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -53,7 +64,7 @@ class DiffPaneController extends DisposableController {
       defaultName: selectedIsolateName ?? '<isolate-not-detected>',
     );
 
-    await _addSnapshot(_heapGraphLoader, item);
+    await _addSnapshot(loader, item);
```
kenzieschmoll, line 67: nit: no need to store loader in a local variable. Just null assert it here `loader!`

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_pane_controller.dart
```diff
@@ -15,6 +15,18 @@ import 'model.dart';
 
 class ProfilePaneController extends DisposableController
     with AutoDisposeControllerMixin {
+  ProfilePaneController();
+
+  factory ProfilePaneController.parse(Map<String, dynamic> map) {
```
kenzieschmoll, line 20: use Object? instead of dynamic. Here and elsewhere in this PR

packages/devtools_app/lib/src/screens/memory/shared/heap/class_filter.dart
```diff
@@ -50,11 +59,30 @@ class ClassFilter {
...
+      only: json[_Json.only] as String?,
+    );
+  }
+
+  Map<String, dynamic> prepareForOffline() {
+    return {
+      _Json.type: filterType.name,
+      _Json.except: except,
+      _Json.only: only,
+    };
```
kenzieschmoll, line 83: use an extension type for the json parsing in this class. Search "extension type" in the codebase for examples

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
```diff
@@ -65,26 +75,82 @@ class MemoryController extends DisposableController
...
-  DiffPaneController _createDiffController() =>
-      DiffPaneController(HeapGraphLoaderRuntime(chart.memoryTimeline));
+  Future<void> _init(
+    @visibleForTesting DiffPaneController? connectedDiff,
+    @visibleForTesting ProfilePaneController? connectedProfile,
+  ) async {
+    assert(!isInitialized.value);
+    switch (_mode) {
+      case DevToolsMode.disconnected:
+        throw StateError('Memory screen does not support disconnected mode.');
```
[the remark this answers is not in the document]
polina-c: Supported it.
Still need to connect to screen via override.

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
```diff
@@ -0,0 +1,59 @@
...
+    this.diff,
+    this.profile,
+    this.chart,
+    this.filter, {
+    this.isEmpty = false,
+    required this.selectedTab,
+  });
+
+  factory OfflineMemoryData.parse(Map<String, dynamic> json) {
+    Map<String, dynamic> item(String key) =>
```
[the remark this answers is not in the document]
polina-c: Added TODO

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
```diff
@@ -0,0 +1,59 @@
...
+  final bool isEmpty;
+
+  final int selectedTab;
+  final ClassFilter filter; // filter is shared between tabs, so it's here
+
+  final DiffPaneController diff;
+  final ProfilePaneController profile;
+  final MemoryChartPaneController chart;
+
+  Map<String, dynamic> prepareForOffline() {
```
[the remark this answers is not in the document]
polina-c: It is consistent with OfflineMemoryData. 

I would update both with what is recommended for serialization: https://docs.flutter.dev/data-and-backend/serialization/json#serializing-json-inside-model-classes

How does it sound?

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
```diff
@@ -0,0 +1,59 @@
...
+  factory OfflineMemoryData.parse(Map<String, dynamic> json) {
+    Map<String, dynamic> item(String key) =>
+        json[key] as Map<String, dynamic>? ?? {};
+    return OfflineMemoryData(
+      DiffPaneController.parse(item(_Json.diffData)),
+      ProfilePaneController.parse(item(_Json.profileData)),
+      MemoryChartPaneController.parse(item(_Json.chartData)),
+      ClassFilter.parse(item(_Json.classFilter)),
+      selectedTab: json[_Json.selectedTab] as int? ?? 0,
+      isEmpty: json.isEmpty,
```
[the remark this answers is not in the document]
polina-c: It is easier to process data than to detect if it is empty. To many defaults to verify they are equal to the values. So, for memory screen I will always return false.

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
polina-c: not relevant any more

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_pane_controller.dart
polina-c: It is consistent with what is recommended for serialization: https://docs.flutter.dev/data-and-backend/serialization/json#serializing-json-inside-model-classes

Map<String, dynamic> is a standard for json.

packages/devtools_app/lib/src/screens/memory/shared/heap/class_filter.dart
polina-c: Added TODO.

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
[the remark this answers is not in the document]
kenzieschmoll: If is recommended to never use `dynamic`. We should be using `Object?` instead. These have different implications in the compiler and for type checking. Couldn't find official docs but here's an issue with some discussion on this topic: https://github.com/dart-lang/language/issues/3192

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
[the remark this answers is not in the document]
kenzieschmoll: Fine to return false for now, but can you add a TODO to implement this in the future? Once all the serialization work is done, we should have a simple way to do this.

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_pane_controller.dart
kenzieschmoll: Commented above. Those docs look outdated.

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
```diff
@@ -91,22 +96,21 @@ class MemoryController extends DisposableController
...
         await maybeLoadOfflineData(
-          PerformanceScreen.id,
+          ScreenMetaData.memory.id,
           createData: (json) => OfflineMemoryData.parse(json),
-          shouldLoad: (data) => !data.isEmpty,
+          shouldLoad: (data) => true,
         );
-        // If shouldLoad returns false, previous line is noop, so data should be initialized.
-        if (!isInitialized.value) _initializeData();
+        assert(_initialized.isCompleted);
```
kenzieschmoll, line 103: this will not be completed if maybeLoadOfflineData was a noop

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
```diff
@@ -91,22 +96,21 @@ class MemoryController extends DisposableController
...
+    assert(_initialized.isCompleted);
   }
 
   void _initializeData({
     OfflineMemoryData? offlineData,
     @visibleForTesting DiffPaneController? diffPaneController,
     @visibleForTesting ProfilePaneController? profilePaneController,
   }) {
-    assert(!isInitialized.value);
+    assert(!_initialized.isCompleted);
```
kenzieschmoll, line 113: we don't need this assert. `_initialized.complete();` will fail if the completer has already been completed

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
```diff
@@ -21,10 +21,11 @@ class OfflineMemoryData {
     this.profile,
     this.chart,
     this.filter, {
-    this.isEmpty = false,
     required this.selectedTab,
   });
 
+  // TODO: use an extension type for the Json parsing, https://github.com/flutter/devtools/issues/6972
```
kenzieschmoll, line 27: is this the right issue?

packages/devtools_app/lib/src/screens/memory/framework/screen_body.dart
```diff
@@ -52,7 +63,7 @@ class _ConnectedMemoryBodyState extends State<ConnectedMemoryBody>
 
   @override
   Widget build(BuildContext context) {
-    if (!memoryController.isInitialized.value) {
+    if (!_initialized) {
```
kenzieschmoll, line 66: use a FutureBuilder instead so we don't need to add the boiler plate above. See other screens for examples.

packages/devtools_app/lib/src/screens/memory/shared/heap/class_filter.dart
```diff
@@ -75,6 +75,8 @@ class ClassFilter {
     );
   }
 
+  // TODO: use an extension type for the Json parsing, https://github.com/flutter/devtools/issues/6972
```
kenzieschmoll, line 78: same comment. incorrect issue link?

packages/devtools_app/lib/src/screens/memory/framework/screen_body.dart
kenzieschmoll: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/profiler/profiler_screen.dart/#L122

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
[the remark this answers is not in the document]
polina-c: I suggest to do it as separate clean up, including `OfflineMemoryData.parse(Map<String, dynamic> json)` and others.
I expect unexpected side effects that may cause us to change our mind here. I remember trying to use Object for json instead of dynamic and facing some difficulties that convinced me to stay with standard for json.

Serialization is a specific mechanism that picks up all objects that meet this standard:  https://docs.flutter.dev/data-and-backend/serialization/json#serializing-json-inside-model-classes

This mechanism may break with switch to Object.

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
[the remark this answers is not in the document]
polina-c: Why it is important to have empty data as a special case? If data is empty, processing of data will not save any performance, it will be fast.
But, in case of memory screen it will add code complexity. For example, I will need to verify that filter value is default. And there are many others like this.

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
polina-c: `_initialized.complete()` happens 20 lines later. 
This assert helps with readability as it shows precondition of the method.

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
polina-c: Yes, I want to have it done before completing the disconnect experience. Most likely next PR.

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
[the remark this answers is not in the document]
kenzieschmoll: We wouldn't need to check the filter value for this case. A good check here would be that chart data is empty, profile data is empty, and diff snapshots is empty. The offline import framework does a check if data is empty and shows an error notification. It's not critical, but would be nice to have for consistency.

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
kenzieschmoll: I don't think it would if the wrapped if check returns false `offlineDataController.shouldLoadOfflineData(screenId)`

packages/devtools_app/lib/src/screens/memory/shared/heap/class_filter.dart
polina-c: looks correct for me

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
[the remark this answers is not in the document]
polina-c: Hm.
If user did not do anything with screen, but crashed and selected to see historical data they will see error?
No filter? What if user typed something smart into filter and crashed, and just wants to copy/paste this value?

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_pane_controller.dart
polina-c: Commented above.

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
polina-c: Fixed.

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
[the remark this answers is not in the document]
polina-c: Actually, memory screen is never empty, because user always see profile data.

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_pane_controller.dart
polina-c: If docs are outdated, how about first updating the doc and then starting following the updated guidance?

kenzieschmoll (approved):
lgtm with one nit

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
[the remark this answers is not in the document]
kenzieschmoll: @srawlins do you know if the linked docs above are up to date? I know you've done a pretty large overhaul of all the JSON parsing we were doing in DevTools using extension types. Can you advise here WRT using `dynamic` vs `Object?`?

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
```diff
@@ -65,26 +79,85 @@ class MemoryController extends DisposableController
...
-  DiffPaneController _createDiffController() =>
-      DiffPaneController(HeapGraphLoaderRuntime(chart.memoryTimeline));
+  Future<void> _init(
+    @visibleForTesting DiffPaneController? connectedDiff,
+    @visibleForTesting ProfilePaneController? connectedProfile,
+  ) async {
+    assert(!_initialized.isCompleted);
+    switch (_mode) {
+      case DevToolsMode.disconnected:
+        // TODO: load memory screen in disconnected mode, https://github.com/flutter/devtools/issues/6972
```
kenzieschmoll, line 89: TODO format (here and elsewhere in this PR) should be `TODO(username OR github issue)`. This will trigger lints in g3 if these are not formatted correctly.

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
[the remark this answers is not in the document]
srawlins: I think the docs are still accurate. Anything produced by `dart:convert` uses `dynamic` (e.g. `List<dynamic>` and `Map<String, dynamic>` or maybe even `Map<dynamic, dynamic>`) instead of `Object?`.

We can also reference the [general recommendation to avoid `dynamic`](https://dart.dev/effective-dart/design#avoid-using-dynamic-unless-you-want-to-disable-static-checking), and I stick strongly to that. I avoid `dynamic` anywhere it isn't necessary. So then then question is: when working with objects produced by `dart:convert`, what should we use? I think `Object?` is safe everywhere here, but I actually haven't worked through the edge cases, etc. I can't promise it, so I think I used `dynamic` in a lot of my devtools PRs, just for consistency.

The runtime danger is that the objects returned from `dart:convert` don't "promote" types of objects when you might expect it. For example, if you call `x = jsonDecode('["hello", "there"]');`, you get a `List<dynamic>` thing at runtime, and you might think "well, I know it's always a List of Strings, so I'll use `x as List<String>`." That is ok at compile-time, but will throw at runtime. You might try `if (x is List<String>)` as defensive, but now that condition _will never be true_ at runtime. It's super awkward but you have to do `if (x is List) x.cast<String>()` (which could still throw if you have bad data) or (`if (x is List) for (var y in x) if (y is String) ...`). Super tedious.

Back to `Object?` vs `dynamic`. Either of these will work, I think all the time, but I'm not sure. E.g.

```dart
import 'dart:convert';

void main() {
  var b = jsonDecode('[1, "hello"]');
  if (b is List<dynamic>) {
    print('yay a List<dynamic>');
  }
  if (b is List<Object?>) {
    print('yay a List<Object?>');
  }
  if (b is List<Object>) {
    print('yay a List');
  }
  f1();
  f2();
  f3();
}

List<dynamic> f1() => jsonDecode('[1, "hello"]');
List<Object?> f2() => jsonDecode('[1, "hello"]');
List<Object> f3() => jsonDecode('[1, "hello"]');
```

This prints "yay a List<dynamic>" and "yay a List<Object?>" because of how is-checks work (assignability?). But not "yay a List<Object>". That test fails. And the calls to `f1` and `f2` pass, but the call to `f3` throws an exception, because the returned object is not a `List<Object>`.

I think guidance w.r.t. dynamic-and-JSON is a team-by-team decision. We can't all become type-system experts and JSON experts (like I can never if JSON explicitly _does not_ support a `null` concept, or does, or if Dart does something special...). So whatever consistent policy keeps the team writing safe code and understandable (for the team) code, is a good policy. I don't think analyzer team has a consistent policy, and we definitely could have JSON-parsing bugs, or YAML-parsing bugs, in malformed analysis_options.yaml, or package_config.json, etc. It's something I keep punting on, hoping `dart:convert` will just change to returning `Object?` objects instead of `dynamic`, or maybe a better JSON-parsing story. 🤷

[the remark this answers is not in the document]
polina-c: Thanks. I suggest to stay compliant with public recommendation.
If we want to invest into switch to <String, Object?>, we need to reflect it in our style rules and migrate the entire project at once, to keep it consistent.

PR #7584, opened 2024-04-15
Rename serialization methods in DevTools to meet recommended standard.

polina-c:
https://docs.flutter.dev/data-and-backend/serialization/json#serializing-json-inside-model-classes

### DIFF
--- a/packages/devtools_app/lib/src/example/conditional_screen.dart
+++ b/packages/devtools_app/lib/src/example/conditional_screen.dart
@@ -91,7 +91,7 @@ class ExampleController extends DisposableController
     } else {
       await maybeLoadOfflineData(
         ExampleConditionalScreen.id,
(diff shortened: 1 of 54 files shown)


kenzieschmoll (commented):
Can we revert all the renames in `devtools_app_shared`, `devtools_extensions`, and `devtools_shared`? These are breaking changes that will cause unnecessary churn for users of these packages.

kenzieschmoll:
Or another option for these packages is to mark the existing methods deprecated with a message to use the new methods. But this may be overkill for a rename.

polina-c:
> Can we revert all the renames in `devtools_app_shared`, `devtools_extensions`, and `devtools_shared`? These are breaking changes that will cause unnecessary churn for users of these packages.

Reverted

kenzieschmoll (approved):

PR #7589, opened 2024-04-16
Skip serialization in case of in-process data passing.

polina-c:


packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
```diff
@@ -141,13 +147,18 @@ class MemoryController extends DisposableController
...
-          profile,
-          chart,
-          profile.classFilter.value,
-          selectedTab: selectedFeatureTabIndex,
-        ).toJson(),
+        data: {
+          // Passing serializable data without conversion to json here
+          // to save time when data are passed in-process,
+          // without serialization.
+          'data': OfflineMemoryData(
```
kenzieschmoll, line 154: how do we make sure we serialize to json when exporting the entire screen for trying to load it later?

What is the performance cost of calling `toJson()` in profile mode?

polina-c: I tested it. Encoding will invoke toJson for all serializable not-json items.

polina-c: Saving to file works:
[dart_devtools_2024-04-16_13:36:51.339.json](https://github.com/flutter/devtools/files/15002263/dart_devtools_2024-04-16_13.36.51.339.json)

kenzieschmoll: interesting. Out of curiosity, would it work if you changed the `toJson` method name to 'foo`? It the "toJson" method name really is "magic", we may be able to clean this up for other screens.

kenzieschmoll, line 154: use `_jsonKey` instead of 'data' string here

polina-c: It will not:

`[FlutterError]: Converting object to an encodable object failed: Instance of 'OfflineMemoryData'`

kenzieschmoll: TIL, that is useful!

kenzieschmoll (approved):

Issue #7444, opened 2024-03-25
Migrate to new flutter bootstrapping


kenzieschmoll:
Warnings we are now seeing during build:
![Screenshot 2024-03-25 at 4 17 49 PM](https://github.com/flutter/devtools/assets/43759233/4dc00303-a1b1-4a42-829f-184909321c5b)

I took a stab at migrating this but it was not as straightforward as I would have hoped.
![Screenshot 2024-03-25 at 4 21 55 PM](https://github.com/flutter/devtools/assets/43759233/c0665b16-6aff-4a9c-8dc1-8a6010c19407)

CC @eyebrowsoffire maybe the warning message can be tweaked to make the migration path more clear?

Bootstrapping changes: https://github.com/flutter/flutter/pull/144434

eyebrowsoffire:
BTW, we are planning for the warnings to link to the documentation once it is finalized. Issue tracking the link in the warning: https://github.com/flutter/flutter/issues/145559

And the issue for documentation for the new bootstrapping in general: https://github.com/flutter/website/issues/10126

I can assist you getting devtools unblocked if you need. The TL;DR is:

* Add a file called `flutter_bootstrap.js` in the `web` directory in your project, next to `index.html`
* Put this in `flutter_bootstrap.js`:
```
{{flutter_js}}
{{flutter_build_config}}
_flutter.loader.load({
  serviceWorkerSettings: {
    serviceWorkerVersion: {{flutter_service_worker_version}},
  },
  config: {
    canvasKitBaseUrl: 'canvaskit/',
  }
});
```
* Get rid of any `flutter.js` script tags in the index.html, and just put the following script tag somewhere:
```
<script src="flutter_bootstrap.js" async></script>
```

If you need help with it, just let me know, I'm happy to help iron out any issues.

PR #7599, opened 2024-04-16
Migrate to new bootstrapping methods.

eyebrowsoffire:
This fixes https://github.com/flutter/devtools/issues/7444

* Moves bootstrapping logic into templated `flutter_bootstrap.js` file.
* Changes to `FlutterLoader.load()` API from `FlutterLoader.loadEntrypoint()` API.

### DIFF
--- a/packages/devtools_app/web/flutter_bootstrap.js
+++ b/packages/devtools_app/web/flutter_bootstrap.js
@@ -1,0 +1,46 @@
+{{flutter_js}}
+{{flutter_build_config}}
+
+// Unregister the old custom DevTools service worker (if it exists). It was
+// removed in: https://github.com/flutter/devtools/pull/5331
+function unregisterDevToolsServiceWorker() {
+  if ('serviceWorker' in navigator) {
+    const DEVTOOLS_SW = 'service_worker.js';
+    const FLUTTER_SW = 'flutter_service_worker.js';
+    navigator.serviceWorker.getRegistrations().then(function(registrations) {
+        for (let registration of registrations) {
+            const activeWorker = registration.active;
+            if (activeWorker != null) {
+                const url = activeWorker.scriptURL;
+                if (url.includes(DEVTOOLS_SW) && !url.includes(FLUTTER_SW)) {
+                    registration.unregister();
+                }
+            }
+        }
+    });
+  }
+}
+
+// Bootstrap app for 3P environments:
+function bootstrapAppFor3P() {
+  _flutter.loader.load({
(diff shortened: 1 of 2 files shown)


packages/devtools_app/web/flutter_bootstrap.js
```diff
@@ -0,0 +1,46 @@
...
+      canvasKitBaseUrl: 'canvaskit/'
+    }
+  });
+}
+
+// Bootstrap app for 1P environments:
+function bootstrapAppFor1P() {
+  _flutter.loader.load({
+    config: {
+      canvasKitBaseUrl: 'canvaskit/'
```
kenzieschmoll, line 40: Looks like this is a change from the previous implementation, where we did not set the `canvasKitBaseUrl` field for 1P. @elliette do you remember why we needed separate bootstrapping for 1P? And does this look okay to you?

eyebrowsoffire: Ah, that's just an oversight by me. I can remove that.

packages/devtools_app/web/flutter_bootstrap.js
```diff
@@ -0,0 +1,42 @@
...
+    },
+    config: {
+      canvasKitBaseUrl: 'canvaskit/'
+    }
+  });
+}
+
+// Bootstrap app for 1P environments:
+function bootstrapAppFor1P() {
+  _flutter.loader.load();
```
elliette, line 38: I'm guessing we also need to specify the entrypointUrl as `main.dart.js`, see https://github.com/flutter/devtools/pull/5705/

packages/devtools_app/web/flutter_bootstrap.js
elliette: I forget why the `canvasKitBasUrl` is only required for 3P, but in general the assets are located in slightly different locations for 1P vs 3P which is why there are two different implementations.

packages/devtools_app/web/flutter_bootstrap.js
eyebrowsoffire: We should not need that, and the new API does not take an `entryPointUrl` as a parameter. The old API should have had that as the default and wouldn't have needed that anyway, so I'm not sure why that was needed before.

elliette:
Note that we will need to update the DevTools copybara script as well, otherwise this will be break DevTools in google3. This is where we replace `bootstrapAppFor3P` with `bootstrapAppFor1P`: cl/513846717

elliette (approved):

PR #7601, opened 2024-04-16
Small tweaks in MemoryController

srawlins:
Just small tweaks I found while poking around this class:

* Make the `_mode` field non-late
* Make the `_init` method `void` so it does not have to be un-awaited.

## Pre-launch Checklist

- [x] I signed the [CLA].

![build.yaml badge]

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
+++ b/packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
@@ -56,22 +56,24 @@ class MemoryController extends DisposableController
   /// instead of the widget state.
   int selectedFeatureTabIndex = 0;
 
-  late final DiffPaneController _diff;
+  late final DiffPaneController diff;
 
-  late final ProfilePaneController _profile;
+  late final ProfilePaneController profile;
 
-  late final MemoryChartPaneController _chart;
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
```diff
@@ -81,7 +78,7 @@ class MemoryController extends DisposableController
 
   static const _jsonKey = 'data';
 
-  Future<void> _init(
+  void _init(
```
bkonyi, line 81: I'm not concerned about this in this particular scenario, but out of curiosity, is this generally considered best practice? I only ask because since this is an `async` method, it's possible that asynchronous gaps can be introduced that impact execution ordering in some contexts that could lead to some nasty race conditions, especially if it's not clear that asynchronous work is being done. Having to wrap unawaited asynchronous functions `unawaited` makes it clear that asynchronous work is being done from the call site.

srawlins: Oh ho! There is a linter rule. https://dart.dev/tools/linter-rules/avoid_void_async

I had no idea. I will drop this :D

Issue #7569, opened 2024-04-11
Support static devtools extensions


kenzieschmoll:
This issue tracks supporting extensions that do not require a running application. Public design doc: https://docs.google.com/document/d/11J4h1oEfB9wBoRYUy4KzclnKITBY-234jKpU3kq4V4w/edit?resourcekey=0-pPo85Svkb9lKdtlrhMCTbw&tab=t.0

- [x] Implement static extensions support in DevTools 
  - https://github.com/flutter/devtools/pull/7570
  - https://github.com/flutter/devtools/pull/7602
  - https://github.com/flutter/devtools/pull/7610
  - https://github.com/flutter/devtools/pull/7611
  - https://github.com/flutter/devtools/pull/7625
  - https://github.com/flutter/devtools/pull/7612
- [x] Add an affordance for available static DevTools extensions to the Flutter sidebar.
  - https://github.com/flutter/devtools/pull/7598
  - https://github.com/flutter/devtools/pull/7604
- [x] Update devtools extensions readme
  - https://github.com/flutter/devtools/pull/7659
  - https://github.com/flutter/devtools/pull/7661

PR #7604, opened 2024-04-17
Include static DevTools extensions in the Flutter sidebar

kenzieschmoll:
Screenshot 2024-04-25 at 9 40 18 AM

Work towards https://github.com/flutter/devtools/issues/7569


packages/devtools_app/lib/src/standalone_ui/vs_code/devtools.dart
```diff
@@ -84,6 +88,94 @@ class DevToolsSidebarOptions extends StatelessWidget {
...
+          'DevTools Extensions',
+          style: theme.textTheme.titleMedium,
+        ),
+        ValueListenableBuilder(
+          valueListenable: _extensionService!.visibleExtensions,
+          builder: (context, extensions, _) {
+            if (extensions.isEmpty) {
+              return const Padding(
+                padding: EdgeInsets.only(left: borderPadding),
+                child: Text('No extensions detected.'),
```
DanTup, line 138: Is it worth just hiding the section if there are none? (it'll probably be empty for most people, at least for a while, so this might just be extra noise)

DanTup (approved):

packages/devtools_app/lib/src/standalone_ui/vs_code/devtools.dart
kenzieschmoll: I was wondering if having a section here might help with discoverability. We could event have a help link that opens the DevTools extensions docs. Eventually, it would be nice if the devtools extensions docs could link to a pub.dev link that queries all packages that provide extensions (https://github.com/dart-lang/pub-dev/issues/7216).

@anderdobo @jacob314 do you have an opinion here on what we should show in the sidebar WRT to extensions? Should we hide the DevTools Extensions section when there are none available, or leave the section with an empty message "No extensions detected" for discoverability of the feature?

packages/devtools_app/lib/src/standalone_ui/vs_code/devtools.dart
```diff
@@ -84,6 +88,94 @@ class DevToolsSidebarOptions extends StatelessWidget {
...
+  ExtensionService? _extensionService;
+
+  @override
+  void initState() {
+    super.initState();
+    _initExtensions();
+  }
+
+  void _initExtensions() {
+    _extensionService = ExtensionService(ignoreServiceConnection: true);
```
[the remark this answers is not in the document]
kenzieschmoll: Yep, this PR depends on https://github.com/flutter/devtools/pull/7612 landing first. 👍

packages/devtools_app/lib/src/standalone_ui/vs_code/devtools.dart
anderdobo: For a cleaner UI, I think we should hide the DevTools Extensions section when there are none available.

PR #7609, opened 2024-04-17
Refactor chart data.

polina-c:
1. Remove live vs offline data from chart memory timeline.
2. Factor out chart data to separate class.

No functional changes.

### DIFF
--- a/packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_data.dart
+++ b/packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_data.dart
@@ -12,9 +12,9 @@ import '../data/primitives.dart';
 class ChartData {
   ChartData({required this.isDeviceAndroid});
 
-  /// Wether device is android, if it [mode] is not [DevToolsMode.connected].
+  /// Wether device is android, if [mode] is not [DevToolsMode.connected].
   ///
-  /// If [mode] is [DevToolsMode.connected], this value  is null
+  /// If [mode] is [DevToolsMode.connected], this value is null
   /// and chart visibility should be detected based on connected app.
   final bool? isDeviceAndroid;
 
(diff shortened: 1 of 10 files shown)


packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_pane_controller.dart
```diff
@@ -34,13 +34,22 @@ class MemoryChartPaneController extends DisposableController
     );
   }
 
+  Map<String, dynamic> toJson() {
```
kenzieschmoll, line 37: nit: class member order: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#order-other-class-members-in-a-way-that-makes-sense

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_data.dart
```diff
@@ -0,0 +1,41 @@
...
+  final _displayInterval =
+      ValueNotifier<ChartInterval>(ChartInterval.theDefault);
+  set displayInterval(ChartInterval interval) {
+    _displayInterval.value = interval;
+  }
+
+  ValueListenable<bool> get isLegendVisible => _legendVisibleNotifier;
+  final _legendVisibleNotifier = ValueNotifier<bool>(true);
+  bool toggleLegendVisibility() =>
+      _legendVisibleNotifier.value = !_legendVisibleNotifier.value;
```
kenzieschmoll, line 35: nit: I don't know if it is important to save this data in the offline export

polina-c: I save everything but state configured in preferences.
It is cheap and create good UX. User will like to see the same thing, without having to re-hide legend.

kenzieschmoll (approved):
a couple nits

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_pane_controller.dart
polina-c: They say 'If no particular order is obvious'.

fromJson and toJson are better readable when close to each other.

PR #7614, opened 2024-04-18
Add serialization and show offline chart.

polina-c:
https://github.com/flutter/devtools/assets/12115586/13e27323-52ad-4f98-9304-9bc7e484745c


packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -38,28 +40,59 @@ class ChartConnection extends DisposableController
...
+
+  late final isDeviceAndroid = _isDevToolsCurrentlyConnected()
+      ? serviceConnection.serviceManager.vm?.operatingSystem == 'android'
+      : false;
+
+  /// True if DevTools is in connected mode and the connection to the app is alive.
+  bool _isDevToolsCurrentlyConnected() =>
+      !offlineDataController.showingOfflineData.value &&
+      serviceConnection.serviceManager.connectedState.value.connected &&
+      serviceConnection.serviceManager.connectedApp != null;
```
kenzieschmoll, line 54: I think you can drop this check. Checking `connectedState.value.connected` should be enough

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -38,28 +40,59 @@ class ChartConnection extends DisposableController
   Timer? _pollingTimer;
   bool _connected = false;
 
-  late final isDeviceAndroid =
-      serviceConnection.serviceManager.vm?.operatingSystem == 'android';
+  /// If true, the connection to application was lost.
+  bool _disconnected = false;
+
+  late final isDeviceAndroid = _isDevToolsCurrentlyConnected()
```
kenzieschmoll, line 46: this can be 
`bool get isDeviceAndroid => serviceConnection.serviceManager.vm?.operatingSystem == 'android';`

This will return false if vm is null, so no need to manually check

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -38,28 +40,59 @@ class ChartConnection extends DisposableController
   Timer? _pollingTimer;
   bool _connected = false;
 
-  late final isDeviceAndroid =
-      serviceConnection.serviceManager.vm?.operatingSystem == 'android';
+  /// If true, the connection to application was lost.
+  bool _disconnected = false;
```
kenzieschmoll, line 44: the fact that we are locally storing `_connected` and `_disconnected` states seems like a bit of code smell that could lead to things getting out of sync. I don't think we should be storing these locally, but instead using the service manager as the source of truth to determine whether DevTools is connected to an app

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -38,28 +40,59 @@ class ChartConnection extends DisposableController
...
+      ? serviceConnection.serviceManager.vm?.operatingSystem == 'android'
+      : false;
+
+  /// True if DevTools is in connected mode and the connection to the app is alive.
+  bool _isDevToolsCurrentlyConnected() =>
+      !offlineDataController.showingOfflineData.value &&
+      serviceConnection.serviceManager.connectedState.value.connected &&
+      serviceConnection.serviceManager.connectedApp != null;
 
   Future<void> maybeConnect() async {
```
kenzieschmoll, line 56: is it possible that maybeConnect will get called more than once? We should be  calling cancelStreamSubscriptions before we add the auto dispose subscriptions in this method

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_data.dart
```diff
@@ -2,40 +2,88 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:collection/collection.dart';
 import 'package:flutter/foundation.dart';
 
-import '../../../../../shared/primitives/simple_items.dart';
+import '../../../../../../devtools_app.dart';
```
kenzieschmoll, line 8: don't import directly

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_data.dart
```diff
@@ -2,40 +2,88 @@
...
+              (mode == DevToolsMode.offlineData &&
+                  isDeviceAndroid != null &&
+                  timeline != null &&
+                  interval != null &&
+                  isLegendVisible != null),
+        ) {
+    this.timeline = timeline ?? MemoryTimeline();
+    _displayInterval =
+        ValueNotifier<ChartInterval>(interval ?? ChartInterval.theDefault);
+    _isLegendVisible = ValueNotifier<bool>(isLegendVisible ?? true);
```
kenzieschmoll, line 38: instead of making these late variables, leave them as final and then just call `.value = ...` here in the constructor body

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_data.dart
```diff
@@ -2,40 +2,88 @@
...
+  }
+
+  factory ChartData.fromJson(Map<String, dynamic> json) {
+    final result = ChartData(
+      mode: DevToolsMode.offlineData,
+      isDeviceAndroid: json[_Json.isDeviceAndroid] as bool? ?? false,
+      timeline:
+          MemoryTimeline.fromJson(json[_Json.timeline] as Map<String, dynamic>),
+      interval: ChartInterval.values
+              .firstWhereOrNull((i) => i.name == json[_Json.interval]) ??
```
kenzieschmoll, line 48: make this a static method on the ChartInterval enum: `ChartInterval.fromName`

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_pane_controller.dart
```diff
@@ -51,15 +40,33 @@ class MemoryChartPaneController extends DisposableController
...
-      await resume();
+  Future<void> get initialized => _initialized.future;
+  final _initialized = Completer<void>();
+
+  Future<void> _init(ChartData? offlineData) async {
+    assert(!_initialized.isCompleted);
+    if (mode == DevToolsMode.connected) {
+      data = ChartData(mode: DevToolsMode.connected);
+    } else {
+      assert(offlineData!.isDeviceAndroid != null);
```
kenzieschmoll, line 51: this is already asserted in the constructor so we can delete this one

packages/devtools_app/lib/src/screens/memory/panes/profile/model.dart
```diff
@@ -10,33 +10,43 @@ import '../../../../shared/table/table_data.dart';
 import '../../../vm_developer/vm_service_private_extensions.dart';
 import '../../shared/heap/class_filter.dart';
 
+class _ProfileJson {
```
kenzieschmoll, line 13: Use an enum instead

packages/devtools_app/lib/src/screens/memory/panes/profile/model.dart
```diff
@@ -84,32 +115,70 @@ class AdaptedProfile {
   final GCStats totalGCStats;
 }
 
+/// Constants ar short here, as they repeat and we want to save space.
```
kenzieschmoll, line 118: r/ar/are

also, if this is a comment, use // instead of ///. /// is for dart doc that defines the class or member

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_pane_controller.dart
```diff
@@ -10,24 +10,38 @@ import 'package:vm_service/vm_service.dart';
 
 import '../../../../shared/config_specific/import_export/import_export.dart';
 import '../../../../shared/globals.dart';
+import '../../../../shared/primitives/simple_items.dart';
 import '../../shared/heap/class_filter.dart';
 import 'model.dart';
 
+class _Json {
+  static const profile = 'profile';
+}
```
kenzieschmoll, line 19: use enum or just a private static variable since there is only one field. same comment elsewhere

packages/devtools_app/lib/src/screens/memory/shared/heap/class_filter.dart
```diff
@@ -57,13 +57,20 @@ class ClassFilter {
   })  : except = _trimByLine(except),
         only = only == null ? null : _trimByLine(only);
 
-  ClassFilter.empty()
+  ClassFilter.theDefault()
```
kenzieschmoll, line 60: rename to default instead of theDefault

packages/devtools_app/lib/src/screens/memory/shared/primitives/memory_timeline.dart
```diff
@@ -8,12 +8,34 @@ import 'package:intl/intl.dart';
 
 import '../../../../shared/primitives/utils.dart';
 
+class _Json {
+  static const data = 'data';
+}
+
 /// All Raw data received from the VM or offline data.
 class MemoryTimeline {
+  MemoryTimeline({List<HeapSample>? data}) {
```
kenzieschmoll, line 17: this can be `MemoryTimeline({this.data = const []})`

packages/devtools_app/lib/src/screens/memory/shared/primitives/memory_timeline.dart
```diff
@@ -8,12 +8,34 @@ import 'package:intl/intl.dart';
...
+    return {
+      _Json.data: data.map((e) => e.toJson()).toList(),
+    };
+  }
+
   int get endingIndex => data.isNotEmpty ? data.length - 1 : -1;
 
   /// Raw Heap sampling data from the VM.
-  final List<HeapSample> data = [];
+  late final List<HeapSample> data;
```
kenzieschmoll, line 38: remove late

packages/devtools_app_shared/lib/src/service/connected_app.dart
```diff
@@ -50,8 +50,10 @@ class ConnectedApp {
   String? _operatingSystem;
 
   // TODO(kenz): investigate if we can use `libraryUriAvailableNow` instead.
-  Future<bool> get isFlutterApp async => _isFlutterApp ??=
-      await serviceManager!.libraryUriAvailable(flutterLibraryUri);
+  Future<bool> get isFlutterApp async =>
+      (_isFlutterApp ??=
+          await serviceManager?.libraryUriAvailable(flutterLibraryUri)) ??
```
kenzieschmoll, line 55: I don't think this change is necessary. `serviceManager` should never be null. If you did get a null exception my guess is that it was a bug due to hot restart

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
```diff
@@ -85,55 +85,55 @@ class MemoryController extends DisposableController
...
           },
           shouldLoad: (data) => true,
-          loadData: (data) => _initializeData(offlineData: data),
+          loadData: (data) async => await _initializeData(offlineData: data),
         );
         // [maybeLoadOfflineData] will be a noop if there is no offline data for the memory screen,
         //  so ensure we still call [_initializedData] if it has not been called.
-        if (!_initialized.isCompleted) _initializeData();
-        assert(_initialized.isCompleted);
+        if (!loaded) {
```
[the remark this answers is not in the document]
polina-c: it seems `loaded` is already local here....

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
polina-c: It is not enough. I do not know what is going on, but these values are not always in sync during loading offline data and/or app disconnect.
Added comment.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
polina-c: 1. If globals[ServiceConnectionManager] is null this will fail.
2. This value does not change during lifetime of the instance, so there is no need to detect it every time, with risk of failure at disconnect

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
polina-c: These values are not about connection between DevTools and application, but about connection between this instance and application. Added comment to make this clear. Thanks.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
polina-c: Yes, but connection is established only once. Replaced booleans with completers to make it clearer. Thanks.

packages/devtools_app/lib/src/screens/memory/shared/heap/class_filter.dart
polina-c: default is a keyword

packages/devtools_app/lib/src/screens/memory/shared/primitives/memory_timeline.dart
polina-c: I cannot use const here because this array will grow.
So, i cannot use initializer, so i cannot remove late below. Added comment.

packages/devtools_app/lib/src/screens/memory/shared/primitives/memory_timeline.dart
polina-c: see above

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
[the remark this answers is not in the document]
kenzieschmoll: The `loaded` local variable was added in this PR on line 100. I'm saying that I think we can remove the `loaded` var (i.e. call `await maybeLoadOfflineData` without saving the result), and then revert this line back to `if (!_initialized.isCompleted) _initializeData();` without checking `if (loaded)`.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
kenzieschmoll: Okay, then we can still make this more verbose but leave as late final:

`late final isDeviceAndroid = serviceConnection.serviceManager.vm?.operatingSystem == 'android';`

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -36,30 +38,67 @@ class ChartConnection extends DisposableController
...
   Timer? _pollingTimer;
-  bool _connected = false;
 
-  late final isDeviceAndroid =
-      serviceConnection.serviceManager.vm?.operatingSystem == 'android';
+  /// If completed, this instance was connected to the application.
+  final Completer<void> _connected = Completer();
+
+  /// If true, the instance lost connection to the application.
+  final Completer<void> _disconnected = Completer();
```
kenzieschmoll, line 46: I really do not think we should be

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -36,30 +38,67 @@ class ChartConnection extends DisposableController
...
+  bool _isDevToolsCurrentlyConnected() =>
+      // Theoretically it should be enough to check only connectedState.value.connected,
+      // but practically these values are not always in sync and, if at least
+      // on of them means disconnection this class consider the connection as lost,
+      // and stops interaction with it.
+      !offlineDataController.showingOfflineData.value &&
+      serviceConnection.serviceManager.connectedState.value.connected &&
+      serviceConnection.serviceManager.connectedApp != null;
 
   Future<void> maybeConnect() async {
```
kenzieschmoll, line 62: can we rename this to `initialize`? The "connect" term is overloaded with DevTools being connected to an app. This is really a method to initialize the chart listeners. So renaming this to `intiailze()` and having a single completer `_initialized` would make this code less confusing.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -36,30 +38,67 @@ class ChartConnection extends DisposableController
...
   Future<void> _onPoll() async {
-    _pollingTimer = null;
-    await _memoryTracker.pollMemory();
-    _pollingTimer = Timer(chartUpdateDelay, _onPoll);
+    assert(_connected.isCompleted);
+    if (_disconnected.isCompleted) return;
+    await _runSafely(() async {
+      _pollingTimer = null;
+      await _memoryTracker.pollMemory();
+      _pollingTimer = Timer(chartUpdateDelay, _onPoll);
```
kenzieschmoll, line 85: two things here:

1. Shouldn't this be `Timer.periodic` so we are actually polling and not just calling this twice?
2. I think the callback is supposed to be `_memoryTracker.pollMemory`, not `_onPoll`. Otherwise this just resets the value of the `_pollingTimer` var. Instead of using recursion, `Timer.periodic` is what we want here to implement polling.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
kenzieschmoll: Okay I've put together a suggestion of what to do with this class. As is, this is starting to feel a bit messy to have to call everything in a `_runSafely` wrapper. Other code in DevTools does not have to do this by making sure that the `connectedState` notifier is being listened to. If you apply these changes, you will not see an error on app disconnect and the polling timer will be handled properly:

https://github.com/flutter/devtools/compare/master...kenzieschmoll:polina-code-review?expand=1

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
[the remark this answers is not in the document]
polina-c: Yes, it will be equivalent. Do you think removing `loaded` will improve readability?

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
polina-c: Yes, what you suggest may pass tests, but I have concerns about reliability.
The code that requires connection is async. That means the connection state can change in an async gap. That means checking connection state before operation will not protect from failures.

And, the actual connection may become unavailable before DevTools detected it and updated connectedState.

Thoughts?

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
polina-c: The timer was not periodic in the original code, and it was making sense as we do not want polls to have risk to overlap.

polina-c: Added comment.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
polina-c: Updated as discussed offline.

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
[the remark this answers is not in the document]
kenzieschmoll: Checking `loaded` seems like we are adding another variable to check initialized state, when really, we want to gate calling `_initializeData` on whether or not `_dataInitialized` is completed. Adding another variable adds the risk that `loaded` and `_dataInitialized` will get out of sync. So yes, I think it is best that we check the completer directly, especially since `_initializeData` will throw an exception if the completer is already completed.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -36,34 +36,86 @@ class ChartConnection extends DisposableController
...
 
-  late final isDeviceAndroid =
-      serviceConnection.serviceManager.vm?.operatingSystem == 'android';
+  /// If completed, this instance was connected to the application.
+  final Completer<void> _initialized = Completer();
+
+  void _stopConnection() {
+    _pollingTimer?.cancel();
+    cancelStreamSubscriptions();
+    cancelListeners();
```
kenzieschmoll, line 46: these are called automatically by `dispose`, so I don't think we should be calling this here.

The reason I suggested to call these prior to adding the listeners / stream subscriptions is so that we do not add duplicate listeners and stream subs if the body of `maybeInitialize` will be called multiple times. You can see this pattern throughout the DevTools codebase when some initialization may occur in `didChangeDependencies`, for example, which can be called more than once.

If the body of `maybeInitialize` where the listeners / stream subs are added will not be executed more than once, then there is no need to call the cancel calls. However, given that they are a no-op when no listeners / stream subs are active, it does not hurt to call them before adding the listeners / stream subs, for robustness.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -36,34 +36,86 @@ class ChartConnection extends DisposableController
...
+      _pollingTimer = Timer(chartUpdateDelay, _startPolling);
+    } catch (e) {
+      if (_checkConnection()) rethrow;
+    }
   }
 
   @override
   void dispose() {
+    // Not nulling out timer, because we need timer to be not null and inactive
+    // for `_isConnectionStopped` to return true.
```
kenzieschmoll, line 118: this is not actually true. `_isConnectionStopped` handles a null `_pollingTimer` object as intended, so we should null it out here.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -36,34 +36,86 @@ class ChartConnection extends DisposableController
...
-  Future<void> _onPoll() async {
-    _pollingTimer = null;
-    await _memoryTracker.pollMemory();
-    _pollingTimer = Timer(chartUpdateDelay, _onPoll);
+  Future<void> _startPolling() async {
+    assert(_initialized.isCompleted);
+    if (!_checkConnection()) return;
+    try {
+      await _memoryTracker.pollMemory();
+      // Timer is not periodic because we do not want polls to overlap.
```
kenzieschmoll, line 108: I think this may leak Timer objects because the Timer never actually finishes its callback, due to the recursion. I think what you want is something like this, a periodic timer with a RateLimiter:

https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/performance/panes/timeline_events/timeline_events_controller.dart/#L160-L171

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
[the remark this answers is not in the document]
bkonyi: +1, it's better to just use `!_initialized.isCompleted` and leave a comment if we're concerned about it not being obvious what the condition is meant to check.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_data.dart
bkonyi: +1, `ValueNotifier`s should always be assigned to avoid introducing `late`, which does insert the equivalent of a null check to each access of the variable.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_pane_controller.dart
```diff
@@ -95,19 +100,34 @@ class MemoryChartPaneController extends DisposableController
...
-  Future<void> resume() async {
-    if (!_paused.value) return;
-    if (mode != DevToolsMode.connected) throw StateError('Not connected.');
-    await _chartConnection!.maybeConnect();
+  void resume() {
+    assert(mode == DevToolsMode.connected && _chartConnection != null);
     _paused.value = false;
   }
 
+  /// Connects when chart is first time expanded.
```
bkonyi, line 108: Ubernit: 

"Establishes the chart connection when the chart is expanded for the first time.

Returns true if the chart was already connected or the connection was established."

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_pane_controller.dart
```diff
@@ -95,19 +100,34 @@ class MemoryChartPaneController extends DisposableController
...
+    data.isDeviceAndroid ??= _chartConnection!.isDeviceAndroid;
+    isAndroidChartVisible.value = data.isDeviceAndroid! &&
+        preferences.memory.androidCollectionEnabled.value;
+  }
 
-  void _updateAndroidChartVisibility() {
-    final isAndroid = data.isDeviceAndroid ?? _chartConnection!.isDeviceAndroid;
-    isAndroidChartVisible.value =
-        isAndroid && preferences.memory.androidCollectionEnabled.value;
+  ValueListenable<bool> isChartVisible = preferences.memory.showChart;
```
bkonyi, line 126: Should this be grouped with the other variables? Also, could this be made into a getter instead?

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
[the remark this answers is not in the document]
polina-c: How about different perspective: we assume that _initializeData will not be completed not completed if data are not loaded by maybeLoadOfflineData. 
By adding variable `loaded` we verify this assumption. Without it we leave risk that data are actually loaded but _dataInitialized is not completed because of an error. Or another one: data are not loaded but _dataInitialized is completed.

In short: yes we have this risk to get out of sync, but with `loaded` we will catch it, and without it we will miss it.

Does it change something?

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
polina-c: I am cancelling them, because I want them to stop triggering functions, not because I want to dispose them.
They will not trigger events in case of disconnect. How about switching to offline mode?
Connection is still here and may trigger events.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
polina-c: maybeInitialize has async gap when _initialized is already completed, but _polling is still null. 
I updated _isConnectionStopped

Observation: if we had a flag `isDisconnected` we would not spend this much time discussing this question.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_data.dart
```diff
@@ -2,40 +2,88 @@
...
+    bool? isLegendVisible,
+  }) : assert(
+          mode == DevToolsMode.connected ||
+              (mode == DevToolsMode.offlineData &&
+                  isDeviceAndroid != null &&
+                  timeline != null &&
+                  interval != null &&
+                  isLegendVisible != null),
+        ) {
+    this.timeline = timeline ?? MemoryTimeline();
```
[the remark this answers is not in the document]
polina-c: Yes, it needs to be a constant.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -36,34 +36,86 @@ class ChartConnection extends DisposableController
   );
 
   Timer? _pollingTimer;
-  bool _connected = false;
 
-  late final isDeviceAndroid =
-      serviceConnection.serviceManager.vm?.operatingSystem == 'android';
+  /// If completed, this instance was connected to the application.
+  final Completer<void> _initialized = Completer();
```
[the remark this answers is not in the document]
polina-c: I realized introducing enum significantly simplifies the code.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -36,34 +36,86 @@ class ChartConnection extends DisposableController
...
+
+  /// True if DevTools is in connected mode.
+  ///
+  /// If DevTools is in offline mode, stops connection and returns false.
+  bool _checkConnection() {
+    assert(_initialized.isCompleted);
+    if (_isConnectionStopped) return false;
+
+    // If connection is up and running, return true.
+    if (!offlineDataController.showingOfflineData.value &&
```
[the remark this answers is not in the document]
polina-c: I want to stop listeners right away because I do not want to take events. Disconnect consists of many parts and it is not responsibility of this class to know or guarantee that events will not happen in case of disconnect. So, it just should stop listening.

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_pane_controller.dart
```diff
@@ -45,7 +57,18 @@ class ProfilePaneController extends DisposableController
 
   /// Current class filter.
   ValueListenable<ClassFilter> get classFilter => _classFilter;
-  final _classFilter = ValueNotifier(ClassFilter.empty());
+  final _classFilter = ValueNotifier(ClassFilter.theDefault());
+  void setFilter(ClassFilter filter) {
```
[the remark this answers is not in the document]
polina-c: See #7 here: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#order-other-class-members-in-a-way-that-makes-sense

packages/devtools_app/lib/src/screens/memory/panes/profile/model.dart
polina-c: If I use enum, I will have to add `.name` each time i need string value that will crowd serialization code.

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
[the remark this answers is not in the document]
polina-c: Added assert as discussed with Kenzie.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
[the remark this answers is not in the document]
polina-c: After discussion with Kenzie I realized my understanding of cycle is wrong. With updated understanding, yes, `dispose` is enough.

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
```diff
@@ -85,55 +85,56 @@ class MemoryController extends DisposableController
...
           shouldLoad: (data) => true,
-          loadData: (data) => _initializeData(offlineData: data),
+          loadData: (data) async => await _initializeData(offlineData: data),
         );
         // [maybeLoadOfflineData] will be a noop if there is no offline data for the memory screen,
         //  so ensure we still call [_initializedData] if it has not been called.
-        if (!_initialized.isCompleted) _initializeData();
-        assert(_initialized.isCompleted);
+        assert(loaded == _dataInitialized.isCompleted);
+        if (!loaded) {
```
kenzieschmoll, line 113: Now that we have the assert, we want to be checking `_dataInitialized.isCompleted` here instead.

packages/devtools_app/lib/src/screens/memory/framework/memory_controller.dart
```diff
@@ -109,31 +111,32 @@ class MemoryController extends DisposableController
         );
         // [maybeLoadOfflineData] will be a noop if there is no offline data for the memory screen,
         //  so ensure we still call [_initializedData] if it has not been called.
-        if (!_initialized.isCompleted) _initializeData();
-        assert(_initialized.isCompleted);
+        assert(loaded == _dataInitialized.isCompleted);
+        if (_dataInitialized.isCompleted) {
```
kenzieschmoll, line 115: this should be negated

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -35,36 +34,64 @@ class ChartConnection extends DisposableController
...
+  late final bool isDeviceAndroid;
+
+  void maybeInit() async {
+    if (initialized) return;
+    initialized = true;
+
+    if (!serviceConnection.serviceManager.connectedState.value.connected) {
+      isDeviceAndroid = false;
+      return;
+    }
```
kenzieschmoll, line 50: we shouldn't need this. ChartVmConnection.maybeInit should never be called if we are not connected to an app.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -35,36 +34,64 @@ class ChartConnection extends DisposableController
...
-  bool _connected = false;
+  bool initialized = false;
+
+  DebounceTimer? _polling;
+
+  late final bool isDeviceAndroid;
+
+  void maybeInit() async {
+    if (initialized) return;
+    initialized = true;
```
kenzieschmoll, line 45: put this at the end of the method so we are setting it to true after all the initialization has occurred.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_pane_controller.dart
```diff
@@ -17,51 +15,45 @@ import 'charts/vm_chart_controller.dart';
...
+      _chartConnection = ChartVmConnection(
+        this.data.timeline,
+        isAndroidChartVisible: isAndroidChartVisible,
+      );
+    } else {
+      this.data = data!;
+      _chartConnection = null;
+      _paused.value = false;
+      recomputeChartData();
+      _paused.value = true;
```
kenzieschmoll, line 37: since `recomputeChartData` is synchronous, we do not need to set _paused.value to false and true around this call. The UI thread will never be released to update listeners, and the default value for `_paused.value` is already true.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_pane_controller.dart
```diff
@@ -95,19 +85,28 @@ class MemoryChartPaneController extends DisposableController
   ValueListenable<bool> get paused => _paused;
   final _paused = ValueNotifier<bool>(true);
   void pause() => _paused.value = true;
-  Future<void> resume() async {
-    if (!_paused.value) return;
-    if (mode != DevToolsMode.connected) throw StateError('Not connected.');
-    await _chartConnection!.maybeConnect();
+  void resume() {
```
kenzieschmoll, line 88: nit: use => since this will fit on one line now

packages/devtools_app/test/test_infra/scenes/memory/default.dart
```diff
@@ -154,9 +154,12 @@ class MemoryDefaultScene extends Scene {
...
       connectedProfile: profileController,
-    )
-      ..chart.data.timeline.data.clear()
-      ..chart.data.timeline.data.addAll(memoryJson.data);
+    );
+
+    await controller.initialized;
+
+    controller.chart.data.timeline.data.clear();
+    controller.chart.data.timeline.data.addAll(memoryJson.data);
```
kenzieschmoll, line 162: use cascades: `..clear()..addAll(`

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
polina-c: Then we will have to remember to do it in every early return. Moved.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_pane_controller.dart
polina-c: Added comment.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
polina-c: We discussed it yesterday. Added comment.

kenzieschmoll: We discussed returning early if not connected. Looking back through the commits, what we had was this:
```
if (initialized || !serviceConnection.serviceManager.connectedState.value.connected) {
  return;
}
```

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -42,10 +42,12 @@ class ChartVmConnection extends DisposableController
 
   void maybeInit() async {
     if (initialized) return;
-    initialized = true;
 
+    // We do this check for cases of accidental disconnect, in order to avoid
+    // failure for initialization of `isDeviceAndroid` next line.
```
kenzieschmoll, line 47: we don't need to worry about this. `maybeInit` is called synchronously, and we only call it after we have checked that we in fact are connected.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
polina-c: Yes, and (1) I got failure for not initialized late isDeviceAndroid and (2) i believe the state of the instance here should become `initialized and disconnected`, not `not initialized`.

How does it sound?

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
polina-c: Ok, added comment and assert.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -35,36 +34,66 @@ class ChartConnection extends DisposableController
...
+  bool initialized = false;
+
+  DebounceTimer? _polling;
+
+  late final bool isDeviceAndroid;
+
+  /// Initializes the connection.
+  ///
+  /// This method should be called without async gap after validation
+  /// the application is not disconnected.
```
kenzieschmoll, line 46: nit: avoid double negatives "validation that an application is connected."

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
polina-c: Yes, but i got failure because of not initialized late isDeviceAndroid.
Replaced with assert.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -35,36 +34,66 @@ class ChartConnection extends DisposableController
...
+  void maybeInit() async {
+    if (initialized) return;
+
+    assert(!serviceConnection.serviceManager.connectedState.value.connected);
```
kenzieschmoll, line 50: we should be asserting that we are connected, not that we are disconnected.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
kenzieschmoll: if we are getting a late initialized error on `isDeviceAndroid`, then that seems like a bug we should be tracking down. We should not be checking any values on ChartVMConnection before it has been initialized.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_connection.dart
```diff
@@ -35,36 +34,66 @@ class ChartConnection extends DisposableController
...
+  bool initialized = false;
+
+  DebounceTimer? _polling;
+
+  late final bool isDeviceAndroid;
+
+  /// Initializes the connection.
+  ///
+  /// This method should be called without async gap after validation that
+  /// the application still connected.
```
kenzieschmoll, line 46: nit: is still connected

kenzieschmoll (approved):

PR #7626, opened 2024-04-19
Fix record experience so we only get requests that occur while recording.

CoderDake:
If pausing and resuming recording, DT would pick up everying since recording was paused. This PR makes it so that only requests that occur during recording are polled for.

Fixes https://github.com/flutter/devtools/issues/7618

https://github.com/flutter/devtools/assets/1386322/9fb0e1c3-5807-49d7-a515-2ba16438faaf


packages/devtools_app/lib/src/screens/network/network_controller.dart
```diff
@@ -244,6 +244,15 @@ class NetworkController extends DisposableController
     _recordingNotifier.value = state;
   }
 
+  /// Updtates the last refresh time of the socket and http data refresh times.
```
bkonyi, line 247: Nit: Updates

bkonyi (approved):

kenzieschmoll (approved):
lgtm with one comment

kenzieschmoll (approved):

PR #7677, opened 2024-04-29
Fix disconnect process.

polina-c:


kenzieschmoll:
This can be closed. I already landed this fix here: https://github.com/flutter/devtools/pull/7679

PR #7681, opened 2024-04-30
Factor out PeriodicTimerWithOverlapProtection.

polina-c:


packages/devtools_app/lib/src/shared/primitives/utils.dart
```diff
@@ -359,6 +359,32 @@ extension JsonMap on Map<String, Object?> {
   String prettyPrint() => const JsonEncoder.withIndent('  ').convert(this);
 }
 
+const _microsecondsInSecond = 1000000;
+
+class PeriodicTimerWithOverlapProtection {
```
kenzieschmoll, line 364: nit: how about rename to `Poller`? Also, add dart doc

packages/devtools_app/lib/src/shared/primitives/utils.dart
```diff
@@ -359,6 +359,32 @@ extension JsonMap on Map<String, Object?> {
   String prettyPrint() => const JsonEncoder.withIndent('  ').convert(this);
 }
 
+const _microsecondsInSecond = 1000000;
```
kenzieschmoll, line 362: make this const part of the class below

packages/devtools_app/lib/src/shared/primitives/utils.dart
```diff
@@ -359,6 +359,32 @@ extension JsonMap on Map<String, Object?> {
...
+
+class PeriodicTimerWithOverlapProtection {
+  PeriodicTimerWithOverlapProtection(
+    this.delay,
+    this.callback,
+    double? ratePerSecond,
```
kenzieschmoll, line 368: nit: how about rename to `maxPollsPerSecond`?

PR #7699, opened 2024-05-05
Serialization for diff and profile tabs.

polina-c:


packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -26,19 +26,62 @@ import '../data/csv.dart';
...
+      core._snapshots.value.addAll(snapshots);
+    }
+    derived._updateValues();
+  }
 
   factory DiffPaneController.fromJson(Map<String, dynamic> json) {
-    // TODO(polina-c): implement, https://github.com/flutter/devtools/issues/6972
-    return DiffPaneController(loader: null);
+    final snapshots = (json[_Json.snapshots] as List)
+        .map((e) => SnapshotDataItem.fromJson(e))
```
kenzieschmoll, line 49: I think this is using dynamic invocation in the `map` call. `e` is dynamic here. To avoid this, you can cast before calling .map

`(json[_Json.snaphots] as List).cast<Map<String, dynamic>>().map(...).toList()`

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -26,19 +26,62 @@ import '../data/csv.dart';
...
+        .toList();
+
+    final diffWith = json[_Json.diffWith] as List<int?>;
+
+    for (var i = 0; i < snapshots.length; i++) {
```
kenzieschmoll, line 54: i can be final

kenzieschmoll, line 54: before this loop can we assert that `snapshots.length` == `diffWith.length`? otherwise we may hit index out of range errors in the loop

packages/devtools_app/lib/src/screens/memory/panes/diff/widgets/snapshot_control_pane.dart
```diff
@@ -76,7 +76,7 @@ class _DiffDropdown extends StatelessWidget {
 
   List<DropdownMenuItem<SnapshotDataItem>> items() =>
       controller.core.snapshots.value
-          .where((item) => item.hasData)
+          .where((item) => item is SnapshotDataItem && item.isProcessed)
```
kenzieschmoll, line 79: you can use `whereType` here so that `item` will already be typed properly

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -26,19 +26,62 @@ import '../data/csv.dart';
...
-    return {};
+    final snapshots = core.snapshots.value
+        .whereType<SnapshotDataItem>()
+        .where((s) => s.heap != null)
+        .toList();
+
+    final snapshotToIndex =
+        snapshots.asMap().map((index, item) => MapEntry(item, index));
+
+    final diffsWith = snapshots.map((item) {
```
kenzieschmoll, line 76: nit rename to `diffWithSnapshots`

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/snapshot_item.dart
```diff
@@ -11,34 +13,56 @@ import '../../../../../shared/memory/heap_graph_loader.dart';
...
+      nameOverride: json[_Json.nameOverride] as String?,
+    );
+
+    final loader = HeapGraphLoaderFromChunks(
+      chunks: json[_Json.chunks] as List<ByteData>,
+      created: json[_Json.created] as DateTime,
+    );
+
+    unawaited(
+      result.loadHeap(loader),
```
kenzieschmoll, line 51: formatting: remove trailing comma and then put comment above this line

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/snapshot_item.dart
```diff
@@ -11,34 +13,56 @@ import '../../../../../shared/memory/heap_graph_loader.dart';
...
+    ); // Start the loading process, that will result in progress indicator in UI.
+
+    return result;
+  }
+
+  Map<String, dynamic> toJson() {
+    final heap = _heap!; // Not processed heaps are not serializable.
```
kenzieschmoll, line 58: we do not want to throw exceptions at runtime. Should return an empty map instead?

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
```diff
@@ -455,12 +455,6 @@ class AllocationProfileTableView extends StatefulWidget {
 
 class AllocationProfileTableViewState
     extends State<AllocationProfileTableView> {
-  @override
-  void didChangeDependencies() {
-    super.didChangeDependencies();
-    widget.controller.initialize();
```
kenzieschmoll, line 461: this is changing the lifecycle of when we pay the cost to initialize the `ProfilePaneController`. The way this code used to be delayed controller initialization until we actually needed it when the Allocation profile view was opened. Are we certain this is what we want to do?

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -26,19 +26,62 @@ import '../data/csv.dart';
...
+      final diffWith = item.diffWith.value;
+      return diffWith == null ? null : snapshotToIndex[diffWith];
+    }).toList();
+
+    return {
+      _Json.snapshots: snapshots,
+      _Json.diffWith: diffsWith,
```
kenzieschmoll, line 83: is this as space efficient as it can be? Or are we storing snapshots twice if there is a diff with value?

packages/devtools_app/lib/src/shared/memory/heap_graph_loader.dart
```diff
@@ -44,3 +46,16 @@ class HeapGraphLoaderFile implements HeapGraphLoader {
     );
   }
 }
+
+class HeapGraphLoaderFromChunks implements HeapGraphLoader {
```
kenzieschmoll, line 50: dartdoc

packages/devtools_app/lib/src/shared/memory/heap_data.dart
```diff
@@ -12,25 +14,28 @@ import 'retainers.dart';
...
-  HeapData._(
-    this.graph,
-    this.classes,
-    this.footprint, {
+  HeapData(
+    this.graph, {
     required this.created,
-    required this.retainedSizes,
-  });
+    @visibleForTesting bool startCalculation = true,
```
kenzieschmoll, line 21: this is unused, so you can remove

kenzieschmoll (commented):
Please add tests for the new serialization and any other affected code as part of this change.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
polina-c: Screenshot 2024-05-07 at 2 48 53 PM

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
polina-c: It is job of memory controller, not widget, to initiate initialization of ProfilePaneController.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
polina-c: diffsWith is just array of indexes

packages/devtools_app/lib/src/shared/memory/heap_data.dart
polina-c: Added comment.

packages/devtools_app/lib/src/screens/memory/panes/profile/profile_view.dart
kenzieschmoll: Performing the initialization in the widget lifecycle makes this lazy, which we probably still want. Otherwise we will pay the cost of listeners firing in the background for a feature that may not even be in use. This is similar to how we don't initialize the memory controller if we are on the flutter inspector, for example. We initialize on first use.

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
kenzieschmoll: Oh whoops. I was thinking this was `for (final foo in someIterable)`.

packages/devtools_app/lib/src/screens/memory/shared/primitives/memory_timeline.dart
```diff
@@ -123,7 +123,7 @@ class MemoryTimeline {
     _eventSamples.add(
       EventSample.gcEvent(
         timestamp,
-        events: extensionEvents,
+        theEvents: extensionEvents,
```
kenzieschmoll, line 126: rename to events

packages/devtools_shared/lib/src/memory/event_sample.dart
```diff
@@ -163,12 +163,12 @@ class EventSample {
     this.extensionEvents,
   );
 
-  EventSample.gcEvent(this.timestamp, {ExtensionEvents? events})
+  EventSample.gcEvent(this.timestamp, {ExtensionEvents? theEvents})
```
kenzieschmoll, line 166: rename to events. The in scope var should be used.

packages/devtools_app/lib/src/screens/memory/shared/primitives/memory_timeline.dart
polina-c: Then it will conflict with the field name

packages/devtools_shared/lib/src/memory/event_sample.dart
polina-c: Then it will conflict with the field name

polina-c:
> Please add tests for the new serialization and any other affected code as part of this change.

Added tests for serialization.
Most of other code is test covered.

I will add more coverage after completing the feature and before  removing the flag.

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
```diff
@@ -25,31 +26,29 @@ class OfflineMemoryData {
...
-        json[key] as Map<String, dynamic>? ?? {};
     return OfflineMemoryData(
-      DiffPaneController.fromJson(item(_Json.diffData)),
-      ProfilePaneController.fromJson(item(_Json.profileData)),
-      ChartData.fromJson(item(_Json.chartData)),
-      ClassFilter.fromJson(item(_Json.classFilter)),
+      deserialize(json[_Json.diffData], DiffPaneController.fromJson),
+      deserialize(json[_Json.profileData], ProfilePaneController.fromJson),
+      deserialize(json[_Json.chartData], ChartData.fromJson),
+      deserialize(json[_Json.classFilter], ClassFilter.fromJson),
```
kenzieschmoll, line 33: add the generic type argument to `deserialize`

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_data.dart
```diff
@@ -38,11 +38,15 @@ class ChartData {
   }
 
   factory ChartData.fromJson(Map<String, dynamic> json) {
+    final timeline = json[_Json.timeline];
     final result = ChartData(
       mode: ControllerCreationMode.offlineData,
       isDeviceAndroid: json[_Json.isDeviceAndroid] as bool? ?? false,
-      timeline:
-          MemoryTimeline.fromJson(json[_Json.timeline] as Map<String, dynamic>),
+      timeline: timeline is MemoryTimeline
```
kenzieschmoll, line 45: can you use the deserialize helper here instead?

packages/devtools_app/lib/src/shared/primitives/serialization.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+typedef FromJson<T> = T Function(Map<String, dynamic> json);
+
+/// Deserialized an object if it is serialized.
+// ignore: avoid-dynamic, serialization is exception for the rule.
+T deserialize<T>(dynamic json, FromJson<T> deserializer) {
```
kenzieschmoll, line 9: add tests for this helper.

packages/devtools_app/lib/src/shared/primitives/serialization.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: 2024

packages/devtools_app/test/memory/chart/controller/chart_data_test.dart
```diff
@@ -0,0 +1,32 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: 2024

packages/devtools_app/test/memory/chart/controller/chart_data_test.dart
```diff
@@ -0,0 +1,32 @@
...
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  test(
+    '$ChartData serializes and deserializes correctly',
+    () {
+      final item = ChartData(
+        mode: ControllerCreationMode.offlineData,
+        isDeviceAndroid: true,
+        timeline: MemoryTimeline(),
```
kenzieschmoll, line 18: can you add a test with non-empty timeline data?

packages/devtools_app/test/memory/chart/controller/chart_data_test.dart
```diff
@@ -0,0 +1,32 @@
...
+        interval: ChartInterval.theDefault,
+        isLegendVisible: true,
+      );
+
+      final json = item.toJson();
```
kenzieschmoll, line 23: assert the value of this:

```
expect(item.toJson(), {...});
```

packages/devtools_app/test/memory/diff/controller/diff_pane_controller_test.dart
```diff
@@ -0,0 +1,68 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
kenzieschmoll, line 3: remove duplicate copyright

packages/devtools_app/test/memory/diff/controller/snapshot_item_test.dart
```diff
@@ -0,0 +1,36 @@
...
+    '$SnapshotDataItem serializes and deserializes correctly',
+    () async {
+      final item = SnapshotDataItem(
+        defaultName: 'defaultName',
+        displayNumber: 1,
+        nameOverride: 'nameOverride',
+      );
+      await item.loadHeap(HeapGraphLoaderGoldens());
+
+      final json = item.toJson();
```
kenzieschmoll, line 21: assert the values created by toJson to verify expectations

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -19,26 +19,72 @@ import '../../../../../shared/memory/classes.dart';
...
+      core._snapshots.addAll(snapshots);
+    }
+    derived._updateValues();
+  }
 
   factory DiffPaneController.fromJson(Map<String, dynamic> json) {
-    // TODO(polina-c): implement, https://github.com/flutter/devtools/issues/6972
-    return DiffPaneController(loader: null);
+    final snapshots = (json[_Json.snapshots] as List)
+        .map((e) => deserialize(e, SnapshotDataItem.fromJson))
```
kenzieschmoll, line 50: add type to deserialize

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -19,26 +19,72 @@ import '../../../../../shared/memory/classes.dart';
...
-    return {};
+    final snapshots = core.snapshots.value
+        .whereType<SnapshotDataItem>()
+        .where((s) => s.heap != null)
+        .toList();
+
+    final snapshotToIndex =
+        snapshots.asMap().map((index, item) => MapEntry(item, index));
+
+    final diffsWithSnapshots = snapshots.map((item) {
```
kenzieschmoll, line 79: nit: now that I better understand what this var is, how about "diffWithIndices" and then below r/diffWith/diffWithIndex

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -19,26 +19,72 @@ import '../../../../../shared/memory/classes.dart';
...
+      final diffWith = item.diffWith.value;
+      return diffWith == null ? null : snapshotToIndex[diffWith];
+    }).toList();
+
+    return {
+      _Json.snapshots: snapshots,
+      _Json.diffWith: diffsWithSnapshots,
```
kenzieschmoll, line 86: same here: _Json.diffWithIndices

packages/devtools_app/lib/src/shared/memory/heap_data.dart
```diff
@@ -12,25 +14,27 @@ import 'retainers.dart';
...
 /// Raw and calculated data of the heap snapshot.
-@immutable
 class HeapData {
-  HeapData._(
-    this.graph,
-    this.classes,
-    this.footprint, {
+  HeapData(
+    this.graph, {
     required this.created,
```
kenzieschmoll, line 20: nit: this will all fit on one line so no need for trailing comma

packages/devtools_app/lib/src/shared/memory/heap_graph_loader.dart
```diff
@@ -44,3 +46,17 @@ class HeapGraphLoaderFile implements HeapGraphLoader {
...
   }
 }
+
+/// Loads a heap snapshot from `List<ByteData>` created with HeapSnapshotGraph.toChunks.
+class HeapGraphLoaderFromChunks implements HeapGraphLoader {
+  HeapGraphLoaderFromChunks({required this.chunks, required this.created});
+
+  List<ByteData> chunks;
+
+  DateTime created;
```
kenzieschmoll, line 56: should these be final?

packages/devtools_app/lib/src/shared/memory/heap_graph_loader.dart
```diff
@@ -44,3 +46,17 @@ class HeapGraphLoaderFile implements HeapGraphLoader {
...
+
+  @override
+  Future<(HeapSnapshotGraph, DateTime)> load() async {
+    return (HeapSnapshotGraph.fromChunks(chunks), created);
+  }
```
kenzieschmoll, line 61: this doesn't need to be async. If the super class returns FutureOr, then we can remove the unnecessary async here

packages/devtools_app/test/memory/diff/controller/diff_pane_controller_test.dart
```diff
@@ -0,0 +1,68 @@
...
+
+      final controller = scene.controller.diff;
+
+      final snapshots =
+          controller.core.snapshots.value.whereType<SnapshotDataItem>();
+
+      expect(snapshots.length, 2);
+      snapshots.first.diffWith.value = snapshots.last;
+
+      final fromJson = DiffPaneController.fromJson(controller.toJson());
```
kenzieschmoll, line 56: can you assert the content of .toJson() and .fromJson() so we can verify the values of the serialization are as expected?

packages/devtools_app/test/memory/framework/offline_data_test.dart
```diff
@@ -0,0 +1,52 @@
...
+          isDeviceAndroid: true,
+          timeline: MemoryTimeline(),
+          interval: ChartInterval.theDefault,
+          isLegendVisible: true,
+        ),
+        ClassFilter.empty(),
+        selectedTab: 0,
+      );
+
+      final json = item.toJson();
```
kenzieschmoll, line 33: assert values of toJson result

packages/devtools_shared/lib/src/memory/event_sample.dart
kenzieschmoll: you can use `this.events = events`

packages/devtools_shared/lib/src/memory/heap_sample.dart
```diff
@@ -53,9 +53,9 @@ class HeapSample {
         'used': used,
         'external': external,
         'gc': isGC,
-        'adb_memoryInfo': adbMemoryInfo.toJson(),
-        'memory_eventInfo': memoryEventInfo.toJson(),
-        'raster_cache': rasterCache.toJson(),
+        'adb_memoryInfo': adbMemoryInfo,
+        'memory_eventInfo': memoryEventInfo,
+        'raster_cache': rasterCache,
```
kenzieschmoll, line 58: I recommend that we don't make unnecessary changes to devtools_shared that aren't test covered. I don't think this code has good test coverage to verify nothing is broken here.

packages/devtools_app/test/memory/diff/controller/snapshot_item_test.dart
polina-c: IMHO, it will increase maintenance cost of the test, without adding value.
We do not care what is in json, we just want it to produce the same values when deserialized, right?

packages/devtools_app/test/memory/diff/controller/diff_pane_controller_test.dart
polina-c: about `toJson` see https://github.com/flutter/devtools/pull/7699#discussion_r1596044223

added some validation of `from`

packages/devtools_app/test/memory/framework/offline_data_test.dart
polina-c: see https://github.com/flutter/devtools/pull/7699#discussion_r1596044223

packages/devtools_shared/lib/src/memory/heap_sample.dart
polina-c: Added test coverage. Heap sample is part of chart data, stat needs to be serialized.

packages/devtools_app/test/memory/diff/controller/snapshot_item_test.dart
kenzieschmoll: If the `.toJson()` value is not what we expect (like a field is missing for example or a default value is wrong), bugs could still slip in. For example both `.toJson()` and `fromJson()` could be missing a field.

packages/devtools_shared/lib/devtools_shared.dart
```diff
@@ -16,4 +16,5 @@ export 'src/utils/compare.dart';
 export 'src/utils/file_utils.dart';
 export 'src/utils/retry.dart';
 export 'src/utils/semantic_version.dart';
+export 'src/utils/serialization.dart';
```
kenzieschmoll, line 19: add a changelog entry for changes to public package devtools_shared

packages/devtools_app/test/memory/diff/controller/snapshot_item_test.dart
polina-c: I added validation that all keys are used. I skipped validation for values. 
Yes, bugs still can squeeze, but no test coverage is full. I believe verification for values is overhead in this case.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_data.dart
```diff
@@ -9,11 +9,15 @@ import '../../../../../shared/primitives/simple_items.dart';
...
-class _Json {
+// ignore: avoid_classes_with_only_static_members, enum-like classes are ok
+@visibleForTesting
+class Json {
   static const isDeviceAndroid = 'isAndroid';
   static const timeline = 'timeline';
   static const interval = 'interval';
   static const isLegendVisible = 'isLegendVisible';
+
+  static final all = [isDeviceAndroid, timeline, interval, isLegendVisible];
```
kenzieschmoll, line 20: make this a set for consistency above. Or preferably, just switch to using an enum so that you can use `Json.values`.

If you leave the `all` field, mark visible for testing if that is the only place it is used. Here and elsewhere.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_data.dart
```diff
@@ -11,13 +11,15 @@ import '../data/primitives.dart';
...
-class Json {
-  static const isDeviceAndroid = 'isAndroid';
-  static const timeline = 'timeline';
-  static const interval = 'interval';
-  static const isLegendVisible = 'isLegendVisible';
+enum Json {
+  isDeviceAndroid('isAndroid'),
+  timeline('timeline'),
+  interval('interval'),
+  isLegendVisible('isLegendVisible');
```
kenzieschmoll, line 18: instead of passing the string representation to each of these, consider this pattern where the constructor has an optional name override, and then you don't have to pass in the string for the values where `Json.enumValue.name` is the equivalent. 

example: https://github.com/flutter/devtools/blob/master/packages/devtools_app/integration_test/test_infra/run/_test_app_driver.dart/#L417

packages/devtools_app/lib/src/screens/memory/panes/diff/controller/diff_pane_controller.dart
```diff
@@ -31,11 +31,13 @@ import 'snapshot_item.dart';
...
-  static const snapshots = 'snapshots';
-  static const diffWith = 'diffWith';
+enum Json {
+  snapshots('snapshots'),
+  diffWith('diffWith');
 
-  static final all = [snapshots, diffWith];
+  const Json(this.key);
+
+  final String key;
```
kenzieschmoll, line 40: unnecessary. Json.snapshots.name and Json.diffWith.name are equivalent. This comment applies here and elsewhere to other enums.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_data.dart
```diff
@@ -9,17 +9,18 @@ import '../../../../../shared/primitives/simple_items.dart';
...
+  interval,
+  isLegendVisible;
 
-  const Json(this.key);
+  const Json({String? nameOverride}) : _nameOverride = nameOverride;
 
-  final String key;
+  final String? _nameOverride;
+
+  String get key => _nameOverride ?? name;
```
kenzieschmoll, line 23: since the override is not used, we don't need any of this.

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_data.dart
```diff
@@ -16,11 +16,7 @@ enum Json {
   interval,
   isLegendVisible;
 
-  const Json({String? nameOverride}) : _nameOverride = nameOverride;
-
-  final String? _nameOverride;
-
-  String get key => _nameOverride ?? name;
+  String get key => name;
```
kenzieschmoll, line 19: why not just use `.name` directly? Seems like added complexity to add a getter that just returns the unchanged name value.

packages/devtools_app/lib/src/screens/memory/framework/offline_data/offline_data.dart
```diff
@@ -2,17 +2,29 @@
...
+import 'package:meta/meta.dart';
+
 import '../../panes/chart/controller/chart_data.dart';
 import '../../panes/diff/controller/diff_pane_controller.dart';
 import '../../panes/profile/profile_pane_controller.dart';
 import '../../shared/heap/class_filter.dart';
 
-class _Json {
+@visibleForTesting
+class Json {
```
kenzieschmoll, line 14: make this an enum too

packages/devtools_shared/lib/src/memory/heap_sample.dart
```diff
@@ -4,9 +4,37 @@
...
 import 'dart:convert';
 
+import 'package:meta/meta.dart';
+
+import '../utils/serialization.dart';
 import 'adb_memory_info.dart';
 import 'event_sample.dart';
 
+@visibleForTesting
+class Json {
```
kenzieschmoll, line 14: make this an enum

packages/devtools_app/lib/src/screens/memory/panes/chart/controller/chart_data.dart
polina-c: For me it is alert for future maintainer that key may be different from name, and it is needed for backward compatibility.
But, yes, it is debatable. 
Removed.

kenzieschmoll (approved):

packages/devtools_shared/CHANGELOG.md
```diff
@@ -1,10 +1,14 @@
 # 10.0.0-dev.2
 * Support detecting package roots for nested Dart projects in the
 `packageRootFromFileUriString` utility method.
+* Added helper `deserialize`
+* Extended serialization for `HeapSample` and `ExtensionEvents`
```
kenzieschmoll, line 5: 10.0.0-dev.2 is already published. Please add a new section for dev.3 and bump the pubspec version. Then add these entires to the dev.3 section

packages/devtools_shared/CHANGELOG.md
```diff
@@ -1,10 +1,14 @@
...
 
 # 10.0.0-dev.1
 * Export `DTDConnectionInfo` from the `devtools_shared.dart` library instead
 of the `devtools_server.dart` library.
+* Support detecting package roots for nested Dart projects in the
+`packageRootFromFileUriString` utility method.
```
kenzieschmoll, line 11: remove this added entry. This is part of dev.2 above

PR #7704, opened 2024-05-06
Logging improvements flag

CoderDake:
Adding a flag for feature gating the new logging experience.

### DIFF
--- a/packages/devtools_app/lib/src/shared/feature_flags.dart
+++ b/packages/devtools_app/lib/src/shared/feature_flags.dart
@@ -38,9 +38,6 @@ bool get enableBeta => enableExperiments || !isExternalBuild;
 const bool _kMemoryOfflineExperiment =
     bool.fromEnvironment('memory_offline_experiment');
 
-const bool _kLoggingImprovementsExperiment =
-    bool.fromEnvironment('logging_improvements_experiment');
-
 // It is ok to have enum-like static only classes.
 // ignore: avoid_classes_with_only_static_members
 /// Flags to hide features under construction.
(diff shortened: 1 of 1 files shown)


packages/devtools_app/lib/src/shared/feature_flags.dart
```diff
@@ -70,10 +73,17 @@ abstract class FeatureFlags {
 
   /// Flag to enable DevTools extensions.
   ///
-  /// TODO(https://github.com/flutter/devtools/issues/6443): remove this flag
+  /// TODO(https://github.com/flutter/devztools/issues/6443): remove this flag
```
kenzieschmoll, line 76: typo

packages/devtools_app/lib/src/shared/feature_flags.dart
```diff
@@ -70,10 +73,17 @@ abstract class FeatureFlags {
...
   /// once extension support is added in g3.
   static bool devToolsExtensions = isExternalBuild;
 
+  /// Flag to enable the new Logging experience.
+  ///
+  /// https://github.com/flutter/devtools/issues/7703
+
+  static bool loggingImprovements =
```
kenzieschmoll, line 84: can we just set this equal to `enableExperiments`? I don't think we need a separate flag.

packages/devtools_app/lib/src/shared/feature_flags.dart
CoderDake: dang what a catch

kenzieschmoll (approved):
one comment but lgtm

PR #7715, opened 2024-05-07
Add support for hiding all screens except extension screens

kenzieschmoll:
This PR adds support for hiding all screens except extension screens with the `hide` query param: `hide=all-except-extensions`. 

Screenshot 2024-05-07 at 11 10 19 AM

Work to unblock https://github.com/flutter/flutter-intellij/issues/7195 in order to support an extensions-only tool window. CC @helin24

### DIFF
--- a/packages/devtools_app/lib/src/framework/scaffold.dart
+++ b/packages/devtools_app/lib/src/framework/scaffold.dart
@@ -43,7 +43,7 @@ class DevToolsScaffold extends StatefulWidget {
     this.page,
     List<Widget>? actions,
     this.embedMode = EmbedMode.none,
-  }) : actions = actions ?? defaultActions();
+  }) : actions = actions ?? (embedMode.embedded ? [] : defaultActions());
 
   DevToolsScaffold.withChild({
     Key? key,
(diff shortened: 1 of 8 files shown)


packages/devtools_app/lib/src/framework/scaffold.dart
```diff
@@ -315,17 +315,20 @@ class DevToolsScaffoldState extends State<DevToolsScaffold>
             serviceConnection.serviceManager.connectedAppInitialized &&
                 !offlineDataController.showingOfflineData.value &&
                 _currentScreen.showConsole(widget.embedMode);
-
+        final containsSingleSimpleScreen =
```
elliette, line 318: OOC, what is a `SimpleScreen`? Is that just a default DevTools as opposed to an Extension screen?

elliette (approved):
LGTM! One question but that's just for my knowledge (not a suggestion)

packages/devtools_app/lib/src/framework/scaffold.dart
kenzieschmoll: SimpleScreen is what we use for displaying rudimentary screens that are basically just a single child body: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/framework/scaffold.dart/#L425. The extends the Screen class though, so it lets us show simple content as part of the DevTools scaffolding structure. This is used by the `DevToolsScaffold.withChild` constructor for things like page not found messages, displaying a single offline screen, etc.

Issue #7183, opened 2024-02-08
DevTools extensions don't load for a test ran with `flutter test` or `dart test`


kenzieschmoll:
- When running `flutter test test/app_that_uses_foo_test.dart --start-paused ` , the `package:foo` DevTools extension does not load
- When running `flutter run test/app_that_uses_foo_test.dart --start-paused -d flutter-tester`, the `package:foo` DevTools extension does load.

In the `flutter test` case, `rootLibraryForMainIsolate` is returning file:///var/folders/gf/14ctmb656m3100h5z857669800fv4z/T/flutter_tools.9i1YVt/flutter_test_listener.ASUv35/listener.dart, whereas in the `flutter run` case it returns the expected path to the test file.

Since we have DevTools extensions intended for test files (e.g. Patrol), we should fix this and cherry pick to stable.

CC @bkonyi @DanTup for any ideas you have on how we can get the proper root lib or isolate for the `flutter test` case.

kenzieschmoll:
What is odd is that the Flutter sidebar DevTools menu properly detects the foo extension even though DevTools itself cannot:
![Screenshot 2024-02-07 at 4 20 01 PM](https://github.com/flutter/devtools/assets/43759233/8f460f3f-2760-4c1e-8e5f-10db49dbbea8)

We get the app root directly from VS code in the sidebar case: https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart/#L166

Whereas in DevTools itself, we try to get the URI from the isolate:
https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/extensions/extension_service.dart#L185

@DanTup how are you passing the proper URI for the debug session, and is this something we can get on the DevTools side too from the VM service?

DanTup:
> @DanTup how are you passing the proper URI for the debug session, and is this something we can get on the DevTools side too from the VM service?

VS Code knows where the root of the project is because when we start a debugging session we need to work out where the project root is to know if it's a Dart/Flutter project. So against the debug session information, we keep track of this (and we include it when we tell the sidebar a debug session started).

I guess we could pass this info along to DevTools somehow, but I'm not sure what the best way would be (because it's tied to a specific debug session). We could perhaps provide the running debug sessions to DTD and include this, or we could perhaps include it when we tell the DevTools server about the VM Services:

[...]

PR #7206, opened 2024-02-13
Detect extensions for test targets

kenzieschmoll:
fixes https://github.com/flutter/devtools/issues/7183

PR #7717, opened 2024-05-07
Detect extensions for test targets

kenzieschmoll:
Picking up work from https://github.com/flutter/devtools/pull/7206, which was closed due to merge conflicts and some changes that have since become obsolete.

Fixes https://github.com/flutter/devtools/issues/7183


packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -180,22 +175,82 @@ class ServiceConnectionManager {
     await serviceManager.isolateManager.init(isolates);
   }
 
+  /// Returns the package root URI for the connected app.
+  ///
+  /// This should be the directory up the tree from the debugging target that
+  /// contains the .dart_tool/package_config.json file.
+  ///
+  /// This method contains special logic for detecting the package root for
+  /// test targets (i.e. a VM service connections spawned from `dart test` or
```
bkonyi, line 184: The uberist of nits: `i.e.,`

packages/devtools_app/lib/src/service/service_manager.dart
```diff
@@ -180,22 +175,82 @@ class ServiceConnectionManager {
...
+  Future<Library?> _mainIsolateRootLibrary() async {
+    final ref = (await serviceManager.isolateManager.waitForMainIsolateState())
+        ?.isolateNow
+        ?.rootLib;
+    if (ref == null) return null;
+    final library = await serviceManager.service!.getObject(
+      serviceManager.isolateManager.mainIsolate.value!.id!,
+      ref.id!,
+    );
+    return library is Library ? library : null;
```
bkonyi, line 234: Should we `assert(library is Library)` and then `return library as Library` instead? This should never resolve to anything but a `Library`.

bkonyi (approved):

packages/devtools_app/lib/src/service/service_manager.dart
kenzieschmoll: couldn't this technically be a Sentinel?

kenzieschmoll: added a try / catch in case a SentinelException is thrown. PTAL.

bkonyi: Yep, that's the right thing to do. LGTM.

PR #7738, opened 2024-05-10
Copy logging screen and controller to V2 locations

CoderDake:
This PR copies over the logging controller and logging screen. The only difference is that the table contents have been replaced with empty `Container()`s. These empty containers were used to avoid having to copy over too many unnecessary components.


packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_controller_v2.dart
```diff
@@ -0,0 +1,879 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: 2024

kenzieschmoll (approved):

PR #7740, opened 2024-05-10
Remove fragile test

gspencergoog:
## Description

In order to land a Flutter change that modifies the initial parts of the widget tree (it adds a `FocusScope` around views), I'm removing this fragile test.  It can either be permanently (since it's fragile), or it can be temporarily until the Flutter change has rolled in.

I can't really just modify this test, since that would break the devtools build until the Flutter change is in, and also I need to make devtools customer tests pass _with_ the change, which are mutually exclusive choices.

## Related PRs
 - https://github.com/flutter/flutter/pull/143259


kenzieschmoll (commented):
Instead of removing this, can we just comment this line out with a TODO and issue? once the flutter change lands, we can re-enable this with the updated golden.

gspencergoog:
Well, I was just removing it because we don't usually check in commented-out code (per the style guide).  How about I just add a TODO to revert this PR (with an updated golden) once the Flutter change lands?

kenzieschmoll (approved):

PR #7761, opened 2024-05-15
Update extensions example readme and assets

kenzieschmoll:
The only file that needs a review is the `example/README.md`. The rest of the changes are generated changes from updating the extension assets for the `foo` extension.


elliette (approved):
One typo but otherwise lgtm!

PR #7792, opened 2024-05-21
Make SnapshotDataItem encodable.

polina-c:


packages/devtools_app/lib/src/shared/memory/heap_graph_loader.dart
```diff
@@ -49,15 +48,29 @@ class HeapGraphLoaderFile implements HeapGraphLoader {
...
   final DateTime created;
 
   @override
   (HeapSnapshotGraph, DateTime) load() {
-    return (HeapSnapshotGraph.fromChunks(chunks), created);
+    return (graph, created);
   }
 }
+
+// /// Loads a heap snapshot from `List<ByteData>` created with HeapSnapshotGraph.toChunks.
```
kenzieschmoll, line 64: remove?

packages/devtools_app/lib/src/shared/primitives/encoding.dart
```diff
@@ -0,0 +1,101 @@
+// Copyright 2024 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:typed_data';
+
+import 'package:vm_service/vm_service.dart';
+
+abstract class EncodeDecode<T> {
```
kenzieschmoll, line 10: dart doc for all these classes

packages/devtools_app/lib/src/shared/primitives/encoding.dart
```diff
@@ -0,0 +1,101 @@
...
+
+  T decode(Object value);
+
+  T? decodeNullable(Object? value) {
+    if (value == null) return null;
+    return decode(value);
+  }
+}
+
+class HeapSnapshotGraphEncodeDecode extends EncodeDecode<HeapSnapshotGraph> {
```
kenzieschmoll, line 21: add tests for all these encode / decode classes

packages/devtools_app/lib/src/shared/primitives/encoding.dart
```diff
@@ -0,0 +1,101 @@
...
+
+  if (value is ByteData) {
+    return ByteDataEncodeDecode.instance.toEncodable(value);
+  }
+
+  if (value is DateTime) {
+    return DateTimeEncodeDecode.instance.toEncodable(value);
+  }
+
+  return value;
```
kenzieschmoll, line 99: Use a switch expression here instead.
```
return switch (value) {
  HeapSnapshotGraph => HeapSnapshotGraphEncodeDecode.instance.toEncodable(value),
  ...
  _ => null,
};
```
or even simpler (you'll have to try this to see if the switch expression is recognized as an EncodeDecode object to make this valid)
```
return (switch (value) {
  HeapSnapshotGraph => HeapSnapshotGraphEncodeDecode.instance,
  ...
  _ => null,
})?.toEncodable(value);
```

polina-c: I tried. There is no type promotion here :(

kenzieschmoll: You should still be able to use the first suggestion though right?
```
return switch (value) {
  HeapSnapshotGraph => HeapSnapshotGraphEncodeDecode.instance.toEncodable(value),
  ...
  _ => null,
};
```

packages/devtools_app/lib/src/shared/primitives/encoding.dart
```diff
@@ -0,0 +1,100 @@
...
+
+  T decode(Object value);
+
+  T? decodeNullable(Object? value) {
+    if (value == null) return null;
+    return decode(value);
+  }
+}
+
+/// Encodes and decodes HeapSnapshotGraph.
```
kenzieschmoll, line 22: put HeapSnapshotGraph in brackets []. Same for the other class names referenced in dart doc below.

packages/devtools_app/lib/src/shared/primitives/encoding.dart
polina-c: toEncodable takes HeapSnapshotGraph as parameter, so I will need to do as HeapSnapshotGraph, with risk to make mistake and fail casting, that I do not want

kenzieschmoll: This should give you type promotion
```
return switch (value) {
  (HeapSnapshotGraph graph) => HeapSnapshotGraphEncodeDecode.instance.toEncodable(graph),
  ...
  _ => null,
};
```

polina-c: it did!!!
thanks :)

packages/devtools_app/lib/src/shared/primitives/encoding.dart
```diff
@@ -0,0 +1,93 @@
...
+  DateTime decode(Object value) {
+    if (value is DateTime) return value;
+    return DateTime.fromMicrosecondsSinceEpoch(value as int);
+  }
+}
+
+/// Function to be passed to [jsonEncode] to enable encoding for more types.
+Object? toEncodable(Object? value) {
+  return switch (value) {
+    (null) => null,
```
kenzieschmoll, line 86: you can remove this. this is caught by the default case

polina-c: nope, default case should return value
fixed

kenzieschmoll (approved):

Issue #4880, opened 2022-12-02
Add source links to Rebuild Stats location column


kenzieschmoll:
We should either link to show this file in the DevTools debugger or link straight to the user IDE when we have custom stream functionality to do this.

![Screen Shot 2022-12-02 at 10 49 16 AM](https://user-images.githubusercontent.com/43759233/205364850-5174430f-98d9-4eaa-b4b7-1066efdaa3ee.png)

Issue #4564, opened 2022-09-30
Performance screen: implement rebuild indicators


jacob314:
- [x] land changes to Flutter needed to handle reconnecting to a running app (https://github.com/jacob314/flutter/tree/rebuild_indicators)
- [ ] support showing the source location for a rebuild.

srawlins:
Closing as obsolete.

PR #7810, opened 2024-05-23
Post ToolEvent when widget in rebuild table is selected

helin24:
Part of https://github.com/flutter/devtools/issues/4564

This sends a navigate to code `ToolEvent` to IDEs so that clicking on a widget in the rebuilds table opens the associated code in an editor.

This change still has an issue in that IntelliJ expects an event to include a field `type`, which this one doesn't. I'm not sure if that will require a change in IntelliJ code or in DDS somewhere. It works without further changes for VS Code.

It seems like since this is still gated by a feature flag that I shouldn't add a release note yet, but let me know if otherwise.

Fixes https://github.com/flutter/devtools/issues/4880

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/panes/rebuild_stats/rebuild_stats.dart
+++ b/packages/devtools_app/lib/src/screens/performance/panes/rebuild_stats/rebuild_stats.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:dds_service_extensions/dds_service_extensions.dart';
 import 'package:devtools_app_shared/service.dart';
 import 'package:devtools_app_shared/ui.dart';
 import 'package:devtools_app_shared/utils.dart';
@@ -10,6 +11,7 @@ import 'package:flutter/material.dart';
 
 import '../../../../service/service_extension_widgets.dart';
 import '../../../../service/service_extensions.dart' as extensions;
+import '../../../../service/vm_service_wrapper.dart';
 import '../../../../shared/analytics/constants.dart' as gac;
 import '../../../../shared/common_widgets.dart';
 import '../../../../shared/globals.dart';
@@ -171,6 +173,8 @@ class _RebuildTableState extends State<RebuildTable> {
   /// column objects for the same column.
   final _columnCache = <String, _RebuildCountColumn>{};
 
+  VmServiceWrapper? get _service => serviceConnection.serviceManager.service;
+
   List<_RebuildCountColumn> get _metricsColumns {
     final columns = <_RebuildCountColumn>[];
     for (var i = 0; i < widget.metricNames.length; i++) {
@@ -207,8 +211,15 @@ class _RebuildTableState extends State<RebuildTable> {
         defaultSortColumn: _metricsColumns.first,
         defaultSortDirection: sortDirection,
         onItemSelected: (item) {
-          // TODO(jacobr): navigate to selected widget in IDE or display the
-          // content side by side with the rebuild table.
+          final location = item?.location;
+          if (location?.path != null) {
+            _service?.postEvent('ToolEvent', 'navigate', <String, Object>{
+              'fileUri': location?.path ?? '',
+              'line': location?.line ?? 0,
+              'column': location?.column ?? 0,
+              'source': 'flutter.performance'
+            });
+          }
         },
       ),
     );
--- a/packages/devtools_app/lib/src/screens/performance/panes/rebuild_stats/rebuild_stats.dart
+++ b/packages/devtools_app/lib/src/screens/performance/panes/rebuild_stats/rebuild_stats.dart
@@ -2,7 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'package:dds_service_extensions/dds_service_extensions.dart';
 import 'package:devtools_app_shared/service.dart';
 import 'package:devtools_app_shared/ui.dart';
 import 'package:devtools_app_shared/utils.dart';
@@ -213,12 +212,12 @@ class _RebuildTableState extends State<RebuildTable> {
         onItemSelected: (item) async {
           final location = item?.location;
           if (location?.path != null) {
-            await _service?.postEvent('ToolEvent', 'navigate', <String, Object>{
-              'fileUri': location?.path ?? '',
-              'line': location?.line ?? 0,
-              'column': location?.column ?? 0,
-              'source': 'flutter.performance',
-            });
+            await _service?.navigateToCode(
+              fileUri: location?.path ?? '',
+              line: location?.line ?? 0,
+              column: location?.column ?? 0,
+              source: 'flutter.performance',
+            );
           }
         },
       ),
--- a/packages/devtools_app/lib/src/screens/performance/panes/rebuild_stats/rebuild_stats.dart
+++ b/packages/devtools_app/lib/src/screens/performance/panes/rebuild_stats/rebuild_stats.dart
@@ -211,12 +211,12 @@ class _RebuildTableState extends State<RebuildTable> {
         defaultSortDirection: sortDirection,
         onItemSelected: (item) async {
           final location = item?.location;
-          if (location?.path != null) {
+          if (location?.fileUri != null) {
             await _service?.navigateToCode(
-              fileUri: location?.path ?? '',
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/screens/performance/panes/rebuild_stats/rebuild_stats.dart
```diff
@@ -206,9 +210,16 @@ class _RebuildTableState extends State<RebuildTable> {
             ValueKey<String?>('${location.location.id}'),
         defaultSortColumn: _metricsColumns.first,
         defaultSortDirection: sortDirection,
-        onItemSelected: (item) {
-          // TODO(jacobr): navigate to selected widget in IDE or display the
-          // content side by side with the rebuild table.
+        onItemSelected: (item) async {
+          final location = item?.location;
+          if (location?.path != null) {
+            await _service?.postEvent('ToolEvent', 'navigate', <String, Object>{
```
kenzieschmoll, line 216: Can you add this as an extension method in this file? https://github.com/flutter/devtools/blob/master/packages/devtools_app_shared/lib/src/service/service_utils.dart/#L7

We have had DevTools extension authors inquiring about how to do this, so would be good to expose this there.

packages/devtools_app/lib/src/screens/performance/panes/rebuild_stats/rebuild_stats.dart
```diff
@@ -206,9 +210,16 @@ class _RebuildTableState extends State<RebuildTable> {
...
+              'fileUri': location?.path ?? '',
+              'line': location?.line ?? 0,
+              'column': location?.column ?? 0,
+              'source': 'flutter.performance',
```
kenzieschmoll, line 220: how about 'devtools.rebuildStats' for source?

packages/devtools_app/lib/src/screens/performance/panes/rebuild_stats/rebuild_stats.dart
```diff
@@ -206,9 +210,16 @@ class _RebuildTableState extends State<RebuildTable> {
...
         defaultSortColumn: _metricsColumns.first,
         defaultSortDirection: sortDirection,
-        onItemSelected: (item) {
-          // TODO(jacobr): navigate to selected widget in IDE or display the
-          // content side by side with the rebuild table.
+        onItemSelected: (item) async {
+          final location = item?.location;
+          if (location?.path != null) {
+            await _service?.postEvent('ToolEvent', 'navigate', <String, Object>{
+              'fileUri': location?.path ?? '',
```
kenzieschmoll, line 217: also nit: if the value of `location.path` is really a file URI, then the `path` field should be renamed to `fileUri`

packages/devtools_app/lib/src/screens/performance/panes/rebuild_stats/rebuild_stats.dart
helin24: Moved to `devtools_app_shared` - let me know if that interface looks good.

How does versioning work here? devtools_app seems to pick up the changes to the shared package, but I see that the dependency in devtools_app is for `devtools_app_shared: ^0.2.0-dev.0` - does this need to be updated?

packages/devtools_app/lib/src/screens/performance/panes/rebuild_stats/rebuild_stats.dart
```diff
@@ -244,7 +243,7 @@ class _LocationColumn extends ColumnData<RebuildLocationStats> {
 
   @override
   String getValue(RebuildLocationStats dataObject) {
-    final path = dataObject.location.path;
+    final path = dataObject.location.fileUri;
```
kenzieschmoll, line 246: rename var to fileUriString

packages/devtools_app/lib/src/screens/performance/panes/rebuild_stats/rebuild_stats_model.dart
```diff
@@ -276,17 +276,17 @@ class RebuildCountModel {
...
 
 class Location {
-  Location({required this.id, this.path, this.line, this.column, this.name});
+  Location({required this.id, this.fileUri, this.line, this.column, this.name});
 
   final int id;
 
   /// Either all of path, line, column, and name are null or none are.
-  String? path;
+  String? fileUri;
```
kenzieschmoll, line 284: sorry for bad initial suggestion. Can we change this to fileUriString? This is consistent with other parts of DevTools where we follow this pattern:

- variables named "path" => /actually/a/file/path
- variables named "fileUriString" => "file:///this/is/a/string/that/equals/uri/toFilePath"
- variables named "uri" => this is a Uri object

packages/devtools_app_shared/lib/src/service/service_utils.dart
```diff
@@ -46,6 +47,23 @@ extension VmServiceExtension on VmService {
   ) async {
     await forEachIsolateHelper(this, callback);
   }
+
+  /// Posts an event to jump to code at the specified [line] and [column] in the
+  /// specified [fileUri]. The [source] should indicate the tool and/or feature
+  /// that is making the jump to code request.
+  Future<void> navigateToCode({
+    required String fileUri,
```
kenzieschmoll, line 55: rename to fileUriString

packages/devtools_app_shared/lib/src/service/service_utils.dart
```diff
@@ -46,6 +47,23 @@ extension VmServiceExtension on VmService {
   ) async {
     await forEachIsolateHelper(this, callback);
   }
+
+  /// Posts an event to jump to code at the specified [line] and [column] in the
+  /// specified [fileUri]. The [source] should indicate the tool and/or feature
+  /// that is making the jump to code request.
```
kenzieschmoll, line 53: can you add (i.e. "someTool.someFeature")

packages/devtools_app/lib/src/screens/performance/panes/rebuild_stats/rebuild_stats_model.dart
helin24: Ah yeah, that makes sense.

packages/devtools_app_shared/lib/src/service/service_utils.dart
```diff
@@ -48,17 +48,21 @@ extension VmServiceExtension on VmService {
...
-  /// Posts an event to jump to code at the specified [line] and [column] in the
-  /// specified [fileUri]. The [source] should indicate the tool and/or feature
-  /// that is making the jump to code request.
+  /// Posts an event to jump to code.
+  ///
+  /// The event directs an IDE to move to the [line] and [column] in the
+  /// specified [fileUriString], which is expected to be a proper file URI (i.e.
+  /// starts with "file://"). The [source] should indicate the tool and/or
+  /// feature (i.e. in the format "someTool.someFeature") that is making the
+  /// jump to code request.
```
kenzieschmoll, line 57: add one more section. If there are no IDEs listening for this event, then this will be a no-op.

packages/devtools_app/lib/src/screens/performance/panes/rebuild_stats/rebuild_stats.dart
```diff
@@ -233,7 +243,7 @@ class _LocationColumn extends ColumnData<RebuildLocationStats> {
 
   @override
   String getValue(RebuildLocationStats dataObject) {
-    final path = dataObject.location.path;
+    final path = dataObject.location.fileUriString;
```
kenzieschmoll, line 246: super nit: rename path to fileUriString here too

kenzieschmoll (commented):
Please also add a changelog entry for devtools_app_shared. You'll need to bump the dev version in the changelog to create a new section and the pubspec to match

kenzieschmoll (approved):
lgtm after changelog entry and pubspec update.

packages/devtools_app/pubspec.yaml
```diff
@@ -24,7 +24,7 @@ dependencies:
   collection: ^1.15.0
   dap: ^1.1.0
   dds_service_extensions: ^2.0.0
-  devtools_app_shared: ^0.2.0-dev.0
+  devtools_app_shared: ^0.3.0-dev.0
```
kenzieschmoll, line 27: can we just bump the dev version instead of doing a minor bump?
`0.2.0-dev.1`

kenzieschmoll (approved):

Issue #7799, opened 2024-05-22
[deep link] Support ios app check


hangyujin:
Deep link validator currently support Android app check and web check,

Plan to add some ios app check 
---> if a domain has AASA file on web, but it's not added in flutter project's ios app settings, we can display some check result to hint developer to update their ios app settings.

DD: go/flutter-deeplink-cross-checking

PR #7812, opened 2024-05-23
[deep link] [ios app check] Request ios build option and add a dropdown for it

hangyujin:
issue : https://github.com/flutter/devtools/issues/7799 
This is part of adding ios app check,

my plan is :
step 1  add ios build option  <---- this PR
step 2  load ios links
step 3  cross check them with android links

image

## Pre-launch Checklist

- [ ] I signed the [CLA].

![build.yaml badge]

### DIFF
--- a/packages/devtools_app/lib/src/screens/deep_link_validation/deep_link_list_view.dart
+++ b/packages/devtools_app/lib/src/screens/deep_link_validation/deep_link_list_view.dart
@@ -39,20 +39,17 @@ class DeepLinkListView extends StatefulWidget {
 
 class _DeepLinkListViewState extends State<DeepLinkListView>
     with ProvidedControllerMixin<DeepLinksController, DeepLinkListView> {
-  List<String> get androidVariants =>
-      controller.selectedProject.value!.androidVariants;
-
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
     initController();
     callWhenControllerReady((_) {
-      int releaseVariantIndex = controller
-          .selectedProject.value!.androidVariants
-          .indexWhere((variant) => variant.toLowerCase().contains('release'));
-      // If not found, default to 0.
-      releaseVariantIndex = max(releaseVariantIndex, 0);
-      controller.selectedVariantIndex.value = releaseVariantIndex;
+      controller.selectedAndroidVariantIndex.value = _getReleaseVariantIndex(
+        controller.selectedProject.value!.androidVariants,
+      );
+      controller.selectedIosConfigurationIndex.value = _getReleaseVariantIndex(
+        controller.selectedProject.value!.iosBuildOptions.configurations,
+      );
     });
   }
 
@@ -70,6 +67,13 @@ class _DeepLinkListViewState extends State<DeepLinkListView>
         ),
       ),
     );
+  }
+
+  int _getReleaseVariantIndex(List<String> variants) {
+    final index = variants
+        .indexWhere((variant) => variant.toLowerCase().contains('release'));
+    // If not found, default to 0.
+    return max(index, 0);
   }
 }
 
@@ -273,24 +277,19 @@ class _DeepLinkListViewTopPanel extends StatelessWidget {
       includeBottomBorder: false,
       tall: true,
       title: Row(
-        mainAxisAlignment: MainAxisAlignment.spaceBetween,
         children: [
           Text(
             'Validate and fix',
             style: Theme.of(context).textTheme.titleSmall,
           ),
-          ValueListenableBuilder(
-            valueListenable: controller.selectedVariantIndex,
-            builder: (_, value, __) {
-              return _AndroidVariantDropdown(
-                androidVariants:
-                    controller.selectedProject.value!.androidVariants,
-                index: value,
-                onVariantIndexSelected: (index) {
-                  controller.selectedVariantIndex.value = index;
-                },
-              );
-            },
+          const Spacer(),
+          _VariantDropdown(
+            os: PlatformOS.android,
+            controller: controller,
+          ),
+          _VariantDropdown(
+            os: PlatformOS.ios,
+            controller: controller,
           ),
         ],
       ),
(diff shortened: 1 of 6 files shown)


packages/devtools_app/lib/src/screens/deep_link_validation/deep_link_list_view.dart
```diff
@@ -248,61 +252,64 @@ class _DeepLinkListViewTopPanel extends StatelessWidget {
...
-                  controller.selectedVariantIndex.value = index;
-                },
-              );
-            },
+          const Spacer(),
+          _VariantDropdown(
+            os: PlatformOS.android,
+            controller: controller,
+          ),
+          _VariantDropdown(
```
kenzieschmoll, line 265: add a SizedBox(width: denseSpacing) widget between these

kenzieschmoll, line 265: If we are connected to a running application, would it make more sense to only show whichever of these is relevant for the connected app platform instead of both?

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_controller.dart
```diff
@@ -139,18 +139,20 @@ class DisplayOptions {
 
 class DeepLinksController extends DisposableController {
   DeepLinksController() {
-    selectedVariantIndex.addListener(_handleSelectedVariantIndexChanged);
+    selectedAndroidVariantIndex
+        .addListener(_handleSelectedAndroidVariantIndexChanged);
```
kenzieschmoll, line 143: do we also need to add a listener for the ios variant index here?

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_controller.dart
```diff
@@ -139,18 +139,20 @@ class DisplayOptions {
 
 class DeepLinksController extends DisposableController {
```
kenzieschmoll, line 140: mix in AutoDisposeControllerMixin and then you can add these listeners with addAutoDisposeListener, which means you don't have to manually remove them in dispose

packages/devtools_app/lib/src/screens/deep_link_validation/project_root_selection/select_project_view.dart
```diff
@@ -59,25 +60,47 @@ class _SelectProjectViewState extends State<SelectProjectView>
...
+  }
+
+  Future<XcodeBuildOptions> _requestiOSBuildOptions(String directory) async {
+    ga.timeStart(
+      gac.deeplink,
+      gac.AnalyzeFlutterProject.loadIosBuildOptions.name,
+    );
+    final XcodeBuildOptions iosBuildOptions =
+        await server.requestIosBuildOptions(directory);
+    ga.timeEnd(
```
kenzieschmoll, line 87: do we need the same logic to check if the iosBuildOptions are empty like we have to the android variants?

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_controller.dart
hangyujin: Yes, let me add a todo here

this pr just added ios build option, will add the listener  when loading ios links in next PR,

packages/devtools_app/lib/src/screens/deep_link_validation/project_root_selection/select_project_view.dart
```diff
@@ -59,25 +60,47 @@ class _SelectProjectViewState extends State<SelectProjectView>
...
-    setState(() {
-      _retrievingFlutterProject = true;
-    });
+  Future<List<String>> _requestAndridVariants(String directory) async {
     ga.timeStart(gac.deeplink, gac.AnalyzeFlutterProject.loadVariants.name);
     final List<String> androidVariants =
         await server.requestAndroidBuildVariants(directory);
-    if (!mounted) {
+    if (androidVariants.isEmpty) {
+      // If the project is not a Flutter project, cancel timing and return an empty list.
```
kenzieschmoll, line 68: should this still check `|| !mounted`?

packages/devtools_app/lib/src/screens/deep_link_validation/deep_link_list_view.dart
hangyujin: If we only have only one dropdown here, we should also run checks and show result for only one platform.

I think it's more consistent to always display the check result for both platform, but will discuss this with team today for sure.

if a domain is detected on both Android and iOS, the connected running app is on Android, and the result table displayed as :
[Domain]                        [OS]
[example.com]               [Android]

I might wonder why iOS is not here

packages/devtools_app/lib/src/screens/deep_link_validation/project_root_selection/select_project_view.dart
hangyujin: i think it's fine, the check here is just to check if the directory is a flutter project, check them in either Android/iOS for once should be enough, no need to do it twice.

packages/devtools_app/lib/src/screens/deep_link_validation/project_root_selection/select_project_view.dart
hangyujin: done

packages/devtools_app/lib/src/screens/deep_link_validation/deep_links_controller.dart
hangyujin: nice! thanks!

packages/devtools_app/lib/src/screens/deep_link_validation/deep_link_list_view.dart
```diff
@@ -71,6 +68,13 @@ class _DeepLinkListViewState extends State<DeepLinkListView>
       ),
     );
   }
+
+  int _getReleaseVariantIndex(List<String> variants) {
+    final index = variants
+        .indexWhere((variant) => variant.toLowerCase().contains('release'));
```
kenzieschmoll, line 74: use the `caseInsensitiveContains` utility method from `shared/primitives/utils.dart`

packages/devtools_app/lib/src/screens/deep_link_validation/project_root_selection/select_project_view.dart
```diff
@@ -59,25 +60,47 @@ class _SelectProjectViewState extends State<SelectProjectView>
...
-  void _handleValidateProject(String directory) async {
-    setState(() {
-      _retrievingFlutterProject = true;
-    });
+  Future<List<String>> _requestAndridVariants(String directory) async {
     ga.timeStart(gac.deeplink, gac.AnalyzeFlutterProject.loadVariants.name);
     final List<String> androidVariants =
         await server.requestAndroidBuildVariants(directory);
-    if (!mounted) {
+    if (androidVariants.isEmpty || !mounted) {
```
kenzieschmoll, line 67: is it possible that `androidVariants` could be empty even if this is a Flutter project?

packages/devtools_app/lib/src/screens/deep_link_validation/project_root_selection/select_project_view.dart
```diff
@@ -59,25 +60,47 @@ class _SelectProjectViewState extends State<SelectProjectView>
...
+      gac.AnalyzeFlutterProject.loadIosBuildOptions.name,
+    );
+    return iosBuildOptions;
+  }
+
+  void _handleValidateProject(String directory) async {
+    setState(() {
+      _retrievingFlutterProject = true;
+    });
+    final List<String> androidVariants =
```
kenzieschmoll, line 98: before we request android variants and ios build options, I would recommend checking if this is a flutter app directly instead of inferring this value based on the set of android build variants returned from the server.

```
final connected = serviceConnection.serviceManager.connectedState.value.connected;
if (connected && !serviceConnection.serviceManager.connectedApp!.isFlutterAppNow) {
  // show dialog for non-flutter project
  return;
}
// then check android build variants, which will still need to
// handle the case where the build variants come back empty
// which could mean that a non-flutter project was selected when
// we do not have a live app connection
```

hangyujin:
I will add a flag first https://github.com/flutter/devtools/pull/7842 and put the feature changes behind the flag.

kenzieschmoll (approved):

PR #7820, opened 2024-05-24
Base Logging V2 behaviour

CoderDake:
*Replace this paragraph with a description of what this PR is changing or adding, and why. Consider including before/after screenshots.*

Implement the main interaction for the logging experience.

Basic tests are working, more developed tests will be added in future PRs as functionality is added.
This PR is already way too big :P

The ability to load all of the log, resize, and scroll performantly has been added.


packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -297,3 +298,78 @@ Future<void> launchUrlWithErrorHandling(String url) async {
...
+  Future<bool> doWork(
+    int length,
+  ) {
+    final completer = Completer<bool>();
+    final localWorkId = ++_workId;
+    final sw = Stopwatch();
+
+    Function(int i)? doChunkWork;
+
+    doChunkWork = (i) {
```
kenzieschmoll, line 341: make this a local function directly instead of having a variable to store the function:
```
void doChunkWork(int i) {
  ...
}
```

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -297,3 +298,78 @@ Future<void> launchUrlWithErrorHandling(String url) async {
...
+      if (i >= length) {
+        sw.stop();
+        return completer.complete(true);
+      }
+
+      _sw.reset();
+      _sw.start();
```
kenzieschmoll, line 348: use cascades `..`

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -297,3 +298,78 @@ Future<void> launchUrlWithErrorHandling(String url) async {
...
+
+      progressCallback(j / length);
+      Future.delayed(const Duration(), () {
+        doChunkWork!.call(i + _chunkSize);
+      });
+    };
+    sw.start();
+
+    if (length <= 0) {
+      return Future.value(true);
```
kenzieschmoll, line 367: put this at the very beginning of the doWork method since it is an early exit

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -297,3 +298,78 @@ Future<void> launchUrlWithErrorHandling(String url) async {
...
+      }
+
+      _sw.reset();
+      _sw.start();
+
+      final J = min(length, i + _chunkSize);
+      int j = i;
+      for (; j < J; j++) {
+        // If our localWorkId is no longer active, then do not continue working
+        if (localWorkId != _workId) return completer.complete(false);
```
kenzieschmoll, line 354: do you need to stop the stopwatch before returning here?

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -297,3 +298,78 @@ Future<void> launchUrlWithErrorHandling(String url) async {
...
+    required this.callback,
+    required this.progressCallback,
+  }) : _chunkSize = chunkSize;
+
+  final int _chunkSize;
+  int _workId = 0;
+  void Function(int) callback;
+  void Function(double progress) progressCallback;
+
+  final _sw = Stopwatch();
```
kenzieschmoll, line 322: what is this needed for? it doesn't look like we use the value of _sw anywhere

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -297,3 +298,78 @@ Future<void> launchUrlWithErrorHandling(String url) async {
...
+    doChunkWork = (i) {
+      if (i >= length) {
+        sw.stop();
+        return completer.complete(true);
+      }
+
+      _sw.reset();
+      _sw.start();
+
+      final J = min(length, i + _chunkSize);
```
kenzieschmoll, line 350: can you use a more descriptive name here than `J`? Also, variable names should be camelcase

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -297,3 +298,78 @@ Future<void> launchUrlWithErrorHandling(String url) async {
...
+      int j = i;
```
kenzieschmoll, line 351: put this in the for loop specification

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -297,3 +298,78 @@ Future<void> launchUrlWithErrorHandling(String url) async {
...
+        // If our localWorkId is no longer active, then do not continue working
+        if (localWorkId != _workId) return completer.complete(false);
+        callback(j);
+      }
+      _sw.stop();
+
+      progressCallback(j / length);
+      Future.delayed(const Duration(), () {
+        doChunkWork!.call(i + _chunkSize);
+      });
```
kenzieschmoll, line 362: nit: this can be replaced with an existing utility:
```
await delayToReleaseUiThread();
doChunkWork(i + _chunkSize);
```

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
```diff
@@ -0,0 +1,83 @@
+import 'dart:async';
```
kenzieschmoll, line 1: copyright

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
```diff
@@ -0,0 +1,83 @@
...
+
+import 'package:flutter/foundation.dart';
+import '../../../shared/utils.dart';
```
kenzieschmoll, line 4: new line between package imports and project imports

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
```diff
@@ -0,0 +1,83 @@
...
+import 'logging_controller_v2.dart';
+import 'logging_table_row.dart';
+
+class LoggingTableModel extends ChangeNotifier {
```
kenzieschmoll, line 8: dart doc

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
```diff
@@ -0,0 +1,83 @@
...
+  final Map<int, double> cachedOffets = {};
+
+  late final InterruptableChunkWorker _worker;
+
+  /// Represents the state of reloading the height caches.
+  ///
+  /// When null, then the cache is not loading.
+  /// When double, then the value is represents how much progress has been made.
+  ValueListenable<double?> get cacheLoadProgress => _cacheLoadProgress;
+  final _cacheLoadProgress = ValueNotifier<double?>(null);
```
kenzieschmoll, line 30: this notifier needs to be disposed. Add a `dispose` method, and then call it from the appropriate place in from the owner of the LoggingTableModel object

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,109 @@
+import 'package:devtools_app_shared/ui.dart';
+import 'package:flutter/material.dart';
+
+import '../../../../devtools_app.dart';
```
kenzieschmoll, line 4: don't import directly

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,109 @@
+import 'package:devtools_app_shared/ui.dart';
```
kenzieschmoll, line 1: copyright

packages/devtools_app/test/shared/utils_test.dart
```diff
@@ -1,3 +1,5 @@
+import 'dart:async';
```
kenzieschmoll, line 1: copyright

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,325 @@
...
+          context: context,
+          builder: (BuildContext context) => AlertDialog(title: Text(message)),
+        ),
+      ),
+    );
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return LayoutBuilder(
```
kenzieschmoll, line 78: break this up into helper widgets. Long build methods are hard to follow

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
```diff
@@ -0,0 +1,83 @@
...
+
+class LoggingTableModel extends ChangeNotifier {
+  LoggingTableModel() {
+    _worker = InterruptableChunkWorker(
+      callback: getRowHeight,
+      progressCallback: (progress) => _cacheLoadProgress.value = progress,
+    );
+  }
+
+  final List<LogDataV2> _logs = [];
```
kenzieschmoll, line 16: types on RHS for all these

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,109 @@
...
+      text: text,
+      style: _metaDataStyle,
+    );
+  }
+
+  static const TextStyle _detailsStyle = TextStyle();
+  static const TextStyle _metaDataStyle = TextStyle();
+
+  static const double _dividerHeight = 10.0;
+  static double calculateRowHeight(LogDataV2 log, double width) {
```
kenzieschmoll, line 37: new line between class members

packages/devtools_app/lib/src/shared/utils.dart
CoderDake: Doing this seems to freeze the UI.

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
CoderDake: gosh that auto import always gets me.

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_controller_v2.dart
```diff
@@ -634,15 +627,21 @@ class LogDataV2 with SearchableDataMixin {
 
   String? get details => _details;
 
-  bool get needsComputing => detailsComputer != null;
+  ValueListenable<bool> get needsComputing => _needsComputing;
+  final ValueNotifier<bool> _needsComputing = ValueNotifier<bool>(true);
```
kenzieschmoll, line 631: can we change this to be `computed` or `detailsComputed` with a default value of false? This would make the logic a bit easier to follow.

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
```diff
@@ -0,0 +1,117 @@
...
+  /// in the background.
+  set tableWidth(double width) {
+    _tableWidth = width;
+    cachedHeights.clear();
+    cachedOffets.clear();
+    unawaited(_preFetchRowHeights());
+  }
+
+  /// Get the filtered log at [index].
+  LogDataV2 getFilteredLog(int index) => _filteredLogs[index];
```
kenzieschmoll, line 62: avoid starting method names with 'get' per effective dart style guidelines. Maybe 'filteredLogAt` instead?

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
```diff
@@ -0,0 +1,117 @@
...
+  double _tableWidth = 0.0;
+
+  /// The total number of logs being held by the [LoggingTableModel].
+  int get logCount => _logs.length;
+
+  /// The number of filtered logs.
+  int get filteredLogCount => _filteredLogs.length;
+
+  /// The number of selected logs.
+  int get selectedLogCount => _selectedLogs.length;
```
kenzieschmoll, line 73: how can there be more than one selection? Are you referring to text selection here?

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
```diff
@@ -0,0 +1,117 @@
...
+
+  /// Clears all of the logs from the model.
+  void clear() {
+    _logs.clear();
+    _filteredLogs.clear();
+    notifyListeners();
+  }
+
+  /// Get the offset of a filtered log, at [index], from the top of the list of filtered logs.
+  double getFilteredLogOffset(int _) {
```
kenzieschmoll, line 92: same comment wrt starting methods with "get". Maybe `filteredLogOffsetAt` instead

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
```diff
@@ -0,0 +1,117 @@
...
+    throw 'Implement this when needed';
```
kenzieschmoll, line 93: throw and Exception object instead of throwing a String

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /// Get the height of a filtered Log at [index].
+  double getFilteredLogHeight(int index) {
```
kenzieschmoll, line 97: same

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
```diff
@@ -0,0 +1,117 @@
...
+    final cachedHeight = cachedHeights[index];
+    if (cachedHeight != null) return cachedHeight;
+    final newHeight = LoggingTableRow.calculateRowHeight(
+      _logs[index],
+      _tableWidth,
+    );
+    cachedHeights[index] = newHeight;
+    return newHeight;
```
kenzieschmoll, line 105: nit: this can be condensed:
```
return cachedHeights[index] = LoggingTableRow.calculateRowHeight(
      _logs[index],
      _tableWidth,
    );
```

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  Future<bool> _preFetchRowHeights() async {
+    final didComplete = await _worker.doWork(
+      _logs.length,
```
kenzieschmoll, line 110: nit: I think this will fit on one line

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,117 @@
...
+  static TextSpan _metadataSpan(String text) {
+    return TextSpan(
+      text: text,
+      style: _metaDataStyle,
+    );
+  }
+
+  static const _detailsStyle = TextStyle();
+
+  static const _metaDataStyle = TextStyle();
```
kenzieschmoll, line 41: why do we need these empty styles?

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,117 @@
...
+
+  static const double _dividerHeight = 10.0;
```
kenzieschmoll, line 43: remove type on LHS

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,117 @@
...
+    // TODO(danchevalier): Improve row2 height by manually flowing metadas into another row
+    // if theyoverflow.
+    final row2 = _textSize(
+      _metadataSpan('always a single line of text'),
+      width: width,
+    );
+    return row1.height + row2.height + _dividerHeight;
+  }
+
+  static Size _textSize(
```
kenzieschmoll, line 59: we already have a method for this in `ui/utils.dart`. We should use that instead `calculateTextSpanHeight`

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,117 @@
...
+    );
+  }
+
+  static const _detailsStyle = TextStyle();
+
+  static const _metaDataStyle = TextStyle();
+
+  static const double _dividerHeight = 10.0;
+
+  static double calculateRowHeight(LogDataV2 log, double width) {
```
kenzieschmoll, line 45: instead of having all these static methods on the widget class, I recommend putting these in the logging_model.dart file where they are used.

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,117 @@
...
+    final size = textPainter.size;
+    textPainter.dispose();
+    return size;
+  }
+}
+
+class _LoggingTableRowState extends State<LoggingTableRow> {
+  @override
+  Widget build(BuildContext context) {
+    Color? color = alternatingColorForIndex(
```
kenzieschmoll, line 76: use `var` instead

kenzieschmoll, line 76: use `var` instead

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,117 @@
...
+      widget.index,
+      Theme.of(context).colorScheme,
+    );
+
+    if (widget.isSelected) {
+      color = Colors.blueGrey;
+    }
+
+    return Container(
+      decoration: BoxDecoration(color: color),
```
kenzieschmoll, line 86: can you use the color attribute instead?

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,117 @@
...
+      child: ValueListenableBuilder(
```
kenzieschmoll, line 87: add type `ValueListenableBuilder<bool>`

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,117 @@
...
+      style: _detailsStyle,
+    );
+  }
+
+  static TextSpan _metadataSpan(String text) {
+    return TextSpan(
+      text: text,
+      style: _metaDataStyle,
+    );
+  }
```
kenzieschmoll, line 37: define these directly in the build method below instead of having them be static methods on the widget class. You can just use the TextSpan directly since having methods is just adding an unnecessary level of abstraction:

```
RichText(
  text: TextSpan(widget.data.asLogDetails()),
),

File: packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart [hunk 35]
```diff
@@ -0,0 +1,117 @@
...
+            children: [
+              RichText(
+                text: LoggingTableRow._detailsSpan(widget.data.asLogDetails()),
+              ),
+              Row(
+                children: [
+                  RichText(
+                    text: LoggingTableRow._metadataSpan('Some METADATA'),
+                  ),
+                  const SizedBox(width: 20.0),
```
kenzieschmoll, line 101: use devtools defined sizes like `defaultSpacing`

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
kenzieschmoll, line 43: also move this const to the state class

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,117 @@
...
+                  RichText(
+                    text: LoggingTableRow._metadataSpan('Goes Here'),
+                  ),
+                ],
+              ),
+              const Divider(
+                height: LoggingTableRow._dividerHeight,
+                color: Colors.black,
```
kenzieschmoll, line 109: can you use PaddedDivider here instead?

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,369 @@
...
+import 'logging_table_row.dart';
+
+/// A builder that includes an Offset to draw the context menu at.
+typedef ContextMenuBuilder = Widget Function(
+  BuildContext context,
+  Offset offset,
+);
+
+class LoggingTableV2 extends StatefulWidget {
+  /// Creates a screen that demonstrates the TableView widget.
```
kenzieschmoll, line 22: no need for dart doc on this constructor, but adding to the class would be helpful

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,369 @@
...
+  final LoggingTableModel model;
+
+  @override
+  State<LoggingTableV2> createState() => _LoggingTableV2State();
+}
+
+class _LoggingTableV2State extends State<LoggingTableV2> {
+  late final ScrollController _verticalController = ScrollController();
+  final Set<int> selections = <int>{};
+  final Map<int, double> cachedOffets = {};
```
kenzieschmoll, line 34: remove types on LHS in favor of types on RHS

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,369 @@
...
+  final normalTextStyle = const TextStyle(color: Colors.black, fontSize: 14.0);
```
kenzieschmoll, line 35: use theme.regularTextStyle instead of defining your own

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,369 @@
...
+  String lastSearch = '';
+  final metadataTextStyle = const TextStyle(
```
kenzieschmoll, line 37: same here. theme.regularTextStyle.copyWith(...) to apply the italic

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,369 @@
...
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Column(
+      children: [
+        Row(
+          children: [
+            Expanded(
+              child: TextField(
```
kenzieschmoll, line 69: use DevToolsClearableTextField for both of these

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,369 @@
...
+
+  final LoggingTableModel model;
+  final ScrollController _verticalController;
+
+  @override
+  State<_LoggingTableContents> createState() => _LoggingTableContentsState();
+}
+
+class _LoggingTableContentsState extends State<_LoggingTableContents> {
+  final Stopwatch _progressStopwatch = Stopwatch();
```
kenzieschmoll, line 117: remove type. on LHS

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,369 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return LayoutBuilder(
+      builder: (context, constraints) {
+        widget.model.tableWidth = constraints.maxWidth;
+        _progressStopwatch.reset();
+        return ListenableBuilder(
```
kenzieschmoll, line 125: use a ValueListenableBuilder so that we don't have to define a local variable for the value

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,369 @@
...
+}
+
+class _LoggingTableContents extends StatefulWidget {
+  _LoggingTableContents({
+    required this.model,
+    required ScrollController verticalController,
+  }) : _verticalController = verticalController;
+
+  static const _millisecondsUntilCacheProgressShows = 500;
+  static const _millisecondsUntilCacheProgressHelperShows = 2000;
```
kenzieschmoll, line 107: put these in the state class where they are used

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,369 @@
...
+                    const SizedBox(height: defaultSpacing),
+                    SizedBox(
+                      width: defaultLinearProgressIndicatorWidth,
+                      height: defaultLinearProgressIndicatorHeight,
+                      child: LinearProgressIndicator(
+                        value: progress,
+                      ),
+                    ),
+                    const SizedBox(height: defaultSpacing),
+                    if (_progressStopwatch.elapsedMilliseconds >
```
kenzieschmoll, line 156: this build method is not retriggered for updated to the stopwatch elapsed time is it?

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,369 @@
...
+                            ._millisecondsUntilCacheProgressHelperShows)
+                      const Text(
+                        'To make this process faster, then reduce the "Log Retention" setting.',
+                      ),
+                  ],
+                ),
+              );
+            } else {
+              _progressStopwatch.stop();
+              _progressStopwatch.reset();
```
kenzieschmoll, line 167: use cascades `..`

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,369 @@
...
+              return Scrollbar(
+                thumbVisibility: true,
+                controller: widget._verticalController,
+                child: ListenableBuilder(
```
kenzieschmoll, line 171: does the `SliverVariedExtentList` really only build the rows it needs? or does it build everything? Having the list be defined below the listenable builder makes me suspicious.

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,369 @@
...
+          },
+        );
+      },
+    );
+  }
+
+  @override
+  void dispose() {
+    super.dispose();
+    _progressStopwatch.stop();
```
kenzieschmoll, line 204: does the stopwatch object need disposed too?

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,369 @@
...
+  }
+}
+
+class _LoggingTableContextMenu extends StatelessWidget {
```
kenzieschmoll, line 208: can we pull all the context menu code out into a follow up PR? This seems like it will need an equally thorough review and maybe some discussion, so better to do in a separate PR.

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 The Chromium Authors. All rights reserved.
+// Copyight 2018 The Chromium Authors. All rights reserved.
```
kenzieschmoll, line 1: typo

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -297,3 +298,69 @@ Future<void> launchUrlWithErrorHandling(String url) async {
...
+      for (int indexIterator = chunkStartingIndex;
+          indexIterator < chunkUpperIndexLimit;
+          indexIterator++) {
+        // If our localWorkId is no longer active, then do not continue working
+        if (localWorkId != _workId) return completer.complete(false);
+        callback(indexIterator);
+      }
+
+      progressCallback(chunkUpperIndexLimit / length);
+      Future.delayed(const Duration(), () {
```
kenzieschmoll, line 352: use Duration.zero, and add a comment explaining why we are delaying a zero duration.

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
CoderDake: In a future iteration, the ability to select multiple logs will be added. With checkboxes on the left side likely.
Otherwise the user will be limited to actions that are just in the current view. Text selection on scrolling views means that some of the widgets disapears and cause buggy behaviour.

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
CoderDake: I put these here so I could hand bake the text styles.
The model and the table both need access to the text styles so that we can calculate the proper height.

Getting the BuildContext is kind of weird though outside of the widget tree.

Do you know of any better ways to just get the default textStyles without having to hand bake them?

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
CoderDake: These are added so that the same spans can be used in calculating the heights. I wanted to avoid duplicating these parts in case changes get made.

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
CoderDake: I'm not sure I understand the question. Could you please elaborate?

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
CoderDake: I'm not sure about best practices so I switched to using setState. Is that the better way to go about these things?

I can also do some testing if there is further doubt, but I remember in my demo that it was only loading a limited number of widgets at any time.

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
kenzieschmoll: We should raise the priority of text selection issues like this with the framework team. @Piinks @goderbauer do you know if text selection in scrolling views is something that is actively being worked on?

Piinks: I think it is, is there an open issue you can point us to?

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_controller_v2.dart
```diff
@@ -634,15 +627,21 @@ class LogDataV2 with SearchableDataMixin {
 
   String? get details => _details;
 
-  bool get needsComputing => detailsComputer != null;
+  ValueListenable<bool> get detailsComputed => _detailsComputed;
+  final ValueNotifier<bool> _detailsComputed = ValueNotifier<bool>(false);
```
kenzieschmoll, line 631: remove type on LHS

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_controller_v2.dart
```diff
@@ -634,15 +627,21 @@ class LogDataV2 with SearchableDataMixin {
...
 
   Future<void> compute() async {
-    _details = await detailsComputer!();
-    detailsComputer = null;
+    if (!detailsComputed.value) {
+      if (detailsComputer != null) {
+        _details = await detailsComputer!();
+      }
+      detailsComputer = null;
+      _detailsComputed.value = true;
```
kenzieschmoll, line 639: do we still want to set this to true even if `detailsComputer()` was never called?

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
kenzieschmoll: Looking at other areas in the code, we may just have to define it directly here. That being said, I think we can remove these 4 static members related to text style and text span, and just have the calculation methods below write:

`TextSpan(text: text, style: TextStyle(fontSize: defaultFontSize))`

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
kenzieschmoll: From what I can tell, we are not rebuilding this widget when the value of `_progressStopwatch.elapsedMilliseconds` changes. So how will this if statement be hit? Is this build method triggered again later on once we expect the stopwatch to have progressed?

CoderDake: Progress updates will refresh every chunk so this should get hit enough to be fresh.
Future changes can be made if people like to have huge retention limits

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
CoderDake: Note: look into whether the cache around the window needs to be shrunk.

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
CoderDake: I've posted a new issue for a different selection bug though.
https://github.com/flutter/devtools/issues/7893

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_controller_v2.dart
CoderDake: if we got to this point then any work that needed to be done is done, so I think this is fine :)

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_model.dart
```diff
@@ -97,11 +104,12 @@ class LoggingTableModel extends ChangeNotifier {
   }
 
   /// Get the height of a filtered Log at [index].
-  double getFilteredLogHeight(int index) {
+  double getFilteredLogHeight(BuildContext context, int index) {
     final cachedHeight = cachedHeights[index];
     if (cachedHeight != null) return cachedHeight;
 
     return cachedHeights[index] = LoggingTableRow.calculateRowHeight(
+      context,
```
kenzieschmoll, line 112: we should not be passing BuildContext around. Just rebuild the text style with the defaultFontSize instead of looking up the theme.

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,108 @@
...
+
+  static TextStyle get metadataStyle =>
+      Theme.of(DevToolsAppState.navigatorKey.currentContext!).subtleTextStyle;
+  static TextStyle get detailsStyle =>
+      Theme.of(DevToolsAppState.navigatorKey.currentContext!).regularTextStyle;
+
+  static final _padding = scaleByFontFactor(8.0);
+
+  static double calculateRowHeight(
+    BuildContext context,
```
kenzieschmoll, line 35: remove context

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,108 @@
...
+  final int index;
+  final LogDataV2 data;
+  final bool isSelected;
+
+  @override
+  State<LoggingTableRow> createState() => _LoggingTableRowState();
+
+  static TextStyle get metadataStyle =>
+      Theme.of(DevToolsAppState.navigatorKey.currentContext!).subtleTextStyle;
+  static TextStyle get detailsStyle =>
```
kenzieschmoll, line 29: new line between class members. Also, order static members before methods.

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -18,45 +18,38 @@ class LoggingTableRow extends StatefulWidget {
...
       maxWidth: width - _padding * 2,
     );
 
     // TODO(danchevalier): Improve row2 height by manually flowing metadas into another row
     // if theyoverflow.
     final row2Height = calculateTextSpanHeight(
-      _metadataSpan(context, 'always a single line of text'),
-      maxWidth: width,
+      TextSpan(text: text, style: metadataStyle),
+      maxWidth: width - _padding * 2,
```
kenzieschmoll, line 52: put `width - _padding * 2` in a const since it is used twice

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -85,22 +78,25 @@ class _LoggingTableRowState extends State<LoggingTableRow> {
               crossAxisAlignment: CrossAxisAlignment.start,
               children: [
                 RichText(
-                  text: LoggingTableRow._detailsSpan(
-                    context,
-                    widget.data.asLogDetails(),
+                  text: TextSpan(
+                    text: widget.data.asLogDetails(),
+                    style: Theme.of(context).regularTextStyle,
```
kenzieschmoll, line 83: use the details style from above

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -85,22 +78,25 @@ class _LoggingTableRowState extends State<LoggingTableRow> {
...
                 ),
                 Row(
                   children: [
                     RichText(
-                      text: LoggingTableRow._metadataSpan(
-                        context,
-                        'Some METADATA',
+                      text: TextSpan(
+                        text: 'Some META data',
+                        style: Theme.of(context).subtleTextStyle,
```
kenzieschmoll, line 91: use the meta data style from above. This will ensure consistency with what we are using and what we are measuring

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -85,22 +78,25 @@ class _LoggingTableRowState extends State<LoggingTableRow> {
...
                       ),
                     ),
                     const SizedBox(width: defaultSpacing),
                     RichText(
-                      text: LoggingTableRow._metadataSpan(context, 'Goes Here'),
+                      text: TextSpan(
+                        text: 'Some OTHER META data',
+                        style: Theme.of(context).subtleTextStyle,
```
kenzieschmoll, line 98: same

packages/devtools_app/lib/src/app.dart
```diff
@@ -130,8 +130,9 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
   final hoverCardController = HoverCardController();
 
   late ReleaseNotesController releaseNotesController;
+  static final navigatorKey = GlobalKey<NavigatorState>();
```
kenzieschmoll, line 133: new line between class members

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,110 @@
...
+  @override
+  Widget build(BuildContext context) {
+    var color = alternatingColorForIndex(
+      widget.index,
+      Theme.of(context).colorScheme,
+    );
+
+    if (widget.isSelected) {
+      color = Colors.blueGrey;
+    }
```
kenzieschmoll, line 68: this can be made more terse:
```
final color = widget.isSelected ? Colors.blueGrey : alternatingColorForIndex(...)
```

Also, use `selectedRowBackgroundColor` instead of blueGrey

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,110 @@
...
+              children: [
+                RichText(
+                  text: TextSpan(
+                    text: widget.data.asLogDetails(),
+                    style: Theme.of(context).regularTextStyle,
+                  ),
+                ),
+                Row(
+                  children: [
+                    RichText(
```
kenzieschmoll, line 88: since these row children are not wrapped in Flexible or Expanded, this will overflow when the screen is narrow.

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,309 @@
...
+class LoggingTableV2 extends StatefulWidget {
+  const LoggingTableV2({super.key, required this.model});
+
+  final LoggingTableModel model;
+
+  @override
+  State<LoggingTableV2> createState() => _LoggingTableV2State();
+}
+
+class _LoggingTableV2State extends State<LoggingTableV2> {
```
kenzieschmoll, line 32: can you add a TODO to use `SearchControllerMixin` and `FilterControllerMixin` for search and filter functionality in this table?

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,309 @@
...
+                onSubmitted: (value) {},
+              ),
+            ),
+            Expanded(
+              child: DevToolsClearableTextField(
+                labelText: 'Search',
+              ),
+            ),
+          ],
+        ),
```
kenzieschmoll, line 74: add a small amount of padding here using a SizedBox. I recommend denseSpacing or densePadding

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,309 @@
...
+            }
+          },
+        );
+      },
+    );
+  }
+
+  @override
+  void dispose() {
+    super.dispose();
```
kenzieschmoll, line 159: super.dispose should always be called last

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,309 @@
...
+  late final ScrollController _verticalController = ScrollController();
+
+  void onModelUpdate() {
+    setState(() {});
+  }
+
+  @override
+  void initState() {
+    widget.model.addListener(onModelUpdate);
+    super.initState();
```
kenzieschmoll, line 185: super.initState should always be called first

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,309 @@
...
+class _LoggingTableRowsState extends State<_LoggingTableRows> {
+  late final ScrollController _verticalController = ScrollController();
+
+  void onModelUpdate() {
+    setState(() {});
+  }
+
+  @override
+  void initState() {
+    widget.model.addListener(onModelUpdate);
```
kenzieschmoll, line 184: use the addAutoDisposeListener instead of manually adding and removing the listener. You will also be able to get rid of the onModelUpdate callback by doing this since this is the default when addAutoDisposeListener is not given a callback

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,309 @@
...
+  final Widget child;
+
+  @override
+  State<_ContextMenuRegion> createState() => _ContextMenuRegionState();
+}
+
+class _ContextMenuRegionState extends State<_ContextMenuRegion> {
+  Offset? _longPressOffset;
+
+  final ContextMenuController _contextMenuController = ContextMenuController();
```
kenzieschmoll, line 243: remove type on LHS

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,309 @@
...
+
+  static bool get _longPressEnabled {
+    switch (defaultTargetPlatform) {
+      case TargetPlatform.android:
+      case TargetPlatform.iOS:
+        return true;
```
kenzieschmoll, line 249: devtools is not ran on these platforms, so we can remove

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,309 @@
...
+
+  void _onLongPressStart(LongPressStartDetails details) {
+    _longPressOffset = details.globalPosition;
+  }
+
+  void _onLongPress() {
+    assert(_longPressOffset != null);
+    _show(_longPressOffset!);
+    _longPressOffset = null;
+  }
```
kenzieschmoll, line 277: long press is a mobile-focused gesture. Why are we using it here? We should be using right click or a three dot context menu since that is native for web.

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -0,0 +1,309 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return GestureDetector(
+      behavior: HitTestBehavior.opaque,
+      onSecondaryTapUp: _onSecondaryTapUp,
+      onTap: _onTap,
+      onLongPress: _longPressEnabled ? _onLongPress : null,
+      onLongPressStart: _longPressEnabled ? _onLongPressStart : null,
```
kenzieschmoll, line 305: remove

packages/devtools_app/lib/src/shared/utils.dart
kenzieschmoll: this comment remains

packages/devtools_app/test/logging/logging_screen_v2/logging_controller_v2_test.dart
```diff
@@ -17,12 +17,9 @@ import 'package:devtools_test/helpers.dart';
 import 'package:flutter_test/flutter_test.dart';
 
 void main() {
-  setUp(() {
-    setGlobal(MessageBus, MessageBus());
-  });
-
   group('LoggingControllerV2', () {
     late LoggingControllerV2 controller;
+    setGlobal(MessageBus, MessageBus());
```
kenzieschmoll, line 22: put this in the setup method below

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
CoderDake: In a future PR I will be handling metadata in more detail. I will ensure the tiles flow nicely.

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -72,6 +74,9 @@ class _LoggingTableV2State extends State<LoggingTableV2> {
             ),
           ],
         ),
+        const SizedBox(
+          height: denseSpacing,
```
kenzieschmoll, line 78: no need for trailing comma on sized box widgets. Helps to make build methods more verbose

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_v2.dart
```diff
@@ -156,13 +161,13 @@ class _LoggingTableProgressState extends State<_LoggingTableProgress> {
...
-    super.dispose();
     _progressStopwatch.stop();
+    super.dispose();
   }
 }
 
 class _LoggingTableRows extends StatefulWidget {
-  const _LoggingTableRows({
+  _LoggingTableRows({
     required this.model,
```
kenzieschmoll, line 171: no need for tc on dart code that will fit on one line

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -0,0 +1,109 @@
...
+    );
+    return row1Height + row2Height + _padding * 2;
+  }
+}
+
+class _LoggingTableRowState extends State<LoggingTableRow> {
+  @override
+  Widget build(BuildContext context) {
+    final color = widget.isSelected
+        ? Colors.blueGrey
```
kenzieschmoll, line 63: theme.colorScheme.selectedRowBackgroundColor

packages/devtools_app/lib/src/shared/utils.dart
```diff
@@ -334,7 +336,7 @@ class InterruptableChunkWorker {
     final completer = Completer<bool>();
     final localWorkId = ++_workId;
 
-    void doChunkWork(int chunkStartingIndex) {
+    void doChunkWork(int chunkStartingIndex) async {
```
kenzieschmoll, line 339: This should return `Future<void>` since it is async

packages/devtools_app/lib/src/screens/logging/logging_screen_v2/logging_table_row.dart
```diff
@@ -59,11 +59,12 @@ class LoggingTableRow extends StatefulWidget {
...
   @override
   Widget build(BuildContext context) {
+    final theme = Theme.of(context);
     final color = widget.isSelected
-        ? Colors.blueGrey
+        ? theme.colorScheme.selectedRowBackgroundColor
         : alternatingColorForIndex(
             widget.index,
-            Theme.of(context).colorScheme,
+            theme.colorScheme,
```
kenzieschmoll, line 67: nit: we can store the colorScheme in the local variable here instead of the theme to save a lookup

kenzieschmoll (approved):

PR #7831, opened 2024-05-27
Save/load tracing.

polina-c:


packages/devtools_app/lib/src/screens/memory/panes/tracing/tracing_data.dart
```diff
@@ -3,18 +3,28 @@
...
+import '../../../../shared/primitives/encoding.dart';
+import '../../../../shared/primitives/simple_items.dart';
 import '../../../../shared/primitives/utils.dart';
 import '../../../../shared/table/table_data.dart';
 import '../../../profiler/cpu_profile_model.dart';
 import '../../../profiler/cpu_profile_transformer.dart';
 
+@visibleForTesting
+enum TracedClassJson {
+  cls,
```
kenzieschmoll, line 21: nit: how about clazz instead so we can avoid abbreviated names

packages/devtools_app/test/test_infra/scenes/memory/default.dart
```diff
@@ -182,5 +182,10 @@ class MemoryDefaultScene extends Scene {
     await tester.pumpAndSettle();
   }
 
+  Future<void> goToTraceTab(WidgetTester tester) async {
```
kenzieschmoll, line 185: where is this used in this PR?

kenzieschmoll (approved):
just a couple comments

packages/devtools_app/test/test_infra/scenes/memory/default.dart
polina-c: It is not used. Will use it in next PR for testing.

PR #7837, opened 2024-05-28
Enable memory screen in disconnected mode.

polina-c:
We need to enable it before offline data mode to create test automation.
Tests cannot flip constant flags. So, we need disconnected mode to be without constant flag.


kenzieschmoll:
Tests can enable experiments. See example: https://github.com/flutter/devtools/blob/master/packages/devtools_app/test/performance/performance_screen_test.dart/#L138

polina-c:
Yes, if they are not constants :)
In this case enum ScreenMetaData requires constant flag that cannot be flipped by tests.

kenzieschmoll:
Then we should enable this in the same PR where tests are added. We are doing a beta release today and should not flip this flag to true without proper testing.

PR #7847, opened 2024-05-29
Cleanup UX for widget build counts feature

kenzieschmoll:
This PR:

- contains general UI polish (borders, padding, etc.)
- cleans up setting description and tooltip verbiage
Screenshot 2024-05-29 at 9 18 06 AM
- fixes the frame analysis tab so that widget rebuild counts are still shown when available, even when we do not have timeline events to build the frame hints and frame time views:
Screenshot 2024-05-29 at 9 32 14 AM
- fixes a bug causing the Rebuild Stats tab to be navigated away from on every frame selection


helin24 (approved):
Thanks!

PR #7863, opened 2024-05-30
Clean up usages of `late`

parlough:
Remove unnecessary ones and clean up a few others.

### DIFF
--- a/packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_controller_web.dart
+++ b/packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_controller_web.dart
@@ -99,7 +99,7 @@ class PerfettoControllerImpl extends PerfettoController {
   ///
   /// See [_viewIdIncrementer] for an explanation of why we use an incrementer
   /// in the id.
-  late final String viewId = 'embedded-perfetto-${_viewIdIncrementer++}';
+  late final viewId = 'embedded-perfetto-${_viewIdIncrementer++}';
 
   /// Url when running Perfetto locally following the instructions here:
   /// https://perfetto.dev/docs/contributing/build-instructions#ui-development
--- a/packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_controller_web.dart
+++ b/packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_controller_web.dart
@@ -99,7 +99,7 @@ class PerfettoControllerImpl extends PerfettoController {
   ///
   /// See [_viewIdIncrementer] for an explanation of why we use an incrementer
   /// in the id.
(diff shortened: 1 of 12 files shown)


packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_controller_web.dart
```diff
@@ -99,7 +99,7 @@ class PerfettoControllerImpl extends PerfettoController {
   ///
   /// See [_viewIdIncrementer] for an explanation of why we use an incrementer
   /// in the id.
-  late final viewId = 'embedded-perfetto-${_viewIdIncrementer++}';
```
kenzieschmoll, line 102: remove type on left hand side since it is inferred

packages/devtools_app/lib/src/shared/memory/class_name.dart
```diff
@@ -176,7 +176,7 @@ class HeapClassName with Serializable {
     return false;
   }
 
-  late final shortName =
+  late final String shortName =
```
kenzieschmoll, line 179: same here

packages/devtools_app/lib/src/shared/memory/class_name.dart
```diff
@@ -208,9 +208,9 @@ class HeapClassName with Serializable {
       (!library.startsWith(PackagePrefixes.dart) &&
           !library.startsWith(PackagePrefixes.genericDartPackage));
 
-  /// True, if the package has prefix `dart:` or has perfix `package:` and is
+  /// True, if the package has prefix `dart:` or has prefix `package:` and is
   /// published by Dart or Flutter org.
-  late final isDartOrFlutter = _isDartOrFlutter(library);
+  late final bool isDartOrFlutter = _isDartOrFlutter(library);
```
kenzieschmoll, line 213: remove type

kenzieschmoll (approved):
a few requests, but lgtm

packages/devtools_app/lib/src/screens/performance/panes/timeline_events/perfetto/_perfetto_controller_web.dart
parlough: This seems inconsistent across DevTools. Is this preference documented somewhere?

Many other Dart team projects will leave the type for fields since they end up being part of the class's API.

kenzieschmoll: This isn't documented anywhere that I am aware of, but this is a convention we have followed in the DevTools codebase for a long time. Mirroring more Dart projects as opposed to Flutter, which requires the types on the left and the right. The argument is that it makes the code unnecessarily verbose. 

We do have some style conventions documented [here](https://github.com/flutter/devtools/blob/master/STYLE.md), but preferring types on the RHS only where possible is not documented there (it probably should be).

Issue #7856, opened 2024-05-30
Ship DevTools with dart2wasm


kenzieschmoll:
The DevTools codebase is prepared for building DevTools with dart2wasm thanks to the unblocking work [here](https://github.com/flutter/devtools/issues/6606).

There are still a few requirements that we need to meet before moving this effort forward:

- [ ] Land the [bootstrapping changes](https://github.com/flutter/devtools/pull/7599) to DevTools in a way that does not break our g3 roll. 
  - CC @eyebrowsoffire @xster we will need support from the Fl@G team to fully support the new bootstrapping syntax. If adding full support is infeasible right now, we may have to workaround this by diverging the g3 bootstrapping code from the external. This is not ideal since this will cause a continued maintenance burden and potential regressions in g3, given we will not be testing externally with the same bootstrapping used in g3.
- [ ] Ensure that the VS Code embedded web views support WasmGC (see [background information on Chromium and V8](https://docs.flutter.dev/platform-integration/web/wasm#background))
  - @DanTup can you confirm?
- [ ] Ensure that the IntelliJ embedded web views (JXBrowser) support WasmGC (see [background information on Chromium and V8](https://docs.flutter.dev/platform-integration/web/wasm#background))
  - @helin24 can you confirm?
- [ ] Add support for opt in / opt out behavior. See these [web bootstrapping docs](https://docs.flutter.dev/platform-integration/web/bootstrapping#example-customizing-flutter-configuration-based-on-url-query-parameters) for an example of customizing configuration based on query parameters.

Compare the dart2wasm DevTools build to dart2js and file any issues that come up:
- TODO

kevmoo:
@kenzieschmoll – keep in mind, we can use query param magic to force loading certain versions. So if there are contexts where wasm won't work, we can just disable them. Or we can make loading the wasm version opt-in for testing, etc.

kevmoo:
Ha! just read your last bullet. 🤦

DanTup:
> Ensure that the VS Code embedded web views support WasmGC (see [background information on Chromium and V8](https://docs.flutter.dev/platform-integration/web/wasm#background))

The current version of VS Code uses Chromium v120 which sounds like it should support this. However, I don't know if there are things in Electron/VS Code themselves that might impact the user of this. @kenzieschmoll what's the easiest way for me to build DevTools using WASM (can I do it through `devtools_tool serve`? hacking scripts locally to make this work would be fine for testing).

[...]

PR #7688, opened 2024-05-01
Revert "Migrate to new bootstrapping methods. (#7599)"

elliette:
This reverts commit 694d552596761a94b1e3c330a003cef649b2cb3a.

We should hold off on submitting these changes until the template tokens are supported in google3.

PR #7865, opened 2024-05-30
Reland migration to use new flutter bootstrapping methods

kenzieschmoll:
This is a reland of https://github.com/flutter/devtools/pull/7599, which was reverted in https://github.com/flutter/devtools/pull/7688 because the new bootstrapping syntax is not supported in google3. This is still true, but I have prepared cl/638797386 to use the existing bootstrapping in google3 until the new syntax is supported.

Work toward https://github.com/flutter/devtools/issues/7856.

### DIFF
--- a/packages/devtools_app/web/flutter_bootstrap.js
+++ b/packages/devtools_app/web/flutter_bootstrap.js
@@ -1,0 +1,42 @@
+{{flutter_js}}
+{{flutter_build_config}}
+
+// Unregister the old custom DevTools service worker (if it exists). It was
+// removed in: https://github.com/flutter/devtools/pull/5331
+function unregisterDevToolsServiceWorker() {
+  if ('serviceWorker' in navigator) {
+    const DEVTOOLS_SW = 'service_worker.js';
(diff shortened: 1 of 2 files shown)


packages/devtools_app/web/flutter_bootstrap.js
```diff
@@ -0,0 +1,42 @@
...
+      serviceWorkerVersion: {{flutter_service_worker_version}},
+    },
+    config: {
+      canvasKitBaseUrl: 'canvaskit/'
+    }
+  });
+}
+
+// Bootstrap app for 1P environments:
+function bootstrapAppFor1P() {
```
elliette, line 37: Do we need this?

elliette (approved):
LGTM with question

packages/devtools_app/web/flutter_bootstrap.js
kenzieschmoll: I would opt to leave it so that things can work without needing to change anything once b/343786318 is resolved.

PR #7866, opened 2024-05-31
Enable prefer_final_in_for_each for consistency

parlough:
The repository already enables `prefer_final_locals` so for-loop declared iteration variables should match for consistency and to avoid accidental/potentially misunderstood assignments.

This matches `flutter/flutter` which [enables it as well](https://github.com/flutter/flutter/blob/master/analysis_options.yaml#L163).

### DIFF
--- a/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
+++ b/packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
@@ -687,7 +687,7 @@ class AppSizeController {
     int totalByteSize = 0;
 
     // Given a child, build its subtree.
-    for (Map<String, dynamic> child in rawChildren) {
+    for (final Map<String, dynamic> child in rawChildren) {
       final childTreemapNode = showDiff
           ? generateDiffTree(child, diffTreeType!)
           : generateTree(child);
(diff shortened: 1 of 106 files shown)


packages/devtools_app/lib/src/screens/app_size/app_size_controller.dart
```diff
@@ -721,7 +721,7 @@ class AppSizeController {
     }
     final childrenMap = <String, TreemapNode>{};
 
-    for (TreemapNode child in children) {
+    for (final TreemapNode child in children) {
```
kenzieschmoll, line 724: since we are using final now, can we remove the type on the left?

packages/devtools_app/lib/src/screens/inspector/inspector_tree_controller.dart
```diff
@@ -582,7 +582,7 @@ class InspectorTreeController extends DisposableController
     }
     final inlineProperties = parent.inlineProperties;
 
-    for (RemoteDiagnosticsNode property in inlineProperties) {
+    for (final RemoteDiagnosticsNode property in inlineProperties) {
```
kenzieschmoll, line 585: remove type here and elsewhere where possible

kenzieschmoll (approved):
a couple comments about types, but lgtm

PR #7896, opened 2024-06-07
Fix a regression with accessing the Flutter store file.

kenzieschmoll:
Add back access to the Flutter store file from devtools_shared. This regressed in https://github.com/flutter/devtools/commit/f4bc98d5c5c0a69fbbb25f170c4189804ded8851 and was causing errors when making requests to the DevTools server for the flutter GA values. See [discussion](https://github.com/flutter/devtools/issues/7856#issuecomment-2152964381) for more context.

### DIFF
--- a/packages/devtools_shared/lib/src/server/file_system.dart
+++ b/packages/devtools_shared/lib/src/server/file_system.dart
@@ -7,7 +7,7 @@ import 'dart:io';
 
 import 'package:path/path.dart' as path;
 
-import 'usage.dart';
+import 'devtools_store.dart';
 
 // ignore: avoid_classes_with_only_static_members, requires refactor.
 class LocalFileSystem {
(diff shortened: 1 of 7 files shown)


packages/devtools_shared/lib/src/server/file_system.dart
```diff
@@ -76,9 +76,95 @@ class LocalFileSystem {
...
+    }
+    syncSettings();
+  }
+
+  late File _file;
+
+  late Map<String, Object?> _map;
+
+  @override
+  // ignore: avoid-dynamic, necessary here.
```
bkonyi, line 118: Ubernit: can this not be `Object?`?

packages/devtools_shared/lib/src/server/file_system.dart
```diff
@@ -76,9 +76,95 @@ class LocalFileSystem {
...
+abstract class PersistentProperties {
+  PersistentProperties(this.name);
+
+  final String name;
+
+  // ignore: avoid-dynamic, dynamic by design.
+  dynamic operator [](String key);
+
+  // ignore: avoid-dynamic, dynamic by design.
+  void operator []=(String key, dynamic value);
```
bkonyi, line 162: Same here

packages/devtools_shared/lib/src/server/file_system.dart
```diff
@@ -76,9 +76,95 @@ class LocalFileSystem {
     return jsonEncode(json);
   }
 
+  /// The location of the Flutter store file, ~/.flutter.
+  static String flutterStoreLocation() {
```
bkonyi, line 80: Is this used anywhere else? If not, it might just be cleaner to keep it the way it was.

If you want to keep this factored out, it should probably be a static getter.

bkonyi (approved):
LGTM overall

DanTup (approved):
LGTM (w/ the nits Ben noted)!

PR #7907, opened 2024-06-11
Add missing dispose method in sidebar api and remove unused code.

kenzieschmoll:


packages/devtools_app/lib/src/standalone_ui/api/dart_tooling_api.dart
```diff
@@ -14,4 +14,6 @@ abstract interface class DartToolingApi {
   /// Access to APIs provided by VS Code and/or the Dart/Flutter VS Code
   /// extensions.
   Future<VsCodeApi?> get vsCode;
+
+  void dispose();
```
bkonyi, line 18: Nit: maybe annotate this with `@mustCallSuper`?

bkonyi (approved):

packages/devtools_app/lib/src/standalone_ui/api/dart_tooling_api.dart
kenzieschmoll: This is abstract so subclasses must implement this method and there is no method body in the superclass to call.

bkonyi: Is it worth adding an empty body so this annotation can be applied?

kenzieschmoll: Adding an empty body would allow subclasses to get away without overriding this method, which I don't think we want since this is written as a base class that another class is expected to implement. Is there a reason having an empty body with mustCallSuper would be better in your opinion?

bkonyi: I was mostly trying to figure out if there's a way to ensure `dispose()` is always called since I saw `super.dispose()` in one of your changes. However, I missed that was in the context of a `StatefulWidget` rather than an extension of `DartToolingApiImpl`. Disregard my comment.

DanTup: I presume the idea is that if a user has multiple level of classes, having `@mustCallSuper` would avoid them forgetting to call super. Arguably they could add `@mustCallSuper` to their own first implementation, but they'd have to remember.

One thing that might work would be to use add both `@mustCallSuper` _and_ `@mustBeOverridden` with an empty body? (if it's worth it.. I'm on the fence 🙂)

Issue #7911, opened 2024-06-11
Unify the summary tree and widget details tree


elliette:
Currently, the inspector contains two trees:

**Inspector summary tree**
* Labeled as the “Widget Tree” in the UI
* This tree shows only widgets that were created in the user’s project

**Inspector details tree**
* Labeled as “Widget Details Tree” in the UI
* This tree shows all the children widgets of the currently selected widget

The two trees can be unified, so that all the widgets are viewable in a single tree. As part of this, we can add styling distinguish between different types of widgets:

**Widgets that were created in the user’s project** 
* These are the widgets that are currently in the summary tree

**Widgets that were created in the framework or other external packages** 
* These widgets can be collapsed using an affordance like <8 more…>

PR #7912, opened 2024-06-11
[Inspector V2] Unify summary tree and widget details tree

elliette:
Work towards https://github.com/flutter/devtools/issues/7911

This PR uses the new API added in https://github.com/flutter/flutter/pull/150010 for the widget tree. 

For the V2 widget tree, it calls it with `isSummaryTree=false` to get all the widgets in a user's widget tree, not just those created in the project. The widgets that weren't created in the project are greyed out.

Screenshot 2024-06-11 at 11 31 55 AM

Follow-up: collapsing the widgets not created by the user's project


packages/devtools_app/lib/src/shared/diagnostics/inspector_service.dart
```diff
@@ -948,23 +948,46 @@ class ObjectGroup extends InspectorObjectGroupBase {
...
-        {'groupName': groupName},
-      ),
-    );
+  Future<RemoteDiagnosticsNode?> getRootWidgetTree({
+    required bool isSummaryTree,
+  }) {
+    const newApi = 'getRootWidgetTree';
+    final supportsNewApi =
+        WidgetInspectorServiceExtensions.values.asNameMap().containsKey(newApi);
+    if (supportsNewApi) {
```
kenzieschmoll, line 969: Since DevTools is tied to the Flutter SDK, we don't need to worry about backwards compatibility - we can just migrate to the new API. The CI will pass once the flutter candidate check picks up a version of Flutter that includes your new service extension change in flutter/flutter.

packages/devtools_app/lib/src/shared/console/widgets/description.dart
```diff
@@ -382,14 +382,16 @@ class DiagnosticsNodeDescription extends StatelessWidget {
         }
       }
 
-      if (!diagnosticLocal.isSummaryTree &&
-          diagnosticLocal.isCreatedByLocalProject) {
-        textStyle = textStyle.merge(DiagnosticsTextStyles.regularBold);
```
kenzieschmoll, line 387: should we still be bolding the nodes that are from the local project or do you think using the subtle text style for the others is enough distinction?

packages/devtools_app/test/shared/diagnostics/inspector_service_test.dart
```diff
@@ -264,92 +264,131 @@ void main() {
...
-        );
+          await group.dispose();
+        });
+
+        test('details tree', () async {
+          await env.setupEnvironment();
+          final group = inspectorService!.createObjectGroup('test-group');
+          final RemoteDiagnosticsNode root = (await group.getRoot(
+            FlutterTreeType.widget,
+            isSummaryTree: true,
```
kenzieschmoll, line 334: should this be false to test the details tree? If not can you add a comment to explain why?

kenzieschmoll (approved):

packages/devtools_app/lib/src/shared/console/widgets/description.dart
elliette: Ah you're right this shouldn't have been removed! This was bolding the local widgets in the **details** tree (not the summary tree). I've added this back in with a TODO to remove it once the details tree is gone.

packages/devtools_app/test/shared/diagnostics/inspector_service_test.dart
elliette: Added some comments :) This test is first getting a node in the summary tree, then using it to get the details tree for that node.

PR #7918, opened 2024-06-12
Refactor VS Code sidebar to make more generic and align with new DTD APIs

DanTup:
I'm working on supporting the Sidebar over DTD. The full in-progress change is at https://github.com/flutter/devtools/pull/7919 but it's getting quite large since the DTD API is slightly different (better) than the `postMessage` one and we should support both for some transition period (to handle old editor plugin versions). The plan is to build the new DTD APIs and then just provide wrappers around the `postMessage` API to make it the same (so the sidebar code can just work against one interface).

This PR takes all the non-DTD implementation from that PR which is basically a refactor of the `postMessage` classes, the sidebar (to use the new DTD-like APIs) and tests (again, to use the new DTD-like APIs). This will make the next PRs smaller because they're just adding the DTD implementation without changing the sidebar implementation, tests etc.

There are some differences in how the APIs work.. the new API will get `getDevices()` and then `deviceAdded` events, whereas the `postMessage` API just sends a `devicesUpdated` event with the full set of devices on every change (and always sends an initial set as part of initialization). The wrapper (`PostMessageEditorClient`) handles these differences.

A summary of the changes:

- `devtools_app/lib/src/editor/` contains the new API classes and interfaces for a client and server of the editor functionality
- Several of the old "DartToolingApi" classes have been renamed to have "PostMessage" in the name to avoid confusion when the DTD parts start landing (I will mark them as legacy when that happens)
- The new `PostMessageEditorClient` class is a wrapper over the postMessage API (`DartToolingApiImpl`, now `PostMessageToolApiImpl`) that implements the new DTD interface (including handling differences between how the events work, such as having no getDevices() and getting a full set of devices in each device event)
- Some extra layers and classes were flattened/removed from the API classes (for example the VsCodeDevice and VsCodeDebugSessions were removed and the replacement "EditorDevice" and "EditorDebugSessions" that are mostly the same are used in their places)
- Widget classes that previously used the old API class now use `EditorClient` (which currently is always the wrapper over postMessage, but can be a DTD implementation), and the Sidebar ("FlutterPanel") widget is updated to handle the new events
- Tests for debug_sessions and devtools extensions are updated to use a mock `EditorClient` in place of the old postMessage API
- The old mock API (`FakeDartToolingApi`) was replaced with a `FakeEditor` mixin that is agnostic to postMessage/DTD and provides the mock editor functionality for the mock editor stager app (and maybe can be used in more tests)


DanTup:
@kenzieschmoll @CoderDake I have a mostly-working sidebar over DTD (in the mock editor... haven't done any real Dart-Code side yet) in https://github.com/flutter/devtools/pull/7919 though it still needs some tweaks.

Since the change is quite large, I'm trying to split it up. This one adds the new DTD API interface, but doesn't use DTD yet. It just wraps the old `postMessage` API to behave like the new DTD one, allowing the sidebar code itself to be updated for the new API but still work with the existing `postMessage` API from VS Code. This is _essentially_ a no-op refactor, but it means implementing the DTD part will not involve changes to some of these tests and the sidebar UI code.

packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
```diff
@@ -82,43 +86,153 @@ class DartToolingApiImpl implements DartToolingApi {
...
+
+    // Trigger the initial initialization now we have the handlers set up.
+    // In the old postMessage world, this is how we get the initial set of
+    // devices/sessions.
+    unawaited(_api.initialize());
   }
 
-  @protected
-  final json_rpc_2.Peer rpc;
+  final VsCodeApi _api;
```
kenzieschmoll, line 165: is this code expected to be IDE agnostic in the future? Will there be one IdeApi that both VS Code and IntelliJ confirm to, or will we maintain two separate APIs - one for VS code and one for IJ?

packages/devtools_app/lib/src/service/editor/api_classes.dart
```diff
@@ -0,0 +1,418 @@
...
+  // Device.
+  deviceAdded,
+  deviceRemoved,
+  deviceChanged,
+  deviceSelected,
+
+  // Debug Session.
+  debugSessionStarted,
+  debugSessionChanged,
+  debugSessionStopped,
```
kenzieschmoll, line 20: can you add more dart doc here on what these events are for and also what their payload will look like if they come with arguments that are expected to be parsed?

packages/devtools_app/lib/src/service/editor/api_classes.dart
```diff
@@ -0,0 +1,418 @@
...
+  static const supported = 'supported';
+  static const vmServiceUri = 'vmServiceUri';
+}
+
+/// A base class for all known events that an editor can produce.
+///
+/// The set of subclasses is not guaranteed to match actual events from any
+/// given editor as the editor might not implement all functionality or may be a
+/// future version running against an older version of this code/DevTools.
+sealed class EditorEvent {
```
kenzieschmoll, line 51: can you mixin `Serializable`? and then you can remove the abstract override of toJson here?

https://github.com/flutter/devtools/blob/master/packages/devtools_shared/lib/src/utils/serialization.dart/#L11

packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
DanTup: It's not, this is the "legacy" API. In the complete PR that adds the DTD bits, this gets renamed to `PostMessageToolApi` (https://github.com/flutter/devtools/pull/7919/files#diff-10d85e415369fb2761370af4cc25dceab8b5f50bd4a432fdfd40bb6ee8310714). Some of that API is mapped onto the new DTD classes (`EditorDevice`, `EditorDebugSession`) in this PR, but the main interface of this is postMessage only and won't be used when communicating with DTD (the `EditorClient` class is essentially the replacement).

packages/devtools_app/lib/src/service/editor/api_classes.dart
DanTup: I've added a short comment to each with a reference to the class that the event goes with (which is the payload for such an event). Is this enough or do you think it needs more?

packages/devtools_app/lib/src/service/editor/api_classes.dart
DanTup: Done! Any objection to me changing `dynamic` to `Object?` in that mixin? I dislike `dynamic` because it's easy to pass where it shouldn't go, but `Object?` prevents that :-)

packages/devtools_app/lib/src/service/editor/api_classes.dart
```diff
@@ -0,0 +1,432 @@
...
+      other is DeviceRemovedEvent &&
+      other.runtimeType == runtimeType &&
+      other.deviceId == deviceId;
+
+  @override
+  int get hashCode => deviceId.hashCode;
+}
+
+/// An event sent by an editor when the current selected device was changed.
+/// This could be as a result of the client itself calling the `selectDevice`
```
kenzieschmoll, line 164: nit: new line after first sentence of dart doc

packages/devtools_app/lib/src/service/editor/api_classes.dart
```diff
@@ -0,0 +1,432 @@
...
+      };
+
+  @override
+  bool operator ==(Object other) =>
+      other is DebugSessionStartedEvent &&
+      other.runtimeType == runtimeType &&
+      other.debugSession == debugSession;
+
+  @override
+  int get hashCode => debugSession.hashCode;
```
kenzieschmoll, line 224: what is the reason we need to override equality and hash code for all of these? are we doing comparisons somewhere?

packages/devtools_app/lib/src/service/editor/api_classes.dart
```diff
@@ -0,0 +1,432 @@
...
+/// This could be as a result of the client itself calling the `selectDevice`
+/// method or because the device changed for another reason (such as the user
+/// selecting a device in the editor directly, or the previously selected device
+/// is being removed).
+class DeviceSelectedEvent extends EditorEvent {
+  DeviceSelectedEvent({required this.deviceId});
+
+  DeviceSelectedEvent.fromJson(Map<String, Object?> map)
+      : this(
+          deviceId: map[_Field.deviceId] as String?,
```
kenzieschmoll, line 173: why is this one nullable?

packages/devtools_app/lib/src/service/editor/api_classes.dart
```diff
@@ -0,0 +1,432 @@
...
+      other is DebugSessionStoppedEvent &&
+      other.runtimeType == runtimeType &&
+      other.debugSessionId == debugSessionId;
+
+  @override
+  int get hashCode => debugSessionId.hashCode;
+}
+
+/// A debug session running in the editor.
+class EditorDebugSession {
```
kenzieschmoll, line 289: mixin serializable

packages/devtools_app/lib/src/service/editor/api_classes.dart
```diff
@@ -0,0 +1,432 @@
...
+        vmServiceUri,
+        flutterMode,
+        flutterDeviceId,
+        debuggerType,
+        projectRootPath,
+      ]);
+}
+
+/// A device that is available in the editor.
+class EditorDevice {
```
kenzieschmoll, line 354: same here

packages/devtools_app/lib/src/service/editor/editor_client.dart
```diff
@@ -0,0 +1,56 @@
...
+  /// A stream of [EditorEvent]s from the editor.
+  Stream<EditorEvent> get event;
+
+  /// Gets the set of currently available devices from the editor.
+  Future<List<EditorDevice>> getDevices();
+
+  /// Gets the set of currently active debug sessions from the editor.
+  Future<List<EditorDebugSession>> getDebugSessions();
+
+  /// Requests the editor selects a specific device. It should not be assumed
```
kenzieschmoll, line 31: nit: new line after first sentence of dart doc

packages/devtools_app/lib/src/service/editor/editor_server.dart
```diff
@@ -0,0 +1,70 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'api_classes.dart';
+
+/// A base for classes that can act as an Editor (agnostic to the communication
+/// channel).
+abstract class EditorServer {
```
kenzieschmoll, line 11: The term "server" is a bit overloaded. Can we use something like `EditorBase` instead? TBH I am not completely understanding what an Editor Client is vs this class here, conceptually.

kenzieschmoll (commented):
General comment: it may be helpful to add a simple mermaid diagram to the editor/ folder that explains the concepts for this API

packages/devtools_app/lib/src/service/editor/api_classes.dart
kenzieschmoll: +1 to change to `Object?`.

packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
```diff
@@ -82,43 +86,153 @@ class DartToolingApiImpl implements DartToolingApi {
...
+        ..clear()
+        ..addAll(newDevices);
+
+      // Finally, handle if the selection changed.
+      if (e.selectedDeviceId != _currentSelectedDeviceId) {
+        _currentSelectedDeviceId = e.selectedDeviceId;
+        _eventController.add(
+          DeviceSelectedEvent(deviceId: _currentSelectedDeviceId),
+        );
+      }
```
kenzieschmoll, line 135: nit: can you put this handler into a private method. Same for below.

packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
```diff
@@ -82,43 +86,153 @@ class DartToolingApiImpl implements DartToolingApi {
...
+    // we must figure out what the actual changes are so we can produce the
+    // same kinds of events as the new DTD version.
+    _api.devicesChanged.listen((e) {
+      final supportedDevices = e.devices.map(
+        (d) => EditorDevice.fromJson({...d.toJson(), 'supported': true}),
+      );
+      final unsupportedDevices = e.unsupportedDevices?.map(
+            (d) => EditorDevice.fromJson({...d.toJson(), 'supported': false}),
+          ) ??
+          [];
```
kenzieschmoll, line 106: add type here

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
```diff
@@ -10,24 +10,25 @@ import 'package:flutter/material.dart';
...
+    required this.devices,
     super.key,
   });
 
-  final VsCodeApi api;
-  final List<VsCodeDebugSession> sessions;
-  final Map<String, VsCodeDevice> deviceMap;
+  final EditorClient editor;
+  final Map<String, EditorDebugSession> sessions;
+  final Map<String, EditorDevice> devices;
```
kenzieschmoll, line 31: can you add dart doc for these? What is the string representing in the map?

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
```diff
@@ -66,29 +68,47 @@ class _VsCodeFlutterPanelState extends State<VsCodeFlutterPanel> {
...
-      widget.api.debugSessionsChanged.listen((event) {
+      widget.editor.event.listen((event) {
         setState(() {
-          debugSessions = event.sessions;
+          switch (event) {
+            // Devices.
+            case DeviceAddedEvent(:final device):
+              devices[device.id] = device;
+            case DeviceChangedEvent(:final device):
+              devices[device.id] = device;
```
kenzieschmoll, line 99: can these two cases be combined?
```
case DeviceAddedEvent(:final device)
case DeviceChangedEvent(:final device): 
  devices[device.id] = device;
```

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
```diff
@@ -66,29 +68,47 @@ class _VsCodeFlutterPanelState extends State<VsCodeFlutterPanel> {
...
+            case DeviceRemovedEvent(:final deviceId):
+              devices.remove(deviceId);
+            case DeviceSelectedEvent(:final deviceId):
+              selectedDeviceId = deviceId;
+            // Debug sessions.
+            case DebugSessionStartedEvent(:final debugSession):
+              debugSessions[debugSession.id] = debugSession;
+            case DebugSessionChangedEvent(:final debugSession):
+              debugSessions[debugSession.id] = debugSession;
```
kenzieschmoll, line 108: same question here

packages/devtools_app/test/test_infra/scenes/standalone_ui/mock_editor_widget.dart
```diff
@@ -8,33 +8,29 @@ import 'dart:convert';
...
 
-/// A simple UI that acts as a stand-in host IDE to simplify the development
-/// workflow when working on embedded tooling.
-///
-/// This UI interacts with [FakeDartToolingApi] to allow triggering events that
-/// would normally be fired by the IDE and also shows a log of recent requests.
-class VsCodeFlutterPanelMockEditor extends StatefulWidget {
-  const VsCodeFlutterPanelMockEditor({
+/// A simple UI that acts as a stand-in host editor to simplify the development
+/// workflow when working on embedded tooling. Uses a [FakeEditor] to provide
```
kenzieschmoll, line 14: nit: new line after first sentence

packages/devtools_app/test/test_infra/test_data/editor_service/fake_editor.dart
```diff
@@ -0,0 +1,177 @@
+// Copyright 2024 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:devtools_app/src/service/editor/api_classes.dart';
+import 'package:devtools_app/src/service/editor/editor_server.dart';
+
+/// An mixin for [EditorServer]s that provides some useful mock editor
```
kenzieschmoll, line 10: nit: A mixin

kenzieschmoll (commented):
Done with first pass

packages/devtools_app/lib/src/service/editor/api_classes.dart
```diff
@@ -0,0 +1,432 @@
...
+  static const vmServiceUri = 'vmServiceUri';
+}
+
+/// A base class for all known events that an editor can produce.
+///
+/// The set of subclasses is not guaranteed to match actual events from any
+/// given editor as the editor might not implement all functionality or may be a
+/// future version running against an older version of this code/DevTools.
+sealed class EditorEvent with Serializable {
+  EditorEventKind get event;
```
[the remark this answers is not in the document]
DanTup: Yes, thanks! Fixed in 847045273

packages/devtools_app/lib/src/service/editor/api_classes.dart
DanTup: Done in 8ec9692b9

packages/devtools_app/lib/src/service/editor/editor_client.dart
DanTup: Done in 8ec9692b9

packages/devtools_app/lib/src/service/editor/editor_server.dart
DanTup: I've expanded the comment here - is it clearer now?

Basically, the "Server" part is for being the editor (eg. what VS Code and IntelliJ would do) and the "Client" part is for connecting to DTD and consuming the editors services.

I added "Server" here just to try and make it a bit clearer (since "Editor" is quite vague). In this repo, we will use this class for testing and for the "Mock Editor" stager app. In theory this class could live elsewhere, but I thought it would be nice to keep the client/server together since their interfaces must match and maybe in future if we break `EditorClient` out so it could be used by other tools, it may make sense to do the same for `EditorServer` (for ex. if you're writing an editor that wants to provide services to host the sidebar).

packages/devtools_app/lib/src/standalone_ui/vs_code/debug_sessions.dart
DanTup: Done in 8ec9692b9 (the keys are the IDs).

packages/devtools_app/test/test_infra/scenes/standalone_ui/mock_editor_widget.dart
DanTup: Done in 8ec9692b9

packages/devtools_app/test/test_infra/test_data/editor_service/fake_editor.dart
DanTup: Done in 8ec9692b9

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
DanTup: Yep, done in 401baf04e

packages/devtools_app/lib/src/standalone_ui/vs_code/flutter_panel.dart
DanTup: Done in 401baf04e

packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
DanTup: Done in 401baf04e

packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
DanTup: Done in 0f63b7d59

packages/devtools_app/lib/src/service/editor/api_classes.dart
DanTup: Done in 256056b34

packages/devtools_app/lib/src/service/editor/api_classes.dart
DanTup: Done in 256056b34

packages/devtools_app/lib/src/service/editor/api_classes.dart
DanTup: It's possible to unselect the active device (for example, you have only one device and disconnect it in the editor).

I added a comment in f168e1d73 to make this clearer.

packages/devtools_app/lib/src/service/editor/api_classes.dart
DanTup: Good question. We're not currently, but I started working on some integration tests for this in the DTD version (they're not in the other PR yet because I had issues making them work, because tests run in flutter_tester here and I was trying to spawn DTD).

I could remove them here and re-add them with the tests later?

packages/devtools_app/lib/src/service/editor/api_classes.dart
DanTup: Done in 7199e0d98

DanTup:
@kenzieschmoll I also added [a readme](https://github.com/DanTup/devtools/blob/7dadfc3094597fbf80cc5c04e1007c0e33a6bc0e/packages/devtools_app/lib/src/service/editor/readme.md) with some notes to try and make things a little clearer.

packages/devtools_app/lib/src/service/editor/editor_server.dart
DanTup: Seems reasonable to me. It felt nice having them together, but I agree it's a little weird having code we don't expect to use inside DevTools in there.

I had to move a few other files too from test_infra/test_data into test_infra/scenes/standalone_ui with this, as I discovered the Stager app had stopped working in Chrome. Seems like desktop can import in test_infra/scenes/... from test_infra/test_data, but web cannot (possibly because it's outside what it considered the project root?).

packages/devtools_app/lib/src/service/editor/api_classes.dart
kenzieschmoll: Sure, let's remove here and see if we need them later 👍

DanTup: Done! (7044c02e3)

kenzieschmoll (approved):

PR #7943, opened 2024-06-18
Add API fields to allow passing requiresDebugSession to openDevTools

DanTup:
This adds a new capability and parameter for openDevTools that allows DevTools to indicate to the editor that a call requires a debug session.

This allows calling openDevTools() without a debug session ID and letting the editor prompt so that things like Inspector could be top-level in the sidebar, but also supports having static tooling at the top level where this prompting is not desired.

If the `openDevToolsWithRequiresDebugSessionFlag` capability is `false`, this flag will not work and the behaviour (of whether an editor prompts or not) is not well defined, but for existing VS Code will prompt only for known DevTools pages that are not-static (that is, DevTools extensions will never prompt and always be assumed to be static if an explicit debug session is not passed).

This does not make any changes to the UI to use this functionality.

### DIFF
--- a/packages/devtools_app/lib/src/service/editor/editor_client.dart
+++ b/packages/devtools_app/lib/src/service/editor/editor_client.dart
@@ -45,6 +45,7 @@ abstract class EditorClient {
     String? debugSessionId, {
     String? page,
     bool? forceExternal,
+    bool? requiresDebugSession,
   });
 
   /// Requests the editor enables a new platform (for example by running
--- a/packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
+++ b/packages/devtools_app/lib/src/standalone_ui/api/impl/dart_tooling_api.dart
@@ -211,11 +211,13 @@ class PostMessageEditorClient implements EditorClient {
     String? debugSessionId, {
     String? page,
     bool? forceExternal,
+    bool? requiresDebugSession,
   }) async {
     await _api.openDevToolsPage(
       debugSessionId,
       page: page,
       forceExternal: forceExternal,
+      requiresDebugSession: requiresDebugSession,
     );
   }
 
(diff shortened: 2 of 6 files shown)


DanTup:
@kenzieschmoll I've pushed a pre-release of Dart-Code that includes this capability and the related changes (and ofc the current stable does _not_ have it).

kenzieschmoll (approved):

kenzieschmoll:
@DanTup after testing these changes out, we also need a field 'prefersDebugSession'. For example, the Deep Links page does not technically require a debug session, since the user can manually enter their project path or select from a list of possible roots. However, if a user has an active debug session, then opening the tool against the current debug session is preferred. This is the same pattern for other screens like Cpu Profiler and Performance, which don't require a connection (you can load offline data into them), but a connection is preferred if there are debug sessions available.

DanTup:
> However, if a user has an active debug session, then opening the tool against the current debug session is preferred.

What if there are multiple debug sessions active? Would we prompt for one, or just one arbitrarily? (perhaps the one selected by VS Code?)

DanTup:
> This is the same pattern for other screens like Cpu Profiler and Performance, which don't require a connection (you can load offline data into them), but a connection is preferred if there are debug sessions available.

Do we want to handle these by just having the same single button in the sidebar? It could seem odd if a single "CPU Profiler" button sometimes shows the offline page (when no session) and sometimes shows a live page (when there is a session). Users might not understand this difference (for example you might imagine you could start the CPU profile just before starting the debug session). Might it be better to separate these on the sidebar in some way (and pass the `requiresDebugSessions` flag as `true`/`false` accordingly)?

DanTup:
I've opened with https://github.com/flutter/devtools/pull/7949 for this (and detailed the behaviour I went with for now - which we can easily change).

PR #7949, opened 2024-06-19
Add "prefersDebugSession" + rename capabilities flag

DanTup:
This adds a new flag `prefersDebugSession` similar to the one added yesterday `requiresDebugSession` to better handle tools that should have debug sessions if they exist but can also be launched without.

Since the change landed yesterday has not shipped, I renamed the capabilities flag to cover both options.

The editor behaviour I have implemented currently (but am happy to tweak is) that if this is flag is true then:

- if there are no debug sessions, launch without
- if there is one debug session, launch with that
- if there are multiple debug sessions, prompt the user

We could just use the "active debug session" (according to VS Code) for the last one, but since we don't do that for other things yet (we always prompt) and we're not showing the "active" debug session in the sidebar yet, I think prompting makes most sense. In most cases, probably users don't have multiple debug sessions anyway.

@kenzieschmoll I will push a new Dart-Code pre-release with those changes once this merges. I don't want to do it now in case there is feedback or you're still working against the already-merged code (since the capability flag is renamed).

### DIFF
--- a/packages/devtools_app/lib/src/service/editor/editor_client.dart
+++ b/packages/devtools_app/lib/src/service/editor/editor_client.dart
@@ -46,6 +46,7 @@ abstract class EditorClient {
     String? page,
     bool? forceExternal,
     bool? requiresDebugSession,
+    bool? prefersDebugSession,
   });
 
   /// Requests the editor enables a new platform (for example by running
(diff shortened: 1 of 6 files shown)


kenzieschmoll (approved):

DanTup:
@kenzieschmoll I've published an updated pre-release of Dart-Code (3.91.20240620) that includes the Dart-Code side of this.

Issue #7879, opened 2024-06-03
Memory/Network Page in blank (DevTools)


JeffreyPSA:
When I try to load the memory page or network page from vs code I just get a blank page as shown below:
![image](https://github.com/flutter/devtools/assets/114021588/4b43e601-2a3c-4830-92d0-77a8f27d7a3e)
Also if I try to load the devtools from my web browser (Chrome), the performance tab only shows me this info below
![image](https://github.com/flutter/devtools/assets/114021588/94b65d31-729e-49e3-8045-6864adb5e887)
This is a web app and I'm trying running it in debug mode.
My flutter version is 3.22.1 devtools version is 2.34.3

I need critically check what's going on with the memory performance of this app, but devtools is not working as an shown.

kenzieschmoll:
For Flutter web apps, the following tools are supported by Dart DevTools: Inspector, Logging, and Debugger. For all other needs like Performance, Network, or Memory, you should use Chrome DevTools from the Chrome window your web app is running in.

It looks like we are not properly disabling the DevTools screens that are not supported for web from VS code, which is causing some confusion. @DanTup 

CC @helin24 since we will need to use the `hide=`  param in the IntelliJ tool window URI to hide the unsupported DevTools screens when connected to a web app.

DanTup:
> It looks like we are not properly disabling the DevTools screens that are not supported for web from VS code, which is causing some confusion. @DanTup

We don't have a great way in VS Code to know all the conditions in which we should enable/disable pages (because users can specify their own flags in a run configuration that could target different devices, modes, etc. or use custom launch scripts). I assume these conditions could (in the future) change between SDK versions.

The most reliable thing may be to have some kind of API (via the server, or a service extension, or DTD) that would allow us to determine which pages are available. Although, in the past I thought DevTools showed a message like "this page isn't available for x" that doesn't appear to be working here - would fixing that be good enough so it's at least clear (in some ways, this might be better than just hiding all of the commands, because a use might wonder why they can't find the command to launch a given page).


> CC @helin24 since we will need to use the hide= param in the IntelliJ tool window URI to hide the unsupported DevTools screens when connected to a web app.

Doesn't DevTools already hide the tabs that aren't supported? `hide=` was originally intended only to hide things that otherwise would be supported (for ex. Debugger). VS Code doesn't use `hide=` for anything else and expected DevTools to only show what's available.

helin24:
[...]

PR #7958, opened 2024-06-20
Show a screen unavailable message for disabled screens.

kenzieschmoll:
When a first party DevTools screen is loaded but is not available for the current app, we now show a message explaining why. Before we just silently navigated to a different screen or showed a blank page when embedded.

Now, we show this:
Screenshot 2024-06-20 at 3 40 27 PM

Embedded mode (hide the navigate home button in this case):
Screenshot 2024-06-20 at 3 39 21 PM

Fixes https://github.com/flutter/devtools/issues/7879


packages/devtools_app/lib/src/app.dart
```diff
@@ -475,6 +494,25 @@ class DevToolsAppState extends State<DevToolsApp> with AutoDisposeMixin {
...
+          _originalScreens.firstWhere((s) => s.screenId == screenMetaData.id);
+      final reason = shouldShowScreen(screen).disabledReason;
+      if (reason == ScreenDisabledReason.requiresDartLibrary) {
+        // Special case for screens that require a library since the message
+        // needs to be generated dynamically.
+        disabledReason = 'The ${screen.title} screen requires library '
+            '${screen.requiresLibrary}, but the library was not detected.';
+      } else if (reason?.message != null) {
+        disabledReason = 'The ${screen.title} screen ${reason!.message!}';
+      }
```
DanTup, line 512: Not important, but FWIW using `case String` is a nice way to avoid the `!`s here:

```dart
if (reason?.message case String message) {
  disabledReason = 'The ${screen.title} screen $message';
}
```

DanTup (approved):
LGTM!

CoderDake (approved):

packages/devtools_app/lib/src/app.dart
kenzieschmoll: Nice! will add to a follow up PR.

DanTup: Oh and you can avoid writing the type (`String`) again like this:

```dart
if (reason?.message case final message?) {
```

